PR #13, opened 2020-12-19
fix name and logic for AWS CIS 1.10

zscholl:
CIS 1.10 is titled `Ensure IAM password policy prevents password reuse` and the benchmark from AWS should pass as long as there are fewer than 24 remembered passwords.

https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-cis-controls.html#securityhub-cis-controls-1.10


yevgenypats:
Awesome, merged. Thanks!

PR #25, opened 2021-01-03
Apply ...zapcore.Option during logger creation

dan-compton:
Prior to this fix, functional options were dropped in the constructor `NewLogger`.  Now they are applied to the logger post-configuration.  Note that verbose could be implemented as a zap.Option to create a uniform interface.


yevgenypats:
Thanks for the fix!

PR #32, opened 2021-01-10
Add support for CQ_ env variable via viper

yevgenypats:
This adds support for environment variables with `CQ_` prefix to cloudquery.

Currently supports:

```
CQ_DSN
CQ_DRIVER
CQ_CONFIG_PATH
CQ_QUERY_PATH
```


yevgenypats:
@zscholl I've added support for environment variables via viper/cobra integration. Hopefully this will help to lambda deployment as well.

PR #43, opened 2021-01-26
Add role and user tags

mvaerle:
Instead of just raising issues I thought it would be nice to try and make a PR. Note that this is my first interaction with Go, so treat this code with caution and feel free to change it.

I experienced an issue where tags are not added to the database for both iam users and roles. Even though the tables  `aws_iam_user_tags` and `aws_iam_role_tags` are created, these tables never get populated.

The code looked correct, and even the API documentation and the aws go sdk specify that tags should be returned when calling `ListRoles` and `ListUsers`, but they never are.

I've found that this is a known issue in the AWS API https://github.com/boto/boto3/issues/1855. The issue has been open for a long time with no end in sight.

I've now implemented another call to `ListUserTags` and `ListRoleTags` to fetch the tags for every user and role. This does add another call per user/role, but I don't see any way around that if you want to access the tags. I've verified on our AWS account that the tables are now populated!

### DIFF
--- a/providers/aws/iam/roles.go
+++ b/providers/aws/iam/roles.go
@@ -1,11 +1,12 @@
 package iam
 
 import (
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/mitchellh/mapstructure"
 	"go.uber.org/zap"
-	"time"
 )
 
 type Role struct {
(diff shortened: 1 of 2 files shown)


yevgenypats:
@mvaerle AMAZING! The code looks good and adheres to the current style. Thanks for finding this weird issue with AWS SDK and fixing. merging and pushing a new release.

yevgenypats:
https://github.com/cloudquery/cloudquery/releases/tag/v0.8.9

PR #45, opened 2021-01-28
Use account authorization details to fetch policy documents

mvaerle:
I wanted to fetch the actual policy documents so that I can query the contents of them, for example filter on all policies that have AssumeRole in them. Since we have 17k+ policies in our accounts doing a call for each of them is probably not practical.

The solution for this the `get-account-authorization-details` call that allows you to fetch all the users, groups, roles and policies of an account in a single call. (https://docs.aws.amazon.com/cli/latest/reference/iam/get-account-authorization-details.html)

I've implemented this in this PR for just the policies call, since all the fields matched perfectly, but of course it can be used to replace (parts of) the users, groups and roles implementations too. It will probably make it faster. The issue is there are slight differences in the fields returned. For example for the policies call it returns more fields, but for others resources it returns less, so this will take a bit more time and research.

I've also implemented urldecode for the contents of documents to make it easier to parse.

Let me know what you think!

### DIFF
--- a/providers/aws/iam/policies.go
+++ b/providers/aws/iam/policies.go
@@ -1,11 +1,13 @@
 package iam
 
 import (
+	"net/url"
+	"time"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/mitchellh/mapstructure"
 	"go.uber.org/zap"
-	"time"
 )
 
 type Policy struct {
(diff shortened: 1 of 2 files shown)


yevgenypats:
@mvaerle Thanks! I think this is definitely the right direction. I've one questions/suggestion. Currently the document is stored as string - maybe it makes sense parse it also into tables so it will be easier to query. The JSON is well structured and has well-defined fields so it should be possible. What do you think?

mvaerle:
@yevgenypats That would be really cool. I assumed there was too much complexity in there to convert it to a relational structure. I just took a quick look at the grammar (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html) and I think you are right that it's possible to parse it into tables.

yevgenypats:
Sounds good. Do you want me to open a new PR or do you like to add this to yours?

providers/aws/iam/policies.go
```diff
@@ -22,14 +24,49 @@ type Policy struct {
...
+
+func (PolicyVersion) TableName() string {
+	return "aws_iam_policy_versions"
+}
+
+func (c *Client) transformPolicyVersionList(values []*iam.PolicyVersion) ([]*PolicyVersion, error) {
+	var tValues []*PolicyVersion
+
+	for _, value := range values {
+		decodedDocument, err := url.QueryUnescape(*value.Document)
```
yevgenypats, line 52: I think there should be a check here `if value.Document != nil` and initialise `decodedDocument` with with `nil` before that. Otherwise the program may panic (theoretically as I'm not sure if it possible for the document to be empty) .

yevgenypats:
@mvaerle I looked into this. Let's do this in two steps. First I'll merge this PR (I only had one small comment there) and then I'll add the improvement of parsing the document to a table.

mvaerle:
@yevgenypats I added a nil check, I'm not quite sure if this is the most elegant, but I'm new to go so I don't know anything better ;-)

Issue #15, opened 2020-12-21
Throttling: Rate exceeded, status code: 400


mvaerle:
Hi,

I'm trying to use cloudquery to extract some resources for compliance purposes. I'm fetching IAM data for 23 accounts, so including all policies, users, etc.

After a seemingly random amount of time but always within a couple of minutes, I get the following error:

```
2020/12/21 15:34:19 Throttling: Rate exceeded
	status code: 400, request id: [left this one out]
```

UPDATE:
Just did a bit more research and a workaround is to comment out the accounts in the config.yml and then run fetch in batches of ~3 accounts.

Using cloudquery version 0.4.3

sethcenterbar:
Typically, AWS will work with you to raise the soft limits on your accounts. That's likely the best way to tackle this issue if you need to query all of the accounts at once on a regular basis. I don't believe anything could be done on the cloudquery side other than doing client side throttling.

yevgenypats:
Thanks @mvaerle and @sethcenterbar I believe raising the AWS soft limit might work indeed.
I can also try and add a configurable option to increase the number of retries done by AWS SDK.
Also, I'll add a logging option so you can rerun it and we can see on which retry you get the error. 
As a last resort I guess we will need to implement client side throttling.

mvaerle:
Thanks for the quick reply!

Asking AWS to raise limits might work, but I'm not even sure which limits I'm hitting and by how much. We're never running into these except when using cloudquery. I'm trying to get some insight through cloudtrail but no luck so far, I'll report back if I can find something.

Once there is logging available I'm happy to test and try again. For now I can use the workaround of commenting out accounts and/or contact AWS to raise our limits.

yevgenypats:
@mvaerle I've added `log_level: debug` and `max_retries: 20` to your AWS `config.yml` as specified [here](https://docs.cloudquery.io/aws/tables-reference).

This should show us both what limit is being reached and also if changing `max_retries` to a higher number helps or not?

Those keywords are available at [v0.5.3](https://docs.cloudquery.io/aws/tables-reference)

dan-compton:
1. It's good to gather data on your API usage which can be done with cloudtrail and an ELK stack.  I think there's a one-click cfn link solution out there.

[...]

PR #46, opened 2021-01-29
add AWS multi account.id fetch capability

Rackme:
I wanted to be able to fetch all my accounts at the same time with such config.yml :
```
providers:
  - name: aws
    accounts: # Optional. if you want to assume role to multiple account and fetch data from them
      - id: production
      - id: sharedstorage
      - id: test-us
      - id: test-eu
      - id: test-rds
      - id: prod-kms
      - id: uballot
      - id: mmkpm
```

Working like a charm, however with sqlite when fetching the eighth account I've encountered io limits or concurrency issue. I will open a proper issue for this.

### DIFF
--- a/providers/aws/provider.go
+++ b/providers/aws/provider.go
@@ -210,7 +210,8 @@ func (p *Provider) Run(config interface{}) error {
 	for _, account := range p.config.Accounts {
 
 		if account.ID != "default" && account.RoleARN != "" {
-			// assume role if different account
+			// assume role if specified (SDK takes it from default or env var: AWS_PROFILE)
+			p.session, err = session.NewSession()
 			cred := stscreds.NewCredentials(p.session, account.RoleARN)
 			p.session, err = session.NewSession(&aws.Config{
 				Credentials: cred,
--- a/providers/aws/provider.go
+++ b/providers/aws/provider.go
@@ -2,6 +2,10 @@ package aws
 
 import (
 	"fmt"
+	"log"
+	"strings"
+	"sync"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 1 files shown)


yevgenypats:
@Rackme Thanks for creating this. can you try and use `role_arn:` instead of `id:` I think this should work.

Regarding limits see this discussion and possible solution - https://github.com/cloudquery/cloudquery/issues/15#issuecomment-754896666

Rackme:
Hello @yevgenypats,

Using profiles' name in `role_arn:` would return :
```
2021/01/29 14:21:14 InvalidParameter: 1 validation error(s) found.
- minimum field size of 20, AssumeRoleInput.RoleArn.
```

Indeed the aws sdk would expect a role (ex: _arn:aws:iam::123456789:role/OrganizationAccessRole_ ) with `stscreds.NewCredentials(p.session, account.RoleARN)` to assume. 
I'm here in the case where no role is used, only one account per Organization.

More over even when using a valid role i'm getting :
`2021/01/29 14:28:25 NoCredentialProviders: no valid providers in chain. Deprecated.`

Regarding the concurrency, i'm not really sure this is about the max_retries fix, i'm already using this, please find the associated stack trace :
[stack_trace.log](https://github.com/cloudquery/cloudquery/files/5894327/stack_trace.log)

providers/aws/provider.go
```diff
@@ -207,11 +208,8 @@ func (p *Provider) Run(config interface{}) error {
...
 	for _, account := range p.config.Accounts {
-		p.session, err = session.NewSession()
-		if err != nil {
-			return err
-		}
-		if account.ID != "default" {
+
+		if account.ID != "default" && account.RoleARN != "" {
 			// assume role if different account
 			cred := stscreds.NewCredentials(p.session, account.RoleARN)
```
yevgenypats, line 214: @Rackme looks like `p.session` will be `nil` here. Does AWS SDK support it?

PR #50, opened 2021-01-29
fix: query --output should write file as list of json objects

### LINKED ISSUE #48: Request for query command to optionally return output in a machine readable format e.g. json/yaml
For anyone looking to utilize the policy packs programmatically or in CI, it would be useful to be able to have the output from the `query` command be in format more easily parsed, such as json or yaml.

James-Quigley:
Instead of writing each query result as an individual object line by line, write the whole thing at once as a list.

Previously the output would have been written to look like:
```
{}
{}
```
which isn't valid JSON. Now it will be written as:
```
[
  {},
  {}
]
```

This should fix #48

### DIFF
--- a/cloudqueryclient/client.go
+++ b/cloudqueryclient/client.go
@@ -182,10 +182,15 @@ func (c *Client) runQueries(config *PolicyConfig, outputPath string) error {
 		defer f.Close()
 	}
 
+	_, err = f.WriteString("[")
+	if err != nil {
+		return err
+	}
+
 	c.log.Info("Executing queries", zap.Int("count", len(config.Queries)))
-	for _, query := range config.Queries {
+	for idx, query := range config.Queries {
 		queryResult := QueryResult{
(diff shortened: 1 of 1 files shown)


yevgenypats:
@James-Quigley Thanks for this. I have one suggestion/ask - accumulating all the results before writing to the file might be an unnecessary memory consumption (it might get big depending on the query). Currently it writes the result after each iteration, I suggest keeping it the same way but just writing a `[` at the beginning and `,` after each row (apart from the last one).

James-Quigley:
> @James-Quigley Thanks for this. I have one suggestion/ask - accumulating all the results before writing to the file might be an unnecessary memory consumption (it might get big depending on the query). Currently it writes the result after each iteration, I suggest keeping it the same way but just writing a `[` at the beginning and `,` after each row (apart from the last one).

That's a good point. I'll update it

yevgenypats:
Thanks!

PR #51, opened 2021-01-29
Add list of accounts

mvaerle:
I tried adding a list of accounts call to cloudquery. The use case that I have is that I want to join this from the other resources by accountId to provide a human readable name for the account instead of the id.

I'm running into some trouble, but I can't really finish it on my own, which is why I'm submitting the PR a bit incomplete, there are three issues:

- Deleting the tables doesn't work, since it depends on the accountId being in the where clause. I tried removing the where clause but that doesn't work. I see a lot of stuff about neo4j in the database layer, and I have no knowledge of neo4j so I'm not confident in changing that part of the code.
- This call only works from an account that can actually access the organizations resources. If not it will throw an access denied warning when running cloudquery, in my case it throws many of these. I'm not sure how to approach suppressing these warnings.
- When you have multiple accounts that can access the organizations resources they will both fetch the same accounts and create duplicate rows in the accounts table. I'm not sure how to solve this.

I hope this PR helps and you can pick it up further since I'm running into my limits here.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -2,10 +2,11 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/spf13/cobra"
 	"io/ioutil"
 	"os"
 	"strings"
+
+	"github.com/spf13/cobra"
 )
 
 var headerConfig = `providers:`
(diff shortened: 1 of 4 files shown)


mvaerle:
closing since #52 is now merged

PR #52, opened 2021-01-30
Add organizations

yevgenypats:
Continue PR: https://github.com/cloudquery/cloudquery/pull/51

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -2,10 +2,11 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/spf13/cobra"
 	"io/ioutil"
 	"os"
 	"strings"
+
+	"github.com/spf13/cobra"
 )
 
 var headerConfig = `providers:`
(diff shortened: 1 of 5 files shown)


yevgenypats:
@mvaerle I've added a small commit on top of your PR. Can you please test it? I think it's almost ready.

mvaerle:
@yevgenypats just tested it and it works for me, I don't get any double rows anymore. I ran it multiple times as well to test that.

Note that I still get a lot of warnings, since in `config.yml` I've added all the accounts and this call only works for two of those. For all the other accounts it throws an `AccessDeniedException`. Note sure what the best approach there is.

yevgenypats:
@mvaerle Thanks for testing! Can you share with me the snippet of the log (you can redact sensitive information). I think you can ignore this exception I just want to make sure it just print the error and continues and not stops the fetching.

yevgenypats:
I've changed the Warning to Info and added an account_id so it will be more verbose. you can ignore this message. Rebasing and releasing!

PR #54, opened 2021-02-04
Add sns topics

nick123pig:
Hello! I am interested in contributing to this project; I decided to add SNS Topics. Please let me know if this looks good!

### DIFF
--- a/providers/aws/sns/topics.go
+++ b/providers/aws/sns/topics.go
@@ -30,32 +30,58 @@ func (Topic) TableName() string {
 	return "aws_sns_topics"
 }
 
-func GetTopicAttributes(c *Client, TopicArn *string) map[string]*string {
+func GetTopicAttributes(c *Client, TopicArn *string) (map[string]*string, error) {
 	params := &sns.GetTopicAttributesInput{
 		TopicArn: TopicArn,
 	}
-	output, _ := c.svc.GetTopicAttributes(params)
-	return output.Attributes
+	output, err := c.svc.GetTopicAttributes(params)
+	if err != nil {
+		return nil, err
+	}
+	return output.Attributes, nil
 }
 
-func getOrDefault(attrs map[string]*string, keyName string) string {
+func getOrZero(attrs map[string]*string, keyName string) string {
 	if v, found := attrs[keyName]; found {
 		return *v
 	}
 	return "0"
 }
 
-func (c *Client) transformTopics(values []*sns.Topic) []*Topic {
+func (c *Client) transformTopics(values []*sns.Topic) ([]*Topic, error) {
 	var tValues []*Topic
 	for _, value := range values {
 
-		output := GetTopicAttributes(c, value.TopicArn)
+		// All topic attributes are returned as a string; we have to handle type conversion
+		output, err := GetTopicAttributes(c, value.TopicArn)
+		if err != nil {
+			return nil, err
+		}
 
-		subsConfirmed, _ := strconv.Atoi(getOrDefault(output, "SubscriptionsConfirmed"))
-		subsDeleted, _ := strconv.Atoi(getOrDefault(output, "SubscriptionsDeleted"))
-		subsPending, _ := strconv.Atoi(getOrDefault(output, "SubscriptionsPending"))
-		isFifo, _ := strconv.ParseBool(getOrDefault(output, "FifoTopic"))
-		isContentBasedDeduped, _ := strconv.ParseBool(getOrDefault(output, "ContentBasedDeduplication"))
+		subsConfirmed, subsConfirmedErr := strconv.Atoi(getOrZero(output, "SubscriptionsConfirmed"))
+		if subsConfirmedErr != nil {
+			return nil, subsConfirmedErr
+		}
+
+		subsDeleted, subsDeletedErr := strconv.Atoi(getOrZero(output, "SubscriptionsDeleted"))
+		if subsDeletedErr != nil {
+			return nil, subsDeletedErr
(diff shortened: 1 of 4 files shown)


yevgenypats:
Thanks for the contribution! I'll review shortly.

providers/aws/sns/topics.go
```diff
@@ -0,0 +1,110 @@
...
+	var tValues []*Topic
+	for _, value := range values {
+
+		output := GetTopicAttributes(c, value.TopicArn)
+
+		subsConfirmed, _ := strconv.Atoi(getOrDefault(output, "SubscriptionsConfirmed"))
+		subsDeleted, _ := strconv.Atoi(getOrDefault(output, "SubscriptionsDeleted"))
+		subsPending, _ := strconv.Atoi(getOrDefault(output, "SubscriptionsPending"))
+		isFifo, _ := strconv.ParseBool(getOrDefault(output, "FifoTopic"))
+		isContentBasedDeduped, _ := strconv.ParseBool(getOrDefault(output, "ContentBasedDeduplication"))
```
yevgenypats, line 58: what happens if the `strconv` fails?

providers/aws/sns/topics.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (Topic) TableName() string {
+	return "aws_sns_topics"
+}
+
+func GetTopicAttributes(c *Client, TopicArn *string) map[string]*string {
+	params := &sns.GetTopicAttributesInput{
+		TopicArn: TopicArn,
+	}
+	output, _ := c.svc.GetTopicAttributes(params)
```
yevgenypats, line 37: I think it best to check the error here and propagate it.

yevgenypats:
@nick123pig looks good! two minor comments/fixes.

providers/aws/sns/topics.go
```diff
@@ -0,0 +1,110 @@
...
+	return output.Attributes
+}
+
+func getOrDefault(attrs map[string]*string, keyName string) string {
```
James-Quigley, line 41: nitpick, but this really is `getOrZero`. Could adapt to allow for a default to be passed as a param and return that instead of `"0"`.

providers/aws/sns/topics.go
```diff
@@ -0,0 +1,110 @@
...
+	&Topic{},
+}
+
+func (c *Client) topics(gConfig interface{}) error {
+	var config sns.ListTopicsInput
+	err := mapstructure.Decode(gConfig, &config)
+	if err != nil {
+		return err
+	}
+	c.db.Where("region", c.region).Where("account_id", c.accountID).Delete(TopicTables...)
```
James-Quigley, line 94: should the delete happen after successfully calling `c.svc.ListTopics(&config)`? If that errors you'd end up with an empty table, right?

yevgenypats: It's a good questions - I don't think it really matters as the other-way around has it's advantages of not having old data when an error occurs. What I think is important is for this to be consistent across all resources and it's currently not so I'll open an issue and push a fix in other resources. 

From a quick glance I actually think that deleting first is better like implemented here. Because otherwise the delete will be in the [for loop](https://github.com/cloudquery/cloudquery/pull/54/files#diff-76628305ad10b059b8b3dcaac97629b56833706aed226f61538dbb0d76d18c3fR97) which is not the expected behaviour.

James-Quigley: Good point. Definitely don't want to delete in the loop!

But would be a bummer if you had some transient API error when loading the data from the API and ended up with an empty table

nick123pig: Agreed that the behavior is non-ideal; however, you could theoretically just re-run to get the correct results

providers/aws/sns/topics.go
nick123pig: Changed!

providers/aws/sns/topics.go
nick123pig: Addressed

providers/aws/sns/topics.go
nick123pig: I propagated these out

providers/aws/sns/topics.go
yevgenypats: I'm more inclined into the current behaviour but anyway I don't think this change should be part of this PR so I'm good with this one. @James-Quigley Thanks for chiming in! You can open a seperate issue on this one so we can collect a bit more feedback before deciding one way or another.

providers/aws/sns/topics.go
```diff
@@ -0,0 +1,142 @@
...
+	var tValues []*Topic
+	for _, value := range values {
+
+		// All topic attributes are returned as a string; we have to handle type conversion
+		output, err := GetTopicAttributes(c, value.TopicArn)
+		if err != nil {
+			return nil, err
+		}
+
+		subsConfirmed, subsConfirmedErr := strconv.Atoi(getOrZero(output, "SubscriptionsConfirmed"))
```
yevgenypats, line 61: @nick123pig the standard convention in Go is just to use `err` no need to use `prefixErr`.

yevgenypats:
@nick123pig Thanks! one last comment and it's ready to be merged.

providers/aws/sns/topics.go
nick123pig: ok thank you (this is my first go project, so feel free to nitpick my style :D)

nick123pig: resolved in 968bdc6

PR #61, opened 2021-02-23
Refactor plugin lifecycle management, client and logging

roneli:
First of multiple refactor branches

- Added plugin.Manager to handle plugin lifecycle
- Added configurable and readable logging with zerolog (replacing zap)
- Client provider fetch runs in parallel
- updated plugin SDK
- plugin self-debug now can be run on any plugin when the CQ_DEBUG_PROVIDER=1 is turned on.


providers/README.md
```diff
@@ -1,10 +0,0 @@
-Providers are now shipped as plugins to enable users develop their own providers in a seperate repository.
-
-CloudQuery Hub: https://hub.cloudquery.io
-
-Official plugins are located:
-- https://github.com/cloudquery/cq-provider-aws
-- https://github.com/cloudquery/cq-provider-azure
-- https://github.com/cloudquery/cq-provider-gcp
-- https://github.com/cloudquery/cq-provider-okta
-- https://github.com/cloudquery/cq-provider-ok8s
```
yevgenypats, line 10: Why this file is deleted? I think some people go through the code and look for the providers so it's nice to have a stub with README here imo?

sdk/serve.go
```diff
@@ -0,0 +1,71 @@
...
+	// Optional: Set NoLogOutputOverride to not override the log output with an hclog
+	// adapter. This should only be used when running the plugin in
+	// acceptance tests.
+	NoLogOutputOverride bool
+}
+
+
+func ServePlugin(opts ServeOpts) {
+
+	if opts.Name == "" {
```
yevgenypats, line 32: if it's requried why not put it in the function signature and the options in the struct?

sdk/serve.go
```diff
@@ -0,0 +1,71 @@
...
+	// CloudQuery main command line accepting any args same as the main binary. The client will execute the
+	// only this plugin instead of using the downloaded provider plugins
+	if os.Getenv("CQ_PROVIDER_DEBUG") != "" {
+		plugin.GetManager().SetProvider(opts.Name, opts.Provider)
+		cmd.Execute()
+	}
+
+	if !opts.NoLogOutputOverride {
+		// In order to allow go-plugin to correctly pass log-levels through to
+		// terraform, we need to use an hclog.Logger with JSON output. We can
```
yevgenypats, line 50: it's cloudquery not terraform:)

sdk/serve.go
```diff
@@ -0,0 +1,71 @@
...
+	if opts.Provider == nil {
+		panic("missing provider instance")
+	}
+
+	// Check of CQ_PROVIDER_DEBUG is turned on. In case it's true we self register the plugin and execute
+	// CloudQuery main command line accepting any args same as the main binary. The client will execute the
+	// only this plugin instead of using the downloaded provider plugins
+	if os.Getenv("CQ_PROVIDER_DEBUG") != "" {
+		plugin.GetManager().SetProvider(opts.Name, opts.Provider)
+		cmd.Execute()
```
yevgenypats, line 45: I think there should be a return statement here

plugin/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	client.Kill()
+	delete(p.activeClients, providerName)
+	return nil
+}
+
+func (p *Manager) GetOrCreateProvider(providerName, version string) (CQProvider, error) {
+
+	provider, err := p.GetProvider(providerName, version)
+	if provider != nil {
```
yevgenypats, line 72: I think it's missing an error check here.

plugin/manager.go
```diff
@@ -0,0 +1,129 @@
...
+		return provider, err
+	}
+
+	p.lock.Lock()
+	version = ""
```
yevgenypats, line 77: I think this shouldn't be here as it always override this.

plugin/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	provider, ok := raw.(CQProvider)
+	if !ok {
+		client.Kill()
+		return nil, fmt.Errorf("failed to cast plugin")
+	}
+
+	p.activeClients[providerName] = client
+	p.lock.Unlock()
```
yevgenypats, line 112: maybe we should use defer? otherwise if there is an error it never releases the lock.

roneli:
Task linked: [CU-6nx7hg Logging refactor](https://app.clickup.com/t/6nx7hg)

yevgenypats (changes requested):
1st iteration

providers/README.md
roneli: Added it again, although eventually it should be removed

plugin/manager.go
roneli: removed

sdk/serve.go
roneli: It allows for better backwards compatibility

plugin/manager.go
roneli: added

plugin/manager.go
```diff
@@ -69,17 +69,25 @@ func (p *Manager) KillProvider(providerName string) error {
 func (p *Manager) GetOrCreateProvider(providerName, version string) (CQProvider, error) {
 
 	provider, err := p.GetProvider(providerName, version)
-	if provider != nil {
+	if provider != nil && err != nil {
```
yevgenypats, line 72: I think it should be  an `||` not an `&&`

plugin/manager.go
```diff
@@ -69,17 +69,25 @@ func (p *Manager) KillProvider(providerName string) error {
...
-
-	p.lock.Lock()
-	version = ""
-	if version == "" {
-		version = "latest"
+	// Create RPC client and initialize CQProvider
+	provider, err = p.createProvider(providerName, version)
+	if err != nil {
+		return nil, err
 	}
```
yevgenypats, line 79: I think it's possible there will be a race condition here. between create a provider and set a provider.

plugin/manager.go
```diff
@@ -105,16 +113,12 @@ func (p *Manager) GetOrCreateProvider(providerName, version string) (CQProvider,
 	provider, ok := raw.(CQProvider)
 	if !ok {
 		client.Kill()
-		return nil, fmt.Errorf("failed to cast plugin")
+		return provider, fmt.Errorf("failed to cast plugin")
```
yevgenypats, line 116: why not `nil`?

yevgenypats (changes requested):

example.aws_cis.policy.yml
```diff
@@ -85,7 +85,7 @@ queries:
       SELECT aws_cloudtrail_trails.account_id, trail_arn, is_multi_region_trail, read_write_type, include_management_events FROM aws_cloudtrail_trails
```
yevgenypats, line 85: I think file should be unchanged. why `logging` -> `log` ?

plugin/manager.go
```diff
@@ -0,0 +1,124 @@
...
+var (
+	doOnce   sync.Once
+	instance *Manager
+)
+
+type Manager struct {
+	activeClients map[string]*plugin.Client
+	registry      map[string]CQProvider
+
+	lock sync.RWMutex
```
yevgenypats, line 22: No need for the locks in the manager as we are not calling it from multiple go routines.

plugin/manager.go
```diff
@@ -0,0 +1,124 @@
...
+func (p *Manager) Setup() error {
+	return nil
+}
+
+func (p *Manager) Shutdown() {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	for n, c := range p.activeClients {
+		c.Kill()
+		delete(p.activeClients, n)
```
yevgenypats, line 34: I guess you can `p.activeClient = make()` just like you do with `p.registry`. why different initialisation?

plugin/manager.go
```diff
@@ -0,0 +1,124 @@
...
+func (p *Manager) KillProvider(providerName string) error {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	client, ok := p.activeClients[providerName]
+	if !ok {
+		return fmt.Errorf("client for provider %s does not exist", providerName)
+	}
+	client.Kill()
+	delete(p.activeClients, providerName)
```
yevgenypats, line 65: Missing `delete(p.registry, providerName)`

plugin/manager.go
```diff
@@ -0,0 +1,124 @@
...
+	"sync"
+)
+
+var (
+	doOnce   sync.Once
+	instance *Manager
+)
+
+type Manager struct {
+	activeClients map[string]*plugin.Client
```
yevgenypats, line 19: Looks like there is a bit of duplication here (I know it was also in my code but if we are already refactoring...). This also causes duplicate add and remove in setup and cleanup (also found some minor bug in cleanup of registry).

I suggest `registry map[string]struct{
    client *plugin.Client
    provider CQProvider
}`

client/config.go
```diff
@@ -0,0 +1,101 @@
...
+			return err
+		}
+		p, err := plugin.GetManager().GetOrCreateProvider(provider, "latest")
+		if err != nil {
+			return err
+		}
+		log.Debug().Str("provider", provider).Msg("Building provider configuration yaml")
+		configYaml, err := p.GenConfig()
+		if err != nil {
+			_ = plugin.GetManager().KillProvider(provider)
```
yevgenypats, line 83: I agree that we need to return the first error but I'd at least print this error as well.

client/hub.go
```diff
@@ -139,55 +145,75 @@ func getProviderPath(name string, version string) (string, error) {
...
 	}
 
-	pluginDir := fmt.Sprintf("%s/.cq/providers/%s/%s", workingDir, org, name)
+	pluginDir := filepath.Join(workingDir, ".cq", "providers", org, pluginName)
 	err = os.MkdirAll(pluginDir, os.ModePerm)
 	if err != nil {
 		return err
 	}
-	pluginPath := fmt.Sprintf("%s/%s-%s-%s", pluginDir, version, runtime.GOOS, runtime.GOARCH)
+	pluginPath, _ := getProviderPath(pluginName, version)
```
yevgenypats, line 213: why ignore this error here?

yevgenypats (changes requested):
Final review!:) Looks good.

client/config.go
roneli: refactoring this in a different branch

client/hub.go
roneli: should be fixed in later refactor

yevgenypats (approved):
Hooray!

PR #64, opened 2021-02-24
Minor config/hub refactor

roneli:
minor config/hub refactor
- allow config generate append mode


client/client.go
```diff
@@ -29,48 +31,59 @@ type PolicyConfig struct {
...
+	config  *config.Config
+	// access to CloudQuery plugin hub
+	hub hub.Hub
 }
 
-func New(driver string, dsn string) (*Client, error) {
-	client := Client{
-		driver:  driver,
-		dsn:     dsn,
+func New(configPath, driver string, dsn string) (Client, error) {
```
yevgenypats, line 52: why you decided to not return an address of. This way you can pass `nil` and not an un-initialised object.

plugin/hub/hub.go
```diff
@@ -57,240 +62,222 @@ I8Hs4e6+i4/g8yyp1aO9jClsLVJL4Xp9o6O6aYpSDj17MEXhV5U053grDEuvvNCA
...
+type Hub struct {
+	registryURL     string
+	skipRegistry    bool
+	pluginDirectory string
+}
+
+func NewHub(skipRegistry bool) Hub {
+	dir, err := os.Getwd()
 	if err != nil {
-		return "", err
```
yevgenypats, line 70: why not return `nil`, err ?

Also same question as above - why not `*Hub` as return value?

yevgenypats (changes requested):
Looks good. Minor questions/suggestions.

yevgenypats (approved):

PR #68, opened 2021-03-01
README.md: Add basic sqlite schema describes

nitrocode:
Changes

* Added how to quickly load the database after the fetch
* Added basic sqlite schema describes
* Most people aren't building this so I removed the `./` from `./cloudquery` from the quick start commands


yevgenypats:
Thanks @nitrocode !

PR #70, opened 2021-03-03
Allow config to be loaded independently of the client

zscholl:
## Description

Previously the config was loaded by passing a filename to `client.New()`, this made it difficult to load a config from a source other than a file.

For the Lambda deployment it is convenient to provide the config as a payload when triggering the Lambda. In order to support this, this change breaks up the configuration loading and client initialization.


Makefile
```diff
@@ -56,10 +56,6 @@ build: config
 proto:
 	@protoc --go_out=. --go-grpc_out=. ./sdk/proto/plugin.proto
 
-.PHONY: config
-config:
-	@go run main.go gen config aws --force --path ./bin/config.yml
-
```
zscholl, line 62: Remove the config step as it's no longer necessary for the lambda deployment

client/client.go
```diff
@@ -49,11 +49,7 @@ type Client struct {
 	hub *hub.Hub
 }
 
-func New(configPath, driver string, dsn string) (*Client, error) {
-	cfg, err := config.Parse(configPath)
-	if err != nil {
-		return nil, err
-	}
+func New(driver, dsn string, cfg *config.Config) (*Client, error) {
```
zscholl, line 52: Pass in a reference to a config object instead of a file path to parse. This is the main change. It requires users of the client to create a config outside of the client initialization

client/client.go
```diff
@@ -75,15 +71,13 @@ func (c Client) Initialize() error {
 	return nil
 }
 
-func (c *Client) Run(path string) error {
-
-	cfg, err := config.Parse(path)
```
zscholl, line 80: No need to parse a config path if the client has already been initialized.

client/client_test.go
```diff
@@ -14,7 +15,11 @@ func TestMigrationSQLServers(t *testing.T) {
 	if err != nil {
 		log.Fatalf("Could not connect to docker: %s", err)
 	}
-	c, err := client.New("./testdata/config.yml", "", "")
+	cfg, err := config.Parse("./testdata/config.yml")
+	if err != nil {
+		t.Fatal(err)
+	}
+	c, err := client.New("", "", cfg)
```
zscholl, line 22: This is how users of the library now have to initialize a client. First they create a config object, then pass it into `client.New`

client/config.go
```diff
@@ -50,20 +51,20 @@ func GenerateConfig(configPath string, providers []string, allowAppend bool, for
...
 	defer plugin.GetManager().KillProvider(providerName)
 
 	log.Debug().Str("provider", providerName).Msg("Building provider configuration yaml")
 	configYaml, err := p.GenConfig()
 	if err != nil {
-		return config.Config{}, err
+		return &config.Config{}, err
 	}
-	return config.LoadFromString(configYaml)
+	return config.LoadFromString(fmt.Sprintf("providers:\n%s", configYaml))
```
zscholl, line 69: This was being handled by `config.ParseFromString`. It's a corner case that should be handled by this function to leave `ParseFromString` as more general

config/parser.go
```diff
@@ -23,15 +22,11 @@ func Parse(configPath string) (*Config, error) {
 }
 
 // Load Config data from string input
-func LoadFromString(data string) (Config, error) {
-	var b bytes.Buffer
-	b.WriteString("providers:\n")
-	b.WriteString(data)
```
zscholl, line 29: Corner case for the provider config generation that's been moved into that function instead.

roneli (approved):

cmd/query.go
```diff
@@ -17,7 +18,7 @@ var queryCmd = &cobra.Command{
 		dsn := viper.GetString("dsn")
 		queryConfigPath := viper.GetString("policy_path")
 		queryOutputJsonPath := viper.GetString("output")
-		client, err := client.New("", driver, dsn)
+		client, err := client.New(driver, dsn, &config.Config{})
```
yevgenypats, line 21: maybe pass `nil` instead of an uninitialised config object?

deploy/lambda.go
```diff
@@ -48,7 +54,7 @@ func Fetch(driver, dsn string) {
 func Policy(driver, dsn string) {
 	outputPath := "/tmp/result.json"
 	queryPath := os.Getenv("CQ_QUERY_PATH") // TODO: if path is an S3 URI, pull file down
-	c, err := client.New("", driver, dsn)
+	c, err := client.New(driver, dsn, &config.Config{})
```
yevgenypats, line 57: same here - maybe pass `nil` instead of empty config?

yevgenypats (changes requested):
Thanks! few minor comments

zscholl:
@yevgenypats thanks for the feedback! I've incorporated the changes you suggested, let me know when you have a chance to review 🙏

yevgenypats:
Thanks!! merged your branch via this PR https://github.com/cloudquery/cloudquery/pull/74

PR #71, opened 2021-03-03
Optionally read in provider dir from environment

zscholl:
## Description

It may be desirable in some environments to use a different directory for plugins. This change adds an environment variable `CQ_PLUGIN_DIR` that if present will be used as the directory for plugins.

If the environment variable is not set, the default behavior of using the current script directory is preserved.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -48,7 +48,11 @@ func init() {
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxSize, "maxSize", 30, "MaxSize the max size in MB of the logfile before it's rolled")
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxBackups, "maxBackups", 3, "MaxBackups the max number of rolled files to keep")
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxAge, "maxAge", 3, "MaxAge the max age in days to keep a logfile")
-	rootCmd.PersistentFlags().String( "plugin-dir", ".", "Directory to save and load Cloudquery plugins from (env: CQ_PLUGIN_DIR)")
+	workingDir, err := os.Getwd()
+	if err != nil {
+		workingDir = "."
+	}
+	rootCmd.PersistentFlags().String( "plugin-dir", workingDir, "Directory to save and load Cloudquery plugins from (env: CQ_PLUGIN_DIR)")
 	_ = viper.BindPFlag("plugin-dir", rootCmd.PersistentFlags().Lookup("plugin-dir"))
 	cobra.OnInitialize(initConfig, initLogging)
 }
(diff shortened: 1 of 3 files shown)


plugin/hub/hub.go
```diff
@@ -71,14 +71,21 @@ type Hub struct {
 }
 
 func NewHub(skipRegistry bool) *Hub {
-	dir, err := os.Getwd()
-	if err != nil {
-		dir = "."
+
+	var pluginDir string
+	var err error
+	pluginDir = os.Getenv("CQ_PLUGIN_DIR")
```
roneli, line 77: Instead of using os.Getenv i'd suggest to use viper

roneli: You can add the following to cmd/root.go

rootCmd.PersistentFlags().String( "pluginDir", ".", "Directory to save and load Cloudquery plugins from (env: CQ_PLUGINDIR)")
_ = viper.BindPFlag("plugindir", rootCmd.PersistentFlags().Lookup("pluginDir"))

Then you can change dir, err := os.Getwd() to just dir = viper.GetString("pluginDir")

roneli (changes requested):
Minor suggestion to use viper instead of os.Getenv so pluginDir can be also controlled via CLI

yevgenypats:
@zscholl Thanks for this! In [v0.11.3](https://github.com/cloudquery/cloudquery/releases/tag/v0.11.3) the fetch command now also downloads the providers so it should make life simpler when running a docker and I believe the lambda deployment can now also work without this feature. 


Unless this is a blocker for the lambda deployment we might add this a bit later as we wanted to refactor some of this code so it will be easier to pass configurations, etc for future changes.

zscholl:
Unfortunately this is a blocker for the lambda deployment. We need to write the downloaded providers to `/tmp` otherwise it attempts to write to a read-only filesystem.

`2021/03/06 14:02:11 Unable to initialize client: mkdir /var/task/.cq: read-only file system`


I updated the config to use Viper, let me know if this is okay to merge for now 🙏

cmd/root.go
```diff
@@ -48,6 +48,8 @@ func init() {
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxSize, "maxSize", 30, "MaxSize the max size in MB of the logfile before it's rolled")
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxBackups, "maxBackups", 3, "MaxBackups the max number of rolled files to keep")
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxAge, "maxAge", 3, "MaxAge the max age in days to keep a logfile")
+	rootCmd.PersistentFlags().String( "plugin-dir", ".", "Directory to save and load Cloudquery plugins from (env: CQ_PLUGIN_DIR)")
```
yevgenypats, line 51: the default value should be `.` is used only when `Getwd` is not working

```
	workingDir, err := os.Getwd()
	if err != nil {
		return "", err
	}
```

yevgenypats (changes requested):
@zscholl Thanks! one comment and we are good to merge!

cmd/root.go
zscholl: Roger that! How does the latest commit look? I used `Getwd()` as the default and only fallback to "." if that command fails.

yevgenypats:
Looks good. Merging!

PR #75, opened 2021-03-05
Feat: Linting action

roneli:
Add golintci action to the repo and fix all current errors

### DIFF
--- a/plugin/proto/plugin.pb.go
+++ b/plugin/proto/plugin.pb.go
@@ -7,11 +7,12 @@
 package proto
 
 import (
+	reflect "reflect"
+	sync "sync"
+
 	proto "github.com/golang/protobuf/proto"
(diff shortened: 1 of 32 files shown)


plugin/proto/plugin.pb.go
```diff
@@ -7,11 +7,12 @@
 package proto
 
 import (
+	reflect "reflect"
```
yevgenypats, line 10: I think we need the linter to skip the proto files as it's auto-generated?

plugin/proto/plugin_grpc.pb.go
```diff
@@ -4,6 +4,7 @@ package proto
 
 import (
 	context "context"
+
```
yevgenypats, line 7: same here.

yevgenypats (changes requested):
Awesome! one minor comment.

PR #76, opened 2021-03-06
bugFix: Add custom unmarshaler for json config

zscholl:
## Description

Currently the native Json unmarshaling will not parse out the `Rest` portion of each of the providers. 
`encoding/json` does not have an `inline` struct tag similar to `yaml` so there is no way to parse the extra data from json without a custom unmarshaler.

I'm proposing an implementation of a custom json unmarshaler here, also added some basic tests.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -45,34 +45,34 @@ func (c *Config) ProviderExists(providerName string) bool {
 //
 // Function assumes config keys are lowercase
 func (c *Config) UnmarshalJSON(data []byte) error {
-	var v map[string][]map[string]interface{}
+	var conf map[string][]map[string]interface{}
 
-	if err := json.Unmarshal(data, &v); err != nil {
+	if err := json.Unmarshal(data, &conf); err != nil {
 		return err
 	}
 
-	for _, d := range v["providers"] {
+	for _, provMap := range conf["providers"] {
 		prov := Provider{}
 		var ok bool
 		numKnownKeys := 0
 
-		if _, ok := d["name"]; ok {
+		if _, ok := provMap["name"]; ok {
 			numKnownKeys++
 		}
-		if _, ok := d["version"]; ok {
+		if _, ok := provMap["version"]; ok {
 			numKnownKeys++
 		}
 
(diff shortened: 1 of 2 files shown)


yevgenypats:
@zscholl thanks! Actually I just double tested it now and `yaml:",inline"` does [exist](https://stackoverflow.com/questions/53666084/how-to-unmarshal-embedded-structs-from-yaml) (I removed it and added it again to see that it works).

Otherwise it means that cloudquery wouldn't work at all. Am I missing something? Can you share your scenario?

yevgenypats:
@zscholl Thanks! (Sorry for previous comment, I misread the description (too early in morning:) )).

Cool, We'll look into this, can you share a use-case for json as config instead of yaml?

zscholl:
No worries :)

The lambda currently [accepts config](https://github.com/cloudquery/cloudquery/blob/6b5fd9fd2fa97ee35804a9e270102750da870622/deploy/lambda.go#L19) as part of the json payload for each request. It naively unpacks the json into the struct using a [struct tag](https://github.com/cloudquery/cloudquery/blob/6b5fd9fd2fa97ee35804a9e270102750da870622/deploy/lambda.go#L15).

I previously thought this was working fine, but after some testing yesterday I realized it wasn't actually using any of the `Rest` portion of the `Provider` object.

That led me down the `inline` rabbit hole, and eventually to writing this custom unmarshaler. 

With this custom unmarshaler the lambda will accept a payload like this:
```json
{
	"taskName": "fetch",
	"config": {
		"providers": [
			{
				"name": "aws",
				"regions": [
					"us-east-1"
				],
				"resources": [
					{
					  "name": "ec2.images"
					},
					{
					  "name": "ec2.instances"
					},
					{
					  "name": "s3.buckets"
					}
				]
			}
		]
	}
}
```

And run the config as defined in that payload.

Why this is really cool is that you can parallelize fetch tasks by scheduling a bunch of Cloudwatch triggers for the lambda that each have a different config. You could have one trigger for each region, each aws account, or even each resource type.

roneli (approved):

PR #83, opened 2021-04-08
feat: move plugin to sdk

roneli:
provider no longer need to import cloudquery to serve themselves only sdk is enough. Better debugging ability


yevgenypats (changes requested):
Is there any tests for that change. It's not a big change but with room for "problems" maybe add some basic tests with this PR?

PR #100, opened 2021-05-20
Aws lambda

zagronitay:


roneli (approved):
Looks Good! lets check why that test flakes out randomly and we can merge this.

PR #101, opened 2021-05-22
Add FreeBSD Package Mention to README.md

neelchauhan:
I maintain the FreeBSD Port/Package for cloudquery. We should mention it in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,15 @@ brew install cloudquery/tap/cloudquery
 brew upgrade cloudquery
 ```
 
+FreeBSD
+```shell script
+# via packages
+pkg install cloudquery
+# via Ports
+cd /usr/ports/net/cloudquery
+make config-recursive install clean
+```
+
 ## Quick Start
 
 ### Running


yevgenypats:
Hi @neelchauhan Thanks for contribution! A few questions, comments:

One of the strength of Go that it can easily compile to any architecture and OS very easily. So if freeBSD is needed we can easily add it to the release. This will also remove the need to maintain a port and for the user to compile it locally instead of just running a pre-compiled binary. Let me know what you think?

neelchauhan:
FYI "FreeBSD Ports" is the name of the packaging system used on FreeBSD, like Debian/Ubuntu `apt-get`, macOS `homebrew`, or Arch `pacman`.

yevgenypats:
Yes, I know but still it will be easier to package a pre-compiled binary then to compile it. Also, I believe once you add only the pre-compiled binary this line: `make config-recursive install clean` won't be necessary

yevgenypats:
Closing for now.

PR #109, opened 2021-05-27
Builde Release binaries for arm64 architecture

GennadySpb:
Include Apple Silicon.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -18,6 +18,10 @@ builds:
       - darwin
     goarch:
       - amd64
+      - arm64
+    ignore:
+      - goos: windows
+        goarch: arm64
 archives:
   -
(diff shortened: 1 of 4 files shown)


.goreleaser.yaml
```diff
@@ -18,6 +18,10 @@ builds:
       - darwin
     goarch:
       - amd64
+      - arm64
+    ignore:
+      - goos: windows
+        goarch: arm64
 archives:
   -
     name_template: "{{ .Binary }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
```
yevgenypats, line 27: Can you remove here the `{{ if .Arm }}v{{ .Arm }}` It's not necessary and it will break the hub functionality/convention.

yevgenypats (changes requested):
Looks good, one comment.

PR #113, opened 2021-05-27
BugFix/panic on empty configuration

roneli:
- allowed to use source based on the provider config source=cloudquery previously was hard coded as cloudquery,  although it defaults to cloudquery as well.
-  fix panic when not adding a configuration block to a provider, that means provider uses the defaults values.


roneli:
fixes #113

PR #131, opened 2021-06-15
Updated the documentation

madhuakula:
I have updated the documentation at README.md and also added some images, screenshots for references to use the cloudquery.


README.md
```diff
@@ -125,47 +207,64 @@ queries:
         SELECT account_id, arn FROM ...
 ```
 
-The `policy` command uses the policy file path `./policy.yml` by default, but this can be overridden via the `--path` flag, or the `CQ_POLICY_PATH` environment variable.
+> The `policy` command uses the policy file path `./policy.yml` by default, but this can be overridden via the `--path` flag, or the `CQ_POLICY_PATH` environment variable
```
igrant-2ndWatch, line 210: Can you confirm if this information is still accurate? Neither the --path flag or the CQ_POLICY_PATH environment variable seem to be supported.

roneli: No this is an old PR please reference please docs.cloudquery.io for more info regarding running policies. Also all available policies exist in our hub with run examples. Hope this helps :)

igrant-2ndWatch: @roneli Thank you for the prompt response and the resource. I was hopeful this would open some dialogue as the current README also reflects this configurability. I am familiar with the docs, but given our use case I was searching for expanded functionality, some of which appears to be part of a previous iteration 😞.

Would you recommend I continue this dialogue and ask targeted questions in a separate issue or is there a better location?

Best

roneli: Yes I recommend we open a new Issue, it will be easier to continue there instead of this PR.

PR #133, opened 2021-06-17
Policy v2

michelvocks:


michelvocks:
@roneli Should be ready for review

pkg/policy/manager.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+	policy := &Policy{
+		SubPath: subPolicyPath,
+	}
+
+	// Parse and validate org/repository
+	orgRepoSplit := strings.Split(args[0], pathDelimiter)
+
+	// Parse org/repo
+	switch len(orgRepoSplit) {
```
roneli, line 109: maybe use [this](https://github.com/cloudquery/cloudquery/blob/21d9b753117aa1bfecbf730b7065f8f32165558e/pkg/plugin/registry/organization.go#L15) function can be upgraded, seems like similar solutions no? or it can be used for the first part

pkg/policy/manager.go
```diff
@@ -0,0 +1,455 @@
...
+
+	// Output progress information if necessary
+	if ui.IsTerminal() {
+		// Print initial information
+		switch {
+		case p.Version != "":
+			ui.ColorizedOutput(ui.ColorProgress, fmt.Sprintf("Cloning Policy %s/%s@%s\n", p.Organization, p.Repository, p.Version))
+		default:
+			ui.ColorizedOutput(ui.ColorProgress, fmt.Sprintf("Cloning Policy %s/%s\n", p.Organization, p.Repository))
+		}
```
roneli, line 163: this can be set outside of the condition since it's checked inside, maybe we should upgrade ui.ColorizedOutput to print to global log if ui isn't terminal.

pkg/policy/manager.go
```diff
@@ -0,0 +1,455 @@
...
+			return execResults, nil
+		}
+	}
+
+	return nil, fmt.Errorf("cannot find sub query %s in %s", elementName, p.SubPath)
+}
+
+// traversePolicies is a recursive function that traverses p until all policies are resolved.
+// All traversed policies gets stored into m where the policy level is used as a key.
+func traversePolicies(p []*config.Policy, levelPath string, m map[string]*config.Policy) map[string]*config.Policy {
```
roneli, line 313: I would add some debug logging here if we ever want to understand what got added to the policy map etc'.

pkg/policy/manager.go
```diff
@@ -0,0 +1,455 @@
...
+		}
+	default:
+		return fmt.Errorf("failed to clone repository: %s", err.Error())
+	}
+
+	return nil
+}
+
+// RunPolicy runs the given policy.
+func (m *ManagerImpl) RunPolicy(ctx context.Context, execReq *ExecuteRequest) (*ExecutionResult, error) {
```
roneli, line 187: I would add some info/debug log lines here. Sometimes users run --enable-console-log (cancelling the console UI) without any log prints (debug/info) you can't see any progress or debug etc'

pkg/policy/manager.go
```diff
@@ -0,0 +1,455 @@
...
+	policyMap := make(map[string]*config.Policy)
+	policyMap = traversePolicies(policies.Policies, "", policyMap)
+
+	// No sub path provided. Execute everything.
+	executor := NewExecutor(conn)
+	if p.SubPath == "" {
+		return executor.ExecutePolicies(ctx, execReq, policyMap)
+	}
+
+	// If we are here, the user only wants to execute a sub policy/view/query so we have to
```
roneli, line 253: suggestion: wouldn't it be better to split this function to the 3 functions -> normal execution, sub policy execution, single query execution?

cmd/policy_run.go
```diff
@@ -0,0 +1,53 @@
...
+	subPath       string
+	outputPath    string
+	stopOnFailure bool
+)
+
+func init() {
+	flags := policyRunCmd.Flags()
+	flags.StringVar(&subPath, "subpath", "", "Forces the policy run command to only execute this sub policy/query")
+	flags.StringVar(&outputPath, "output", "", "Generates a new file at the given path with the output")
+	flags.BoolVar(&stopOnFailure, "stoponfailure", false, "Stops the execution on the first failure")
```
roneli, line 51: stop-on-failure

roneli (changes requested):
Looks amazing! I think I only found 1 major issue, tell me your opinion about that. In general I would sprinkle more log info/debug lines in the execution/policy run|download so more info will be available in the cloudquery log.

pkg/client/client.go
```diff
@@ -252,59 +253,54 @@ func (c Client) GetProviderConfiguration(ctx context.Context, providerName strin
...
-	if err := yaml.Unmarshal(data, &policy); err != nil {
-		return nil, err
-	}
-	// Create Views
-	c.Logger.Debug("creating policy views", "policy", request.PolicyPath)
-	if err := createViews(ctx, conn, policy.Views); err != nil {
-		return nil, fmt.Errorf("failed to create policy views %w", err)
+func (c Client) RunPolicy(ctx context.Context, req PolicyRunRequest) error {
+	c.Logger.Info("Running policy", "args", req.Args)
+	m := policy.NewManager(c.PolicyDirectory, c.pool)
```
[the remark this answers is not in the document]
michelvocks: It wouldn't! The policy managers functions are thread-safe. However, I agree that it may makes sense to document that in the code so future developers are aware of that.

pkg/policy/execute_test.go
```diff
@@ -0,0 +1,158 @@
...
+			if tc.ShouldBeEmpty {
+				assert.Empty(t, res.Data)
+			} else {
+				assert.NotEmpty(t, res.Data)
+			}
+		})
+	}
+}
+
+func TestExecutor_ExecutePolicies(t *testing.T) {
```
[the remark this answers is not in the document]
michelvocks: Sure 👍

pkg/policy/execute.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// UpdateCallback gets called when the client receives updates on policy execution (optional).
+	UpdateCallback ExecutionCallback
+
+	// StopOnFailure if true policy execution will stop on first failure
+	StopOnFailure bool
+}
+
+// NewExecutor creates a new executor singleton instance and/or returns one.
+func NewExecutor(conn *pgxpool.Conn) *Executor {
```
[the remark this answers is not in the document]
michelvocks: Good catch! I agree. Let me refactor this.

pkg/policy/manager.go
michelvocks: The error that gets returned by that function doesn't fit the use-case here. We could refactor this but I wonder if it is worth the effort since we're talking here about a few lines of code? WDYT?

pkg/policy/manager.go
michelvocks: Good point. Yeah that sounds reasonable.

pkg/policy/manager.go
michelvocks: Good idea 👍

pkg/policy/manager.go
michelvocks: I agree. Sounds reasonable.

cmd/policy_run.go
michelvocks: 👍

pkg/config/config_parser.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (p *Parser) decodeConfig(body hcl.Body, diags hcl.Diagnostics) (*Config, hcl.Diagnostics) {
+
+	config := &Config{}
+
+	content, contentDiags := body.Content(configFileSchema)
+	diags = append(diags, contentDiags...)
+
+	for _, block := range content.Blocks {
```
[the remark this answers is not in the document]
michelvocks: Maybe but I think we shouldn't do that now. Let's keep that for future improvements. WDYT?

pkg/policy/manager.go
michelvocks: Makes sense. Thanks!

pkg/client/client.go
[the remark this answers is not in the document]
michelvocks: Decided to make the policy manager not a singleton anymore which also fixes the executor singleton issue.

pkg/policy/manager_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// Setup database
+	pool, tearDownFunc := setupDatabase(t, "test_policy_table")
+	defer tearDownFunc(t)
+
+	m := NewManager(tmpDir, pool)
+
+	// TODO: Add test for official cloudquery org
+	// Checkout repo with "main" branch
+	policyHubPath := []string{"michelvocks/my-cq-policy"}
```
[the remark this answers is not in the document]
michelvocks: Fixed. Thanks!

roneli (approved):
Looks good! minor changes are left. I feel that we should log a little more but maybe that's just me.  I am approving we can merge once fix these minor issues.

pkg/policy/manager.go
```diff
@@ -0,0 +1,473 @@
...
+	// Define clone options
+	cloneOptions := &git.CloneOptions{
+		URL:  gitURL,
+		Tags: git.AllTags,
+	}
+
+	// Print initial information
+	switch {
+	case p.Version != "":
+		ui.ColorizedOutput(ui.ColorProgress, fmt.Sprintf("Cloning Policy %s/%s@%s\n", p.Organization, p.Repository, p.Version))
```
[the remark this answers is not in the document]
michelvocks: Done. Thanks!

cmd/root.go
```diff
@@ -60,8 +60,10 @@ func init() {
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxBackups, "max-backups", 3, "MaxBackups the max number of rolled files to keep")
 	rootCmd.PersistentFlags().IntVar(&loggerConfig.MaxAge, "max-age", 3, "MaxAge the max age in days to keep a logfile")
 	rootCmd.PersistentFlags().String("plugin-dir", "./.cq/providers", "Directory to save and load CloudQuery plugins from (env: CQ_PLUGIN_DIR)")
+	rootCmd.PersistentFlags().String("policy-dir", ".", "Directory to save and load CloudQuery policies from (env: CQ_POLICY_DIR)")
```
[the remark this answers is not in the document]
michelvocks: Done. Thanks!

cmd/init.go
```diff
@@ -51,6 +51,7 @@ func Initialize(providers []string) {
 	// TODO: build this manually with block and add comments as well
 	cqBlock := gohcl.EncodeAsBlock(&config.CloudQuery{
 		PluginDirectory: ".",
+		PolicyDirectory: ".",
```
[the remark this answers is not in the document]
michelvocks: Done. Thanks!

PR #137, opened 2021-06-23
Implement file HCL function

michelvocks:
This PR implements the `file` function into the HCL parser that allows policies to specify an external path.
It also fixes two small issues with the policy execution that I noticed (error output malformed and repository path is evaluated before repository checkout).


roneli (approved):
Looks good

PR #144, opened 2021-07-09
skip versioning flag added for policy run

fdistorted:


pkg/config/config_parser.go
```diff
@@ -74,6 +74,7 @@ func (p *Parser) decodeConfig(body hcl.Body, diags hcl.Diagnostics) (*Config, hc
 					config.CloudQuery.PolicyDirectory = dir
 				}
 			}
+			config.CloudQuery.SkipVersioning = viper.GetBool("skip-versioning")
```
michelvocks, line 77: I think this is not required since you don't set a `viper.BindPFlag`. This flag should be also only required if you run a policy so a command argument is enough (IMO).

michelvocks (approved):
LGTM. One small suggestion.

pkg/config/config_parser.go
fdistorted: agree

PR #154, opened 2021-07-22
added error check to return error if view creation query failed.

fdistorted:


roneli (approved):
Looks Good, would add a test for this if possible.

michelvocks (commented):
I wonder if this really fixes the issue because we actually don't return the error during view creation: https://github.com/cloudquery/cloudquery/blob/main/pkg/policy/execute.go#L101

PR #162, opened 2021-08-03
fix/consistent_policy_execution

roneli:


pkg/policy/manager.go
```diff
@@ -282,7 +286,7 @@ func (m *ManagerImpl) runSubPolicyOrQuery(
 				subPolicyMap[k] = v
 			}
 		}
-		return exec.ExecutePolicies(ctx, execReq, subPolicyMap)
+		return exec.ExecutePolicy(ctx, execReq, subPolicyMap[subPath])
```
michelvocks, line 289: I guess the above stuff isn't needed anymore since you only pass the policy which already contains all its subpolicies?

michelvocks (approved):
LGTM

pkg/policy/manager.go
roneli: fixed suggestion

PR #167, opened 2021-08-06
Implement `provider download` command.

### LINKED ISSUE #163: Download Providers Only
I am trying to build some docker containers that have my providers pre-downloaded. Is there a way in the CLI to download providers without running the full resource fetch? 

Basically, is there a way to singularly call `DownloadProviders` from the CLI? If not, would it be possible to add a command like that?

irmatov:
This will just verify/download all providers mentioned in the
configuration.

Closes #163

### DIFF
--- a/cmd/provider.go
+++ b/cmd/provider.go
@@ -1,0 +1,19 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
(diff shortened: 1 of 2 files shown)


roneli (approved):
LGTM

PR #168, opened 2021-08-07
hide global flags under options

yevgenypats:
Currently all global flags that are rarely used are being printed to the screen on every help command which makes it very hard to read the relevant commands. 

This PR hides the global flags in all commands help messages and they can be viewed by using `cloudquery options`. Very similar idea is used in `kubectl options` for the same purposes. 

```
% cloudquery options

The following are global options and can be passed to any commands

Global Flags:
      --config string               path to configuration file. can be generated with 'gen config' command (env: CQ_CONFIG_PATH) (default "./config.hcl")
      --dsn string                  database connection string (env: CQ_DSN) (example: 'host=localhost user=postgres password=pass DB.name=postgres port=5432')
      --enable-console-log          Enable console logging
      --enable-file-logging         enableFileLogging makes the framework logging to a file (default true)
      --encode-json                 EncodeLogsAsJson makes the logging framework logging JSON instead of KV
      --log-directory string        Directory to logging to to when file logging is enabled (default ".")
      --log-file string             Filename is the name of the logfile which will be placed inside the directory (default "cloudquery.log")
      --max-age int                 MaxAge the max age in days to keep a logfile (default 3)
      --max-backups int             MaxBackups the max number of rolled files to keep (default 3)
      --max-size int                MaxSize the max size in MB of the logfile before it's rolled (default 30)
      --no-verify                   NoVerify is true registry won't verify the plugins
      --plugin-dir string           Directory to save and load CloudQuery plugins from (env: CQ_PLUGIN_DIR) (default "./.cq/providers")
      --policy-dir string           Directory to save and load CloudQuery policies from (env: CQ_POLICY_DIR) (default "./.cq/policies")
      --reattach-providers string   Path to reattach unmanaged plugins, mostly used for testing purposes (env: CQ_REATTACH_PROVIDERS)
      --skip-build-tables           Skip building tables on run, this should only be true if tables already exist.
  -v, --verbose                     Enable Verbose logging
```

This hides global flags in all commands and can be show by using `cloudquery options`. This is very similar to `kubectl options` design. 

The new top-level help is now nice and clean:

```
CloudQuery CLI

Query your cloud assets & configuration with SQL for monitoring security, compliance & cost purposes.

Find more information at:
        https://docs.cloudquery.io

Usage:
cloudquery [command]

Available Commands:
completion  Generate completion script (run --help for full instructions)
fetch       Fetch resources from configured providers
init        Generate initial config.hcl for fetch command
options     Prints list of global CLI options (applies to all commands)
policy      Download and run CloudQuery policy
version     Print full version info of cloudquery

Additional help topics:

Use "cloudquery [command] --help" for more information about a command.
Use "cloudquery options" for a list of global CLI options.
```


roneli (approved):
Looks good to me, we should get the tests working, it seems like it prints the console info in a terminal situation which is weird

pkg/policy/manager.go
```diff
@@ -24,7 +24,7 @@ const (
 	pathDelimiter    = "/"
 	versionDelimiter = "@"
 
-	cloudQueryOrg = "cloudquery"
+	cloudQueryOrg = "cloudquery-policies"
```
roneli, line 27: I think this broke the tests

PR #176, opened 2021-08-16
Add query type support

irmatov:
Query can have a field named "type" which defaults to "automatic".
Another possible value is "manual".

CLI result output will highlight manual query results with warning
emoji and text.


irmatov:
Closes cloudquery/cloudquery-issues#85

pkg/config/policy.go
```diff
@@ -1,9 +1,28 @@
...
+// Validate validates all contained policies.
+func (pw *PolicyWrapper) Validate() error {
+	for i, p := range pw.Policies {
+		if err := p.Validate(); err != nil {
+			return fmt.Errorf("policy %s (%d): %w", p.Name, i, err)
+		}
+	}
+	return nil
+}
+
```
roneli, line 25: I would put this logic in the policy decode parser, see config_parser.go and config.go, when parsing and decoding we can have logic when decoding the block see provider.go

roneli (changes requested):
Looks good, I would suggest one change with the normalize/validate moving it to part of the policy config parsing phase.

michelvocks (approved):
LGTM

pkg/config/policy.go
irmatov: Done

roneli (approved):
LGTM

PR #178, opened 2021-08-18
Partial Fetch

michelvocks:
Depends on https://github.com/cloudquery/cq-provider-sdk/pull/76

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/golang-migrate/migrate/v4"
 
@@ -250,6 +251,7 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (*FetchRespons
 			c.Logger.Error("failed to create provider plugin", "provider", providerConfig.Name, "error", err)
 			return nil, err
 		}
+		// TODO: move this into an outer function
 		errGroup.Go(func() error {
 			var cfg []byte
 			var partialFetchResults []*cqproto.PartialFetchFailedResource
@@ -259,7 +261,8 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (*FetchRespons
 					return err
 				}
 			}
-			c.Logger.Info("requesting provider to configure", "provider", providerPlugin.Name(), "version", providerPlugin.Version())
+			pLog := c.Logger.With("provider", providerConfig.Name, "alias", providerConfig.Alias, "version", providerPlugin.Version())
+			pLog.Info("requesting provider to configure")
 			_, err = providerPlugin.Provider().ConfigureProvider(gctx, &cqproto.ConfigureProviderRequest{
 				CloudQueryVersion: Version,
 				Connection: cqproto.ConnectionDetails{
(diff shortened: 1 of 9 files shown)


pkg/client/client.go
```diff
@@ -238,6 +242,15 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) error {
 				resp, err := stream.Recv()
 				if err == io.EOF {
 					c.Logger.Info("provider finished fetch", "provider", providerPlugin.Name(), "version", providerPlugin.Version())
+					// Print out partial fetch results if given
+					if len(partialFetchResults) != 0 {
```
roneli, line 246: questioning if this is not better to be added in the FetchCallback? Because I would assume I would actually want to mark  the fetch as from V to ! in case I get partial fetch errors. Basically on each callback I can get an update of Errors rather than error.

pkg/config/provider.go
```diff
@@ -81,6 +86,9 @@ var providerBlockSchema = &hcl.BodySchema{
 		{
 			Name: "alias",
 		},
+		{
```
roneli, line 89: This is good, we should do it per provider not globally.

pkg/ui/console/client.go
```diff
@@ -253,6 +258,28 @@ func buildFetchProgress(ctx context.Context, providers []*config.Provider) (*Pro
 	return fetchProgress, fetchCallback
 }
 
+func buildPrintPartialFetchResult() func(providerName string, results []*cqproto.PartialFetchFailedResource) {
+	return func(providerName string, results []*cqproto.PartialFetchFailedResource) {
+		ui.ColorizedOutput(ui.ColorWarning, "\n\nPartial Fetch Errors for Provider %s:\n\n", providerName)
```
roneli, line 263: maybe add some spacing formating so it will be more uniformed?

roneli (changes requested):
Looks good, not sure about the callback and maybe we should improve how the console UI reacts based on fetchcallback error.

pkg/client/client.go
```diff
@@ -229,7 +232,8 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) error {
 			}
 			c.Logger.Info("provider configured successfully", "provider", providerPlugin.Name(), "version", providerPlugin.Version())
 			c.Logger.Debug("requesting provider fetch", "provider", providerPlugin.Name(), "version", providerPlugin.Version())
-			stream, err := providerPlugin.Provider().FetchResources(gctx, &cqproto.FetchResourcesRequest{Resources: providerConfig.Resources})
+			c.Logger.Info("partial fetch", "enabled", providerConfig.PartialFetchEnabled)
```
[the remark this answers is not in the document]
michelvocks: done

pkg/client/client.go
```diff
@@ -50,6 +52,15 @@ type FetchUpdate struct {
 	ResourceCount uint64
 	// Error if any returned by the provider
 	Error string
+	// PartialFetchResults contains the partial fetch results for this update
+	PartialFetchResults []*cqproto.PartialFetchFailedResource
+}
+
+// FetchFinishCallbackRequest represents a request for the FetchFinishCallback
+type FetchFinishCallbackRequest struct {
```
roneli, line 60: nit: this isn't the request but the result no? FetchSummary perhaps?

roneli (approved):
Looks good, minor nit.

pkg/client/client.go
michelvocks: I thought this is more the Request for the Callback function.

PR #179, opened 2021-08-19
fix unmanaged plugin panic

### LINKED ISSUE #177: Panic on debugging yandex cloud provider
I try to run yandex cloud plugin in debug mode with `config.hcl`:
```
cloudquery {
  plugin_directory = "./cq/providers"
  policy_directory = "./cq/policies"

  provider "yandex-cloud/yandex" {
    source  = "yandex-cloud"
    version = "latest"
  }

  connection {
    dsn = "host=localhost user=postgres password=pass database=postgres port=5432"
  }
}

// All Provider Configurations

provider "yandex" {
  configuration {
    // Optional. Filter as described https://cloud.yandex.com/docs/resource-manager/grpc/folder_service#List
    // folder_filter = ""
    // Optional. If not specified either using all folders accessible.
    // folder_ids = [<CHANGE_THIS_TO_YOUR_FOLDER_ID>]
    cloud_id = "<my_cloud_id>"
  }
  resources = [
    "compute.disks"
  ]
}
```

it panics with:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1b3a1ad]

goroutine 101 [running]:
github.com/cloudquery/cloudquery/pkg/ui/console.buildFetchProgress.func2(0xc000114a90, 0x6, 0x1d96b7a, 0x9, 0xc00045b560, 0xb3, 0x0, 0x0)
        /home/runner/work/cloudquery/cloudquery/pkg/ui/console/client.go:163 +0xad
github.com/cloudquery/cloudquery/pkg/client.(*Client).Fetch.func1(0x0, 0x0)
        /home/runner/work/cloudquery/cloudquery/pkg/client/client.go:257 +0x1069
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc0002a7110, 0xc0000cc930)
        /home/runner/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
        /home/runner/go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:54 +0x66
```

roneli:
fixes #177

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -125,7 +125,7 @@ func newUnmanagedPlugin(providerName string, config *plugin.ReattachConfig) (*un
 		return nil, fmt.Errorf("failed to cast plugin")
 	}
 	return &unmanagedPlugin{
-		name:     providerName,
+		name:     fmt.Sprintf("%s_%s", providerName, providerName),
 		config:   config,
 		client:   client,
 		provider: provider,
--- a/pkg/ui/console/progress.go
+++ b/pkg/ui/console/progress.go
@@ -156,6 +156,15 @@ func (u *Progress) GetBar(name string) *Bar {
 	return bar
 }
 
+func (u *Progress) AbortAll() {
+	u.lock.RLock()
+	defer u.lock.RUnlock()
+	for _, b := range u.bars {
+		b.b.Abort(true)
+	}
+	u.Wait()
+}
+
 func defaultStatusUpdater(u *Bar, _ decor.Statistics) string {
 	return emojiStatus[u.Status]
 }
--- a/pkg/ui/console/progress.go
+++ b/pkg/ui/console/progress.go
@@ -35,7 +35,7 @@ func (b *Bar) SetTotal(total int64, triggerComplete bool) {
 type Progress struct {
 	p       *mpb.Progress
 	bars    map[string]*Bar
-	lock    sync.RWMutex
+	lock    sync.Mutex
 	options *ProgressOptions
 }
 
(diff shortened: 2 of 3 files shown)


pkg/plugin/plugin.go
```diff
@@ -125,7 +125,7 @@ func newUnmanagedPlugin(providerName string, config *plugin.ReattachConfig) (*un
 		return nil, fmt.Errorf("failed to cast plugin")
 	}
 	return &unmanagedPlugin{
-		name:     providerName,
+		name:     fmt.Sprintf("%s_%s", providerName, providerName),
```
michelvocks, line 128: Do we leave this now in? Just want to make sure that this is not an oversight

pkg/ui/console/progress.go
```diff
@@ -156,6 +156,15 @@ func (u *Progress) GetBar(name string) *Bar {
 	return bar
 }
 
+func (u *Progress) AbortAll() {
+	u.lock.RLock()
```
michelvocks, line 160: What happens if (theoretically) another thread accesses this progress bar instance while we are aborting? The reason I ask is if we should use a general lock instead of a read lock?

michelvocks (approved):
LGTM just two questions

pkg/ui/console/progress.go
roneli: I think general lock would be better, i'll swap into that.

pkg/plugin/plugin.go
roneli: not an oversight

PR #181, opened 2021-08-21
Add sslmode=disable to init

roneli:


roneli:
Closes https://github.com/cloudquery/cq-provider-aws/issues/161

PR #183, opened 2021-08-23
add test init command

roneli:


.github/workflows/test.yml
```diff
@@ -46,9 +46,12 @@ jobs:
       - name: Build
         run: go build -v .
 
-      - name: Sanity
+      - name: Sanity Fetch
         run: go run ./main.go fetch --config=internal/test/test_config.hcl --enable-console-log
 
+      - name: Sanity Init
+        run: go run ./main.go init test --config=test_init_config.hcl
```
michelvocks, line 53: What does `--config` do here for init?

michelvocks (approved):
LGTM. Just a small question.

.github/workflows/test.yml
roneli: instead of creating config.hcl I give it a different name, this is to control where the config is created

michelvocks: Gotcha. Shouldn't fetch then also use the same config? The name is currently different.

PR #186, opened 2021-08-30
feat/Support load existing providers

roneli:
support load existing providers into manager control, this removes the need to download a provider.


michelvocks (approved):
LGTM

PR #200, opened 2021-09-16
Check PostgreSQL version

irmatov:
Obtain current PostgreSQL version and print a warning if it is lower
than required (11.0 currently).

Fix a few linter warning in tests.


irmatov:
Tests seem like overkill for such a simple thing but still it's nice to have them anyway :)

pkg/client/pg_version_check.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type connection interface {
+	QueryRow(ctx context.Context, sql string, args ...interface{}) scanner
+	Release()
+}
+
+type pooler interface {
+	Acquire(context.Context) (connection, error)
+}
```
roneli, line 27: I think you can use the QueryExecer interface instead of defining our own interfaces. It should make this code smaller.

roneli (changes requested):
I think we can make this code shorter with using a known pgx Interfaces.

I would also make a wrapper method that gets a pool does Acquire -> Run Check -> Release. rather than this struct which feels a little redundant for just this usage.

pkg/client/pg_version_check.go
irmatov: Can't find that interface. Have a link?

irmatov:
Added a wrapper as suggested.

pkg/client/pg_version_check.go
roneli: https://pkg.go.dev/github.com/jackc/pgtype/pgxtype#Querier

roneli (approved):
LGTM

PR #202, opened 2021-09-19
feat/allow disable delete

### LINKED ISSUE #199: Feature: Delete after Fetch
Currently when CloudQuery fetches a resource (from any provider) it will first delete this resource and only then will start inserting the fetched data. 

This causes the database to appear empty for the specific resource before a fetch is done, which makes it hard to understand if there are no resources or it is just fetching them. This issue can be encountered if running CloudQuery continuously.

To combat this issue/bug the following is needed:
- The delete action should be move to post-fetch instead of pre-fetch (for every resource)
- fetch_date should be added to be able to recognize and delete old data (from previous fetch cycle) after the fetch is complete.
- If fetch failed for specific resource in a specific fetch cycle it shouldn't delete old resource

Follow-up technical tasks will be linked to this issue.

roneli:
Fixes #199

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -36,7 +36,9 @@ var fetchCmd = &cobra.Command{
 
 func init() {
 	fetchCmd.SetUsageTemplate(usageTemplateWithFlags)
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):
LGTM. looks like there is a linting issue.

PR #237, opened 2021-11-01
Print Ignore diags only on --verbose

roneli:
Fixes https://github.com/cloudquery/cq-provider-aws/issues/235


yevgenypats (changes requested):
Suggest not to merge it for now and keep the ignore warning prints as it was now fixed in the AWS (for the byoip resource)

PR #238, opened 2021-11-02
support policy local running

zagronitay:


pkg/client/client.go
```diff
@@ -88,8 +88,11 @@ func (p ProviderFetchSummary) HasErrors() bool {
 
 // PolicyRunRequest is the request used to run a policy.
 type PolicyRunRequest struct {
-	// Args are the given arguments from the policy run command.
-	Args []string
+	// Remote is the given remote policy from the run command.
+	Remote string
+
+	// Local is the given remote policy from the run command.
```
zagronitay, line 94: Will change the comment

michelvocks (approved):
LGTM

spangenberg (approved):
LGTM

yevgenypats (changes requested):
Something feels a bit weird here with the flags. Why two flags are necessary? what happens if someone does `--remote` and `--local`. I feel like one should be the default then. 

let's do remote by default? unless `--local` is specified?

roneli (approved):
LGTM

PR #253, opened 2021-11-08
drift: Cleaner result output with --enable-console-log

disq:


pkg/ui/console/client.go
```diff
@@ -225,10 +225,10 @@ func (c Client) CallModule(ctx context.Context, req ModuleCallRequest) error {
 	}
 
 	if outString, ok := out.Result.(stringer); ok {
-		ui.ColorizedOutput(ui.ColorInfo, "Module output\n%s\n", outString.String())
+		ui.ColorizedOutput(ui.ColorInfo, "Module output: \n%s\n", outString.String())
```
disq, line 228: The additional colon and space is added so that when `ui.ColorizedOutput` strips the newline (`if !IsTerminal() {` block) we have some kind of a separation between the first line of the result and the string.

roneli (approved):
LGTM

PR #255, opened 2021-11-08
drift: Exit with nonzero code if drift detected

disq:
Exit with exit code 1 if any of these are detected:
- drifted resources
- extra resources
- missing resources

### DIFF
--- a/cmd/util.go
+++ b/cmd/util.go
@@ -3,12 +3,17 @@ package cmd
 import (
 	"os"
 
+	"github.com/cloudquery/cloudquery/pkg/ui/console"
 	"github.com/spf13/cobra"
 )
 
 func handleError(f func(cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) {
 	return func(cmd *cobra.Command, args []string) {
 		if err := f(cmd, args); err != nil {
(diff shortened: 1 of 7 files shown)


cmd/drift.go
```diff
@@ -53,12 +54,20 @@ var (
...
 
-			return c.CallModule(ctx, console.ModuleCallRequest{
+			err = c.CallModule(ctx, console.ModuleCallRequest{
 				Name:          driftModuleID,
 				Params:        driftParams,
 				ModConfigPath: driftConfigPath,
 				OutputPath:    driftOutputPath,
 			})
+			if err != nil {
+				if ee, ok := err.(*console.ExitCodeError); ok {
```
disq, line 64: This code could be simplified if we decide we only need just one nonzero exit code.

spangenberg:
@disq can you add the handleError from #257, guess you can and should still use the os.Exit as you have it for specific codes.

disq:
> @disq can you add the handleError from #257, guess you can and should still use the os.Exit as you have it for specific codes.

Already added 4 minutes before your comment 👀

spangenberg (approved):
LGTM

roneli (approved):

PR #257, opened 2021-11-08
Fixes #250 - Do not print unrelated help string on cmd error

### LINKED ISSUE #250: On CloudQuery errors (not args errors) - don't print help message
Currently when running `cloudquery fetch` or `cloudquery policy run` or anything else with **correct** arguments but cloudquery returns an error to `RunE` it will print the help message which just add unrelated info to the screen and makes it difficult to find the real error message

Looks like there is a solution suggestion in this cobra [issue](https://github.com/spf13/cobra/issues/340#issuecomment-374617413)

spangenberg:
IMHO the `SilenceUsage` option isn't the real solution to this kind of problem as it bringt other drawbacks.


disq (approved):
Looks ok but this will also affect #255

PR #261, opened 2021-11-09
update old policy documentation

yevgenypats:
Update old policy documentation. It still needs more work and basically it should point to the right section in the docs which we are still missing but this should be a good first iteration.


roneli (approved):
LGTM

PR #264, opened 2021-11-09
Handle --fail-on-violation flag higher up

irmatov:
Otherwise there is a bug where policy result is not saved to output file
on violation.


roneli (approved):
good catch

PR #265, opened 2021-11-09
Set nonzero exit code if module returned an error

disq:
This is different from the case of drift module detecting a drift.


fdistorted (approved):

roneli (approved):
Not sure what I see, but it removed processed so I like :)

PR #266, opened 2021-11-09
Handle subresources

disq:
Subresources are (often multiple) related resources contained in a terraform.Instance as attributes, under an array-typed key.

Issues:
- Some arrays are in fact "sets", ie. the order doesn't matter. We currently can't include them as PKs (identifiers) ~and can't reliably match them in deep mode either~. To fix this I tried sorting them handling TF values, but Postgres returns them unsorted so that didn't help... Looked for a function to sort an `ARRAY` type in PG, but what's recommended is to use `UNNEST()` and then do a subquery to order and `ARRAY_AGG`, which seemed to be too much. (`aws_cloudfront_distribution_cache_behaviours`)
- "root." syntax and handling in `parseTerraformInstance` is not very nice, need better handling/jsonpath to do this for us ideally...
- ~Some type issues (int vs. string typed int) throws off the deep mode comparisons, can be seen in `aws_cloudfront_distribution_custom_error_responses.response_code` (not handled) and `aws_cloudtrail_trail_event_selectors.include_management_events` (handled using `CASE WHEN`)~ fixed

### DIFF
--- a/pkg/module/drift/drift.go
+++ b/pkg/module/drift/drift.go
@@ -270,7 +270,7 @@ func handleIdentifiers(identifiers []string) (exp.Expression, error) {
 		}
 
 		if !usingVariable && !strings.Contains(id, ".") {
-			id = "c." + id
+			id = "c." + `"` + id + `"`
 		}
 
 		concatArgs = append(concatArgs, id, "'"+idSeparator+"'")
--- a/pkg/module/drift/drift.go
+++ b/pkg/module/drift/drift.go
@@ -150,7 +150,7 @@ func queryIntoResourceList(ctx context.Context, logger hclog.Logger, conn *pgxpo
 		var pgErr *pgconn.PgError
 		if errors.As(err, &pgErr) && pgErr.Code == pgerrcode.UndefinedTable {
 			// ERROR: relation %q does not exist
-			return nil, fmt.Errorf("cloud provider tables don't exist: Did you run `cloudquery fetch`?%q", pgErr.TableName)
+			return nil, fmt.Errorf("cloud provider tables don't exist: Did you run `cloudquery fetch`? %w", pgErr)
 		}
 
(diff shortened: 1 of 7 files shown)


pkg/module/drift/drift.go
```diff
@@ -248,30 +248,39 @@ func handleFilters(sel *goqu.SelectDataset, res *ResourceConfig) *goqu.SelectDat
 
 var idRegEx = regexp.MustCompile(`(?ms)^\$\{sql:(.+?)\}$`)
 
-func handleIdentifier(identifiers []string) (exp.Expression, error) {
-	switch l := len(identifiers); {
-	case l == 0:
+const idSeparator = "|"
+
+func handleIdentifiers(identifiers []string) (exp.Expression, error) {
```
roneli, line 253: would document this function and its inner logic, I have some changes I think that would be better to implement such as defining the table and doing table.Col instead of c + "." + name etc', but we can do this in a later PR.

pkg/module/drift/terraform.go
```diff
@@ -93,45 +173,52 @@ func driftTerraform(ctx context.Context, logger hclog.Logger, conn *pgxpool.Conn
 	resData := resources[resName]
 	deepMode := runParams.ForceDeep || (resData.Deep != nil && *resData.Deep)
```
roneli, line 174: is there a point for Deep to be nil ? shouldn't it just be False/True?

roneli (approved):
Looks okay, had a few comments.

pkg/module/drift/terraform.go
disq: @roneli It could be unset so that would make it nil. In theory you could only enable this for some of the resources and not all. If it's unset then it would fall back to the wild resource entry. (which can again be `nil`, then which would default to false until you run with the cli --deep flag...)

CLI always overrides the config, but since we need the set/unset logic under a non-wild resource...

PR #269, opened 2021-11-10
Feat/policy config

zagronitay:


pkg/policy/manager.go
```diff
@@ -87,49 +88,54 @@ func NewManager(policyDir string, pool *pgxpool.Pool, logger hclog.Logger) Manag
...
-	case 2:
-		policy.Organization = orgRepoSplit[0]
-		policy.Repository = orgRepoSplit[1]
-	case 1:
-		policy.Repository = orgRepoSplit[0]
-		policy.Organization = cloudQueryOrg
-	default:
-		return nil, fmt.Errorf("invalid policy path. Repository name malformed: %s", args[0])
+	var totalQueriesToRun = policyWrapper.getQueriesCount()
+	var finishedQueries = 0
```
roneli, line 100:
```suggestion
	var (
	 totalQueriesToRun = policyWrapper.getQueriesCount()
	finishedQueries = 0
	)
```

pkg/policy/manager.go
```diff
@@ -87,49 +88,54 @@ func NewManager(policyDir string, pool *pgxpool.Pool, logger hclog.Logger) Manag
...
-		policy.RepositoryPath = args[1]
+	executor := NewExecutor(conn, m.logger)
+
+	var selector []string
+	if execReq.Policy.SubPath != "" {
+		selector = strings.Split(execReq.Policy.SubPath, "/")
 	}
-	return policy, nil
+
+	return executor.ExecutePolicies(ctx, progressUpdate, execReq, policyWrapper.Policies, selector)
```
roneli, line 134:
```suggestion
	return NewExecutor(conn, m.logger).ExecutePolicies(ctx, progressUpdate, execReq, policyWrapper.Policies, selector)
```

pkg/policy/manager.go
```diff
@@ -87,49 +88,54 @@ func NewManager(policyDir string, pool *pgxpool.Pool, logger hclog.Logger) Manag
...
+				QueriesCount:    totalQueriesToRun,
+				Error:           "",
+			})
+		}
 	}
 
-	// Parse repository path if given
-	if len(args) == 2 {
-		policy.RepositoryPath = args[1]
+	executor := NewExecutor(conn, m.logger)
```
roneli, line 127: I think we can pass the updatecallback, and the execute as well perhaps when initializing the Executor, seems like it will be reduce the amount of arguments passed

pkg/policy/execute.go
```diff
@@ -2,17 +2,44 @@ package policy
 
 import (
```
roneli, line 3: nit: General rule of thumb we want private functions to be defined after public functions, the order here is not by convention

pkg/policy/manager.go
```diff
@@ -237,49 +228,71 @@ func (m *ManagerImpl) RunPolicy(ctx context.Context, execReq *ExecuteRequest) (*
...
-		return nil, fmt.Errorf("failed to load policy file: %#v", diags.Error())
+	// If repository path was specified, also check if that exists
+	policyFolder = repoFolder
+	// if remotePolicy.Repository != "" {
+	//	policyFolder = filepath.Join(repoFolder, remotePolicy.Repository)
+	//	if info, err := osFs.Stat(policyFolder); err != nil || !info.IsDir() {
+	//		return nil, fmt.Errorf("could not find policy '%s' in the folder '%s'. Try to download the policy first", orgPolicyStr, p.RepositoryPath)
+	//	}
+	//	m.logger.Debug("internal repo folder set", "path", policyFolder)
+	//}
```
roneli, line 277: Is this code relevant ? if not I suggest to remove.

pkg/policy/policy.go
```diff
@@ -1,4 +1,4 @@
-package config
+package policy
 
 type PolicyWrapper struct {
```
roneli, line 3: Not sure if policy wrapper is the best name, but lacking any better suggestion for now.

pkg/policy/policy.go
```diff
@@ -1,4 +1,4 @@
-package config
+package policy
```
roneli, line 1: can't comment on unchanged files but PolicyProvider -> Provider since the package name denotes the namespace i.e policy.Provider instead of policy.PolicyProvider

pkg/policy/policy.go
roneli: maybe defined a type called Policies i.e 
```suggestion
type Policies []*Policy
```

roneli (changes requested):
First Review for Policy config.
Looks great overall, commented some suggestions and changes, but the structure is great and very clean. I have a feeling that the logger should be used a little more to log more information for debugging / info on progress etc'

pkg/ui/console/client.go
```diff
@@ -494,3 +523,89 @@ func loadConfig(path string) (*config.Config, error) {
 	}
 	return cfg, nil
 }
+
+func buildPolicyRunProgress(ctx context.Context, policies []*config.Policy, failOnViolation bool) (*Progress, policy.UpdateCallback) {
```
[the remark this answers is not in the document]
zagronitay: No sure i understand you.
We do have counter of running queries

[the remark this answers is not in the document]
roneli: Similar to the progress in fetch where it has a count of failed queries and changes the Progress -> Warning if there is 1 or more failures that returned by the callback

pkg/client/client.go
```diff
@@ -550,83 +543,122 @@ func (c *Client) DropProvider(ctx context.Context, providerName string) error {
...
+			c.Logger.Info("Writing policy to output directory", "args", policyConfig)
+			err = policy.GenerateExecutionResultFile(result, req.OutputDir)
+
+			if err != nil {
+				if handleError(err) {
+					return results, nil
+				} else {
+					continue
+				}
+			}
```
roneli, line 657:
```suggestion
			err = policy.GenerateExecutionResultFile(result, req.OutputDir)
			if handleError(err) != nil; err != nil {
				return results, nil
			}
			
```

pkg/client/client.go
```diff
@@ -710,6 +742,46 @@ func (c *Client) getProviderConfig(providerName string) (*config.RequiredProvide
 	return providerConfig, nil
 }
 
+func GetPoliciesToRun(args []string, configPolicies []*config.Policy, policyName string) ([]*config.Policy, error) {
```
roneli, line 745: pehaps change name to FilterPolicies

roneli (commented):
Added a few more comments

pkg/policy/manager.go
```diff
@@ -386,23 +390,147 @@ func (p *Policy) checkoutPolicyVersion(repoFolder string) error {
...
+
+	policy := &RemotePolicy{}
+
+	if strings.HasPrefix(args[0], "http") {
+		// Parse the policy from a URL
+		return ParsePolicyFromSource(&config.Policy{
+			Type:   config.Remote,
+			Source: args[0],
+		})
+	} else {
```
roneli, line 444: we can remove this else imo

roneli (approved):
LGTM minor nitpick

PR #270, opened 2021-11-11
Clean up module profile handling

disq:
`--profile` and profile selection is now done by the console client. (Still, we don't load/parse it unless we're invoking a module)


roneli (approved):
LGTM

PR #272, opened 2021-11-12
Enable environment variables in drift config

disq:
ie. set `CQ_VAR_ACCOUNT_ID=123` and use `ACCOUNT_ID` in the `aws` block as the variable.


roneli (approved):
LGTM, would maybe also add a test on drift config parser

PR #273, opened 2021-11-12
drift: Fix for wild resource on wild provider application for user supplied config

disq:
It's in the docs but we forgot to actually implement it


spangenberg (approved):
LGTM

PR #276, opened 2021-11-13
Update Quick Start

patrickpierson:
Added 

```
?sslmode=disable
```

to fetch command to make the Quick Start process easier to complete for a new person because the process currently errors out saying 

```
pq: SSL is not enabled on the server
```


roneli (approved):
Good catch, Thanks for the fix!

PR #280, opened 2021-11-15
telemetry/analytics

disq:
- [x] Set up collector/receiver end
- [x] Hostname and connection details


spangenberg:
So far I really like it :)

pkg/plugin/manager.go
```diff
@@ -76,7 +82,16 @@ func (m *Manager) DownloadProviders(ctx context.Context, providers []*config.Req
 			return err
 		}
 		m.providers[providerName] = details
+		traceData[i] = providerName + "@" + details.Version
```
roneli, line 85: would add a method to Plugin Interface to get Provider Full name which should include details.Provider + details.Version, this is also regarding line 76

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,297 @@
...
+		return "", err
+	}
+	return id, nil
+}
+
+func (c *Client) defaultExporter(ctx context.Context) (trace.SpanExporter, error) {
+	return otlptracegrpc.New(
+		ctx,
+		otlptracegrpc.WithInsecure(), // TODO change
+		otlptracegrpc.WithEndpoint("localhost:4317"), // TODO change. env var?
```
roneli, line 256: I would say this should be hardcoded, maybe make it configurable in the future so we can support proxies

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,297 @@
...
+	return resource.New(ctx,
+		resource.WithTelemetrySDK(),
+		resource.WithProcessRuntimeName(),
+		resource.WithProcessRuntimeVersion(),
+		resource.WithProcessRuntimeDescription(),
+		resource.WithAttributes(attr...),
+	)
+}
+
+func (c *Client) cookie() (string, error) {
```
roneli, line 222: document the cookie function :)

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,297 @@
...
+	exporter trace.SpanExporter
+	closer   io.Closer
+
+	logger hclog.Logger
+	fs     afero.Afero
+	err    error
+
+	version, commit, buildDate string
+
+	debug, disabled bool
```
roneli, line 40: please document the parameters

roneli (changes requested):
Looking good, added a few comments

pkg/plugin/manager.go
disq: The only problem is `registry.ProviderDetails` doesn't seem to implement the `Plugin` interface.

internal/telemetry/telemetry.go
disq: Done

internal/telemetry/telemetry.go
disq: Done

cmd/util.go
```diff
@@ -1,22 +1,82 @@
...
+
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
-func handleError(f func(cmd *cobra.Command, args []string) error) func(cmd *cobra.Command, args []string) {
+func handleError(f func(context.Context, *console.Client, *cobra.Command, []string) error) func(cmd *cobra.Command, args []string) {
```
roneli, line 21: I liked this fix, makes command much cleaner, not sure handleError is correct anymore since it does more, maybe handleCommand / executeCommand, you add the handle error inside this function, or just keep it as is in the current function.

roneli (commented):
LGTM, I won't approve yet, unless you don't have any more changes, ill have another go over it to double check

cmd/util.go
disq: it's now `handleCommand`

pkg/plugin/manager.go
roneli: So add a method to `registry.ProviderDetails` ?

cmd/telemetry.go
```diff
@@ -0,0 +1,36 @@
...
+		opts = append(opts, telemetry.WithDisabled())
+	}
+
+	if viper.GetBool("debug-telemetry") {
+		opts = append(opts, telemetry.WithDebug())
+	}
+
+	if viper.GetBool("inspect-telemetry") {
+		fs := afero.NewOsFs()
+		f, err := fs.Create("cq-telemetry.txt")
```
roneli, line 28: maybe add a log line explaining that cq-telemetry was created locally ?

roneli (approved):
LGTM

cmd/telemetry.go
disq: Done

pkg/plugin/manager.go
disq: It has to be the whole interface, so 5/6 methods. It just reads the struct fields without utilizing any interface abstraction. I don't want to touch that in this PR.

spangenberg (approved):

Issue #281, opened 2021-11-16
Feature request: Allow regex patterns for drift configuration (both local files and s3 bucket + keys)


eistrati:
Right now the drift configuration example looks like this:
```hcl
modules {
  drift "drift-example" {
    terraform {
      backend = "local" # "local" or "s3"
      files   = ["/path/to.tfstate"]
      /*
      bucket   = "<terraform state bucket>"
      keys     = ["<terraform state key>"]
      region   = "us-east-1"
      role_arn = ""
      */
    }
  }
}
```

Please improve the drift configuration to allow some kind of regex patterns. For example:
```hcl
modules {
  drift "drift-example" {
    terraform {
      backend = "local" # "local" or "s3"
      files   = ["/path/to/**/terraform.tfstate"]
      /*
      bucket   = "<terraform state bucket>"
      keys     = ["terraform/my_project/**/terraform.tfstate"]
      region   = "us-east-1"
      role_arn = ""
      */
    }
  }
}
```

Our use cases include multiple `terraform.tfstate` files stored in multi-layers folders structure (or multi-layers s3 bucket prefixes) and enumerating exactly each file in the drift configuration doesn't scale well.

roneli:
Hi @eistrati we support this feature from [v0.16.0](https://github.com/cloudquery/cloudquery/releases/tag/v0.16.0). Thank you for the feature request :)

eistrati:
Thank you for quick turn around! Looking forward to give it a try :)

PR #282, opened 2021-11-16
drift: Glob support in source configuration

disq:
To keep things simple, only partial support (`*` and `**`) in the S3 backend (no character ranges with `[` and `]`)
Local filesystem glob matching is done using `afero.Glob`, which still uses `filepath.Match` underneath so it should be compatible with `filepath.Glob`.

Implements #281.


roneli (approved):
LGTM

PR #291, opened 2021-11-18
Improve error message if config could not be read

### LINKED ISSUE #277: Confusing Error when config.hcl doesn't exist
### Describe the Bug
When config.hcl doesn't exist the following is printed

```
Configuration Error Diagnostics:
❌ Failed to read file; The file "./config.hcl" could not be read.
bad configuration
```

### Expected Behavior
Should print what file it tried to read and that it doesn't exist (not bad configuration)

### CloudQuery Version
Version: 0.15.10
Commit: 00e576804344a17360d3ad94f79f6b30a4118c78
Date: 2021-11-14 09:54:28.360668 +0200 IST m=+0.005966954
### Steps to Reproduce
```
cloudquery fetch # or any other command
```

spangenberg:
Fixes #277

### DIFF
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -36,7 +36,7 @@ func CreateClient(ctx context.Context, configPath string, opts ...client.Option)
 	cfg, ok := loadConfig(configPath)
 	if !ok {
 		// No explicit error string needed, user information is in diags
-		return nil, HandledError{}
+		return nil, &ExitCodeError{ExitCode: 1}
 	}
 	return CreateClientFromConfig(ctx, cfg, opts...)
 }
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -16,12 +16,13 @@ import (
 	"github.com/spf13/viper"
 	"github.com/vbauerster/mpb/v6/decor"
 
+	"github.com/cloudquery/cq-provider-sdk/cqproto"
+
 	"github.com/cloudquery/cloudquery/pkg/client"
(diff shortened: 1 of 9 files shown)


pkg/ui/console/client.go
```diff
@@ -30,9 +31,9 @@ type Client struct {
 }
 
 func CreateClient(ctx context.Context, configPath string, opts ...client.Option) (*Client, error) {
-	cfg, err := loadConfig(configPath)
-	if err != nil {
-		return nil, err
+	cfg, ok := loadConfig(configPath)
+	if !ok {
+		return nil, fmt.Errorf("")
```
disq, line 36: Add comment to explain why we're using empty error string? or maybe use a typed error, or even `console.ExitCodeError`

disq (approved):
LGTM with nit

pkg/ui/console/client.go
roneli: I agree we should maybe define console error codes, this important to diff between an actual error or a failure for example fetch failed/drift scan returned unmanaged resources/policy fails. Maybe out of scope from this PR. we can open a new issue on that.

roneli (commented):
I suggest adding a test to verify this use case and we get the expected error

pkg/ui/console/client.go
disq: @roneli console.ExitCodeError already exists (but it takes an `int`, not a special-typed exitcode or const) my point was to use it instead of an empty error message, since the error message is already handled in the caller as a diag.

Issue #228, opened 2021-10-24
Feature Request: Historical records


somerska:
It would be significantly more useful/powerful if I can see how resources are trending. This feature request is for CQ to not delete records when fetch is ran (or at least allow for a flag to control the behavior) and also to include a column for a date time field on the table(s).   Please thumb up if you would like to see this feature in CloudQuery.

roneli:
History support added from v0.17.0! Check out how to configure [here](https://docs.cloudquery.io/docs/cli/history/configuration)

Issue #55, opened 2021-02-09
Feature Request: Keeping History


James-Quigley:
In the current implementation, CloudQuery wipes tables for each selected resource during execution. This prevents tracking the state of an object throughout history. 

One potential solution would be to include a time stamp column on each resource that is the same for all resources tracked in a given execution of CloudQuery. You could then query by all resources from a given snapshot, or query a single resource over time. 

Then it would be great to include options such as only keeping _x_ versions of a resource, or only keeping resources less than _y_ hours old. 

If primary keys are set on resource ID, they might need to be shifted to be a joint key of resource ID and time stamp. This would mean that you couldn't have two instance of CloudQuery fetch running in the same timestamp, but that likely isn't an issue

dan-compton:
> In the current implementation, CloudQuery wipes tables for each selected resource during execution.   This prevents tracking the state of an object throughout history.

Effectively this is equivalent to out-of-order, parallal, causal A/B integration test execution and maybe a desired outcome.  You have some number of tests that must be executed by a user prior to the experiments conclusion which happens within some time window t_T.  

I do not know if there is a threshold outcome for the completion of the execution (some minimal number of resources which must be modified or if there is a time limit and how if ever that will tend towards a live stream.  That almost sounds on its face like a problem for kstreams and ksql if it weren't for the overhead of fragmentation.  I wouldn't want to be in the business of managing individual tables and joins for an innumerable number of resource types.  I would think storing the immutable state entity which is a pointer to a prior immutable state entity {resources and ops and temporal data} (or resource primitive) primitive and the operational data would be enough.

Not an expert on that though (the query aspect).  This delta lake concept seems reasonable: https://databricks.com/blog/2020/05/27/modernizing-risk-management-part-1-streaming-data-ingestion-rapid-model-development-and-monte-carlo-simulations-at-scale.html.   You could get time-windowed joins and alternate universes if you fed it with kafka and changed causality.

dan-compton:
Or https://www.confluent.io/blog/real-time-analytics-with-kafka-and-pinot/

Rackme:
I used this little workaround for now If needed, cloning the DB for each version.
```
str_date_postgres=$(date '+%Y_%m_%d')
docker-compose exec postgres psql -U postgres -c "CREATE DATABASE dump_$str_date_postgres WITH TEMPLATE postgres;"
# Prepare new postegre database (to avoid duplicates)
docker-compose exec postgres dropdb -U postgres postgres
[...]

PR #300, opened 2021-11-26
Add support for history

roneli:
Related #55 #228


pkg/client/client.go
```diff
@@ -759,49 +768,28 @@ func (c *Client) getProviderConfig(providerName string) (*config.RequiredProvide
...
-		policyConfig.SubPath = subPath
-		if err != nil {
-			return nil, err
-		}
-		policies = append(policies, policyConfig)
-	} else {
-		policies = configPolicies
+func (c *Client) setupTableCreator(ctx context.Context) error {
+	if c.TableCreator != nil {
+		c.Logger.Debug("table creator already set")
```
disq, line 773: Warning level instead? Shouldn't happen anyway

pkg/client/history/table.go
```diff
@@ -0,0 +1,144 @@
...
+	_, err := conn.Exec(ctx, "SELECT history.build_trigger($1, $2, $3);", p.Name, t.Name, c.Name)
+	if err != nil {
+		return fmt.Errorf("failed to create trigger: %w", err)
+	}
+	return nil
+}
+
+func (h TableCreator) findParentIdColumn(t *schema.Table) *schema.Column {
+	for _, c := range t.Columns {
+		if strings.HasSuffix(c.Name, "cq_id") && c.Name != "cq_id" {
```
disq, line 98: Looks like a polyfill for older providers that don't support `c.Meta()`, so maybe move this after the Meta check and add comment to clarify?

Also, maybe go over `t.Columns` two times, so that if the `foo_cq_id` isn't the parent reference but meta is supported, we don't end up choosing the wrong column just based on the name (first one we encounter basically)

pkg/client/history/table.go
```diff
@@ -0,0 +1,144 @@
...
+	return nil
+}
+
+func (h TableCreator) buildTableSQL(table *schema.Table, parentTable *schema.Table) (string, error) {
+	// Build SQL to create a table.
+	ctb := sqlbuilder.CreateTable(fmt.Sprintf("history.%s", table.Name)).IfNotExists()
+	var uniques []string
+	for _, c := range schema.GetDefaultSDKColumns() {
+		if c.CreationOptions.Unique {
+			ctb.Define(c.Name, schema.GetPgTypeFromType(c.Type))
```
disq, line 114: Simplify by moving ctb.Define to a single line before the if on line 113

pkg/ui/console/client.go
```diff
@@ -105,7 +106,8 @@ func (c Client) Fetch(ctx context.Context, failOnError bool) error {
 	request := client.FetchRequest{
 		Providers:         c.cfg.Providers,
 		UpdateCallback:    fetchCallback,
-		DisableDataDelete: viper.GetBool("disable-delete"),
+		DisableDataDelete: true, // TODO: use bool + also if history block is enable always turn into true
```
disq, line 109: It's already like this in Fetch code, maybe remove TODO? Rename field to `PreferDisableDataDelete`? Add comment? add `c.HistoryCfg != nil` here as well?

disq (commented):
Minor nits

pkg/client/client.go
roneli: it can happen if set as a functional option, so maybe keep it as debug.

pkg/ui/console/client.go
roneli: Missed this one good catch.

roneli: Considering removing disable delete to be true by default.

pkg/client/history/table.go
```diff
@@ -67,16 +67,20 @@ func (h TableCreator) CreateTable(ctx context.Context, conn *pgxpool.Conn, t *sc
 	return nil
 }
 
-func (h TableCreator) createHyperTable(ctx context.Context, t *schema.Table, conn *pgxpool.Conn) error {
+func (h TableCreator) createHyperTable(ctx context.Context, t *schema.Table, p *schema.Table, conn *pgxpool.Conn) error {
```
disq, line 70: `t, p *schema.Table`

pkg/client/client.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/cloudquery/cloudquery/pkg/client/history"
```
disq, line 12: nit: import order/goimports

pkg/client/history/table.go
```diff
@@ -0,0 +1,149 @@
...
+	ctb := sqlbuilder.CreateTable(fmt.Sprintf("history.%s", table.Name)).IfNotExists()
+	var uniques []string
+	for _, c := range schema.GetDefaultSDKColumns() {
+		ctb.Define(c.Name, schema.GetPgTypeFromType(c.Type))
+		if c.CreationOptions.Unique {
+			uniques = append(uniques, c.Name)
+		}
+	}
+	// TODO check fetch_date not defined
+	ctb.Define("fetch_date", schema.GetPgTypeFromType(schema.TypeTimestamp))
```
disq, line 127: Maybe prefix `fetch_date`? As in `cq_fetch_date`? So that the TODOs are removed and there's no chance of a collusion.

pkg/config/config.go
```diff
@@ -3,6 +3,8 @@ package config
 import (
 	"fmt"
 
+	"github.com/cloudquery/cloudquery/pkg/client/history"
+
```
disq, line 7: Rogue newline (or in the wrong place)

pkg/config/config_parser.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/creasty/defaults"
+
```
disq, line 8: The war on rogue newlines.

pkg/policy/filter.go
```diff
@@ -0,0 +1,52 @@
+package policy
```
disq, line 1: erm

pkg/ui/console/client.go
```diff
@@ -152,7 +153,7 @@ func (c Client) DownloadPolicy(ctx context.Context, args []string) error {
 func (c Client) RunPolicies(ctx context.Context, args []string, policyName, outputDir, subPath string, stopOnFailure, skipVersioning, failOnViolation, noResults bool) error {
 	c.c.Logger.Debug("Received params: args: %v, policyName: %s, outputDir: %s, stopOnFailure: %v, skipVersioning: %v, failOnViolation: %v, noResults: %v", args, policyName, outputDir, stopOnFailure, skipVersioning, failOnViolation, noResults)
 
-	policiesToRun, err := client.FilterPolicies(args, c.cfg.Policies, policyName, subPath)
+	policiesToRun, err := policy.FilterPolicies(args, c.cfg.Policies, policyName, subPath)
```
disq, line 156: Move to the other PR I suppose

disq (changes requested):
LGTM overall

pkg/ui/console/client.go
roneli: I actually after going further into that these functions belong to the console, since its console argument parsing its only used here. Should move to `ui/console/policy.go` I will refactor that after I merge this PR and #322

pkg/policy/filter.go
roneli: ?

disq: A subtle way of saying unrelated change?

disq (approved):
LGTM

PR #315, opened 2021-11-30
policy parser: Don't assume we have a valid Policy entry

disq:


pkg/policy/parser.go
```diff
@@ -98,7 +98,7 @@ func decodePolicyContent(labels []string, content *hcl.BodyContent, ctx *hcl.Eva
 		diags = append(diags, contentDiags...)
 		inner, innerDiags := decodePolicyContent([]string{""}, innerContent, ctx, r)
 		diags = append(diags, innerDiags...)
-		policy.Policies = append(policy.Policies, inner.Policies[0])
+		policy.Policies = append(policy.Policies, inner.Policies...)
```
roneli, line 101: I did this because I assumed, and still assume that when you file a policy it has a root policy i.e there isn't more than one block of policy in a single file, although this is more safer.

roneli: And ofc possible to do multiple policy blocks in a policy.hcl file

disq: the problem was the policy had an error... so we could keep inner.Policies[0] but add a length check, and also if length > 1 then add an error (too many policies, unhandled policies, etc)

roneli (approved):

PR #319, opened 2021-11-30
Telemetry: Expose diagnostic counts on fetch

disq:
Something like this? I can't really get them to fail in this particular way so can't really test the output.

Supposed to generate `int` attributes like
`fetch.diag.error.access.aws`
`fetch.diag.warning.throttle.aws`
etc. so:
`fetch.diag.<severity lowercase>.<type lowercase>.<provider>`

And total attributes per severity/provider pair, like
`fetch.diag.error.total.aws`
`fetch.diag.warning.total.aws`
`fetch.diag.ignore.total.aws`


pkg/client/client.go
```diff
@@ -93,6 +94,46 @@ func (p ProviderFetchSummary) HasErrors() bool {
 	return false
 }
 
+func (p ProviderFetchSummary) Attributes() []attribute.KeyValue {
```
roneli, line 97: Maybe you should allow conversion of Provider into a map, and create a generic function to convert map -> attribute.KeyValue perhaps?

roneli (changes requested):
LGTM overall, had one question/discussion, before we approve and merge.

pkg/client/client.go
disq: @roneli done

roneli (approved):
LGTM

PR #326, opened 2021-12-06
Fix provider failure cancel

### LINKED ISSUE #325: Provider fetch failure cancels other provider fetches
### Describe the Bug
if multiple providers are executed, and one provider fails i.e returns an error, all providers get cancelled. 

### Expected Behavior

Failed provider should fail, other provider should finish fetching, errors printed in the end.

### CloudQuery Version
v0.16.2

### Steps to Reproduce

Config hcl with two providers
```
cloudquery fetch
```

roneli:
Fix #325
Fixed bug where total resource count wasn't added up.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -322,7 +322,7 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 	c.Logger.Info("received fetch request", "disable_delete", request.DisableDataDelete, "extra_fields", request.ExtraFields)
 
 	fetchSummaries := make(chan ProviderFetchSummary, len(request.Providers))
-	errGroup, gctx := errgroup.WithContext(ctx)
+	errGroup, _ := errgroup.WithContext(ctx)
 	for _, providerConfig := range request.Providers {
 		providerConfig := providerConfig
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -322,7 +322,7 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 	c.Logger.Info("received fetch request", "disable_delete", request.DisableDataDelete, "extra_fields", request.ExtraFields)
 
 	fetchSummaries := make(chan ProviderFetchSummary, len(request.Providers))
-	errGroup, gctx := errgroup.WithContext(ctx)
+	errGroup, _ := errgroup.WithContext(ctx)
```
disq, line 325: Maybe add a comment regarding why we're ignoring the errgroup ctx? It's apparent now we're fixing that specific issue, but future generations will forget this and this bug might resurface. A test would've been better but it seems to be buried deep within fetch and would get complicated fast.

zagronitay (approved):
LGTM

disq (approved):
LGTM

pkg/client/client.go
roneli: Eventually this will need a refactor here, as if we aren't using gctx, I don't see much reason for the errGroup to be honest.

Issue #303, opened 2021-11-26
Output UUID from Query/Policy


bbernays:
### Describe the Bug
When executing a policy and saving the output, Cloudquery does not correctly convert the [][]interface{} to strings for printing/savings when the underlying datatype is a UUID.

in the output file it is an array of ints. If you cast the UUID column in SQL then the correct value will be printed

Although I have not tested it this behavior might occur with other complex data types as well 

### Expected Behavior

cloudquery would print out uuid as a string

bbernays:
Fixed in #332

PR #332, opened 2021-12-07
Fix UUID Output

bbernays:
Retain UUID type information when returning values from PGX
#303


pkg/policy/execute.go
```diff
@@ -203,6 +212,23 @@ func (e *Executor) executeQuery(ctx context.Context, q *Query) (*QueryResult, er
...
+			if values[i] == nil {
+				continue
+			}
+			switch val.DataTypeOID {
+			case 2950:
+				fmt.Printf("val: %v\n", values[i])
+				t := uuid.FromBytesOrNil(ByteSlice(values[i].([16]uint8)))
+				fmt.Printf("Underlying Type: %T, %s\n", t, t.String())
+				values[i] = t
+			}
```
bbernays, line 228: I'm not sure if there is a more generic way where we wouldn't have to re-implement for every type

pkg/policy/execute.go
```diff
@@ -203,6 +212,23 @@ func (e *Executor) executeQuery(ctx context.Context, q *Query) (*QueryResult, er
...
+
+		fmt.Printf("Underlying Type: %+v\n", values)
+		for i, val := range data.FieldDescriptions() {
+			// val.DataTypeOID
+			if values[i] == nil {
+				continue
+			}
+			switch val.DataTypeOID {
+			case 2950:
+				fmt.Printf("val: %v\n", values[i])
```
bbernays, line 224: This is in for local debugging/dev will be removed when pr is complete

pkg/client/uuid.go
```diff
@@ -0,0 +1,273 @@
+package client
```
disq, line 1: If this was copied from pgtype, link to the original file + our PR maybe?

bbernays: Added a link to the file on Github.
Also added comments to the two places that changed `\\CQ-Change:`

bbernays: If this is not what you were looking for then let me know!

pkg/client/uuid.go
```diff
@@ -0,0 +1,276 @@
...
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/jackc/pgtype"
+)
+
+// Code came from: https://github.com/jackc/pgtype/blob/75446032b914bb0be5e07da29c976034c0a666cf/uuid.go#L3-L230
+
+type UUID struct {
```
roneli, line 17: Would embedding the original/calling the original code

roneli (changes requested):
LGTM overall, had a general suggestion that can reduce the amount of code we have.

pkg/client/uuid.go
bbernays: Great suggestion! Fixed!

disq (approved):

PR #342, opened 2021-12-14
Integrate Sentry

disq:
What more do we need?


spangenberg (commented):
LGTM

yevgenypats (dismissed):
Where is the opt-out flag?

disq:
> Where is the opt-out flag?

oops. added.

spangenberg (approved):

pkg/client/client.go
```diff
@@ -1050,3 +1052,15 @@ func collectFetchSummaryStats(span otrace.Span, fetchSummaries map[string]Provid
...
 }
+
+// reportFetchSummaryStats reads provided fetch summaries and sends them to sentry
+func reportFetchSummaryStats(fetchSummaries map[string]ProviderFetchSummary) {
+	for _, ps := range fetchSummaries {
+		for _, e := range ps.Diagnostics() {
+			if e.Severity() == diag.ERROR {
+				d := e.Description()
+				sentry.CaptureException(fmt.Errorf("%s: %s: %s: %s: %s", ps.ProviderName, e.Type().String(), d.Resource, d.Summary, d.Detail))
+			}
```
roneli, line 1063:
```suggestion
			if e.Severity() != diag.ERROR {
			  continue
			}
				d := e.Description()
				sentry.CaptureException(fmt.Errorf("%s: %s: %s: %s: %s", ps.ProviderName, e.Type().String(), d.Resource, d.Summary, d.Detail))
```

pkg/client/client.go
```diff
@@ -518,6 +519,7 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 	}
 
 	collectFetchSummaryStats(otrace.SpanFromContext(ctx), response.ProviderFetchSummary)
+	reportFetchSummaryStats(response.ProviderFetchSummary)
```
roneli, line 522:
```suggestion
	reportFetchSummaryErrors(response.ProviderFetchSummary)
```

roneli, line 522: I think you could probably combine both methods to reduce the looping, wdyt?

roneli (changes requested):
Would also suggest to do some manual tests like running AWS provider with bad credentials

Also write some tests for example test provider and see that errors get captured and sent.

For AWS provider I would test that the no creds are leaked, they shouldn't but just to double check.

internal/telemetry/error.go
```diff
@@ -11,6 +12,8 @@ func RecordError(span otrace.Span, err error, opts ...otrace.EventOption) {
 		return
 	}
 
+	sentry.CaptureException(err)
+
 	//  TODO for fetch get table name / error type
```
roneli, line 17: we do this no with resource name etc' so we can remove this TODO

roneli (approved):
LGTM, minor nit

PR #350, opened 2021-12-16
Classify errors

disq:
Maybe we need something like this?

Not sure about classes, maybe just `shouldIgnoreError(error) bool` and set a generic "error" state...


internal/telemetry/error.go
```diff
@@ -1,19 +1,70 @@
...
+func classifyError(err error) errClass {
+	if st, ok := status.FromError(err); ok {
+		if st.Code() == gcodes.Canceled {
+			return errCancellation
+		}
+		// if strings.Contains(st.Message(), `AWS Error: operation error STS: GetCallerIdentity,`) {
+		if strings.HasPrefix(st.Message(), `failed to create aws client for`) ||
+			strings.HasPrefix(st.Message(), `failed to retrieve credentials for`) {
+			return errAuth
+		}
```
roneli, line 52: perhaps makes this a file (embedded with errors) I have a feeling this might grow as currently its very specific

internal/telemetry/error.go
```diff
@@ -1,19 +1,70 @@
...
+
+const (
+	errNoClass      = errClass("")
+	errCancellation = errClass("cancelled")
+	errAuth         = errClass("auth")
+	errConn         = errClass("connection")
+	errDatabase     = errClass("database")
+)
+
+func classifyError(err error) errClass {
```
roneli, line 43: food for taught we can make this more generic i.e define a static data structure where we have for each error type possible messages that will classify / known errors etc' and then just loop over each type with the error returned to classify.

roneli (changes requested):
LGTM, had some ideas, tell me what you think before we approve and merge

internal/telemetry/error.go
disq: Every error type has something to check for before it's "accepted" (as seen with the `net.OpError.Op` and `pq.Error.Code.Class()` cases) so unless it's fairly generic ("ignore all pqErrors") it would just complicate things for the sake of being able to loop through a list.

internal/telemetry/error.go
disq: At this point it would be too much complication, we don't know if all errors will use `HasPrefix` (as I've only checked with the current reported "issues" in Sentry, which these types of provider-errors are mostly from AWS)

roneli (approved):

PR #352, opened 2021-12-17
hide source on config generation

amanenk:
fixes cloudquery/cloudquery-issues#176


pkg/client/client_test.go
```diff
@@ -609,6 +613,7 @@ func Test_collectProviderVersions(t *testing.T) {
 }
 
 func Test_CheckForProviderUpdates(t *testing.T) {
+	soruce := "cloudquery"
```
roneli, line 616: Typo

pkg/client/client_test.go
```diff
@@ -467,6 +469,8 @@ func setupTestPlugin(t *testing.T) context.CancelFunc {
 	defer watcher.Close()
 
 	go provider.ServeTestPlugin(debugCtx)
+	// sleep to prevent ParseReattachProviders from reading empty file
+	time.Sleep(time.Millisecond * 50)
```
roneli, line 473: This feels flaky to me, would use a file watcher I think such a check already exists in our tests

roneli (changes requested):
Looks good, two minor issues and we can merge

pkg/client/client_test.go
amanenk: it seems that error was related to github request limits. Now it works without these lines

roneli (approved):
LGTM

PR #354, opened 2021-12-20
telemetry: Instrument runPolicy

disq:
Too little or just right? Should I dive into PolicyManager and instrument LoadPolicy/RunPolicy as well?

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -786,7 +786,19 @@ func (c *Client) LoadPolicies(ctx context.Context, req *PoliciesRunRequest) (pol
 }
 
 func (c *Client) runPolicy(ctx context.Context, policyConfig *config.Policy, req *PoliciesRunRequest) (res *policy.ExecutionResult, retErr error) {
-	ctx, spanEnder := telemetry.StartSpanFromContext(ctx, "runPolicy", otrace.WithAttributes(attribute.String("policy", req.PolicyName)))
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -785,9 +785,11 @@ func (c *Client) LoadPolicies(ctx context.Context, req *PoliciesRunRequest) (pol
 	return loadedPolicies, nil
 }
 
-func (c *Client) runPolicy(ctx context.Context, policyConfig *config.Policy, req *PoliciesRunRequest) (*policy.ExecutionResult, error) {
-	c.Logger.Info("preparing to run policy")
+func (c *Client) runPolicy(ctx context.Context, policyConfig *config.Policy, req *PoliciesRunRequest) (res *policy.ExecutionResult, retErr error) {
+	ctx, spanEnder := telemetry.StartSpanFromContext(ctx, "runPolicy", otrace.WithAttributes(attribute.String("policy", req.PolicyName)))
```
roneli, line 789: Would add subpath, but also only span perhaps only on policies that are hub only.

roneli (changes requested):
Added some taughts, might require more discussion

pkg/client/client.go
disq: Something like this maybe?

roneli (approved):

PR #356, opened 2021-12-21
feat: Added store fetch summary routine

### LINKED ISSUE #284: Store Fetch Summary

It is important to understand what resource(s) were fetched and which resource(s) had issues or warnings. It is also important to understand what provider was used to grab the data. This data can be used during policy evaluation to ensure correctness

At the end of every fetch we should store a summary of all actions that were attempted and their result.

amanenk:
closes #284


pkg/client/fetch_summary/fetch_summary.go
```diff
@@ -0,0 +1,108 @@
...
+);
+`
+
+type FetchSummarizer struct {
+	pool     *pgxpool.Pool
+	dbStruct *sqlbuilder.Struct
+}
+
+// NewFetchSummarizer creates cq_fetches table and returns a summarizer that saves fetch summary to cq_fetches table
+func NewFetchSummarizer(ctx context.Context, pool *pgxpool.Pool) (*FetchSummarizer, error) {
```
disq, line 39: just `New` maybe?

pkg/client/fetch_summary/fetch_summary.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return &FetchSummarizer{
+		pool:     pool,
+		dbStruct: sqlbuilder.NewStruct(new(FetchSummary)),
+	}, nil
+}
+
+// FetchSummary includes a summarized report of fetch, such as fetch id, fetch start and finish,
+// resources fetch results
+type FetchSummary struct {
```
disq, line 59: `Summary`

pkg/client/fetch_summary/fetch_summary.go
```diff
@@ -0,0 +1,108 @@
...
+	Finish             time.Time              `db:"finish"`
+	TotalResourceCount uint64                 `db:"total_resource_count"`
+	ProviderName       string                 `db:"provider_name"`
+	ProviderVersion    string                 `db:"provider_version"`
+	ProviderMeta       []byte                 `db:"provider_meta"` // reserved field to store providers metadata such as
+	FetchedResources   []ResourceFetchSummary `db:"fetch_results"`
+}
+
+// ResourceFetchSummary includes a data about fetching specific resource
+type ResourceFetchSummary struct {
```
disq, line 73: `ResourceSummary`

pkg/client/fetch_summary/fetch_summary.go
```diff
@@ -0,0 +1,108 @@
...
+	Status string `json:"status"`
+	// Total Amount of resources collected by this resource
+	ResourceCount uint64 `json:"resource_count"`
+	// Diagnostics of failed resource fetch, the diagnostic provides insights such as severity, summary and
+	// details on how to solve this issue
+	Diagnostics diag.Diagnostics `json:"diagnostics"`
+}
+
+// SaveFetchSummary saves fetch summary into cq_fetches database
+func (c *FetchSummarizer) SaveFetchSummary(ctx context.Context, fs FetchSummary) error {
```
disq, line 92: `Save`

pkg/client/fetch_summary/fetch_summary.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/huandu/go-sqlbuilder"
+	"github.com/jackc/pgx/v4/pgxpool"
+
+	"github.com/cloudquery/cq-provider-sdk/cqproto"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema/diag"
+	"github.com/google/uuid"
```
disq, line 12: I would move this line to the upper block

roneli (changes requested):
Didn't post this review for some reason, sorry for the delay. Looks like a good start, have some structural changes to purpose  + we need to support migrate for core tables that are created, otherwise it will be hard to manage it.

go.mod
```diff
@@ -133,3 +133,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
+
+replace (
+	github.com/cloudquery/cq-provider-sdk => ../cq-provider-sdk
```
disq, line 138: go.mod testing mode leak

pkg/client/migrations/1_v0.17.4.down.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE cq_fetches;
```
disq, line 1:
```suggestion
DROP TABLE cq_fetches;
```

Also drop schema?

pkg/client/migrations/1_v0.17.4.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS cq_fetches
+(
+    cq_id                UUID NOT NULL,
+    cq_fetch_id          UUID NOT NULL,
```
disq, line 4: What's the difference between `cq_id` and `cq_fetch_id`? Also why the `cq_` prefix here?

go.mod
roneli: I think we can upgrade to latest SDK @amanenk

pkg/client/client.go
```diff
@@ -291,6 +296,12 @@ func New(ctx context.Context, options ...Option) (*Client, error) {
 			c.Logger.Warn("postgrest validation warning", "err", err)
 		}
 	}
+	// migrate cloudquery core tables to latest version
+	err = c.MigrateCore(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to migrate cloudquery_core tables: %w", err)
+	}
```
roneli, line 303:
```suggestion
	if err := c.MigrateCore(ctx); err != nil {
		return nil, fmt.Errorf("failed to migrate cloudquery core tables: %w", err)
	}
```

pkg/client/migrations/1_v0.17.4.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS cq_fetches
```
roneli, line 1: we can call this table fetches as users will do cloudquery.fetches to access this table `cq` will be redundant here i.e cloudquery.cq_fetches

pkg/client/migrations/1_v0.17.4.up.sql
roneli: I agree we should remove the cq prefix here, as this table resides in its own schema

pkg/client/client.go
```diff
@@ -922,6 +968,49 @@ func (c *Client) buildProviderMigrator(migrations map[string][]byte, providerNam
 	return m, providerConfig, err
 }
 
+func createCoreSchema(ctx context.Context, pool *pgxpool.Pool) error {
```
roneli, line 971: move this function down, usually we want to put private functions after public functions for the reader to first only see the exported interface.

roneli (changes requested):
Looking good, had a few extra comments

disq (approved):

yevgenypats (dismissed):
I would run the migration on `fetch` and not on `new Client` because otherwise we will have to have a database running even for commands that do not require a database such as `cloudquery policy describe` and potentially others.

roneli:
> I would run the migration on `fetch` and not on `new Client` because otherwise we will have to have a database running even for commands that do not require a database such as `cloudquery policy describe` and potentially others.

I would add there are some places that client can be initialized without dsn otherwise it tries to connect, I think also in policy describe it tries. So if dsn != "" its safe to run migrations

pkg/client/client.go
```diff
@@ -1097,3 +1174,46 @@ func reportFetchSummaryErrors(span otrace.Span, fetchSummaries map[string]Provid
...
+	defer conn.Release()
+
+	_, err = conn.Exec(ctx, "CREATE SCHEMA IF NOT EXISTS cloudquery")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func readMigrationFiles(log hclog.Logger, migrationFiles embed.FS) (map[string][]byte, error) {
```
roneli, line 1192: why not use SDK's function?

pkg/client/client.go
```diff
@@ -944,6 +992,35 @@ func (c *Client) buildProviderMigrator(migrations map[string][]byte, providerNam
...
+
+	defer func() {
+		if err := m.Close(); err != nil {
+			c.Logger.Error("failed to close migrator connection", "error", err)
+		}
+	}()
+
+	version := "latest"
+	err = m.UpgradeProvider(version)
+	if err != migrate.ErrNoChange {
```
roneli, line 1018:
```suggestion
        const latestVersion := "latest" // todo move this out of the function
	if err := m.UpgradeProvider(latestVersion); err != migrate.ErrNoChange {
```

pkg/client/client.go
```diff
@@ -944,6 +992,35 @@ func (c *Client) buildProviderMigrator(migrations map[string][]byte, providerNam
...
+		return err
```
roneli, line 1019:
```suggestion
		return fmt.Errorf("failed to migrate cloudquery core schema: %w",  err)
```

roneli (changes requested):
More comments I missed in prior review

roneli (changes requested):
Another comment I missed

pkg/client/client.go
amanenk: it was not released

amanenk:
> > I would run the migration on `fetch` and not on `new Client` because otherwise we will have to have a database running even for commands that do not require a database such as `cloudquery policy describe` and potentially others.
> 
> I would add there are some places that client can be initialized without dsn otherwise it tries to connect, I think also in policy describe it tries. So if dsn != "" its safe to run migrations

If we will have some other core tables related to other commands it is better to migrate in the client initialization

pkg/client/client.go
```diff
@@ -512,6 +551,16 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
...
+
+				fs.addResourceFetchSummary(ResourceFetchSummary{
+					ResourceName:                resp.ResourceName,
+					FinishedResources:           resp.FinishedResources,
+					Status:                      strconv.Itoa(int(resp.Summary.Status)), // todo use human readable representation of status
+					Error:                       resp.Error,
+					PartialFetchFailedResources: resp.PartialFetchFailedResources,
+					ResourceCount:               resp.ResourceCount,
+					Diagnostics:                 resp.Summary.Diagnostics,
+				})
```
roneli, line 563:
```suggestion
				fs.Resources = append(fs.Resources, ResourceFetchSummary{
					ResourceName:                resp.ResourceName,
					FinishedResources:           resp.FinishedResources,
					Status:                      strconv.Itoa(int(resp.Summary.Status)), // todo use human readable representation of status
					Error:                       resp.Error,
					PartialFetchFailedResources: resp.PartialFetchFailedResources,
					ResourceCount:               resp.ResourceCount,
					Diagnostics:                 resp.Summary.Diagnostics,
				})
```

pkg/client/fetch.go
```diff
@@ -0,0 +1,85 @@
...
+	//  Unique Id of fetch session
+	FetchId            uuid.UUID               `db:"fetch_id"`
+	Start              time.Time               `db:"start"`
+	Finish             time.Time               `db:"finish"`
+	IsSuccess          bool                    `db:"is_success"`
+	TotalResourceCount uint64                  `db:"total_resource_count"`
+	TotalErrorsCount   uint64                  `db:"total_errors_count"`
+	ProviderName       string                  `db:"provider_name"`
+	ProviderVersion    string                  `db:"provider_version"`
+	FetchedResources   *[]ResourceFetchSummary `db:"results"`
```
roneli, line 29:
```suggestion
	Resources   *[]ResourceFetchSummary `db:"results"`
```

roneli (approved):
LGTM, two minor nits

PR #361, opened 2021-12-23
telemetry: Detect lambda env

disq:


roneli (approved):

internal/telemetry/telemetry.go
```diff
@@ -311,6 +312,21 @@ func IsCI() bool {
 	return false
 }
 
+// IsLambda determines if we're running under a Lambda env by checking Lambda-specific env vars
+func IsLambda() bool {
```
bbernays, line 316: If we are checking for any ephemeral/FaaS then maybe we should name this something different?

PR #363, opened 2021-12-23
Always disable data delete, remove the option

disq:
Do we also want an SDK counterpart removing this from the proto?


roneli:
To keep backwards compatability, I would pass it as true here always, wouldn't touch the protocol

roneli (approved):

PR #367, opened 2021-12-24
Fix is_faas misdetection

### LINKED ISSUE #366: Detect FaaS
### Describe the Bug
The env variable `AWS_EXECUTION_ENV` is actually set when using AWS ECS Fargate and EC2 as well as Lambda

### Expected Behavior
Only find FaaS environments

### References

- [Source-1](https://stackoverflow.com/questions/54177061/is-it-possible-to-detect-fargate-without-trying-the-metadata-api)
- [Source-2](https://github.com/jeshan/fargate-environment-variables)

disq:
Fixes #366


bbernays (approved):
LGTM

PR #374, opened 2021-12-30
Better error for HCL bounds issues

disq:
We don't really know what caused it other than a stack trace, so I just enriched the error, preventing panic.


disq:
If #375 is merged this can be closed/

PR #377, opened 2021-12-31
Classify more errors

disq:
- Classify context.Canceled, SSL setup and query cancellations 
- Persist the random ID to better identify installations

### DIFF
--- a/cmd/sentry.go
+++ b/cmd/sentry.go
@@ -69,19 +69,16 @@ func initSentry() {
 	}
 }
 
-func setSentryVars(traceID string) {
-	hub := sentry.CurrentHub()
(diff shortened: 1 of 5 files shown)


yevgenypats:
Looks good. Question: If we are in CI, can we use the hash of the `config.hcl` + the type of CI ?

disq:
> Looks good. Question: If we are in CI, can we use the hash of the `config.hcl` + the type of CI ?

CI type detection may or may not work due to how the thing works (some CI's may not have their own identifier but set a generic env var)

yevgenypats (approved):

roneli (approved):

PR #375, opened 2021-12-30
remove the need for json converter

yevgenypats:

PR #379, opened 2022-01-02
upgrade to protocol v3 due to #375 PR

yevgenypats:
https://github.com/cloudquery/cloudquery/pull/375

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/cloudquery/cloudquery/internal/logging"
 	"github.com/cloudquery/cloudquery/pkg/plugin/registry"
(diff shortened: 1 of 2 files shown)


pkg/plugin/plugin.go
```diff
@@ -45,7 +46,7 @@ func newRemotePlugin(details *registry.ProviderDetails, alias string, env []stri
 	client := plugin.NewClient(&plugin.ClientConfig{
 		HandshakeConfig: serve.Handshake,
 		VersionedPlugins: map[int]plugin.PluginSet{
-			2: pluginMap,
+			3: pluginMap,
```
roneli, line 49:
```suggestion
			3: pluginMap,
			2: pluginMap,
```

You need to support also previous protocol, I think the above should do the trick

roneli (changes requested):

pkg/plugin/plugin.go
yevgenypats: This wont work here as we will need to get the json back and we removed this code. This is why we release first the providers and then cloudquery core. so users will be able to upgrade to latest provider

pkg/plugin/plugin.go
```diff
@@ -55,6 +56,10 @@ func newRemotePlugin(details *registry.ProviderDetails, alias string, env []stri
 	rpcClient, err := client.Client()
 	if err != nil {
 		client.Kill()
+		// give a more clear message to the user
+		if strings.Contains(err.Error(), "Incompatible API version") {
```
disq, line 60: Are you sure the error comes in `client.Client()` and not few lines below in `Dispense()`?

yevgenypats: yeah. tested it.

roneli (approved):

PR #381, opened 2022-01-03
Upgrade history: drop view first

disq:
See the comment for explanation and link


roneli (approved):
Looks good have one comment would suggest maybe to run them in a tx or combine them into a single query i.e DROP ... ; CREATE OR REPLACE; actually create will be enough

PR #382, opened 2022-01-03
Don't check for update if development build

disq:
This prevents the warning-level log


yevgenypats (approved):
A bit unrelated to this PR but the `DefaultVersion` confused me a bit. can we name it `DevelopmentVersion` ?

disq:
> A bit unrelated to this PR but the `DefaultVersion` confused me a bit. can we name it `DevelopmentVersion` ?

Good idea, done

PR #385, opened 2022-01-03
Fix isLevel functions for log level comparison

### LINKED ISSUE #368: Bug: isLevel is incorrect
I believe this comparision in all IsLevel function should be the other way around
https://github.com/cloudquery/cloudquery/blob/098070efba6974fe0b6dd6d24838dcddf70584a3/internal/logging/hclog.go#L68

spangenberg:
Fixes #368


yevgenypats (approved):

yevgenypats:
@spangenberg did you give it a manual test to make sure it's working and we didn't mix up something. i.e that we continue to see the logs as necessary when running with `--enable-console-log`

spangenberg:
> @spangenberg did you give it a manual test to make sure it's working and we didn't mix up something. i.e that we continue to see the logs as necessary when running with `--enable-console-log`

Works as expected, normal mode gives info, warn, err...
Verbose mode additionally gives debug and trace

PR #386, opened 2022-01-04
Prevent reporting of errors to Sentry twice

disq:
Once in end of span (with `spanEnder` business) and the other in the cli handler.


roneli (approved):
Looks good, shame no way to actually test this

PR #388, opened 2022-01-05
Improve provider update check error message

irmatov:
Rework unit tests to not require network access

### DIFF
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/cloudquery/cloudquery/pkg/client"
 	"github.com/cloudquery/cloudquery/pkg/config"
 	"github.com/cloudquery/cloudquery/pkg/module"
+	"github.com/cloudquery/cloudquery/pkg/plugin/registry"
 	"github.com/cloudquery/cloudquery/pkg/policy"
 	"github.com/cloudquery/cloudquery/pkg/ui"
 )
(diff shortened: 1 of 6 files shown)


pkg/ui/console/client.go
```diff
@@ -92,7 +93,7 @@ func (c Client) DownloadProviders(ctx context.Context) error {
 		c.updater.Wait()
 	}
 	ui.ColorizedOutput(ui.ColorProgress, "Finished provider initialization...\n\n")
-	updates, _ := c.c.CheckForProviderUpdates(ctx)
+	updates := client.CheckForProviderUpdates(ctx, registry.GetLatestRelease, c.c.Providers, c.c.Logger)
```
yevgenypats, line 96: why migrate it from method level and then pass everything as parameters?

yevgenypats (changes requested):
Still when running with latest and --enable-console-log I see a pretty confusing message:

```
2:28PM INF Finished provider initialization...
2:28PM WRN Update check failed error="bad local version: provider aws, version latest" provider=aws
```

irmatov:
Now it's clear what's the problem..

pkg/ui/console/client.go
irmatov: It makes it more of a pure function, easier to test and easier to reason about. Doesn't matter much in this case, I reverted it.

irmatov:
Should be fixed now.

yevgenypats:
@irmatov wdyt about this approach - https://github.com/cloudquery/cloudquery/pull/388 ?

irmatov:
@yevgenypats it is mostly the same. Here I rework unit tests a bit to not depend on external repo (and network access, and even the DB availability) just to check how our update detection works.

spangenberg (approved):
LGTM

yevgenypats:
Fixed by https://github.com/cloudquery/cloudquery/pull/391 . Per the tests I think if the tests are simple and short it's better to focus on integration e2e tests.

PR #391, opened 2022-01-05
remove confusing message on upgrade when version is latest

yevgenypats:
If version was latest the log would always print `fmt.Errorf("bad version: provider %s, version %s", providerName, requestedProvider.Version)` . 

Also there were just incorrect code comparing `if err != nil` 3 times one after another 😄 

There is still confusing logs on each run but I'll address them in a different PR.

```
7:45PM INF upgrading provider version provider=aws version=latest
7:45PM INF ✓ Upgraded provider aws to latest successfully.
7:45PM INF Finished upgrading providers...
7:45PM INF plugin doesn't exist, creating.. alias= provider=aws
7:45PM INF Starting provider fetch...
7:45PM INF plugin doesn't exist, creating.. alias= provider=aws
7:45PM INF received fetch request extra_fields=null history_enabled=false
7:45PM INF plugin doesn't exist, creating.. alias=aws provider=aws
7:45PM INF requesting provider to configure alias=aws provider=aws version=v0.8.5
```


irmatov:
https://github.com/cloudquery/cloudquery/pull/388 ?

yevgenypats:
Suggestion to https://github.com/cloudquery/cloudquery/pull/388

pkg/client/client.go
```diff
@@ -320,13 +320,18 @@ type ProviderUpdateSummary struct {
...
+		}
 		version, err := c.Hub.CheckProviderUpdate(ctx, p)
 		if err != nil {
-			c.Logger.Warn("Failed check provider update", "provider", p.Name)
+			c.Logger.Warn("Failed check provider update", "provider", p.Name, "error", err)
 			continue
 		}
-
 		if version == nil {
+			c.Logger.Debug("version is latest already", "provider", p.Name, "version", p.Version)
```
disq, line 334:
```suggestion
			c.Logger.Debug("already at latest version", "provider", p.Name, "version", p.Version)
```

Because otherwise it would conflict with the new message at line 325, and not really clear either

pkg/client/client.go
```diff
@@ -320,13 +320,18 @@ type ProviderUpdateSummary struct {
 func (c *Client) CheckForProviderUpdates(ctx context.Context) ([]ProviderUpdateSummary, error) {
 	var summary []ProviderUpdateSummary
 	for _, p := range c.Providers {
+		// if version is latest it means there is no update as DownloadProvider will download the latest version automatically
+		if strings.Compare(p.Version, "latest") == 0 {
+			c.Logger.Debug("version is latest already", "provider", p.Name, "version", p.Version)
```
disq, line 325:
```suggestion
			c.Logger.Debug("version is latest", "provider", p.Name, "version", p.Version)
```
I think `already` is not necessary in this case

disq (approved):
Minor nits

pkg/client/client.go
```diff
@@ -320,13 +320,18 @@ type ProviderUpdateSummary struct {
 func (c *Client) CheckForProviderUpdates(ctx context.Context) ([]ProviderUpdateSummary, error) {
 	var summary []ProviderUpdateSummary
 	for _, p := range c.Providers {
+		// if version is latest it means there is no update as DownloadProvider will download the latest version automatically
+		if strings.Compare(p.Version, "latest") == 0 {
+			c.Logger.Debug("version is latest", "provider", p.Name, "version", p.Version)
```
roneli, line 325: maybe Info level?

roneli (commented):
LGTM, minor nit

roneli (approved):

pkg/client/client.go
irmatov: That is not useful to see at the info level. I would even drop the debug given this all is executed after DownloadProviders, which should complain if it fails to update the provider to the latest version

pkg/client/client.go
```diff
@@ -314,13 +314,18 @@ type ProviderUpdateSummary struct {
...
+		}
 		version, err := c.Hub.CheckProviderUpdate(ctx, p)
 		if err != nil {
-			c.Logger.Warn("Failed check provider update", "provider", p.Name)
+			c.Logger.Warn("Failed check provider update", "provider", p.Name, "error", err)
 			continue
 		}
-
 		if version == nil {
+			c.Logger.Debug("already at latest version", "provider", p.Name, "version", p.Version)
```
roneli, line 328: would also make this info
```suggestion
			c.Logger.Info("already at latest version", "provider", p.Name, "version", p.Version)
```

spangenberg (approved):

PR #397, opened 2022-01-07
fix lambda json conversion

yevgenypats:
Description is still wip:

As we removed the jsonconverter we didn't update the lambda correctly yet.

Related to https://github.com/cloudquery/terraform-aws-cloudquery/pull/9

Also relates to this - https://github.com/hashicorp/hcl/issues/294

Supporting both json and HCL is possible but it's a real hassle and room for edge-cases so I suggest dropping this for not completly sticking to HCL unless in the future this will be requested somehow again.

### DIFF
--- a/deploy/lambda.go
+++ b/deploy/lambda.go
@@ -2,7 +2,6 @@ package deploy
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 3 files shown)


roneli (approved):
Would add error if file type is not of correct extension to user. and not we don't support json hcl.

PR #399, opened 2022-01-09
fix empty policy edge case

roneli:


pkg/policy/parser.go
```diff
@@ -69,7 +71,12 @@ func DecodePolicy(body hcl.Body, diags hcl.Diagnostics, basePath string) (*Polic
 			panic("unexpected block")
 		}
 	}
-	return nil, diags
```
yevgenypats, line 72: Isn't the `for _, block := range content.Blocks` a bit strange as you can't really have multiple `policies`?

yevgenypats (approved):
looks good, one comment on the logic (not in the PR but in `DecodePolicy`)

PR #412, opened 2022-01-14
fix: Exit with 1 on policy error

yevgenypats:
Remove confusing and unused flags:
stop-on-failure
fail-on-violations

Reasoning:

If there is an internal error such as incorrect version, missing tables, there is no reason really to not stop, so this just creates additional documentation and not-intuitive default behaviour (such as not returning 1 on internal error)

 `stop-on-failure` - policies run very fast so I would eliminate that flag as well, just have a default behaviour that shows all the results.

Why I started it:

In the CI we always had to add special flags to detect errors and sometimes we missed errors because the CI didn't fail on internal errors.


roneli (approved):
LGTM

PR #413, opened 2022-01-14
feat: Expose max_parallel_resource_fetch_limit

roneli:
expose max_parallel_resource_fetch_limit to allow users to limit amount of concurrent resources fetched by provider, if 0 is provider or attribute is not defined it acts as if no limit is set


roneli:
might fix https://github.com/cloudquery/cq-provider-gcp/issues/97

yevgenypats (approved):

PR #416, opened 2022-01-17
feat: Increase ulimit in unix environment

spangenberg:
The ulimit in Unix environment will be increased automatically.
Currently, we fail a lot of runs due to low limits.
In case of an error, we just log and continue.

### DIFF
--- a/cmd/ulimit_unix.go
+++ b/cmd/ulimit_unix.go
@@ -1,0 +1,45 @@
+//go:build darwin || linux
+// +build darwin linux
+
+package cmd
+
+import (
+	"fmt"
+	"syscall"
+
+	zerolog "github.com/rs/zerolog/log"
+)
+
+const ulimitUnix uint64 = 16384
+
+func init() {
+	fileDescriptorF = checkAndSetUlimitUnix
+}
+
+func checkAndSetUlimitUnix() {
+	logger := zerolog.Logger
+	if err := setUlimit(ulimitUnix); err != nil {
+		logger.Err(fmt.Errorf("error setting ulimit: %w", err))
+	}
+}
+
+func setUlimit(ulimit uint64) error {
+	logger := zerolog.Logger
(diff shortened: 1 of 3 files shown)


cmd/ulimit_unix.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"syscall"
+
+	zerolog "github.com/rs/zerolog/log"
+)
+
+const ulimitUnix uint64 = 16384
+
+func init() {
+	fileDescriptorF = checkAndSetUlimitUnix
```
yevgenypats, line 16: why we can call `checkAndSetUlimitUnix` directly ? This should also remove the need for this [check](https://github.com/cloudquery/cloudquery/pull/416/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR145) unless Im missing some edge-case

yevgenypats (changes requested):
Great. I think this should work - I've one questions regarding initialization and if you were able to test it that it also sets the ulimit for child process (I think it should but would double verify it manually).

yevgenypats (approved):
ok I see now regarding the initialization, it is because of the build tags. Can you add a small comment for that.

yevgenypats:
@spangenberg can you only fix that the commit will be the same as the pr title (it's when there is one commit, GitHub is using the commit instead of the PR Title and it doesn't have the `feat` prefix so it's "unconventional")

cmd/ulimit_unix.go
spangenberg: It should fail on windows if I'm not completely mistaken. Since `syscall` is not available.

spangenberg:
Sure, will take care about that and the extended manual testing.

spangenberg:
The PR was manually tested on MacOS, Linux and Windows.

PR #422, opened 2022-01-19
fix: Updated the policy download command message

zagronitay:


cmd/policy_download.go
```diff
@@ -17,17 +17,11 @@ var (
...
   
   # The following will be the same as above
   # Official policies are hosted here: https://github.com/cloudquery-policies
-  cloudquery policy download cloudquery-policies/aws-cis-1.2.0
-
-  # Download community policy
-  cloudquery policy download COMMUNITY_GITHUB_ORG/aws-cis-1.2.0
-
-  # Download community from any source control
-  cloudquery policy download https://github.com/COMMUNITY_GITHUB_ORG/aws-cis-1.2.0
```
bbernays, line 30: Are these not valid examples anymore? Why do we not want to provide these examples?

roneli: I agree we should add them, would double check them just in case

roneli (approved):
LGTM

PR #426, opened 2022-01-20
fix: Store fetch summary adjustements

### LINKED ISSUE #423: Fetch Summary violation on multiple providers
### Describe the Bug
If running two providers (two aws provider blocks) you receive 
failed to save fetch summary err="ERROR: duplicate key value violates unique constraint \"fetches_pk\" (SQLSTATE 23505)"

### Expected Behavior
Summary saved for each block

### CloudQuery Version
v0.19.1

### Steps to Reproduce
```
cloudquery fetch
```

Add the following providers to config.hcl

```
provider "aws" {
  configuration {
    aws_debug = false
  }
  // list of resources to fetch
  resources = [
    "lambda.runtimes"
  ]
  // enables partial fetching, allowing for any failures to not stop full resource pull
  enable_partial_fetch = true
}


provider "aws" {
  alias = "aws-2"
  configuration {
    aws_debug = false
  }
  // list of resources to fetch
  resources = [
    "lambda.runtimes"
  ]
  // enables partial fetching, allowing for any failures to not stop full resource pull
  enable_partial_fetch = true
}
```

amanenk:
- Fetch summary constraint now uses alias to allow using multiple definitions of the same provider in config
- Console fetch result reporting now use alias to report fetch results correctly
- added createdAt fetch summary column to store time of creation of fetch summary object
- add core_version column 
- Fetch summary start, finish, created_at columns are now nullable  
            - if start is null that means that fetch hasn't started (err during initialization, etc)
            - if finish is null that means that fetch hasn't finished 
- added store of fetch summary when createProvider function failed
 
closes #423


pkg/client/client.go
```diff
@@ -423,28 +423,30 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 
 	for _, providerConfig := range request.Providers {
 		providerConfig := providerConfig
+		createdAt := time.Now().UTC()
+		fs := FetchSummary{
+			FetchId:       fetchId,
```
disq, line 428: Maybe also save CloudQuery core version?

pkg/client/client.go
```diff
@@ -507,15 +510,16 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 							pLog.Warn("received partial fetch error", parsePartialFetchKV(fetchError)...)
 						}
 						fetchSummaries <- ProviderFetchSummary{
-							ProviderName:          providerConfig.Name,
+							ProviderName:          providerConfig.Alias,
```
disq, line 513: I think this is for diagnostics and not related to SaveFetchSummary?

disq (changes requested):

pkg/client/client.go
amanenk: that is a good suggestion

pkg/client/client.go
amanenk: current configuration allows to create more than one instance of a provider only in case if instances have different aliases. Also fetch summaries look wrong if we use only provider name since only results of latest fetch summary are displayed in console.

amanenk: image
here how output looks now when we use 2 aliases
before it was only 1 line which is wrong

disq (approved):
LGTM

pkg/client/client.go
```diff
@@ -410,6 +411,19 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 
 	for _, providerConfig := range request.Providers {
 		providerConfig := providerConfig
+		createdAt := time.Now().UTC()
+		fs := FetchSummary{
```
roneli, line 415: nit: fs collides with fs package

roneli (approved):
LGTM, another minor nit also L:489 `status`  collides with package name fix that.

Issue #428, opened 2022-01-20
Migrate Executor to New Filter


bbernays:
We should migrate the executor functionality to use the `policy.filter("selector")`. This will make the executor simpler as all checks/policies passed to the executor will be executed and there will be no filtering needed

roneli:
Fixed in #429

PR #429, opened 2022-01-20
feat: Remove filtering from executor

bbernays:
Implements #428 

A few comments:
1. selector path needs to be of the form `rootPolicy/policy1/subpolicy/check`
2. ~Error message has changed, need to understand if we need to put the old error message back in~
3. **Change**: If a policy has no Checks or sub-policies then no views will be invoked


pkg/policy/execute.go
```diff
@@ -119,65 +120,54 @@ func (e *Executor) with(policy string, args ...interface{}) *Executor {
...
 		}
 	}
-	if !found && len(selector) > 0 {
-		e.log.Error("policy/query not found with provided sub-policy selector", "selector", selector, "available_policies", policy.Policies.All())
-		return nil, fmt.Errorf("%s//%s: %w", policy.Name, path.Join(selector...), ErrPolicyOrQueryNotFound)
-	}
+	// if !found && len(selector) > 0 {
+	// 	e.log.Error("policy/query not found with provided sub-policy selector", "selector", selector, "available_policies", policy.Policies.All())
+	// 	return nil, fmt.Errorf("%s//%s: %w", policy.Name, path.Join(selector...), ErrPolicyOrQueryNotFound)
+	// }
```
bbernays, line 170: Still need to add this error hand back in

roneli (changes requested):
Had some suggestions, also in retrospect, feels like hack that the filter returns a Policy with one policy from the selector, not sure if its better to add a policy.FilterCheck. I would maybe suggest Filter only returns the last policy level i.e policy/subpolicy/subsubpolicy/check so it will return subsubpolicy and then in the execute u will do subsubPolicy.GetCheck(lastPart) ? maybe less hacky but must pass check to execute.

pkg/policy/manager.go
```diff
@@ -101,14 +102,15 @@ func (m *ManagerImpl) Run(ctx context.Context, request *ExecuteRequest) (*Execut
 			})
 		}
 	}
-
-	var selector []string
-	if request.Policy.meta.subPolicy != "" {
-		selector = strings.Split(request.Policy.meta.subPolicy, "/")
+	selector := strings.ReplaceAll(request.Policy.meta.subPolicy, "//", "/")
```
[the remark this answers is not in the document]
bbernays: I couldn't 100% verify that subPolicy() never returns something in the form of `aws//policy/subpolicy/subpolicy/control` so I cleaned the input so it removes the `//`

bbernays:
> Had some suggestions, also in retrospect, feels like hack that the filter returns a Policy with one policy from the selector, not sure if its better to add a policy.FilterCheck. I would maybe suggest Filter only returns the last policy level i.e policy/subpolicy/subsubpolicy/check so it will return subsubpolicy and then in the execute u will do subsubPolicy.GetCheck(lastPart) ? maybe less hacky but must pass check to execute.

With that way of doing it, how do you tell if the last part is a check or a policy?

pkg/policy/execute_test.go
```diff
@@ -313,8 +319,8 @@ func TestExecutor_Execute(t *testing.T) {
 				UpdateCallback: nil,
 				StopOnFailure:  tc.StopOnFailure,
 			}
-
-			res, err := executor.Execute(context.Background(), execReq, tc.Policy, tc.Selector)
+			filtered := tc.Policy.Filter(tc.Selector)
```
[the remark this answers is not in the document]
bbernays: I was trying to minimize the changes to ensure we don't break anything

roneli:
> > Had some suggestions, also in retrospect, feels like hack that the filter returns a Policy with one policy from the selector, not sure if its better to add a policy.FilterCheck. I would maybe suggest Filter only returns the last policy level i.e policy/subpolicy/subsubpolicy/check so it will return subsubpolicy and then in the execute u will do subsubPolicy.GetCheck(lastPart) ? maybe less hacky but must pass check to execute.
> 
> With that way of doing it, how do you tell if the last part is a check or a policy?

You give me a path lets say Px is a policy and C is a check. So by passsing p//p1/p2/p3/C or p//p1/p2/p3 the function always returns p3 + C in the latter case it will return p3 + "". The Execute function gets P3 + C if C != "" it runs C in P3 if it exists, otherwise runs all P3 checks.

bbernays:
> You give me a path lets say Px is a policy and C is a check. So by passsing p//p1/p2/p3/C or p//p1/p2/p3 the function always returns p3 + C in the latter case it will return p3 + "". The Execute function gets P3 + C if C != "" it runs C in P3 if it exists, otherwise runs all P3 checks.

While I see that your suggestion works in our current codebase but I am wary about tying policy filtering execution together. I implemented it this way so that any function that needs to interact with a policy never needs any more information than the policy that is passed to it. This fully decouples filtering from execution/evaluation/anything other action we implement. I don't think that actions should have to know any information about the selector or any other processing that might go on.

roneli:
> > You give me a path lets say Px is a policy and C is a check. So by passsing p//p1/p2/p3/C or p//p1/p2/p3 the function always returns p3 + C in the latter case it will return p3 + "". The Execute function gets P3 + C if C != "" it runs C in P3 if it exists, otherwise runs all P3 checks.
> 
> While I see that your suggestion works in our current codebase but I am wary about tying policy filtering execution together. I implemented it this way so that any function that needs to interact with a policy never needs any more information than the policy that is passed to it. This fully decouples filtering from execution/evaluation/anything other action we implement. I don't think that actions should have to know any information about the selector or any other processing that might go on.

Yes that makes sense, maybe what bugs me is the fact that if you ask for a control you get a copy of that policy only with that control. But I can see that the my initial suggestion remove the decoupling.

pkg/policy/manager.go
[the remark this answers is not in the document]
roneli: subPolicy should always never have the // it splits when it is loaded, as its also an internal member it shouldn't be changed other then where it is set which is in loadSource.

roneli (commented):
I think this looks good, lets wrap up our conversation and we can merge.

pkg/policy/manager.go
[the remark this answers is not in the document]
bbernays: Thank you for that assurance, I have removed the `ReplaceAll()` call

pkg/policy/manager.go
```diff
@@ -102,13 +102,14 @@ func (m *ManagerImpl) Run(ctx context.Context, request *ExecuteRequest) (*Execut
 		}
 	}
 
-	var selector []string
-	if request.Policy.meta.subPolicy != "" {
-		selector = strings.Split(request.Policy.meta.subPolicy, "/")
+	filteredPolicy := request.Policy.Filter(request.Policy.meta.subPolicy)
+	if len(filteredPolicy.Policies.All()) == 0 && len(filteredPolicy.Checks) == 0 {
```
roneli, line 106: nit: maybe make this a method? HasChecks()? which checks if policy has checks or any of its sub policies have checks

pkg/policy/execute.go
```diff
@@ -119,65 +118,56 @@ func (e *Executor) with(policy string, args ...interface{}) *Executor {
...
```
roneli, line 170: if this is commented out either remove it or uncomment it, or add a todo explaining why its commented out

pkg/policy/manager.go
roneli, line 106: isn't this checked in the .Execute function?

roneli (changes requested):
Looks good, a few questions nits.

pkg/policy/manager.go
bbernays: The check needs to be in both places in order to maintain backwards compatibility... We need to sanitize any policies being passed into the function and in order to maintain the same error message we have to return the error from within the `manager Run()` function otherwise we would have to pass the selector to the function which we are trying to avoid.

pkg/policy/execute.go
bbernays: removed

pkg/policy/manager.go
bbernays: I like this idea

bbernays: Implemented with tests and replaced all of the manual checks

roneli (approved):
LGTM

pkg/policy/manager.go
```diff
@@ -73,7 +73,12 @@ func (m *ManagerImpl) Run(ctx context.Context, request *ExecuteRequest) (*Execut
 		totalQueriesToRun = request.Policy.TotalQueries()
 		finishedQueries   = 0
 	)
-
+	filteredPolicy := request.Policy.Filter(request.Policy.meta.subPolicy)
+	if !filteredPolicy.HasChecks() {
+		m.logger.Error("policy/query not found with provided sub-policy selector", "selector", request.Policy.meta.subPolicy, "available_policies", filteredPolicy.Policies.All())
+		return nil, fmt.Errorf("%s//%s: %w", request.Policy.Name, request.Policy.meta.subPolicy, ErrPolicyOrQueryNotFound)
+	}
+	totalQueriesToRun = filteredPolicy.TotalQueries()
```
bbernays, line 81: This is rearranged so that the total Queries to run is calculated after filtering and not before

roneli (approved):
Looks good!

PR #432, opened 2022-01-24
feat: Policy Testing- Create Snapshot

bbernays:
PR Overview:
- adds new command: `cloudquery snapshot <selector> <output-directory>`
example: `cloudquery snapshot aws//foundational_security/ec2/EC2.18 database-data`

Output:
- stores 2 files in the following directory structure
    `<output-directory>/query-<CONTROL-ID>/tests/<UUID>`
        - `pg-dump.sql` stores all relevant tables and views including data
        - `data.json` this stores the output of control against the dumped data

Known issues:
- Panic occurs if DSN  is a blank string
- `pg_dump`- exports views as tables... This adds complications for migrations in the future as any tooling we might have will operate at the table level rather than the newly derived table from the view
- Logging statements are missing or wrong
~- When SQL queries returns no data output is `\N` (postgres syntax for `null`)~

To Do:
- [x] Parse DSN to replace hard coded credentials
- [x] Respect 2nd arg that defines snapshot directory for storage


yevgenypats:
Trying to run and getting the following error:
```
ep@macbook-pro-73 cloudquery % go run main.go policy snapshot aws//foundational_security/ec2/EC2.18 dest 
selector must specify only a single control
exit status 1
```

PR #434, opened 2022-01-24
feat: Update drift for AWS IOT resources

disq:
`billing_groups`, `ca_certificates` and `streams` aren't supported (as we don't have data for them in the tfstate I received)

Looks like those resources aren't supported in TF yet


disq:
Fixing stuff about iam.roles

roneli (approved):

PR #437, opened 2022-01-25
feat: Report on multiple aliased providers

disq:


pkg/ui/console/client.go
```diff
@@ -338,13 +338,19 @@ func (c Client) UpgradeProviders(ctx context.Context, args []string) error {
 		return err
 	}
 	ui.ColorizedOutput(ui.ColorProgress, "Upgrading CloudQuery providers %s\n\n", args)
+	dupes := make(map[string]struct{}, len(providers))
```
roneli, line 341: dupes shouldn't be allowed, as it won't parse in the config.

pkg/client/client.go
```diff
@@ -383,6 +384,31 @@ func (c *Client) Fetch(ctx context.Context, request FetchRequest) (res *FetchRes
 	ctx, spanEnder := telemetry.StartSpanFromContext(ctx, "Fetch")
 	defer spanEnder(retErr)
 
+	{
```
roneli, line 387: would move this to a function, but also just to request.Providers a function that returns providers/count

roneli (changes requested):
Some requests, fetch function is getting to big, lets not make it any bigger

pkg/ui/console/client.go
disq: It works with aliased things... and upgrades aws 2 times

roneli: I see what this fix does now, actually would make a function on the provider list to return a unique set and then iterate over that, or actually for upgrade it should iterate on the top level providers and not the provider blocks.

pkg/client/client.go
disq: Fixed

pkg/ui/console/client.go
disq: for 0.21 I assume.

roneli: Open an issue for this?

roneli (approved):
LGTM, lets open an issue around the duplicate upgrade

PR #438, opened 2022-01-25
fix: Panic On Policy Describe

bbernays:


disq:
typo on title: Descibe

pkg/ui/console/client.go
```diff
@@ -538,11 +538,18 @@ func (c Client) describePolicy(ctx context.Context, p *policy.Policy, selector s
...
 	}
+
 	ui.ColorizedOutput(ui.ColorHeader, "Describe Policy %s output:\n\n", p.String())
 	t := &Table{writer: tablewriter.NewWriter(os.Stdout)}
 	t.SetHeaders("Path", "Description")
-	pol := p.Filter(strings.ReplaceAll(selector, "//", "/"))
-	buildDescribePolicyTable(t, policy.Policies{&pol}, selector[:strings.LastIndexAny(selector, "/")])
+	selector = strings.ReplaceAll(selector, "//", "/")
+	pol := p.Filter(selector)
+	outputSelector := selector
```
disq, line 546: We don't need the `outputSelector`, just overwrite `selector` below?

pkg/ui/console/client.go
```diff
@@ -538,11 +538,18 @@ func (c Client) describePolicy(ctx context.Context, p *policy.Policy, selector s
...
+	if strings.Count(selector, "/") > 0 {
```
disq, line 547: `strings.Contains` as we don't need the actual count. (could even overoptimize by calling LastIndexAny and putting it in a variable first, and then checking if it's `> -1`... but would hurt readability)

disq (changes requested):

pkg/ui/console/client.go
bbernays: That works too

pkg/ui/console/client.go
bbernays: changed

PR #442, opened 2022-01-25
fix: Nil connection

disq:
If connection info is not specified we allow the client to get created (missing DSN just logs a line)
But this causes problems the first time we want to access the DB. For fetch this is always in buildProviderMigrator, for drift and policy functions it's just before the respective module executor calls.


roneli (approved):
Good catch

PR #443, opened 2022-01-26
fix: Panic when canceled

disq:


spangenberg (approved):

pkg/ui/console/client.go
```diff
@@ -130,6 +130,9 @@ func (c Client) Fetch(ctx context.Context, failOnError bool) error {
 			return err
 		}
 
+		if response == nil {
```
roneli, line 133: I think this is the wrong fix, so if we got cancelled we don't want to print fetch complete, probably fetch cancelled.  (talking about L:145, should if there). Moreover, we might need to improve the fetch cancel to still return what was finished and mark the rest cancelled in the client's Fetch but that is a bigger PR.

roneli (changes requested):
Good catch, but this is a band aid fix, we can merge it, but I think the user should get an appropriate message or handling if we cancelled.

pkg/ui/console/client.go
disq: Fixed, I'm not sure if that's OK or we should return error (if failOnError is true maybe? so that error code is not affected?)

roneli (approved):
LGTM

Issue #446, opened 2022-01-27
Persistent data folder should be configurable


Saare:
### Is your feature request related to a problem? Please describe.

I am trying to create an isolated environment for cloudquery due to company policy.

The cli allows me to configure the paths for plugins, policies and configuration - but the default `.cq` directory is still being created when running `cloudquery init` with two files in it: `last-update-check` and `telemtry-random-id`. 

```bash
cloudquery init aws --plugin-dir /.my-cd-directory/providers --policy-dir /.my-cd-directory/policies --config /.my-cd-directory/config.hcl --enable-file-logging=false
```

### Describe the solution you'd like

My suggestion is to add a cli parameter (something like `--persistent-data-dir`), that defaults to `./.cq`, in which such data is stored.

PR #456, opened 2022-01-31
feat: Configurable data folder

disq:
Implements #446.

```
	rootCmd.PersistentFlags().String("data-dir", "./.cq", "Directory to save and load CloudQuery persistent data to (env: CQ_DATA_DIR)")
	rootCmd.PersistentFlags().String("plugin-dir", "./.cq/providers", "Directory to save and load CloudQuery plugins from (env: CQ_PLUGIN_DIR)")
	rootCmd.PersistentFlags().String("policy-dir", "./.cq/policies", "Directory to save and load CloudQuery policies from (env: CQ_POLICY_DIR)")
```

Little bit too much. Maybe migrate away from plugin-dir and policy-dir in the future


spangenberg (approved):
LGTM

roneli (approved):
LGTM, agree we should deprecate plugin/policy dir or default them to data-dir/policies|plugin if the flag isn't given. Docs need update PR.

disq:
@roneli so, I'll make those two options hidden and empty... if empty, it'll default to datadir/[policies|providers]. wdyt?

roneli (approved):
LGTM

PR #459, opened 2022-02-01
feat: Small improvements

irmatov:
* Remove double import of a package
* Rewrite Client.CheckForProviderUpdates test
    Rewrite the test so that it does not depend on database availability
    and network access to a git repo.
* Add a unit test for Hub.CheckProviderUpdate
* A few cleanups
    * introduce ProviderRepoName function
    * drop unused Hub.getLatestReleaseVersion

### DIFF
--- a/pkg/plugin/registry/hub.go
+++ b/pkg/plugin/registry/hub.go
@@ -43,10 +43,6 @@ type Hub struct {
 	providers map[string]ProviderDetails
 }
 
-const (
-	providerDisplayMsg = "cq-provider-%s@%s"
-)
-
 type Option func(h *Hub)
 
 func NewRegistryHub(url string, opts ...Option) *Hub {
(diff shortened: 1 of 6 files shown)


pkg/plugin/registry/hub.go
```diff
@@ -41,20 +45,26 @@ type Hub struct {
...
+	}
+}
+
 func NewRegistryHub(url string, opts ...Option) *Hub {
 	h := &Hub{
-		PluginDirectory: filepath.Join(".", ".cq", "providers"),
-		Logger:          logging.NewZHcLog(&zerolog.Logger, ""),
-		url:             url,
-		providers:       make(map[string]ProviderDetails),
+		PluginDirectory:  filepath.Join(".", ".cq", "providers"),
```
disq, line 63: This needs an update now that #456 is merged?

disq (changes requested):

roneli (approved):
LGTM, nothing more to add that @disq didn't point out.

pkg/plugin/registry/hub.go
irmatov: Right, I'll do a rebase

irmatov: Ehm, looks like at least for this PR purposes there is no change.

disq (approved):

Issue #462, opened 2022-02-02
Target data fetches to a subset of accounts or resource types


paul-e-allen:
### Is your feature request related to a problem? Please describe.
n/a

### Describe the solution you'd like
Since data fetching can be time-consuming if there are a lot of AWS accounts, Azure subscriptions, etc. configured in `config.hcl`, it would be nice to be able to target a fetch to a subset of the accounts configured. 

Another targeting dimension might be the resource type. E.g., refresh the data for all accounts, but just for the `ec2.vpcs` resource type.

### Describe alternatives you've considered
You could use a secondary `config.hcl` that contains configuration for just the target account(s) or resource types.

roneli:
lets say we have the following config:
```
provider "azure" {
  resources = ["*"]
}

provider "aws" {
  configuration {
    aws_debug = false
  }
  // list of resources to fetch
  resources = [
    "s3.buckets"
  ]
  // enables partial fetching, allowing for any failures to not stop full resource pull
  enable_partial_fetch = true
  // Limit provider to fetch only 5 resources at a given time
  max_parallel_resource_fetch_limit = 5
}


provider "aws" {
  alias = "aws-2"
  configuration {
    aws_debug = false
  }
  // list of resources to fetch
  resources = [
    "*"
  ]
  // enables partial fetching, allowing for any failures to not stop full resource pull
  enable_partial_fetch = true
}
```

Would adding a keyword to fetch i.e `cq fetch azure aws-2` be helpful? Where this would run aws-2 block and azure provider blocks only.

 I am trying to think how to substitute resources via command line that will make it simpler to give it more granular control.

paul-e-allen:
Yes, that seems like a reasonable solution.

paul-e-allen:
I've been thinking a little more about that possible solution. As I look forward to support for AWS Organizations and Org-aware configuration for the AWS provider, I'm not sure the alias idea would work well. It _could_ still work, but it would destroy the utility of the Org-aware provider configuration (where we just provide a role in the master account, and name a role to assume in linked accounts) because you would end up having to list out every account that you might want to target directly (with an alias).

Still, the alias approach is better than nothing.

roneli:
Hi @paul-e-allen we release this feature in v0.20.2, see examples in #468, you can close this issue if this feature solves this for you.

paul-e-allen:
I've finally had a chance to try out this new feature. I like it! Thanks!

PR #468, opened 2022-02-04
feat: Fetch only specified providers/resources

disq:
```
Usage:
cloudquery fetch
cloudquery fetch aws
cloudquery fetch aws gcp
cloudquery fetch aws:ec2.instances,s3.buckets gcp
cloudquery fetch aws_2:ec2.instances,s3.buckets
cloudquery fetch aws_2 gcp_2
```

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -41,7 +41,7 @@ func init() {
 
 // filterConfigProviders gets a list of "providerAlias:resource1,resource2" items and updates the given config, removing non-matching providers
 // valid usages:
-// "aws" or "aws:*" (all resources)
+// "aws" or "aws:*" (all resources specified in the config)
 // "aws:ec2.instances,s3.buckets" (only ec2.instances and s3.buckets)
 func filterConfigProviders(list []string) func(*config.Config) error {
 	return func(cfg *config.Config) error {
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -2,7 +2,10 @@ package cmd
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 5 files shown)


roneli:
Related https://github.com/cloudquery/cloudquery/issues/462

cmd/fetch.go
```diff
@@ -35,3 +38,85 @@ func init() {
 	_ = fetchCmd.Flags().MarkHidden("redact-diags")
 	rootCmd.AddCommand(fetchCmd)
 }
+
+// filterConfigProviders gets a list of "providerAlias:resource1,resource2" items and updates the given config, removing non-matching providers
+// valid usages:
+// "aws" or "aws:*" (all resources)
+// "aws:ec2.instances,s3.buckets" (only ec2.instances and s3.buckets)
+func filterConfigProviders(list []string) func(*config.Config) error {
```
roneli, line 46: I am still not a fan of this implementation, as it feels harder, also its limited as it will be harder to upgrade it for example I can't do aws:ec2.* for example.

I think the following should be done -> parse fetch arg into a structure i.e RequestedProviders.
-> pass this to the fetch command. 
-> in fetch would loop over requested providers if its empty would loop over all providers in config. 
 ---> if requested providers is not empty, for each provider then get the provider from config, if we also requested resources swap the resources in the fetch request.

roneli (changes requested):
I think we can make this more simplified, maybe I am missing something, Added a comment how I think it should be achieved we can discuss this further if what I suggested wasn't clear enough.

roneli (approved):

PR #470, opened 2022-02-06
feat: Remove auto-prefixing env variables in config

yevgenypats:
We will have also to update docs and lambda but I think there is no need for this auto-prefixing of environment variables as it's just confusing (hit that in our k8s deployment). This way what you see in the config is what you get in the env variable.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -74,7 +74,7 @@ func Initialize(ctx context.Context, providers []string) error {
 
 	rootBody.AppendBlock(cqBlock)
 	cfg, diags := config.NewParser(
-		config.WithEnvironmentVariables(config.EnvVarPrefix, os.Environ()),
+		config.WithEnvironmentVariables(os.Environ()),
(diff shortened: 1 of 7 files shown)


bbernays:
If we do this, how are we going to be sure we don't having a naming collision with other tools?

disq:
Instead of removing auto-prefixing, why not _allow_ only CQ_ prefixed env variables to be used? other variables (in the same "hey I'm using an env variable" format) should generate an error.

bbernays:
While I like that idea, would this only apply to CQ specific variables? I am not sure how we could do that with regards to the env variables that the Cloud providers use to setup credentials

disq:
But we never put credentials into config.hcl (except the GCP _JSON thing maybe?)? It's only db credentials I think

erezrokah:
Should we close this PR until we figure out the path we'd like to take?

erezrokah:
Going to close this as stale. If I got this wrong please let me know and I'll re-open

PR #474, opened 2022-02-07
fix: New release not found while building

roneli:
while creating a new release, the tag is created but the binary hasn't compiled yet, we want to avoid this by fetching the prior release until all assets are added.


roneli:
Closed in favor of #483

PR #475, opened 2022-02-07
feat: Use provider name/version as exception type when reporting diagnostics to Sentry

disq:
The commented part will be removed once we finalize on a convention


roneli (approved):
LGTM

PR #476, opened 2022-02-07
fix: Added error when connection configuration is not set

amanenk:
fixes nil pointer exception if connection config is not set


disq:
When does this happen?

disq (approved):
I don't know how this happens, but in general we're against panics, so approving

disq (commented):
erm, why is `cfg.CloudQuery.Connection` a pointer, maybe make it a non-ptr value?

PR #483, opened 2022-02-09
feat: Use firestore for releases

disq:
[This](https://firestore.googleapis.com/v1/projects/hub-cloudquery/databases/(default)/documents/orgs/cloudquery/providers/aws/versions?mask.fieldPaths=tag&orderBy=v_major+desc%2C+v_minor+desc%2C+v_patch+desc%2C+published_at+desc&pageSize=1) is the URL it builds


go.mod
```diff
@@ -33,6 +33,7 @@ require (
 )
 
 require (
+	cloud.google.com/go/firestore v1.1.0
```
roneli, line 36: no need this package

PR #486, opened 2022-02-11
fix: Don't return an error if encountering an empty subpolicy

shimonp21:
Also removed the "multilayer policies with invalid selector" test (since now,
the "executor" won't actually return an error in the "invalid selector" case - it
will be up to the "manager").

### DIFF
--- a/pkg/policy/execute.go
+++ b/pkg/policy/execute.go
@@ -123,9 +123,13 @@ func (e *Executor) with(policy string, args ...interface{}) *Executor {
 
 // Execute executes given policy and the related sub queries/views.
 func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Policy) (*ExecutionResult, error) {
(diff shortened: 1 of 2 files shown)


pkg/policy/execute.go
```diff
@@ -124,7 +124,7 @@ func (e *Executor) with(policy string, args ...interface{}) *Executor {
 // Execute executes given policy and the related sub queries/views.
 func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Policy) (*ExecutionResult, error) {
 	if !policy.HasChecks() {
-		return nil, fmt.Errorf("no checks or policies to execute")
+		e.log.Warn("no checks or policies to execute")
```
roneli, line 127: Would set result := ExecutionResult{PolicyName: req.Policy.Name, Passed: true, Results: make([]*QueryResult, 0)} and the if below do return result. 

This define is instead of line 136

roneli (changes requested):
Looks good, have a minor suggestion.

roneli (approved):

PR #494, opened 2022-02-17
fix: Upgrade schema in history mode

disq:
Views need to be dropped before any migrations can be applied (except creating new tables)


roneli (approved):
LGTM, can we add a test for this? upgrade ->  works -> downgrade -> works -> upgrade -> works ?

roneli (approved):
LGTM

PR #500, opened 2022-02-22
fix: Snapshot Persistance

bbernays:


pkg/policy/snapshots.go
```diff
@@ -207,19 +207,21 @@ func (ce *Executor) checkTableExistence(ctx context.Context, tableName string) (
 	explainQuery := fmt.Sprintf("select pg_get_viewdef('%s'::regclass::oid) ", tableName)
 	rows, err := ce.conn.Query(ctx, explainQuery)
 	if err != nil {
+		ce.log.Error("error running explain", "tableName", tableName)
```
disq, line 210: log the `err` as well?

pkg/policy/snapshots.go
```diff
@@ -207,19 +207,21 @@ func (ce *Executor) checkTableExistence(ctx context.Context, tableName string) (
...
 		return "", err
 	}
+	if err := rows.Err(); err != nil {
```
disq, line 213: As per doc, `rows.Err` returns errors occured while reading, so moving this before the reading part (which is `Scan`) is not correct? But there's a similar thing in line 152 of the same file where the rows.Err block should be moved to after rows.Next around line 111.

disq (changes requested):

pkg/policy/snapshots.go
bbernays: Thanks for that catch! Fixed it

disq (approved):

disq:
LGTM :)

PR #502, opened 2022-02-23
fix: Init command cleanup

disq:
- [x] Don't set `source` if default org
- [x] Include drift config only with AWS
- [x] Support `cloudquery init provider@version`
- [x] Add hint to use `cloudquery init <provider>` if `config.hcl` loading fails with `fs.ErrNotExist`

### DIFF
--- a/pkg/plugin/registry/organization.go
+++ b/pkg/plugin/registry/organization.go
@@ -3,6 +3,8 @@ package registry
 import (
 	"fmt"
 	"strings"
+
+	"github.com/Masterminds/semver/v3"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


pkg/plugin/registry/organization.go
```diff
@@ -18,12 +20,36 @@ func ParseProviderName(name string) (string, string, error) {
 		return strings.ToLower(names[0]), names[1], nil
 	}
 	if len(names) == 1 {
-		return defaultOrganization, name, nil
+		return DefaultOrganization, name, nil
 	}
-	return "", "", fmt.Errorf("invalid provider name %s", name)
+	return "", "", fmt.Errorf("invalid provider name %q", name)
```
bbernays, line 25: Why do you use `%q` and not `%w`?

pkg/plugin/registry/organization.go
```diff
@@ -18,12 +20,36 @@ func ParseProviderName(name string) (string, string, error) {
...
 }
 
 // ProviderRepoName returns a repository name for a given provider name.
 func ProviderRepoName(name string) string {
 	return fmt.Sprintf("cq-provider-%s", name)
 }
+
+// ParseProviderNameWithVersion parses the given string as "repository/provider@version" or "provider@version" (or @version omitted, defaulting to "latest")
+func ParseProviderNameWithVersion(nameWithVersion string) (string, string, string, error) {
+	versionParts := strings.Split(nameWithVersion, "@")
```
bbernays, line 35: I think we should be documenting what characters can be in a provider name

disq: Screen Shot 2022-02-23 at 17 35 15

Yes but a repo with `@` in the name is not possible with github anyway...

pkg/plugin/registry/organization.go
bbernays: Ohh I see, for some reason my brain was assuming you were wrapping the error like you did above
`return fmt.Errorf("could not parse requested provider: %w", err)`

pkg/plugin/registry/organization.go
bbernays: For my own knowledge, do providers have to be on Github? I know that policies can be in many different locations

disq: Looks like it can be anywhere as long as it's indexed in our Firestore (`registry/hub.go`)

bbernays (approved):
lgtm

yevgenypats (approved):

PR #503, opened 2022-02-23
chore: Add codeowners

bbernays:


yevgenypats (changes requested):
Can you add this in cloudquery/.github repository so it will automatically will add to all providers,policies as well?

bbernays:
Closing because moved to `cloudquery/.github`

PR #510, opened 2022-03-02
fix: Validate DB connection with an explicit timeout, rather than the…

shimonp21:
… default timeout

- The default timeout is many minutes, way too long to give the user feedback about failed db connection.
- Note that cq-provider-sdk/database/postgres uses lazy-connect, so it won't return an error in case of failed DB connection.

### DIFF
--- a/pkg/client/database/postgres/executor.go
+++ b/pkg/client/database/postgres/executor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	sdkpg "github.com/cloudquery/cq-provider-sdk/database/postgres"
 	"github.com/hashicorp/go-hclog"
@@ -36,6 +37,10 @@ func (e Executor) Validate(ctx context.Context) (bool, error) {
 		return false, err
 	}
 
+	if err := ValidatePostgresConnection(ctx, pool); err != nil {
+		return false, err
+	}
+
 	if err := ValidatePostgresVersion(ctx, pool, MinPostgresVersion); err != nil {
 		return true, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/client/database/postgres/executor.go
```diff
@@ -51,6 +56,18 @@ func (e Executor) Finalize(_ context.Context, err error) error {
 	return err
 }
 
+// Validates that we can actually connect to the postgres database.
+func ValidatePostgresPing(ctx context.Context, pool *pgxpool.Pool) error {
```
disq, line 60: Why not keep it as `ValidatePostgresConnection`? Ping is implementation detail.

pkg/client/database/postgres/executor.go
```diff
@@ -51,6 +56,18 @@ func (e Executor) Finalize(_ context.Context, err error) error {
 	return err
 }
 
+// Validates that we can actually connect to the postgres database.
```
disq, line 59:
```suggestion
// ValidatePostgresConnection validates that we can actually connect to the postgres database.
```
(valid godoc format as described [here](https://go.dev/blog/godoc))

disq (changes requested):

pkg/client/database/postgres/executor.go
```diff
@@ -51,6 +56,18 @@ func (e Executor) Finalize(_ context.Context, err error) error {
 	return err
 }
 
+// ValidatePostgresConnection validates that we can actually connect to the postgres database.
+func ValidatePostgresConnection(ctx context.Context, pool *pgxpool.Pool) error {
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	if err := pool.Ping(ctx); err != nil {
```
disq, line 64: this could just be `return pool.Ping(ctx)`

disq (approved):

PR #515, opened 2022-03-03
feat: Drift: Add @getbool modifier

disq:
This is used if the specified object may or may not be nil, and we want to extract a bool value from it


roneli (approved):
LGTM

PR #522, opened 2022-03-08
feat: Add file() func to config parser

disq:
So one can do `dsn = file("dsn.txt")`


disq:
`convert/stdlib_test.go` already has a `TestFileFunc`

PR #525, opened 2022-03-08
fix: Foreign Keys in Snapshots

bbernays:
When restoring a snapshot prior to copying the data drop all FK requirements

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -666,7 +666,7 @@ func (c *Client) BuildProviderTables(ctx context.Context, providerName string) (
 		}
 
 		c.Logger.Error("BuildProviderTables failed", "error", retErr)
-		retErr = fmt.Errorf("incompatible provider schema: Please drop provider tables and recreate, alternatively execute `cloudquery provider drop %s`", providerName)
+		retErr = fmt.Errorf("Incompatible provider schema: Please drop provider tables and recreate, alternatively execute `cloudquery provider drop %s`", providerName)
 	}()
 
 	// create migration table and set it to version based on latest create table
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -666,7 +666,7 @@ func (c *Client) BuildProviderTables(ctx context.Context, providerName string) (
 		}
 
 		c.Logger.Error("BuildProviderTables failed", "error", retErr)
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -666,7 +666,7 @@ func (c *Client) BuildProviderTables(ctx context.Context, providerName string) (
 		}
 
 		c.Logger.Error("BuildProviderTables failed", "error", retErr)
-		retErr = fmt.Errorf("Incompatible provider schema: Please drop provider tables and recreate, alternatively execute `cloudquery provider drop %s`", providerName)
+		retErr = fmt.Errorf("incompatible provider schema: Please drop provider tables and recreate, alternatively execute `cloudquery provider drop %s`", providerName)
```
disq, line 669: This was done (casing) deliberately as this error message is not internal and it's what the user sees.

bbernays: Ahh. Ok, that makes sense I wonder if there is a way to let the gofmt know it is on purposes because I was getting errors that it was "wrong"

disq: maybe something like `//nolint` ?

pkg/policy/snapshots.go
```diff
@@ -60,6 +60,10 @@ func RestoreSnapshot(ctx context.Context, conn LowLevelQueryExecer, log hclog.Lo
 		return fmt.Errorf("invalid filename: %q", fileName)
 	}
 	source_name := strings.TrimPrefix(strings.TrimSuffix(fileName, ".csv"), "table_")
+	err = deleteFKs(ctx, conn, log, source_name)
+	if err != nil {
+		return fmt.Errorf("error removing fks from %q: %w", source_name, err)
+	}
```
roneli, line 66: question around this why isn't the data restore order consistent i.e parent -> child ? Then there shouldn't be any FK issue and no need to delete snapshots.

roneli (changes requested):
Added a comment, other than that looks good.

pkg/policy/snapshots.go
bbernays: The idea around snapshots was to store the minimum amount of data required for the sql query to run

roneli (approved):

PR #528, opened 2022-03-15
fix: Warn on providers with no resources requested

### LINKED ISSUE #514: Add log that indicates that no resources are specified (or maybe a warning)

If not resources are specified it can be tricky to see why fetch is just succeeding and exiting with success.

Add a log:

`WRN: no resources specified, cloudquery will configure the provider and exit.`

roneli:
Fixes #514 

Run example:
image

### DIFF
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -96,7 +96,7 @@ func (c Client) DownloadProviders(ctx context.Context) error {
 	}
 	// sleep some extra 300 milliseconds for progress refresh
 	if ui.IsTerminal() {
-		time.Sleep(300 * time.Millisecond)
+		time.Sleep(500 * time.Millisecond)
 		c.updater.Wait()
 	}
 	ui.ColorizedOutput(ui.ColorProgress, "Finished provider initialization...\n\n")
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):

pkg/ui/console/client.go
```diff
@@ -96,7 +96,7 @@ func (c Client) DownloadProviders(ctx context.Context) error {
 	}
 	// sleep some extra 300 milliseconds for progress refresh
 	if ui.IsTerminal() {
-		time.Sleep(300 * time.Millisecond)
+		time.Sleep(500 * time.Millisecond)
```
bbernays, line 99: Maybe also update the comment above

bbernays (approved):
Minor nit

pkg/ui/console/client.go
```diff
@@ -96,7 +96,7 @@ func (c Client) DownloadProviders(ctx context.Context) error {
 	}
 	// sleep some extra 300 milliseconds for progress refresh
```
bbernays, line 97:
```suggestion
	// sleep some extra 500 milliseconds for progress refresh
```

PR #529, opened 2022-03-15
Use of sha1 is insecure

matthew694694:
Hello,

I found the use of sha1 within the project, and I would suggest updating the crypto package to something with greater integrity.

The use of sha1 is not recommended (https://pkg.go.dev/crypto/sha1), so I have updated it to the sha512 package in the files I found sha1 in.

Thanks

PS I am not sure of the testing edict or process on this repo do forgive me if I have missed a step :)

### DIFF
--- a/internal/telemetry/telemetry.go
+++ b/internal/telemetry/telemetry.go
@@ -2,7 +2,7 @@ package telemetry
 
 import (
 	"context"
-	"crypto/sha1"
+	"crypto/sha512"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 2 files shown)


disq:
Thank you for the PR. This is a non-issue because we don't use this cipher cryptographically. Could've used MD5 or CRC32 and it would still be OK.

There's no reason to break backwards compatibility with previously hashed entries (and have longer hashes) in our internal systems just to feel better because we've shunned an older hashing method.

matthew694694:
Thanks for the feedback @disq :)

PR #531, opened 2022-03-15
fix: Downgrade provider

### LINKED ISSUE #496: provider downgrade doesn't work
`cloudquery provider downgrade [provider]` will not work:
- if `config.hcl` contains the current provider version, it will exit with `no change`
- if `config.hcl` contains the provider version we want to downgrade to, down migrations from the future (from the version we're currently at) don't exist yet, so migrate fails

Maybe downgrade should first take/upgrade to `latest` for the provider, then downgrade to the specified (in config) version.

related #378

disq:
Closes #496.

This is extra bad, but given the way everything is set up this is what I ended up with.

### DIFF
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -418,8 +418,6 @@ func (c Client) DowngradeProviders(ctx context.Context, args []string) error {
 		return err
 	}
 
-	ui.ColorizedOutput(ui.ColorProgress, "Will upgrade providers first, before downgrade... %s\n\n", strings.Join(providers.Names(), ", "))
-
 	provVersions := make(map[string]string, len(providers))
 	for _, p := range providers {
 		provVersions[p.Name] = p.Version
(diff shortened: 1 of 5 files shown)


pkg/ui/console/client.go
```diff
@@ -407,28 +407,60 @@ func (c Client) UpgradeProviders(ctx context.Context, args []string) error {
...
+
+	provVersions := make(map[string]string, len(providers))
+	for _, p := range providers {
+		provVersions[p.Name] = p.Version
+		p.Version = migrator.Latest
+	}
+
+	newCfg := *c.cfg
+	newCfg.CloudQuery.Providers = providers
+	conWithLatest, err := CreateClientFromConfig(ctx, &newCfg)
```
roneli, line 431: wouldn't conWithLatest (version X > Y) be able to downgrade Y -> Z  (Y > Z) I don't think we need to do an upgrade with the cqWithLatest.

roneli (changes requested):
Looks okay, I think we can avoid the extra upgrade, but fix me if I am incorrect in my assumption.

pkg/ui/console/client.go
disq: Fixed

PR #534, opened 2022-03-16
fix: Make Date a static value

### LINKED ISSUE #509: cloudquery version shows runtime date
`cloudquery version` seems to show the runtime date (time.Now) instead of the date of the compile.

in `cmd/root.go`, `Date` variable is by default set to `time.Now().String()` (on initialization) but apparently it's not updated during build-time.

irmatov:
Make Date a static value so it can be overriden during build time.

Closes #509


disq (approved):
LGTM, `Date` could also have a friendy default (without any forced date-formatting) like "unset" etc. as well...

PR #540, opened 2022-03-18
fix: Validate Timescale Version

bbernays:
Output when running against pg12 with tsdb 2.6.0:
```
cloudquery % go run main.go fetch
Initializing CloudQuery Providers...

✓ cq-provider-aws@v0.10.14 verified    0s  100 %

Finished provider initialization...

Upgrading CloudQuery providers aws
```

Output when running against pg 12 with tsdb 1.7.5:
```
cloudquery % go run main.go fetch
❌ Failed to initialize client. Error: validate: unsupported Timescale version: 1.7.5. (should be >= 2.0.0)

validate: unsupported Timescale version: 1.7.5. (should be >= 2.0.0)
```

### DIFF
--- a/pkg/client/database/timescale/timescale.go
+++ b/pkg/client/database/timescale/timescale.go
@@ -20,8 +20,9 @@ import (
 const (
 	validateTimescaleInstalled = `SELECT EXISTS(SELECT 1 FROM pg_extension where extname = 'timescaledb')`
 	timescaleVersionQuery      = `SELECT extversion FROM pg_catalog.pg_extension WHERE extname='timescaledb';`
-	MinTimescaleVersion        = version.Must(version.NewVersion("2.0"))
 )
+
+var MinTimescaleVersion = version.Must(version.NewVersion("2.0"))
 
 // queryRower helps with unit tests
 type queryRower interface {
--- a/pkg/client/database/timescale/timescale.go
+++ b/pkg/client/database/timescale/timescale.go
@@ -20,9 +20,8 @@ import (
 const (
 	validateTimescaleInstalled = `SELECT EXISTS(SELECT 1 FROM pg_extension where extname = 'timescaledb')`
 	timescaleVersionQuery      = `SELECT extversion FROM pg_catalog.pg_extension WHERE extname='timescaledb';`
+	MinTimescaleVersion        = version.Must(version.NewVersion("2.0"))
 )
-
-var MinTimescaleVersion = version.Must(version.NewVersion("2.0"))
 
 // queryRower helps with unit tests
 type queryRower interface {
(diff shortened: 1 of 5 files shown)


pkg/client/database/timescale/timescale.go
```diff
@@ -97,3 +113,39 @@ func (e Executor) Finalize(ctx context.Context, retErr error) error {
...
+	if len(fields) != 1 {
+		return nil, fmt.Errorf("failed to parse version: %s, %q", result, fields)
+	}
+	return version.NewVersion(fields[0])
+}
+
+// ValidateTimescaleVersion checks that Timescale plugin version available through pool is not lower than wanted version.
+// In this case it returns nil. Otherwise returns error describing current and desired version or any other error encountered
+// during the check.
+func ValidateTimescaleVersion(ctx context.Context, pool *pgxpool.Pool, want *version.Version) error {
```
disq, line 133: I would use the const `MinTimescaleVersion` here and not require `want`. (in fact the same logic applies to `ValidatePostgresVersion`)

pkg/client/database/timescale/timescale.go
```diff
@@ -97,3 +113,39 @@ func (e Executor) Finalize(ctx context.Context, retErr error) error {
...
+	return doValidateTimescaleVersion(ctx, conn, want)
+}
+
+func doValidateTimescaleVersion(ctx context.Context, q queryRower, want *version.Version) error {
+	got, err := runningTimescaleVersion(ctx, q)
+	if err != nil {
+		return fmt.Errorf("error getting Timescale version: %w", err)
+	}
+	if got.LessThan(want) {
+		return fmt.Errorf("unsupported Timescale version: %v. (should be >= %v)", got, want)
```
disq, line 148: not sure if needed, but:

```suggestion
		return fmt.Errorf("unsupported Timescale version: %s. (should be >= %s)", got.String(), want.String())
```

disq (approved):
LGTM with nits

pkg/client/database/timescale/timescale.go
bbernays: You can't make it a const:
```
version.Must(version.NewVersion("2.0")) (value of type *version.Version) is not constant 
```

pkg/client/database/timescale/timescale.go
bbernays: Fixed in both TS and PG

pkg/client/database/timescale/timescale.go
disq: no use the const in the top line with the string.

disq: ok saw the edit, `var` is also ok. but my point was, to make it like:
`func ValidateTimescaleVersion(ctx context.Context, pool *pgxpool.Pool) error`

ValidateTimescaleVersion uses the const inside of the fn, caller doesn't need to specify or know about what version or what const is required.

bbernays: This change would require changes to the tests

disq: Tests would use `doValidateTimescaleVersion`?

PR #546, opened 2022-03-21
fix: Policy describe should run without config.hcl requirement

zagronitay:


disq:
Yes but only if the cq files are under `.cq`. It won't work if you have them somewhere else because you can't provide config.hcl if you wanted to now. You're locked into the null client.

cmd/util.go
```diff
@@ -87,6 +87,13 @@ func handleConsole(ctx context.Context, tele *telemetry.Client, cmd *cobra.Comma
 		delayMessage = false
 	case "init":
 		// No console client created here
+	case "describe":
+		var err error
+		c, err = console.ClientFactory(ctx, &cfgPath, cfgMutator)
+		if err != nil {
+			return err
+		}
+		defer c.Client().Close()
```
roneli, line 96: nit: why not use the console.CerateClient that will return a NullClient / Normal client based on if cfg path exists and returns a console client.

roneli (approved):
LGTM, had one nit.

PR #556, opened 2022-03-26
feat: Limit scope of views

shimonp21:


pkg/policy/execute.go
```diff
@@ -158,6 +155,11 @@ func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Pol
 		}
 	}
 
+	if err := e.createViews(ctx, policy); err != nil {
+		return nil, err
+	}
+	defer e.deleteViews(ctx, policy)
+
```
roneli, line 162: In theory the better design here is to create a session at this point (pool.Acquire) run the checks and close the session rather then create delete. The original passing of only a connection is a flaw by design. (I am not sure if its a simple fix as this, so if it isn't maybe add a todo here).

roneli (commented):
Had one comment, once it's resolved either with todo or the actual fix will approve and we can merge.

pkg/policy/execute.go
shimonp21: Added TODO for now. Definitely agree that a new session per policy is a better idea, but this refactor is not trivial.

roneli (approved):
LGTM

PR #559, opened 2022-03-29
feat: Add console colour disable

roneli:
Allow to disable color formatting in logger


disq:
Why not use something like https://github.com/jwalton/go-supportscolor ?

bbernays:
> Why not use something like https://github.com/jwalton/go-supportscolor ?

Are we sure that the source of the issue is Windows 10? Maybe we want to use that package to automatically adjust, but still give the user to force disable

disq (approved):

roneli:
Closes #560

PR #563, opened 2022-03-30
feat: Drift: Improvements

disq:
- Turns out the "combination feature" was kind of doable with the `path` statement but I still wanted to leave it in for now.
- Added @if to GJSON `value|@if:this,that` (if value is `this` it will be replaced with `that`. if not, it will be left as-is)
- Added @split to GJSON to split a string into an array: `value1,value2|@split:,` => `[ value1, value2 ]`
- Added ability to run GJSON filters in the cloud provider attributes (in the attribute map) and updated drift protocol to v2
- Added ability to define a single cloud resource multiple times (explicitly, using a 'hash' syntax, so that it doesn't get confused/lost in the map accesses) to be able to cover a blanket resource (think `iam.accounts` or `ec2.regional_config`) using multiple TF resources.
- Removed `-tf-mode` cli param (which was pointless)

### DIFF
--- a/pkg/module/drift/config.go
+++ b/pkg/module/drift/config.go
@@ -451,17 +451,19 @@ func (d *Drift) applyProvider(cfg *ProviderConfig, p *cqproto.GetProviderSchemaR
 	checkEnabled := len(cfg.CheckResources) > 0
 
 	for resName, res := range cfg.Resources {
+		resNameParts := strings.SplitN(resName, "#", 2)
+
 		// CheckResources / IgnoreResources broad strokes...
 		if checkEnabled {
 			res.acl.AllowEnabled = true
-			res.acl.Allow = append(cfg.CheckResources.ByType(resName), allChecks...)
+			res.acl.Allow = append(cfg.CheckResources.ByType(resNameParts[0]), allChecks...)
 			if !res.acl.Allow.AllInstances() && !res.acl.Allow.HasTags() {
 				delete(cfg.Resources, resName)
 				continue
 			}
 		}
 
-		res.acl.Ignore = append(cfg.IgnoreResources.ByType(resName), allIgs...)
+		res.acl.Ignore = append(cfg.IgnoreResources.ByType(resNameParts[0]), allIgs...)
 		if res.acl.Ignore.AllInstances() {
 			delete(cfg.Resources, resName)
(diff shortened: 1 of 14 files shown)


pkg/module/drift/config.go
```diff
@@ -451,17 +451,19 @@ func (d *Drift) applyProvider(cfg *ProviderConfig, p *cqproto.GetProviderSchemaR
 	checkEnabled := len(cfg.CheckResources) > 0
 
 	for resName, res := range cfg.Resources {
+		resNameParts := strings.SplitN(resName, "#", 2)
```
roneli, line 454: nit: maybe create a small function that splits and returns string string ?

pkg/module/drift/model.go
```diff
@@ -147,8 +147,16 @@ func (rs *Results) process() {
 		if len(ids) == 0 {
 			return
 		}
+		typeParts := strings.SplitN(r.ResourceType, "#", 2)
```
roneli, line 150: see comment above

pkg/module/drift/model.go
```diff
@@ -285,3 +292,27 @@ func (rs *Results) process() {
 
 	rs.Text = strings.Join(lines, "\n")
 }
+
+// UniqueSort sorts input alphabetically and returns only non-duplicate entries
+func UniqueSort(input []string) []string {
```
roneli, line 297: feels like this function should be in a different package, maybe in internal package under sort.go

roneli (changes requested):
Looks good, had some comments and suggestions around new added function etc'

pkg/module/drift/terraform.go
```diff
@@ -316,6 +334,14 @@ func RenderDriftTable(resName string, resources map[string]*ResourceConfig, clou
 		return table
 	}
 
+	for i := range differentIDs {
+		for j := range differentIDs[i].Attributes {
+			if alist[j].CloudMod != nil {
+				differentIDs[i].Attributes[j] = alist[j].CloudMod(differentIDs[i].Attributes[j])
+			}
+		}
+	}
```
roneli, line 343: nit: document this loop

roneli (approved):
LGTM, had one nit.

Issue #562, opened 2022-03-29
Policy Tests broken


bbernays:
### Describe the Bug

We run a `cq policy run ` for every pr in all of our policy repos. We used to run it as a sanity check to ensure that no sql syntax errors would block customers. Now that we enforce a fetch prior to running a policy that test stops before it ever even executes the policy

### Expected Behavior

SQL policies should run

### CloudQuery Version

0.22.0

### Additional Context

We have a few options:
1. run a fetch prior to execution
2. Pin cq version to <0.22
3. Insert a record to make it look like fetch occurred
4. expose an override flag to disable fetch check

shimonp21:
Discussed with Ron, chosen solution is overriding the 'fetch check' with a flag

PR #567, opened 2022-04-06
feat: Add '--disable-fetch-check' flag

shimonp21:
fixes: https://github.com/cloudquery/cloudquery/issues/562


cmd/policy_run.go
```diff
@@ -47,6 +48,10 @@ func init() {
 	flags := policyRunCmd.Flags()
 	flags.StringVar(&outputDir, "output-dir", "", "Generates a new file for each policy at the given dir with the output")
 	flags.BoolVar(&noResults, "no-results", false, "Do not show policies results")
+	flags.Bool("disable-fetch-check", false, "Disable checking if a respective fetch happened before running policies")
```
disq, line 51: `no-fetch-check`?
same goes for `disable-log-color` btw... we seem to have most of our "turn off" options as "no-option"

shimonp21: Hmmm, I like "enable/disable" much more, because it's a verb. @roneli - what do you think?

bbernays: I would vote for consistency

roneli: I think disable/enable is better, we should change all flags to disable/enable (but keep the old ones as well for backwards compatability eventually remove them)

irmatov, line 51: nit: This seems a bit out of style with other flags (which are bound to variables and pass their values as arguments to `RunPolicies`). As a consequence introduces dependency on viper package into pkg/policy package. Maybe we could combine those flags into a structure and still pass it as an argument..

roneli (approved):
LGTM, can merge once we decide on flag naming convention

cmd/policy_run.go
shimonp21: discussed in poll and changed back to disable/enable.

shimonp21: Discussed on slack :)

PR #575, opened 2022-04-07
fix: Drift: Remove version constraints, always include all parents

disq:
- Version constraints are not needed since HCL is bundled with the provider now
- Second commit fixes the issue about not being able to filter two resources down, using check_resources:

This would fail:
```hcl
      check_resources = [
        // "codepipeline.pipelines:*",
        // "aws_codepipeline_pipeline_stages:*",
        "aws_codepipeline_pipeline_stage_actions:*"
	]
```

but this would work:
```hcl
      check_resources = [
        // "codepipeline.pipelines:*",
        "aws_codepipeline_pipeline_stages:*",
        "aws_codepipeline_pipeline_stage_actions:*"
	]
```
(now the former also works)

### DIFF
--- a/pkg/module/drift/config.go
+++ b/pkg/module/drift/config.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/cloudquery/cq-provider-sdk/cqproto"
 	"github.com/hashicorp/go-hclog"
-	"github.com/hashicorp/go-version"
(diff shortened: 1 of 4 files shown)


roneli (approved):
LGTM, always great to see code simplified.

PR #576, opened 2022-04-09
fix: Replace non-supported satori/go.uuid with google/uuid

dronnix:
Hi!

The satori/go.uuid library isn't longer supported, and contains security [issues](https://github.com/satori/go.uuid/issues/115). So I've replaced this with google/uuid.


yevgenypats:
Thanks @dronnix ! This looks good though not sure about the severity of this as we are not using this for anything crypto/security related but prob will be a good best practice / future-proof PR in any case. I'll let @disq review.

disq (approved):
LGTM

PR #580, opened 2022-04-13
fix: Add safeguard for drop provider

disq:


cmd/provider.go
```diff
@@ -117,6 +122,7 @@ func init() {
 		"last-update is the duration from current time we want to remove resources from the database. "+
 			"For example 24h will remove all resources that were not update in last 24 hours. Duration is a string with optional unit suffix such as \"2h45m\" or \"7d\"")
 	providerRemoveStaleCmd.Flags().BoolVar(&dryRun, "dry-run", true, "")
+	providerDropCmd.Flags().BoolVar(&iAmSure, "i-am-sure", false, "Really drop tables for the provider")
```
roneli, line 125: nit: --force ? --approve-drop ?

roneli (approved):
LGTM had 1 nit.

cmd/provider.go
bbernays: I like --force

disq: Done!

disq: ...but `--force` implies "it's normally possible to do it without forcing, but for some reason, in this instance, we can't" which is not the case here.

bbernays (approved):

PR #594, opened 2022-04-22
feat: Configure to return diagnostics, upgrade SDK

disq:
goes with https://github.com/cloudquery/cq-provider-sdk/pull/226 and https://github.com/cloudquery/cq-provider-aws/pull/755
~a problem: provider.Configure calls fail with `diag.ACCESS` / `ERROR`. Since it's `ERROR` it's currently reported.~

### DIFF
--- a/internal/telemetry/error.go
+++ b/internal/telemetry/error.go
@@ -17,6 +17,8 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const ConfigureProviderResource = "provider.configure"
+
 var sqlStateRegex = regexp.MustCompile(`\(SQLSTATE ([0-9A-Z]{5})\)`)
 
 // RecordError should be called on a span to mark it as errored. Returns true if err was recorded.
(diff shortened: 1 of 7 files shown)


irmatov (approved):

roneli (approved):
LGTM

internal/telemetry/error.go
```diff
@@ -42,6 +45,13 @@ func RecordError(span trace.Span, err error, opts ...trace.EventOption) bool {
 
 // ShouldIgnoreDiag checks the wire-transferred diagnostic against errors we don't want to process.
 func ShouldIgnoreDiag(d diag.Diagnostic) bool {
+	if d.Severity() == diag.IGNORE || (d.Severity() == diag.WARNING && d.Type() == diag.ACCESS) || d.Type() == diag.USER {
+		return true
+	}
+	if d.Type() == diag.ACCESS && d.Description().Resource == ConfigureProviderResource {
```
roneli, line 51: not sure I like this ConfigureProviderResoruce, there is no such resource etc'; I understand we want to distinct diagnostic by "Type" but that is why we have Type function. My guess is we want to return ERROR + ACCESS. So there is an issue with the diagnostic design, we need maybe to add a new type / convert diags from the protocol perhaps to a new Type i.e CoreDiagnostic and have that have a new method perhaps? (not sure)

roneli (commented):
Added a comment, around the ConfigureProviderResoruce hack

internal/telemetry/error.go
disq: we already return ERROR+ACCESS but the issue is we _don't ignore_ ERROR+ACCESS. So the error wouldn't be ignored in that case and would be reported, as explained in the PR body.

roneli: see #596 created Fetch Diagnostics, I would argue to do the same for ConfigureDiagnostics. Moreover, would probably  create ResourceDiagnostic in SDK (but would do that in future PR), and that would hold resource id / resource name etc'.

disq: Doesn't work if we want to use squashed diagnostics.

PR #596, opened 2022-04-23
feat: User Diagnostics

roneli:
Clean console.Client, remove old telemetry and add print diags for all commands

- [x] Capture Diags on All commands (policy/drift/etc')
- [x] Pass cfghash etc' to analytics
- [x] Fix random id issues (on windows context parsing is incorrect) #599 
- [x] Make all commands return diags
- [x] Add diag.User, diag.Severity stringer -> will be added in rebase
- [x] ~Add diags classification calls~ - separate PR
- [ ] drift / policy run events


pkg/ui/console/diagnostics.go
```diff
@@ -27,6 +27,11 @@ func ClassifyDiagnostics(diags diag.Diagnostics) diag.Diagnostics {
 }
 
 func printDiagnostics(_ string, diags diag.Diagnostics, redactDiags, verbose bool) {
+	// Nothing to print
+	if diags.Len() == 0 {
```
disq, line 31: `!diags.HasDiags()`?

internal/analytics/client.go
```diff
@@ -0,0 +1,132 @@
...
+		return uuid.New()
+	}
+	return id
+}
+
+func (c Client) Capture(eventType string, providers registry.Providers, data interface{}, diags diag.Diagnostics) {
+	if c.disabled {
+		return
+	}
+	_ = Event{
```
disq, line 93: add TODO here

internal/analytics/client.go
```diff
@@ -0,0 +1,132 @@
...
+
+func SummarizeDiagnostics(diags diag.Diagnostics) DiagnosticsSummary {
+	summary := DiagnosticsSummary{
+		Total:      0,
+		ByType:     make(map[diag.DiagnosticType]int),
+		BySeverity: make(map[diag.Severity]int),
+	}
+	for _, d := range diags {
+		summary.Total += 1
+		if _, ok := summary.BySeverity[d.Severity()]; ok {
```
disq, line 119: why not just `summary.BySeverity[d.Severity()]++` or some form of that?

internal/analytics/client.go
```diff
@@ -0,0 +1,132 @@
...
+			summary.BySeverity[d.Severity()] = 1
+		} else {
+			summary.BySeverity[d.Severity()] += 1
+		}
+
+		if _, ok := summary.ByType[d.Type()]; ok {
```
disq, line 125: same here with `summary.ByType[d.Type()]++`

disq (commented):
Not sure it covers all cases, but here are some comments

pkg/core/diagnostics.go
```diff
@@ -0,0 +1,69 @@
+package core
+
+import "github.com/cloudquery/cq-provider-sdk/provider/diag"
+
+type FetchDiagnostic struct {
+	base diag.Diagnostic
```
disq, line 6: no need for base, just embed it directly. That way you don't need the overloaded methods unless you really want to overload.

pkg/errors/sentry.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"github.com/cloudquery/cloudquery/pkg/core"
+	"github.com/cloudquery/cloudquery/pkg/plugin"
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
+	"github.com/getsentry/sentry-go"
+	"github.com/spf13/viper"
+)
+
+func CaptureError(err error, tags map[string]string) {
```
disq, line 11: Add `CaptureCommand(err error, commandName string)` to call this one with the `tags` auto filled in with the given command?

pkg/errors/sentry.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		sentry.WithScope(func(scope *sentry.Scope) {
+			if t, ok := d.(core.FetchDiagnostic); ok {
+				if allowUnmanaged && t.Version == plugin.Unmanaged {
+					return
+				}
+				scope.SetTags(map[string]string{"provider": t.Provider, "provider_version": t.Version, "resource": t.Description().Resource})
+			}
+			// set any extra tags to this scope
+			scope.SetTags(tags)
```
disq, line 42: also add `diag_type` as tag

disq:
missing features:
- [ ] `setSentryVars` is removed so CQTeamID-auto-disable is gone (`sentry.ConfigureScope` is not called)
- [ ] `--debug-sentry` exists to allow sending of unmanaged provider events, but it's not used to override CQTeamID-auto-disable (which is gone)
- [ ] `sentry.User` is not set (`scope.SetUser`)
- [ ] `trace_id` is not set to correlate with telemetry events
- [ ] `telemetryOpts` is removed, so no more `--debug-telemetry` and `--inspect-telemetry` handling
- [ ] `diag_type` is not set as tag (to `theDiag.Type().String()`)
- [ ] `fetch.resources.<provider>` and `fetch.<severity>.total` are gone (expected)

pkg/errors/sentry.go
roneli: added

yevgenypats:
@disq what do you mean with trace_id ? I think for the new way of analytics we dont need `trace_id` or is it just additional data as of `execution_id` ?

disq:
> what do you mean with trace_id

yes the analytics trace_id. won't need it so it's scratched out in the list above.

pkg/ui/console/client.go
```diff
@@ -433,6 +437,10 @@ func (c Client) DescribePolicies(ctx context.Context, policySource string) error
 // =====================================================================================================================
 
 func (c Client) CallModule(ctx context.Context, req ModuleCallRequest) error {
+	if _, diags := c.SyncProviders(ctx); diags.HasErrors() {
+		return errors.New(diags.Error())
```
disq, line 441: `return diags`

PR #599, opened 2022-04-26
fix: Trim UTF BOM from telemetry random ID

disq:
~don't merge: only trims, doesn't convert~ hope it works now


roneli (approved):
LGTM

disq:
btw I think this much code to accomodate manual edits (on certain platforms) is unnecessary. Just use iconv if your editor doesn't have the ascii option, or `edit.com`.

PR #605, opened 2022-04-28
feat: Shuffle resource order

disq:


bbernays (approved):

disq:
@roneli I don't think randomize is a job for the SDK as it's a cq-core managed operation (fetching) the question is, do we want this functionality or not? I'm opposed to putting this in SDK (when the sole caller will be cq-core again)

PR #607, opened 2022-04-28
feat: Add policy analytics

roneli:
- add policy analytics
- fix edges of fetch failed with configure
- improve fetch analytics event


pkg/core/fetch_test.go
```diff
@@ -123,17 +123,17 @@ func Test_Fetch(t *testing.T) {
 			ExpectedDiags: []diag.FlatDiag{
 				{
 					Err:         "rpc error: code = DeadlineExceeded desc = context deadline exceeded",
-					Type:        6,
+					Type:        7,
```
disq, line 126: why are these int? at least do `int(diag.ERROR)`?

PR #613, opened 2022-05-01
fix(policy): Add missing GitHub policy getter

erezrokah:
Related to https://github.com/cloudquery/cloudquery-issues/issues/324 (internal issue).

To reproduce the issue run: `cloudquery policy run github::github.com/cloudquery-policies/aws`, and see that it fails with `Failed to run policies: failed to get source github::github.com/cloudquery-policies/aws: download not supported for scheme 'github'.`

### DIFF
--- a/internal/getter/getter.go
+++ b/internal/getter/getter.go
@@ -46,11 +46,12 @@ var (
 	}
 
 	getters = map[string]getter.Getter{
-		"file": new(getter.FileGetter),
-		"gcs":  new(getter.GCSGetter),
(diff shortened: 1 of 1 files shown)


bbernays (approved):
LGTM... only question is do we document somewhere what schemes we support?

erezrokah:
> LGTM... only question is do we document somewhere what schemes we support?

Thanks. I created a separate issue for the docs bit https://github.com/cloudquery/cloudquery-issues/issues/326

PR #615, opened 2022-05-01
chore: Refactor cleanups

roneli:


cmd/drift.go
```diff
@@ -30,6 +32,9 @@ var (
 				Profile:    driftProfile,
 				OutputPath: driftOutputPath,
 			})
+			if err := analytics.Capture("drift", c.Providers, nil, diag.FromError(err, diag.INTERNAL)); err != nil {
+				log.Warn().Err(err).Msg("analytics send failed")
```
disq, line 36: I don't think analytics send failed messages should be in the log unless we specify `--debug-telemetry`. So the logger should be inside the analytics package and this shouldn't return an `error`.

internal/analytics/client.go
```diff
@@ -155,17 +157,22 @@ func Capture(eventType string, providers registry.Providers, data Message, diags
...
 		"buildDate":   c.version.BuildDate,
 		"env":         c.env,
-		"data":        data.Properties(),
 		"instanceId":  c.instanceId,
 		"success":     !diags.HasErrors(),
 		"providers":   providers,
 		"diagnostics": core.SummarizeDiagnostics(diags),
 	}
+
+	if data != nil && !reflect2.IsNil(data) {
```
disq, line 172:
```suggestion
	if !reflect2.IsNil(data) {
```

`reflect2.IsNil` already does the `data != nil` check

pkg/policy/execute.go
```diff
@@ -154,7 +146,7 @@ func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Pol
 		r, err := executor.Execute(ctx, req, p)
 		if err != nil {
 			executor.log.Error("failed to execute policy", "err", err)
-			return nil, fmt.Errorf("%s/%w", policy.Name, err)
+			return nil, diag.FromError(fmt.Errorf("%s/%w", policy.Name, err), diag.DATABASE)
```
disq, line 149:
```suggestion
			return nil, diag.FromError(fmt.Errorf("%s: %w", policy.Name, err), diag.DATABASE)
```

pkg/policy/execute.go
```diff
@@ -166,7 +158,7 @@ func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Pol
 	// TODO: A better idea here is to create a new session, create the views, run queries, and close the session.
 	//       This will remove the need for 'deleteViews'.
 	if err := e.createViews(ctx, policy); err != nil {
-		return nil, err
+		return nil, diag.FromError(fmt.Errorf("%s/%w", policy.Name, err), diag.DATABASE)
```
disq, line 161:
```suggestion
		return nil, diag.FromError(fmt.Errorf("%s: %w", policy.Name, err), diag.DATABASE)
```

pkg/policy/execute.go
```diff
@@ -175,7 +167,7 @@ func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Pol
 		qr, err := e.executeQuery(ctx, q)
 		if err != nil {
 			e.log.Error("failed to execute query", "err", err)
-			return nil, fmt.Errorf("%s/%w", policy.Name, err)
+			return nil, diag.FromError(fmt.Errorf("%s/%w", policy.Name, err), diag.DATABASE)
```
disq, line 170:
```suggestion
			return nil, diag.FromError(fmt.Errorf("%s: %w", policy.Name, err), diag.DATABASE)
```

This `%s/%w` syntax isn't good, let's make it the Go way which is `<related object>: <error about object>`

pkg/policy/run.go
```diff
@@ -89,52 +91,52 @@ type RunRequest struct {
...
+		if dd.HasErrors() {
 			// this error means error in execution and not policy violation
 			// we should exit immediately as this is a non-recoverable error
 			// might mean schema is incorrect, provider version
 			log.Error().Err(err).Msg("policy execution finished with error")
-			return results, err
+			return results, diags.Add(dd)
 		}
 
+		diags = diags.Add(dd)
```
disq, line 117: maybe do diags = diags.Add(dd) on line 109, and do the if dd.HasErrors() afterwards, returning the diags itself if true

pkg/policy/run.go
```diff
@@ -89,52 +91,52 @@ type RunRequest struct {
...
 		results = append(results, result)
 		if req.OutputDir == "" {
 			continue
 		}
 		log.Info().Str("policy", p.Name).Str("version", p.Version()).Str("subPath", p.SubPolicy()).Msg("writing policy to output directory")
 		if err := GenerateExecutionResultFile(result, req.OutputDir); err != nil {
-			return nil, err
+			return nil, diags
```
disq, line 124: `err` not handled here

disq (changes requested):

cmd/drift.go
roneli: fixed

pkg/policy/execute.go
disq: Resolved but not fixed?

pkg/policy/run.go
```diff
@@ -89,52 +91,51 @@ type RunRequest struct {
...
+		diags = diags.Add(dd)
 		log.Info().Msg("policy execution finished")
-		if err != nil {
+		if diags.HasErrors() {
 			// this error means error in execution and not policy violation
 			// we should exit immediately as this is a non-recoverable error
 			// might mean schema is incorrect, provider version
 			log.Error().Err(err).Msg("policy execution finished with error")
-			return results, err
+			return results, diags
```
disq, line 115: maybe move the "policy execution finished" message on line 109 to line 117 so we have a single line for each case (error/no error)

disq:
so `--debug-telemetry` is still not handled? it does log.Error() now which is the global logger no?

roneli:
> so `--debug-telemetry` is still not handled? it does log.Error() now which is the global logger no?

added debug event in case debug is set true

disq (approved):

PR #618, opened 2022-05-02
fix(policy): Use firebase instead of GitHub to get latest version

erezrokah:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/324 (internal issue).

This PR gets the policy latest version from Firebase instead of GitHub to avoid hitting rate limits.
To reproduce run `cloudquery policy run github::github.com/cloudquery-policies/aws` multiple times.

This is the smallest change I could think of to fix the issue.

Example data  https://firestore.googleapis.com/v1/projects/hub-cloudquery/databases/(default)/documents/orgs/cloudquery/policies/aws/versions?mask.fieldPaths=policy.Name

### DIFF
--- a/internal/firebase/client.go
+++ b/internal/firebase/client.go
@@ -24,7 +24,7 @@ type FirebaseClient struct {
 	url string
 }
 
-func NewFirebaseClient(registryUrl string) *FirebaseClient {
+func New(registryUrl string) *FirebaseClient {
 	f := &FirebaseClient{
 		url: registryUrl,
 	}
--- a/internal/firebase/client.go
+++ b/internal/firebase/client.go
@@ -14,9 +14,10 @@ import (
 )
 
 const (
-	CloudQueryRegistryURL = "https://firestore.googleapis.com/v1/projects/hub-cloudquery/databases/(default)/documents/orgs/"
-	providersPath         = "%s/providers/%s/versions"
-	policiesPath          = "%s/policies/%s/versions"
+	CloudQueryRegistryURL              = "https://firestore.googleapis.com/v1/projects/hub-cloudquery/databases/(default)/documents/orgs/"
+	CloudQueryRegistryURLWithProviders = CloudQueryRegistryURL + "%s/providers/%s"
+	providersVersionsPath              = "%s/providers/%s/versions"
+	policiesVersionPath                = "%s/policies/%s/versions"
 )
 
 type FirebaseClient struct {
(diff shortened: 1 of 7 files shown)


internal/firebase/client.go
```diff
@@ -0,0 +1,122 @@
...
+			Name string `json:"name"`
+		} `json:"documents"`
+	}
+
+	if err := json.NewDecoder(res.Body).Decode(&doc); err != nil {
+		return "", err
+	}
+
+	valid := make([]*version.Version, 0)
+	for _, d := range doc.Documents {
```
erezrokah, line 107: This is how our UI does it https://github.com/cloudquery/cq-website/blob/ef3e714533610edfbd50d842d87bd13019dcad91/hub/src/services/firebase.ts#L37 (internal link).

We could make the data easier to consume, but that will probably be a breaking change for our Website

internal/getter/github.go
```diff
@@ -6,7 +6,13 @@ import (
 	"net/url"
 	"strings"
 
-	"github.com/google/go-github/v35/github"
+	"github.com/cloudquery/cloudquery/internal/firebase"
+)
+
+var (
+	repoToFirebasePath = map[string]string{
+		"cloudquery-policies": "cloudquery",
```
erezrokah, line 14: Mapping between the repo name to the Firebase path

pkg/plugin/registry/hub.go
```diff
@@ -259,45 +256,9 @@ func (h Hub) downloadProvider(ctx context.Context, organization, providerName, p
 }
 
 func (h Hub) getLatestRelease(ctx context.Context, organization, providerName string) (string, error) {
-	versions, err := url.Parse(fmt.Sprintf(h.url+"/versions", organization, providerName))
```
erezrokah, line 262: Extracted to a new Firebase struct

erezrokah:
Looking at the failed tests

internal/getter/github_test.go
```diff
@@ -14,13 +14,6 @@ func TestGitHubDetector_Detect(t *testing.T) {
 		ExpectedFound  bool
 		ExpectedError  error
 	}{
-		{
```
erezrokah, line 17: The `test_policy` is not in our Hub

internal/firebase/client.go
```diff
@@ -0,0 +1,123 @@
...
+	CloudQueryRegistryURLWithProviders = CloudQueryRegistryURL + "%s/providers/%s"
+	providersVersionsPath              = "%s/providers/%s/versions"
+	policiesVersionPath                = "%s/policies/%s/versions"
+)
+
+type FirebaseClient struct {
+	url string
+}
+
+func NewFirebaseClient(registryUrl string) *FirebaseClient {
```
disq, line 27:
```suggestion
func New(registryUrl string) *Client {
```

disq (approved):
One nit

internal/firebase/client.go
erezrokah: Thanks for the review! Good point, done in https://github.com/cloudquery/cloudquery/pull/618/commits/1c8f7762d50bcbd1e3e6b291b7e1fd298189e8b9

disq: Missed the type rename though 😂

erezrokah: Ah, good catch. Done in https://github.com/cloudquery/cloudquery/pull/622

PR #623, opened 2022-05-02
chore(refactor): Refactor core code

roneli:


cmd/command.go
```diff
@@ -97,24 +68,22 @@ func handleConsole(ctx context.Context, tele *telemetry.Client, cmd *cobra.Comma
...
-		c, err = console.CreateClient(ctx, cfgPath, cfgMutator)
+		c, err = console.CreateClient(ctx, cfgPath, false, cfgMutator)
 		if err != nil {
 			return err
 		}
-		defer c.Client().Close()
 	}
 
-	if tele.Enabled() && tele.NewRandomId() {
+	if analytics.Enabled() {
```
disq, line 86: There's a functionality disparity here: we only do this notice/delay thing if we had created a "new" ID, so it was only shown once (unless the filesystem isn't persistent)

New code shows/delays every time

cmd/provider.go
```diff
@@ -42,7 +46,12 @@ var (
 		Short: providerUpgradeHelpMsg,
 		Long:  providerUpgradeHelpMsg,
 		Run: handleCommand(func(ctx context.Context, c *console.Client, cmd *cobra.Command, args []string) error {
-			return c.UpgradeProviders(ctx, args)
+			_, diags := c.SyncProviders(ctx, args...)
+			errors.CaptureDiagnostics(diags, nil)
```
disq, line 50: Missing command in the map

cmd/provider.go
```diff
@@ -52,7 +61,12 @@ var (
 		Short: providerDowngradeHelpMsg,
 		Long:  providerDowngradeHelpMsg,
 		Run: handleCommand(func(ctx context.Context, c *console.Client, cmd *cobra.Command, args []string) error {
-			return c.DowngradeProviders(ctx, args)
+			_, diags := c.SyncProviders(ctx, args...)
+			errors.CaptureDiagnostics(diags, nil)
```
disq, line 65: Missing command in the map

cmd/command.go
roneli: I think it should show every time to be honest even if Id is already created.

disq: but it also delays for 2 seconds.

roneli: we can live with that for now ?

PR #641, opened 2022-05-03
fix: Encode json

### LINKED ISSUE #639: Console JSON Logging not possible
### Describe the Bug

`--encode-json` parameter is not respected

### Expected Behavior

`cloudquery fetch --enable-console-log --encode-json` should output all of the logs that went to `cloudquery.log` to stdout in a json format.

### CloudQuery Version

cloudquery version 0.22.3

roneli:
Closes #639

### DIFF
--- a/internal/logging/logger.go
+++ b/internal/logging/logger.go
@@ -25,8 +25,6 @@ type Config struct {
 	Verbose bool `hcl:"verbose,optional"`
 	// EncodeLogsAsJson makes the logging framework logging JSON
 	EncodeLogsAsJson bool `hcl:"encode_logs_as_json,optional"`
-	// EncodeConsoleAsJson makes the logging framework logging JSON
-	EncodeConsoleAsJson bool `hcl:"encode_logs_as_json,optional"`
 	// FileLoggingEnabled makes the framework logging to a file
 	// the fields below can be skipped if this value is false!
 	FileLoggingEnabled bool `hcl:"file_logging_enabled,optional"`


pkg/ui/output.go
```diff
@@ -14,20 +15,26 @@ import (
...
+var emojiStatus = []string{color.GreenString("✓"), "📋", color.RedString("❌"), "⚠️", "⌛"}
+
 // ColorizedOutput outputs a colored message directly to the terminal.
 // The remaining arguments should be interpolations for the format string.
 func ColorizedOutput(c *color.Color, msg string, values ...interface{}) {
 	if viper.GetBool("enable-console-log") {
 		// Print output to log
-		if logMsg := strings.ReplaceAll(msg, "\n", ""); logMsg != "" {
-			log.Info().Msgf(logMsg, values...)
+		if logMsg := strip(fmt.Sprintf(msg, values...)); logMsg != "" {
```
erezrokah, line 26: [non blocking] Can you explain the reason we need to strip emojis in the console (maybe good to comment it)?

Also, we could consider using https://github.com/acarl005/stripansi/blob/5a71ef0e047df0427e87a79f27009029921f1f9b/stripansi.go#L7 or https://github.com/chalk/ansi-regex/blob/02fa893d619d3da85411acc8fd4e2eea0e95a9d9/index.js#L3

erezrokah (approved):
Looks good @roneli. Left a non blocking commnet

pkg/ui/output.go
roneli: I tried them they didn't work cleanly for some reason

internal/logging/logger.go
```diff
@@ -25,6 +25,8 @@ type Config struct {
 	Verbose bool `hcl:"verbose,optional"`
 	// EncodeLogsAsJson makes the logging framework logging JSON
 	EncodeLogsAsJson bool `hcl:"encode_logs_as_json,optional"`
+	// EncodeConsoleAsJson makes the logging framework logging JSON
```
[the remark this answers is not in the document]
roneli: removed

Issue #647, opened 2022-05-03
CloudQuery Policy Selector


bbernays:
### Describe the Bug
CloudQuery does not respect a selector when the same policy is also defined in the config.hcl

config.hcl:
``` 
policy "aws" {
    source = "aws"
}
```

`cloudquery policy run aws//cis_v1.2.0`

This will run the entire AWS policy instead of just the cis_v1.2.0

### Expected Behavior

When a selector is passed it should override any policies defined in config.hcl

bbernays:
@roneli - I think this was fixed and merged correct?

roneli:
Correct closing this issue.

PR #649, opened 2022-05-04
fix(policy): Don't crash when setting policy via config.hcl

erezrokah:
I encountered this while trying to reproduce https://github.com/cloudquery/cloudquery/issues/647

To reproduce:
1. Add the following in your `config.hcl` file (as a top level object)
```hcl
policy "aws" {
    source = "aws"
}
```
2. On the `main` branch run `go run main.go policy run`

### DIFF
--- a/pkg/policy/policy.go
+++ b/pkg/policy/policy.go
@@ -19,7 +19,7 @@ func (pp Policies) Properties() map[string]interface{} {
 	usedCustom := 0
 	policies := make([]*Meta, 0)
 	for _, p := range pp {
-		if p.meta.Type != "hub" {
+		if p.meta != nil && p.meta.Type != "hub" {
 			usedCustom++
 			// we don't add info about custom policies that were executed
 			continue


.github/workflows/sanity.yml
```diff
@@ -93,6 +93,16 @@ jobs:
         env:
           CQ_NO_TELEMETRY: 1
 
+      - name: Policy Run from config.hcl configuration
+        run: |
+          go run ./main.go policy run --config=internal/test/test_policy.hcl --disable-fetch-check &> policy_run.txt
```
erezrokah, line 98: `policy run` fails if the policy tests failed, so this is the best way I could come up with to verify it didn't panic

erezrokah:
I removed the test to unblock the PR. I'll see why it's still failing on GitHub action later on

pkg/policy/policy.go
```diff
@@ -19,7 +19,7 @@ func (pp Policies) Properties() map[string]interface{} {
 	usedCustom := 0
 	policies := make([]*Meta, 0)
 	for _, p := range pp {
-		if p.meta.Type != "hub" {
+		if p.meta != nil && p.meta.Type != "hub" {
```
roneli, line 22: should be if p.meta == nil, but actually if you set a config policy we should need info to read where its from. Will debug locally.

roneli:
Closes for #651

PR #652, opened 2022-05-05
feat: Track db installations

disq:
TODO: call `executor.Identifier()` to get the identifier and inject it into telemetry.

First we should fix the call to `executor.Validate()` though. (#653 fixes this)


roneli (approved):
LGTM, would add it to analytics in a separate PR

roneli (commented):
maybe write a test, we have e2e tests to a PG, that just checks we get an identifier and call it twice perhaps?

PR #653, opened 2022-05-05
fix: Validate db version before proceeding

disq:
This was broken in the refactor.

### DIFF
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -110,6 +110,15 @@ func CreateClientFromConfig(ctx context.Context, cfg *config.Config) (*Client, e
 		if err != nil {
 			return nil, err
 		}
+		if ok, err := dialect.Validate(ctx); err != nil {
+			if !ok {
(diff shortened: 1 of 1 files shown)


pkg/ui/console/client.go
```diff
@@ -110,6 +110,15 @@ func CreateClientFromConfig(ctx context.Context, cfg *config.Config) (*Client, e
 		if err != nil {
 			return nil, err
 		}
+		if ok, err := dialect.Validate(ctx); err != nil {
+			if !ok {
+				return nil, fmt.Errorf("validate: %w", err)
+			}
+			log.Warn().Err(err).Msg("database validation warning")
```
bbernays, line 117: Do we want a better message?

disq: the `err` is either:
`unsupported PostgreSQL version: %s. (should be >= %s)`
or
`error getting PostgreSQL version: %w`

bbernays (approved):
LGTM

PR #670, opened 2022-05-09
feat: DSN credentials

### LINKED ISSUE #600: DSN Parsing
### Describe the Bug
`postgres://cloudquery:yMLh4SV0j#gPP>T@cluster-id.us-east-2.rds.amazonaws.com:5432/postgres`

This is a valid password for Postgres. CQ Parsing fails to recognize the characters in the password and outputs the following error:

```
parse "DSN redacted": invalid port ":yMLh4SV0j" after host
```

### Expected Behavior

Should parse DSN cleanly

### CloudQuery Version
```
Version: 0.22.3
Commit: d9170a5140a2a8b0bc151fc15502607be103e070
Date: 2022-04-12T09:17:49Z
```

### LINKED ISSUE #612: DSN in config should be optional
### Describe the Bug

`dsn` in config.hcl should be optional (as long as `--dsn` is specified) but it doesn't seem to be, if `connection` block is specified.

### Expected Behavior

`config.hcl` without `dsn` should work as long as DSN is specified with `--dsn` cli param.

### CloudQuery Version

`v0.22.10`

### Steps to Reproduce

- Remove DSN from config, but keep an empty `connection` block.
- Run cq fetch with --dsn param.

disq:
Introduce all DSN parts separately in connection block.
Old DSN syntax (URL or non-URL) can still be used, but the two types can't be combined.

Closes #600.
Fixes #612.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -73,18 +73,17 @@ func Initialize(ctx context.Context, providers []string) error {
 		providers[i] = providerName // overwrite "provider@version" with just "provider"
 	}
 	// TODO: build this manually with block and add comments as well
-	defaultPort := 5432
 	cqBlock := gohcl.EncodeAsBlock(&config.CloudQuery{
 		PluginDirectory: "./cq/providers",
 		PolicyDirectory: "./cq/policies",
 		Providers:       requiredProviders,
 		Connection: &config.Connection{
-			Username: ptr("postgres"),
-			Password: ptr("pass"),
-			Host:     ptr("localhost"),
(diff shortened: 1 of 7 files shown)


bbernays:
What if we were to expose all of the required DSN attributes (`host`, `database`, `port`, and `sslmode`) in our config not just the 3 we have exposed here? Then users don't have to dynamically generate any strings they only have to specify single attributes.

bbernays (approved):
LGTM

PR #684, opened 2022-05-10
fix(policy): Improve file detector to detect non existing files

erezrokah:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/333 (internal issue).

Underlying issue - our file detector only detects existing files:
https://github.com/cloudquery/cloudquery/blob/38004809f7a0e4ae8a7ef86403524f0af3037eff/internal/getter/file.go#L24

This causes the code to return `found=false` here https://github.com/cloudquery/cloudquery/blob/38004809f7a0e4ae8a7ef86403524f0af3037eff/pkg/ui/console/policy.go#L43

And throw the error. The error itself is not that good either as it reports that no policies were found in the configuration.

This PR fixes the issue by improving the file detector to detect non existing files.
We need the `allowNonExisting` configuration in order not to break the `hub` detector logic.

### Todo
- [ ] Add a test

### DIFF
--- a/internal/getter/file.go
+++ b/internal/getter/file.go
@@ -39,7 +39,7 @@ func fmtFileURL(path string) string {
 	}
 
 	// Make sure that we don't start with "/" since we add that below.
-	if path[0] == '/' {
+	if len(path) > 0 && path[0] == '/' {
 		path = path[1:]
 	}
 	return fmt.Sprintf("file:///%s", path)
--- a/internal/getter/file.go
+++ b/internal/getter/file.go
@@ -1,7 +1,9 @@
 package getter
 
 import (
+	"fmt"
 	"path/filepath"
+	"runtime"
 
 	"github.com/hashicorp/go-getter"
 	"github.com/spf13/afero"
@@ -9,7 +11,38 @@ import (
 
 // fileDetector is a replacement for go-getter's own file detector which also verifies the source file exists
 type fileDetector struct {
-	fd getter.FileDetector
+	fd               getter.FileDetector
+	allowNonExisting bool
+}
+
+func isValidPath(fs afero.Fs, path string) bool {
+	err := fs.MkdirAll(path, 0700)
+	if err == nil {
+		_ = fs.Remove(path)
+	}
+
+	return err == nil
+}
+
+func NewFileDetector(allowNonExisting bool) *fileDetector {
+	p := new(fileDetector)
+	p.allowNonExisting = allowNonExisting
+	return p
+}
+
+// Extracted from getter.FileDetector
+func fmtFileURL(path string) string {
+	if runtime.GOOS == "windows" {
+		// Make sure we're using "/" on Windows. URLs are "/"-based.
+		path = filepath.ToSlash(path)
+		return fmt.Sprintf("file://%s", path)
+	}
+
+	// Make sure that we don't start with "/" since we add that below.
+	if path[0] == '/' {
+		path = path[1:]
+	}
+	return fmt.Sprintf("file:///%s", path)
 }
 
 func (d *fileDetector) Detect(src, pwd string) (string, bool, error) {
@@ -20,7 +53,11 @@ func (d *fileDetector) Detect(src, pwd string) (string, bool, error) {
 	if pwd != "" && !filepath.IsAbs(src) {
 		checkPath = filepath.Join(pwd, src)
 	}
-	if ok, _ := afero.Exists(afero.NewOsFs(), checkPath); !ok {
+	fs := afero.NewOsFs()
+	if ok, _ := afero.Exists(fs, checkPath); !ok {
+		if d.allowNonExisting && isValidPath(fs, checkPath) {
+			return fmtFileURL(checkPath), true, nil
+		}
 		return "", false, nil
 	}
 	return d.fd.Detect(src, pwd)
--- a/internal/getter/file.go
+++ b/internal/getter/file.go
@@ -15,15 +15,6 @@ type fileDetector struct {
 	allowNonExisting bool
 }
 
-func isValidPath(fs afero.Fs, path string) bool {
-	err := fs.MkdirAll(path, 0700)
-	if err == nil {
-		_ = fs.Remove(path)
-	}
-
-	return err == nil
-}
-
 func NewFileDetector(allowNonExisting bool) *fileDetector {
 	p := new(fileDetector)
 	p.allowNonExisting = allowNonExisting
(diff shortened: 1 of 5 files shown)


pkg/ui/console/client.go
```diff
@@ -375,7 +376,7 @@ func (c Client) RunPolicies(ctx context.Context, policySource, outputDir string,
 		policyRunProgress.MarkAllDone()
 		policyRunProgress.Wait()
 	}
-	if !noResults {
+	if !noResults && resp != nil {
```
erezrokah, line 379: Unrelated bug fix

internal/getter/file.go
```diff
@@ -1,15 +1,48 @@
...
 type fileDetector struct {
-	fd getter.FileDetector
+	fd               getter.FileDetector
+	allowNonExisting bool
+}
+
+func isValidPath(fs afero.Fs, path string) bool {
+	err := fs.MkdirAll(path, 0700)
+	if err == nil {
+		fs.Remove(path)
```
erezrokah, line 21: This deletes only the top level directory. We could recurse and delete all empty ones going up, but I'm not sure I'm happy doing that

internal/getter/hub.go
```diff
@@ -14,7 +14,7 @@ func (h HubDetector) Detect(src, pwd string) (string, bool, error) {
 	if len(src) == 0 {
 		return "", false, nil
 	}
-	fileDetector := fileDetector{}
+	fileDetector := NewFileDetector(false)
```
erezrokah, line 17: We pass `false` here otherwise the file detector will detect `aws` as a file (for example), so we let the hub detector run in that case.

erezrokah:
Blocked by https://github.com/cloudquery/cloudquery/pull/688 and https://github.com/cloudquery/cloudquery/pull/690

internal/getter/hub.go
roneli:
```
policy aws {
  source = "./some/path/to/policy/aws"
}
```
Would this work?

erezrokah: Let me verify that use case.

The logic for `hub.go` doesn't change (only for the file detector), but I'll verify this works.

internal/getter/file.go
```diff
@@ -1,15 +1,48 @@
...
 
 // fileDetector is a replacement for go-getter's own file detector which also verifies the source file exists
 type fileDetector struct {
-	fd getter.FileDetector
+	fd               getter.FileDetector
+	allowNonExisting bool
+}
+
+func isValidPath(fs afero.Fs, path string) bool {
+	err := fs.MkdirAll(path, 0700)
```
disq, line 19: This doesn't check for validity but for write access also. Why not fs.Stat first, if `os.ErrNotExist` just use methods from `filepath` to make sure it's a valid syntax? (or just check syntax and don't do any stat)

Creating just to remove doesn't shouldn't even be an option.

disq (changes requested):

internal/getter/file.go
```diff
@@ -1,15 +1,48 @@
...
+// Extracted from getter.FileDetector
+func fmtFileURL(path string) string {
+	if runtime.GOOS == "windows" {
+		// Make sure we're using "/" on Windows. URLs are "/"-based.
+		path = filepath.ToSlash(path)
+		return fmt.Sprintf("file://%s", path)
+	}
+
+	// Make sure that we don't start with "/" since we add that below.
+	if path[0] == '/' {
```
disq, line 42: This will panic if path is an empty string.

internal/getter/file.go
erezrokah: We do an `Exists` check before https://github.com/cloudquery/cloudquery/blob/8520eae35d47a89a86406a8e36831f7903edec52/internal/getter/file.go#L23 so I'm not sure we need another one here.

Checking for valid syntax is not straightforward as it's highly dependent on OS flavor and version (e.g. OSX, Windows, Linux and even different version within that OS).

We could use `WriteFile` maybe but that will fail if the directory structure is missing (however maybe we can detect that from the error).

More in https://stackoverflow.com/questions/35231846/golang-check-if-string-is-valid-path

internal/getter/file.go
erezrokah: Good catch. I'll add a check for that

pkg/policy/source.go
```diff
@@ -34,6 +35,13 @@ func DetectPolicy(name string, subPolicy string) (*Policy, bool, error) {
 	}, true, nil
 }
 
+func classifyError(detectorType string, err error) error {
+	if (detectorType == "file") && strings.Contains(err.Error(), "no such file or directory") {
```
disq, line 39: use `errors.Is(err, fs.ErrNotExist)`

disq:
detecting non existing is a confusing concept. why not just check if that file exists (or if the error is ErrNotExist) and move on? don't touch the detector etc.

pkg/policy/source.go
erezrokah: Ah, I tried it and it doesn't work as the original error has been formatted multiple times once it gets to this line.
I tried unwrapping it to get the original, but that doesn't work as well, since `go-getter` doesn't use the `%w` operand.

erezrokah:
> detecting non existing is a confusing concept. why not just check if that file exists (or if the error is ErrNotExist) and move on? don't touch the detector etc.

We already check if the file exists and that works well. The issue is that when the file doesn't exists, it won't match any getter and the following error will be printed:

```
ERR no policy with name ./non-existing-dir found in configuration or remote. Available in config: []
```

Which is super confusing.

I think that what we want is:
1. Detect that the user intended to use a file (even if non existing).
2. Error out if the file doesn't exist when trying to download it

erezrokah:
> or if the error is ErrNotExist

We could also have the file detector return always true, assuming it runs last and skip the path check if that is what you meant

roneli:
> > detecting non existing is a confusing concept. why not just check if that file exists (or if the error is ErrNotExist) and move on? don't touch the detector etc.
> 
> We already check if the file exists and that works well. The issue is that when the file doesn't exists, it won't match any getter and the following error will be printed:
> 
> ```
> ERR no policy with name ./non-existing-dir found in configuration or remote. Available in config: []
> ```
> 
> Which is super confusing.
> 
> I think that what we want is:
> 
> 1. Detect that the user intended to use a file (even if non existing).
> 2. Error out if the file doesn't exist when trying to download it
> 
> Please let me know if that makes sense

What did you run here? `cloudquery policy run ./non-existing-policy` ?

If so the error makes sense, unless you want to make custom errors per detector, which is not true, its the getter error that should return the correct error as with detectors "false" i.e not found is okay, its only when we use the "getter" to actually load from the source an error of not exists should occur.

The follow is as following:

- by given run -> check if policy is defined in configuration -> use that source, if not use the given path as source
- try best to detect what getter to use for that source, unless source is forced i.e file:// or github://
- use the getter to get the source, return correct error to user.

Because `policy run aws` can also use the config which can have a different source.

The internal issue doesn't have enough context to why it happened.

erezrokah:
Going to close this in favor of https://github.com/cloudquery/cloudquery/pull/697.
We decided it's best not to change the logic and only focus on improving the error to the user

PR #688, opened 2022-05-10
fix: Handle `nil` policy run response

erezrokah:
Extracted from https://github.com/cloudquery/cloudquery/pull/684 to a dedicated PR


roneli (approved):
LGTM

PR #690, opened 2022-05-10
fix: Run policy detectors in order

erezrokah:
Extracted from https://github.com/cloudquery/cloudquery/pull/684.

Specifically https://github.com/cloudquery/cloudquery/pull/684#discussion_r869393112.

Using an array to iterate over the detectors (instead of a map) ensures we run them in order and provide a deterministic result


disq (approved):
LGTM

PR #697, opened 2022-05-11
fix: Classify not found policies and improve errors

erezrokah:
Replaces https://github.com/cloudquery/cloudquery/pull/684

### DIFF
--- a/pkg/ui/console/policy.go
+++ b/pkg/ui/console/policy.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cloudquery/cloudquery/internal/getter"
 	"github.com/cloudquery/cloudquery/pkg/policy"
 	"github.com/cloudquery/cloudquery/pkg/ui"
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
 
 	"github.com/fatih/color"
 	"github.com/olekukonko/tablewriter"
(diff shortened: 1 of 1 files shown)


disq (commented):
I would use `%q` when referring to user inputs.

pkg/ui/console/policy.go
```diff
@@ -43,15 +44,22 @@ func FilterPolicies(policyPath string, policies policy.Policies) (policy.Policie
...
 		return nil, err
 	}
 	if found {
 		return policy.Policies{p}, nil
 	}
-	return nil, fmt.Errorf("no policy with name %s found in configuration or remote. Available in config: %s", policyName, policies.All())
+
+	configPolicies := policies.All()
+	if len(configPolicies) == 0 {
+		return nil, diag.FromError(fmt.Errorf("no valid policy with name %s found. If using a local policy directory, ensure the path is correct and the directory exists", policyName), diag.USER)
```
disq, line 58: I would use `%q` when referring to user inputs. Line 61 also.

erezrokah: Thanks for the tip! Done in https://github.com/cloudquery/cloudquery/pull/697/commits/7353ee0fc13cc7165471437df302f0fff230927a (I didn't wrap the array as it wraps each individual element and doesn't look that nice).

disq: ty, array isn't user input anyway.

disq (approved):

PR #701, opened 2022-05-11
fix: Classify subdir not found error

erezrokah:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/334 (internal issue).

I could not reproduce this (probably needs a Windows machine). However the error is descriptive enough for the user to take action, so we should probably just filter it from our analytics.

The reported error is:
```
*errors.errorString: subdir "<redacted>" not found
*fmt.wrapError: failed to get source file://C://<redacted>: subdir "<redacted>" not found
```

### DIFF
--- a/pkg/policy/source.go
+++ b/pkg/policy/source.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/cloudquery/cloudquery/internal/getter"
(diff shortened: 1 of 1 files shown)


pkg/policy/source.go
```diff
@@ -37,6 +38,16 @@ func DetectPolicy(name string, subPolicy string) (*Policy, bool, error) {
 	}, true, nil
 }
 
+func classifyError(source string, err error) error {
+	matched, _ := regexp.MatchString("subdir .+? not found", err.Error())
+	formattedError := fmt.Errorf("failed to  %s: %w", source, err)
```
disq, line 43: double space between `to` and `%s`

erezrokah: Thanks! The error should not change actually. Fixed in https://github.com/cloudquery/cloudquery/pull/701/commits/60d5f565e823f3fb8a1a22e8f337f215680ed2f1

disq (approved):

PR #705, opened 2022-05-12
feat: Use database id as unique id

roneli:


cmd/sentry.go
```diff
@@ -100,9 +100,10 @@ func initSentry() {
 			ID: userId.String(),
 		})
 		scope.SetTags(map[string]string{
-			"terminal": strconv.FormatBool(ui.IsTerminal()),
-			"ci":       strconv.FormatBool(analytics.IsCI()),
-			"faas":     strconv.FormatBool(analytics.IsFaaS()),
+			"cookie_id": userId.String(),
```
disq, line 103: This would lead to high cardinality which is not good for tags.

roneli: Maybe ill add it to extras then.

internal/analytics/client.go
```diff
@@ -196,13 +199,17 @@ func Capture(eventType string, providers registry.Providers, data Message, diags
...
+	event := analytics.Track{UserId: c.userId, Event: eventType, Timestamp: time.Now().UTC(), Properties: eventProps}
 	if err := c.client.Enqueue(event); err != nil {
 		if c.debug {
 			log.Error().Err(err).Msg("failed to send analytics")
 		}
 	}
 }
+func SetUserId(userId string) {
+	c := currentHub
+	c.userId = userId
```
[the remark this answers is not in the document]
roneli: I am not sure I 100% follow, analytics don't have scopes, or do you mean do the change in one place with ConfigureScope of sentry?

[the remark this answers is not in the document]
disq: Sorry I thought this was related to Sentry but it's not

internal/analytics/client.go
```diff
@@ -106,7 +107,8 @@ func Init(opts ...Option) error {
 func New(opts ...Option) *Client {
 	c := &Client{
 		version:    VersionInfo{},
-		userId:     GetUserId(),
+		userId:     GetUserId().String(),
+		cookieId:   GetUserId(),
```
disq, line 111: since we're using "User ID" as cookie id and eventually db id as user id, I would rename `GetUserId` to be `GetCookieId`

roneli: Although it does become a user id if no database id is set (for example pg < 10)

pkg/ui/console/client.go
```diff
@@ -778,6 +779,15 @@ func setAnalyticsProperties(props map[string]interface{}) {
 	})
 }
 
+func setUserId(databaseId string) {
```
disq, line 782:
```suggestion
func setUserId(newId string) {
```
maybe? calling it databaseId leaks upstream implementation details into this fn, which innocently sets the id and moves on 😂

disq (approved):
there's a nit but LGTM

Issue #687, opened 2022-05-10
Support `NotResource` Attribute for provider


bbernays:
### Is your feature request related to a problem? Please describe.
Right now a user can either hard code a list of all resources they want to fetch or use the `*` to get all resources. There is no way to explicitly deny a set of resources

### Describe the solution you'd like

Add an HCL attribute `not_resource`  which takes a list of resources that are not fetched

disq:
We have `ignore_resources` in drift side so maybe a similar keyword? But the syntax is different for drift (it's `resource_type:<tags, identifier or *>`) so this one wouldn't be the same syntax.

bbernays:
maybe `skip_resources`? I am thinking that it takes an array like `resources`

PR #706, opened 2022-05-12
feat: Resource list enhancements

disq:
Implements #687 and https://github.com/cloudquery/cq-provider-aws/issues/879

~Maybe don't use "unlimited" globbing but allow only for class-level wildcards? ie. don't allow `s3.bu*` but only `s3.*`~


bbernays:
I agree that we should only support blobs at the service level

pkg/core/fetch_test.go
```diff
@@ -261,56 +260,106 @@ func Test_doNormalizeResources(t *testing.T) {
...
+			true,
+		},
+		{
+			"invalid glob 2",
+			[]string{"c1.*res1"},
+			nil,
+			map[string]*schema.Table{"c1.res1": nil, "c1.res2": nil, "c2.res3": nil, "c2.res4": nil},
+			nil,
+			true,
+		},
```
bbernays, line 347: The only other test I would add is a test to ensure `c1.res*` will fail

bbernays (approved):
LGTM other than the single comment

pkg/core/fetch.go
```diff
@@ -418,44 +420,85 @@ func executeFetch(ctx context.Context, pLog zerolog.Logger, plugin plugin.Plugin
...
-// It replaces wildcard resource with all resources. Error is returned if:
+// doGlobResources returns a canonical list of resources given a list of requested and all known resources.
+// It replaces wildcard resource with all resources in non-wild mode. Error is returned if:
 //
 // * wildcard is present and other explicit resource is requested;
 // * one of explicitly requested resources is not present in all known;
 // * some resource is specified more than once (duplicate).
-func doNormalizeResources(requested []string, all map[string]*schema.Table) ([]string, diag.Diagnostics) {
-	if len(requested) == 1 && requested[0] == "*" {
+func doGlobResources(requested []string, allowWild bool, all map[string]*schema.Table) ([]string, diag.Diagnostics) {
```
roneli, line 445: nit: I think this function has multiple jobs, its better to have small functions rather then one big one.

- one to handle "*" 
- one to handle skip
- one to open glob patterns

roneli (commented):
Looks great had a nit on the function as I think the functions should be small and with 1 job rather then do many jobs on the resource list, other than that the logic looks great.

pkg/core/fetch.go
disq: skip is handled already in the same fn that's the point of normalize calling glob.
the glob pattern here is special and very basic so I don't think a new fn is needed just to wrap strings.HasSuffix and few exceptions.

disq: Updated

roneli (approved):
LGTM

PR #710, opened 2022-05-13
fix: Remove lambda support

yevgenypats:
Lambda is limited by 15min and thus is not a good fit for CQ.

Currently it is advised to deploy CQ on k8s via helm

https://github.com/cloudquery/helm-charts
https://github.com/cloudquery/terraform-aws-cloudquery


bbernays (approved):

roneli (approved):
LGTM. Yes lets get rid of this, it is also hard to maintain this here, any lambda CQ runner should be external / community

disq (approved):

PR #711, opened 2022-05-14
feat: Change HCL to YAML

yevgenypats:
This is still WIP.

Context:

HCL is a powerful language but brings a lot of complexity and interoperability issues between languages and stacks.
    
CloudQuery doesn't use any of the powerful HCL features and mainly uses it for very simple configs and HCL parser and configuration introduce code burden, development and room for bugs.
    
Moreover YAML and JSONs have good ecosystems of schema validators which will remove the need for code validation and will also help streamline providers configuration (schema for example is available in the PR).

Next steps:

Discuss what is needed to make the transition and if don't miss anything where we have to have HCL.


yevgenypats:
cc @roneli @bbernays @disq for discussion next week. once discussed and we have a plan for transition I can continue the PR and make it workable.

bbernays:
I am all for this change, the biggest caveat I would say is that we should not remove HCL in the same release we add yaml. On top of that I think we should give customers a conversion tool to make the transition easy

yevgenypats:
I think it won't be possible to create a good tool to convert from HCL to yaml in a reasonable amount of time so I think we just have to create a blog explaining why we did it and tutorial on how to convert and so on. 

The reasoning behind it is: It is better to say: we broke backward compatibility: here is semi-automated/manual work  you will need to do to migrate, instead of here is a buggy software that will help you migrate and actually will take more time and frustration to the user then the manual but working way imo.

I think it this stage we will break backward compatibility from time to time (no way around it or development will stall completely).

bbernays:
If we are making this change in order to reduce overall complexity, why can't we leave the HCL parser in (for a release or two)? We make overall very few config file schema changes that the added complexity (in the short term) wouldn't be too bad I think. Not only is this a better experience for users it also gives us the ability to slowly roll it out before making it the default and only option

erezrokah:
👍  For keeping both for a few releases. At least until we update the docs. We could print a warning message if someone is using the old format saying it's going to be deprecated and link to the blog post/docs. Extra bonus points for a command to convert (found https://github.com/sclevine/yj and https://www.hcl2json.com/ but not sure how well they suite our use case).

disq:
PR also removes history related things for some reason

disq:
Closing in favor of #887

PR #714, opened 2022-05-15
feat: Add global log id

### LINKED ISSUE #671: Log Global Execution ID
When digging through log file it is annoying/hard/impossible to understand where a command ends and a new one starts. Also if a user uses the same log file with multiple simultaneous executions there is no way of identifying which logs go together.

When the global logger is instantiated we should inject som type of uuid that is unique to that command execution and should be included in all log messages. Ideally this would be able to be correlated with the fetch id stored in the DB

roneli:
Closes #671

### DIFF
--- a/pkg/core/fetch.go
+++ b/pkg/core/fetch.go
@@ -188,15 +188,21 @@ type FetchOptions struct {
 	ExtraFields map[string]interface{}
 	// Optional: whether fetch is executed in history mode or not
 	History *history.Config
+	// Optional: unique identifier for the fetch, if this isn't given, a random one is generated.
+	FetchId uuid.UUID
 }
 
 func Fetch(ctx context.Context, storage database.Storage, pm *plugin.Manager, opts *FetchOptions) (res *FetchResponse, diagnostics diag.Diagnostics) {
-	fetchId, err := uuid.NewUUID()
-	if err != nil {
-		return nil, diag.FromError(err, diag.INTERNAL)
+	var err error
+	fetchId := opts.FetchId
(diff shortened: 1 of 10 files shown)


erezrokah (approved):

bbernays (commented):
This looks good. By looking only at the logs is it possible to correlate a fetch_id with this global id?

roneli:
> This looks good. By looking only at the logs is it possible to correlate a fetch_id with this global id?

Yes that is possible, I can make the fetch_id be the global id, ill need to pass it via the request, will have it committed shortly

bbernays (approved):

pkg/core/fetch.go
```diff
@@ -188,15 +188,21 @@ type FetchOptions struct {
...
+	fetchId := opts.FetchId
+	if fetchId == uuid.Nil {
+		fetchId, err = uuid.NewUUID()
+		if err != nil {
+			return nil, diag.FromError(err, diag.INTERNAL)
+		}
 	}
 	// set metadata we want to pass to
-	metadata := map[string]interface{}{"cq_fetch_id": fetchId.String()}
+	metadata := map[string]interface{}{"cq_fetch_id": fetchId}
```
disq, line 205: booo `fetchId.String()`

bbernays: Why is this an issue?

disq: because https://github.com/cloudquery/cq-provider-sdk/blob/66f762ee64ed563ae6646d337493b24b2abc5c7a/provider/schema/meta.go#L35

PR #729, opened 2022-05-17
fix: Don't generate telemetry-random-id in current dir

erezrokah:

#### Summary

Fixes an issue that on each run of `cloudquery` a telemetry id file is created in the current dir

---

- [x] Update or add tests 🧪

### DIFF
--- a/internal/analytics/client.go
+++ b/internal/analytics/client.go
@@ -29,7 +29,7 @@ type VersionInfo struct {
 }
 
 // Consumers must call analytics.Init before using this package
-var currentHub *Client = nil
+var currentHub *Client
 
 type Client struct {
 	version    VersionInfo
--- a/internal/analytics/client.go
+++ b/internal/analytics/client.go
@@ -28,8 +28,8 @@ type VersionInfo struct {
 	CommitId  string `json:"commit_id,omitempty"`
 }
 
-// currentHub is the initial Hub with no Client bound and an empty Scope.
-var currentHub = New()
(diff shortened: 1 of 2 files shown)


internal/analytics/client.go
```diff
@@ -28,8 +28,8 @@ type VersionInfo struct {
 	CommitId  string `json:"commit_id,omitempty"`
 }
 
-// currentHub is the initial Hub with no Client bound and an empty Scope.
-var currentHub = New()
+// Consumers must call analytics.Init before using this package
+var currentHub *Client = nil
```
erezrokah, line 32: Not sure this is the right fix.
We could also call `Init()` before each method call in this package if `currentHub = nil` (and maybe add a lock to make it safe)

roneli: Ye that is fine, you can actually do in initAnalytics in cmd package to set the default hub or even wrap it in a sync.Once

disq:
```suggestion
var currentHub *Client
```

disq (approved):
LGTM with nit/styling

internal/analytics/client.go
erezrokah: I'm probably missing something here. We already call `initAnalytics` here https://github.com/cloudquery/cloudquery/blob/811855475622955dcfb323298292bde958f4372d/cmd/root.go#L157.

Where should I add another call?

disq: I think this is fine

PR #731, opened 2022-05-17
feat!: Remove provider migrations

disq:


pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+			return diag.FromError(fmt.Errorf("CreateTableDefinitions failed for %s: %w", t, err), diag.INTERNAL)
+		}
+		for _, sql := range up {
+			if err := db.Exec(ctx, sql); err != nil {
+				return diag.FromError(fmt.Errorf("error creating table %s: %w", t, err), diag.INTERNAL)
+			}
+		}
+	}
+
+	sta, err := state.NewMigratedClient(ctx, dsn, logging.NewZHcLog(&log.Logger, "sync"))
```
roneli, line 199: maybe if we touched the state remove the need to pass the logger and use the global one?

pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	DownloadLatest bool
+}
+
+type SyncResult struct {
+	State      SyncState
+	OldVersion string
+	NewVersion string
+}
+
+func Sync(ctx context.Context, storage database.Storage, pm *plugin.Manager, opts *SyncOptions) (*SyncResult, diag.Diagnostics) {
```
roneli, line 55: important to note: every time sync is called and we have a new version we drop all data, i.e if I run a policy with latest version or any command that does sync (most commands do) we will drop data and policy won't work for example.

pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+		res.State = NoChange
+	case cur.ParsedVersion.LessThan(want.ParsedVersion): // Upgrade
+		log.Debug().Stringer("provider", opts.Provider).Str("version", opts.Provider.Version).Msg("upgrading provider schema")
+		res.State = Upgraded
+	case cur.ParsedVersion.GreaterThan(want.ParsedVersion): // Downgrade
+		log.Debug().Stringer("provider", opts.Provider).Str("version", opts.Provider.Version).Msg("downgrading provider schema")
+		res.State = Downgraded
+	default:
+		return nil, diag.FromError(fmt.Errorf("sync: unhandled case"), diag.INTERNAL)
+	}
```
roneli, line 120: move to a function?

pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	defer db.Close()
+
+	rp.Version = s.Version
+
+	want := state.ProviderFromRegistry(rp)
+	if want.ParsedVersion == nil {
+		return nil, diag.FromError(fmt.Errorf("failing provider with invalid version %q", rp.Version), diag.INTERNAL)
+	}
+
+	sta, err := state.NewMigratedClient(ctx, storage.DSN(), logging.NewZHcLog(&log.Logger, "sync"))
```
roneli, line 86: I don't see you call in all places NewMigratedClient, why not just call it NewClient?

pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	for _, t := range sort.StringSlice(funk.Keys(resourceTables).([]string)) {
+		up, _, err := tc.CreateTableDefinitions(ctx, resourceTables[t], nil)
+		if err != nil {
+			return diag.FromError(fmt.Errorf("CreateTableDefinitions failed for %s: %w", t, err), diag.INTERNAL)
+		}
+		for _, sql := range up {
+			if err := db.Exec(ctx, sql); err != nil {
+				return diag.FromError(fmt.Errorf("error creating table %s: %w", t, err), diag.INTERNAL)
+			}
+		}
```
roneli, line 196: if any table creation fails, maybe we want to try to create all of them and collect all failed?

pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	tc := migration.NewTableCreator(logger, schema.PostgresDialect{})
+
+	for _, t := range sort.StringSlice(funk.Keys(resourceTables).([]string)) {
+		up, _, err := tc.CreateTableDefinitions(ctx, resourceTables[t], nil)
+		if err != nil {
+			return diag.FromError(fmt.Errorf("CreateTableDefinitions failed for %s: %w", t, err), diag.INTERNAL)
+		}
+		for _, sql := range up {
+			if err := db.Exec(ctx, sql); err != nil {
+				return diag.FromError(fmt.Errorf("error creating table %s: %w", t, err), diag.INTERNAL)
```
roneli, line 194: add for what resource the table creation failed

pkg/core/sync.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func dropProvider(ctx context.Context, dsn string, provider registry.Provider, resourceTables map[string]*schema.Table) error {
+	dType, _, err := sdkdb.ParseDialectDSN(dsn)
+	if err != nil {
+		return err
+	}
+
+	m, err := migrator.New(logging.NewZHcLog(&log.Logger, "migrator"), dType, nil, dsn, fmt.Sprintf("%s_%s", provider.Source, provider.Name))
```
roneli, line 155: feels like the migrator shouldn't exist anymore, didn't see the code in the SDK, so take this comment with a grain of salt.

pkg/core/state/client.go
```diff
@@ -32,22 +35,45 @@ func NewClient(db execution.QueryExecer, logger hclog.Logger) *Client {
 	}
 }
 
-func (c *Client) MigrateCore(ctx context.Context, de database.DialectExecutor) error {
-	err := c.db.Exec(ctx, "CREATE SCHEMA IF NOT EXISTS cloudquery")
-	if err != nil {
-		return err
+func NewMigratedClient(ctx context.Context, dsn string, logger hclog.Logger) (*Client, error) {
```
roneli, line 38: nit: NewClient is enough imo

roneli (changes requested):
Looks good, gave initial batch of comments, waiting for SDK PR so I can see the bigger picture and maybe have more comments. The only main one to be aware as we drop data, any sync with "newer" version will drop all data, which can cause issues for any command that isn't fetch basically.

pkg/core/state/client.go
disq: I don't want to init db and call MigrateCore every time I want an updated core. I'll fix this when I circle back to fix storage / make it transactional etc.

pkg/core/sync.go
disq: it wouldn't migrate core.

pkg/core/sync.go
disq: It wasn't in a function before.

pkg/core/sync.go
disq: migrator is now only used for drop provider (and schema_migrations tables)

also it exists because we still have the core schema to worry about.

pkg/core/sync.go
disq: `t` ? for subresource, I can't, unless I print the whole SQL statement.

pkg/core/sync.go
disq: it will be transactional later on, and won't break the schema.

pkg/core/sync.go
disq: Removed the logger param from both constructors, now they build one from the global logger in the method. The requirement was mostly because sdkdb and sdkmigrations also require logger in their constructors, which will/might soon go away.

pkg/core/sync.go
roneli: I think we should always migrate core when creating the client, so it doesn't really matter to give it a special name.

pkg/core/sync.go
roneli: still, even if its transactional, you want to run them all and then rollback, if two tables are problematic, I rather get two errors, then 1 error, fix, another error.

pkg/core/sync.go
disq: but then policy stuff doesn't work because of interfaces.

disq: it will be fixed, you shouldn't think too hard about this issue

pkg/core/sync.go
disq: the main sync part is now a fn

pkg/core/sync.go
disq: the signature stuff takes care of that now

pkg/core/sync.go
disq: Fixed, but in a tx after a failed query, all following queries would also fail:

```
cq=# begin;
BEGIN
cq=*# drop table aws_non_table;
ERROR:  table "aws_non_table" does not exist
cq=!# select count(1) from aws_ec2_instances;
ERROR:  current transaction is aborted, commands ignored until end of transaction block
cq=!# rollback;
ROLLBACK
```

pkg/core/fetch.go
```diff
@@ -205,30 +200,13 @@ func Fetch(ctx context.Context, storage database.Storage, pm *plugin.Manager, op
...
-			opts.ExtraFields = make(map[string]interface{})
-		}
-		opts.ExtraFields["cq_fetch_date"] = fd
-	}
-	log.Info().Interface("extra_fields", opts.ExtraFields).Bool("history_enabled", opts.History != nil).Msg("received fetch request")
+	log.Info().Interface("extra_fields", opts.ExtraFields).Msg("received fetch request")
 
-	// TODO: in future more components will want state, so make state more generic and passable via database.Storage
-	db, err := sdkdb.New(ctx, logging.NewZHcLog(&log.Logger, "fetch"), storage.DSN())
+	stateClient, err := state.NewClient(ctx, storage.DSN())
```
roneli, line 205: why not pass the state as part of the arguments, and create it once in the console.Client

pkg/core/state/migrations/postgres/3_v0.24.0.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    name TEXT NOT NULL,
+    version TEXT NOT NULL,
+    v_major INT NOT NULL,
+    v_minor INT NOT NULL,
+    v_patch INT NOT NULL,
+    v_pre TEXT NOT NULL,
+    v_meta TEXT NOT NULL,
+    tables JSONB NOT NULL, -- key: resource name, value: array of table names
+    signatures JSONB NOT NULL, -- key: resource name, value: table signature
+    CONSTRAINT "providers_id" PRIMARY KEY (source, name)
```
roneli, line 13: nit: would create some index here for name, and version, although number will be quite low, maybe its an over optimization

pkg/core/state/provider.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (a *stringMap) Scan(value interface{}) error {
+	b, ok := value.([]byte)
+	if !ok {
+		return errors.New("type assertion to []byte failed")
+	}
+
+	return json.Unmarshal(b, &a)
+}
```
roneli, line 161: nit: maybe put this in a separate file in state maybe types.go?

pkg/core/sync.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if err := tx.Commit(ctx); err != nil {
+		return diag.FromError(err, diag.DATABASE)
+	}
+
+	return nil
+}
+
+func dropProvider(ctx context.Context, db execution.QueryExecer, provider registry.Provider, tableNames map[string][]string, curSignatures, wantSignatures map[string]string) diag.Diagnostics {
+	{
```
roneli, line 200: nit: why the brackets?

pkg/core/sync.go
```diff
@@ -0,0 +1,258 @@
...
+	return "Unknown"
+}
+
+type SyncResult struct {
+	State      SyncState
+	OldVersion string
+	NewVersion string
+}
+
+const dropTableSQL = "DROP TABLE IF EXISTS %s CASCADE"
```
roneli, line 52: nit: why not use goqu?

pkg/core/sync.go
```diff
@@ -0,0 +1,258 @@
...
+
+	for _, t := range sort.StringSlice(funk.Keys(resourceTables).([]string)) {
+		up, err := migration.CreateTableDefinitions(ctx, schema.PostgresDialect{}, resourceTables[t], nil)
+		if err != nil {
+			diags = diags.Add(diag.FromError(err, diag.INTERNAL, diag.WithSummary("failed to get create table definition"), diag.WithResourceName(t)))
+			continue
+		}
+		for _, sql := range up {
+			if err := db.Exec(ctx, sql); err != nil {
+				return diags.Add(diag.FromError(err, diag.INTERNAL, diag.WithSummary("error creating table"), diag.WithResourceName(t)))
```
roneli, line 236: diag.Database

pkg/core/sync.go
```diff
@@ -0,0 +1,258 @@
...
+	case cur.ParsedVersion.GreaterThan(want.ParsedVersion): // Downgrade
+		log.Debug().Stringer("provider", provider).Str("version", provider.Version).Msg("downgrading provider schema")
+		res.State = Downgraded
+	default:
+		return nil, diag.FromError(fmt.Errorf("sync: unhandled case"), diag.INTERNAL)
+	}
+
+	if res.State != NoChange {
+		dd := syncTables(ctx, sta, cur, want, s.ResourceTables)
+		diags = diags.Add(dd)
```
roneli, line 112:
```suggestion
		diags = diags.Add(syncTables(ctx, sta, cur, want, s.ResourceTables))
```

pkg/core/sync.go
```diff
@@ -0,0 +1,258 @@
...
+		return diags
+	}
+
+	if err := tx.UninstallProvider(ctx, want.Registry()); err != nil {
+		return diag.FromError(fmt.Errorf("uninstall provider failed: %w", err), diag.INTERNAL)
+	}
+
+	if err := tx.InstallProvider(ctx, want); err != nil {
+		return diag.FromError(fmt.Errorf("state failed: %w", err), diag.INTERNAL)
+	}
```
roneli, line 190: why not have an UpdateProvider ?

pkg/core/sync.go
```diff
@@ -0,0 +1,258 @@
...
+			if err := db.Exec(ctx, fmt.Sprintf(dropTableSQL, strconv.Quote(t))); err != nil {
+				return diag.FromError(err, diag.DATABASE, diag.WithSummary("drop table failed"), diag.WithResourceName(t))
+			}
+		}
+	}
+
+	return nil
+}
+
+func installProvider(ctx context.Context, db execution.QueryExecer, resourceTables map[string]*schema.Table) diag.Diagnostics {
```
roneli, line 225: got confused with the state.InstallProvider, maybe call this createProviderTables or buildProviderTables?

roneli (changes requested):
Looks good, didn't approve as day off today and didn't run it / check it out locally. But overall mostly minor comments.

pkg/core/state/migrations/postgres/3_v0.24.0.up.sql
disq: it's an over optimization as every source/name can have only one row.

pkg/core/state/provider.go
disq: done

pkg/core/sync.go
disq: scoping. it's a block, migTable and q isn't used anywhere else, which can be complicated, so it eases reading.

pkg/core/sync.go
disq: this is more clear than unknown goqu. This was moved here from the SDK which had the line exactly like this.

pkg/core/sync.go
disq: fixed

pkg/core/sync.go
disq: sure

pkg/core/sync.go
disq: this works if we have no provider installed (no tables dropped + no row deleted) as well as updating a provider (`cur` can be nil, remember)

having an UpdateProvider would need either `INSERT ON CONFLICT UPDATE` (ugly) or an `UPDATE` statement (which wouldn't work in this case). And since this is all done inside a TX there's no gain, just complications?

pkg/core/sync.go
disq: dropProviderTables + createProviderTables

pkg/core/fetch.go
disq: Done

pkg/core/sync.go
```diff
@@ -219,6 +232,38 @@ func createProviderTables(ctx context.Context, db execution.QueryExecer, resourc
...
+	}
+
+	var dependentViews []string
+	if err := pgxscan.ScanAll(&dependentViews, rows); err != nil {
+		return diag.FromError(err, diag.DATABASE)
+	}
+
+	if len(dependentViews) > 0 {
+		return diag.Diagnostics{diag.NewBaseError(nil, diag.USER,
+			diag.WithSummary("One or more views depend on table %q, which is about to be dropped. Run with --force-drop to override this message which will drop all dependent views.", table),
```
disq, line 258: maybe capitalize `ALL`? because it will drop dependent views from ALL DROPPED tables

roneli (approved):
LGTM

PR #736, opened 2022-05-18
fix: Progressbar

disq:


erezrokah:
What does this PR fix? Is it still relevant?

disq:
closing, this progressbar pkg needs removing instead of workarounds anyway.

Issue #719, opened 2022-05-16
test: Test_fetch unit test is flaky


erezrokah:
See https://github.com/cloudquery/cloudquery/runs/6455492440?check_suite_focus=true#step:10:621

```
--- FAIL: Test_Fetch (35.45s)
    --- PASS: Test_Fetch/fetch-errors (8.58s)
    --- PASS: Test_Fetch/fetch-simple (8.38s)
    --- PASS: Test_Fetch/fetch-unique-fetch-id (8.39s)
    --- FAIL: Test_Fetch/fetch-timeout (4.29s)
    --- PASS: Test_Fetch/fetch-default-config (5.38s)
    --- PASS: Test_Fetch/fetch-duplicates (0.35s)
```

erezrokah:
I'll have to find a reproduction first (probably only happens in CI) or add more data to the logs when it fails

erezrokah:
Still flaky https://github.com/cloudquery/cloudquery/runs/6556240275?check_suite_focus=true#step:11:518

erezrokah:
Should be fixed by https://github.com/cloudquery/cloudquery/pull/790

PR #741, opened 2022-05-19
fix: Handle DeadlineExceeded errors

erezrokah:

#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/719

The test timeout test expects the fetch to fail at a very specific point in the code, while the `DeadlineExceeded` error can be returned in multiple places (every place that uses the context).
I'm guessing this happens more often in CI as the machines are slower so we reach the deadline before and fail in a different way.

You can reproduce using a debugger and wait more than 4 seconds before stepping into `resp, diags := Fetch(ctx, storage, pManager, &tc.Options)`

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/core/fetch.go
+++ b/pkg/core/fetch.go
@@ -226,7 +226,8 @@ func Fetch(ctx context.Context, storage database.Storage, pm *plugin.Manager, op
 	defer db.Close()
 	stateClient := state.NewClient(db, logging.NewZHcLog(&log.Logger, "fetch"))
 	// migrate CloudQuery core tables to latest version
-	if err := stateClient.MigrateCore(ctx, storage.DialectExecutor()); err != nil {
+	// context.DeadlineExceeded is handled inside runProviderFetch that returns a response and summary
+	if err := stateClient.MigrateCore(ctx, storage.DialectExecutor()); err != nil && !errors.Is(err, context.DeadlineExceeded) {
 		return nil, diag.FromError(err, diag.DATABASE, diag.WithSummary("failed to migrate cloudquery_core tables"))
 	}
 
@@ -303,13 +304,26 @@ func runProviderFetch(ctx context.Context, pm *plugin.Manager, info ProviderInfo
 	})
 	if err != nil {
 		pLog.Error().Err(err).Msg("failed to configure provider")
+		var (
+			d      diag.Diagnostics
+			status FetchStatus
+		)
+
+		if isErrorCancellation(err) {
+			d = getCancellationDiag(err)
+			status = FetchCanceled
+		} else {
+			d = diag.FromError(err, diag.INTERNAL)
+			status = FetchConfigureFailed
+		}
+
 		return &ProviderFetchSummary{
 			Name:             info.Provider.Name,
 			Alias:            info.Config.Alias,
 			Version:          providerPlugin.Version(),
 			FetchedResources: make(map[string]ResourceFetchSummary),
-			Status:           FetchConfigureFailed,
-		}, diag.FromError(err, diag.INTERNAL)
+			Status:           status,
+		}, d
 	}
 	if resp.Diagnostics.HasErrors() {
 		return &ProviderFetchSummary{
@@ -324,6 +338,18 @@ func runProviderFetch(ctx context.Context, pm *plugin.Manager, info ProviderInfo
 	pLog.Info().Msg("provider configured successfully")
 	summary, diags := executeFetch(ctx, pLog, providerPlugin, info, metadata, opts.UpdateCallback)
 	return summary, convertToFetchDiags(diags, info.Provider.Name, providerPlugin.Version())
+}
+
+func isErrorCancellation(err error) bool {
+	if st, ok := status.FromError(err); ok && (st.Code() == gcodes.Canceled || st.Code() == gcodes.DeadlineExceeded) {
+		return true
+	}
+
+	return errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)
+}
+
+func getCancellationDiag(err error) diag.Diagnostics {
+	return diag.FromError(err, diag.USER, diag.WithSummary("provider fetch was canceled by user / fetch deadline exceeded"))
 }
 
 func executeFetch(ctx context.Context, pLog zerolog.Logger, plugin plugin.Plugin, info ProviderInfo, metadata map[string]interface{}, callback FetchUpdateCallback) (*ProviderFetchSummary, diag.Diagnostics) {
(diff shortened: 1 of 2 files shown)


pkg/core/fetch_test.go
```diff
@@ -256,7 +256,20 @@ func Test_Fetch(t *testing.T) {
 			}
 			resp, diags := Fetch(ctx, storage, pManager, &tc.Options)
 			if tc.ExpectedDiags != nil {
-				assert.ElementsMatch(t, tc.ExpectedDiags, diag.FlattenDiags(diags, false))
+				flattenedDiags := diag.FlattenDiags(diags, false)
```
erezrokah, line 259: So we can do partial matching.
Not sure if there's a better way to do it

pkg/core/fetch.go
```diff
@@ -226,7 +226,8 @@ func Fetch(ctx context.Context, storage database.Storage, pm *plugin.Manager, op
 	defer db.Close()
 	stateClient := state.NewClient(db, logging.NewZHcLog(&log.Logger, "fetch"))
 	// migrate CloudQuery core tables to latest version
-	if err := stateClient.MigrateCore(ctx, storage.DialectExecutor()); err != nil {
+	// context.DeadlineExceeded is handled inside runProviderFetch that returns a response and summary
+	if err := stateClient.MigrateCore(ctx, storage.DialectExecutor()); err != nil && !errors.Is(err, context.DeadlineExceeded) {
```
erezrokah, line 230: I don't like this, but a better fix might require a bigger change in the code

disq: this will change in #731 but welcome to add the fix later

better fix needs a touch to `classifyError` maybe?

erezrokah: >this will change in #731 but welcome to add the fix later

Thanks, I figured this will have to change soon. Happy to apply the fix later.

>better fix needs a touch to classifyError maybe?

Can you explain this a bit?
Do you mean add `DeadlineExceeded` (code and error) here https://github.com/cloudquery/cloudquery/blob/a3ecfc9166170e1bb77011befd11a5fbe1c86007/pkg/errors/error.go#L54 and https://github.com/cloudquery/cloudquery/blob/a3ecfc9166170e1bb77011befd11a5fbe1c86007/pkg/errors/error.go#L65

so they don't get reported?

disq: yes, `if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {`

pkg/core/fetch.go
```diff
@@ -326,6 +337,18 @@ func runProviderFetch(ctx context.Context, pm *plugin.Manager, info ProviderInfo
 	return summary, convertToFetchDiags(diags, info.Provider.Name, providerPlugin.Version())
 }
 
+func isErrorDeadlineExceeded(err error) bool {
+	if st, ok := status.FromError(err); ok && st.Code() == gcodes.Canceled || st.Code() == gcodes.DeadlineExceeded {
+		return true
+	}
+
+	return false
```
disq, line 345:
```suggestion
	return errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)
```

pkg/core/fetch.go
disq: or just `if isErrorDeadlineExceeded(err)` with the addition below.

disq: since we have the isErrorDeadlineExceeded in the classifier this line/block change can be reverted (and not cause a conflict with my PR 😂 )

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/741/commits/cfb8b202b522fd453ad55dca2294967c64a90020

pkg/errors/error.go
```diff
@@ -50,22 +50,35 @@ const (
...
+}
+
 // classifyError classifies given error by type and internals. Successfully classified (not errNoClass) errors don't get reported to sentry.
 func classifyError(err error) errClass {
+	if isErrorCancellation(err) {
+		return errCancellation
+	}
+
 	if st, ok := status.FromError(err); ok {
-		if st.Code() == gcodes.Canceled {
```
erezrokah, line 56: Removed this to group related logic together (cost is re-running `status.FromError`)

pkg/errors/error.go
```diff
@@ -50,22 +50,35 @@ const (
 	errDatabase     = errClass("database")
 )
 
+func isCodeCancellation(code gcodes.Code) bool {
+	return code == gcodes.Canceled || code == gcodes.DeadlineExceeded
```
erezrokah, line 54: In other places we use a map, but since it's only 2 values this feels simpler to me.
Happy to change it if needed

pkg/errors/error.go
```diff
@@ -50,22 +50,35 @@ const (
...
+}
+
+func isErrorCancellation(err error) bool {
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
erezrokah, line 58: Same here, we could use an array and iterate, but kept it like this since we only compare to 2 values

pkg/core/fetch.go
erezrokah: I don't mind waiting with this after you merge your PR.
It's not a major issue and probably only happens in CI tests

pkg/errors/error.go
```diff
@@ -52,20 +52,17 @@ const (
 
 // classifyError classifies given error by type and internals. Successfully classified (not errNoClass) errors don't get reported to sentry.
 func classifyError(err error) errClass {
+	if core.IsErrorCancellation(err) {
```
disq, line 55: I would move IsErrorCancellation in `errors` (here) and call it from core as errors.IsErrorCancellation

disq: errors shouldn't import core? but it does... maybe check how I did `IsConfigureDiagnostic`

erezrokah:
For anyone watching this PR @disq and I decided to do the refactor to remove the `core` import from `errors` in another PR

pkg/core/fetch.go
```diff
@@ -303,13 +304,23 @@ func runProviderFetch(ctx context.Context, pm *plugin.Manager, info ProviderInfo
 	})
 	if err != nil {
 		pLog.Error().Err(err).Msg("failed to configure provider")
+		var d diag.Diagnostics
```
disq, line 307: nit:
```suggestion
		var (
			d diag.Diagnostics
			status FetchStatus
		)
```

disq (approved):
lgtm with nit. I think you should merge this before the migrations stuff.

PR #744, opened 2022-05-19
fix: Diag import clean up

disq:
- pkg/errors can be imported from anywhere now (doesn't import core)
- new methods from https://github.com/cloudquery/cloudquery/pull/741 moved to errors pkg
- ~~tried importing pkg/errors as `cqerrors` (to leave stdlib errors as `errors`) which didn't work in the workflow, had to switch around (`errors` and `stderrors` respectively)~~ my bad

### DIFF
--- a/pkg/core/diagnostics.go
+++ b/pkg/core/diagnostics.go
@@ -1,23 +1,20 @@
 package core
 
-import "github.com/cloudquery/cq-provider-sdk/provider/diag"
+import (
+	"github.com/cloudquery/cloudquery/pkg/plugin"
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
+	"github.com/spf13/viper"
+)
 
-type ConfigureDiagnostic struct {
+type SentryDiagnostic struct {
 	diag.Diagnostic
+
+	Tags   map[string]string
+	Ignore bool
 }
 
-func (d *ConfigureDiagnostic) IsConfigureDiagnostic() bool {
-	return true
-}
-
-type FetchDiagnostic struct {
-	diag.Diagnostic
-	Provider string
-	Version  string
-}
-
-func (d *FetchDiagnostic) IsFetchDiagnostic() (bool, string, string) {
-	return true, d.Provider, d.Version
+func (d *SentryDiagnostic) IsSentryDiagnostic() (bool, map[string]string, bool) {
+	return true, d.Tags, d.Ignore
 }
 
 type DiagnosticsSummary struct {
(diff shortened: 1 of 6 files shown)


pkg/core/diagnostics.go
```diff
@@ -10,32 +10,14 @@ func (d *ConfigureDiagnostic) IsConfigureDiagnostic() bool {
...
-	for i, d := range diags {
-		fd[i] = FetchDiagnostic{
-			Diagnostic: d,
-			Provider:   provider,
-			Version:    version,
-		}
-	}
-	return fd
+func (d *FetchDiagnostic) IsFetchDiagnostic() (bool, string, string) {
+	return true, d.Provider, d.Version
```
disq, line 20: This is not very scalable as we should stick to base types (to avoid importing this pkg from errors)

pkg/errors/sentry.go
```diff
@@ -22,7 +22,7 @@ func CaptureError(err error, tags map[string]string) {
 }
 
 func CaptureDiagnostics(dd diag.Diagnostics, tags map[string]string) {
-	allowUnmanaged := core.Version == core.DevelopmentVersion && viper.GetBool("debug-sentry")
+	allowUnmanaged := viper.GetBool("debug-sentry")
```
disq, line 25: `core` reference removed because we shouldn't care which binary we're on if `--debug-sentry` is specified? (or we can embed `core.Version == core.DevelopmentVersion` as a bool inside the fetch diag as a fourth arg... which also kinda works)

pkg/errors/sentry.go
```diff
@@ -32,11 +32,11 @@ func CaptureDiagnostics(dd diag.Diagnostics, tags map[string]string) {
 			continue
 		}
 		sentry.WithScope(func(scope *sentry.Scope) {
-			if t, ok := d.(core.FetchDiagnostic); ok {
-				if allowUnmanaged && t.Version == plugin.Unmanaged {
+			if ok, p, v := isFetchDiagnostic(d); ok {
+				if !allowUnmanaged && v == plugin.Unmanaged {
```
disq, line 36: Logic was slightly off here.

pkg/errors/sentry.go
```diff
@@ -32,11 +32,11 @@ func CaptureDiagnostics(dd diag.Diagnostics, tags map[string]string) {
...
 					return
 				}
-				scope.SetTags(map[string]string{"provider": t.Provider, "provider_version": t.Version, "resource": t.Description().Resource})
+				scope.SetTags(map[string]string{"provider": p, "provider_version": v, "resource": d.Description().Resource})
```
disq, line 39: Thought about making a `TaggedDiagnostic` to embed tags but turns out we also use `IsConfigureDiag` to determine some things about error classification.

pkg/errors/cancel.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+	stderrors "errors"
+
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
+
+	gcodes "google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func IsCancelation(err error) bool {
```
disq, line 13: Cancelation or Cancellation? 🤔 Both is correct but since `context.Canceled` is with one L I wanted to keep it the same.

erezrokah (approved):

pkg/errors/sentry.go
disq: Done in [1c11f31](https://github.com/cloudquery/cloudquery/pull/744/commits/1c11f3130a446a99761ee4b072da8239fd288b18)

pkg/core/diagnostics.go
disq: Fixed in [1c11f31](https://github.com/cloudquery/cloudquery/pull/744/commits/1c11f3130a446a99761ee4b072da8239fd288b18)

erezrokah (approved):
Very cool with the tags 😀

roneli (approved):
LGTM

PR #745, opened 2022-05-20
fix: FetchId column regression

disq:
Fix regression introduced in https://github.com/cloudquery/cloudquery/pull/714


disq:
Although this is still broken because there's an SDK issue

roneli (approved):

Issue #746, opened 2022-05-22
Unable to test a new provider in debug mode


andrewthetechie:
### Describe the Bug
While trying to develop my own provider, I am having issues in attempting to test it. Cloudquery always does a lookup against the hub and when it fails to find my provider, it fails. My provider is not listed in the hub.

### Expected Behavior
In debug mode, cloudquery uses the local version of my provider via CQ_REATTACH_PROVIERS without talking to the hub.

### CloudQuery Version
Version: 0.23.4
Commit: 96ec2c96eb6fd8dc67e8f7f666618a8bcfbdd262
Date: 2022-05-17T14:33:18Z

### Debug Output
```CQ_PROVIDER_DEBUG=1 .cq/providers/cloudquery/datadog/v0.0.1-darwin_arm64 --help
{"@level":"debug","@message":"plugin address","@module":"Datadog","@timestamp":"2022-05-22T00:25:00.884348-05:00","address":"/var/folders/dl/4jg1h4h53dv_pzzpbt6sntyr0000gn/T/plugin3220677151","network":"unix"}
Provider started, to attach Cloudquery set the CQ_REATTACH_PROVIDERS env var:

        export CQ_REATTACH_PROVIDERS=/Users/andrew/Documents/code/cq-provider-datadog/.cq_reattach
```

In a second terminal
```
CQ_PROVIDER_DEBUG=1 CQ_REATTACH_PROVIDERS=/Users/andrew/Documents/code/cq-provider-datadog/.cq_reattach ./cloudquery fetch --no-verify
Anonymous telemetry collection and crash reporting enabled. Run with --no-telemetry to disable, or check docs at https://docs.cloudquery.io/docs/cli/telemetry
Syncing CloudQuery providers [datadog@v0.0.1]

Initializing CloudQuery Providers...

⚠️ cq-provider-datadog@v0.0.1 skipped verification...    0s  100 %

Finished provider initialization...

Checking available provider updates...

Diagnostics:

Type: Internal Severity: Error
        Summary: failed to find provider[datadog] latest version

Sync Diagnostics:

Type: Internal Severity: Error
        Summary: failed to find provider[datadog] latest version
```

Config.hcl
```
cloudquery {
  plugin_directory = "./cq/providers"
  policy_directory = "./cq/policies"

  provider "datadog" {
    version = "latest"
  }

  connection {
    username = "postgres"
    password = "pass"
    host     = "localhost"
    port     = 5432
    database = "postgres"
[...]

PR #748, opened 2022-05-22
fix: Don't attempt to download provider in re-attach mode

erezrokah:

#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/746

---

- [x] Update or add tests 🧪

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -154,7 +154,7 @@ func init() {
 
 	rootCmd.SetHelpCommand(&cobra.Command{Hidden: true})
 	rootCmd.SetUsageTemplate(usageTemplate)
-	cobra.OnInitialize(initConfig, initLogging, initUlimit, initSentry, initAnalytics, initReattachMode)
+	cobra.OnInitialize(initConfig, initLogging, initUlimit, initSentry, initAnalytics)
 }
 
 func initUlimit() {
@@ -217,10 +217,4 @@ func logInvocationParams(cmd *cobra.Command, args []string) {
 	})
 
 	l.Str("command", cmd.CommandPath()).Strs("args", args).Msg("Invocation parameters")
-}
-
-func initReattachMode() {
-	if viper.GetString("reattach-providers") != "" {
-		viper.Set("no-verify", true)
-	}
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -109,9 +109,7 @@ func init() {
 
 	// Logging Flags
 	rootCmd.PersistentFlags().BoolVarP(&logging.GlobalConfig.Verbose, "verbose", "v", false, "Enable Verbose logging")
-	_ = viper.BindPFlag("verbose", rootCmd.PersistentFlags().Lookup("verbose"))
 	rootCmd.PersistentFlags().BoolVar(&logging.GlobalConfig.ConsoleLoggingEnabled, "enable-console-log", false, "Enable console logging")
-	_ = viper.BindPFlag("enable-console-log", rootCmd.PersistentFlags().Lookup("enable-console-log"))
 	rootCmd.PersistentFlags().BoolVar(&logging.GlobalConfig.EncodeLogsAsJson, "encode-json", false, "EncodeJson makes the logging framework log files with JSON instead of KV")
 	rootCmd.PersistentFlags().BoolVar(&logging.GlobalConfig.FileLoggingEnabled, "enable-file-logging", true, "enableFileLogging makes the framework logging to a file")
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -135,6 +133,8 @@ func init() {
 	_ = rootCmd.PersistentFlags().MarkHidden("plugin-dir")
 	_ = rootCmd.PersistentFlags().MarkHidden("policy-dir")
 
+	_ = viper.BindPFlag("verbose", rootCmd.PersistentFlags().Lookup("verbose"))
```
erezrokah, line 136: Moved the binding here to have everything in the same place (we could also put each bind below each flag definition, but we should be consistent with either approach)

cmd/root.go
```diff
@@ -218,3 +218,9 @@ func logInvocationParams(cmd *cobra.Command, args []string) {
 
 	l.Str("command", cmd.CommandPath()).Strs("args", args).Msg("Invocation parameters")
 }
+
+func initReattachMode() {
+	if viper.GetString("reattach-providers") != "" {
```
erezrokah, line 223: Unrelated to the fix, but I don't think there's a case we'd want to verify when `reattach-providers` is set (correct me if I'm wrong).
When verifying we download the checksum from GitHub, so not sure how that can work with a local provider (unless the user ensures those match).

pkg/plugin/manager.go
```diff
@@ -65,8 +66,8 @@ func (m *Manager) DownloadProviders(ctx context.Context, providers []registry.Pr
 	log.Debug().Interface("providers", providers).Msg("Downloading required providers")
 	downloaded := make([]registry.ProviderBinary, len(providers))
 	for i, rp := range providers {
-		if _, ok := m.clients[rp.String()]; ok {
-			log.Debug().Str("name", rp.Name).Str("version", rp.Version).Msg("Skipping provider download, using reattach instead")
+		if _, ok := m.clients[rp.Name]; ok {
```
erezrokah, line 69: `clients` are de-serialized from the `.cq_reattach` file which includes only the name of the provider. `rp.String()` format is `name@version`, hence this was failing to match the providers.

pkg/plugin/manager.go
```diff
@@ -65,8 +66,8 @@ func (m *Manager) DownloadProviders(ctx context.Context, providers []registry.Pr
...
+			ui.ColorizedOutput(ui.ColorInfo, fmt.Sprintf("Skipping provider %s download, using reattach instead\n", ui.Colorize(ui.ColorSuccessBold, false, rp.Name)))
```
erezrokah, line 70: I think this is quite a useful log message to have to ensure the re-attach was successful

cmd/root.go
roneli: that is correct, but also incorrect, if I have one provider (aws) not reattached and one provider in reattach (datadog) if we disable verify than aws provider is not verified.

the plugin manager should not verify/download plugins that are reattached, so I would add the fix in the pluginmanager to avoid this check for reattached providers.

roneli (changes requested):
LGTM, I was guessing it was the no-verify flag that should be turned on, but a user shouldn't guess that as per you comment if reattach on a provider is on it shouldn't verify it.

Pr

erezrokah:
>that is correct, but also incorrect, if I have one provider (aws) not reattached and one provider in reattach (datadog) if we disable verify than aws provider is not verified.
the plugin manager should not verify/download plugins that are reattached, so I would add the fix in the pluginmanager to avoid this check for reattached providers.

Cool. I'll remove it from this PR and we can do it separately

erezrokah:
Done in https://github.com/cloudquery/cloudquery/pull/748/commits/6dd3c9994b86f6a32a2f6b4f581a047c407e2f82

pkg/plugin/manager.go
erezrokah, line 69: `rp.String()` -> `rp.Name` is the bug fix

erezrokah:
>Cool. I'll remove it from this PR and we can do it separately

Actually with the fix in this PR, we won't get to the code that uses `--no-verify` when using a re-attach provider.

roneli (approved):
LGTM, maybe we should add a sanity test with an provider maybe use cq-unverified-provider download it and run it with debug? (we can do in other PR) so this can be merged so users can work with it.

erezrokah:
> LGTM, maybe we should add a sanity test with an provider maybe use cq-unverified-provider download it and run it with debug? (we can do in other PR) so this can be merged so users can work with it.

Yes please :) I opened https://github.com/cloudquery/cloudquery-issues/issues/360 to track this

PR #778, opened 2022-05-24
fix: Remove always nil return value

erezrokah:

#### Summary

New linter caught this. I'm doing this in a separate PR to make verifying the change easier

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/module/drift/config_parser.go
+++ b/pkg/module/drift/config_parser.go
@@ -164,7 +164,7 @@ func (p *Parser) Decode(body hcl.Body, allowedProvider string, diags hcl.Diagnos
 }
 
 // interpret iterates over every provider/resource and replaces missing values with the ones in wildprovider/wildresource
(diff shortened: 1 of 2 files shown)


pkg/module/drift/config_parser.go
```diff
@@ -175,7 +175,6 @@ func (p *Parser) interpret(cfg *BaseConfig) hcl.Diagnostics {
 			}
 		}
 	}
-	return nil
```
erezrokah, line 178: Return value is always `nil` and will actually cause a panic in callers I believe

disq: https://go.dev/play/p/JxmTqiIB3Wj?v=goprev

erezrokah: Cool, thanks for the context. Should we still remove the return value? It's not really used

disq: Sure.

disq (approved):

PR #779, opened 2022-05-24
fix: Remove unused function return value

erezrokah:

#### Summary

Noticed this while working on the lint errors. Doesn't look like the second return value is used anywhere.

---

- [x] Update or add tests 🧪


disq:
This won't work and will cause panics.

disq:
Also it's not a `Split*` function when it's not returning multiple values so if we were to apply this change the name would need to change as well.

erezrokah:
Not sure if we want the change in the PR or not based on the comments. Assuming we don't, so closing the PR.
If I got this wrong let me know and I'll re-open

PR #782, opened 2022-05-25
refactor: Use pointer for connection receiver

erezrokah:

#### Summary

Another lint related issue. I don't think it's a "real" issue as we update the DSN outside of `BuildFromConnParams`.
However it might be a bit confusing that the DNS is updated and the port and type aren't.

A smaller fix would be to only change `c Connection` to `c *Connection`

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -80,7 +80,7 @@ func (c Connection) IsAnyConnParamsSet() bool {
 	return c.Type != "" || c.Username != "" || c.Password != "" || c.Host != "" || c.Port != 0 || c.Database != "" || c.SSLMode != "" || len(c.Extras) > 0
 }
 
-func (c Connection) BuildFromConnParams() (*dburl.URL, error) {
+func (c *Connection) BuildFromConnParams() error {
 	if c.Port == 0 {
 		c.Port = 5432
 	}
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -121,10 +121,14 @@ func (c Connection) BuildFromConnParams() (*dburl.URL, error) {
 	}
 	u.RawQuery = v.Encode()
 
-	return &dburl.URL{
+	url := &dburl.URL{
```
disq, line 124: bad variable name.

erezrokah: Yep :) Just realized it as the new linter errored on it

pkg/config/config.go
```diff
@@ -121,10 +121,14 @@ func (c Connection) BuildFromConnParams() (*dburl.URL, error) {
 	}
 	u.RawQuery = v.Encode()
 
-	return &dburl.URL{
+	dbUrl := &dburl.URL{
```
disq, line 124: just do `c.DSN = &dburl.URL{params}.String()`?

erezrokah: Sure https://github.com/cloudquery/cloudquery/pull/782/commits/7138a8767724f70d3f8b024928807a837de0fb3d

roneli (approved):

PR #787, opened 2022-05-26
fix: Panic in loadPolicyFromSource

irmatov:
The reason is DecodePolicy does not actually return error diag when there
is parse error. But the returned policy is nil leading to panic.

### DIFF
--- a/pkg/policy/parser.go
+++ b/pkg/policy/parser.go
@@ -59,28 +59,29 @@ func decodePolicy(body hcl.Body, basePath string) (*Policy, hcl.Diagnostics) {
 		return nil, diags
 	}
 	if len(content.Blocks) > 1 {
-		return nil, hcl.Diagnostics{{
+		return nil, diags.Append(&hcl.Diagnostic{
 			Severity: hcl.DiagError,
 			Summary:  `Only single root policy block allowed`,
 			Detail:   `Only a single policy block is allowed in root level policy`,
 			Subject:  &content.MissingItemRange,
-		}}
+		})
 	}
 
 	for _, block := range content.Blocks {
 		switch block.Type {
 		case "policy":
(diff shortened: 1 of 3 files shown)


pkg/policy/parser.go
```diff
@@ -53,8 +53,9 @@ var policySchema = &hcl.BodySchema{
 	},
 }
 
-func DecodePolicy(body hcl.Body, diags hcl.Diagnostics, basePath string) (*Policy, hcl.Diagnostics) {
+func decodePolicy(body hcl.Body, diags hcl.Diagnostics, basePath string) (*Policy, hcl.Diagnostics) {
 	content, contentDiags := body.Content(policyWrapperSchema)
+	diags = append(diags, contentDiags...)
 	if contentDiags.HasErrors() {
 		return nil, diags
 	}
```
roneli, line 61:
```suggestion
	content, diags := body.Content(policyWrapperSchema)
	if diags.HasErrors() {
		return nil, diags
	}
```

roneli (approved):
LGTM had one nit/suggestion

pkg/policy/parser.go
irmatov: This way we would lose `diags` that were already passed to us as an argument. How knows what's inside and what errors it might have.

disq: So you're suggesting to change fn signature also, to remove diags?

irmatov: @disq I'm not sure I understand :)

disq: comment was to @roneli we just happened to post at the same minute :)
I agree with your points.

irmatov: Removed unnecessary diags argument

disq (approved):

PR #788, opened 2022-05-26
fix: Error classifier improvements

disq:
Some issues from sentry.

The context canceled error is reproduced by hitting CTRL+C as soon as seeing the `INF All regions specified in config.yml. Assuming all regions` message in log, during `cloudquery fetch --enable-console-log`.


roneli (approved):
LGTM

PR #791, opened 2022-05-26
chore: Refactor OsFs.DownloadFile

irmatov:
Extract actual http part from OsFs as it is not its responsibility to do
network requests.

### DIFF
--- a/internal/file/file.go
+++ b/internal/file/file.go
@@ -1,15 +1,12 @@
 package file
 
 import (
-	"context"
-	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"sync"
 
+	"github.com/cloudquery/cloudquery/internal/network"
 	"github.com/cloudquery/cloudquery/pkg/ui"
-
 	"github.com/spf13/afero"
 )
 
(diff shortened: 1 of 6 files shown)


disq:
LGTM code-wise but maybe we need a bigger refactor around OsFs?

irmatov:
@disq one step at a time

yevgenypats:
What issue does it fix?

yevgenypats:
Im on the fence here because two main things:
1) as @disq said bigger refactor might be coming anyway
2) The focus is sentry erros, which is still through the roof and we are not able to maintain them nor to configure them correctly yet and it's a 1st priority to build a sustainable product with sustainable team and support team. Once we are able to do that and move sentry errors to manageable place where we can do on-call I would go about more refactoring.

yevgenypats:
Sorry for being a blocker here but I think right it's a no-merge from me especially because we have so many bugs and I don't want to introduce more because of a refactor that Im also personally not sure about this one and I also don't think everything should be functional as it complicates the code and even though it is easier to test it is actually becomes hard to test the real world scenario.

PR #793, opened 2022-05-26
fix: Normalize selector

bbernays:

#### Summary

---

- [ ] Lint your changes via `golangci-lint run --new-from-rev main` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [ ] Update or add tests 🧪

### DIFF
--- a/pkg/policy/execute.go
+++ b/pkg/policy/execute.go
@@ -500,11 +500,11 @@ func parseRow(columns []string, values []interface{}, identifiers []string, reas
 	return r, nil
 }
 
-func normalizeCheckSelector(policyExecution *state.PolicyExecution, policyPath []string, policyName string) string {
+func normalizeCheckSelector(policyExecution *state.PolicyExecution, policyPath []string, checkName string) string {
 	selector := []string{policyExecution.PolicyName}
 	if !strings.HasPrefix(selector[0], policyExecution.Location+"//") {
 		selector[0] = policyExecution.Location + "/"
 	}
 	selector = append(selector, policyPath...)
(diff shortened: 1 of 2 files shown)


pkg/policy/execute.go
```diff
@@ -245,15 +245,24 @@ func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Pol
 	return &total, nil
 }
 
-func (e *Executor) createCheckResult(ctx context.Context, selector []string, policyExecution *state.PolicyExecution, q *Check, qr *QueryResult, err error) error {
+func normalizeCheckSelector(policyExecution *state.PolicyExecution, policyPath []string, policyName string) string {
```
disq, line 248: Bare functions (`normalizeCheckSelector`) should be towards the end of the file

bbernays: fixed

pkg/policy/execute.go
```diff
@@ -245,15 +245,24 @@ func (e *Executor) Execute(ctx context.Context, req *ExecuteRequest, policy *Pol
...
+	selector := []string{policyExecution.PolicyName}
+	if !strings.HasPrefix(selector[0], policyExecution.Location+"//") {
+		selector[0] = policyExecution.Location + "/"
```
disq: Not sure I get the logic of overwriting policyName here, and then appending it in L254 regardless of if it was overridden or not.

bbernays: Good catch. I made a mistake in naming... `policyName` should be `checkName`

PR #810, opened 2022-05-29
feat: Add database info metrics

roneli:

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/pkg/core/database/postgres/executor.go
+++ b/pkg/core/database/postgres/executor.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cloudquery/cloudquery/pkg/core/database/model"
+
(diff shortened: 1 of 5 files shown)


pkg/core/database/postgres/executor.go
```diff
@@ -39,13 +41,19 @@ func (e *Executor) Validate(ctx context.Context) (bool, error) {
 	var dbIdErr error
 	e.dbId, dbIdErr = GetDatabaseId(ctx, pool)
 
+	e.info, _ = GetDatabaseInfo(ctx, pool)
```
erezrokah, line 44: [non-blocking] If we don't use the error anywhere I'd say remove it altogether. We could also return a default info model so we'd know it errored out

erezrokah (approved):
LGTM with a non blocking comment on handling errors

erezrokah (approved):
Looks great, I think you need to update the test file

PR #822, opened 2022-06-01
fix: Changed level of bad config diagnostics

amanenk:

#### Summary

- changed level of bad config to USER
- added printing of diags 

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/ui/console/client.go
+++ b/pkg/ui/console/client.go
@@ -193,7 +193,7 @@ func (c Client) DownloadProviders(ctx context.Context) (diags diag.Diagnostics) 
 	return diags
 }
 
-func (c Client) Fetch(ctx context.Context) (result *core.FetchResponse, diags diag.Diagnostics) {
+func (c Client) Fetch(ctx context.Context) (*core.FetchResponse, diag.Diagnostics) {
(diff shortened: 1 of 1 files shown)


pkg/ui/console/client.go
```diff
@@ -193,10 +193,12 @@ func (c Client) DownloadProviders(ctx context.Context) (diags diag.Diagnostics)
 	return diags
 }
 
-func (c Client) Fetch(ctx context.Context) (*core.FetchResponse, diag.Diagnostics) {
+func (c Client) Fetch(ctx context.Context) (result *core.FetchResponse, diags diag.Diagnostics) {
 	if _, dd := c.SyncProviders(ctx, c.cfg.Providers.Names()...); dd.HasErrors() {
 		return nil, dd
 	}
+	defer printDiagnostics("Fetch", &diags, viper.GetBool("redact-diags"), viper.GetBool("verbose"))
+
```
roneli, line 201: you already have printDiagnostics below, so one of the two neds to be used, otherwise it will print twice

roneli (changes requested):

pkg/ui/console/client.go
amanenk: Oh. you are right. In my case it was printing only once so I didn't notice it.

roneli (approved):
LGTM

Issue #740, opened 2022-05-19
More policy output improvements


shimonp21:
### Is your feature request related to a problem? Please describe.
Two separate issues here:

1. Currently, when outputting policy results to a json file, the user needs to specify `--output-dir=<some-dir>`.
    Then, `cloudquery` will create a file `<policy-name>.json` in this directory. For example, `cloudquery policy run aws//cis_v1.2.0 --output-dir=.` will create a file named `aws.json`.
    This means that when a user wants to use policy output in the CI or similar, this user can't know ahead of time what will be the name of the file that will be created by `cloudquery policy run`, and will probably need to do a 'dry-run' and somehow hardcode the filename.
    A corollary is that a user can't use the same output-dir for two different subpolicies' results:
    ```
    cloudquery policy run aws//cis_v1.2.0 --output-dir=.
    cloudquery policy run aws//pci_dss_v3.2.1  --output-dir=.
    ```
    The second `policy run` will overwrite the same `aws.json` file used in the first `policy run`.
2. Currently, in policy output, every check only has the check name, rather than full path. This makes it impossible to tell which check is which (except maybe by description, which is **not** part of our public interface and **should not** be used by user's CI or scripts.
    For example, in `aws` policy output results, there are 19 checks name `1`:
![image](https://user-images.githubusercontent.com/37939765/169258411-726757b7-4498-418d-af8c-1998cbc69e33.png)

### Describe the solution you'd like
1. A `--output-file` flag. Should be mutually-exclusive with `--output-dir`.
2. A `path` or `fullPath` property in check results, in addition to `name`. TBD if the path should contain name as well (leaning hard toward "yes").

shimonp21:
3. Currently, it is allowed to do `cloudquery policy run` (without specifying any policy), if there are policies specified in `config.hcl`. If we move to `--output-file` (we should!), then this option needs to be disallowed. 

```
policy "AAA" {
  source = "aws"
}

policy "BBB" {
  source = "gcp"
}
```

4. A more radical change that I think needs to be considered is disallowing specifying policies in the config.hcl.

shimonp21:
4. A more radical change is to just remove the policies from the config file completely. 
    - A config file is not the right place for data/policies. We have a perfectly good way right now of running policies from the local filesystem (`cloudquery policy run ./my-custom-policy`).
    - Code like
       ```
       policy "AAA" {
          source = "aws"
       }
       ```
[...]

PR #841, opened 2022-06-03
breaking!: Disable policies in config file, Disallow running more tha…

shimonp21:
…n one policy

- A config file is not the right place for a data file.
- This is a prerequisite to replacing "--output-dir" flag with "--output-file" flag (a "--output-file" flag
  can only work if every "policy run" command can only run a single policy.

Related to: https://github.com/cloudquery/cloudquery/issues/740

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/pkg/config/config_parser.go
+++ b/pkg/config/config_parser.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"github.com/cloudquery/cloudquery/internal/logging"
-	"github.com/cloudquery/cloudquery/pkg/policy"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
 	"github.com/spf13/viper"
@@ -45,6 +44,8 @@ func (p *Parser) decodeConfig(body hcl.Body, diags hcl.Diagnostics) (*Config, hc
 	content, contentDiags := body.Content(configFileSchema)
 	diags = append(diags, contentDiags...)
 
+	hasPolicyBlock := false
+
 	for _, block := range content.Blocks {
 		switch block.Type {
 		case "cloudquery":
@@ -60,11 +61,7 @@ func (p *Parser) decodeConfig(body hcl.Body, diags hcl.Diagnostics) (*Config, hc
 				config.Providers = append(config.Providers, cfg)
 			}
 		case "policy":
-			cfg, cfgDiags := policy.DecodePolicyBlock(block, &p.HCLContext)
-			diags = append(diags, cfgDiags...)
-			if cfg != nil {
-				config.Policies = append(config.Policies, cfg)
-			}
+			hasPolicyBlock = true
 		case "modules":
 			// Module manager will process this for us
 			config.Modules = block.Body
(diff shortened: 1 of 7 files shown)


bbernays:
Why not just have an array in the single output file? I am concerned about how users will run multiple policies in CI after this change. After this change they will have to run multiple commands in order to do something they could have done in a single command before

bbernays:
This is especially true when working with containerized CQ images. Prior to this change a user could have put all of the policies they wanted to run in the `config.hcl` and then just ran `cloudquery policy run` now they will have to have a separate job for each policy which just increases the amount of coordination/configuration they have to manage

bbernays:
Also why is the policy file any more of a data file than the provider binaries? Aren't they both just "executables" that generate data?

yevgenypats:
what the output policy should contain? list of results? single result?

amanenk (approved):
LGTM

yevgenypats:
Thought about this a bit more and given that we are going to transform it to YAML it will 1) help with the need to migrate this code
2) given the not a lot of policies that we have right now the work-around for that will be just running this twice but having a consistent behaviour  of what exactly is the output / where and so forth + less bugs. It doesn't look like anyone is running 5 policies in parallel so the worst case scenario is to run the command two times.

yevgenypats (approved):

shimonp21:
Added a deprecation warning + [link to docs] in case a policy is specified in the config.

disq:
- Making the result an array and enabling multiple policies to be specified in CLI could also bring order-related issues (or maybe not, as we wouldn't in parallel...)

- Removing the list from config seems OK as long as we ignore how we do things with fetch: Removing actually creates a disparity with fetch configuration (unless we also remove `resources[]` from config)

yevgenypats:
per our call I agree (as everyone) this has to go but Im running this locally to see what to advise users is a best practice to run policies and potentially will open another PR in docs. Will report back with docs and any additional bugs we need to take care of here.

pkg/config/config_parser.go
```diff
@@ -74,6 +71,17 @@ func (p *Parser) decodeConfig(body hcl.Body, diags hcl.Diagnostics) (*Config, hc
...
+
+	if hasPolicyBlock {
+		diags = append(diags,
+			&hcl.Diagnostic{
+				Severity: hcl.DiagWarning,
+				Summary:  "Deprecated 'policy' block in config file",
+				Detail:   "Specifying 'policy' blocks in 'config.hcl' has been deprecated. See https://docs.cloudquery.io/docs/tutorials/policies/writing-your-first-policy for instructions on running a local policy.",
+			},
+		)
+	}
```
roneli, line 83: why not append the diag in the switch case?

pkg/ui/console/client.go
```diff
@@ -478,7 +478,7 @@ func (c Client) DescribePolicies(ctx context.Context, policySource string) error
 
 func (c Client) ValidatePolicy(ctx context.Context, policySource string) (diags diag.Diagnostics) {
 	defer printDiagnostics("", &diags, viper.GetBool("redact-diags"), viper.GetBool("verbose"))
-	policyToValidate, err := FilterPolicies(policySource, c.cfg.Policies)
+	policyToValidate, err := FilterPolicies(policySource)
```
roneli, line 481: nit: not sure filter policies is required anymore, as it will always just wrap the policy source after detect, maybe the name is confusing.

roneli (commented):
LGTM, added a few comments

pkg/config/config_parser.go
shimonp21: Don't want 10 different diags in case the user has more then one policy block (one is enough to let them know of the deprecation)

pkg/ui/console/client.go
shimonp21: Not sure what to name it - It does "parse path" + "detect" and returns a policy struct.. 
"ParseAndDetct" maybe?

yevgenypats (approved):

roneli (approved):
LGTM

PR #862, opened 2022-06-07
feat: Improve policy analytics

roneli:

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/pkg/policy/policy.go
+++ b/pkg/policy/policy.go
@@ -4,9 +4,37 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"strings"
+
+	"github.com/cloudquery/cloudquery/internal/getter"
 )
 
 type Policies []*Policy
+
+func (pp Policies) All() []string {
+	policyNames := make([]string, len(pp))
+	for i, p := range pp {
+		policyNames[i] = p.Name
+	}
+	return policyNames
+}
+
+func (pp Policies) Get(policyName, subPath string) Policies {
+	for _, p := range pp {
+		if policyName != p.Name {
+			continue
+		}
+		if subPath != "" && p.SubPolicy() == "" {
+			return Policies{
+				&Policy{
+					Name:   p.Name,
+					Source: p.Source + "//" + subPath,
+				},
+			}
+		}
+		return Policies{p}
+	}
+	return nil
+}
 
 type Meta struct {
 	Type      string `json:"type,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/policy/policy.go
```diff
@@ -42,6 +44,21 @@ type Check struct {
 	Reason       string    `hcl:"reason,optional"`
 }
 
+type Analytic struct {
+	// Whether policy will persist in database
+	Persistence bool
+	// Whether policy was defined in configuration
+	UsedConfig bool
```
shimonp21, line 51: If we merge https://github.com/cloudquery/cloudquery/pull/841 , I think we can avoid the "UsedConfig" field. It will also simplify the logic of the `Get` function.

pkg/policy/policy.go
```diff
@@ -78,21 +95,50 @@ func (pp Policies) All() []string {
 	return policyNames
 }
 
-func (pp Policies) Properties() map[string]interface{} {
-	usedCustom := 0
-	policies := make([]*Meta, 0)
+func (pp Policies) Get(policyName, subPath string) Policies {
```
erezrokah, line 98: [non blocking] What do you think about returning a single Policy? Both places we call `Get` checks for `len > 0`. We could return a single policy pointer and check for `nil` instead.
Also the function always returns a single policy so it will better reflect what it does.

pkg/policy/policy.go
erezrokah, line 51: [non blocking] What are the other options to define policies? Via a CLI argument?

How about changing it to `UsageOrigin` or something like that? Then it can have multiple values.

Not this this is very relevant to feel free to ignore

pkg/policy/policy.go
```diff
@@ -78,21 +95,50 @@ func (pp Policies) All() []string {
...
+}
+
+func (p Policy) Analytic(dbPersistence, usedConfig bool) Analytic {
+	pa := Analytic{
+		Persistence: dbPersistence,
+		UsedConfig:  usedConfig,
+		Name:        p.Name,
+		Type:        p.SourceType(),
+		Selector:    p.SubPolicy(),
+		Private:     p.SourceType() != "hub",
```
erezrokah, line 127: [non blocking] Could we remove `Private`? Seems like `Type` has that information already

pkg/policy/policy_test.go
```diff
@@ -587,3 +588,61 @@ func TestFilterPolicies(t *testing.T) {
 		})
 	}
 }
+
+func TestPolicies_Get(t *testing.T) {
```
erezrokah, line 592: Great refactor 🧹

pkg/ui/console/client.go
```diff
@@ -400,9 +400,10 @@ func (c Client) RunPolicies(ctx context.Context, policySource, outputDir string,
 		DBPersistence: dbPersistence,
 	})
 	if resp == nil {
-		analytics.Capture("policy run", c.Providers, policiesToRun, diags)
-	} else {
-		analytics.Capture("policy run", c.Providers, resp.Policies, diags)
+		policiesToRun = resp.Policies
```
erezrokah, line 403: [blocking] I believe `resp` is `nil` here

erezrokah (changes requested):
Cool @roneli, great refactor. Left a few comments

pkg/policy/policy.go
roneli: Will change this

pkg/policy/policy.go
roneli: Yeah I can change the name no problem, good idea

pkg/policy/policy.go
roneli: Its more for analytics ease of use

pkg/ui/console/client.go
roneli: good catch meant `!=`

erezrokah (approved):

bbernays (approved):

PR #867, opened 2022-06-07
fix: Redact errors again

irmatov:


pkg/core/diagnostics.go
```diff
@@ -8,17 +8,28 @@ import (
...
-	Total      int            `json:"total,omitempty"`
-	ByType     map[string]int `json:"by_type,omitempty"`
-	BySeverity map[string]int `json:"by_severity,omitempty"`
+func (d SentryDiagnostic) Redacted() diag.Diagnostic {
+	v, ok := d.Diagnostic.(diag.Redactable)
+	if !ok {
+		return d
+	}
+	if r := v.Redacted(); r != nil {
+		return r
```
disq, line 30: Shouldn't this create and return a `SentryDiagnostic` to keep the "enrichments" in the redacted diag as well?

irmatov: Not sure, that's why I asked you for the review :)

It was hard enough for me to track the issue down across everything that happens around diags. I'lll update, thanks!

disq (approved):
LGTM

roneli (approved):

PR #874, opened 2022-06-08
feat: Sentry: Report provider version as release

disq:
Reports provider version as the release, if we're reporting a provider-generated diagnostic.


yevgenypats:
@disq can you test it locally that it will looks as we think it will look in sentry?

yevgenypats:
@irmatov multiple sentry project doesn't necessarly scale well with provider especially around when we want to configure alerts in every project

yevgenypats (approved):

PR #879, opened 2022-06-12
fix: Improve invalid provider version error

erezrokah:

#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/387 (internal issue)

To reproduce:
1. `cloudquery init aws`
2. `git clone https://github.com/cloudquery/cq-provider-aws`
3. `make build` in the repo ⬆️ 
4. Copy complied provider to `.cq/providers/cloudquery/aws` and replace the existing one.
5. Run `cloudquery fetch --no-verify`

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/core/sync.go
+++ b/pkg/core/sync.go
@@ -68,7 +68,7 @@ func Sync(ctx context.Context, sta *state.Client, pm *plugin.Manager, provider r
 	}
 
 	if want.ParsedVersion == nil {
-		return nil, diag.FromError(fmt.Errorf("failing provider with invalid version %q", provider.Version), diag.INTERNAL)
+		return nil, diag.FromError(fmt.Errorf("unsupported version %q for provider. If you're trying to debug a provider see https://docs.cloudquery.io/docs/developers/debugging", provider.Version), diag.USER)
 	}
 
 	cur, err := sta.GetProvider(ctx, provider)


pkg/core/sync.go
```diff
@@ -68,7 +68,7 @@ func Sync(ctx context.Context, sta *state.Client, pm *plugin.Manager, provider r
 	}
 
 	if want.ParsedVersion == nil {
-		return nil, diag.FromError(fmt.Errorf("failing provider with invalid version %q", provider.Version), diag.INTERNAL)
+		return nil, diag.FromError(fmt.Errorf("unsupported version %q for provider. If you're trying to debug a provider see https://docs.cloudquery.io/docs/developers/debugging", provider.Version), diag.USER)
```
roneli, line 71: The message is 100% a better improvement, but I am not sure its correct. In line 65 we check that if parsed version is nil (so if it returns in 71 it should be nil) and provider version is unmanaged so it should not be nil in that case. 

Would not that returning version in GetProviderSchema was introduced later so might be that edge case maybe?

roneli (commented):
LGTM, added a comment with some side thoughts about whether this solves the issue or not or just hides it under "user" type.

pkg/core/sync.go
erezrokah: >and provider version is unmanaged so it should not be nil in that case.

Agree this should not happen. I tested with unmanaged provider flow (with `CQ_REATTACH_PROVIDERS`) and couldn't reproduce.

The only way I could reproduce is by building the provider locally, overriding the official downloaded binary and running with `--no-verify` (without `CQ_REATTACH_PROVIDERS`).
Hence I'm assuming this is related somehow to a user trying to debug a provider by copying the binary under `.cq/providers`.

erezrokah: >Would not that returning version in GetProviderSchema was introduced later so might be that edge case maybe?

Also I'm not sure I understand your intention with this comment

roneli (approved):

PR #884, opened 2022-06-13
feat: Added ability to configure proxy via config file

### LINKED ISSUE #499: Ensure Support for HTTP Proxy

Many organizations require all traffic to go through specific proxies, we need to ensure that users are able to direct their traffic to these proxies

- Users should be able to define proxy settings via environment variables as well as the config.hcl
- Providers need to respect these settings as well

amanenk:

#### Summary

I have tested proxy configuration with cq-provider-aws cq-provider-gcp cq-provider-azure. All of them work with proxy setting environment variables so they can be used to set proxy config.  Added proxy block to hcl config. added setting of proxy env vars if they are exist in config.

closes #499 

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -28,15 +28,16 @@ type History struct {
 }
 
 type CloudQuery struct {
-	// Deprecated. Value from hcl is overrwritten.
+	// Deprecated. Value from hcl is overwritten.
 	PluginDirectory string `hcl:"plugin_directory,optional"`
-	// Deprecated. Value from hcl is overrwritten.
+	// Deprecated. Value from hcl is overwritten.
 	PolicyDirectory string `hcl:"policy_directory,optional"`
 
 	Logger     *logging.Config   `hcl:"logging,block"`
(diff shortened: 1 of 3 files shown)


erezrokah:
I might be missing something here, but if users can use env variables for configure a proxy, what's the use case for having it in our configuration file too? Won't using `HTTP_PROXY=<proxy> cloudquery fetch` work?

amanenk:
> I might be missing something here, but if users can use env variables for configure a proxy, what's the use case for having it in our configuration file too? Won't using `HTTP_PROXY=<proxy> cloudquery fetch` work?

The example you provided will work.
I forgot to link an issue for this PR. Ability to set proxy in config file was in requirements of the issue (#499)

bbernays:
If we are just setting the os environment variables, I am not sure that we should own that complexity in CQ rather than just verifying and documenting that all of our http clients respect those HTTP Variables

erezrokah:
@amanenk what do you think about closing this PR in favor of documenting the env variables support?

PR #887, opened 2022-06-13
feat!: Remove drift

disq:
Continued from https://github.com/cloudquery/cloudquery/pull/752

- [x] `cq init` should generate yaml
- [x] after config parse, actions must be taken to set/reinit values from config
- [x] make sure json-schema passes
- [x] all providers will need to change, as example configs are hcl
- [x] sdk needs to be updated (`GetProviderConfig`)
- [ ] ~policies will need updating/converting~
- [x] `cq init` should get yaml config from providers and correctly inject it?


bbernays:
I don't think we should be removing `hcl` support in the same PR we are adding yaml support. This would allow users to transition over at their own pace, while also lowering the testing bar for us as we won't be breaking any functionality

disq:
> I don't think we should be removing `hcl` support in the same PR we are adding yaml support. This would allow users to transition over at their own pace, while also lowering the testing bar for us as we won't be breaking any functionality

Why not introduce a converter? Otherwise the code gets very complicated (especially provider config logic in the SDK)

yevgenypats:
I also not sure about having two code bases and bug types to support - what is the value of customers upgrading at their own pace? eventually we will have to deprecate anyway. I think converter might be a better take and even that I think it should be a big deal to run `cloudquery init aws` and convert manually - I think given in terms of where we are in terms of users this should be ok to introduce such breaking change as long as we have good documentation, explanation and it's a major version etc...

yevgenypats:
Nice! Going to run locally and play with so I can give a more thorough review but one nit in the meanwhile. I think given it will be a breaking change the default should be the yaml generator and if someone want to use the old hcl format he will need to use the flag. This way more users will migrate faster and we can deprecate hcl faster.

yevgenypats (approved):

pkg/config/config_parser.go
```diff
@@ -1,47 +1,160 @@
...
+			Type: "cloudquery",
+		},
+		{
+			Type:       "provider",
+			LabelNames: []string{"name"},
+		},
+		{
+			Type:       "policy",
+			LabelNames: []string{"name"},
+		},
```
roneli, line 39: don't think we allow policy blocks in config anymore.

pkg/config/config_parser.go
```diff
@@ -1,47 +1,160 @@
...
+		CloudQuery: yc.CloudQuery,
+	}
+	for k := range yc.Providers {
+		v := yc.Providers[k]
+		v.Name = k
+		c.Providers = append(c.Providers, v)
+	}
+
+	diags := diag.Diagnostics{}
+	diags = diags.Add(ValidateCQBlock(&c.CloudQuery))
```
roneli, line 144:
```suggestion
	diags := ValidateCQBlock(&c.CloudQuery)
```

roneli (approved):
Looks good, two comments :)

pkg/config/config_parser.go
disq: Yes we don't but there's a deprecation message. This was moved from another file, so when the PR to really remove it lands, this line can also be removed by that.

PR #891, opened 2022-06-14
fix: Classify "failed to open file" as USER diags

shimonp21:

#### Summary

fixes: https://github.com/cloudquery/cloudquery-issues/issues/378

---

- [ ] Update or add tests 🧪

### DIFF
--- a/pkg/policy/execute.go
+++ b/pkg/policy/execute.go
@@ -415,11 +415,15 @@ func (e *Executor) deleteViews(ctx context.Context, policy *Policy) {
 	}
 }
 
-func GenerateExecutionResultFile(result *ExecutionResult, outputDir string) error {
+func GenerateExecutionResultFile(result *ExecutionResult, outputDir string) diag.Diagnostics {
 	fs := afero.NewOsFs()
 
 	if err := fs.MkdirAll(outputDir, 0755); err != nil {
-		return err
+		return diag.FromError(
+			err,
+			diag.USER,
+			diag.WithDetails(fmt.Sprintf("failed to create directory %q", outputDir)),
+		)
 	}
 
 	// result.PolicyName is the full selector for this policy run.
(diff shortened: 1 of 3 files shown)


pkg/policy/execute_test.go
```diff
@@ -712,8 +712,9 @@ func TestGenerateExecutionResultFile(t *testing.T) {
 
 	for _, tc := range testCases {
 		executiionResult := ExecutionResult{PolicyName: tc.PolicyName}
-		err := GenerateExecutionResultFile(&executiionResult, tc.OutputDir)
-		if !assert.NoError(t, err) {
+		diags := GenerateExecutionResultFile(&executiionResult, tc.OutputDir)
+
+		if !assert.False(t, diags.HasErrors()) {
```
irmatov, line 717: nit: just use `assert.True` instead of `!assert.False` ?

shimonp21: I don't believe that's right :)
I want to assert that there are no errors (i.e. `assert.False(diags.HasErrors())`).
But if there are errors (i.e. `HasErrors` is `true`), I want to `continue` to the next testcase.
An assertion returns "true" if the condition specified is true, and false otherwise (i.e. it will return false if there are errors in diags)..

erezrokah: Will:
```go
if diags.HasErrors() {
	assert.Fail(t, "expected no errors, got %v", diags)
	continue
}
```

work?
That way we'll get the received errors in the message

shimonp21: done :)

pkg/policy/execute.go
```diff
@@ -415,36 +415,51 @@ func (e *Executor) deleteViews(ctx context.Context, policy *Policy) {
...
-	f, err := fs.Create(fmt.Sprintf("%s.json", filepath.Join(outputDir, basePolicyName)))
+	filePath := filepath.Join(outputDir, basePolicyName)
+
+	f, err := fs.Create(fmt.Sprintf("%s.json", filePath))
 	if err != nil {
-		return err
+		return diag.FromError(
+			err,
+			diag.USER,
+			diag.WithDetails(fmt.Sprintf("failed to create file \"%s\"", filePath)),
```
disq, line 444: nit: Use `%q` instead of `\"%s\"`
Also wouldn't the `err` itself contain the file and the "failed to create" msg already?

pkg/policy/execute.go
```diff
@@ -415,36 +415,51 @@ func (e *Executor) deleteViews(ctx context.Context, policy *Policy) {
...
 	if err != nil {
-		return err
+		return diag.FromError(err, diag.INTERNAL)
 	}
 	if _, err := f.Write(data); err != nil {
-		return err
+		return diag.FromError(
+			err,
+			diag.USER,
+			diag.WithDetails(fmt.Sprintf("failed to write to file \"%s\"", filePath)),
```
disq, line 459: Would use `%q` instead of `\"%s\"`

pkg/policy/execute.go
shimonp21: 1) Done.
2) Not sure what are all the possible error messages from the library so opted for providing more context just in case. No big feelings here.

roneli (approved):
LGTM

PR #855, opened 2022-06-06
fix: Remove deprecated 'policy_directory' and 'plugin_directory' from hcl file

shimonp21:
Closes: #843.

🎉 Thank you for making CloudQuery awesome by submitting a PR 🎉

#### Summary

---

Use the following steps to ensure your PR is ready to be reviewed

- [x] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `go fmt` to format your code 🖊
- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [ ] Update or add tests 🧪
- [ ] Ensure the status checks below are successful ✅

PR #898, opened 2022-06-15
fix: Don't parse plugin and policy dirs from HCL

erezrokah:

#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/855

ignores `plugin_directory` and `policy_directory` when parsing the configuration (also related to https://github.com/cloudquery/cloudquery-issues/issues/384 (internal issue)

---

- [x] Update or add tests 🧪

### DIFF
--- a/pkg/config/config_parser.go
+++ b/pkg/config/config_parser.go
@@ -127,6 +127,8 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery, 
 	cq.Logger = &logging.GlobalConfig
 	var diags hcl.Diagnostics
 	diags = diags.Extend(gohcl.DecodeBody(block.Body, ctx, &cq))
+	// We don't care about additional configuration blocks, so just ignore them
+	cq.Remain = nil
 
 	// TODO: decode in a more generic way
 	if cq.Connection == nil {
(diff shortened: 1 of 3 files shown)


erezrokah:
We could also consider removing `data-dir` but that flag works

pkg/config/config_parser.go
```diff
@@ -127,6 +127,8 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery,
 	cq.Logger = &logging.GlobalConfig
 	var diags hcl.Diagnostics
 	diags = diags.Extend(gohcl.DecodeBody(block.Body, ctx, &cq))
+	// We don't care about additional configuration blocks, so just ignore them
+	cq.Remain = nil
```
erezrokah, line 131: A possible improvement can be to print warnings on unexpected configuration blocks, migrating blocks that changed, etc.

erezrokah:
The concern with still parsing the values from the configuration (current code) is that we might use those values in our code and get unexpected results

erezrokah:
Closing as if we decide to ignore additional fields in the configuration we should have a mechanism to alert the user.
This should be done as a part of https://github.com/cloudquery/cloudquery-issues/issues/384 (internal issue)

Issue #908, opened 2022-06-16
feature: Change the default version of providers in `config.hcl` from `latest` to `~version-of-latest`


erezrokah:
When someone runs `cloudquery init aws` we generate the following configuration:

```hcl
cloudquery {
  provider "aws" {
    version = "latest"
  }
....
```

This means that if we break providers we automatically break all existing versions of CloudQuery that use the default setting.
A possible solution will be to use:

```hcl
cloudquery {
  provider "aws" {
    version = "~0.12.8"
  }
....
```

If we add range support, this will prevent us from breaking existing CloudQuery versions when breaking providers, while still ensuring updates to latest patch version.

This solution assumes the user is running on a recent version of CloudQuery that supports the latest provider during init. To support matching old CloudQuery versions to supported providers we'll need a different mechanism.

>We could even print a notice that a newer provider version exists, but say it's a breaking one and require an update to the core (and might also require other migrations like config changes, updating postgres, etc.)

erezrokah:
Should be resolved in [v2](https://github.com/cloudquery/cloudquery/pull/1463) by pinning the version

PR #912, opened 2022-06-19
refactor: Rename parser.go to loader.go, move loadRemoteFile

erezrokah:
🎉 Thank you for making CloudQuery awesome by submitting a PR 🎉

#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/908

We have 2 different files implementing methods for the `Parser` struct, one named `config_parser.go` and the other `parser.go`. The latter (`parser.go`) mostly loads the config (it does parse the HCL file into a block, but it's different from `config_parser.go` that returns a valid CloudQuery configuration).
This PR renames `parser.go` to `loader.go` for clarity.

Also, `pkg/config/load_config.go` had a single function `loadRemoteFile` used in `parser.go`, so I moved it to that file, along with the tests from `pkg/config/load_config_test.go`.

I'll add comments on specific changes for more clarity

---

Use the following steps to ensure your PR is ready to be reviewed

- [x] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `go fmt` to format your code 🖊
- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Update or add tests 🧪
- [x] Ensure the status checks below are successful ✅

PR #913, opened 2022-06-19
refactor: Split configuration parsing into steps

erezrokah:

#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/908
This refactors the code that parses the configuration into steps:
1. Assigning defaults
2. Validation - that is, verifying that the configuration supplied by the user is valid
3. Normalization - that is, transforming the configuration supplied by the user so it can be consumed by CloudQuery CLI

---

- [ ] Update or add tests 🧪


pkg/config/config.go
```diff
@@ -119,20 +118,13 @@ func (c Connection) IsAnyConnParamsSet() bool {
...
 
-func (c *Connection) BuildFromConnParams() error {
+func (c *Connection) BuildFromConnParams() {
 	if c.Port == 0 {
 		c.Port = 5432
 	}
 	if c.Type == "" {
 		c.Type = "postgres"
 	}
-	if c.Host == "" {
```
erezrokah, line 129: Moved this to validation step and added tests for it

pkg/config/config.go
```diff
@@ -119,20 +118,13 @@ func (c Connection) IsAnyConnParamsSet() bool {
...
-		return errors.New("missing host")
-	}
-	if c.Database == "" {
```
erezrokah, line 132: Moved this to validation step and added tests for it

pkg/config/config_parser.go
```diff
@@ -120,27 +119,14 @@ func ReadModuleConfigProfiles(module string, block hcl.Body) (map[string]hcl.Bod
...
-func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery, hcl.Diagnostics) {
-	var cq CloudQuery
-	// Pre-populate with existing values
-	cq.Logger = &logging.GlobalConfig
-	var diags hcl.Diagnostics
-	diags = diags.Extend(gohcl.DecodeBody(block.Body, ctx, &cq))
-
+func assignDefaults(cq *CloudQuery) {
 	// TODO: decode in a more generic way
 	if cq.Connection == nil {
```
erezrokah, line 124: Moved to assign defaults step

pkg/config/config_parser.go
```diff
@@ -151,7 +137,14 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery,
...
-	// validate provider versions
+	if ds := viper.GetString("dsn"); ds != "" {
+		cq.Connection.DSN = ds
+	}
+}
+
+func validate(block *hcl.Block, cq *CloudQuery) hcl.Diagnostics {
+	var diags hcl.Diagnostics
+
 	for _, cp := range cq.Providers {
```
erezrokah, line 148: Moved this to validation step

yevgenypats:
I like it but will it make it sense to do it post hcl deprecation? especially as we move to jsonschema validation

erezrokah:
> I like it but will it make it sense to do it post hcl deprecation? especially as we move to jsonschema validation

I think this will help us with the deprecation as we would only need to replace a single function `validate` in https://github.com/cloudquery/cloudquery/pull/913/files#diff-a985f21601efa99f89f2f497e2c36cfcb899000b335c36b09f830594ba49ad78R197

Definitely better to merge after https://github.com/cloudquery/cloudquery/pull/887 to make it easier to handle conflicts

erezrokah:
> I would also get rid of hcl.Range and hcl.Diags as much as possible as return values and completely avoid them in shared methods between multiple config formats.

Thanks for the tip. Good idea. I'll rebase all the config related PRs once we're done with YAML

erezrokah:
Will handle this after https://github.com/cloudquery/cloudquery/pull/912 is merged

pkg/config/config.go
```diff
@@ -200,19 +189,3 @@ func (r RequiredProviders) Get(name string) *RequiredProvider {
 	}
 	return nil
 }
-
-func handleConnectionConfig(c *Connection) error {
```
erezrokah, line 204: Split this code into validation (non mutating) and normalization (mutating)

pkg/config/config_parser.go
```diff
@@ -45,15 +45,20 @@ var configSchemaHCL = &hcl.BodySchema{
...
+	var config *Config
+	var diags diag.Diagnostics
 	if IsNameYAML(name) {
-		return decodeConfigYAML(bytes.NewBuffer(data))
+		config, diags = decodeConfigYAML(bytes.NewBuffer(data))
+	} else {
+		config, diags = p.decodeConfigHCL(name, data)
 	}
 
-	body, diags := p.LoadFromSource(name, data)
```
erezrokah, line 52: Inlined `LoadFromSource` and handling its result into `decodeConfigHCL`.

Also not sure where `body` can be `nil` as `LoadFromSource` always returns a non `nil` value

pkg/config/loader.go
```diff
@@ -110,17 +110,6 @@ func (p *Parser) LoadFile(path string) ([]byte, diag.Diagnostics) {
 	return contents, nil
 }
 
-func (p *Parser) LoadFromSource(name string, data []byte) (hcl.Body, diag.Diagnostics) {
```
erezrokah, line 113: Inlined into `decodeConfigHCL`

pkg/config/provider.go
```diff
@@ -53,27 +53,6 @@ func decodeProviderBlock(block *hcl.Block, ctx *hcl.EvalContext, existingProvide
 	if attr, exists := content.Attributes["alias"]; exists {
 		valDiags := gohcl.DecodeExpression(attr.Expr, ctx, &provider.Alias)
 		diags = append(diags, valDiags...)
-		if _, ok := existingProviders[provider.Alias]; ok {
```
erezrokah, line 56: Moved this into the validation step

pkg/config/config_parser.go
```diff
@@ -73,66 +78,47 @@ func IsNameYAML(name string) bool {
 	}
 }
 
-func ValidateCQBlock(cq *CloudQuery) diag.Diagnostics {
+func Validate(config *Config) diag.Diagnostics {
 	var diags diag.Diagnostics
 
-	if cq.Connection == nil {
```
erezrokah, line 79: Assigning `&Connection{}` was moved to `applyDefaults` step

pkg/config/config_parser.go
```diff
@@ -73,66 +78,47 @@ func IsNameYAML(name string) bool {
...
+
+func assignDefaults(config *Config) {
+	// TODO: decode in a more generic way
+	if config.CloudQuery.Connection == nil {
+		config.CloudQuery.Connection = &Connection{}
 	}
+}
 
+func overrideFromFlags(config *Config) {
 	datadir := viper.GetString("data-dir")
```
erezrokah, line 99: Moved to `overrideFromFlags` from flags step

pkg/config/config_parser.go
```diff
@@ -73,66 +78,47 @@ func IsNameYAML(name string) bool {
...
+		config.CloudQuery.PluginDirectory = filepath.Join(datadir, "providers")
 	}
 
 	if datadir != "" {
-		cq.PolicyDirectory = filepath.Join(datadir, "policies")
+		config.CloudQuery.PolicyDirectory = filepath.Join(datadir, "policies")
 	}
 
-	// validate provider versions
-	for _, cp := range cq.Providers {
```
erezrokah, line 97: Moved to validate step

pkg/config/config_parser.go
```diff
@@ -235,3 +217,84 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery,
...
+	}
+
+	return diags
+}
+
+func validateProvidersBlock(config *Config) diag.Diagnostics {
+	var diags diag.Diagnostics
+	existingProviders := make(map[string]bool, len(config.Providers))
+
+	for _, provider := range config.Providers {
```
erezrokah, line 257: This should also remove some code duplication we had

cmd/init.go
```diff
@@ -86,7 +86,7 @@ func initialize(ctx context.Context, providers []string) error {
 			},
 		},
 	}
-	if diags := config.ValidateCQBlock(&mainConfig.CloudQuery); diags.HasErrors() {
```
erezrokah, line 89: `ValidateCQBlock` did more stuff that validation (for example, setting the providers and policies download directories). This is done as a part of `HandleDecodedConfig` now

pkg/config/loader_test.go
```diff
@@ -39,10 +39,9 @@ provider "aws" {
   resources = ["slow_resource"]
 }`
 
-const expectedDuplicateProviderError = "fixtures/duplicate_provider_name.hcl:21,1-15: Provider Alias Required; Provider with name aws already exists, use alias in provider configuration block."
```
erezrokah, line 42: With `YAML` we don't get the line and column numbers

pkg/config/provider.go
```diff
@@ -43,37 +43,16 @@ var providerBlockSchema = &hcl.BodySchema{
...
 }
 
-func decodeProviderBlock(block *hcl.Block, ctx *hcl.EvalContext, existingProviders map[string]bool) (*Provider, hcl.Diagnostics) {
+func decodeProviderBlock(block *hcl.Block, ctx *hcl.EvalContext) (*Provider, hcl.Diagnostics) {
 	var diags hcl.Diagnostics
 
 	content, _, moreDiags := block.Body.PartialContent(providerBlockSchema)
 	diags = append(diags, moreDiags...)
 	name := block.Labels[0]
-	provider := &Provider{Name: name, Alias: name}
```
erezrokah, line 52: Setting the `Alias` is done during normalization

yevgenypats:
Im also working on this in https://github.com/cloudquery/cloudquery/pull/1003 basically our cobra-viper is a mess. (was waiting for the HCL to go away because it was impossible to do it when we had both). Once we put in place the viper stuff then the whole priority will start to make sense:
1) flags - highest priority
2) Environment variable
3) File

Viper also support reading from a file and we should use it because otherwise we introduce some kind of another out of bounds config.

erezrokah:
> * flags - highest priority
> * Environment variable
> * File

💯 agree with this priority.

I think our work can be decoupled and this PR might make your changes easier to implement.

The purpose of this PR is to ensure that once we finish reading the config (broken into steps: loading the file, parsing the format, validating the config and normalizing it), we won't have to mess with it later on in the code.
You can see example in https://github.com/cloudquery/cloudquery/pull/917 on how it moves handling setting the `latest` version to a single place in the code.

pkg/config/config_parser.go
```diff
@@ -169,28 +167,37 @@ func decodeConfigYAML(r io.Reader) (*Config, diag.Diagnostics) {
...
 	c := &Config{
 		CloudQuery: yc.CloudQuery,
+		Providers:  providers,
 		format:     cqproto.ConfigYAML,
 	}
 
+	return c, nil
+}
+
+func (p *Parser) decodeConfigHCL(name string, data []byte) (*Config, diag.Diagnostics) {
```
erezrokah, line 189: This function is designated to be removed https://github.com/cloudquery/cloudquery/pull/1044

pkg/config/config_parser.go
```diff
@@ -45,16 +45,21 @@ var configSchemaHCL = &hcl.BodySchema{
 }
 
 func (p *Parser) LoadConfigFromSource(name string, data []byte) (*Config, diag.Diagnostics) {
+	var config *Config
+	var diags diag.Diagnostics
 	if IsNameYAML(name) {
```
erezrokah, line 50: This `else` part of this condition should be removed after we drop HCL support

pkg/config/config_parser.go
```diff
@@ -236,3 +241,72 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery,
...
+		// allow using a DSN flag even if the config file has explicitly attributes (user, password, host, database, etc.)
+		dsnFromFlag := viper.GetString("dsn")
+		if dsnFromFlag == "" && connection.IsAnyConnParamsSet() {
+			diags = append(diags, diag.NewBaseError(
+				fmt.Errorf("invalid connection configuration"),
+				diag.USER, diag.WithOptionalSeverity(diag.ERROR),
+				diag.WithDetails("DSN specified along with explicit attributes, only one type is supported")),
+			)
+		}
+	} else {
```
roneli, line 257: this function can be simplified in the first if, its enough to just return if the first if occurs otherwise return nil, and then you don't need the else.

pkg/config/config_parser.go
```diff
@@ -236,3 +241,72 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery,
 	cq.Logger = &logging.GlobalConfig
 	return cq, gohcl.DecodeBody(block.Body, ctx, &cq)
 }
+
+func validateConnection(connection *Connection) diag.Diagnostics {
```
roneli, line 245: document this function

roneli, line 245: add tests to this function

pkg/config/config_parser.go
```diff
@@ -236,3 +241,72 @@ func decodeCloudQueryBlock(block *hcl.Block, ctx *hcl.EvalContext) (CloudQuery,
...
+				diag.USER, diag.WithOptionalSeverity(diag.ERROR),
+				diag.WithDetails("missing database")),
+			)
+		}
+	}
+
+	return diags
+}
+
+func validateProvidersBlock(config *Config) diag.Diagnostics {
```
roneli, line 277: document this function

roneli (changes requested):
Added some review comments, most importantly, would require @disq to review this as he removed the hcl and added the yaml so he has most expertise regarding this area of the code.

pkg/config/config_parser.go
erezrokah: This function already had some test in https://github.com/cloudquery/cloudquery/pull/913/files#diff-23627973acbb422484ff2af98b8a1dcd8485d1703b280cae1b041867ac44f32fR9, and I added some more in https://github.com/cloudquery/cloudquery/pull/913/files#diff-23627973acbb422484ff2af98b8a1dcd8485d1703b280cae1b041867ac44f32fR52

pkg/config/config_parser.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/913/commits/67837b2ad64d2cadb4358963cd78c6414a4fceb7

pkg/config/config_parser.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/913/commits/67837b2ad64d2cadb4358963cd78c6414a4fceb7

pkg/config/config_parser.go
erezrokah: Great catch! I would expect the linter to report it actually, not sure why it didn't.
Fixed in 6d4caed5dec28aeb754b12ed0e964aa4d1e92528

pkg/config/config_parser.go
```diff
@@ -32,66 +39,59 @@ func (p *Parser) LoadConfigFile(path string) (*Config, diag.Diagnostics) {
...
-func ValidateCQBlock(cq *CloudQuery) diag.Diagnostics {
+func HandleDecodedConfig(config *Config) diag.Diagnostics {
 	var diags diag.Diagnostics
-
-	if cq.Connection == nil {
-		cq.Connection = &Connection{}
+	assignDefaults(config)
+	overrideFromFlags(config)
+	diags = append(diags, validate(config)...)
+	if len(diags) == 0 {
```
disq, line 47:
```suggestion
	if diags.HasErrors() {
```

pkg/config/config_parser.go
```diff
@@ -32,66 +39,59 @@ func (p *Parser) LoadConfigFile(path string) (*Config, diag.Diagnostics) {
 	return p.LoadConfigFromSource(contents)
 }
 
-func ValidateCQBlock(cq *CloudQuery) diag.Diagnostics {
+func HandleDecodedConfig(config *Config) diag.Diagnostics {
```
disq, line 42: Not a great name (`Handle` is too generic), maybe document?

pkg/config/config_parser.go
```diff
@@ -32,66 +39,59 @@ func (p *Parser) LoadConfigFile(path string) (*Config, diag.Diagnostics) {
...
 
-	if err := handleConnectionConfig(cq.Connection); err != nil {
-		diags = diags.Add(diag.FromError(err, diag.USER, diag.WithSummary("Invalid DSN configuration")))
+	return diags
+}
+
+func validate(config *Config) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	diags = append(diags, validateCloudQueryProviders(config.CloudQuery.Providers)...)
```
disq, line 57: `diags.Add` could also work, would enable to omit the `...`

disq (changes requested):

pkg/config/config_parser.go
erezrokah: Good point. Improved it a bit in https://github.com/cloudquery/cloudquery/pull/913/commits/9850e6950466a4015d58386851e2cf233081cb68#diff-a985f21601efa99f89f2f497e2c36cfcb899000b335c36b09f830594ba49ad78R42.

disq (approved):

pkg/config/config_parser.go
```diff
@@ -32,66 +39,64 @@ func (p *Parser) LoadConfigFile(path string) (*Config, diag.Diagnostics) {
...
 	var diags diag.Diagnostics
 
-	if cq.Connection == nil {
-		cq.Connection = &Connection{}
-	}
+	diags = diags.Add(validateCloudQueryProviders(config.CloudQuery.Providers))
+	diags = diags.Add(validateConnection(config.CloudQuery.Connection))
+	diags = diags.Add(validateProvidersBlock(config))
+
+	return diags
```
roneli, line 66:
```suggestion
	return diags.Add(validateProvidersBlock(config))
```

roneli (approved):
LGTM, 1 suggestion

pkg/config/config_parser.go
disq: @roneli how about
```go
return diag.Diagnostics{}.Add(validateCloudQueryProviders(config.CloudQuery.Providers)).Add(validateConnection(config.CloudQuery.Connection)).Add(validateProvidersBlock(config))
```
😝

erezrokah: Ah, I did that intentionally to make it easier to add more validations functions before/after each function, but doesn't really matter now

erezrokah: It's all the same to me expect that having it line by line makes it easier to debug

PR #915, opened 2022-06-19
fix(config): Align provider version validation

erezrokah:

#### Summary

Builds on top of https://github.com/cloudquery/cloudquery/pull/913.
Correct diff available in https://github.com/erezrokah/cloudquery/compare/refactor/config_parsing...erezrokah:fix/version_validation?expand=1

When someone runs `init <provider>@<version>`, `version` can be `latest` or a valid loose Semver version (e.g. `0.10`, `v0.10`, etc.).
However the logic to validate the configuration file is a bit different and we allow `latest` and strings starting with `v`.
This PR align the logic so validating the configuration is loose SemVer compliant.

---

- [x] Update or add tests 🧪

### DIFF
--- a/cmd/init/init.go
+++ b/cmd/init/init.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/cloudquery/cloudquery/cmd/utils"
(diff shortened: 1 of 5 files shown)


roneli (approved):
LGTM

PR #916, opened 2022-06-19
feat!: Remove fail-on-error

yevgenypats:
Remove flag that should be the default behaviour.
If someone wants to override it he should use bash "|| true" mechanism.

This should also help remove bloat of flags that we have right now.

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -65,41 +65,8 @@ jobs:
       - name: Build
         run: go build -v .
 
-      - name: Fetch
-        run: go run ./main.go fetch --config=internal/test/test_config.hcl --enable-console-log --fail-on-error
-
-
-      - name: Fetch Multiple Provider
-        run: go run ./main.go fetch --config=internal/test/test_double_provider_config.hcl --enable-console-log --fail-on-error
-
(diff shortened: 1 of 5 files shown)


disq (approved):

erezrokah (approved):
Nice 👍 I suggest merging with `feat!` as this is a breaking change that can cause existing flows to fail unexpectedly (if they update to latest patch)

erezrokah:
Also will require an update to https://docs.cloudquery.io/docs/cli/commands/fetch#usage

yevgenypats:
> Also will require an update to https://docs.cloudquery.io/docs/cli/commands/fetch#usage

I think we should remove this part from the docs as it mostly auto generated and it get stale quickly and just adds burden. so users should just address the `--help` as well as all our tutorials that we should of-course maintain. wdyt ?

erezrokah:
> I think we should remove this part from the docs as it mostly auto generated and it get stale quickly and just adds burden. so users should just address the `--help` as well as all our tutorials that we should of-course maintain. wdyt ?

How about we move this discussion to https://github.com/cloudquery/docs/issues/197?

PR #848, opened 2022-06-06
test(integration): Add snapshot tests

erezrokah:
🎉 Thank you for making CloudQuery awesome by submitting a PR 🎉

#### Summary

Trying a new testing approach to catch issues like https://github.com/cloudquery/cloudquery/pull/830

---

Use the following steps to ensure your PR is ready to be reviewed

- [x] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `go fmt` to format your code 🖊
- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Update or add tests 🧪
- [x] Ensure the status checks below are successful ✅

PR #918, opened 2022-06-19
feat: Remove handle_command wrapper function

yevgenypats:
#### Summary

This should keep all logic of each command in one place and make code easier to read, maintain and debug.

Also moved to use `RunE` in cobra and using `cmd.Context`.

There is some repeat your self in `provider.go` but this cmd should be refactored anyway. other then that it should be the same amount of code as in command.go there was a `switch` for every command. 

There is still some more testing to do here on my end but should be read for first review

---

- [ ] Update or add tests 🧪


yevgenypats:
Also looks like this uncovered bugs for our sanity tests that were returning always true as we just used `Run` instead of `RunE` - https://github.com/cloudquery/cloudquery/runs/6955807347?check_suite_focus=true

.github/workflows/sanity.yml
```diff
@@ -78,8 +78,9 @@ jobs:
       - name: Policy Describe
         run: go run ./main.go policy describe k8s//nsa_cisa_v1/pod_security --config=internal/test/test_config.hcl
 
-      - name: Policy Download
-        run: go run ./main.go policy download github.com/roneli/aws//cis_v1.2.0 --config=internal/test/test_config.hcl
+      # TODO: Fix this in later PR
```
erezrokah, line 81: We might be able to use the approach in https://github.com/cloudquery/cloudquery/pull/848/files#diff-3fb3b260bf8d8453a9e3eed595af8d91e1f28d5c5681743103007b1d60c96a51R48 as an alternative

erezrokah (approved):
Looks good and I always like a good 🧹 

I can also run a manual test when it's ready

.github/workflows/sanity.yml
yevgenypats: Yeah potentially - it's another issue but I don't think this should be defined in the GH workflow. If we use this appraoch it should be defined under `scripts/sanity.sh` and then also added to make file so both the GH action  and the developer can run this in the same way.

yevgenypats:
@erezrokah This is ready for manual test (Im also running this manually and giving it some more thought to make sure I didn't miss any edge case and thinking a bit more about the sentry initialization place)

Also opened this bug - https://github.com/cloudquery/cloudquery/issues/921

cmd/command.go
```diff
@@ -1,93 +0,0 @@
...
-	}
-
-	if telemetryMsg && analytics.Enabled() {
-		ui.ColorizedOutput(ui.ColorInfo, "Anonymous telemetry collection and crash reporting enabled. Run with --no-telemetry to disable, or check docs at https://docs.cloudquery.io/docs/cli/telemetry\n")
-		if ui.IsTerminal() {
-			if err := helpers.Sleep(ctx, 2*time.Second); err != nil {
-				return err
-			}
-		}
-	}
```
roneli, line 90: where is this called?

yevgenypats: Good catch. Now here - https://github.com/cloudquery/cloudquery/pull/918/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR91

yevgenypats:
Also potentially found another bug, where our architecture of cobra and viper bindings is incorrect. Viper is one way finding. i.e if a user pass an env variable it doesn't set the flag variable.

so this will not work

```
CQ_VERBOSE=true go run main.go init aws --enable-console-log
```

because we reference this https://github.com/cloudquery/cloudquery/blob/main/cmd/root.go#L108

In theory it is suggested to just use viper thing but we can we can try two way binding as in this example: https://github.com/carolynvs/stingoftheviper/blob/main/main.go#L104

I'll refactor the flags in this PR as well.

.github/workflows/sanity.yml
yevgenypats: Added this in this PR

disq (approved):
Looks good but will need updates after we merge #887 (`configPath`)

PR #968, opened 2022-06-21
fix: Request correct config format (YAML) from provider

disq:
Otherwise ConfigureProvider is called with no format, which would default to HCL.

init/config providers format also moved from map to list of providers.

### DIFF
--- a/pkg/config/config_parser.go
+++ b/pkg/config/config_parser.go
@@ -102,10 +102,43 @@ func ValidateCQBlock(cq *CloudQuery) diag.Diagnostics {
 	return diags
 }
 
+func ProcessValidateProviderBlock(plist []*Provider) (Providers, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	existingProviders := make(map[string]struct{}, len(plist))
+
+	var ret Providers
+
+	for _, v := range plist {
+		if v.Alias != "" {
+			if _, ok := existingProviders[v.Alias]; ok {
+				diags = diags.Add(diag.FromError(fmt.Errorf("provider with alias %s for provider %s already exists, give it a different alias", v.Alias, v.Name), diag.USER, diag.WithSummary("Duplicate Alias")))
+				continue
+			}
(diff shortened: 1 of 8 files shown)


erezrokah (approved):

pkg/config/config_parser.go
```diff
@@ -102,10 +102,43 @@ func ValidateCQBlock(cq *CloudQuery) diag.Diagnostics {
...
+				diags = diags.Add(diag.FromError(fmt.Errorf("provider with name %s already exists, use alias in provider configuration block", v.Name), diag.USER, diag.WithSummary("Provider Alias Required")))
+				continue
+			}
+			existingProviders[v.Name] = struct{}{}
+			v.Alias = v.Name
+		}
+		var err error
+		v.Configuration, err = yaml.Marshal(v.ConfigKeys)
+		if err != nil {
+			diags = diags.Add(diag.FromError(err, diag.INTERNAL, diag.WithSummary("ConfigKeys marshal failed")))
```
erezrokah, line 129: [non-blocking] Should this be `diag.User` too?

disq: I don't think so, it would mean yaml is acting weirdly or we changed ConfigKeys type to be something else etc? or there's something inside the interface{} that we don't handle but we'd ideally like to?

erezrokah: Ah cool, thanks for explaining. Lets keep it this way and see what gets reported

roneli (approved):
LGTM. I guess, maybe, add a test to the parser?

erezrokah:
> add a test to the parser?

I can follow up with another PR with tests as I added some in https://github.com/cloudquery/cloudquery/pull/969/files#diff-0f1a4fd4caa8778f443d8cb2b88e52020357ceac4ba0302c6d844aa784bf36e7R112

disq:
Yeah I'll merge this and we'll follow up with tests later

PR #985, opened 2022-06-22
fix: Disallow running of db-using policy operations without config

disq:
`cloudquery policy run` shouldn't allow running without db (panics on c.Storage access later on)


erezrokah (approved):

yevgenypats (approved):

irmatov:
`policy validate` willl panic too

PR #986, opened 2022-06-23
fix: Panic in policy commands

irmatov:
If there is no config file then all policy commands will panic on
database access.

### DIFF
--- a/cmd/policy.go
+++ b/cmd/policy.go
@@ -103,7 +103,7 @@ var (
   # See https://hub.cloudquery.io for additional policies.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			source := args[0]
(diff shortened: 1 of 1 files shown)


erezrokah:
Thanks @irmatov I think https://github.com/cloudquery/cloudquery/pull/985 covers this, WDYT?

cmd/policy.go
```diff
@@ -103,7 +103,7 @@ var (
   # See https://hub.cloudquery.io for additional policies.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			source := args[0]
-			c, err := console.CreateClient(cmd.Context(), getConfigFile(), true, nil, instanceId)
```
irmatov, line 106: what is the use for `allowDefaultConfig` argument? should we remove it completely (assume `false`)?

irmatov:
@erezrokah thanks for the heads up.

PR #992, opened 2022-06-23
fix: Show identifiers values correctly

irmatov:
In policy run output identifer values were not shown correctly because
`cast.ToStringSlice(row.Identifiers)` cannot handle
`map[string]interface{}` type and always returns empty slice.


roneli (approved):
Good catch maybe this fixes the table panic with the index in footer.

irmatov:
@roneli no, it shouldn't fix the panic.

Not sure exactly how the panic happens (one of the data rows should be larger than len(columns) and, thus, the footer), but the workaround for that will be in follow up PR.

PR #993, opened 2022-06-23
chore!: Remove support for cq init HCL

disq:
- [x] `cq init --config config.hcl` should fail
- [x] `config.hcl` should be a fallback only if `config.yml` is not found.


erezrokah (approved):
Looks good. We should merge with `chore!:` to trigger a breaking change release

PR #1029, opened 2022-06-27
chore: Release 0.29.0

erezrokah:
Release-As: 0.29.0


erezrokah:
Doing this PR since https://github.com/cloudquery/cloudquery/pull/993 was merged not as a breaking change

PR #1030, opened 2022-06-27
chore: Rename default config to be cloudquery.yml

disq:


yevgenypats (approved):

erezrokah:
> is it a "chore" or a "chore!"? no one knows

should be `chore!` for a breaking change. `chore` won't generate a release/changelog

PR #1037, opened 2022-06-29
fix: Update 'policy run' help message

shimonp21:
#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/cmd/policy.go
+++ b/cmd/policy.go
@@ -78,7 +78,7 @@ var (
 	cloudquery policy run github.com/<repo-owner>/<repo-name>
 
 	# Run a policy from a local directory
-	cloudquery policy run ./PATH_TO_POLICY_DIRECTORY/
+	cloudquery policy run ./<path-to-local-directory>
 
 	# See https://hub.cloudquery.io for additional policies.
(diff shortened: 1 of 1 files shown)


cmd/policy.go
```diff
@@ -63,25 +63,25 @@ var (
...
-  # Run community policy
-  cloudquery policy run github.com/COMMUNITY_GITHUB_ORG/aws
+	# Run community policy
+	cloudquery policy run github.com/COMMUNITY_GITHUB_ORG/POLICY_NAME
 
-  # See https://hub.cloudquery.io for additional policies.`,
+	# Run a policy from a local directory
+	cloudquery policy run ./PATH_TO_POLICY_DIRECTORY/
+
+	# See https://hub.cloudquery.io for additional policies.
```
erezrokah, line 83:
```suggestion
	# See https://hub.cloudquery.io for additional policies
```

Also same comment about clickable links

erezrokah (approved):
Thanks for the PR @shimonp21 👍 

I added a few comments, all non blocking in my opinion:
1. I think it's common to use `<param>` to denote a value supplied by the user
2. I think we should either reference the Hub or GitHub, but not both

erezrokah (approved):

PR #1048, opened 2022-06-30
fix: Use SDK GetULimit

erezrokah:
#### Summary

---

- [x] Update or add tests 🧪

### DIFF
--- a/cmd/ulimit_unix.go
+++ b/cmd/ulimit_unix.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"syscall"
 
+	"github.com/cloudquery/cq-provider-sdk/helpers/limit"
 	zerolog "github.com/rs/zerolog/log"
 )
 
@@ -24,7 +25,7 @@ func checkAndSetUlimitUnix() {
 
 func setUlimit(ulimit uint64) error {
 	logger := zerolog.Logger
-	rLimit, err := getUlimit()
+	rLimit, err := limit.GetUlimit()
 	if err != nil {
 		return fmt.Errorf("error getting ulimit: %w", err)
 	}
@@ -41,11 +42,5 @@ func setUlimit(ulimit uint64) error {
 		rLimit.Cur = ulimit
 	}
 
-	return syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
-}
-
-func getUlimit() (syscall.Rlimit, error) {
-	var rLimit syscall.Rlimit
-	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
-	return rLimit, err
+	return syscall.Setrlimit(syscall.RLIMIT_NOFILE, &syscall.Rlimit{Cur: rLimit.Cur, Max: rLimit.Max})
 }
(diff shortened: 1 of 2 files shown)


cmd/ulimit_unix.go
```diff
@@ -44,7 +44,7 @@ func setUlimit(ulimit uint64) error {
 	return syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
 }
 
-func GetUlimit() (syscall.Rlimit, error) {
+func getUlimit() (syscall.Rlimit, error) {
```
erezrokah, line 47: No need for this to be public, and I think we should use the SDKs `limit.GetUlimit()` but will leave that to another PR

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/1048/commits/4ad7eb7f87dd26a0fece3f27b5c24d7339c3611e

cmd/ulimit_unix_test.go
```diff
@@ -0,0 +1,35 @@
...
+package cmd
+
+import (
+	"syscall"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_setUlimit(t *testing.T) {
```
erezrokah, line 12: Very stupid test, probably can be improved by running on MacOS and also setting ulimit before and ensuring it was changed later

cmd/ulimit_unix_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_setUlimit_when_low(t *testing.T) {
+	// set to low values
+	err := syscall.Setrlimit(syscall.RLIMIT_NOFILE, &syscall.Rlimit{Cur: 10, Max: 20})
+	require.NoError(t, err)
+
+	err = setUlimit(ulimitUnix)
```
erezrokah, line 17: Setting the limit the second time fails

cmd/ulimit_unix.go
```diff
@@ -41,11 +42,5 @@ func setUlimit(ulimit uint64) error {
...
 	}
 
-	return syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
-}
-
-func GetUlimit() (syscall.Rlimit, error) {
-	var rLimit syscall.Rlimit
-	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
-	return rLimit, err
+	return syscall.Setrlimit(syscall.RLIMIT_NOFILE, &syscall.Rlimit{Cur: rLimit.Cur, Max: rLimit.Max})
```
roneli, line 45: maybe this should be moved to the SDK. Also removing the need to seperate ulimit files, and where we call it in root just do limit.Set etc' and in windows it will do nothing etc'.

roneli (changes requested):
Looks good, suggested to maybe completely remove the low level ulimit calls here and move them 100% to SDK also reducing the ulimit_<os> bloat.

cmd/ulimit_unix.go
erezrokah: Makes sense, I was able to make the test fail so looking into it

erezrokah:
Per an internal discussion @shimonp21 will open a separate PR to handle setting the ulimit as we can only increase the hard limit when running in `root` permissions

PR #1050, opened 2022-06-30
feat: Send telemetry about failed COPY FROMs

irmatov:
Closes cloudquery/cloudquery-issues#374

### DIFF
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -58,8 +58,6 @@ func NewCmdFetch() *cobra.Command {
 				for _, p := range result.ProviderFetchSummary {
 					analytics.Capture("fetch", c.Providers, p, diags, "fetch_id", result.FetchId)
 				}
-			}
-			if result != nil {
 				sendProviderTelemetryEvents(c.Providers, result)
 			}
 			if diags.HasErrors() {
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -26,17 +26,15 @@ const (
 )
 
 // sendProviderTelemetryEvents sends all collected telemetry events from the provider fetch response.
+// It will panic if fr argument is nil.
 func sendProviderTelemetryEvents(providers registry.Providers, fr *core.FetchResponse) {
-	if fr == nil {
-		return
-	}
 	for _, e := range fr.TelemetryEvents {
-		analytics.Capture(e.Type, providers, e)
+		analytics.Capture(e.Category, providers, e, nil)
 	}
 	for _, pfs := range fr.ProviderFetchSummary {
 		for _, rfs := range pfs.FetchedResources {
 			for _, e := range rfs.TelemetryEvents {
-				analytics.Capture(e.Type, providers, e)
+				analytics.Capture(e.Category, providers, e, nil)
 			}
 		}
 	}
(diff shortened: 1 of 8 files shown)


irmatov:
Depends on new SDK with https://github.com/cloudquery/cq-provider-sdk/pull/395

internal/analytics/client.go
```diff
@@ -179,9 +177,7 @@ func Capture(eventType string, providers registry.Providers, data Message, diags
 		"env":                 c.env,
 		"instance_id":         c.instanceId,
 		"cookie_id":           c.cookieId,
-		"success":             !diags.HasErrors(),
 		"installed_providers": pp,
-		"diagnostics":         core.SummarizeDiagnostics(diags),
```
irmatov, line 184: these removals are needed to break the import cycle

erezrokah:
@irmatov looks like you need to update a test for CI to pass https://github.com/cloudquery/cloudquery/runs/7177084872?check_suite_focus=true#step:8:269

irmatov:
@erezrokah looks like a flaky test. it worked after rebase & push.

cmd/fetch/fetch.go
```diff
@@ -23,6 +25,23 @@ const (
 	cloudquery fetch`
 )
 
+// sendProviderTelemetryEvents sends all collected telemetry events from the provider fetch response.
+func sendProviderTelemetryEvents(providers registry.Providers, fr *core.FetchResponse) {
+	if fr == nil {
```
disq, line 30: Since the only place this is called (L64) has a `if result != nil` block before it, you can move the call one line up to L63 and get rid of this `nil` check.

internal/analytics/client.go
```diff
@@ -162,7 +160,7 @@ func GetCookieId() uuid.UUID {
 	return id
 }
 
-func Capture(eventType string, providers registry.Providers, data Message, diags diag.Diagnostics, extra ...interface{}) {
+func Capture(eventType string, providers registry.Providers, data Message, extra ...interface{}) {
```
disq, line 163: Why not keep `diags` param as-is but pass in summarized versions? Repeating the `"diagnostics"` string (and remembering to set `success`) as `extra` is prone to errors (and forgetting). If you don't have any (`sendProviderTelemetryEvents`) could just pass `nil`.

internal/analytics/telemetry_events.go
```diff
@@ -0,0 +1,44 @@
+package analytics
+
+import "github.com/cloudquery/cq-provider-sdk/provider/diag"
+
+// TelemetryEvent holds data of a telemtetry event collected from a Diagnostic.
```
disq, line 5:
```suggestion
// TelemetryEvent holds data of a telemetry event collected from a Diagnostic.
```

cmd/fetch/fetch.go
irmatov: Updated as suggested.

internal/analytics/client.go
irmatov: Reverted

internal/analytics/client.go
irmatov: Can't pass summarized version as it still would introduce `core` import into the analytics and create an import loop. Moved `SummarizeDiagnostics` into the analytics package.

internal/analytics/summary.go
```diff
@@ -0,0 +1,39 @@
+package analytics
```
irmatov, line 1: moved from core without any changes

pkg/core/diagnostics.go
```diff
@@ -40,32 +32,6 @@ func (d *SentryDiagnostic) IsSentryDiagnostic() (bool, map[string]string, bool)
 	return true, d.Tags, d.Ignore
 }
 
-func SummarizeDiagnostics(diags diag.Diagnostics) DiagnosticsSummary {
```
irmatov, line 43: moved into analytics package to break import cycle.

pkg/core/fetch.go
```diff
@@ -239,7 +243,12 @@ func Fetch(ctx context.Context, sta *state.Client, storage database.Storage, pm
 		}(providerInfo)
 	}
 	wg.Wait()
-	response := &FetchResponse{FetchId: fetchId, ProviderFetchSummary: make(map[string]*ProviderFetchSummary, len(opts.ProvidersInfo)), Duration: time.Since(start)}
+	response := &FetchResponse{
+		FetchId:              fetchId,
+		ProviderFetchSummary: make(map[string]*ProviderFetchSummary, len(opts.ProvidersInfo)),
+		Duration:             time.Since(start),
+		TelemetryEvents:      make([]analytics.TelemetryEvent, 0),
```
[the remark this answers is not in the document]
irmatov: Thanks, removed!

PR #1003, opened 2022-06-25
feat!: Refactor cobra

yevgenypats:
Stil WIP.

This huge PR introduce:
- Convention and standards for Cobra commands
- Conventions to flag naming
- Ability to test commands
- Remove of unnecessary abstraction layers
- Cobra + Viper fixes

---

Use the following steps to ensure your PR is ready to be reviewed

- [ ] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [ ] Run `go fmt` to format your code 🖊
- [ ] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [ ] Update or add tests 🧪
- [ ] Ensure the status checks below are successful ✅

Issue #952, opened 2022-06-20
Viper is not setup correctly with cobra


yevgenypats:
### Describe the Bug

Follow-up https://github.com/cloudquery/cloudquery/pull/918#issuecomment-1159811814

yevgenypats:
This is solved now in v2

Issue #858, opened 2022-06-07
feat(logging): Consolidate logging flags


erezrokah:
### Is your feature request related to a problem? Please describe.

We have 10 flags to configure logging:
`verbose, enable-console-log, encode-json, enable-file-logging, disable-log-color, log-directory, log-file, max-size, max-backups, max-age`

Some use `enable-*` notation, one uses `disable-*` notation, some prefix the flag with `log-*` and some don't have a prefix.
Also `enable-file-logging` use `logging` while others use `log`.

We should ensure flags have consistent naming and descriptions.

### Describe the solution you'd like

Suggestions:
- [ ] `verbose` - stays the same
- [ ] `enable-console-log` - stays the same, or rename to `log-format=console` (allowing multiple formats via additional flags)
- [ ] `enable-file-logging` - rename to `enable-file-log` or to `log-format=file` (allowing multiple formats via additional flags)
- [ ] `encode-json` - rename to `console-log-format` or `log-format=console-json` via https://github.com/cloudquery/cloudquery/issues/857
- [ ] `disable-log-color` - rename to `color` with options of `on,off,auto`. Default `auto` to detect if environment supports colors, cloud be done via https://github.com/cloudquery/cloudquery/issues/859
- [ ] `log-directory` - stays the same
- [ ] `log-file` - stays the same (maybe rename to `log-filename`)
- [ ] `max-size, max-backups, max-age` - prefix with `log`

### Describe alternatives you've considered

~~Move some of the options (like `max-size, max-backups, max-age`) to `config.hcl`, but that's tricky since then how do users will be able to control logging before they generate the config~~

Logging configuration can be set via the `config.yml` file but should be removed

### Additional context

Related to https://github.com/cloudquery/cloudquery/issues?q=is%3Aopen+is%3Aissue+label%3A%22theme%3A+logging%22.

We could do this in a non breaking way via `MarkDeprecated` and `SetNormalizeFunc` functions of the Cobra API

erezrokah:
Should be resolved in [v2](https://github.com/cloudquery/cloudquery/pull/1463)

PR #1057, opened 2022-06-30
refactor: Cobra restructure as pre-step to fixing cobra/viper

yevgenypats:
#### Summary

Related PRs and Issues:

https://github.com/cloudquery/cloudquery/pull/1003
https://github.com/cloudquery/cloudquery/issues/952
https://github.com/cloudquery/cloudquery/issues/858

This is just a pre-refactor where I just changed the structure and removed the use of `init` so our command will scale better and wont use globals. There are still alot of bugs that Im aware of and cobra/viper not working in many cases but I didn't want to do it all at once as otherwise it would be impossible to diff between bug fixes and just restructure.

---

- [ ] Update or add tests 🧪


cmd/utils/utils.go
```diff
@@ -0,0 +1,13 @@
+package utils
```
roneli, line 1: utils is a bad package name, it doesn't say what it includes, also this file includes two items that have nothing in common. I suggest maybe to put in root.go

cmd/provider/purge.go
```diff
@@ -0,0 +1,43 @@
...
+		Args:  cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c, err := console.CreateClient(cmd.Context(), utils.GetConfigFile(), false, nil, utils.InstanceId)
+			if err != nil {
+				return err
+			}
+			diags := c.RemoveStaleData(cmd.Context(), lastUpdate, dryRun, args)
+			errors.CaptureDiagnostics(diags, map[string]string{"command": "provider_purge"})
+			if diags.HasErrors() {
+				return fmt.Errorf("failed to remove stale data")
```
roneli, line 33: I know you didn't write this code, but if I am already looking at it, I think we should be consistent with our error value returned. Either with the `%w` or return the where more diags are in the logs.
```suggestion
			return fmt.Errorf("failed to remove stale data: %w", diags)
```
```suggestion
				return fmt.Errorf("failed to remove stale data")
```

roneli (approved):
LGTM, had two minor nits, personally not a fan of a package per command as it creates a bigger mental model,  and some commands are around 40 lines~ so feels weird to give them a file.  But I guess its easier to what belongs to what, and I assume after viper adding changing these files will grow.

cmd/doc.go
```diff
@@ -10,16 +10,24 @@ import (
...
+)
+
+func newCmdDoc() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "doc [directory_path]",
+		Short:  docShort,
+		Args:   cobra.ExactValidArgs(1),
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// This is no danger of infinite recursion here as it just goes through the docs
```
erezrokah, line 24:
```suggestion
			// There is no danger of infinite recursion here as it just goes through the docs
```

erezrokah (approved):
Looks good. ➕ For removing globals

cmd/utils/utils.go
yevgenypats: This should go away in a follow up PR as there will be no use of globals, right now it's not possible to put it in root because it will cause cyclic import.

Regarding the subpackages - the idea is to have a package per sub command so all policy related commands under `policy` and so on. especially as looks like we are growing `cloudquery` CLI interface.

PR #1061, opened 2022-07-01
fix: Ulimit bug

shimonp21:
Ulimit has a 'soft'(current) value and 'hard'(maximum) value. 'soft' can always be increased up to the 'hard' limit.
But, increasing the 'hard' limit requires some very arcane permissions (often even root is not enough).
Before this fix, we did just one 'setrlimit' call for increasing both 'hard' and 'soft' limit. But, since increasing
the 'hard' fails on EPERM, the soft limit is also not increased.. This PR fixes this by splitting the 'set hard limit' and 'set soft limit' into two different invocations.

#### Summary

---

- [ ] Update or add tests 🧪

### DIFF
--- a/cmd/ulimit_unix.go
+++ b/cmd/ulimit_unix.go
@@ -15,37 +15,87 @@ func init() {
 	fileDescriptorF = checkAndSetUlimitUnix
 }
 
+// In this entire file, by 'ulimit' we mean the "num open files ulimit" (RLIMIT_NOFILE)
 func checkAndSetUlimitUnix() {
 	logger := zerolog.Logger
-	if err := setUlimit(ulimitUnix); err != nil {
-		logger.Err(fmt.Errorf("error setting ulimit: %w", err))
-	}
-}
 
-func setUlimit(ulimit uint64) error {
-	logger := zerolog.Logger
-	rLimit, err := GetUlimit()
+	rlimit, err := getUlimit()
 	if err != nil {
-		return fmt.Errorf("error getting ulimit: %w", err)
+		logger.Err(err).Msg("checkAndSetUlimitUnix: failed getting ulimit")
+		return
+	}
+	logger.Info().Uint64("hard_ulimit", rlimit.Max).Uint64("soft_ulimit", rlimit.Cur).Msg("limits (before adjustment)")
+
+	if err := setUlimit(); err != nil {
+		logger.Err(err).Msg("failed setting ulimit")
 	}
 
-	logger.Info().Uint64("hard_ulimit", rLimit.Max).Uint64("soft_ulimit", rLimit.Cur).Msg("limits (before adjustment)")
-
-	if rLimit.Max < ulimit {
-		logger.Info().Uint64("previous_ulimit", rLimit.Max).Uint64("new_ulimit", ulimit).Msg("adjusting max ulimit")
-		rLimit.Max = ulimit
+	rlimit, err = getUlimit()
+	if err != nil {
+		logger.Err(err).Msg("checkAndSetUlimitUnix: failed getting ulimit")
+		return
 	}
-
-	if rLimit.Cur < ulimit {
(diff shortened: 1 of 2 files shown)


roneli (approved):
LGTM, would suggest though to move this logic to the SDK as that also has a limit package, would be cleaner to have everything in one place. Wdyt?

cmd/ulimit_unix.go
```diff
@@ -15,37 +15,87 @@ func init() {
...
 	}
+
+	return setSoftUlimit()
 }
 
-func setUlimit(ulimit uint64) error {
+func setHardUlimit() error {
 	logger := zerolog.Logger
-	rLimit, err := GetUlimit()
+	rLimit, err := getUlimit()
```
erezrokah, line 56: [non-blocking] Can we avoid this extra call to `getUlimit`? We just read it in `checkAndSetUlimitUnix`.
We could pass it via an argument

cmd/ulimit_unix.go
```diff
@@ -15,37 +15,87 @@ func init() {
...
+	return nil
+}
+
+func setSoftUlimit() error {
+	logger := zerolog.Logger
 
-	if rLimit.Max < ulimit {
-		logger.Info().Uint64("previous_ulimit", rLimit.Max).Uint64("new_ulimit", ulimit).Msg("adjusting max ulimit")
-		rLimit.Max = ulimit
+	rlimit, err := getUlimit()
```
erezrokah, line 73: [non-blocking] Similar comment about avoiding the extra call. We can have `setHardUlimit` return the new hard limit and use that value here, WDYT?

erezrokah (approved):
Thanks @shimonp21, this looks great 👍 Added a few non blocking comments to see if we can reduce the number of system calls (mostly thinking about the chance of those erroring out than performance). If we do keep those, maybe prefix each error with the name of the function so we can easily know which one caused the error, e.g:

```go
fmt.Errorf("checkAndSetUlimitUnix: error getting ulimit: %w", err)
fmt.Errorf("setHardUlimit: error getting ulimit: %w", err)
fmt.Errorf("setSoftUlimit: error getting ulimit: %w", err)
```

If we already print the function name when we get the error, you can ignore the latter comment

shimonp21:
I didn't move everything to the SDK because it's a bit of a headache (Different logging frameworks, different behaviours depending on platform...). I decided to just fix the bug in this PR.

cmd/ulimit_unix.go
shimonp21: Leaving as-is for now, but will add the logging.
[Only the syscall in setHardLimit can be avoided - the syscall in setSoftLimit needs to happen because it has no way of knowing if setHardLimit succeeded or not].
[Because only one of them can be avoided, I'd prefer to have a cleaner interface and consistency in code and behaviour between them]...

cmd/ulimit_unix.go
shimonp21: Leaving as-is for now, but will add the logging.
[Only the syscall in setHardLimit can be avoided - the syscall in setSoftLimit needs to happen because it has no way of knowing if setHardLimit succeeded or not].
[Because only one of them can be avoided, I'd prefer to have a cleaner interface and consistency in code and behaviour between them]...

cmd/ulimit_unix.go
```diff
@@ -15,37 +15,87 @@ func init() {
...
 }
 
-func GetUlimit() (syscall.Rlimit, error) {
+func getUlimit() (syscall.Rlimit, error) {
 	var rLimit syscall.Rlimit
 	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
 	return rLimit, err
 }
+
+// Golang doesn't have an out-of-the-box min function xd
```
disq, line 94: `math.Min(float64, float64) float64`

PR #1102, opened 2022-07-05
feat: Send connection issues to analytics

irmatov:
Closes https://github.com/cloudquery/cloudquery-issues/issues/370

### DIFF
--- a/cmd/fetch/fetch.go
+++ b/cmd/fetch/fetch.go
@@ -8,7 +8,9 @@ import (
 	"github.com/cloudquery/cloudquery/internal/analytics"
 	"github.com/cloudquery/cloudquery/pkg/config"
 	"github.com/cloudquery/cloudquery/pkg/errors"
+	"github.com/cloudquery/cloudquery/pkg/plugin/registry"
 	"github.com/cloudquery/cloudquery/pkg/ui/console"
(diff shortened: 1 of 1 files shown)


cmd/fetch/fetch.go
```diff
@@ -56,6 +59,33 @@ func NewCmdFetch() *cobra.Command {
 	return fetchCmd
 }
 
+func diagToEventType(d diag.Diagnostic) string {
+	if strings.HasSuffix(d.Error(), "i/o timeout") {
+		return "io_timeout"
+	}
+	return ""
+}
```
roneli, line 67: I can see this growing into many event types, we currently have fetch and policy, but in this case we can have more.
Maybe call this event type "diag" and give it sub types and some meta information as occurrence count etc' and by no means to pass the "error" itself as that can have user information we don't want.

ill let @bbernays  weigh in some more here.

roneli (changes requested):
Nice idea for adding more meta analytic information, we do need to update the docs after this PR merges. Added some thoughts here.

yevgenypats (changes requested):
Let's hold on with that as it uses Diags which are going away in the upcoming weeks - https://github.com/cloudquery/cq-provider-sdk/pull/437

yevgenypats:
Closing as we move to cloudquery v2 and error reporting is different

PR #1106, opened 2022-07-05
feat: Validate OOB config keys

disq:


pkg/config/config_parser.go
```diff
@@ -150,22 +154,43 @@ func decodeConfig(r io.Reader) (*Config, diag.Diagnostics) {
...
+	if len(extraKeys) == 0 {
+		return nil
+	}
+
+	keys := sort.StringSlice(funk.Keys(extraKeys).([]string))
+	keyOrKeys := "keys"
+	if len(keys) == 1 {
+		keyOrKeys = "key"
+	}
+	return diag.FromError(errors.New("Encountered unknown element in config"), diag.USER, diag.WithSummary("Invalid configuration"), diag.WithDetails("Found unknown %s: %s", keyOrKeys, strings.Join(keys, ", ")))
```
bbernays, line 193: I don't think that this should be an error. I think it should be a warning. That way a newer config file can work with an older cq binary

yevgenypats:
There is also `yaml.StrictUnmarshal` - can it work?

disq:
> There is also `yaml.StrictUnmarshal` - can it work?

There's no such thing in our version but we have `yaml.KnownFields`, but due to the way we decode the provider config (comes in at runtime so we need to decode it freely) it might not work. I'll give it a try.

disq:
> There is also `yaml.StrictUnmarshal` - can it work?

Nope, since yaml doesn't have RawMessage like json, it doesn't work holistically, so we would have two separate messages: one for everything except providers, and another for providers:

```
Configuration Error Diagnostics:
❌ Failed to parse yaml: yaml: unmarshal errors:
  line 13: field blah not found in type struct { CloudQuery config.CloudQuery "yaml:\"cloudquery\" json:\"cloudquery\""; Providers []*config.Provider "yaml:\"providers\" json:\"providers\"" }; 
Error: config not read
```

vs

```
Configuration Error Diagnostics:
❌ Invalid configuration: Encountered unknown element in config; Found unknown key: regions
Error: config not read
```

But it would catch all "extras" in all blocks.

pkg/config/config_parser.go
disq: Not sure, maybe. Maybe different error levels depending on the block they're on (allow for adding a new block (e.g. `modules:`) in top level but don't allow for an extra key inside the `provider` etc.)

pkg/config/config_parser.go
```diff
@@ -141,7 +144,7 @@ func decodeConfig(r io.Reader) (*Config, diag.Diagnostics) {
 		if len(errs) == 0 {
 			return nil, diag.FromError(errors.New("Failed to validate config with schema"), diag.USER, diag.WithSummary("Invalid configuration"))
 		}
-		diags := diag.Diagnostics{}
+		var diags diag.Diagnostics
```
roneli, line 145: good change, I think this would have returned `error: no errors`

pkg/config/config_parser.go
roneli: I think kemal is correct here, this should be an error as the "config" is invalid, although only because of a whitespace perhaps, but it might have unintentional run for example if I wanted region X and Y if we get only a warning CQ will run and fetch all regions, which is not the intended action

pkg/config/config_parser.go
```diff
@@ -150,22 +153,42 @@ func decodeConfig(r io.Reader) (*Config, diag.Diagnostics) {
...
+	var diags diag.Diagnostics
 	providers := yc.Providers
-	diags := diag.Diagnostics{}
 	for _, p := range providers {
 		p.Configuration, err = yaml.Marshal(p.ConfigKeys["configuration"])
 		if err != nil {
 			diags = diags.Add(diag.FromError(err, diag.INTERNAL, diag.WithSummary("ConfigKeys marshal failed")))
 			continue
 		}
+		delete(p.ConfigKeys, "configuration")
```
roneli: nit: document why you are deleting the key

roneli (commented):
Looks good, added a few comments

pkg/config/config_parser.go
disq: nil slice is valid (and empty) slice so both would give us `error: no error` if cast to `error`.

pkg/config/config_parser.go
disq: solved in new commit

roneli (approved):
LGTM

PR #1112, opened 2022-07-07
fix: Add schema option to connection config

### LINKED ISSUE #901: Specify Schema
### Describe the Bug

Right now it's not possible to specify schema when we define the database. It mostly uses the default `search_path` which in some situations can be not in the user control depending on which database it uses. 

Issue noted in GCP on Cloud SQL creating all tables in `cloudquery` schema instead of `public`

disq:
Fixes #901, needs https://github.com/cloudquery/cq-provider-sdk/pull/371 reverted (and a new SDK tag)

This allows to:
- Resume compatibility if `search_path` was being set in `extras`
- Otherwise, use the `schema` key for search path
- If none, defaults to `public`

`search_path` is not touched if `DSN` is being used directly.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -54,6 +54,7 @@ type Connection struct {
 	Port     int      `yaml:"port,omitempty" json:"port,omitempty"`
 	Database string   `yaml:"database,omitempty" json:"database,omitempty"`
 	SSLMode  string   `yaml:"sslmode,omitempty" json:"sslmode,omitempty"`
+	Schema   string   `yaml:"schema,omitempty" json:"schema,omitempty"`
 	Extras   []string `yaml:"extras,omitempty" json:"extras,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


shimonp21 (changes requested):
Is it possible to do this without reverting https://github.com/cloudquery/cq-provider-sdk/pull/371?

Reverting 371 means that for all users with 'latest' provider and 'cloudquery' username, all their tables will move *yet again*, And yet again their fetches will fail: https://github.com/cloudquery/cq-provider-azure/issues/376.

An ideal solution would only have a difference in hehaviour for users who specify the new config.

disq:
> Is it possible to do this without reverting [cloudquery/cq-provider-sdk#371](https://github.com/cloudquery/cq-provider-sdk/pull/371)?

I don't think it's possible unless we introduce a second, `postgres.Connect2` method in the SDK and switch all providers to use it.

yevgenypats (approved):

PR #1128, opened 2022-07-19
chore: Update SDK to v0.14.1

disq:
To go with https://github.com/cloudquery/cq-provider-sdk/pull/424
This isn't a breaking change because it's already not supported any more, we're just updating the SDK to remove it almost completely.

~~Tests won't pass until provider PRs are merged/released because now that the config format flag is gone from the SDK, we can't request yaml config format from the current providers.~~ using 0.14.1 to overcome this issue


cmd/init/init.go
```diff
@@ -141,9 +140,8 @@ func generateConfig(ctx context.Context, c *console.Client, providers []string,
 	for _, p := range providers {
 		pCfg, diags := core.GetProviderConfiguration(ctx, c.PluginManager, &core.GetProviderConfigOptions{
 			Provider: c.ConvertRequiredToRegistry(p),
-			Format:   cqproto.ConfigYAML,
 		})
-		if pCfg != nil && pCfg.Format != cqproto.ConfigYAML {
+		if pCfg != nil && pCfg.Format != 1 /* YAML */ {
```
disq, line 144: Not sure when we should remove this check

yevgenypats (approved):

PR #1178, opened 2022-08-09
chore: Add linting

erezrokah:
Adds linting to monorepo modules


cli/.golangci.yml
```diff
@@ -1,4 +1,3 @@
-# DONT EDIT. This file is synced from https://github.com/cloudquery/.github/.github
```
hermanschaaf, line 1: How do we decide, in cases like these, whether to keep only one global `.golangci.yml` or one per module?

.github/workflows/lint_golang.yml
```diff
@@ -1,34 +1,33 @@
...
 on:
   push:
     branches:
-      - main
+      - monorepo
   pull_request:
     branches:
-      - main
-
+      - monorepo
```
hermanschaaf, line 8: Shall we maybe keep these both `main` and `monorepo` for now, so it will still work when we do eventually make this the main branch? Then we can do some clean-up by removing `monorepo`, rather than having to pro-actively fix the build.

cli/.golangci.yml
erezrokah: Hmm, it's mostly based on the code requirements of the module. Plugins have some auto-generated code that might conflict with linting (we can also make the autogenerated code play nice with liniting).

Specifically in our case for CLI and plugins I think we can use a single file, and will follow up on it in this PR.

scripts/resolve-modules.sh
```diff
@@ -2,7 +2,8 @@
 # Recursively finds all directories with a go.mod file and creates
 # a GitHub Actions JSON output option. This is used by the linter action.
 
+# Based on https://github.com/golangci/golangci-lint/issues/828#issuecomment-658207652
 echo "Resolving modules in $(pwd)"
 
-PATHS=$(find . -mindepth 2 -type f -name go.mod -printf '{"workdir":"%h"},')
+PATHS=$(find . -mindepth 2 -type f -name go.mod -printf '{"workdir":"%h"},' | sed 's/\.\///g')
```
erezrokah, line 8: This removes the `./` prefix from the directories, makes it easier to use later on in the GitHub action

cli/.golangci.yml
erezrokah: I kept 1 file for the CLI and 1 file for the plugins as it got a bit confusing to have a single file for all (e.g. hard to understand why some directories are ignored and where they belong)

scripts/resolve-modules.sh
hermanschaaf: Nice 👍

hermanschaaf (approved):

PR #1186, opened 2022-08-10
fix: CQv2 misc fixes

disq:


cmd/fetch/fetch.go
```diff
@@ -64,6 +64,9 @@ func fetch(cmd *cobra.Command, args []string) error {
 
 func fetchConnection(ctx context.Context, specReader *specs.SpecReader, connSpec specs.ConnectionSpec, pm *plugin.PluginManager) error {
 	sourceSpec := specReader.GetSourceByName(connSpec.Source)
+	if sourceSpec.Name == "" {
```
yevgenypats, line 67: can we move this directly to unmarshal code? that this is required. This way we wont need to do this additional check everywhere in the code?

disq: this is a check to see if `connSpec.Source` was found in the specreader or not. moving it to some other place won't do any good. Other option is to make the `SpecReader.Get<T>ByName` methods return `<T>, bool` or `<T>, error`.

disq: What we could do is make sure it always finds the source, by checking every conn spec against the list of sources? That's more organic and would enable us to skip that check, but if it fails for some other reason, we'll still get the unclear errors.

yevgenypats:
I fixed some of it concurrently as we worked on the same stuff. will try to unblock as quickly as possible. closing for now.

PR #1187, opened 2022-08-10
chore: Rename modules and imports

erezrokah:
Update the modules names in `go.mod` files and relevant imports to match the monorepo structure


erezrokah:
Going to close this as we'll moving plugins under `plugins/source/*`

PR #1189, opened 2022-08-10
chore: Move plugins into source directory (monorepo)

hermanschaaf:


erezrokah (approved):

.github/workflows/check_generated_code_drift.yml
```diff
@@ -30,8 +30,14 @@ jobs:
         id: changed-files
         uses: tj-actions/changed-files@v25
         with:
+<<<<<<< HEAD
```
erezrokah, line 33: You might need to resolve this merge conflict first

hermanschaaf: Ah thanks, I missed that one

erezrokah (approved):

PR #1192, opened 2022-08-10
chore: Update release process to reference plugins/source/<name>

erezrokah:
To be merged after https://github.com/cloudquery/cloudquery/pull/1189


plugins/source/terraform/.gitignore
```diff
@@ -43,3 +43,4 @@ config.hcl
 cloudquery.yml
 terraform/tfplan.binary
 cq-provider*
+!examples/terraform.tfstate
```
erezrokah, line 46: Unrelated test fix

PR #1224, opened 2022-08-12
chore: Update CLI to work with monorepo release conventions

hermanschaaf:
- [x] Download versions from manifest files
 - [x] Download checksums from monorepo releases
 - [x] Download plugins from monorepo releases
 - [x] Fall back to provider repos for older versions

This keeps compatibility with the old release conventions, so that older versions of plugins can still be used with the new version of the CLI. We can also continue doing releases on the other repos until everything is in place here (should it be necessary - we'd just need to update the version manifests in the monorepo)

### DIFF
--- a/pkg/plugin/registry/hub.go
+++ b/pkg/plugin/registry/hub.go
@@ -3,8 +3,6 @@ package registry
 import (
 	"context"
 	"fmt"
-	"github.com/cloudquery/cloudquery/internal/versions"
-	"github.com/rs/zerolog"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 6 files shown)


yevgenypats (approved):
should it go into the monorepo branch ? 

Anyway, looks good! I wasn't too picky on the PR as this part will be partially migrated/re-written into cli v2.

pkg/plugin/registry/hub.go
```diff
@@ -206,6 +203,35 @@ func (h Hub) verifyProvider(ctx context.Context, provider Provider, version stri
...
+		tag := fmt.Sprintf("plugins/source/%s/%s", provider.Name, version)
+		checksumsURL := fmt.Sprintf("https://github.com/%s/cloudquery/releases/download/%s/%s", DefaultOrganization, tag, fileName)
+		l.Debug().Str("url", checksumsURL).Str("path", path).Msg("downloading checksums file from monorepo")
+		err := osFs.DownloadFile(ctx, path, checksumsURL, nil)
+		if err == nil {
+			return path, nil
+		}
+		// if we failed to download checksums from the monorepo, it might be because the plugin version
+		// was released on the older provider repo. Fall through to try there before giving up
+		fallthrough
```
erezrokah, line 223: Nice usage of `fallthrough` 🚀

erezrokah:
> should it go into the monorepo branch ?
> 
> Anyway, looks good! I wasn't too picky on the PR as this part will be partially migrated/re-written into cli v2.

I'll port this to the monorepo branch in a bit

PR #1277, opened 2022-08-15
feat: Add fsx filesystems

irmatov:
Closes https://github.com/cloudquery/cloudquery/issues/1375
Closes https://github.com/cloudquery/cloudquery/issues/1376


plugins/source/aws/resources/services/fsx/filesystems.go
```diff
@@ -0,0 +1,546 @@
...
+						Description: "The ID for an existing Amazon Web Services Managed Microsoft Active Directory instance that the file system is joined to.",
+						Type:        schema.TypeString,
+					},
+					{
+						Name:        "aliases",
+						Description: "An array of one or more DNS aliases that are currently associated with the Amazon FSx file system",
+						Type:        schema.TypeJSON,
+					},
+					{
+						Name:        "audit_log_configuration_file_access_audit_log_level",
```
erezrokah, line 420: What do you think about removing the `audit_log_configuration_` prefix?

erezrokah (approved):
Looks good to me, one non blocking comment

plugins/source/aws/resources/services/fsx/filesystems.go
irmatov: Mmm, good catch, thanks!

erezrokah (approved):

PR #1279, opened 2022-08-15
feat: Add fsx data repo tasks

irmatov:
Closes https://github.com/cloudquery/cloudquery/issues/1377


plugins/source/aws/resources/services/fsx/data_repo_tasks.hcl
```diff
@@ -0,0 +1,48 @@
...
+add_generate     = true
+
+description_modifier "remove_read_only" {
+  words = ["  This member is required."]
+}
+
+resource "aws" "fsx" "data_repo_tasks" {
+  path = "github.com/aws/aws-sdk-go-v2/service/fsx/types.DataRepositoryTask"
+  ignoreError "IgnoreAccessDenied" {
+    path = "github.com/cloudquery/cloudquery/plugins/source/aws/client.IgnoreAccessDeniedServiceDisabled"
```
hermanschaaf, line 13: Could this maybe be `IgnoreCommonErrors`?

plugins/source/aws/resources/services/fsx/data_repo_tasks.hcl
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  column "resource_arn" {
+    rename = "arn"
+  }
+
+  column "tags" {
+    type = "json"
+    generate_resolver = true
```
hermanschaaf, line 46: I'm not 100% sure it'll work in this case, but it seems like it should: did you consider using `client.ResolveTags` here? Like this:

```
    resolver "resolveTags" {
      path = "github.com/cloudquery/cloudquery/plugins/source/aws/client.ResolveTags"
    }
```

(instead of `generate_resolver = true`)

plugins/source/aws/resources/services/fsx/data_repo_tasks.go
```diff
@@ -0,0 +1,170 @@
...
+			return diag.WrapError(err)
+		}
+		res <- result.DataRepositoryTasks
+		if aws.ToString(result.NextToken) == "" {
+			break
+		}
+		input.NextToken = result.NextToken
+	}
+	return nil
+}
```
hermanschaaf, line 167: I think you can use `fsx.NewDescribeDataRepositoryTasksPaginator` here, which should make the code a little simpler since we then don't need to handle the `NextToken` here

plugins/source/aws/resources/services/fsx/data_repo_tasks.hcl
irmatov: Updated, thanks!

plugins/source/aws/resources/services/fsx/data_repo_tasks.hcl
irmatov: Yes, it works, thanks!

plugins/source/aws/resources/services/fsx/data_repo_tasks.go
irmatov: Great suggestion, thanks!

hermanschaaf (approved):

Issue #1221, opened 2022-08-12
Azure Service Principal Auth failing for accounts, subscriptions


hermanschaaf:
Reported by user on Discord:

```
failed to resolve table \"azure_account_locations\": error at github.com/cloudquery/cq-provider-azure/resources/services/account.fetchAccountLocations[locations.go:138] AzureCLICredential: Azure CLI not found on path
```

```
3:46AM INF calculated max goroutines for fetch execution @module=azure instance_id=9754c53e-496e-4dee-bad7-8c529b8ba269 max_goroutines=16384 timestamp=2022-08-12T03:46:32.992Z
3:46AM ERR received resolve resources error error="failed to resolve table \"azure_account_locations\": error at github.com/cloudquery/cq-provider-azure/resources/services/account.fetchAccountLocations[locations.go:138] AzureCLICredential: Azure CLI not found on path" @module=azure client_id=azure_account_locations:1 instance_id=9754c53e-496e-4dee-bad7-8c529b8ba269 subscription_id=aecab9f5-c68e-4837-85ba-2e4a04b2cf4c table=azure_account_locations timestamp=2022-08-12T03:46:33.004Z
3:46AM WRN received resource fetch diagnostics alias=azure instance_id=9754c53e-496e-4dee-bad7-8c529b8ba269 provider=azure resource=account.locations
3:46AM ERR received resolve resources error error="failed to resolve table \"azure_subscription_subscriptions\": error at github.com/cloudquery/cq-provider-azure/resources/services/subscription.fetchSubscriptionSubscriptions[subscriptions.go:99] AzureCLICredential: Azure CLI not found on path" @module=azure client_id=azure_subscription_subscriptions:1 instance_id=9754c53e-496e-4dee-bad7-8c529b8ba269 subscription_id=aecab9f5-c68e-4837-85ba-2e4a04b2cf4c table=azure_subscription_subscriptions timestamp=2022-08-12T03:46:33.006Z
```

My guess is that the problem is likely specific to Service Principal auth, and will need to be handled here: https://github.com/cloudquery/cq-provider-azure/blob/main/client/client.go#L72-L81

PR #1283, opened 2022-08-15
fix!: Fix Azure credential chain

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/1221

This changes the new Azure SDK to use the default credential chain, which steps through the authentication options in this order:
 1. Environment
 2. ManagedIdentity
 3. AzureCLI

I believe this is the correct ordering (Environment should take precedence over CLI), and so am also proposing we update the old SDK auth to use the same ordering (minus ManagedIdentity):

1. Environment
2. AzureCLI

This may be a breaking change for some users that have both environment variables and CLI auth set up in the same environment. In the long run I believe this is the correct behavior though: environment variables should be able to override any CLI variables, rather than the other way round.


erezrokah (approved):

amanenk (approved):
LGTM

PR #1299, opened 2022-08-15
chore: Add PR labels for plugins

hermanschaaf:
#### Summary

Adds labels to PRs according to the plugin that is updated in the PR.

PR #1301, opened 2022-08-15
chore: Remove issue_reporting.md from plugin directories

hermanschaaf:
... and other miscellaneous changes related to PR labeling and typo-fixing.


erezrokah:
I think `.labeler` needs to `labeler`, https://github.com/cloudquery/cloudquery/pull/1299#issuecomment-1214852920

erezrokah (approved):

PR #1322, opened 2022-08-15
feat: Add fsx volumes

irmatov:
Closes https://github.com/cloudquery/cloudquery/issues/1372


plugins/source/aws/resources/services/fsx/volumes.hcl
```diff
@@ -0,0 +1,110 @@
...
+  column "tags" {
+    type = "json"
+    resolver "resolveTags" {
+      path = "github.com/cloudquery/cloudquery/plugins/source/aws/client.ResolveTags"
+    }
+  }
+
+  column "administrative_actions" {
+    skip = true
+  }
```
bbernays, line 53: Why are we skipping these fields?

erezrokah: Since this PR adds more of the information how about we do `administrative_actions` in a follow up PR?
I'll keep the issue open with a mention to this column

erezrokah (approved):

PR #1337, opened 2022-08-16
chore: Handle no tables in resource view

irmatov:
Give more informative message when there are no tables created yet and
an attempt is made to create the resources view.

Closes https://github.com/cloudquery/cloudquery/issues/1347


irmatov:
Resources view were only reformatted. Indentation was replaced to use spaces. At the end there is an `IF` statement that gives a new error message.

disq (approved):

PR #1434, opened 2022-08-16
fix: K8s Mock Testing

bbernays:
#### Summary

In my local IDE (vscode) gpls reports errors. This PR fixes it

---

- [ ] Update or add tests 🧪


erezrokah (approved):

plugins/source/k8s/resources/services/apps/daemon_sets.go
```diff
@@ -98,10 +98,10 @@ func DaemonSets() *schema.Table {
 				IgnoreInTests: true,
 			},
 			{
-				Name:        "cluster_name",
-				Description: "The name of the cluster which the object belongs to. This is used to distinguish resources with same name and namespace in different clusters. This field is not set anywhere right now and apiserver is going to ignore it if set in create or update request.",
+				Name:        "zzz_cluster_name",
```
erezrokah, line 101: This field was deprecated https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta.
We didn't catch it as tests were not running for this plugin

plugins/source/k8s/resources/services/rbac/roles_mock_test.go
```diff
@@ -1,14 +1,37 @@
-//go:build integration
```
erezrokah, line 1: This tagging was wrong as this is a mock test. Regardless I removed it

erezrokah:
Thanks for the PR @bbernays, I added another commit in https://github.com/cloudquery/cloudquery/pull/1434/commits/8b26a6dabb2ad3611630d894dfc3ab0df852bfd5 so `go test ./...` will run the tests. As a result I had to fix a bunch of issues which are in that same commit.

PR #1453, opened 2022-08-17
feat: Add cq-gen hcl and a new field to sqs queues

irmatov:


plugins/source/aws/resources/services/sqs/queues.go
```diff
@@ -62,169 +35,158 @@ func SQSQueues() *schema.Table {
...
 				Description: "The approximate number of messages that are in flight.",
-				Type:        schema.TypeInt,
+				Type:        schema.TypeBigInt,
 			},
 			{
 				Name:        "created_timestamp",
-				Description: "UNIX time when the queue was created.",
-				Type:        schema.TypeTimestamp,
-				Resolver:    client.ResolveTimestampField("CreatedTimestamp"),
+				Description: "The time when the queue was created in seconds (epoch time).",
```
shimonp21, line 66: Could we use a custom resolver to convert this (and other timestamps) to postgresql timestamp without timzeone? (UTC)?

It was requested a few times by customers to have a consistent format for timestamps..

shimonp21 (changes requested):

plugins/source/aws/resources/services/sqs/queues.go
shimonp21: Ah, but actually I guess since it's just a small refactor then this is just optional (there are many places in the code with unix-timestamps instead of Postgres timestamps - out-of-scope of this PR)

shimonp21 (approved):

erezrokah:
@irmatov can we merge this PR?

PR #1457, opened 2022-08-17
fix(docs): Add the missing path 'cli' for compiled bin

hardy4yooz:
#### Summary

Add the missing path 'cli' for compiled binary file.
---

- [ ] Update or add tests 🧪

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ For other providers, you can follow their specific guide on [Cloudquery Hub](htt
 
 ```
 make build-cli
-./bin/cloudquery # --help to see all options
+./bin/cli/cloudquery # --help to see all options
 ```
 
 ## Deployment via Helm


yevgenypats (approved):

hardy4yooz:
The most actions here are useless, we can filter the github actions by file suffix. 
eg: `md` file changes only trigger the `Lint markdown` action is correct. right?
@yevgenypats

erezrokah:
> The most actions here are useless, we can filter the github actions by file suffix. eg: `md` file changes only trigger the `Lint markdown` action is correct. right? @yevgenypats

Thanks @hardy4yooz, we actually ignore those in runtime:
image
https://github.com/cloudquery/cloudquery/blob/d9927c82b9c7f9cb92bf35a6ef8306257fe6450b/.github/workflows/lint_doc.yml#L34

The are 2 reasons we do it that way:
1. Since this is a monorepo, in order to use the built-in GitHub skip feature we'll need a workflow file per component, e.g.
`lint_go_aws.yml`:
```yaml
on:
  pull_request:
    branches:
      - main
    paths:
      - plugins/source/aws/** 
```
Which is hard to maintain as we need to manually add those files per component.
2. If we use the built-in GitHub skip feature, the status checks will show up as pending and won't pass the branch protection rules. See more in https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

PR #1508, opened 2022-08-19
fix(website): Fix position of corner in top image

hermanschaaf:
Fixes the position of the bottom block of section1.svg, which I calculated slightly incorrectly in the previous PR.


anton-kuptsov (approved):
Perfect! Thank you! 👍

PR #1615, opened 2022-08-30
feat: Added aws cis 1.5.0 policy

### LINKED ISSUE #1559: support CIS Amazon Web Services Foundations Benchmark v1.5.0

User requested the support of the lates AWS CIS policy.
 CIS Amazon Web Services Foundations Benchmark v1.5.0 is the latest one for now.

amanenk:
#### Summary
closes #1559 

---

- [ ] Update or add tests 🧪

### DIFF
--- a/plugins/source/aws/policies/queries/accessanalyzer/regions_with_no_accessanalyzers.sql
+++ b/plugins/source/aws/policies/queries/accessanalyzer/regions_with_no_accessanalyzers.sql
@@ -1,0 +1,22 @@
+insert into aws_policy_results
+WITH regions_with_enabled_accessanalyzer AS (
+    SELECT ar.region AS analyzed_region
+    FROM aws_regions ar
+             LEFT JOIN aws_access_analyzer_analyzers aaaa ON
+            ar.region = aaaa.region
+    WHERE aaaa.status = 'ACTIVE' )
+select
+    :'execution_time' as execution_time,
+        :'framework' as framework,
+        :'check_id' as check_id,
+        'certificate has less than 30 days to be renewed' as title,
+    account_id,
+    region AS resource_id,
+    case when
+             aregion.analyzed_region IS NULL
+             AND ar.enabled = TRUE
+         then 'fail'
+         else 'pass'
+        end as status
+FROM aws_regions ar LEFT JOIN regions_with_enabled_accessanalyzer aregion ON
+        ar.region = aregion.analyzed_region
--- a/plugins/source/aws/policies/queries/acm/certificates_expired.sql
+++ b/plugins/source/aws/policies/queries/acm/certificates_expired.sql
@@ -1,0 +1,14 @@
+insert into aws_policy_results
+select
+    :'execution_time' as execution_time,
+        :'framework' as framework,
+        :'check_id' as check_id,
+        'Ensure that all the expired SSL/TLS certificates stored in AWS IAM are removed (Automated)' as title,
+    account_id,
+    arn AS resource_id,
+    case when
+            not_after < NOW() AT TIME ZONE 'UTC'
+        then 'fail'
+        else 'pass'
+        end as status
+FROM aws_acm_certificates
--- a/plugins/source/aws/policies/queries/cloudwatch/alarm_organization_changes.sql
+++ b/plugins/source/aws/policies/queries/cloudwatch/alarm_organization_changes.sql
@@ -1,0 +1,31 @@
+insert into aws_policy_results
+select
+  :'execution_time' as execution_time,
+  :'framework' as framework,
+  :'check_id' as check_id,
+  'Ensure a log metric filter and alarm exists for AWS Organizations changes (Automated)' as title,
+  account_id,
+  cloud_watch_logs_log_group_arn as resource_id,
+  case when
+      pattern = '{ ($.eventName = AcceptHandshake) '
+      || '|| ($.eventName = AttachPolicy) '
+      || '|| ($.eventName = CreateAccount) '
+      || '|| ($.eventName = CreateOrganizationalUnit) '
+      || '|| ($.eventName = CreatePolicy) '
+      || '|| ($.eventName = DeclineHandshake) '
+      || '|| ($.eventName = DeleteOrganization) '
+      || '|| ($.eventName = DeleteOrganizationalUnit) '
+      || '|| ($.eventName = DeletePolicy) '
+      || '|| ($.eventName = DetachPolicy) '
+      || '|| ($.eventName = DisablePolicyType) '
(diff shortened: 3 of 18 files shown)


plugins/source/aws/policies/queries/cloudwatch/alarm_organization_changes.sql
```diff
@@ -0,0 +1,31 @@
...
+      || '|| ($.eventName = DetachPolicy) '
+      || '|| ($.eventName = DisablePolicyType) '
+      || '|| ($.eventName = EnablePolicyType) '
+      || '|| ($.eventName = InviteAccountToOrganization) '
+      || '|| ($.eventName = LeaveOrganization) '
+      || '|| ($.eventName = MoveAccount) '
+      || '|| ($.eventName = RemoveAccountFromOrganization) '
+      || '|| ($.eventName = UpdatePolicy) '
+      || '|| ($.eventName = UpdateOrganizationalUnit) }'
+      then 'pass'
```
bbernays, line 28: This query seems very fragile, If I am reading it correctly if a user were to put the events in a different order then it would fail

plugins/source/aws/policies/queries/cloudwatch/alarm_organization_changes.sql
```diff
@@ -0,0 +1,31 @@
...
+      else 'fail'
+  end as status
+from view_aws_log_metric_filter_and_alarm
```
bbernays, line 31: I don't think that the "unit" of compliance is individual resources, but the account as a whole. So rather than marking all log_metrics as failing we should only be marking the account as pass/fail

plugins/source/aws/policies/queries/iam/unused_creds_disabled_45_days.sql
```diff
@@ -0,0 +1,18 @@
...
+  case when
+      (password_enabled and password_last_used < (now() - '45 days'::INTERVAL)
+        or (last_used < (now() - '45 days'::INTERVAL)))
+      then 'fail'
+      else 'pass'
+  end
+from
+    aws_iam_users
+inner join aws_iam_user_access_keys
+    on aws_iam_users.cq_id = aws_iam_user_access_keys.user_cq_id
```
bbernays, line 18: This won't catch root user keys as there is no way for us to list IAM keys for the root user

jsonpr: I'm not sure if there's a specific table that stores the root access keys as it's technically not an IAM User.  the AWS API call to ListAccessKeys should also list root access keys (I have to verify this).    

On another note from a naming convention, I'm going back and forth if this should be categorized as aws_iam_user_access_keys for root or should it be something like aws_root_access_keys.  (That still belongs under the broader unused_creds_disabled_45_days).  

https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListAccessKeys.html

Are we storing root account information somewhere?  I briefly browsed through and saw an aws_accounts table but that did not seem quite right.  Happy to discuss this further.

plugins/source/aws/policies/queries/acm/certificates_expired.sql
```diff
@@ -0,0 +1,14 @@
...
+        :'check_id' as check_id,
+        'Ensure that all the expired SSL/TLS certificates stored in AWS IAM are removed (Automated)' as title,
+    account_id,
+    arn AS resource_id,
+    case when
+            not_after < NOW() AT TIME ZONE 'UTC'
+        then 'fail'
+        else 'pass'
+        end as status
+FROM aws_acm_certificates
```
bbernays, line 14: This check is not checking ACM certs, it is checking certs in IAM

as returned by `aws iam list-server-certificates` . I am not sure if we support this resource though. We might get the information as a field in the `AccountReport`

jsonpr: ^should be here and yes from AccountReport https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/docs/tables/aws_accounts.md.

plugins/source/aws/policies/queries/accessanalyzer/regions_with_no_accessanalyzers.sql
```diff
@@ -0,0 +1,22 @@
...
+    SELECT ar.region AS analyzed_region
+    FROM aws_regions ar
+             LEFT JOIN aws_access_analyzer_analyzers aaaa ON
+            ar.region = aaaa.region
+    WHERE aaaa.status = 'ACTIVE' )
+select
+    :'execution_time' as execution_time,
+        :'framework' as framework,
+        :'check_id' as check_id,
+        'certificate has less than 30 days to be renewed' as title,
```
bbernays, line 12: This seems to be the wrong title

plugins/source/aws/policies_v1/queries/cloudwatch/alarm_actions_disabled.sql
```diff
@@ -8,4 +8,4 @@ select :'execution_time'           as execution_time,
        'fail'                      as status
 from aws_cloudwatch_alarms
 where actions_enabled = false
-   or array_length(alarm_actions, 1) = 0;
+   or array_length(actions, 1) = 0
```
disq, line 11: This needs to stay as `alarm_actions`? there's no `actions` field.

disq, line 11: This needs to stay as `alarm_actions`? there's no `actions` field.

plugins/source/aws/policies_v1/views/metric_filters_with_active_subscriptions.sql
```diff
@@ -0,0 +1,16 @@
+-- describes metric alarms and pattern of alarm filter that has active subscriptions
+CREATE OR REPLACE VIEW view_aws_metric_filters_with_active_subscriptions AS
```
disq, line 2: This is never used?

amanenk: Oh. it was used in cloud watch checks.  it was used instad of `view_aws_log_metric_filter_and_alarm`. Which one do you think is more correct?

yevgenypats (approved):

plugins/source/aws/policies_v1/views/metric_filters_with_active_subscriptions.sql
disq: @amanenk just so we know the reason why it's there. Not sure which is more correct, but we can keep it for now since it's already there.

disq (approved):

PR #1645, opened 2022-08-31
feat: Write to config file instead of stdout (v2)

hermanschaaf:
Changes the default behavior of `generate` to write to file, rather than stdout. Also adds an `--output` (or `-O`) flag to specify where output should be written. Stdout output is currently explicitly not supported, to keep the featureset minimal.

~This depends on https://github.com/cloudquery/plugin-sdk/pull/29 for getting the plugin name, so go.mod will need to be updated here after that is merged in and released.~


cli/internal/destinations/postgresql/postgresql.go
```diff
@@ -173,9 +173,6 @@ func (p *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		}
 		var columnName pgx.Identifier = []string{c.Name}
 		fieldDef := columnName.Sanitize() + " " + pgType
-		if c.CreationOptions.Unique {
```
hermanschaaf, line 176: @yevgenypats I had to make some changes to this file to handle the `Unique` option not being there anymore (so the code could compile). Could you double-check this please?

cli/cmd/generate.go
```diff
@@ -75,15 +79,29 @@ func genSource(cmd *cobra.Command, path string, pm *plugin.PluginManager, regist
...
+
+	configPath := name + ".yml"
+	if outputFile != "" {
+		configPath = outputFile
+	}
+	err = writeFile(configPath, cfg)
+	if err != nil {
+		return errors.Wrap(err, "failed to write file")
+	}
+	fmt.Fprintln(os.Stderr, "✅ Source plugin config written to "+configPath)
```
yevgenypats, line 100: no emojies please :) otherwise we will have to introduce more flags like no color to stdout and so forth.
also why stderr if it's all prints it successfully ?

cli/cmd/generate.go
```diff
@@ -94,10 +112,38 @@ func genDestination(cmd *cobra.Command, path string, pm *plugin.PluginManager, r
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get plugin name")
+	}
+	cfg, err := client.GetExampleConfig(cmd.Context())
 	if err != nil {
 		return errors.Wrap(err, "failed to get example config")
 	}
-	fmt.Println(res)
+
+	configPath := name + ".yml"
```
yevgenypats, line 125: let's just ask the user to pass something ending with `yml` otherwise it will be `.yml.yml`

cli/cmd/generate_test.go
```diff
@@ -1,14 +1,37 @@
...
+			t.Fatal(err)
+		}
+
+		// sanity check that valid yaml has been produced
+		var cfg interface{}
+		yamlFile, err := ioutil.ReadFile(output)
+		if err != nil {
+			t.Fatalf("error reading config file output: %v ", err)
+		}
+		err = yaml.Unmarshal(yamlFile, &cfg)
```
yevgenypats, line 32: Can we do even further check on the content? just string equality I suggest?

cli/internal/destinations/postgresql/postgresql.go
yevgenypats: Im pushing a pr that removes this atm.

yevgenypats (changes requested):
few questions and also the postgresql code it will create a conflict for me as Im removing the unique atm and pushing a pr as well

cli/cmd/generate.go
hermanschaaf: This is for when the user hasn't passed in the flag (the name comes from the plugin RPC call)

cli/cmd/generate.go
yevgenypats: also let's use `fmt.Errorf` - I was experimenting with errors.Wrap to try and get the line number for sentry but it doesn't work great - I think we should only use it in plugins where it's hard to know where the error came from.

hermanschaaf: stderr because this is a log line and shouldn't be seen as the program's output (unless that is what we want?). Usually log lines go to stderr. At the same time, I didn't want the log line going to the log file, because then the CLI user doesn't see it unless they pass a special flag (that's why I didn't use `log.Info()` here either).

hermanschaaf: I'll take the emoji out 😢

yevgenypats: I think we should just output to the user (we can also write to the log in addition)

hermanschaaf: stderr is shown to the user, and I think it's the right place for something that is not strictly part of the program's output and is subject to change (like how Go's `log` package also writes to stderr). I'd like to keep it stderr, since I think that's better/more correct, but will change if you think we should still use stdout?

cli/cmd/generate_test.go
hermanschaaf: Done 👍

yevgenypats (approved):

PR #1698, opened 2022-09-01
feat: Auto-fill config when calling generate sub-command (v2)

hermanschaaf:
When calling `cloudquery generate source` or `cloudquery generate destination`, we now use information from the plugin to pre-populate the config with good defaults. This also adds comments and commented out lines to show what changes can be made after the initial generate call.

One more thing that changes is that plugins are now expected only to provide the contents of their own `spec` when returning an example. The rest of the config generation is handled by the CLI.


yevgenypats (changes requested):
The logic and templating looks perfect but I think this should go in the sdk level - https://github.com/cloudquery/plugin-sdk/blob/main/plugins/source.go#L107 

as the example is tied to a specific sdk anyway so in the CLI you can just call exampleConfig and that's it and also we will be able to use it in other places if needed in the future.

hermanschaaf:
@yevgenypats Makes sense, I will move it to the SDK :+1:

hermanschaaf:
@yevgenypats Should we perhaps merge this in (after I've rebased and fixed conflicts) to get the functionality? We can later move functionality to the SDK later as an internal (non-user facing) change, once that becomes possible/easier to do

PR #1705, opened 2022-09-01
fix: Handle SNS Subscriptions in Pending State

bbernays:
#### Summary

When a Subscription has not been confirmed, it doesn't have a valid ARN so CloudQuery is not able to make nested call. This PR does two things:
1. If a single call fails it will still try all other calls rather than ending early
2. It will not make a call if the ARN is equal to `PendingConfirmation`
---

- [ ] Update or add tests 🧪


amanenk (approved):
LGTM

erezrokah (approved):

PR #1720, opened 2022-09-04
feat: New go sdk for gcp

yevgenypats:
#### Summary
Apparent we were using an old SDK. We should use https://cloud.google.com/go/docs/reference were possible given the old one is "complete" and the new one has better and more standard interface which makes code generation much easier. 

---

- [ ] Update or add tests 🧪


erezrokah (approved):
Due to the size of the PR I think it's better if we merge and iron out issues as we go

PR #1729, opened 2022-09-05
fix: Provide `Version` in all providers

disq:
But it still needs to be set during build.


erezrokah (approved):
It's added during build https://github.com/cloudquery/cloudquery/blob/448046ef1ccf4cd393eceebcdf9d6cc0e8f2c863/plugins/.goreleaser.yaml#L12.

I think this won't be required for v2 cc @yevgenypats

PR #1742, opened 2022-09-06
WIP: AWS codegen

hermanschaaf:


yevgenypats:
This will be done in https://github.com/cloudquery/cloudquery/pull/1774 pr as complete codegen for resolvers is not feasiable and for structs will take too much time so we just want to migrate stuff manually to new sdk and use new codegen for new resources going forward. Closing.

PR #1765, opened 2022-09-07
chore: Remove deprecated linters (and dec-order)

hermanschaaf:
Similar to https://github.com/cloudquery/plugin-sdk/pull/64


erezrokah:
I think we need to update to `go@1.19` to upgrade the linter https://github.com/cloudquery/cloudquery/pull/1553#issuecomment-1226912575 as the latest version of the linter formats code using `go@1.19` which is a bit different.

You can trigger both updates via https://github.com/cloudquery/cloudquery/issues/760
image

erezrokah (approved):
Approved but I'd keep the current version of the linter per my comment and also merge with `main` to get latest fixes from https://github.com/cloudquery/cloudquery/commit/7a83a65446119b5339f5f3a3759f7f160a3716b4

hermanschaaf:
@erezrokah Ah okay, thanks! I'll revert the linter upgrade part 👍

PR #1773, opened 2022-09-08
chore: Add an .nvmrc file

hermanschaaf:
With this you can run `nvm use` in the directory and it will select the correct version of Node.


website/.nvmrc
```diff
@@ -0,0 +1 @@
+v16.17.0
```
erezrokah, line 1:
```suggestion
lts/*
```

This should ensure we're on latest stable and saves us the trouble with maintaining this version.
Node.js LTS update rarely break things

erezrokah (approved):

website/.nvmrc
erezrokah: Or maybe `v16` to get security updates and bug fixes

hermanschaaf: Agreed, updated to v16 👍 We can upgrade major versions manually

PR #1775, opened 2022-09-08
feat: Generate v2 table docs for website

hermanschaaf:
Uses the new `doc` command in the v2 CLI to generate table docs for our website. Currently only the two ported v2 plugins are supported (GCP and Heroku), but others can be added by uncommenting the lines.

My Node is a bit rusty, but this works locally at least 😄 

@erezrokah Let me know what you think - does it make sense to merge this to main and iterate there, or keep this open as a draft until more plugins are ready and we finalized the formatting more?


erezrokah (approved):
Nice 🚀  We should also disable the "edit this page on GitHub link" for the tables:
image

You can add a condition here https://github.com/cloudquery/cloudquery/blob/12643fe403f65061621ddea115486ad19ef0b274/website/components/EditLink.tsx#L41
and return `null` for the generated pages

hermanschaaf:
@erezrokah Nice, good catch, thanks!

PR #1776, opened 2022-09-08
feat!: Migrate GitHub plugin to v2

candiduslynx:
#### Summary

---

- [x] Update or add tests 🧪


plugins/source/github/docs/index.md
```diff
@@ -19,16 +16,12 @@ CloudQuery needs to be authenticated with your GitHub account's Personal Token i
...
-    access_token: "<YOUR ACCESS TOKEN HERE>"
-    orgs: ["<YOUR ORG NAME>"]
-  resources:
-    - "*"
+First, generate the configuration template using the following command:
+```bash
+cloudquery gen source github
 ```
 
+After that up fill in the `access_token` and `orgs` values.
```
hermanschaaf, line 24:
```suggestion
After that, edit `github.yml` and set the `access_token` and `orgs` values. 
```

plugins/source/github/resources/plugin/plugin.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/github/client"
+	"github.com/cloudquery/plugin-sdk/plugins"
+)
+
+const (
+	exampleConfig = `
+		access_token: <Access Token HERE>
+		orgs:
+		- cloudquery
+`
```
hermanschaaf, line 13:
```suggestion
	exampleConfig = `
access_token: <Access Token HERE>
orgs:
- cloudquery
`
```

I believe this should not be indented

hermanschaaf (changes requested):
I didn't review all the resources in great detail (I think we can merge and fix any issues as they come up). Overall it looks good :+1: I just left a couple of comments

plugins/source/github/resources/plugin/plugin.go
candiduslynx: When we use indentation then we get the properly constructed example config.
Alternatively, we should go to filling in the Spec with the example data and marshal it to YAML altogether.

yevgenypats (changes requested):
Given we have some more time for v2 release, let's introduce the new codegen. We only generate the structs and the resolvers should go into `resource_fetch.go`. See example - https://github.com/cloudquery/cloudquery/pull/1774

plugins/source/github/resources/plugin/plugin.go
hermanschaaf: @candiduslynx I'm not sure how you checked, but using this branch, I just ran the plugin server with `go run main.go serve` and then use the cli to run `go run main.go generate source --registry grpc 127.0.0.1:7777`. The generated config was not correctly indented:

Screenshot 2022-09-12 at 08 55 36

The code for the `generate` command adds the necessary indentation to get it at the right indentation level, we don't need to do it here. 

Personally I think an example string is fine, but if you want to change it to struct marshalling to yaml, that's great too, let's just make sure there's no indent either way.

plugins/source/github/resources/plugin/plugin.go
```diff
@@ -0,0 +1,30 @@
+package plugin
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/github/client"
+	"github.com/cloudquery/plugin-sdk/plugins"
+)
+
+const (
+	exampleConfig = `
+		access_token: <Access Token HERE>
```
yevgenypats, line 10: This should be with no tab - otherwise `gen` wont work.

yevgenypats (approved):
Looks good, I might have some more comments and nits on gen improvements but I think we can do this later on.

plugins/source/github/resources/plugin/plugin.go
yevgenypats: cc @candiduslynx

candiduslynx: done

hermanschaaf (approved):

PR #1777, opened 2022-09-08
feat!: Migrate CloudFlare plugin to v2

disq:
~As of first commit, resource schemas are not migrated yet.~

~Untested but compiles/lints.~


plugins/source/cloudflare/client/spec.go
```diff
@@ -1,16 +1,14 @@
...
+type Spec struct {
+	Token    string   `json:"api_token,omitempty"`
+	ApiKey   string   `json:"api_key,omitempty"`
+	ApiEmail string   `json:"api_email,omitempty"`
+	Accounts []string `json:"accounts,omitempty"`
+	Zones    []string `json:"zones,omitempty"`
 }
 
-func (Config) Example() string {
+func (Spec) Example() string {
```
hermanschaaf, line 11: I'm not sure if this `Example` function is used anywhere, but if it is, the comments should be yaml-formatted (`#`). If this function is not used, let's remove it.

plugins/source/cloudflare/docs/index.md
```diff
@@ -1,54 +0,0 @@
-## Cloudflare Plugin
```
hermanschaaf, line 1: I think we should keep the docs directory with this `index.md` file at least (I believe it gets used to populate the website as well). 

We don't need `docs.go` or the `tables` directory anymore though.

plugins/source/cloudflare/docs/index.md
```diff
@@ -1,54 +0,0 @@
...
-api_key: "<YOUR_API_KEY_HERE>"
-// API email to access Cloudflare resources, also can be set with the CLOUDFLARE_API_EMAIL environment variable
-api_email: "<YOUR_API_EMAIL_HERE>"
-
-// List of accounts to target, if empty, all accounts will be targeted
-//accounts:
-// - "<YOUR_ACCOUNT_ID>"
-
-// List of accounts to target, if empty, all available zones will be targeted
-//zones:
```
hermanschaaf, line 31: This example should also be changed to use `yaml`-style comments.

plugins/source/cloudflare/docs/index.md
```diff
@@ -1,54 +0,0 @@
-## Cloudflare Plugin
-
-The CloudQuery Cloudflare provider pulls configuration out of Cloudflare resources, normalizes them and stores them in PostgreSQL database.
-
-### Install
-
-```bash
-cloudquery init cloudflare
```
hermanschaaf, line 8: This will need to be updated (`cloudquery generate cloudflare`) here and lower down. It would also be good to add a sentence or two about what the command does: it generates a file called `cloudflare.yml` with example config that needs to be tweaked.

plugins/source/cloudflare/plugin/plugin.go
```diff
@@ -0,0 +1,48 @@
...
+token: <token>
+
+## Optional. GRPC Retry/backoff configuration, time units in seconds. Documented in https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md
+# backoff_base_delay: 1
+# backoff_multiplier: 1.6
+# backoff_max_delay: 120
+# backoff_jitter: 0.2
+# backoff_min_connect_timeout = 0
+## Optional. Max amount of retries for retrier, defaults to max 3 retries.
+# max_retries: 3
```
hermanschaaf, line 21: This example config needs to be updated to match the expectations for cloudflare. (and the name in the comment changed from `Heroku` -> `Cloudflare` 😉 )

I also saw @yevgenypats remove these backoff options from the gcp plugin in recent PR. I don't know the reasoning, but we should probably remove it here and from the Heroku plugin as well.

hermanschaaf (changes requested):
Left a couple of comments; nothing major. I haven't gone through all the resources yet :)

plugins/source/cloudflare/plugin/plugin.go
yevgenypats: Good catch, agree we need to remove it here as well. I dont think the backoff stuff in GCP ever worked at least in the expected way and we dont know if they were use so rather then introducing options that confuse users I would prefer for now to remove it and add it later if it will be requested.

disq: oops yeah, I meant to use ExampleConfig but then copied your plugin.

hermanschaaf: I like your way of putting `Example()` as a method on the `Spec` struct - makes it less likely that you'll forget to update the example when the config changes. We should probably adopt that as a standard.

plugins/source/cloudflare/docs/index.md
disq: Updated

plugins/source/cloudflare/docs/index.md
disq: Done.

plugins/source/cloudflare/docs/index.md
disq: Done.

plugins/source/cloudflare/plugin/plugin.go
```diff
@@ -0,0 +1,54 @@
...
+# API key to access Cloudflare resources, also can be set with the CLOUDFLARE_API_KEY environment variable
+#api_key: "<YOUR_API_KEY_HERE>"
+# API email to access Cloudflare resources, also can be set with the CLOUDFLARE_API_EMAIL environment variable
+#api_email: "<YOUR_API_EMAIL_HERE>"
+
+# List of accounts to target, if empty, all accounts will be targeted
+#accounts:
+# - "<YOUR_ACCOUNT_ID>"
+
+# List of accounts to target, if empty, all available zones will be targeted
```
hermanschaaf, line 25:
```suggestion
# List of zones to target, if empty, all available zones will be targeted
```

plugins/source/cloudflare/plugin/plugin.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/cloudflare/client"
+	"github.com/cloudquery/cloudquery/plugins/source/cloudflare/services"
+	"github.com/cloudquery/plugin-sdk/plugins"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+const exampleConfig = `
+# Use can use either the API token or the API key. (API key requires setting the API email field)
```
hermanschaaf, line 11:
```suggestion
# You may use either the API token or the API key. (API key requires setting the API email field)
```

hermanschaaf (approved):

yevgenypats (changes requested):
Can you please open a new cloudflare test account to manually test it?

disq:
> Can you please open a new cloudflare test account to manually test it?

tested, works

yevgenypats (changes requested):
Looks good but I'd like the codegen to look more like in aws and gcp https://github.com/cloudquery/cloudquery/pull/1774. i.e split to files and avoid `parent,combine` and other functional utility functional to make it easier to maintain (imo). if something has relations the convention is to put the relation in it's own file (see aws)

disq:
> Looks good but I'd like the codegen to look more like in aws and gcp #1774. i.e split to files and avoid `parent,combine` and other functional utility functional to make it easier to maintain (imo). if something has relations the convention is to put the relation in it's own file (see aws)

Updated.

erezrokah (approved):

erezrokah (approved):

plugins/source/cloudflare/docs/index.md
```diff
@@ -1,43 +1,48 @@
 ## Cloudflare Plugin
 
-The CloudQuery Cloudflare provider pulls configuration out of Cloudflare resources, normalizes them and stores them in PostgreSQL database.
+The CloudQuery Cloudflare plugin pulls configuration out of Cloudflare resources, normalizes them and stores them in PostgreSQL database.
 
 ### Install
 
 ```bash
-cloudquery init cloudflare
+cloudquery generate cloudflare
```
erezrokah, line 8:
```suggestion
cloudquery generate source cloudflare
```

plugins/source/cloudflare/docs/index.md
```diff
@@ -1,43 +1,48 @@
...
 ```
 
+This will generate a `cloudflare.yml` file containing example configuration.
+
 ### Configuration
 
-The following configuration section can be automatically generated by `cloudquery init cloudflare`:
+The following configuration section will be automatically generated by `cloudquery generate cloudflare`:
```
erezrokah, line 15:
```suggestion
The following configuration section will be automatically generated by `cloudquery generate source cloudflare`:
```

yevgenypats (approved):

PR #1780, opened 2022-09-09
feat!: Migrate Terraform plugin to v2

disq:
~`services/tfdata_test.go` is not ready.~


plugins/source/terraform/client/backend.go
```diff
@@ -21,9 +21,9 @@ type BackendType string
 
 // BackendConfigBlock - abstract backend config
 type BackendConfigBlock struct {
-	BackendName string                 `yaml:"name"`
-	BackendType string                 `yaml:"backend"`
-	ConfigAttrs map[string]interface{} `yaml:",inline"`
+	BackendName string                 `json:"name"`
+	BackendType string                 `json:"backend"`
+	ConfigAttrs map[string]interface{} `json:",inline"`
```
disq, line 26: This will need testing.

disq: Fixed, the new config format has the backend-specific config under backend-specific keys.

plugins/source/terraform/services/tfdata_test.go
```diff
@@ -0,0 +1,23 @@
...
+//        local:
+//          path: "../examples/terraform.tfstate"
+//`
+//
+//	// TODO
+//	providertest.TestResource(t, providertest.ResourceTestCase{
+//		Provider: provider.Provider(),
+//		Config:   cfg,
+//	})
+//}
```
hermanschaaf, line 23: Would it be possible to add this test back in now?

plugins/source/terraform/plugin/plugin.go
```diff
@@ -0,0 +1,37 @@
...
+  - name: mylocal # local backend
+    local:
+      path: ./examples/terraform.tfstate
+#  - name: myremote # s3 backend
+#    s3:
+#      bucket: tf-states
+#      key: terraform.tfstate
+#      region: us-east-1
+#      role_arn: ""
+`
```
hermanschaaf, line 21: I would suggest changing this config format to something like this:

```
backends:
  - name: mylocal # local backend
    type: local
    params:
      path: ./examples/terraform.tfstate
#  - name: myremote # s3 backend
#    type: s3
#    params: 
#      bucket: tf-states
#      key: terraform.tfstate
#      region: us-east-1
#      role_arn: ""
```

Each "type" can then specify its own set of parameters. This way, we don't have to rely on the presence / absence of keys to tell what kind of remote backend it is. We can also more easily check whether a given "type" is supported by the current version of the code, and error out if it is not. WDYT?

plugins/source/terraform/docs/index.md
```diff
@@ -5,31 +5,24 @@ The CloudQuery Terraform plugin extracts terraform state.
...
-           backend: local
-           path: ./examples/terraform.tfstate
-         - name: myremote # s3 backend
-           backend: s3
-           bucket: tf-states
-           key: terraform.tfstate
-           region: us-east-1
-           role_arn: ""
-      # list of resources to fetch
-      resources:
```
hermanschaaf, line 31: I notice the removal of `resources` from the config: is it still possible to filter the resources to fetch?

hermanschaaf (changes requested):

plugins/source/terraform/plugin/plugin.go
disq: This will complicate things, we'll basically need to have a copy of specreader-style code in the plugin. Which is not pretty. (The old `",inline"` thing doesn't work because we're supposed to use json tags/decoder even if we're using yaml config.)

plugins/source/terraform/docs/index.md
disq: It's not supposed to be here, it will be in the outer spec and managed by the core side of things. (It's still possible and there's no reason for it to not work)

hermanschaaf: Oh yeah of course, makes sense 👍

plugins/source/terraform/services/tfdata_test.go
disq: I don't think we have this type of test in any other place, so I'm not sure how it will work (and if it will be accepted by @yevgenypats ). If you have a example plugin that does this already, I'm happy to port that code here.

plugins/source/terraform/plugin/plugin.go
hermanschaaf: Hmm, if we want to keep things simple, can we use a `map[string]interface{}` type for the params, then place the values into appropriate structs with some custom logic?

disq: Fixed

plugins/source/terraform/docs/index.md
```diff
@@ -39,13 +32,13 @@ CloudQuery currently supports LOCAL and S3 backends.
...
-        key: "<terraform state key>"
-        region: us-east-1
-        role_arn: ""
+backends:
+  - name: myremote # s3 backend
+    s3:
+      bucket: "<terraform state bucket>"
+      key: "<terraform state key>"
+      region: us-east-1
+      role_arn: ""
```
hermanschaaf, line 41: Thanks for updating the config! This example (and the local one above it) will also need to be updated now

disq: Fixed

plugins/source/terraform/services/tfdata_test.go
hermanschaaf: I think the terraform plugin is a bit special in this regard, because we want to test that it is able to read and decode a local terraform state file. I ported the test to do this using the new SDK in https://github.com/cloudquery/cloudquery/pull/1780/commits/730cdaa6188d1ea286bc0063190b00214470c4fd (pushed to this branch) - happy to hear @yevgenypats's thoughts here though

erezrokah (approved):

plugins/source/terraform/services/tfdata_test.go
yevgenypats: I don't see a reason for that not to work - just as our mock tests are working.

PR #1781, opened 2022-09-09
feat!: Migrate Okta plugin to v2

disq:


yevgenypats (changes requested):
Can you please open a new okta account and manually test it ?

disq:
> Can you please open a new okta account and manually test it ?

seems to work!

plugins/source/okta/client/spec.go
```diff
@@ -0,0 +1,6 @@
+package client
+
+type Spec struct {
+	Token  string `yaml:"token,omitempty"`
```
yevgenypats, line 4: it should be json (we use json everywhere apart from the first step in the CLI where we read yaml and then convert to json - just like in k8s)

yevgenypats (changes requested):
Looks good. one comment and ready to go.

yevgenypats (approved):

PR #1835, opened 2022-09-16
feat: AWS v2: Split users and credential report users into separate resources

hermanschaaf:


plugins/source/aws/resources/services/iam/users_fetch.go
```diff
@@ -2,116 +2,67 @@ package iam
...
-			Cert1LastRotated:      reportUser.Cert1LastRotated,
-			Cert2LastRotated:      reportUser.Cert2LastRotated,
-			reportUser:            reportUser,
-			isRoot:                true,
+	listUser := listInfo.(types.User)
+	arn := ""
+	if listUser.Arn != nil {
+		arn = *listUser.Arn
+	}
+	if arn == c.AccountGlobalARN(client.IamService, "root") {
```
bbernays, line 50: I don't think this check is needed anymore.

plugins/source/aws/resources/services/iam/users_fetch.go
```diff
@@ -2,116 +2,67 @@ package iam
...
+		// TODO: check if this is still necessary after unbundling credential reports
```
bbernays, line 51: I don't think it is. We should be able to to just return the User object

plugins/source/aws/resources/services/iam/users_fetch.go
bbernays, line 50: I don't think any of this is necessary anymore as `ListUser` never returns the root user

plugins/source/aws/resources/services/iam/users_fetch.go
```diff
@@ -354,7 +163,7 @@ func postIamUserAccessKeyResolver(ctx context.Context, meta schema.ClientMeta, r
 
 func fetchIamUserAttachedPolicies(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
 	var config iam.ListAttachedUserPoliciesInput
-	p := parent.Item.(UserWrapper)
+	p := parent.Item.(*types.User)
 	if aws.ToString(p.UserName) == rootName {
 		return nil
 	}
```
bbernays, line 169: Not necessary anymore as Root user is never returned

plugins/source/aws/resources/services/iam/users_fetch.go
```diff
@@ -289,7 +98,7 @@ func fetchIamUserGroups(ctx context.Context, meta schema.ClientMeta, parent *sch
 
 func fetchIamUserAccessKeys(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
 	var config iam.ListAccessKeysInput
-	p := parent.Item.(UserWrapper)
+	p := parent.Item.(*types.User)
 	svc := meta.(*client.Client).Services().IAM
 	if aws.ToString(p.UserName) == rootName {
 		return nil
```
bbernays, line 104: Not necessary anymore as Root user is never returned

plugins/source/aws/resources/services/iam/user_policies_fetch.go
```diff
@@ -14,7 +15,7 @@ import (
 func fetchIamUserPolicies(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
 	c := meta.(*client.Client)
 	svc := c.Services().IAM
-	user := parent.Item.(UserWrapper)
+	user := parent.Item.(*types.User)
 	if aws.ToString(user.UserName) == rootName {
 		return nil
 	}
```
bbernays, line 21: Not necessary

plugins/source/aws/resources/plugin/tables.go
```diff
@@ -195,6 +195,7 @@ func tables() []*schema.Table {
 		glue.Workflows(),
 		guardduty.Detectors(),
 		iam.Accounts(),
+		iam.CredentialReportUsers(),
```
bbernays, line 198: I would call this just `iam.CredentialReport`

plugins/source/aws/resources/services/iam/credential_report_users_fetch.go
```diff
@@ -0,0 +1,77 @@
...
+	PasswordNextRotation  string    `csv:"password_next_rotation"`
+	MfaActive             bool      `csv:"mfa_active"`
+	AccessKey1Active      bool      `csv:"access_key_1_active"`
+	AccessKey2Active      bool      `csv:"access_key_2_active"`
+	AccessKey1LastRotated string    `csv:"access_key_1_last_rotated"`
+	AccessKey2LastRotated string    `csv:"access_key_2_last_rotated"`
+	Cert1Active           bool      `csv:"cert_1_active"`
+	Cert2Active           bool      `csv:"cert_2_active"`
+	Cert1LastRotated      string    `csv:"cert_1_last_rotated"`
+	Cert2LastRotated      string    `csv:"cert_2_last_rotated"`
```
bbernays, line 29: There are some fields that are missing:

```
access_key_1_last_used_date
access_key_1_last_used_region
access_key_1_last_used_service
access_key_2_last_used_date
access_key_2_last_used_region
access_key_2_last_used_service
password_last_used
```

plugins/source/aws/resources/plugin/tables.go
hermanschaaf: Hmm, I'm just worried that the table will then be called `credential_reports`, which doesn't sound right considering that it contains user entries, not report entries. WDYT?

plugins/source/aws/resources/services/iam/credential_report_users_fetch.go
hermanschaaf: Added 👍

plugins/source/aws/resources/plugin/tables.go
bbernays: Users are the only resource that have credentials though

plugins/source/aws/resources/services/iam/credential_reports.go
```diff
@@ -0,0 +1,131 @@
...
+			},
+			{
+				Name:     "user",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("User"),
+			},
+			{
+				Name:     "user_creation_time",
+				Type:     schema.TypeTimestamp,
+				Resolver: schema.PathResolver("UserCreationTime"),
```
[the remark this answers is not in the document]
hermanschaaf: Good catch, updated!

bbernays (approved):
LGTM

Issue #1808, opened 2022-09-13
feat(cli-v2): When generating a source the CLI should set at least 1 destination for it


erezrokah:
### Is your feature request related to a problem? Please describe.

After generating a configuration for a source plugin, if you try to run a `sync` you'll get:
![image](https://user-images.githubusercontent.com/26760571/189932877-d133bda1-f860-4568-ac1b-adff5bbe6871.png)

Then you'll need to:
1. `cloudquery gen destination postgresql`
2. Manually update the source `yml` with the destination

Since having at least one destination is a requirement, we should make it work more seamlessly

### Describe the solution you'd like

Either always add a default destination, or prompt the user to select from a list

### Describe alternatives you've considered

Manually updating source configuration files

PR #1847, opened 2022-09-18
feat: Add destination when generating sources

erezrokah:
#### Summary

Requires https://github.com/cloudquery/plugin-sdk/pull/120

Somehow fixes https://github.com/cloudquery/cloudquery/issues/1808.
When generating sources, if there's a destination generated already, add it to the source configuration.

Then I think our docs should first instruct users to generate the destination, then the sources.

---

- [x] Update or add tests 🧪

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -5,6 +5,7 @@ import (
 	"embed"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"text/template"
 
@@ -62,6 +63,22 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	}
 }
 
+func getDestinations(directory string) []string {
+	specReader, err := specs.NewSpecReader(directory)
+	if err != nil {
+		// no need to fail here as we only use this for config generation
+		return nil
+	}
+
+	destinations := make([]string, 0)
+	for _, destination := range specReader.GetDestinations() {
+		destinations = append(destinations, destination.Name)
+
+	}
+
+	return destinations
+}
+
 func genSource(cmd *cobra.Command, path string, pm *plugins.PluginManager, registry specs.Registry, outputFile string) error {
 	if registry == specs.RegistryGithub && !strings.Contains(path, "/") {
 		path = "cloudquery/" + path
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):
I think this looks good! but I think we have to be cautions with `gen` turning into a product of itself which generates configs for various use-cases and so on so really important to keep it simple as I would even go far that `gen` is even not mandatory on it's own as it's quite common in most cases for users to create configs from documentation rather from commands that auto generate configs.

erezrokah:
> I think this looks good! but I think we have to be cautions with `gen` turning into a product of itself which generates configs for various use-cases and so on so really important to keep it simple as I would even go far that `gen` is even not mandatory on it's own as it's quite common in most cases for users to create configs from documentation rather from commands that auto generate configs.

💯 This is a feature on its own. I'm trying to prevent the back and forth between CLI commands and manually editing files.
Another approach would be to make config creation/editing DX better (i.e. expose schema, have auto completion in IDE), but I think at the moment this is the fastest thing we can do to improve onboarding.

erezrokah:
> Because for example how do we know the user wants to sync the source with all destinations? so either he will have to add a destination or to delete a destination (not sure which is better I think to explicitly add might be better and the more secure way)

Yes I think this feature might change when we have more destinations (for example I think we can let the user chose from a list)

erezrokah:
Also waiting on https://github.com/cloudquery/plugin-sdk/pull/120 to merge this

yevgenypats:
> > Because for example how do we know the user wants to sync the source with all destinations? so either he will have to add a destination or to delete a destination (not sure which is better I think to explicitly add might be better and the more secure way)
> 
> Yes I think this feature might change when we have more destinations (for example we can let the user chose from a list)

I wouldn't go into the rabbit hole of adding more features like choosing from a list but rather just output the config and make this feature end at this. Im even at the opinion that the `gen` command is a bit too much but given we already have it and people like it we can keep it but wouldn't work on adding more features and interactivity to that as it's not the main goal of CQ (other we branch to auto-generating config product)

yevgenypats (changes requested):
Actually second thought here - Let's keep the `gen` command write only without reading the current directly I think it is really important to keep bugs at bay and things like oh it read a different directory and so on. I think the current `gen` command is way more advanced to the current status of the tool anyway.

erezrokah:
Got it, I'll close the PR

Issue #1803, opened 2022-09-13
test: Add tests on multiple OS


erezrokah:
Since we support MacOS, Linux and Windows we should add sanity tests on each OS

yevgenypats:
In that case, maybe we should just run all unit-tests on all OS  as a matrix?

erezrokah:
> In that case, maybe we should just run all unit-tests on all OS as a matrix?

Yeah I'm assuming only the CLI unit tests. I think this means we'll need a separate workflow file as now we run tests the same way for all modules https://github.com/cloudquery/cloudquery/blob/14d2ca6894bdf51597082fbadd1aad7401316e59/.github/workflows/test_unit.yml#L21.

The challenge is setting up docker on MacOS, I'm trying with https://github.com/docker-practice/actions-setup-docker which [seems to time out after 10m](https://github.com/cloudquery/cloudquery/actions/runs/3077855910).

PR #1848, opened 2022-09-18
test(cli): Add cross env tests for `gen,sync` commands

erezrokah:
#### Summary

Adds basic integration tests. The `sync` test is failing due to a bug in our release pipeline (not embedding versions) which I'll fix in a separate PR

---

- [ ] Update or add tests 🧪

### DIFF
--- a/.github/workflows/test_unit.yml
+++ b/.github/workflows/test_unit.yml
@@ -1,11 +1,12 @@
-name: Unit Tests Plugins
+name: Unit Tests
 
 on:
   pull_request:
     branches:
       - main
-env:
+env: 
   CGO_ENABLED: 0
+  CQ_NO_TELEMETRY: 1
 
 jobs:
   resolve-modules:
@@ -25,7 +26,6 @@ jobs:
       matrix: ${{ fromJson(needs.resolve-modules.outputs.matrix) }}
       fail-fast: false
     runs-on: ubuntu-latest
-    if: matrix.workdir != 'cli'
     services:
       postgres:
         image: postgres:11
@@ -74,6 +74,11 @@ jobs:
         if: steps.changed-files.outputs.any_changed == 'true'
         working-directory: ${{ matrix.workdir }}
         run: go get -v -t -d ./...
+
+      - name: Build
+        if: steps.changed-files.outputs.any_changed == 'true' && matrix.workdir == 'cli'
+        run: go build
+        working-directory: ${{ matrix.workdir }}
 
       - name: Test
         if: steps.changed-files.outputs.any_changed == 'true'
--- a/cli/cmd/generate_test.go
+++ b/cli/cmd/generate_test.go
@@ -68,11 +68,7 @@ spec:
 `
 
 func TestGenerate(t *testing.T) {
-	tmpdir, tmpErr := os.MkdirTemp("", "generate_test_*")
-	if tmpErr != nil {
-		t.Fatalf("failed to create temporary directory: %v", tmpErr)
-	}
-	defer os.RemoveAll(tmpdir)
+	tmpdir := t.TempDir()
 
 	t.Run("generate source", func(t *testing.T) {
 		output := path.Join(tmpdir, "test-source.yml")
--- a/.github/workflows/test_cli.yml
+++ b/.github/workflows/test_cli.yml
@@ -35,7 +35,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        os: [macos-latest, windows-latest]
+        os: [macos-12, windows-2022]
         workdir: [cli]
       fail-fast: false
     steps:
--- a/.github/workflows/test_cli.yml
+++ b/.github/workflows/test_cli.yml
@@ -1,4 +1,4 @@
-name: Build
+name: Unit Tests CLI
 
 on:
   pull_request:
@@ -8,24 +8,11 @@ env:
   CGO_ENABLED: 0
 
 jobs:
-  resolve-modules:
-    name: Resolve Modules
+  check-cli-changes:
+    name: Check for CLI Changes
     runs-on: ubuntu-latest
     outputs:
-      matrix: ${{ steps.set-matrix.outputs.matrix }}
-    steps:
-      - name: Checkout Sources
-        uses: actions/checkout@v3
-      - id: set-matrix
-        run: ./scripts/resolve-modules.sh
-  test-build:
-    name: Build
-    needs: resolve-modules
-    strategy:
-      matrix: ${{ fromJson(needs.resolve-modules.outputs.matrix) }}
-      fail-fast: false
-    runs-on: ubuntu-latest
-    
+      cli-changed: ${{ steps.changed-files.outputs.any_changed }}
     steps:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v3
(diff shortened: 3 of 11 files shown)


.github/workflows/test_build.yml
```diff
@@ -19,7 +18,7 @@ jobs:
         uses: actions/checkout@v3
       - id: set-matrix
         run: ./scripts/resolve-modules.sh
-  test_build:
+  test-build:
```
erezrokah, line 21: Keep casing consistent

.github/workflows/test_build.yml
```diff
@@ -6,7 +6,6 @@ on:
       - main
 env: 
   CGO_ENABLED: 0
-  CQ_NO_TELEMETRY: 1
```
erezrokah, line 9: I don't think we need this anymore

.github/workflows/test_integration.yml
```diff
@@ -0,0 +1,65 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+env: 
+  CGO_ENABLED: 0
+
+jobs:
+  check-cli-changes:
```
erezrokah, line 11: This should save spinning up the `windows,linux,macOS` matrix if the CLI hadn't changed

cli/integration_tests/gen_test.go
```diff
@@ -0,0 +1,48 @@
...
+func TestGen(t *testing.T) {
+	for _, plugin := range pluginsToTest {
+		t.Run(fmt.Sprintf("gen %s %s", plugin.kind, plugin.name), func(t *testing.T) {
+			i := newIntegrationTest(t)
+			output, err := i.runCommand(t, "gen", plugin.kind, plugin.name)
+			fmt.Println(output)
+			require.NoError(t, err)
+			require.Contains(t, output, "plugin config successfully written to "+plugin.name+".yml")
+			config := validateYaml(t, path.Join(i.dir, plugin.name+".yml"))
+			require.Equal(t, plugin.kind, config.Kind)
```
erezrokah, line 45: This is a sanity check. We shouldn't be testing the whole parsing here

cli/integration_tests/sync_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func addDestination(t *testing.T, configFile string) {
```
erezrokah, line 15: We should be able to remove this once https://github.com/cloudquery/cloudquery/pull/1847 is merged

erezrokah:
OK so I found https://github.com/cloudquery/cloudquery/blob/9e3ace775da2d600968ef4275e9e0013d4dfd825/cli/cmd/generate_test.go#L70 and https://github.com/cloudquery/cloudquery/blob/9e3ace775da2d600968ef4275e9e0013d4dfd825/cli/cmd/sync_test.go#L9 so I'll merge the too.
The separation to integration tests is so we can run them on Windows/MacOS too

erezrokah:
OK so this PR improves on the current test by:
1. Runs each test in a separate isolated directory
2. Runs the CLI via an external process closer to how the user runs int
3. Separate the integration tests from the unit tests and runs them on a multiple OS matrix

erezrokah:
Looks like the CLI v2 doesn't work on Windows https://github.com/cloudquery/cloudquery/actions/runs/3077812862/jobs/4972939978#step:7:42

yevgenypats:
> OK so this PR improves on the current tests by:
> 
> 1. Runs each test in a separate isolated directory so easy to add more scenarios
> 2. Runs the CLI via an external process closer to how the user runs it
> 3. Separate the integration tests from the unit tests and runs them on a multiple OS matrix

Im not a big fan of running this test as an external process as it makes testing good and bad test cases much hard as it will involve stdout output. I specifically built the cobra in such way that it should be possible to do so via `NewRootCmd` which is the main entry point so there shouldn't be any difference between this and an external process.

erezrokah:
> I specifically built the cobra in such way that it should be possible to do so via `NewRootCmd` which is the main entry point so there shouldn't be any difference between this and an external process.

Is there a way to run isolated tests in a temp directory and not in the CWD? That should allow us to reduce dependencies between tests and run tests in parallel.

Also I think testing stdout is not necessarily bad in the case of CLIs as that's what the user experiences. We can leverage snapshot testing on the CLI output to avoid the need to parse the output.
Probably better to discuss via a Zoom call :)

yevgenypats:
> > I specifically built the cobra in such way that it should be possible to do so via `NewRootCmd` which is the main entry point so there shouldn't be any difference between this and an external process.
> 
> Is there a way to run isolated tests in a temp directory and not in the CWD? That should allow us to reduce dependencies between tests and run tests in parallel.
> 
> Also I think testing stdout is not necessarily bad in the case of CLIs as that's what the user experiences. We can leverage snapshot testing on the CLI output to avoid the need to parse the it. Probably better to discuss via a Zoom call :)

Yeah I think it should be possible to do in separate directories. Also, Im not a big fan of snapshot testing :) https://medium.com/@sapegin/whats-wrong-with-snapshot-tests-37fbe20dfe8e
but sure let's do zoom 📞

cli/integration_tests/gen_test.go
```diff
@@ -0,0 +1,58 @@
...
```
hermanschaaf, line 51:
```suggestion
			// sanity check only; we shouldn't be testing the full config here
			require.Equal(t, plugin.kind, config.Kind)
```

.github/workflows/test_unit.yml
```diff
@@ -77,5 +76,5 @@ jobs:
 
       - name: Test
         if: steps.changed-files.outputs.any_changed == 'true'
-        run: go test -v ./...
+        run: go test -v ./... --tags unit
```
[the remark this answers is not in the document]
erezrokah: Good point. I think I'll take the suggestion from https://github.com/cloudquery/cloudquery/issues/1803 to run all tests on all OSs

erezrokah:
> Yeah I think it should be possible to do in separate directories. Also, Im not a big fan of snapshot testing :) https://medium.com/@sapegin/whats-wrong-with-snapshot-tests-37fbe20dfe8e

I'm familiar with the article and agree that in most cases (like testing Web apps with Jest) they don't provide much value.
I do think for CLIs they can provide benefits but lets discuss separately as it's out of scope for this PR and for v2 initial iteration. Opened a separate issue to discuss https://github.com/cloudquery/cloudquery/issues/1860

cli/cmd/generate_test.go
```diff
@@ -68,11 +68,7 @@ spec:
 `
 
 func TestGenerate(t *testing.T) {
-	tmpdir, tmpErr := os.MkdirTemp("", "generate_test_*")
-	if tmpErr != nil {
-		t.Fatalf("failed to create temporary directory: %v", tmpErr)
-	}
-	defer os.RemoveAll(tmpdir)
+	tmpdir := t.TempDir()
```
hermanschaaf, line 71: Oh, nice, TIL that `testing.T` provides a `TempDir` function 🚀

hermanschaaf (approved):

yevgenypats (changes requested):
Let's have a live discussion on the integration testing strategy?

cli/cmd/doc_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
```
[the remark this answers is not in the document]
erezrokah: This code it outdated per @hermanschaaf's comment. We don't have a separate between unit tests and integration tests.
This PR introduced a separate workflow that run only the CLI unit tests on Windows and MacOS (yet to be completed):
https://github.com/cloudquery/cloudquery/pull/1848/files#diff-0dbaaba16effff4896e9054770ab112bf49ae5b1e41a70915abafeef253b9ca7R31

yevgenypats (changes requested):
Let's discuss live but see comment here - https://github.com/cloudquery/cloudquery/issues/1860#issuecomment-1259419465 

I don't think that we need to test/maintain output snapshot/predictability of a CLI

erezrokah:
> I don't think that we need to test/maintain output snapshot/predictability of a CLI

To clarify, this PR does not introduce snapshot testing. I adds a workflow to run exiting CLI tests on Windows and Mac and changes the CLI tests so they can be executed in a temporary directory and not the CWD of the tests

erezrokah:
Replaced by https://github.com/cloudquery/cloudquery/pull/2204

PR #1856, opened 2022-09-18
fix: Resolve plugin version from GitHub registry instead of gRPC call

erezrokah:
#### Summary

This PR fixes https://github.com/cloudquery/docs/issues/216 which is still present in v2 and causing failures in https://github.com/cloudquery/cloudquery/pull/1848. Also somehow related to https://github.com/cloudquery/cloudquery/issues/908

If you try:
1. `gen source test`
2. `gen destination postgresql`
3. `gen sync .`
you'll get the follow error:
```
Loading specs from directory:  .
Downloading plugin from: https://github.com/cloudquery/cloudquery/releases/download/plugins/source/test/development/test_darwin_arm64.zip to: .cq/plugins/github/cloudquery/test/development/cq-source-test_darwin_arm64.zip 
Error: failed to sync source test: failed to get source plugin client for test: failed to download plugin: bad status: 404 Not Found. downloading https://github.com/cloudquery/cloudquery/releases/download/plugins/source/test/development/test_darwin_arm64.zip
exit status 1
```

This is because the test plugin doesn't embed the version correctly when released (and probably others too), thus reports a `development` version which is embedded in the config:
```yml
kind: "source"
spec:
  # Name of the plugin.
  name: "test"

  # Version of the plugin to use.
  version: "development"
...
```

Then the CLI tries to download that version from GitHub and fails.
Since we already know the version from GitHub I suggest we use that one instead of the gRPC call. For local/gRPC plugins the version is not important (at the moment).
The version is already set in https://github.com/cloudquery/cloudquery/blob/7a83a65446119b5339f5f3a3759f7f160a3716b4/cli/internal/plugins/plugins.go#L137 so I only needed to pass in a pointer

---

- [ ] Update or add tests 🧪

### DIFF
--- a/cli/cmd/generate_test.go
+++ b/cli/cmd/generate_test.go
@@ -16,7 +16,7 @@ spec:
   name: "test"
 
   # Version of the plugin to use.
-  version: "development"
+  version: "v1.1.4"
 
   # Registry to use (one of "github", "local" or "grpc").
   registry: "github"
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -80,7 +80,7 @@ func genSource(cmd *cobra.Command, path string, pm *plugins.PluginManager, regis
 	}
 	sourceSpec.SetDefaults()
 
-	plugin, err := pm.NewSourcePlugin(cmd.Context(), sourceSpec)
+	plugin, err := pm.NewSourcePlugin(cmd.Context(), &sourceSpec)
 	if err != nil {
 		return fmt.Errorf("failed to create source plugin: %w", err)
 	}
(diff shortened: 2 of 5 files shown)


yevgenypats (approved):

cli/cmd/generate_test.go
```diff
@@ -16,7 +16,7 @@ spec:
   name: "test"
 
   # Version of the plugin to use.
-  version: "development"
+  version: "v1.1.4"
```
erezrokah, line 19: The tests shows the bug actually.
I'll update this test in https://github.com/cloudquery/cloudquery/pull/1848 to avoid having a hard coded version as that will break when we release the test plugin

PR #1861, opened 2022-09-19
fix: Azure policies

disq:
Seems to be running but results (or some rewrites) need checking. See TODO items.


plugins/source/azure/policies/queries/account/locations_without_network_watchers.sql
```diff
@@ -7,7 +7,7 @@ select
   l.subscription_id,
   l.id,
   case
-    when anw.cq_id is null then 'fail' else 'pass'
+    when anw._cq_id is null then 'fail' else 'pass'
   end
-from azure_account_locations l
-  left join azure_network_watchers anw on l.name = anw.location
\ No newline at end of file
+from azure_subscriptions_locations l
```
SCKelemen, line 12: Should this be
```suggestion
from azure_subscription_locations l
```

plugins/source/azure/policies/hipaa_hitrust_v9.2/network_controls.sql
```diff
@@ -19,7 +19,7 @@
 \ir ../queries/keyvault/vaults_with_no_service_endpoint.sql
 \set check_id '0866.09m3Organizational.1516 - 09.m - 1'
 \echo :check_id
-\ir ../queries/storage/accounts_with_not_restricted_asscess.sql
+\ir ../queries/storage/accounts_with_not_restricted_access.sql
```
SCKelemen, line 22: I feel like this should be either:
```suggestion
\ir ../queries/storage/accounts_with_unrestricted_access.sql
```
or
```
accounts_without_restricted_access
```

and probably the first

plugins/source/azure/policies/queries/account/locations_without_network_watchers.sql
disq: Unfortuately your suggestion is not the correct table name in the current state of the azure plugin.

SCKelemen: Thats what I figured. Thanks.

plugins/source/azure/policies/hipaa_hitrust_v9.2/network_controls.sql
disq: Fixed

erezrokah (approved):

PR #1868, opened 2022-09-19
chore: Update SDK

disq:


erezrokah (approved):

plugins/source/heroku/main.go
```diff
@@ -6,7 +6,5 @@ import (
 )
 
 func main() {
-	serve.Serve(serve.Options{
-		SourcePlugin: plugin.Plugin(),
-	})
+	serve.Source(plugin.Plugin())
```
yevgenypats, line 9: Should we put sentry here as well?

yevgenypats (approved):

plugins/source/heroku/main.go
disq: eventually yes, but there's an issue with sentry login

erezrokah: Going to merge this PR to unblock and we can add the sentry DSNs later

PR #1877, opened 2022-09-19
fix: Small fixes to example.yml and typos in cloudtrail errors

hermanschaaf:


plugins/source/aws/resources/plugin/example.yml
```diff
@@ -16,27 +16,27 @@ aws_debug: false
...
-#  # If you are assuming a role in a partition other than the AWS commercial region,
-#  # it is important that this attribute is set.
-#  default_region: < DEFAULT_REGION >
-#  # Optional. Sync only specified regions. Default assumes all regions.
-#  # This per-account setting overrides the global `regions` setting, if any.
-#  regions:
-#    - us-east-1
-#    - us-west-2
-#    # - ...
+  ## Optional account name to use in logs. Defaults to account id.
```
hermanschaaf, line 19: I've found this comment format work better with IDEs; when uncommenting the lines are left at the correct indentation.

hermanschaaf:
@erezrokah oh actually, these docs files should go under `tables`, shouldn't they? In that case we'll break the links we have from the website right now. Maybe best to leave it for now?

hermanschaaf:
@erezrokah I've removed the docs updates from this PR for now, we can revisit again soon.

yevgenypats (approved):

PR #1881, opened 2022-09-19
feat(aws): Remove client side limiter

yevgenypats:
This should solve the following non-retriable errors: `failed to get rate limit token` (https://github.com/aws/aws-sdk-go-v2/blob/c363cb8290bd1cb29739c24cd7f1d873180c552c/aws/retry/standard.go#L242)

Right now the SDK has two mechanism of limits: Client Side and Server Side:
Let's say a server side returns a throttle error -> Client side will wait number amount of seconds and will also remove a counter from the bucket on the client side. if the bucket will be empty the sdk will error out and maxRetrier (maxAttempts) wont be called at all. Given that we don want to limit calls on the bucket side we should set bucketLimiter to NoOp (or at least expose the total size of the bucket though I don't see a reason for that right now). 

 I think this should also work instead of this - https://github.com/cloudquery/cloudquery/pull/1878 as at least from what I observed it achieves the same thing but just with one option.

Additionally, this might remove the need for per region client which should simplify the client by a lot (this assumption need more testing)

---

- [ ] Update or add tests 🧪


plugins/source/aws/client/client.go
```diff
@@ -415,9 +416,21 @@ func getAccountId(ctx context.Context, awsCfg aws.Config) (*sts.GetCallerIdentit
...
 		config.WithDefaultRegion(defaultRegion),
-		config.WithRetryer(newRetryer(logger, awsConfig.MaxRetries, awsConfig.MaxBackoff)),
+		// https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/
+		config.WithRetryer(func() aws.Retryer {
+			// return retry.NewAdaptiveMode()
+			return retry.NewStandard(func(so *retry.StandardOptions) {
+				so.MaxAttempts = maxAttempts
+				so.RateLimiter = &NoRateLimiter{}
+			})
+			// return retry.AddWithMaxAttempts(retry.NewStandard(), 5)
```
hermanschaaf, line 432: Should we remove these commented out lines?

hermanschaaf (approved):
Approving, looks good from a cursory inspection. But I think it would be great if @bbernays also had a look at this

yevgenypats:
Discussed with @bbernays . We are going to try and stay inline with aws SDK functionality as this functionality detects outages and stops retries. We are going to expose the bucket_size parameter so user can increase it and if we see people are affected from this, we can change the default or make the bucket size infinite (for default).

Issue #1882, opened 2022-09-19
AWS Cloudtrail should be multiplexed on region and account


bbernays:
`aws_cloudtrail_trails` I think should be multiplexed on a region as well as an account

current: `client.AccountMultiplex`

PR #1892, opened 2022-09-20
fix(aws): Fix cloudtrail status column not defined

yevgenypats:
This should fix undefined status field in cloudtrail. in the future we can also unnest it but given it seems it was never defined and was silently ignored we can do it like this for now (though need to go back in the old code). In the future we can add support for code gen to generate tables from multiple structs.

Closes https://github.com/cloudquery/cloudquery/issues/1882


bbernays:
Can we also fix [this issue](https://github.com/cloudquery/cloudquery/issues/1882) in this pr?

hermanschaaf:
@yevgenypats We can generate tables from multiple structs by putting them inside a nested struct wrapper. We already have a `CloudtrailWrapper` here (https://github.com/cloudquery/cloudquery/pull/1892/files#diff-2442ebd6411401fb7e355875880a50564c7aadea24d459a9137d7a6591e47d54L16), so we could add the status struct as a nested struct there to get the additional fields generated.

yevgenypats:
> Can we also fix [this issue](https://github.com/cloudquery/cloudquery/issues/1882) in this pr?

should be done.

hermanschaaf (approved):

PR #1902, opened 2022-09-20
fix: Fix date parsing in Credential Report CSVs (v2)

hermanschaaf:
This fixes a couple of issues with credential reports:
 - Credential Report CSVs can have values like `N/A` and `not_supported`, which previously failed to parse as dates, but now will result in nil columns
 - Some columns were strings when they should have been dates

This was quite hard to fix, as it required creation of a custom type to the unmarshal CSV date, as well as a custom resolver to convert this to a normal timestamp. I also added a custom test to test different scenarios, including one where we expect certain columns to be nil, but want to check that date parsing still works.


disq (approved):
LGTM!

PR #1911, opened 2022-09-21
fix: AWS policies

disq:


plugins/source/aws/policies/queries/iam/unused_creds_disabled.sql
```diff
@@ -4,15 +4,13 @@ select
   :'framework',
   :'check_id',
   'Ensure credentials unused for 90 days or greater are disabled (Scored)',
-  account_id,
-  arn,
+  split_part(r.arn, ':', 5) as account_id,
+  r.arn,
   case when
-      (password_enabled and password_last_used < (now() - '90 days'::INTERVAL)
```
amanenk, line 10: this check wont work if password was never used because `password_last_used ` will be `NULL`

```
SELECT CASE WHEN (NULL < (now() - '90 days'::INTERVAL)) 
THEN 'fail'
ELSE 'pass'
  END
```

 `password_last_changed < (now() - '45 days'::INTERVAL)  ` check should be added to queries like this.

disq: @amanenk how about adding `AND password_last_changed IS NOT NULL`? but we already check for `password_status` being truthful, which I'm guessing is to check if the password is set?

erezrokah (approved):

Issue #1815, opened 2022-09-14
EC2 Reserved Instances


pseymournutanix:
### Is your feature request related to a problem? Please describe.
I would like to determine reserved EC2 (not Elasticcache) instances and numbers etc.

### Describe the solution you'd like
Details of EC2 reserved instances. If this exists and I have missed it - apologies.

### Describe alternatives you've considered
Using the AWS CLI for example:-
```
aws ec2 describe-reserved-instances
```
Shows the details not sure if this can be tied in with instances etc.

Thanks.

hermanschaaf:
Thanks for reporting this @pseymournutanix! And apologies for the delay in getting back to you, we are hard at work getting a new v2 version ready for release.

It seems that you are right: we are missing information about reserved instances right now. Here is an in-progress PR to add it, currently planned for post-v2. Feel free to comment and/or track progress there: https://github.com/cloudquery/cloudquery/pull/1912

PR #1912, opened 2022-09-21
feat: Add support for EC2 reserved instances

hermanschaaf:
This adds support for fetching EC2 reserved instances. 

It is meant to address https://github.com/cloudquery/cloudquery/issues/1815, but please note that since this change is being applied to (the as-of-yet unreleased) CloudQuery v2, it will require upgrading to the new version once it is released.

I couldn't see a way to paginate through these results, so I hope that the API returns the full resultset for each region in the first call.


plugins/source/aws/resources/services/ec2/reserved_instances_fetch.go
```diff
@@ -0,0 +1,43 @@
...
+	svc := c.Services().EC2
+
+	// this API does not seem to support any form of pagination
+	output, err := svc.DescribeReservedInstances(ctx, &config, func(options *ec2.Options) {
+		options.Region = c.Region
+	})
+	if err != nil {
+		return err
+	}
+	for _, instance := range output.ReservedInstances {
```
disq, line 26: res <- output.ReservedInstances

plugins/source/aws/resources/services/ec2/reserved_instances_fetch.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func fetchEc2ReservedInstances(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	var config ec2.DescribeReservedInstancesInput
+	c := meta.(*client.Client)
+	svc := c.Services().EC2
+
+	// this API does not seem to support any form of pagination
+	output, err := svc.DescribeReservedInstances(ctx, &config, func(options *ec2.Options) {
+		options.Region = c.Region
```
disq, line 21: We shouldn't need to set the region here.

disq (changes requested):

bbernays (changes requested):
I think you need to add the function to `plugins/source/aws/resources/plugin/tables.go` and then add the docs

plugins/source/aws/resources/services/ec2/reserved_instances_mock_test.go
```diff
@@ -0,0 +1,32 @@
+package ec2
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2Types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
```
disq, line 7: nit:
```suggestion
	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
```

disq (approved):

plugins/source/aws/resources/services/ec2/reserved_instances_mock_test.go
hermanschaaf: Fixed here and in a number of other places

yevgenypats (approved):

PR #1915, opened 2022-09-21
fix: Can't gen with specific version

erezrokah:
#### Summary

`gen` With a specific plugin version doesn't work, which means we can't test pre-releases.

e.g. `gen source heroku@v0.1.0-pre.0` fails.

This PR fixes it

---

- [ ] Update or add tests 🧪

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -77,9 +77,6 @@ func getSourceSpec(path string, registry specs.Registry) specs.Source {
 		version := "latest"
 		if len(versionParts) > 1 {
 			version = versionParts[1]
-			if !strings.HasPrefix(version, "v") {
-				version = "v" + version
-			}
 		}
 		return specs.Source{
 			Name:     name,
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -63,21 +63,42 @@ func runGenerate(cmd *cobra.Command, args []string) error {
 	}
 }
 
-func genSource(cmd *cobra.Command, path string, pm *plugins.PluginManager, registry specs.Registry, outputFile string) error {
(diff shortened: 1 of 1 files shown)


cli/cmd/generate.go
```diff
@@ -63,21 +63,42 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
-	if registry == specs.RegistryGithub && !strings.Contains(path, "/") {
-		path = "cloudquery/" + path
-	}
-	version := "latest"
-	if strings.Contains(path, "@") {
-		version = strings.Split(path, "@")[1]
-	}
-
-	sourceSpec := specs.Source{
+func getSourceSpec(path string, registry specs.Registry) specs.Source {
```
erezrokah, line 66: This needs a test, I'll do it in a follow up PR as this blocks testing pre-releases

cli/cmd/generate.go
```diff
@@ -63,21 +63,42 @@ func runGenerate(cmd *cobra.Command, args []string) error {
...
+		versionParts := strings.Split(nameParts[1], "@")
+
+		org := nameParts[0]
+		name := versionParts[0]
+		version := "latest"
+		if len(versionParts) > 1 {
+			version = versionParts[1]
+			if !strings.HasPrefix(version, "v") {
+				version = "v" + version
+			}
```
hermanschaaf, line 82: hmm, I'm wondering whether this is wise--should we maybe rather let it fail if you don't set the `v`? Otherwise maybe a plugin producer can release without the `v` and then you can't use that tag

hermanschaaf (approved):
LGTM, just left a passing thought 🙂

cli/cmd/generate.go
erezrokah: Yeah good point. Lets leave it out for now, and we can always add it later

PR #1935, opened 2022-09-21
fix: Autofilling aws plugin default config values

### LINKED ISSUE #1884: Default Values in Config Are ignored
When a user doesn't set a value we chose default values that differed from the AWS SDK. We either need to get this `default` working or change the documentation to show we are deferring to AWS for the default values (ie 3 retries and 10 s for backoff)

```
	MaxRetries   int       `json:"max_retries,omitempty" default:"10"`
	MaxBackoff   int       `json:"max_backoff,omitempty" default:"30"`
```

amanenk:
#### Summary

closes #1884

---

- [ ] Update or add tests 🧪


yevgenypats (dismissed):
I think we shouldn't use the default library anymore and set default via code. 

I addressed it here https://github.com/cloudquery/cloudquery/pulls so no need to reimplement.

amanenk:
> I think we shouldn't use the default library anymore and set default via code.
> 
> I addressed it here https://github.com/cloudquery/cloudquery/pulls so no need to reimplement.

I have set defaults via yaml example. no go-defaults usage anymore. if there is not retryer configured it will use the default one

hermanschaaf (approved):

PR #1944, opened 2022-09-21
fix(aws): Remove error parsing from aws

yevgenypats:
Found more dead code we need to remove.

We dont hide errors and log everything in error level via sdk

There some bits of initial documentation here as I started with that and then stumbled upon the quite a bit of code we need to remove.

---

- [ ] Update or add tests 🧪


plugins/source/aws/CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+
+Welcome to CloudQuery contribution guide for AWS Source plugin!
+
+## Links
+
+- Homepage: [https://cloudquery.io](https://cloudquery.io)
+- Documentation: [https://docs.cloudquery.io](https://docs.cloudquery.io)
+- Tables: [./docs/tables/README.md](./docs/tables/README.md)
+- Discord: [https://cloudquery.io/discord](https://cloudquery.io/discord)
+- [Open AWS Issues](https://github.com/cloudquery/cloudquery/issues?q=is%3Aopen+is%3Aissue+label%3Aaws)
```
erezrokah, line 11:
```suggestion
- [Homepage](https://www.cloudquery.io)
- [Documentation](https://www.cloudquery.io/docs)
- [Tables](./docs/tables/README.md)
- [Discord](https://www.cloudquery.io/discord)
- [Open AWS Issues](https://github.com/cloudquery/cloudquery/issues?q=is%3Aopen+is%3Aissue+label%3Aaws)
```

Not sure where we landed with tables docs, but that link is broken at the moment

plugins/source/aws/README.md
```diff
@@ -0,0 +1,350 @@
...
+CloudQuery Destination plugins
+
+## Links
+
+- Homepage: [https://cloudquery.io](https://cloudquery.io)
+- Documentation: [https://docs.cloudquery.io](https://docs.cloudquery.io)
+- Tables: [./docs/tables/README.md](./docs/tables/README.md)
+- Contributing: [./docs/contributing/README.md](./docs/contributing/README.md)
+- Discord: [https://cloudquery.io/discord](https://cloudquery.io/discord)
+
```
erezrokah, line 13: Same suggestion as before. Also should we duplicate this information?

plugins/source/aws/client/spec.go
```diff
@@ -0,0 +1,40 @@
...
+	ChildAccountExternalID      string   `json:"member_external_id,omitempty"`
+	ChildAccountRegions         []string `json:"member_regions,omitempty"`
+}
+
+type Spec struct {
+	Regions      []string  `json:"regions,omitempty"`
+	Accounts     []Account `json:"accounts,omitempty"`
+	Organization *AwsOrg   `json:"org,omitempty"`
+	Debug        bool      `json:"debug,omitempty"`
+	MaxAttempts  int       `json:"max_attemps,omitempty"`
```
erezrokah, line 29:
```suggestion
	MaxAttempts  int       `json:"max_attempts,omitempty"`
```

Is this spelling error intended?

erezrokah (changes requested):
Added a few comments on the doc, maybe worth separating the code removal from the docs changes?

plugins/source/aws/README.md
yevgenypats: we can remove the duplicates at just keep tables ? wdyt ? or just keep convention and then we dont need the links.

erezrokah: Yeah, maybe keep the links in the README.md?

plugins/source/aws/client/spec.go
```diff
@@ -0,0 +1,40 @@
+package client
+
+type Account struct {
+	Name           string   `json:"name,omitempty"`
```
bbernays, line 4: Should name be required?

plugins/source/aws/resources/services/lambda/functions_fetch.go
```diff
@@ -35,13 +35,6 @@ func fetchLambdaFunctions(ctx context.Context, meta schema.ClientMeta, parent *s
 			})
 
 			if err != nil {
-				if c.IsNotFoundError(err) || c.IsAccessDeniedError(err) {
-					c.Logger().Warn().Err(err).Msg("Failed to get function")
-					res <- &lambda.GetFunctionOutput{
-						Configuration: &f,
-					}
-					continue
-				}
```
bbernays, line 44: Without this, a single permission error or function not found will stop all fetching for the entire region

bbernays (commented):
Left some comments. Overall we really need to consider if we should stop fetching on 403/404/NotFound errors... Because it will make it much harder for users to grab data as in order to grab data about any resource you have to 0 errors for the entire resource

plugins/source/aws/client/client.go
```diff
@@ -373,50 +371,45 @@ func verifyRegions(regions []string) error {
...
+		if region != "*" && !availableRegions[region] {
+			return fmt.Errorf("unsupported region %s", region)
+		}
+	}
+
 	// validate regions values
 	var hasWildcard bool
 	for i, region := range regions {
+		if region == "*" && hasWildcard {
+			return fmt.Errorf("region * can appear only once")
```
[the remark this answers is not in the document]
yevgenypats: In Go it's customary to have different error for different places otherwise you can't differentiate where the error came from because this how wrapping works unless we go into the rabbit whole of adding line numbers which is not "Go idiomatic"

plugins/source/aws/resources/services/lambda/functions_fetch.go
yevgenypats: > Left some comments. Overall we really need to consider if we should stop fetching on 403/404/NotFound errors... Because it will make it much harder for users to grab data as in order to grab data about any resource you have to 0 errors for the entire resource

It should be the default behaviour in the SDK otherwise will be implementing error parsing all the time and running after more 404 and so on. So basically all resolvers should be called no matter what and for list/detail we should use the preResourceResolver

bbernays: So should that change be a blocker for v2 release? If we leave this pr as it is many users won't be able to grab resources

yevgenypats:
Going to open a new PR with error removals only as docs were already updated

PR #1985, opened 2022-09-22
feat: Use new tag format to download plugins

erezrokah:
#### Summary

Updates the CLI to take plugins versions from the `v2` versions site which has a different structure for tags.
It still tries to download using the old format too to not break existing configurations.
Once we release all plugins under the v2 site we can remove the fallback code

---

- [ ] Update or add tests 🧪

### DIFF
--- a/cli/cmd/generate_test.go
+++ b/cli/cmd/generate_test.go
@@ -16,7 +16,7 @@ spec:
   name: "test"
 
   # Version of the plugin to use.
-  version: "v1.1.4"
+  version: "v1.1.5"
 
   # Registry to use (one of "github", "local" or "grpc").
   registry: "github"
--- a/cli/internal/plugins/plugins.go
+++ b/cli/internal/plugins/plugins.go
@@ -54,7 +54,7 @@ type downloadPaths struct {
 	destFile    string
 	zip         string
 	binaryInZip string
-	url         string
+	urls        []string
 }
 
 func (p *DestinationPlugin) Close() error {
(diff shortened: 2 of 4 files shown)


cli/cmd/generate_test.go
```diff
@@ -16,7 +16,7 @@ spec:
   name: "test"
 
   # Version of the plugin to use.
-  version: "v1.1.4"
+  version: "v1.1.5"
```
erezrokah, line 19: Every time we release a new test plugin we need to update this, it's fixed in https://github.com/cloudquery/cloudquery/pull/1848/files#diff-d47c553087688ec44981249565e43b874cdc5f189894bd10829dcf0626831af0R53 by check the version is a valid SemVer one instead of a string comparison

cli/internal/plugins/plugins.go
```diff
@@ -143,45 +143,52 @@ func getBinarySuffix() string {
...
+	}
 	// we use convention over configuration and we use github as our registry. Similar to how terraform and homebrew work.
 	// For example:
 	// https://github.com/cloudquery/cloudquery/releases/download/plugins-source-test-v1.1.0/test_darwin_amd64.zip
 	if org != "cloudquery" {
 		// https://github.com/yevgenypats/cq-source-test/releases/download/v1.0.0/cq-source-test_linux_amd64.zip
-		downloadURL = fmt.Sprintf("https://github.com/%s/cq-source-%s/releases/download/%s/cq-source-%s_%s_%s.zip", org, repo, version, repo, runtime.GOOS, runtime.GOARCH)
-		binaryPathInZip = "cq-source-" + repo + binarySuffix
+		downloadURLs = []string{
+			fmt.Sprintf("https://github.com/%s/cq-source-%s/releases/download/%s/cq-source-%s_%s_%s.zip", org, name, version, name, runtime.GOOS, runtime.GOARCH),
```
erezrokah, line 165: There's a bit of negative impact here as if the config points to a plugin with an old tag structure we print this log message twice

PR #1992, opened 2022-09-22
feat: Add first draft of AWS v2 migration guide

hermanschaaf:
This adds an automatically-generated schema migration guide to show which tables and columns changed, and how. It makes a reasonable effort to summarize how things moved, but it might not be 100% correct in all cases. In these cases, we can manually override with `CustomTableComments`. 

You can view the generated output here: https://github.com/cloudquery/cloudquery/blob/aws-v2-migration-guide/plugins/source/aws/docs/v2-migration.md


scripts/v2-migration/internal/convert/pg.go
```diff
@@ -0,0 +1,91 @@
+package convert
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func SchemaTypeToPg(t schema.ValueType) (string, error) {
```
hermanschaaf, line 10: This was copied from the postgresql package, because I can't import from an `internal` package elsewhere.

scripts/v2-migration/main.go
```diff
@@ -0,0 +1,300 @@
...
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/cloudquery/cloudquery/scripts/v2-migration/internal/convert"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/gertd/go-pluralize"
+)
+
+var (
```
hermanschaaf, line 20: This is a once-off script that we'll throw away in a week or two, so I didn't spend too much time cleaning it up. Please forgive stylistic issues :bow:

yevgenypats (approved):

PR #1994, opened 2022-09-22
fix: Rename accessanalyzer -> access_analyzer (v2)

hermanschaaf:


yevgenypats (approved):

hermanschaaf:
Actually, `accessanalyzer` would be more consistent with all the other AWS resources. `elasticsearch`, `elasticbeanstalk`, `directconnect`, `codepipeline`, etc. This is our chance to make it consistent: I'm going to leave it without the underscore and mark it as renamed.

PR #1998, opened 2022-09-22
fix: Remove underscores in ec_2, s_3 and others

hermanschaaf:
Generating names without underscores separating numbers seems like a better default. I'm on the fence about `sha1` and `sha256` maybe needing underscores, but I don't think it's worth making exceptions for those one or two columns 🤷


erezrokah:
Linking https://github.com/cloudquery/plugin-sdk/issues/159 for context as we might want to use the approach in this PR to work around that issue

Issue #1947, opened 2022-09-22
Investigation request(Azure): Wrong subscriptions id


erezrokah:
### Describe the Bug

Initially reported on Discord ([private link](https://discord.com/channels/872925471417962546/963472127699329074/1019326425960038460))

When fetching Azure subscriptions via the `azure_subscription_subscriptions` (`azure_subscriptions` in v2), the `id` column doesn't match the `subscription_id` column.

### Expected Behavior

Values should match

### CloudQuery Version

[v0.32.12](https://github.com/cloudquery/cloudquery/releases/tag/cli%2Fv0.32.12)

### Steps to Reproduce

Not sure how to reproduce, we might need additional information from the user and also test on v2

PR #2018, opened 2022-09-22
fix: Don't multiplex subscriptions

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/1947

We shouldn't be multiplexing subscriptions as the API returns all subscriptions. So we were basically getting all subscriptions for each subscription, and then using the client subscription id to set the value in the DB. Since we use `ID` as the primary key and do an upsert (at least in v2), the last operation is the one that "took".

~~I think this also means subscription locations should not be multiplexed either and be a relation of subscriptions.
Most of the changes are to support relations in the new SDK list pager API.~~

Removed the relations stuff to simplify things

~~Still a draft as needs some cleanup, which I'll do next week~~

Done

---

- [x] Update or add tests 🧪


erezrokah:
> I think this also means subscription locations should not be multiplexed either and be a relation of subscriptions. Most of the changes are to support relations in the new SDK list pager API.

Not sure about that, maybe since we have the subscriptions already we can make it a top level table

plugins/source/azure/resources/services/subscriptions/locations.go
```diff
@@ -7,18 +7,20 @@ import (
...
 		Name:      "azure_subscriptions_locations",
 		Resolver:  fetchSubscriptionsLocations,
-		Multiplex: client.SubscriptionMultiplex,
+		Multiplex: client.SingleSubscriptionMultiplex,
 		Columns: []schema.Column{
 			{
-				Name:     "subscription_id",
-				Type:     schema.TypeString,
-				Resolver: client.ResolveAzureSubscription,
+				Name:     "subscriptions_subscription_id",
```
erezrokah, line 21: This is a bit confusing but `ParentIDResolver` get the CQ Id

plugins/source/azure/resources/services/subscriptions/subscriptions.go
```diff
@@ -58,17 +53,27 @@ func Subscriptions() *schema.Table {
 				Type:     schema.TypeString,
 				Resolver: schema.PathResolver("State"),
 			},
+			{
+				Name:     "subscription_id",
```
erezrokah, line 57: I'll probably revert this and use `ResolveAzureSubscription` with `SingleSubscriptionMultiplex`

plugins/source/azure/resources/services/subscriptions/locations.go
```diff
@@ -53,13 +55,25 @@ func Locations() *schema.Table {
...
+				Name:     "subscription_id",
+				Type:     schema.TypeUUID,
+				Resolver: resolveSubscriptionID,
+			},
 		},
 	}
 }
 
-func fetchSubscriptionsLocations(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- interface{}) error {
+func resolveSubscriptionID(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
erezrokah, line 67: Not sure why, but getting the subscription ID directly from the location sets a `nil` value in the database.
I'll look into it and probably revert this custom resolver

plugins/source/azure/resources/services/subscriptions/subscriptions.go
```diff
@@ -13,13 +13,8 @@ func Subscriptions() *schema.Table {
 	return &schema.Table{
 		Name:      "azure_subscriptions",
 		Resolver:  fetchSubscriptionsSubscriptions,
-		Multiplex: client.SubscriptionMultiplex,
+		Multiplex: client.SingleSubscriptionMultiplex,
 		Columns: []schema.Column{
-			{
-				Name:     "subscription_id",
```
erezrokah, line 19: Subscriptions already have the id in `ID` so no need for `subscription_id`

plugins/source/azure/resources/services/subscriptions/locations.go
erezrokah: Done

erezrokah:
@hermanschaaf marked this as ready and made you reviewer :) I mad the minimal changes to fix the bug and I think we can do the relations bit if we feel that's the right approach, or required by users.

hermanschaaf (approved):
@erezrokah Looks good 👍 I couldn't test it locally the other day due to some unrelated local environment issues, but this fixes the bug so let's not block on that--we can always revisit if something still doesn't work right.

PR #2020, opened 2022-09-22
feat: Provide a decoded policy document field inside aws_iam_policies.policy_version_list

disq:
~This PR adds a `DecodedDocument *string` field inside the `policy_version_list []types.PolicyVersion` field.~

[PolicyVersion type](https://docs.aws.amazon.com/IAM/latest/APIReference/API_PolicyVersion.html) includes an URL-encoded policy document, which is no use to run policies with.

The alternative is [this](https://github.com/cloudquery/cloudquery/pull/1911/commits/c07194454719c1db07ecb474f784e07a187d51d5) which is extra slow.


hermanschaaf (changes requested):
Nice catch! I think we should probably decode the Document and place it back in the `Document` field, rather than introducing a non-standard `DecodedDocument` field that users will not be able to find in the documentation. This seems okay, we're doing the minimal transformation to make the data useful. What do you think?

disq:
> place it back in the `Document` field

My concern here would be an invalid/undecodable document and what to do with its value.

yevgenypats:
> Nice catch! I think we should probably decode the Document and place it back in the `Document` field, rather than introducing a non-standard `DecodedDocument` field that users will not be able to find in the documentation. This seems okay, we're doing the minimal transformation to make the data useful. What do you think?

This seems right. 

@disq What do you mean by invalid?

hermanschaaf:
> My concern here would be an invalid/undecodable document and what to do with its value.

I think we can treat that as an error: it's AWS not upholding its contract with us, since the docs for `PolicyVersion` explicitly say: 

```
The policy document returned in this structure is URL-encoded compliant with RFC 3986 (https://tools.ietf.org/html/rfc3986)
```

disq:
by invalid I thought wrong specified policies would lead to failure to decode, but that obviously can't happen because we're not trying to unmarshal the json (the user supplied...) but merely decoding something from the protocol. So yeah, it shouldn't happen.

plugins/source/aws/resources/services/iam/policies_fetch.go
```diff
@@ -41,3 +42,15 @@ func resolveIamPolicyTags(ctx context.Context, meta schema.ClientMeta, resource
...
 	return resource.Set("tags", client.TagsToMap(response.Tags))
 }
+
+func resolveIamPolicyVersionList(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	r := resource.Item.(types.ManagedPolicyDetail)
+	for i := range r.PolicyVersionList {
+		if v, err := url.PathUnescape(aws.ToString(r.PolicyVersionList[i].Document)); err == nil {
+			r.PolicyVersionList[i].Document = &v
+		} else {
+			meta.Logger().Warn().Err(err).Msg("Failed to unescape policy document, leaving as-is")
```
hermanschaaf, line 52: nit: We may want to put some information about the policy ID or something in this log message, so it's possible to locate & fix it.

hermanschaaf (approved):

plugins/source/aws/resources/services/iam/policies_fetch.go
disq: added

PR #2036, opened 2022-09-23
chore: Fix some typos and missing docs across aws, azure, gcp

hermanschaaf:


plugins/source/aws/resources/services/redshift/clusters_fetch.go
```diff
@@ -20,7 +20,7 @@ func fetchRedshiftClusters(ctx context.Context, meta schema.ClientMeta, parent *
 		if err != nil {
 			return err
 		}
-		res <- response.Clusters
```
disq, line 23: This change doesn't seem right

disq (changes requested):

plugins/source/aws/resources/services/redshift/clusters_fetch.go
hermanschaaf: ah thanks 😮‍💨

hermanschaaf: Reverted 👍

hermanschaaf: Time for me to take a break from the screen I think 😄

plugins/source/aws/README.md
```diff
@@ -73,9 +73,10 @@ export AWS_PROFILE=myprofile
 or, configure your desired profile in the `local_profile` field:
 
 ```yaml title="aws.yml"
-accounts:
-  id: <account_alias>
-  local_profile: myprofile
+...
```
yevgenypats, line 76: I think we should have a convention in our docs where we don't indent and plugin documentation are always assumed it will be under the `spec`, otherwise it just adds boilerplate code and `...` will be in every snippet

plugins/source/aws/docs/tutorials/multi_account.md
```diff
@@ -100,22 +100,23 @@ org:
...
-          local_profile: provider
-          // Optional. Role ARN we want to assume when accessing this account
-          role_arn: <YOUR_ROLE_ARN_2>
+kind: "source"
+spec:
+  name: aws
+  resources:
+    - "*"
+  configuration:
+    accounts:
```
yevgenypats, line 109: Let's start the snippet from here:

yevgenypats (changes requested):
Looks good. few nit comments + conventions I think we need to set around snippet documentation.

plugins/source/aws/README.md
hermanschaaf: Hmm, I think someone is going to be confused if we don't indent (or somehow indicate that it's a partial snippet). But maybe we can go with that convention until the first user comes to Discord asking why their unindented config doesn't work, then talk about it again 😃

plugins/source/aws/docs/tutorials/multi_account.md
hermanschaaf: Updated 👍

plugins/source/aws/README.md
hermanschaaf: Updated 👍

yevgenypats (approved):

PR #2040, opened 2022-09-24
chore: Remove deprecated terraform files

yevgenypats:
Those files are not used and it confuses first time contributors and/or users who looks for guidelines on how to build new plugins.


erezrokah:
Can we move the files to https://github.com/cloudquery/deployments (private repo) instead? It's quite useful to have all these examples for manual testing purposes

yevgenypats:
Im not really sure on the usefullness of that anymore but we can def backup this in the deployment repo. most of those are just copy paste from the terraform registry website .

hermanschaaf (approved):
Agree, let's move it somewhere where we can still access it relatively easily if we need to. Even if they are mostly copy-paste from the terraform website, it can require some fine-tuning to set _all_ the parameters properly.

yevgenypats:
@amanenk mind taking care of that?

amanenk:
@yevgenypats @hermanschaaf  https://github.com/cloudquery/deployments/pull/43

erezrokah:
> Even if they are mostly copy-paste from the terraform website, it can require some fine-tuning to set _all_ the parameters properly.

Fine tuning is my experience too (mostly copy paste didn't work out of the box for me, so I need to take the full examples terraform registry has and trim it to what I wanted).

PR #2042, opened 2022-09-24
chore(pg): Add README

yevgenypats:
Add general and config docs for PostgreSQL destination.


cli/internal/destinations/postgresql/README.md
```diff
@@ -0,0 +1,20 @@
+# CloudQuery PostgreSQL Destination Plugin
+
+This destination plugin let's you sync data from any CloudQuery source to any PostgreSQL compatible database.
+
+Known supported databases:
+
+- PostgreSQL > v10
```
erezrokah, line 7:
```suggestion
- PostgreSQL > v10
```

Do we mean `>= 11`? Or `v10.0.1` is supported to?

cli/internal/destinations/postgresql/README.md
```diff
@@ -0,0 +1,20 @@
+# CloudQuery PostgreSQL Destination Plugin
+
+This destination plugin let's you sync data from any CloudQuery source to any PostgreSQL compatible database.
```
erezrokah, line 3:
```suggestion
This destination plugin let's you sync data from a CloudQuery source to a PostgreSQL compatible database.
```

cli/internal/destinations/postgresql/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Known supported databases:
+
+- PostgreSQL > v10
+
+## PostgreSQL Spec
+
+This is the top level spec used by PostgreSQL Destination Plugin.
```
erezrokah, line 11:
```suggestion
This is the top level spec used by the PostgreSQL destination Plugin.
```

cli/internal/destinations/postgresql/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+- `connection_string` (string) (required)
+
+  Connection string to connect to the database
+
+- `pgx_log_level` (string) (default: "error")
```
erezrokah, line 17:
```suggestion
- `pgx_log_level` (string) (optional, defaults to "error")
```

cli/internal/destinations/postgresql/README.md
erezrokah, line 11: When we write `top level`, what does it mean?

To me it means that the `yaml` file looks like this:

```
kind: destination
connection_string: <>
```

cli/internal/destinations/postgresql/README.md
```diff
@@ -0,0 +1,20 @@
+# CloudQuery PostgreSQL Destination Plugin
+
+This destination plugin let's you sync data from any CloudQuery source to any PostgreSQL compatible database.
+
+Known supported databases:
+
+- PostgreSQL > v10
+
+## PostgreSQL Spec
```
erezrokah, line 9:
```suggestion
## Configuration
```

Can we use configuration instead of `spec`?

erezrokah (changes requested):
Added a few comments, I think the `top level` and `spec` comments could be relevant to other plugins

cli/internal/destinations/postgresql/README.md
yevgenypats: I think v10.0.1 should be supported as well. at least per my quick check locally. I didn't use any new shiny primitives.

cli/internal/destinations/postgresql/README.md
yevgenypats: we can maybe just omit the `top level` or even better just remove this line and go straight to the spec as it's a bit redundant. The user should be aware of our convention and that every plugin has a spec (it is mentioned in our main docs so no need to repeat that).

In any case it means the following:
```
kind: destination
spec:
  name: postgresql
  mode: overwrite
  spec:
     // goes here
     connection_string: stuff..

```

Similar to here - https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/ there is no intro wording that this is a spec and we already have an intro about what is the postgresql destination plugin

cli/internal/destinations/postgresql/README.md
yevgenypats: I think we need to be consistent with the configuration as we use `spec` (we went the k8s way so should be familiar to some users) so would rather leave as is.

cli/internal/destinations/postgresql/README.md
erezrokah: I think the only reason we supported 10 is because of Aurora which now supports 11 cc @bernays

PR #2043, opened 2022-09-24
feat(postgresql): Move to standalone plugin

yevgenypats:
Following changes in this PR:

Moving PostgreSQL to standalone plugin so we can test the workflow of our destinations. We don't want to support at the moment compiled destinations as we want to use and test one workflow only.

Also, I've simplified the plugin manager and downloader given we use convention over configuration and this code should mostly not changed.

Removed unit-tests from versions and made them pure functions for the following reasons:
- They don't maintain any state so it's just makes it harder to invoke with calling NewClient and then the function you want. If in the future we would like to pass an HttpClient then it should pass directly to function rather then to a NewClient as this will cause various race conditions because http client shouldn't be shared between calls (in most cases) 
- the purpose and most of the logic of versions and plugin manager is to download stuff from github so we have to have integration tests for that rather then unit-tests. Once we have integration tests there is not much value to maintaing unit-tests for that. If we don't want to run those tests on every `go test ./...` then we should just add tags to the integration tests (if someone is working on the plugin manager they will need to run integration tests anyway so again not alot of value in maintaing unit-tests for this component imo).


.github/workflows/test_unit.yml
```diff
@@ -26,21 +26,6 @@ jobs:
       matrix: ${{ fromJson(needs.resolve-modules.outputs.matrix) }}
       fail-fast: false
     runs-on: ubuntu-latest
-    services:
```
erezrokah, line 29: Don't we need this to test that we successfully write to the database?

plugins/destination/postgresql/main.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/cloudquery/plugin-sdk/serve"
+	"github.com/cloudquery/plugins/destinations/postgresql/client"
+)
+
+func main() {
+	serve.Destination(client.New())
+}
+
+// https://github.com/cloudquery/cloudquery/releases/download/plugins-source-test-v1.1.0/test_darwin_arm64.zip
+// https://github.com/cloudquery/cloudquery/releases/download/plugins-source-test-v1.1.5/test_darwin_arm64.zip
```
erezrokah, line 13: Can we delete these comments?

erezrokah (changes requested):
I think once we fix the CI errors this should be a good start and we can do the release changes in a follow up PR.

Also we should delete the binary `plugins/destination/postgresql/postgresql` and ignore it.

.github/workflows/test_unit.yml
yevgenypats: I think let's discuss this offline as it's a bigger discussion but every destination plugin will have different requirements for databases and so on (sources dont need any db). so I think we can't automate much here but rather just have one workflow per destination plugin (and I dont think we will have more than 10 so shouldn't be an issue or a lot to automate) even if we will have to copy paste a bit

erezrokah: Maybe keep the tests at least until we have the discussion, otherwise we won't have coverage for the sync command

cli/cmd/sync.go
```diff
@@ -116,20 +123,37 @@ func syncConnection(ctx context.Context, pm *plugins.PluginManager, specReader *
 	format := " Syncing (%d resources) %s"
 	s.Suffix = fmt.Sprintf(format, 0, time.Duration(0))
 	s.Start()
-	failedWrites := 0
+	var failedWrites uint64 = 0
```
hermanschaaf, line 126:
```suggestion
	failedWrites := uint64(0)
```

Absolutely a nit, I just think this style fits better with the `totalResources := 0` declaration below it.

cli/internal/plugins/destination.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (p *DestinationPlugin) GetClient() *clients.DestinationClient {
+	return p.client
+}
+
+
+
+
+func destJsonToLog(msg map[string]interface{}, l zerolog.Logger) {
+	switch msg["level"] {
```
hermanschaaf, line 37:
```suggestion
}

func destJsonToLog(msg map[string]interface{}, l zerolog.Logger) {
	switch msg["level"] {
```

Removing some unnecessary newlines

cli/internal/plugins/manager.go
```diff
@@ -0,0 +1,268 @@
+// pluginmanager take care of lifecycle of plugins.
+// Including: downloading, upgrading, spawning, closing
+// Currently we use github releases as our plugin store.
+package plugins
```
hermanschaaf, line 4:
```suggestion
// pluginmanager takes care of the lifecycle of plugins,
// including downloading, upgrading, spawning and closing. 
// Currently we use GitHub releases as our plugin store.
package plugins
```

hermanschaaf (approved):
LGTM, approving with some minor nit comments

PR #2057, opened 2022-09-26
feat(dest-test): Add new test destination

yevgenypats:
Adding a new test destination for CLI testing purposes. The new destination plugin is basically a dev null.

Also, Im proposing/trying out a .yml per plugin ? maybe this way we can take advantage of the built-in github

```
    paths:
      - "plugins/destination/test/**"
```

For destination plugins I think this is specifaclly important because the workflows vary very much between each plugin as we need different databases and so on. 

I think for source plugin this will also start becoming a thing when we introduce some smoke tests that will require credentials and so on.


.github/workflows/dest_test.yml
```diff
@@ -0,0 +1,32 @@
+name: Destination Plugin Test Workflow
+
+on:
+  pull_request:
+    paths:
+      - "plugins/destination/test/**"
```
hermanschaaf, line 6: I think a yml file per plugin makes sense, but the issue with this type of filter (afaik) is that you can't then make it a requirement for PRs.

hermanschaaf (approved):
Approving, but left a comment regarding how we filter out builds. I would love to find a solution to that problem though

.github/workflows/dest_test.yml
yevgenypats: what a requirement for PR mean? is it just the auto-merge ? can we skip auto-merge all together ? we don't use it that much imo and basically the rule should be just for all actions to be green?

yevgenypats: So I think this wont be a problem if we wont use the feature by github called "mark checks as required" because this feature is half baked and it is by default required already. it's like double require.

hermanschaaf: Correct, it won't be a problem if we don't mark it as required. However, if a check is not required, merging is allowed even if the check fails, so it does serve a purpose. I'd prefer to enforce not allowing merging if e.g. unit tests are failing.

erezrokah, line 6: Regardless where we land, the `paths` is missing `.github/workflows/dest_test.yml`, otherwise changes to the workflow file won't trigger it (for example in a PR)

PR #2058, opened 2022-09-26
chore: PR template

yevgenypats:
Put the steps into comments as otherwise it gets into the commit message when merging and into other notifications which makes it hard to read imo.

It just a nit / my OCD :)


erezrokah (approved):
👍

Issue #1958, opened 2022-09-22
Bug(v1): AWS aws_ec2_instances should have state_transition_reason_time


disq:
Which is regex matched/parsed from the state_transition_reason text to a time struct in the previous provider... Used in policies to determine when the instance was stopped. Or we need to find another way to determine the stop time.

PR #2087, opened 2022-09-26
fix: Re-add AWS EC2 Instance state_transition_reason_time

disq:
Implements https://github.com/cloudquery/cloudquery/issues/1958


yevgenypats (changes requested):
I wouldn't actually have this Transformation here but rather have just `state_transition_reason` column (I think we already have). Is this needed to any policy or we just assumed this is needed in v1 ?

disq:
> I wouldn't actually have this Transformation here but rather have just `state_transition_reason` column (I think we already have). Is this needed to any policy or we just assumed this is needed in v1 ?

@yevgenypats all explained in #1958

hermanschaaf:
Tricky one 😄 This particular case is pretty minor, but it could be the start of a slippery slope... Let's have a discussion about how we want to handle cases like this where we need a non-trivial transformation for use in our policy queries?

yevgenypats (approved):

hermanschaaf (approved):

Issue #1828, opened 2022-09-15
feature(v2): Improve progress bar


erezrokah:
It took me a bit of time to understand that the left counter is total number of resources and the right one is the rate and that's the reason it's decreasing:

https://user-images.githubusercontent.com/26760571/190433528-fe0e3e20-28f2-442d-8735-5ef06fdc0150.mov

I suggest we drop the rate as I don't think it adds much value, and also remove the empty `/-` which I assume should be the total number (that we don't know in advance)

yevgenypats:
This is done.

erezrokah:
Let's discuss this a bit more. I don't think it's done as we still show unrelated information in the counter:
![image](https://user-images.githubusercontent.com/26760571/194744570-7a139535-a588-48f7-9616-1d0d35d66b6d.png)

See the additional `/-` (977 out of something) in the progress.

Removing the `/-` was also suggested in https://github.com/cloudquery/cloudquery/pull/2163

yevgenypats:
I do like the `resource/sec` as it helps understand if the plugin is stuck on slow resources - I don't think we should remove extra information.

Regarding the `977/-` I think it's a nit and it is useful in some sense that we cannot now how much resources there are.
I would say we can def improve it but it is a low priority because we had bunch of issues with the old implementation and it had bigger issues like my terminal stopped working after every run and I had to revert - https://github.com/cloudquery/cloudquery/pull/2089

erezrokah:
Sure so let's close as won't fix due to low priority

PR #2089, opened 2022-09-26
feat: Remove spinner

yevgenypats:
Spinner is already available in a progressbar
 library we use (https://github.com/schollz/progressbar).
Also it incoproates the same spinners available in https://github.com/briandowns/spinner

Also, this fixes a bug where cursor was disappearing in the spinner library


hermanschaaf (approved):
I also tested it out locally, looks good 👍

erezrokah:
The spinner we use displays a lot of redundant information and can't be configured, see screenshot for example:

Progress

I could find a way to remove `resources/sec` with our current spinner, but not the `-/` (out of total) indicator.

This is described in https://github.com/cloudquery/cloudquery/issues/1828.

erezrokah:
Re-opened https://github.com/cloudquery/cloudquery/issues/1828 for further discussion

PR #2093, opened 2022-09-27
chore: Make GH workflow per plugin/cli

yevgenypats:
This is a take at a different approach where we use workflow per plugin. This should give the following advantages:

- native workflow filtering: actions wont run if files not changed which will save lot's of time, GH minutes and money
- we still have the ability to easily customize workflow per plugin when needed. especially true for destinations but also for sources.
- if we have any common functionatliy we can always put it in scripts and then reuse them in the actions rather then colliding the actions.

I think worth doing a live discussion on this tomorrow.


hermanschaaf:
Indeed, this has an effect on kodiak's automated merging abilities (we wouldn't really be able to use it), so let's discuss live tomorrow

yevgenypats:
@erezrokah @hermanschaaf This is ready for first review.

I think this should be a better approach and give the following capabilities:
- keep ability to do automerge with one required check that wait for all other required check (based on the same logic that we created the matrix)
- less minutes, api calls and faster ci
- easy customization of workflows per plugin without the need to understand additional automation on top of github
- we can do even more stuff with auto merge that weren't possible before (though I would personally be careful with the automerge rabbit hole - I understand it's a great feature but it is also complex for a reason so I would keep just for simple things and wont try to over optimize that)

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -0,0 +1,73 @@
...
+          dir_names: true
+      - uses: actions/github-script@v6
+        env:
+          DIRS: ${{ steps.changed-files.outputs.all_changed_files }}
+          REF: ${{ github.ref }}
+        with:
+          script: |
+            const { REF, DIRS } = process.env
+            let now = new Date().getTime()
+            const deadline = now + 50 * 1000
```
hermanschaaf, line 26: Does this mean we only allow 50 seconds for tests to run? We may want to increase that to ~10 minutes.

hermanschaaf (approved):
Looks great, I really like this approach 👍

.github/workflows/validate_pr_title.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 jobs:
   main:
```
erezrokah, line 11:
```suggestion
  validate-pr-title:
```

We can change this and remove the name as the default is the key

.github/workflows/validate_pr_title.yml
```diff
@@ -9,7 +9,7 @@ on:
...
-    name: Validate PR title
+    name: validate-pr-title
```
erezrokah, line 12:
```suggestion
```

We can remove the `name` as the key is the default value

.github/workflows/wait_for_required_workflows.yml
yevgenypats: Ah good catch. 15 min sounds good, anyway if a workflow fails earlier it will stop.

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -0,0 +1,73 @@
...
+            let actions = ["cli",
```
erezrokah, line 27: How about we keep the resolve modules logic we had to get this list?
That should ensure that if someone adds a new module, they are required to add a matching workflow file?

Looks like we can do it with https://github.com/featherweight-design/collect-directory-names-by-file-glob, WDYT?

Also I think we might also want to wait for the checks if someone changes the `source_test.yml` file for example, but I would keep that for another PR as it can complicate this script.

.github/workflows/wait_for_required_workflows.yml
erezrokah, line 26: Same comment as @hermanschaaf on the deadline being too short

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -0,0 +1,73 @@
+name: Wait for all required workflows to pass
+
+on: [pull_request, push]
```
erezrokah, line 3:
```suggestion
on:
  pull_request:
    branches:
      - main
```

Do we need to run this on every push to the repository? Maybe only do it on pushes to `main`, but also since we require branches to be up to date, we can only check PRs I think

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -0,0 +1,73 @@
...
+            while (now <= deadline) {
+              const result = await github.rest.checks.listForRef({
+                owner: 'cloudquery',
+                repo: 'cloudquery',
+                ref: context.sha,
+                status: 'completed',
+                per_page: 100
+              })
+              console.log(result)
+              result.data.check_runs.forEach(item => {
```
erezrokah, line 57:
```suggestion
              const checkRuns = await github.paginate(github.rest.checks.listForRef, {
                owner: 'cloudquery',
                repo: 'cloudquery',
                ref: context.sha,
                status: 'completed',
                per_page: 100
              })
              console.log(checkRuns)
              checkRuns.forEach(item => {
```

I think this should work though we don't expect more than a 100 results, so this comment is non blocking for the PR

erezrokah (changes requested):
Great work @yevgenypats I think this approach should work well 🚀 

Added comments on the implementation, I think mainly adding caching back, fixing the CLI job name (and missing `runs-on`), and inferring the checks from the modules dynamically.

Also I think we can delete https://github.com/cloudquery/cloudquery/blob/1253568397fdca65ea4c967683394fe6d6598a53/.github/workflows/sync_repo_settings.yml too, WDYT?

.github/workflows/lint_doc.yml
```diff
@@ -1,83 +0,0 @@
-name: Lint Docs
```
erezrokah, line 1: I think we want to keep this test to avoid drift in our docs, I couldn't find it in the new approach

erezrokah (changes requested):
Added another comment

erezrokah:
One thing I'm noticing with this approach is we'll need to put all the different tests we have (doc, lint, build, unit) under the same file otherwise we'll end up with A LOT of files which can have a negative impact on our CI time.
Also adding a new test (for example checking for drift in codegen) requires editing quite a lot of files, but that's expected I think

.github/workflows/cli.yml
```diff
@@ -0,0 +1,32 @@
...
+          go-version: 1.19.1
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.49.0
+          args: "--config ${GITHUB_WORKSPACE}/cli/.golangci.yml"
+      - name: Get dependencies
```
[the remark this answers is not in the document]
yevgenypats: I don't think this is relevant anymore as it was caching things between jobs.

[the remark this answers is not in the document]
yevgenypats: Added, but I used native cache option - https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs not sure why implemented that manually before :)

.github/workflows/wait_for_required_workflows.yml
yevgenypats: Regarding the logic that sources that automatically let's do it in a follow-up PR and check that it all works now

.github/workflows/lint_doc.yml
yevgenypats: Let's do it as a follow-up https://github.com/cloudquery/cloudquery/issues/2138

.github/workflows/cli.yml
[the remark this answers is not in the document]
erezrokah: I think caching was only added in `v3` of the setup-actions so maybe that's the reason.
Also how is the cache key determined with the native option? We need to ensure each module uses its own key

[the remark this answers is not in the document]
yevgenypats: There is a special option for monorepos `cache-dependency-path` (I used it)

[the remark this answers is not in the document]
yevgenypats: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

erezrokah (approved):
Lets ship this 🚀

.github/workflows/test_policy_sql.yml
```diff
@@ -1,101 +0,0 @@
-name: SQL Policy Tests
```
erezrokah, line 1: @yevgenypats I think we want to keep this workflow too

Issue #2076, opened 2022-09-26
(code) GetLatestVersion doesn't work in CLI


yevgenypats:
Follow-up on this PR where I commented the test, https://github.com/cloudquery/cloudquery/pull/2074/files 

Seems get latest CLI version stopped working. worth adding this back.

hermanschaaf:
This isn't working because the version file is currently return an empty version:

```
$ curl https://versions.cloudquery.io/v2/cli.json
{ "latest": "" }
```

And it looks like it's deliberate while we're in pre-release: https://github.com/cloudquery/cloudquery/blob/main/.github/workflows/release_cli.yml#L70-L72

I'm trying to understand whether it would be okay to update that file right now, and what the effects will be 🤔

erezrokah:
Yes the version files only get updated for official releases.

But I think it's ok to update those manually since v0 uses `v1` versions site and `v1` (pre-release) uses `v2` versions site.
I created the `v2` versions site per https://github.com/cloudquery/cloudquery/pull/1957 as to not breaking existing versions on the CLi.

TLDR: If we change the versions in `v2` nothing should breaking as the only consumer is the v1 pre-release

erezrokah:
BTW for the new destination plugins, as those were officially releases, the versions are there:
https://github.com/cloudquery/cloudquery/blob/f1a1caa1ac89e38cfb289e15bddf60a4d7710234/sites/versions/v2/destination-postgresql.json#L1
https://github.com/cloudquery/cloudquery/blob/f1a1caa1ac89e38cfb289e15bddf60a4d7710234/sites/versions/v2/destination-postgresql.json#L1

erezrokah:
And if this PR was mostly for the CLI, I think the only consumer of the CLI JSON file is our GitHub action:
https://github.com/cloudquery/setup-cloudquery/blob/1c8abd16eb48f102853bbd8863f84cf2750a60a9/src/main.ts#L16

yevgenypats:
I think this is solved now.

PR #2094, opened 2022-09-27
chore: Update cli latest v2 version to pre-release

hermanschaaf:
This updates `sites/versions/v2/cli.json` to `cli-v1.1.0-pre.1`, which is a pre-release. Normally we wouldn't use pre-releases here, but having this would be better than having an empty string as we can then test the flow of end-to-end (related: https://github.com/cloudquery/cloudquery/issues/2076). It will be updated automatically as soon as we do a real v1 release.

### DIFF
--- a/sites/versions/v2/cli.json
+++ b/sites/versions/v2/cli.json
@@ -1,1 +1,1 @@
-{ "latest": "" }
+{ "latest": "cli-v1.1.0-pre.1" }


yevgenypats (approved):

erezrokah:
Can you explain where we consume this version? I thought the only consumer is the GitHub action

hermanschaaf:
@erezrokah it is used by this code to detect the latest available CLI version: https://github.com/cloudquery/cloudquery/blob/main/cli/internal/plugins/versions.go#L36-L48

AFAIK currently we are not using it in the new CLI, but at some point soon we will be using it to detect whether a new version is available.

erezrokah:
> @erezrokah it is used by this code to detect the latest available CLI version: https://github.com/cloudquery/cloudquery/blob/main/cli/internal/plugins/versions.go#L36-L48

That code is only used in a commented out test I think
https://github.com/cloudquery/cloudquery/search?q=GetLatestCLIRelease

hermanschaaf:
@erezrokah Yes, the test is fixed in this PR: https://github.com/cloudquery/cloudquery/pull/2097

And yeah, we're not using this function right now, but it will be used to check for new versions and then display a message on the CLI if a new version available.

erezrokah:
> And yeah, we're not using this function right now, but it will be used to check for new versions and then display a message on the CLI if a new version available.

Ah makes sense, thanks for clarifying

PR #2097, opened 2022-09-27
fix: Version fetching code and test

hermanschaaf:
This re-instates the test for `GetLatestCLIRelease` and ensures it also works on pre-release versions.


cli/internal/plugins/versions_test.go
```diff
@@ -28,13 +28,31 @@ func TestGetLatestCommunityPluginRelease(t *testing.T) {
 	}
 }
 
-// func TestGetLatestCLIRelease(t *testing.T) {
-// 	version, err := GetLatestCLIRelease(context.Background())
-// 	if err != nil {
-// 		t.Fatalf("error calling GetLatestCLIRelease: %v", err)
-// 	}
+func TestGetLatestCLIRelease(t *testing.T) {
```
hermanschaaf, line 31: The full integration test is still there, to make sure things work end-to-end.

erezrokah (approved):
Great catch 🚀

cli/internal/plugins/versions.go
```diff
@@ -61,11 +63,14 @@ func getLatestCQPluginRelease(ctx context.Context, name string, typ PluginType)
...
 }
 
-// extractVersionFromTag takes a tag of the form "plugins-source-test-v0.1.21" and returns
-// the version, i.e. "v0.1.21"
+// extractVersionFromTag takes a tag of the form "plugins-source-test-v0.1.21" or "cli-v1.1.0-pre.1" and returns
+// the version, i.e. "v0.1.21" or "v1.1.0-pre.1"
 func extractVersionFromTag(tag string) string {
-	parts := strings.Split(tag, "-")
-	return parts[len(parts)-1]
+	m := reVersionFromTag.FindStringSubmatch(tag)
```
yevgenypats, line 69: isn't our versioning and tags are predictable and according to configured format in goreleaser? why do we need a regex? it's usually much harder to understand, debug, slower and so on...

yevgenypats (changes requested):
Looks good. one comment: can we avoid the regex ?

cli/internal/plugins/versions.go
hermanschaaf: @yevgenypats I'm not sure how we would do this reliably without a regex. Or I can think of a solution, but it's not going to be better IMO. We don't know how many dashes there are in the part before the version starts, so we'd have to do something like this (pseudocode):

```
parts = split(tag, "-")
if len(parts) > 1 and parts[-1] matches v{}.{}.{}:
   return parts[-1]
if len(parts) > 2 and parts[-2] matches v{}.{}.{}:
   return parts[-2] + "-" + parts[-1]
return ""
```

We'd still need a regex for the `matches v{}.{}.{}` part there--`hasPrefix("v")` wouldn't cut it as you could have a plugin name that starts with a `v`.

hermanschaaf: @yevgenypats refactored in https://github.com/cloudquery/cloudquery/pull/2097/commits/7ba878310946745c0069356bca862270b62bbe3c, thanks for the suggestion 👍

cli/internal/plugins/versions.go
```diff
@@ -58,14 +62,11 @@ func getLatestCQPluginRelease(ctx context.Context, name string, typ PluginType)
...
-	return extractVersionFromTag(mr.Latest), nil
+	return extractPluginVersionFromTag(mr.Latest, name, typ), nil
 }
 
-// extractVersionFromTag takes a tag of the form "plugins-source-test-v0.1.21" and returns
-// the version, i.e. "v0.1.21"
-func extractVersionFromTag(tag string) string {
-	parts := strings.Split(tag, "-")
-	return parts[len(parts)-1]
+func extractPluginVersionFromTag(tag, name string, typ PluginType) string {
```
yevgenypats, line 68: I don't really think this function is needed and we can inline that but just nit.

yevgenypats (approved):

PR #2035, opened 2022-09-23
feat!: Migrate k8s plugin to v2

shimonp21:
#### Summary

What is left to do for the plugin:
- policies.
- initial config (and test cloudquery generate).
---

Use the following steps to ensure your PR is ready to be reviewed

- [ ] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [ ] Test locally on your own infrastructure
- [ ] Run `go fmt` to format your code 🖊
- [ ] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation)) 
- [ ] Update or add tests 🧪
- [ ] Ensure the status checks below are successful ✅

PR #2107, opened 2022-09-28
chore(ci): Fix workflow that checks docs are up-to-date for v1

hermanschaaf:


erezrokah (approved):

hermanschaaf:
I've proven that it works now, since some plugins are failing the test :joy: I'll update their docs and check that they pass before merging this.

hermanschaaf:
k8s is failing because it still has an open PR to move it to v1: https://github.com/cloudquery/cloudquery/pull/2035 It should be fixed in that PR, and because we only run the check if files changed in that directory, I don't think it'll cause issues for most other PRs right now.

PR #2111, opened 2022-09-28
chore: Git ignore yaml files inside CLI dir

erezrokah:
#### Summary

Might be a bit too much so open for suggestions. I find myself trying to be super careful not to commit my configuration test files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,5 +25,6 @@ go.work
 go.work.sum
 cli/examples
 cli/cli
+cli/*.yml
 plugins/source/aws/aws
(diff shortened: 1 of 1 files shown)


yevgenypats (changes requested):
maybe just use `examples` ? we do have `yml` files in testdata directory ? 

Anyway we shouldn't have any creds there so shouldn't be a big deal even if commited.

erezrokah:
> maybe just use `examples` ? we do have `yml` files in testdata directory ?

Ah, I missed examples, that works 👍

PR #2074, opened 2022-09-26
feat: Move to standalone postgresql plugin

yevgenypats:
This PR got messed up in conflicts https://github.com/cloudquery/cloudquery/pull/2043

PR #2112, opened 2022-09-28
chore: Merge Build,Lint and Unit Tests workflows

erezrokah:
#### Summary

Per our internal discussion to reduce the number of status checks per PR


.github/workflows/test_unit.yml
```diff
@@ -19,13 +19,28 @@ jobs:
...
-  test_unit:
-    name: Unit Tests
+  test:
+    name: Test
     needs: resolve-modules
     strategy:
       matrix: ${{ fromJson(needs.resolve-modules.outputs.matrix) }}
       fail-fast: false
     runs-on: ubuntu-latest
+    services:
```
erezrokah, line 29: This was removed in https://github.com/cloudquery/cloudquery/pull/2074.

I suggest we keep it here at least until we ensure all CI passes without it

hermanschaaf (approved):

yevgenypats (changes requested):
Let's hold off with that a bit. I would like to try a backward solution that I'll share today/tomorrow where there is one check that is the required check and it waits for the checks to pass based on the same logic that we already have via changed files.

erezrokah:
@yevgenypats maybe merge this so at least we have a passing CI :) We can always do another PR with your change

erezrokah:
Closing for https://github.com/cloudquery/cloudquery/pull/2093

PR #2119, opened 2022-09-28
chore(gcp): Regenerate GCP with new timestamp resolver

candiduslynx:


yevgenypats (approved):
Looks ok. Does make sense to use the new caser functionality ?

PR #2122, opened 2022-09-28
chore(ci): Fix dest plugin CI

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/2112.
Our CI is currently failing since the dest plugin needs a Postgres DB.
Until we find a better way to include it only for specific plugins I suggest we at least keep our CI passing


yevgenypats (changes requested):
Let's discuss/consider this tomorrow - https://github.com/cloudquery/cloudquery/pull/2093

erezrokah:
Closing in favor of https://github.com/cloudquery/cloudquery/pull/2093

PR #2131, opened 2022-09-29
chore(aws): Contributing guide update

yevgenypats:
#### Summary

I just moved things around but we need to improve our contribution guide as number of people already asked for it recently. with the latest here - https://github.com/cloudquery/cloudquery/discussions/2092


erezrokah (approved):
Approved as it just moves things around, but we're still referencing `provider` in this doc

hermanschaaf (approved):
Moving things around, let's follow up with another for changes so it's easier to review

PR #2132, opened 2022-09-29
fix(gcp): Re-add missing field in crypto_keys (rotation_period)

### LINKED ISSUE #1995: Bug(v2): gcp_kms_crypto_keys.rotation_period column missing
Required in policies.

disq:
Closes #1995 

Had to use the old KMS api to make it work.


erezrokah (approved):

yevgenypats:
@disq can you please elaborate on what was the issue with the new API or even better add a comment so we wont try to change it to the new API in the future?

disq:
> @disq can you please elaborate on what was the issue with the new API or even better add a comment so we wont try to change it to the new API in the future?

@yevgenypats `rotation_period` field does not exist in the new api which is used by the policies. If there's a way to get that info some other way (or to re-iterate what the policy should check) then we can use the new api? The check is cis 1.2.0 - 1.10: "Ensure KMS encryption keys are rotated within a period of 90 days (Automated)"

yevgenypats (approved):

PR #2134, opened 2022-09-29
fix(gcp): Re-add gcp_iam_service_account_keys

### LINKED ISSUE #1990: Bug(v1): gcp_iam_service_account_keys is missing
Previously fetched with `Projects.ServiceAccounts.Keys.List`.

disq:
Closes #1990


plugins/source/gcp/codegen/recipes/iam.go
```diff
@@ -9,10 +9,8 @@ import (
 
 var iamResources = []*Resource{
 	{
-		SubService:   "roles",
-		Struct:       &iam.Role{},
-		NewFunction:  iam.NewProjectsRolesService,
-		ListFunction: (&iam.ProjectsRolesService{}).List,
```
yevgenypats, line 15: Is this not used ?

yevgenypats (approved):
Looks good. had one question.

plugins/source/gcp/codegen/recipes/iam.go
disq: > Is this not used ?

@yevgenypats because `SkipFetch` is set to true on all resources below.

PR #2045, opened 2022-09-25
chore: Fix website redirects

erezrokah:
#### Summary

Fix redirects to plugins, also uses `source` as to be consistent with other routes where the page title matches the route

---

Use the following steps to ensure your PR is ready to be reviewed

- [ ] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [ ] Test locally on your own infrastructure
- [ ] Run `go fmt` to format your code 🖊
- [ ] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation)) 
- [ ] Update or add tests 🧪
- [ ] Ensure the status checks below are successful ✅

PR #2148, opened 2022-09-29
chore(docs-v1): Point to README in source plugin table 

hermanschaaf:
...and remove duplicate Azure entry.

I think we should point to README for each plugin, otherwise the user is greeted with a long list of files and have to scroll down below the fold to get to the README portion.


hermanschaaf:
(note this is being merged into the [chore/website/docsv2](https://github.com/cloudquery/cloudquery/tree/chore/website/docsv2) branch)

erezrokah:
I remember doing the Azure fix in https://github.com/cloudquery/cloudquery/pull/2045/files and a bunch of other stuff too.

erezrokah (approved):

PR #2151, opened 2022-09-29
chore(docs-v1): AWS Contribution guide

hermanschaaf:
This replaces the old AWS contribution guide with a new one specifically written for V1.


hermanschaaf:
@yevgenypats I think this is ready for a review, I could probably write a lot more but this should be a good starting point.

yevgenypats (approved):

yevgenypats:
Looks great for starters. I'll also give another shot when I'll write my next resource on AWS both to test it and to add anything missing.

PR #2152, opened 2022-09-29
chore(ci-wait-for-all): Use correct sha when listing checks, exit early when done

erezrokah:
#### Summary

It seems the API call for `listForRef` always returns an empty array (even for non forked PRs, see https://github.com/cloudquery/cloudquery/actions/runs/3151802511/jobs/5126220370)

This PR fixes it for non forked PRs by using the correct `sha` and also fixes the workflow to exit early once we finished waiting

### DIFF
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -21,10 +21,9 @@ jobs:
       - uses: actions/github-script@v6
         env:
           DIRS: ${{ steps.changed-files.outputs.all_changed_files }}
-          REF: ${{ github.ref }}
         with:
           script: |
-            const { REF, DIRS } = process.env
+            const { DIRS: dirs } = process.env
             let now = new Date().getTime()
             const deadline = now + 60 * 1000 * 15
             let actions = ["cli",
@@ -43,7 +42,7 @@ jobs:
                            "plugins/destination/postgresql",
                            "plugins/destination/test",
                            ]
-            actions = actions.filter(item => DIRS.includes(item))
+            actions = actions.filter(item => dirs.includes(item))
             if (actions.length === 0) {
               console.log("No actions to wait for")
               return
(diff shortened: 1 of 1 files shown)


.github/workflows/wait_for_required_workflows.yml
```diff
@@ -65,12 +61,14 @@ jobs:
...
               })
+              if (actions.length === 0) {
+                console.log("No actions to wait for")
+                return
+              }
+
               await new Promise(r => setTimeout(r, 1000));
               now = new Date().getTime()
             }
-            if (actions.length === 0) {
```
erezrokah, line 71: I think the current code always wait for the deadline instead of existing early

cli/Makefile
```diff
@@ -6,4 +6,6 @@ lint:
 
 .PHONY: test-unit
 test-unit:
-	go test ./...
\ No newline at end of file
+	go test ./...
```
erezrokah, line 9: A change to trigger the CLI workflow

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -52,7 +52,7 @@ jobs:
               const checkRuns = await github.paginate(github.rest.checks.listForRef, {
                 owner: 'cloudquery',
                 repo: 'cloudquery',
-                ref: context.sha,
```
erezrokah, line 55: This is the wrong sha for pull requests events:
image

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -65,12 +65,14 @@ jobs:
                   actions = actions.filter(action => action !== item.name)
                 }
               })
-              await new Promise(r => setTimeout(r, 1000));
+              if (actions.length === 0) {
+                console.log("No actions to wait for")
+                return
+              }
+              
+              await new Promise(r => setTimeout(r, 5000));
```
erezrokah, line 73: I think 1 second polling is a bit too much, so changed it to 5 seconds to also avoid hitting rate limits

yevgenypats (approved):

PR #2154, opened 2022-09-29
chore(ci): Add go cache

yevgenypats:

### DIFF
--- a/.github/workflows/source_gcp.yml
+++ b/.github/workflows/source_gcp.yml
@@ -21,7 +21,9 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.1
+          go-version-file: plugins/source/gcp/go.mod
+          cache: true
+          cache-dependency-path: plugins/source/gcp/go.sum
       - uses: actions/checkout@v3
         with:
           fetch-depth: 2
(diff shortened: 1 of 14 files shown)


yevgenypats:
either doing something wrong or this is bug - https://github.com/actions/setup-go/issues/256 as can't find the path to `go.sum` and `go.mod`

.github/workflows/source_gcp.yml
```diff
@@ -21,7 +21,9 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.1
+          go-version-file: ./plugins/source/gcp/go.mod
+          cache: true
+          cache-dependency-path: ./plugins/source/gcp/go.sum
```
erezrokah, line 26:
```suggestion
          go-version-file: plugins/source/gcp/go.mod
          cache: true
          cache-dependency-path: plugins/source/gcp/go.sum
```

erezrokah:
>either doing something wrong or this is bug - https://github.com/actions/setup-go/issues/256 as can't find the path to go.sum and go.mod

You first need to checkout the code, then do the `setup-go`

yevgenypats:
> > either doing something wrong or this is bug - [actions/setup-go#256](https://github.com/actions/setup-go/issues/256) as can't find the path to go.sum and go.mod
> 
> You first need to checkout the code, then do the `setup-go`

lol. good catch. I knew I had to go to sleep :)

PR #2158, opened 2022-09-30
chore(docs): Refactor GCP codegen recipes and update contribution docs

hermanschaaf:
- Moves codegen-related files into `codegen` directory and adds a `recipes` subdirectory, similar to AWS source plugin 
- Updates contribution guidelines for GCP. It could probably use more details, but at least it's not completely out-of-date.


yevgenypats (approved):
Let's go!

PR #2161, opened 2022-09-30
chore(docs): Add releases and changelog links to plugins table

hermanschaaf:
Adding links to Releases and Changelogs to the plugins table in the docs. 

Ideally I think we should list the latest version there, but I couldn't think of a good way to automatically update that together with a new release. Thought release-please maybe, but didn't see an option for that. @yevgenypats / @erezrokah let me know if you have any ideas on this :)


hermanschaaf:
Ugh, I shouldn't have merged into this branch 😮‍💨  Let me undo that...

yevgenypats (approved):
Looks good but I think because of the file change `source->sources` we need to update all other links to that file.

erezrokah (approved):
Looks good, and I think we can follow up to autogenerate theses tables.

As for latest versions we could do something similar like the CLI:
https://github.com/cloudquery/cloudquery/blob/a279ecd9097792ea075e217e89cd2aac81bd745b/.github/workflows/release_cli.yml#L81

and add a find/replace in this file after releasing each plugin.
We can add it here:
https://github.com/cloudquery/cloudquery/blob/a279ecd9097792ea075e217e89cd2aac81bd745b/.github/workflows/release_cli.yml#L81

hermanschaaf:
@yevgenypats Thanks, updated the links - let me know if I missed one.

@erezrokah Great, yeah let's do that in a follow-up 👍

PR #2165, opened 2022-09-30
feat(cli): Use SourceClient and DestinationClient directly

yevgenypats:
PluginManager abstraction is not needed anymore. makes less prop drilling and less code. see https://github.com/cloudquery/plugin-sdk/pull/208


yevgenypats:
The versions should go prob also to the SDK but we can do that later if needed. The most important thing was to get rid off the heavy pluginmanager abstraction anyway.

cli/cmd/sync.go
```diff
@@ -58,49 +57,47 @@ func sync(cmd *cobra.Command, args []string) error {
...
 	for i := range destSubscriptions {
 		destSubscriptions[i] = make(chan []byte)
 	}
 	defer func() {
-		for _, destPlugin := range destPlugins {
-			if destPlugin != nil {
-				destPlugin.Close()
+		for _, destClient := range destClients {
+			if destClient != nil {
+				destClient.Close()
```
hermanschaaf, line 83: Not directly related to the work in this PR, but technically this `Close` returns an error that we're ignoring, we may want to at least log it

cli/cmd/sync.go
```diff
@@ -163,10 +160,6 @@ func syncConnection(ctx context.Context, pm *plugins.PluginManager, sourceSpec s
 	}
 	_ = bar.Finish()
 	fmt.Println("Fetch completed successfully.")
```
hermanschaaf, line 162:
```suggestion
	fmt.Println("Sync completed successfully.")
```
?

cli/cmd/sync.go
```diff
@@ -163,10 +160,6 @@ func syncConnection(ctx context.Context, pm *plugins.PluginManager, sourceSpec s
...
-	fmt.Printf("Summary: Resources: %d, Failed Writes: %d, Fetch Errors: %d, Fetch Warnings: %d\n",
-		totalResources, failedWrites, sourcePlugin.Errors(), sourcePlugin.Warnings())
-	if sourcePlugin.Errors() > 0 || sourcePlugin.Warnings() > 0 || failedWrites > 0 {
-		fmt.Println("Please check the logs for more details on errors/warnings.")
-	}
```
hermanschaaf, line 170: Why are these messages being removed, were they moved elsewhere?

cli/cmd/sync.go
```diff
@@ -58,49 +57,47 @@ func sync(cmd *cobra.Command, args []string) error {
...
-func syncConnection(ctx context.Context, pm *plugins.PluginManager, sourceSpec specs.Source, destinationsSpecs []specs.Destination) error {
-	sourcePlugin, err := pm.NewSourcePlugin(ctx, sourceSpec.Registry, sourceSpec.Path, sourceSpec.Version)
+func syncConnection(ctx context.Context, sourceSpec specs.Source, destinationsSpecs []specs.Destination) error {
+	sourceClient, err := clients.NewSourceClient(ctx, sourceSpec.Registry, sourceSpec.Path, sourceSpec.Version)
 	if err != nil {
 		return fmt.Errorf("failed to get source plugin client for %s: %w", sourceSpec.Name, err)
 	}
-	defer sourcePlugin.Close()
-	sourceClient := sourcePlugin.GetClient()
+	defer sourceClient.Close()
```
hermanschaaf, line 73: I think this also ignores any error we may get from `Close`

hermanschaaf (changes requested):

hermanschaaf:
Much simpler :+1: I left a few comments unrelated to this PR really, so feel free to ignore, but we should probably address the error handling there at some point. The only real question for me is around the print lines being removed

cli/cmd/sync.go
yevgenypats: Good catch. Print is good - log is already done on the sdk side (though I should add some more there)

cli/cmd/sync.go
yevgenypats: fixed

cli/cmd/sync.go
yevgenypats: I need to open a follow-up PR but I want us to stop parsing logs and the plugins return number of error and or warnings as they have this information anyway, this will make also consistent experience between debug and release (which I know was quite weird before)

hermanschaaf: Maybe keep these lines in there for now though, we can remove them in the follow-up?

hermanschaaf (approved):

Issue #2062, opened 2022-09-26
Add time interval to the destination


candiduslynx:
In accordance with https://github.com/cloudquery/plugin-sdk/issues/178

PR #2167, opened 2022-09-30
feat(postgresql): Add interval column type for `schema.TimeInterval`

candiduslynx:
Add interval column type for `schema.TimeInterval` (#2062)


yevgenypats (changes requested):
Looks good. Can you also add a default for text and maybe also log a warning that it doesn't support a specific type.

hermanschaaf:
@yevgenypats I think we need to carefully consider whether to default to text here--it can result in changed behaviour whenever the SDK adds a new type. For example, intervals were previously stored as `int`, so if we defaulted to `text` without explicitly handling it in this plugin, then those would have silently started to convert to `text`. I'd rather have it fail very loudly when we upgrade the SDK to a version that added a new type, so it forces you to add support or fall back to the right thing.

Maybe a solution is to add a test that iterates over all types in the SDK, checking they are all handled? Then you wouldn't be able to upgrade the SDK to a version with a new type without also updating the handled types.

plugins/destination/postgresql/client/types.go
```diff
@@ -68,6 +70,8 @@ func (*Client) SchemaTypeToCockroach(t schema.ValueType) string {
 		return "text[]"
 	case schema.TypeTimestamp:
 		return "timestamp without time zone"
+	case schema.TypeTimeInterval:
+		return "interval" // cockroach defaults to interval(6), so we leave it as is
```
hermanschaaf, line 74: why not be explicit and use `interval(6)`?

hermanschaaf (changes requested):
LGTM, just wondering why we chose to rely on defaults for the cockroachdb case

plugins/destination/postgresql/client/types.go
candiduslynx: Cocroachdb returns it back as `interval`, so our tests are failing here (pg works fine).

hermanschaaf: Ah okay interesting. Let's leave it as is then, seems unlikely to cause any real trouble.

hermanschaaf (approved):

yevgenypats (approved):

yevgenypats:
> @yevgenypats I think we need to carefully consider whether to default to text here--it can result in changed behaviour whenever the SDK adds a new type. For example, intervals were previously stored as `int`, so if we defaulted to `text` without explicitly handling it in this plugin, then those would have silently started to convert to `text`. I'd rather have it fail very loudly when we upgrade the SDK to a version that added a new type, so it forces you to add support or fall back to the right thing.
> 
> Maybe a solution is to add a test that iterates over all types in the SDK, checking they are all handled? Then you wouldn't be able to upgrade the SDK to a version with a new type without also updating the handled types.

Maybe we can have an option then. because let's say a community plugin is not up to date, it will be pretty hard requirement always staying up to date with all latest fields. maybe we can ignore this column as well? I think let's wait and see what issues this will cause.

PR #2249, opened 2022-10-03
fix: Drop directory from console log

erezrokah:
#### Summary

Fixes a wrong console message now that we support multiple directories. Also considered using https://github.com/gertd/go-pluralize but seems a bit much for this use case


cli/cmd/sync.go
```diff
@@ -33,16 +34,22 @@ func NewCmdSync() *cobra.Command {
 	return cmd
 }
 
+func getPluralDirectories(count int) string {
```
disq, line 37: Why not something like: `func pluralize(count int, singular, plural string) string`

cli/cmd/sync.go
```diff
@@ -33,16 +34,22 @@ func NewCmdSync() *cobra.Command {
...
 func sync(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
-	directory := "."
-	if len(args) > 0 {
-		directory = args[0]
-	}
-	fmt.Println("Loading specs from directory: ", directory)
+
+	directoryPlural := getPluralDirectories(len(args))
+	fmt.Printf("Loading specs from %s: %s\n", directoryPlural, strings.Join(args, ", "))
```
hermanschaaf, line 49: Btw, since we're here... there is another issue with this log line: if the config being read is a file, it still prints "directory". Sorry to pile onto this PR, but would you mind fixing that too? 😄

erezrokah:
Dropped "directory" entirely to keep things simple

cli/cmd/sync.go
```diff
@@ -35,14 +36,11 @@ func NewCmdSync() *cobra.Command {
...
-	if len(args) > 0 {
-		directory = args[0]
-	}
-	fmt.Println("Loading specs from directory: ", directory)
+
+	fmt.Printf("Loading spec(s) from: %s\n", strings.Join(args, ", "))
 	specReader, err := specs.NewSpecReader(args)
 	if err != nil {
-		return fmt.Errorf("failed to load specs from directory %s: %w", directory, err)
+		return fmt.Errorf("failed to load spec(s) from: %s. Error: %w", strings.Join(args, ", "), err)
```
hermanschaaf, line 43:
```suggestion
		return fmt.Errorf("failed to load spec(s) from %s. Error: %w", strings.Join(args, ", "), err)
```

PR #2250, opened 2022-10-03
fix: Set `.cq` directory when downloading plugins

erezrokah:
#### Summary

Not sure if this should be here or in the SDK


erezrokah:
Will do this in the SDK

PR #2295, opened 2022-10-03
fix(cli): Fix tests and remove push triggers

yevgenypats:


erezrokah:
What's the reason we need to run tests on `push`? If we keep branches up to date before merging running everything on push shouldn't be required

yevgenypats:
> What's the reason we need to run tests on `push`? If we keep branches up to date before merging running everything on push shouldn't be required

They were already running on push but I've added filters to make them not run on every push. so either remove them completely or use filtering. Im fine with anything.

erezrokah:
> They were already running on push but I've added filters to make them not run on every push. so either remove them completely or use filtering. Im fine with anything.

Yeah lets remove them completely (we didn't have them in the single workflow for all plugins flow)

PR #2310, opened 2022-10-04
chore: Update the latest link

erezrokah:
#### Summary

Only official releases get updated automatically, so we have to do this manually for now


yevgenypats:
let's make it official so we dont need to do anymore.

Issue #2308, opened 2022-10-04
chore(ci): Check for spelling errors in MDX files


erezrokah:
Currently we only do it for MD files, but with the move to Nextra many of our docs are MDX files.

Looks like https://vale.sh/docs/topics/config/#format-associations should work:

```
[formats]
mdx = md
```

PR #2320, opened 2022-10-04
chore(docs): Fix typos in blogs, docs, website

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/2308


plugins/source/gcp/CONTRIBUTING.md
```diff
@@ -2,15 +2,16 @@
 
 Thanks for contributing to CloudQuery! You are awesome. This document serves as a guide for adding new services and resources to the GCP source plugin.
 
-There are three main steps to adding a new GCP resource:
-1. [Add a code generation recipe](#1-add-a-code-generation-recipe)
-2. [Writing the resolver function to fetch the resource using the AWS SDK](#2-setting-up-the-resource)
+There are two main steps to adding a new GCP resource:
 
-## 1. Add a Code Generation Recipe
```
hermanschaaf, line 9: Why remove the step numbers from the headings here? I think that helps to add context and ensures readers don't miss a step by skipping around

plugins/source/k8s/dashboards/README.md
```diff
@@ -3,18 +3,19 @@
 This directory contains pre-built dashboards. Currently those are available only for Grafana, but you can create them in any other BI platform:
 
 Checkout those tutorials:
+
 * [Building Open Source Cloud Asset Inventory with CloudQuery and Grafana](https://www.cloudquery.io/blog/open-source-cloud-asset-inventory-with-cloudquery-and-grafana)
 * [Building Open Source Cloud Asset Inventory with CloudQuery and Apache Superset](https://www.cloudquery.io/blog/cloud-asset-inventory-cloudquery-apache-superset)
 * [Building Open Source Cloud Asset Inventory with CloudQuery and AWS QuickSight](https://www.cloudquery.io/blog/cloud-asset-inventory-cloudquery-aws-quicksight)
-* [Building Open Source Cloud Asset Inventory with MetaBase](https://www.cloudquery.io/blog/cloud-asset-inventory-cloudquery-metabase)
+* [Building Open Source Cloud Asset Inventory with Metabase](https://www.cloudquery.io/blog/cloud-asset-inventory-cloudquery-metabase)
```
erezrokah, line 10: This is the correct casing for `Metabase`

plugins/source/k8s/dashboards/README.md
```diff
@@ -3,18 +3,19 @@
...
 
 ## What's inside?
 
 ### Kubernetes Asset Inventory
 
 <img alt="Kubernetes Asset Inventory" src="../dashboards/grafana/asset_inventory.png" width=50% height=50%>
 
-#### Installation
```
erezrokah, line 17: We should not use the same name for headers as it prevents linking to them

plugins/source/test/docs/README.md
```diff
@@ -1,5 +0,0 @@
-# Source Plugin: test
```
erezrokah, line 1: Already under `docs/tables`

plugins/source/test/docs/test_some_table.md
```diff
@@ -1,16 +0,0 @@
-# Table: test_some_table
```
erezrokah, line 1: Already under `docs/tables`

website/pages/blog/aws-encryption-and-multi-account-access.md
```diff
@@ -265,6 +265,6 @@ If you have comments, feedback on this post, follow-up topics you’d like to se
 
 [AWS Key Management Service: How AWS Services use AWS KMS](https://docs.aws.amazon.com/kms/latest/developerguide/service-integration.html)
 
-[AWS Whitepaper: Organizing your AWS Environment Using Multiple Accounts](https://docs.aws.amazon.com/whitepapers/latest/organizing-your-aws-environment/organizing-your-aws-environment.html)
+[AWS White paper: Organizing your AWS Environment Using Multiple Accounts](https://docs.aws.amazon.com/whitepapers/latest/organizing-your-aws-environment/organizing-your-aws-environment.html)
```
erezrokah, line 268: Google says so 
image

website/pages/docs/reference/destination-spec.md
```diff
@@ -4,32 +4,37 @@ This goes through all the available options for the destination plugin `spec` ob
 
 ## Spec
 
-### name
+### `name`
```
hermanschaaf, line 7: Using quotes here doesn't render well on the website (I tried it) - because they're rendered as `code`, the font is small and not the size that a header should be. I don't think it adds much to render it as code in any case, could we keep it as it was maybe?

website/pages/docs/reference/source-spec.md
```diff
@@ -4,42 +4,49 @@ This goes through all the available options for the source plugin `spec` object:
 
 ## Spec
 
-### name
+### `name`
```
hermanschaaf, line 7: Same here

plugins/source/gcp/CONTRIBUTING.md
erezrokah: Ah it feels odd to have numbering for a header. I can keep those I don't mind

website/pages/docs/reference/destination-spec.md
erezrokah: Good point I'l revert that change

hermanschaaf (approved):

plugins/source/gcp/CONTRIBUTING.md
hermanschaaf: Probably not worth spending more time on, but I would also be okay with `Step 1: ...`, `Step 2: ...`, etc if that's better for you :)

erezrokah: I don't have a strong opinion here so we can leave it at that :)

PR #2321, opened 2022-10-04
chore!: Official v1 release

erezrokah:
#### Summary

This should set the next version of the CLI to `v1.0.0`


.release-please-manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "cli": "1.3.0-pre.1",
+  "cli": "0.9.9",
```
erezrokah, line 2: This tells release please the current version, so with a breaking change the next one should be `v1.0.0`

cli/Makefile
```diff
@@ -4,4 +4,4 @@ lint:
 
 .PHONY: test
 test:
-	go test ./...
\ No newline at end of file
+	go test ./...
```
erezrokah, line 7: This is a stub change to trigger a release in the monorepo (as we detect changes based on paths in the repo)

.release-please-manifest.json
hermanschaaf: Don't our breaking changes usually get their minor version bumped? Or how does it work now?

yevgenypats (approved):

.release-please-manifest.json
erezrokah: Yes we did than until https://github.com/cloudquery/cloudquery/pull/2050/files

hermanschaaf: Ah cool, thanks 👍

PR #2335, opened 2022-10-04
chore!: Official v1 release

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/2321 for plugins


erezrokah:
Linting targets don't work anyways as they need to point to the correct configuration file two levels up. We can add them later

hermanschaaf (approved):

PR #2364, opened 2022-10-04
fix(aws): Fix throttle errors

yevgenypats:
#### Summary

Bringing back this PR as number of users reported the same behaviour Im experiencing as well the retry is stuck in infinite amount of retries because he had some kind of our own implementation and users are getting hit by client side throttling.

Turning client side throttling off for now and we can expose variables to control it if needed in the future.


bbernays:
Two things:
1. Before we merge this can we get confirmation about the errors that the impacted users are actually this and not some other error?
2. I am still unsure why this is the best solution rather than [this pr](https://github.com/cloudquery/cloudquery/pull/1878)

yevgenypats:
Im able to reproduce it locally. I dont think we should be afraid to merge things as this was hiting us for over a year now and I was able to find a solution that I believe that work. it will be very easy for them to test it. If it doesn't work then we will know there is something else.

hermanschaaf (approved):

PR #2384, opened 2022-10-04
fix(cli): Dont truncate log (append only)

yevgenypats:


hermanschaaf (approved):

hermanschaaf:
Necessary, but not sufficient 😃 It clearly should be append, but there seems to be another issue here with logs not getting streamed from the plugins

PR #2389, opened 2022-10-04
fix: Add back missing plugin logs

hermanschaaf:
This fixes an issue where plugin logs were not being streamed to `cloudquery.log`. It also adds a test for this to catch future regressions.

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -37,6 +37,8 @@ func NewCmdSync() *cobra.Command {
 
 func sync(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
+	log.Info().Msg("Start sync")
+	defer log.Info().Msg("End sync")
 
 	fmt.Printf("Loading spec(s) from %s\n", strings.Join(args, ", "))
 	specReader, err := specs.NewSpecReader(args)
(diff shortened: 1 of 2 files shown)


cli/cmd/sync.go
```diff
@@ -37,6 +37,8 @@ func NewCmdSync() *cobra.Command {
 
 func sync(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
+	log.Info().Msg("Start sync")
```
yevgenypats, line 40: I think we should move this to `syncConnection` and add `source` and `destinations=[]` to that as otherwise it's missing information and `sync_tmie`

disq (approved):

yevgenypats (approved):
looks good, one nit around logging.

cli/cmd/sync.go
```diff
@@ -65,8 +66,17 @@ func sync(cmd *cobra.Command, args []string) error {
 }
 
 func syncConnection(ctx context.Context, sourceSpec specs.Source, destinationsSpecs []specs.Destination) error {
+	destinationNames := make([]string, len(destinationsSpecs))
+	for i := range destinationsSpecs {
+		destinationNames[i] = destinationsSpecs[i].Name
+	}
+	log.Info().Str("source", sourceSpec.Name).Strs("destinations", destinationNames).Msg("Start sync")
```
disq, line 73: Do you think it would make sense to set and log `syncTime` here as well? or overwrite the log line timestamp somehow?

hermanschaaf: Oh yeah totally, let's add `syncTime` as that is pretty important.

hermanschaaf: Added

cli/cmd/sync.go
```diff
@@ -65,8 +66,18 @@ func sync(cmd *cobra.Command, args []string) error {
...
 
 func syncConnection(ctx context.Context, sourceSpec specs.Source, destinationsSpecs []specs.Destination) error {
+	destinationNames := make([]string, len(destinationsSpecs))
+	for i := range destinationsSpecs {
+		destinationNames[i] = destinationsSpecs[i].Name
+	}
 	syncTime := time.Now().UTC()
-	sourceClient, err := clients.NewSourceClient(ctx, sourceSpec.Registry, sourceSpec.Path, sourceSpec.Version)
+
+	log.Info().Str("source", sourceSpec.Name).Strs("destinations", destinationNames).Time("syncTime", syncTime).Msg("Start sync")
```
disq, line 75: nit: I think all our log args are snake_case, so it should apply to `sync_time` as well

hermanschaaf: Updated

Issue #2382, opened 2022-10-04
CLI: Improve formatting of error output


hermanschaaf:
Currently, if the CLI encounters an error in the configuration (or something else) it can output some messages that are not very human-friendly. For example:

```
cloudquery sync cq
Loading spec(s) from cq
Starting sync for:  heroku -> [postgresql]
| Syncing resources... (0/-, 0 resources/hr) [0s] Error: failed to sync source heroku: failed to fetch resources: failed to sync source heroku: failed to fetch resources from stream: rpc error: code = Unknown desc = failed to sync resources: table heroku_add_on_webhook_events cannot be both included and skipped
```

The relevant part for the user in this case is the very last bit: `heroku_add_on_webhook_events cannot be both included and skipped`.

I'm not sure what a reasonable fix will look like here, but maybe a start would be output only the final part of the error to the terminal, and write the rest to the logs:

```
cloudquery sync cq
Loading spec(s) from cq
Starting sync for:  heroku -> [postgresql]
| Syncing resources... (0/-, 0 resources/hr) [0s] 
Error: table heroku_add_on_webhook_events cannot be both included and skipped
```

yevgenypats:
Closing as I think we shouldn't format go errors but a better approach will be having a future validation layer where we can print warnings/errors on each line instead of wrapping/unwrapping/splitting errors - https://github.com/cloudquery/plugin-sdk/pull/272

PR #2408, opened 2022-10-04
feat: Add newlines to errors printed in CLI

hermanschaaf:
Rather than printing a long string of errors on a single line, this splits errors onto multiple lines (separated by `: `).

Before:

```
../cloudquery/cli/cli sync --log-level debug v2
Loading spec(s) from v2
Starting sync for:  heroku -> [postgresql]
| Syncing resources... (0/-, 0 resources/hr) [0s] Error: failed to sync source heroku: failed to fetch resources: failed to sync source heroku: failed to fetch resources from stream: rpc error: code = Unknown desc = failed to sync resources: invalid "*" resource, with explicit resources
```

After:
```
../cloudquery/cli/cli sync --log-level debug v2
Loading spec(s) from v2
Starting sync for:  heroku -> [postgresql]
| Syncing resources... (0/-, 0 resources/hr) [0s] Error: failed to sync source heroku:
  failed to fetch resources:
  failed to sync source heroku:
  failed to fetch resources from stream:
  rpc error:
  code = Unknown desc = failed to sync resources:
  invalid "*" resource, with explicit resources
```

Hopefully small quality-of-life improvement for CloudQuery users, with little cost for us to maintain.

Closes https://github.com/cloudquery/cloudquery/issues/2382

### DIFF
--- a/cli/cmd/doc.go
+++ b/cli/cmd/doc.go
@@ -20,9 +20,9 @@ func newCmdDoc() *cobra.Command {
 		Short:  docShort,
 		Args:   cobra.ExactValidArgs(1),
 		Hidden: true,
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: commandWithPrettyErrors(func(cmd *cobra.Command, args []string) error {
 			return doc.GenMarkdownTreeCustom(cmd.Parent(), args[0], filePrepender, linkHandler)
-		},
+		}),
 	}
 	return cmd
 }
--- a/cli/cmd/errors.go
+++ b/cli/cmd/errors.go
@@ -1,0 +1,31 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"strings"
+)
+
+type prettyError struct {
+	err error
+}
+
+type command func(*cobra.Command, []string) error
+
+func commandWithPrettyErrors(original command) command {
+	return func(cmd *cobra.Command, args []string) error {
+		err := original(cmd, args)
+		return prettifyError(err)
+	}
+}
+
+func (p prettyError) Error() string {
+	msg := p.err.Error()
+	return strings.Join(strings.Split(msg, ": "), ":\n  ")
+}
+
+func prettifyError(err error) error {
+	if err == nil {
+		return err
+	}
+	return prettyError{err: err}
+}
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -30,7 +30,7 @@ func NewCmdSync() *cobra.Command {
 		Long:    fetchShort,
 		Example: fetchExample,
 		Args:    cobra.MinimumNArgs(1),
-		RunE:    sync,
+		RunE:    commandWithPrettyErrors(sync),
 	}
 	return cmd
 }


yevgenypats:
Ah I just meant a new line after the progressbar. I don't think this will play well with CI systems, logs and so on, it will break a lot of standard tools that work with CLIs. I would follow k8s error handling here as it's the biggest go code base Im aware of  and didn't see custom go error handling in other places. I think we might want to do a bit better error wrapping (if we have to much of wraping but really we should just print the errors as is).

I think we should also rememeber CQ CLI is a just a CLI to a backend (and it is mostly run in CI) so we need to put quite little effort into CLI capabilities and mostly put effort in SDK, data, plugins, concurrency, deployments and so on.

hermanschaaf:
> I don't think this will play well with CI systems, logs and so on, it will break a lot of standard tools that work with CLIs

How would adding newlines break standard tools that work with CLIs? We also maintain a separate log file for logs. If users want to dump the output of the CLI to a log they can, but they shouldn't be parsing that output without expecting it to change from time to time.

We are also including things like progress bars, which you would not expect to be useful (or even good) in a CI environment. 

I think we have to face the fact that at least initially, the CLI is how humans interact with the software. Maybe later it moves to CI, but for the first couple of iterations, users will only know about configuration mistakes through the errors we report. These messages (especially config validation) should be as user-friendly as possible, regardless of how Go error handling works.

yevgenypats:
> > I don't think this will play well with CI systems, logs and so on, it will break a lot of standard tools that work with CLIs
> 
> How would adding newlines break standard tools that work with CLIs? We also maintain a separate log file for logs. If users want to dump the output of the CLI to a log they can, but they shouldn't be parsing that output without expecting it to change from time to time.
> 
> We are also including things like progress bars, which you would not expect to be useful (or even good) in a CI environment.
> 
> I think we have to face the fact that at least initially, the CLI is how humans interact with the software. Maybe later it moves to CI, but for the first couple of iterations, users will only know about configuration mistakes through the errors we report. These messages (especially config validation) should be as user-friendly as possible, regardless of how Go error handling works.

We can discuss this live but we are mixing two differents concepts here: go errors and cloudquery config validation.

go errors should always be printed as one line because logic of go errors wraping and unwrapping can get really hairy and it will grow all the time because we will have to address different use-cases.

Config validation output - I hear you and the right solution here imo is to have SDK logic for that if we really want that but that's another feature. i.e ValidateConfig and it can return number of strings. so in case the validation failed it will print to the screen and just return one error but it wont be this `RPC weird thin` which happens because the rpc should in general never return an error and this is where the confusion comes from so in normal mode of operation where there is a validation error it should print something nicely and just return -1. but I would push this feature down the line for now.

For example in this validation API - https://github.com/xeipuuv/gojsonschema you can see there is a distinguish between errors and valid/non-valid.

yevgenypats:
I agree the progress bar is weird it we should look into what is the right solution there. also right now we sync sources in sequence but let's say users will ask us to support parallel source sync then the CLI actually turns into a a schedule rather then a CLI and it shouldn't exit for example on errors but rather just log and continue and then the question is it a CLI at all. worth discussing / brainstorming  on the responsibilities of the CLI and so on .

yevgenypats:
basically we are mixing between exceptions and errors like in python but because go doesn't really have exception it makes it quite difficult to have a clear approach. I think the validation thing should solve it but it can also become a thing so this is why I worry about that.

PR #2410, opened 2022-10-04
fix!: Update all AWS Client Versions

bbernays:
#### Summary

- Upgrade from `aws-sdk-go` to `aws-sdk-go-v2`
- Update all AWS SDK versions to latest version and regenerate resolvers and docs


plugins/source/aws/resources/services/route53/domains.go
```diff
@@ -34,6 +34,26 @@ func Domains() *schema.Table {
 				Resolver:    resolveRoute53DomainTags,
 				Description: `A list of tags`,
 			},
+			{
+				Name:     "admin_contact",
+				Type:     schema.TypeJSON,
+				Resolver: schema.PathResolver("AdminContact"),
+			},
```
bbernays, line 41: More re ordering of fields/columns

plugins/source/aws/docs/tables/aws_route53_domains.md
```diff
@@ -15,23 +15,24 @@ The composite primary key for this table is (**account_id**, **domain_name**).
...
-|nameservers|JSON|
-|registrant_contact|JSON|
 |registrant_privacy|Bool|
 |registrar_name|String|
 |registrar_url|String|
 |registry_domain_id|String|
 |reseller|String|
-|status_list|JSON|
-|tech_contact|JSON|
+|status_list|StringArray|
```
hermanschaaf, line 34: This looks like a breaking change (changing type JSON to StringArray). Not much we can do about it... but it should technically require a major version bump.

plugins/source/aws/resources/services/route53/domains.go
```diff
@@ -111,14 +116,9 @@ func Domains() *schema.Table {
 			},
 			{
 				Name:     "status_list",
-				Type:     schema.TypeJSON,
+				Type:     schema.TypeStringArray,
```
bbernays, line 119: @hermanschaaf - A StringArray is a subset of JSON, maybe we can override the codegen to keep it as a JSON?

hermanschaaf: Indeed we can... although it is a different question from whether we should 🙂 Future cases will also inevitably arise where we don't have a backwards-compatible option. Plus, just because we keep the column JSON, doesn't mean users don't need to update their queries (since the contents will change). I'd rather indicate that with a major version bump. 

Maybe we should do updates like these on a semi-regular schedule, but not so regularly so as to cause constant toil for users. Will need to think about this some more...

plugins/source/aws/resources/services/fsx/data_repository_tasks.go
```diff
@@ -71,6 +71,16 @@ func DataRepositoryTasks() *schema.Table {
...
+			{
+				Name:     "file_cache_id",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("FileCacheId"),
+			},
+			{
+				Name:     "file_system_id",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("FileSystemId"),
+			},
```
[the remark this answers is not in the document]
hermanschaaf: Right now, with PostgreSQL being the only destination, I don't think it breaks backwards compatibility... However as we add more destinations this may become an issue (e.g. for a CSV destination it absolutely will break, and possibly for some other data lakes too)

plugins/source/aws/resources/services/route53/domains.go
bbernays: This change only occurred because the codegen was using [v1](https://github.com/aws/aws-sdk-go-v2/blob/9c92421d2b2253960dd282b5966f7712b62469dd/service/route53domains/api_op_GetDomainDetail.go#L135) of the AWS Go SDK  which relied heavily on pointers rather than native go types like the [AWS SDK Go V2](https://github.com/aws/aws-sdk-go-v2/blob/9c92421d2b2253960dd282b5966f7712b62469dd/service/route53domains/api_op_GetDomainDetail.go#L135) does

hermanschaaf: Ah, I see! So it shouldn't happen too often then. That's good to know

plugins/source/aws/resources/services/glue/jobs_test.go
```diff
@@ -17,11 +17,12 @@ func buildJobsMock(t *testing.T, ctrl *gomock.Controller) client.Services {
 
 	node := types.CodeGenConfigurationNode{}
 
-	require.NoError(t, faker.FakeDataSkipFields(&node, []string{"ApplyMapping", "JDBCConnectorSource"}))
+	require.NoError(t, faker.FakeDataSkipFields(&node, []string{"ApplyMapping", "JDBCConnectorSource", ""}))
```
hermanschaaf, line 20: Why is there an empty string being skipped here?

plugins/source/aws/resources/services/route53/domains.go
bbernays: AWS is really good about not releasing breaking changes, so unless something in our codegen/sdk changes as then we should always be good regenerating the clients

plugins/source/aws/resources/services/glue/jobs_test.go
bbernays, line 20:
```suggestion
	require.NoError(t, faker.FakeDataSkipFields(&node, []string{"ApplyMapping", "JDBCConnectorSource"}))
```

hermanschaaf:
@bbernays LGTM, there are just some conflicts to resolve now

PR #2431, opened 2022-10-05
chore(cli): Make sentry only send errors on panic

yevgenypats:
Go doesn't really have a great way of distinguishing exceptions and errors. Also we don't want to go into the rabbit hole of classifying every error as user error or internal error as sometimes it's not really possible: Let's say we had an error in our marshaling logic where usually this error means a user error in some cases it can be also an internal one.

We will have to rely on user reports for the most, apart from panics which are clearly internal errors and we would like to send them to sentry.


cli/main.go
```diff
@@ -11,9 +13,18 @@ import (
...
+			sentry.Flush(sentryFlushTimeout)
+			fmt.Println(originalMessage)
+			os.Exit(1)
+		}
+	}()
+
 	if err := cmd.NewCmdRoot().Execute(); err != nil {
-		sentry.CaptureMessage(err.Error())
-		sentry.Flush(sentryFlushTimeout)
 		os.Exit(1)
```
disq, line 28: This is fine, would just add a // lint: nolint here

cli/main.go
```diff
@@ -10,10 +12,19 @@ import (
...
+			panic(err)
+		}
+	}()
+
 	if err := cmd.NewCmdRoot().Execute(); err != nil {
-		sentry.CaptureMessage(err.Error())
-		sentry.Flush(sentryFlushTimeout)
 		os.Exit(1)
 	}
 	sentry.Flush(sentryFlushTimeout)
```
disq, line 30: No need for this line any more, and the const can also be removed.

cli/main.go
```diff
@@ -10,10 +12,19 @@ import (
 
 const sentryFlushTimeout = 5 * time.Second
 
+//nolint:gocritic
```
disq, line 15: Wouldn't add nolint to the whole fn but to the os.Exit line?

cli/main.go
yevgenypats: It is actually needed becayse call to `CaptureMessage` are not sent synchronously by default , but I changed the transport to sync now so we can avoid it as we call sentry only in one place

cli/main.go
yevgenypats: done

cli/main.go
disq: We still don't this line because if it panics and we're sending a sentry report in the defer, this line won't run anyway.

yevgenypats: Oh I see. you commented on another line. sorry.

disq (approved):

PR #2432, opened 2022-10-05
fix(aws): Codegen name transformer

disq:
This will resolve the `e_c2_security_groups` problem and potentially others. Avoided regex, because to do it with regex required a complex solution.

Could also just keep the HasPrefix check and remove the others, but feels like we'll run into this issue with others soon (as we add new resources)


hermanschaaf:
@disq Looks good, but I think this is a slightly simpler solution that uses the functions available in the SDK: https://github.com/cloudquery/cloudquery/pull/2473

PR #2459, opened 2022-10-05
fix(cli): Close stdout when using --log-console

yevgenypats:
when we use --log-console we dont want standard output. logs are by default written to stderr so we just close stdout which make any fmt.Print* silent. This makes it easy to use fmt.Print* in the CLI and not use any new abstractions and prop drilling.

This also adds handling of ctrl+c, that should go together with this as well https://github.com/cloudquery/plugin-sdk/pull/252


cli/cmd/root.go
```diff
@@ -63,6 +64,9 @@ func NewCmdRoot() *cobra.Command {
 				}
 			}
 			if logConsole {
+				if err := os.Stdout.Close(); err != nil {
+					return fmt.Errorf("failed to close stderr: %w", err)
```
hermanschaaf, line 68:
```suggestion
					return fmt.Errorf("failed to close stdout: %w", err)
```

hermanschaaf (approved):

PR #2461, opened 2022-10-05
feat(aws): Use PreResourceResolver to resolve list/describe resources

disq:
Shouldn't change performance as we don't have the proposed "PreResourceResolver parallelism" in the plugin-sdk yet.


hermanschaaf:
@disq Looks like docs need to be regenerated here (the AWS workflow failed)

disq:
> @disq Looks like docs need to be regenerated here (the AWS workflow failed)

nah it was because of the codegen drift in main, before merging #2476. fixed.

plugins/source/aws/codegen/recipes/backup.go
```diff
@@ -23,9 +23,10 @@ func BackupResources() []*Resource {
 			},
 		},
 		{
-			SubService: "plans",
-			Struct:     &backup.GetBackupPlanOutput{},
-			SkipFields: []string{"BackupPlanArn"},
+			SubService:          "plans",
+			Struct:              &backup.GetBackupPlanOutput{},
+			SkipFields:          []string{"BackupPlanArn"},
+			PreResourceResolver: "getPlan",
```
bbernays, line 29: Shouldn't the `PreResourceResolver` be a plural, because it is getting a list of items?

disq: `Pre` is actually not a good name, because it's the last call before we "realize" a resource with its fields, **after** the actual resolver is called (to potentially get the list of items). The "singularization" of the list of items is done by the SDK, and we get a single handle from the resolver, look it up in the `PreResourceResolver` handler and do a `resource.Item` assignment (or `resource.SetItem()`, which is never used for some reason)

PR #2472, opened 2022-10-06
chore(website): Minor nits to migration blog

yevgenypats:

### DIFF
--- a/website/pages/blog/migrating-from-v0-to-v1.md
+++ b/website/pages/blog/migrating-from-v0-to-v1.md
@@ -81,7 +81,7 @@ spec:
     connection_string: "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"```
 ```
 
-Check the documentation [destination spec](/docs/reference/destination-spec), and for each [plugin](/docs/plugins/destinations) for details on how to configure the plugin-specific spec part. However, generally these will be exactly the same as in v0, and all the same authentication functionality is still supported.
+Check the [destination spec documentation](/docs/reference/destination-spec) for general layout, and individual [destination plugin documentation](/docs/plugins/destinations) for details on how to configure the plugin-specific spec part. Generally these will be the same as in v0, and all the same authentication functionality is still supported.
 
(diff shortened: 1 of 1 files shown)


website/pages/blog/migrating-from-v0-to-v1.md
```diff
@@ -23,7 +23,7 @@ The [announcement blog post](/blog/cloudquery-v1-release) lists many of the impo
 
 ## Changes to the Configuration Format
 
-V1 introduces a new config format that is closely related to the old one, but an old config will need some massaging to work with the CloudQuery v1 CLI.
+V1 introduces a new config format that is closely related to the old one, but an old config will need some massaging to work with the CloudQuery v1 CLI. Mostly because we support now multiple destinations, there are seperate configs for source and destination plugins.
```
hermanschaaf, line 26:
```suggestion
V1 introduces a new config format that is closely related to the old one, but an old config will need some massaging to work with the CloudQuery v1 CLI. Mostly because we now support multiple destinations, there are separate configs for source and destination plugins.
```

website/pages/blog/migrating-from-v0-to-v1.md
```diff
@@ -55,7 +55,7 @@ spec:
     # plugin specific configuration.
 ```
 
-Check the documentation for each plugin for details on how to configure the plugin-specific spec part. However, generally these will be exactly the same as in v0, and all the same authentication functionality is still supported.
+Check the documentation [source spec](/docs/reference/source-spec), and for each [plugin](/docs/plugins/sources) for details on how to configure the plugin-specific spec part. However, generally these will be exactly the same as in v0, and all the same authentication functionality is still supported.
```
hermanschaaf, line 58:
```suggestion
Check the [source spec documentation](/docs/reference/source-spec) for general layout, and individual [plugin documentation](/docs/plugins/sources) for details on how to configure the plugin-specific spec. Generally these will be the same as in v0, and all the same authentication functionality is still supported.
```

PR #2473, opened 2022-10-06
fix: AWS codegen: fix ec2 naming

hermanschaaf:
Add a special case for EC2 so that it doesn't get converted to `e_c2` by the caser. I re-ran codegen using this and it resulted in no changes (as expected).


amanenk (approved):
LGTM

PR #2477, opened 2022-10-06
chore: Add step to update versions in docs

hermanschaaf:
This adds steps to the plugin release workflow to update the versions in the docs to the latest. It currently handles two cases:
 - a string like `version: "v2.0.0" # latest version of aws plugin` will get matched and the version replaced with the latest version.
 - a string in the sources/destinations table will get matched and replaced with the latest version for easy reference.

I tried some other, less janky, approaches but couldn't get it working with nextra within a couple of hours and decided to revert to a simpler solution.


yevgenypats (approved):
Looks great. I also wasted a few hours on making it work via nextra but got to the same issues you hit. I think `shiki` might be broken and/or not a good fit for next/mdx. 

Anyway, opened this issue - https://github.com/shuding/nextra/issues/876 

Maybe there will be an update there. The funny thing is that with react-syntax-highlighter it works but they migrated away so it will involve maintaining   two components and two css which will be also quite hard.

yevgenypats:
Maybe, small nit: the magic keyword should be `plugins/source/name` and or something to note that it is a `destination` or `source` plugin

erezrokah (approved):
Looks good and follows the same convention we have for the CLI:
https://github.com/cloudquery/cloudquery/blob/16ed24b0310e8f0a09c412e7bce7377e765cb46e/.github/workflows/release_cli.yml#L76

Not sure how to do it in Nextra as it generates the content during build time (not when users access the Website, unless we use `getStaticProps` with `revalidate` which seems a bit too much for this use case). So we'll still need to trigger a build after a plugin release if we use the Nextra route.

yevgenypats:
> Looks good and follows the same convention we have for the CLI:
> 
> https://github.com/cloudquery/cloudquery/blob/16ed24b0310e8f0a09c412e7bce7377e765cb46e/.github/workflows/release_cli.yml#L76
> 
> Not sure how to do it in Nextra as it generates the content during build time (not when users access the Website, unless we use `getStaticProps` with `revalidate` which seems a bit too much for this use case). So we'll still need to trigger a build after a plugin release if we use the Nextra route.

In nextra I was able to do that. mdx is already generated at compile time. the problem is this `shiki` library which doesn't work. unless this was in codeblock we can already use something like:

```

export {  SOURCE_PLUGIN_AWS_VERSION } from ../versions.tsx;

# README
Please see ${SOURCE_PLUGIN_AWS_VERSION}
```

erezrokah:
> In nextra I was able to do that. mdx is already generated at compile time

Yes that's exactly the issue. If MDX is generated at compile time, you'll need to trigger a build after each plugin release anyways to get the site updated, so easier to do it via a PR that changes the static content of the docs (instead of doing it dynamically in the MDX)

erezrokah:
OK so this has a side effect of creating conflicts in the `sources.md` file, see 
image

https://github.com/cloudquery/cloudquery/pull/2539

I think we can use the same approach as https://github.com/cloudquery/cloudquery/pull/1314 to handle it

PR #2486, opened 2022-10-06
fix: Fixed crashes when json field is nil

amanenk:
#### Summary

 MarshaledJsonResolver  was not able to handle nil fields

### DIFF
--- a/plugins/source/aws/client/resolvers.go
+++ b/plugins/source/aws/client/resolvers.go
@@ -139,15 +139,14 @@ func MarshaledJsonResolver(path string) schema.ColumnResolver {
 		var j map[string]interface{}
 		field := funk.Get(r.Item, path, funk.WithAllowZero())
 
+		if field == nil {
+			return nil
+		}
+
 		var val reflect.Value
+		val = reflect.ValueOf(field)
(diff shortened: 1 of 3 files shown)


plugins/source/aws/client/resolvers.go
```diff
@@ -139,15 +139,14 @@ func MarshaledJsonResolver(path string) schema.ColumnResolver {
 		var j map[string]interface{}
 		field := funk.Get(r.Item, path, funk.WithAllowZero())
 
+		if field == nil {
+			return nil
+		}
```
bbernays, line 144: I would suggest we set the value to `nil`... that way we can always be deterministic about `nil` vs `empty` vs `{}`

bbernays (changes requested):
Can you also add a test case to the SNS Subscriptions resolver that uncovered this whole issue?

https://github.com/cloudquery/cloudquery/pull/2487/files/5ff17a8ca613d3a8b1fc7c5a0563d384f7e3387d#diff-61779f3c80da17b5846ec98a4c15a3802bbde29561a9b2dffe6a60e649b28575

hermanschaaf (approved):

bbernays (approved):

PR #2487, opened 2022-10-06
Fix nil ptr error

bbernays:
This will be closed once @amanenk PR is merged


plugins/source/aws/client/resolvers.go
```diff
@@ -138,7 +138,9 @@ func MarshaledJsonResolver(path string) schema.ColumnResolver {
 	return func(_ context.Context, meta schema.ClientMeta, r *schema.Resource, c schema.Column) error {
```
yevgenypats, line 138: Do we have a test for this function? if not I think we should add it and add the test case that causes this.

bbernays: This PR will not be merged. See @amanenk's PR... #2486

PR #2523, opened 2022-10-07
chore(aws): Delete unused DeleteFilters

yevgenypats:
#### Summary

This is not solved on the sdk level via source `name` and `sync_time`


hermanschaaf:
> This is not solved on the sdk level via source name and sync_time

@yevgenypats Do you mean this is _now_ solved on the sdk level? 😄

PR #2530, opened 2022-10-07
feat: Add Support for HSMv2 Clusters and Backups

bbernays:


plugins/source/aws/codegen/recipes/cloudhsmv2.go
```diff
@@ -0,0 +1,47 @@
...
+			Multiplex:  `client.ServiceAccountRegionMultiplexer("cloudhsmv2")`,
+			ExtraColumns: append(
+				defaultRegionalColumns,
+				[]codegen.ColumnDefinition{
+					{
+						Name:     "arn",
+						Type:     schema.TypeString,
+						Resolver: `resolveClusterArn`,
+						Options:  schema.ColumnCreationOptions{PrimaryKey: true},
+					},
```
hermanschaaf, line 23: For consistency with other AWS resources, let's add a `ResolveTagField` resolver for `tags` here (using field `TagList`)

plugins/source/aws/codegen/recipes/cloudhsmv2.go
```diff
@@ -0,0 +1,47 @@
...
+			ExtraColumns: append(
+				defaultRegionalColumns,
+				[]codegen.ColumnDefinition{
+					{
+						Name:     "arn",
+						Type:     schema.TypeString,
+						Resolver: `resolveBackupArn`,
+						Options:  schema.ColumnCreationOptions{PrimaryKey: true},
+					},
+				}...),
```
hermanschaaf, line 39: Same comment here: for consistency with other AWS resources, let's add a `ResolveTagField` resolver for `tags` (using field `TagList`)

hermanschaaf (changes requested):
LGTM, just that one thing about `tags` I think we should address

plugins/source/aws/codegen/recipes/cloudhsmv2.go
bbernays: Done

yevgenypats (approved):

Issue #1710, opened 2022-09-03
Add Support For CockroachDB


mingfang:
I would like to use CockroachDB for storage instead of Postgres.
Because CockroachDB wire compatible with Postgres, I was about to use similar database connection URL.
But immediately at startup it failed with this error.
```
Error: could not init state: failed to migrate cloudquery_core tables: try lock failed in line 0: SELECT pg_advisory_lock($1) (details: pq: unknown function: pg_advisory_lock())
```
This is a known issue https://github.com/cockroachdb/cockroach/issues/13546
Ideally Cloudquery add native support for CockroachDB.

yevgenypats:
@mingfang Thanks! Actually we are rolling this next week - we are working on CloudQuery v2 - https://github.com/cloudquery/cloudquery/pull/1463 which separates the plugins from the storage layer as well as remove any non standard PG operations so cockroach will work out-of-the-box in addition to new storage layers coming up such as BigQuery/Snowflake etc...

yevgenypats:
@mingfang This is now out 🥳  with destination plugin `postgresql` version `v1.1.0`

https://github.com/cloudquery/cloudquery/pull/2566

mingfang:
@yevgenypats Just tested and confirmed working. Thank you very much!

yevgenypats:
Amazing, we've added cockroach also to our CI tests but please feel free to let us know if you hit any issues on Cockroach (or PostgreSQL).

mingfang:
Not sure if this is cockroachdb related,  but I'm getting this error with postgres plugin v1.3.3 and tried upgrading my aws plugin from v2.1.0 to v3.0.1...
```
Error: failed to sync source aws: failed to migrate source aws on destination postgresql : failed to call Migrate: rpc error: code = Unknown desc = failed to commit transaction to recreate primary keys: ERROR: transaction committed but schema change aborted with error: (23505): failed to ingest index entries during backfill: duplicate key value violates unique constraint "aws_elasticache_engine_versions_cqpk" (SQLSTATE XXA00)
```

PR #2531, opened 2022-10-08
feat(pg): Support CockroachDB

yevgenypats:
#### Summary

CockroachDB is PostgreSQL compliant but there always some differences so this PR address those and adds it to the testing workflow.

Closes https://github.com/cloudquery/cloudquery/issues/1710


plugins/destination/postgresql/client/types.go
```diff
@@ -4,7 +4,16 @@ import (
 	"github.com/cloudquery/plugin-sdk/schema"
 )
 
-func SchemaTypeToPg(t schema.ValueType) string {
+func (c *Client) SchemaTypeToPg(t schema.ValueType) string {
+	switch c.pgType {
+	case pgTypeCockroachDB:
+		return c.SchemaTypeToCockroach(t)
+	default:
+		return c.SchemaTypeToPg10(t)
```
bbernays, line 12: I thought our minimum supported pg version was 11

bbernays:
Overall this looks good. My question is why add support for Cockroach DB to the PG plugin destination? Why not make a standalone plugin for CockroachDB? That way we can add any special optimizations to each plugin without worrying about breaking the other

yevgenypats:
> Overall this looks good. My question is why add support for Cockroach DB to the PG plugin destination? Why not make a standalone plugin for CockroachDB? That way we can add any special optimizations to each plugin without worrying about breaking the other

Because our pg plugin should work with cockroachdb out of the box. The whole point of CockroachDB is that is is PG compliant. Right now we have 95% of the same code so unless the plugin will turn into 50% code for Cockroach and 50% for DB only then we should look into splitting this. It's even the same driver ('pgx').

A good rule of thumb should be `driver/library` per `plugin`

plugins/destination/postgresql/client/types.go
yevgenypats: I tested it locally, we should change our Postgresql In CI to 10. there is no feature in 11 right now that we use that is not available in 10

erezrokah (changes requested):
The code looks good, but a few comments:

1. ~~[non-blocking] Shouldn't this be a separate destination plugin? I get that the logic is mostly the same, but maybe we can re-use the code via a shared module, etc?~~ Saw that you commented to @bbernays but I'd rather duplicate the code first and then pull the duplication into a shared library. That way we won't need to worry about breaking `CockroachDB` when we can Postgres and vice versa. Also we could de-couple the tests.

2. [non-blocking] How would the configuration for `CockroachDB` look like? Looks like the same and we auto detect the DB type. Maybe better to be explicit and set the DB type in the spec?

3. [blocking] I'd keep the minimal requirement as Postgres 11, at least until we verify all of our policies queries work on 10 (or modify them to do so). Is 10 a requirement for this PR?

4. [blocking] [Looks like there a few differences for `CockroachDB`](https://www.cockroachlabs.com/docs/stable/postgresql-compatibility.html#unsupported-postgresql-features):
- It can have only [1 primary key](https://www.cockroachlabs.com/docs/stable/primary-key.html). And dropping a key needs to be via a single transaction. Do we need to handle it in https://github.com/cloudquery/cloudquery/blob/283fca902bcc4cbafb268c8fca798e9d73bcf465/plugins/destination/postgresql/client/migrate.go#L121 
- Maybe there are some other changes too but I'll need to test a bit more

yevgenypats:
> The code looks good, but a few comments:
> 
> 1. ~[non-blocking] Shouldn't this be a separate destination plugin? I get that the logic is mostly the same, but maybe we can re-use the code via a shared module, etc?~ Saw that you commented to @bbernays but I'd rather duplicate the code first and then pull the duplication into a shared library. That way we won't need to worry about breaking `CockroachDB` when we change Postgres and vice versa. Also we could de-couple the tests.
> 2. [non-blocking] How would the configuration for `CockroachDB` look like? Looks like the same and we auto detect the DB type. Maybe better to be explicit and set the DB type in the spec?
> 3. [blocking] I'd keep the minimal requirement as Postgres 11, at least until we verify all of our policies queries work on 10 (or modify them to do so). Is 10 a requirement for this PR?
> 4. [blocking] [Looks like there a few differences for `CockroachDB`](https://www.cockroachlabs.com/docs/stable/postgresql-compatibility.html#unsupported-postgresql-features):
> 
> * It can have only [1 primary key](https://www.cockroachlabs.com/docs/stable/primary-key.html). And dropping a key needs to be via a single transaction. Do we need to handle it in https://github.com/cloudquery/cloudquery/blob/283fca902bcc4cbafb268c8fca798e9d73bcf465/plugins/destination/postgresql/client/migrate.go#L121
> * Maybe there are some other changes too but I'll need to test a bit more



> * 1 primary key

1) I don't think we should separate it, pgx has the save driver and so should we. If we are breaking something in pg it means we are doing something wrong as for an insertion plugin it really should work for both. 
2) Configuration is the same for cockroachDB as this is basically the idea behind Cockroach
3) I don't think we should keep minimal support at 11 because it is just artificial version. our policies are not related to data types supported at insertion time so for policies we should mention that is is minimum 11 but not for a pg destination when it is clearly working for 10. If we will see in the code that we are starting to put a lot of effort to support 10 then sure we can drop it but at the moment this works out-of-the-box
4) This is a PostgreSQL limitation as well it can have only one primary key (it can be composite - and the same in cockroach). The only different that you rightfully mentioned is that cockroach must have a pk and this is already solved as we have `_cq_id` for the exact reason when we don't have other pks. The other thing we might want to do is to alter the pk in transaction and/or use `alter ` statement - but this is a best practices actually also for PostgreSQL so I'll push this in the PR.

yevgenypats:
In anycase even if we will split in the future due to unforseen requirements constraints I don't think we should do it now given how easy it is to support both with current constraints/requirements. So basically let's go with the easy solution first and then see if any requirements will come up that will require splitting it and much more work.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -119,15 +119,43 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table) erro
...
 			return fmt.Errorf("failed to add primary key constraint on table %s: %w", table.Name, err)
 		}
+		if err := tx.Commit(ctx); err != nil {
+			return fmt.Errorf("failed to commit transaction to recreate primary keys: %w", err)
+		}
+	}
+	return nil
+}
+
+func (c *Client) setNullOnPks(ctx context.Context, table *schema.Table) error {
```
erezrokah, line 153: Should this be called `setNotNullOnPks`?

Also let's say someone changes an existing column to be a primary key and it has a null value. What will happen? Will this fail and if so, how do we handle this failure?

erezrokah (approved):
Assuming this doesn't break Postgres I think we can give it a go, though I would maybe keep it as a hidden feature until we can do more testing on it

plugins/destination/postgresql/client/migrate.go
yevgenypats: Yes this will fail. we don't have (and will never have) a way to deal with that. The user will have to either drop the table or migrate columns to contain non-null.

auto-migrations can fail in data integrations platform so it's best effort and it is acceptable/industry standard as maintaining migrations for all databases*all_apis is not really feasible.

yevgenypats:
> Assuming this doesn't break Postgres I think we can give it a go, though I would maybe keep it as a hidden feature until we can do more testing on it

Thanks! I don't think this will break Postgres and I wouldn't keep it under feature flag as otherwise no one will be able to help us test it (I think just one or two users asked for it - so now we wait for people to try out so we can fix the bugs)

erezrokah:
> Thanks! I don't think this will break Postgres and I wouldn't keep it under feature flag as otherwise no one will be able to help us test it (I think just one or two users asked for it - so now we wait for people to try out so we can fix the bugs)

Yeah, no need for a flag, "hidden" means no documentation for it :)

PR #2543, opened 2022-10-09
chore(website): Remove deprecated dev docs

yevgenypats:
#### Summary

Our dev docs still needs much more work but for now it's better just to redirect to examples rather then having deprecated old docs.


erezrokah (approved):
Nice 🚀 

Maybe also add a redirect from the old doc(s) to the new ones here https://github.com/cloudquery/cloudquery/blob/a108d71174ee23a16668d6a6ae0dd5431e6fdb41/website/vercel.json#L28

Issue #2047, opened 2022-09-25
feat(docs-v2): For each plugin we should have a full config example so users can copy paste it


erezrokah:
Looking at the new docs, it's hard for me to understand how the configuration should look like.
For example in https://github.com/cloudquery/cloudquery/blob/8efaec1d7f0a24b4b4ff9834376c7681f2fed8f9/plugins/source/aws/docs/configuration.md I'm not sure if the configuration items should go under the root level of the YAML, e.g.

```yaml
regions: [us-east-1]
```

or they should be nested or at which level.

Same for the readme https://github.com/cloudquery/cloudquery/tree/8efaec1d7f0a24b4b4ff9834376c7681f2fed8f9/plugins/source/aws#shared-configuration-files:
![image](https://user-images.githubusercontent.com/26760571/192136342-6919e3d2-6a20-4f66-b38e-463ad5d32a1b.png)

This puts the configuration at the top level.

Finally, with https://github.com/cloudquery/cloudquery/pull/2038/files#diff-727bced4f01abf7492e76921b5f9103e1c5ad88029c47df644148a724a7eefd7R19 we document one part of the configuration in our Website, and the other in GitHub, so in order to assemble a configuration users will need to look in multiple places.

Providing an example for each plugin with all configuration items should make it easier to understand how to configure it

yevgenypats:
I know it might be confusing in the beginning but I think we need to set a very clear convention over configuration and maybe document it a bit better in the docs in the following way:

Main docs: document `source` and `destination` specs and all available options. When we change SDK or add new version we always update this only in one place.
Plugin docs should never reference full spec and should always document only their own spec, this way we make the documentation consistent and we don't need to go through all the docs updating SDK docs.

I think it is really important we keep it this way, potentially document better in our main docs what is the convention.

erezrokah:
I think we can automate updating plugins docs with SDK fields, but until we do that we could have for example in AWS

```yaml
kind: source
spec:
   # global source configuration
   ...
   spec:
      // AWS regions...
      regions: []
```

WDYT? That way we don't need to update the plugin docs when SDK changes, and it's clear where to put stuff and what the indentation should be.

WDYT?

yevgenypats:
That will work with the following caveat/comments:
- I don't think we should create any kind of automation updating all the options for source and dest in all plugins as this also wont work for community ones and it is just a lot of work.
- Just having an example with a comment like in your example should work well I think

[...]

PR #2575, opened 2022-10-09
chore(website): Add recipes section

yevgenypats:
This adds section of useful configs for most used source/destination plugins that users can just copy paste (with auto-update pinned versions as part of CI).

Closes: https://github.com/cloudquery/cloudquery/issues/2047


hermanschaaf (approved):
Nice ✨

PR #2587, opened 2022-10-10
fix: Update postgresql plugin to work with new sdk (deleteStale feature)

shimonp21:
#### Summary

This will only compile after https://github.com/cloudquery/plugin-sdk/pull/269 is released

### DIFF
--- a/plugins/destination/postgresql/client/deletestale.go
+++ b/plugins/destination/postgresql/client/deletestale.go
@@ -14,9 +14,9 @@ func (p *Client) DeleteStale(ctx context.Context, table string, source string, s
 	sb.WriteString("delete from ")
 	sb.WriteString(pgx.Identifier{table}.Sanitize())
 	sb.WriteString(" where ")
-	sb.WriteString(schema.CqSourceName.Name)
+	sb.WriteString(schema.CqSourceNameColumn.Name)
 	sb.WriteString(" = $1 and ")
-	sb.WriteString(schema.CqSyncTime.Name)
+	sb.WriteString(schema.CqSyncTimeColumn.Name)
 	sb.WriteString(" < $2")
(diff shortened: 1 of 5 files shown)


plugins/destination/postgresql/client/deletestale.go
```diff
@@ -9,15 +9,23 @@ import (
...
-	sb.WriteString(schema.CqSourceName.Name)
+	sb.WriteString(" where (")
+	sb.WriteString(schema.CqSourceNameColumn.Name)
 	sb.WriteString(" = $1 and ")
-	sb.WriteString(schema.CqSyncTime.Name)
-	sb.WriteString(" < $2")
+	sb.WriteString(schema.CqSyncTimeColumn.Name)
+	sb.WriteString(" < $2) or ")
+	sb.WriteString(schema.CqSourceNameColumn.Name)
+	sb.WriteString(" is null or ")
```
yevgenypats, line 26: why is that ? I wouldn't delete anything that is not 100% ours. if users upgrade to a new version it is up to him to drop the database and re-run in `overwrite-delete-stale` mode. we shouldn't go into edgecase otherwise we can delete user's data.

yevgenypats (changes requested):
Looks good. One comment on query.

plugins/destination/postgresql/client/deletestale.go
shimonp21: 1) This section only runs when the users runs in `overwrite-delete-stale` mode.
2) This code removes the need to drop the database (and, if the user doesn't drop the database, removes resources that can hang there forever)
3) The 'dropping' of the data will only happen **once**. After the next run the behaviour will always be as-expected.

yevgenypats (approved):

PR #2619, opened 2022-10-11
feat: Add anonymised analytics

yevgenypats:
#### Summary

This still needs more testing once we fix main.

This introduce anonymised analytics that will help us get a few key metrics:
- Important: source and destination plugins used including versions
- Important: resources/data synced with each plugin
- Semi-Important: Fetch Error, Write Errors, Panics

Few notes:
- We don't need a note on this apart from the documentation because those analytics are fully anonymysed (GDPR compliant, similar to server side analytics). 
- We don't and shouldn't try to identify a user with a cookie or database id because this information is not ours and then we will need also to get a consent. Also, this will still give innacurate information about number of users so it's just best not to collect it in the first place and run in circles of trying to identify a user where in the first place we say it is anonymysed.
- I don't think it will ever become a need unless we introduce our own hub or GitHub will put a limitation on release downloads in which case we might need to introduce login to make sure our egress is not abused.
- We shouldn't turn the analytics into a product of itself and/or turn it into a funnel optimisation thing because we don't really have one as the product is really simple now - it gets data from one place to another.
- Any performance relates things most prob shouldn't go into here as well and in any-case we don't really have performance monitor for CQ, so this should come once we solve this with prometheus or whatever we decide.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,14 +2,12 @@ package cmd
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"strings"
 
 	"github.com/cloudquery/cloudquery/cli/internal/enum"
-	"github.com/getsentry/sentry-go"
-	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"github.com/rudderlabs/analytics-go"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
@@ -25,18 +23,32 @@ Open source data integration that works.
 
 Find more information at:
 	https://cloudquery.io`
+	telemetryLevels = []string{"none", "errors", "all"}
+
+	analyticsClient *Analytics
 )
+
+
+func strInArray(str string, arr []string) bool {
+	for _, s := range arr {
+		if s == str {
+			return true
+		}
+	}
+	return false
+}
+
 
 func NewCmdRoot() *cobra.Command {
 	logLevel := enum.NewEnum([]string{"trace", "debug", "info", "warn", "error"}, "info")
 	logFormat := enum.NewEnum([]string{"text", "json"}, "text")
-	noColor := false
 	logConsole := false
 	noLogFile := false
 	logFileName := "cloudquery.log"
 	sentryDsn := sentryDsnDefault
 
 	var logFile *os.File
+	var analyticsClient analytics.Client
 	cmd := &cobra.Command{
 		Use:     "cloudquery",
 		Short:   rootShort,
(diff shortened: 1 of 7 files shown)


cli/cmd/root.go
```diff
@@ -118,25 +98,13 @@ func NewCmdRoot() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&logFileName, "log-file-name", "cloudquery.log", "Log filename")
 
 	// Telemtry (analytics) flags
-	cmd.PersistentFlags().Bool("no-telemetry", false, "disable telemetry collection")
+	cmd.PersistentFlags().String("telemetry-level", "all", "Telemetry level (none, errors, all)")
```
erezrokah, line 101: What's the use case to have multiple levels? Maybe we can keep it as a boolean for now, due to:
1. Renaming the flag is a breaking change, so better to do it in a backwards compatible
2. I think it will be hard to explain the differences in the levels as it's mostly an internal implementation

cli/cmd/root.go
```diff
@@ -149,4 +117,9 @@ func initViper() {
 	viper.AutomaticEnv()
 	viper.SetEnvPrefix("CQ")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	viper.SetConfigName("config")    // name of config file (without extension)
```
erezrokah, line 120: Where do we use these settings?

cli/cmd/sync.go
```diff
@@ -191,5 +192,21 @@ func syncConnection(ctx context.Context, sourceSpec specs.Source, destinationsSp
 	fmt.Printf("Summary: resources: %d, errors: %d, panic: %d, failed_writes: %d, time: %s\n", totalResources, summary.Errors, summary.Panics, failedWrites, tt.Truncate(time.Second).String())
 	log.Info().Str("source", sourceSpec.Name).Strs("destinations", sourceSpec.Destinations).
 		Int("resources", totalResources).Uint64("errors", summary.Errors).Uint64("panic", summary.Panics).Uint64("failed_writes", failedWrites).Float64("time_took", tt.Seconds()).Msg("Sync completed successfully")
+	if err := analyticsClient.Enqueue(analytics.Track{
```
erezrokah, line 195: Shouldn't this be under `sync` and not `syncConnection`. Meaning we'd like to track each invocation of the `sync` command by every user, and not the invocation of each source?

Then we can send the list of sources and list of destinations in a single event

erezrokah (dismissed):
Not that familiar with Rudder, but added a few comments on the code

cli/cmd/root.go
yevgenypats: 1. I don't think anyone is using this flag since we first released it last week. This was specifcally asked by the team here - https://github.com/cloudquery/plugin-sdk/issues/128 (I think by you :) )
2. I don't mind splitting those we just need to decide. I merged it because it was requested.

cli/cmd/root.go
yevgenypats: It will give people the ability to turn of analytics permanently without passing it on each run

cli/cmd/root.go
erezrokah: Ah, I think we meant to just not call the flag `sentry` and use a non implementation detail named flag like `telemetry`

cli/cmd/root.go
erezrokah: Can you share an example on how this config file will look like? Not sure we'd want to support another configuration API, and if so, maybe call it `cq_config.yaml`? Or `cloudquery.yaml`.

cli/cmd/sync.go
```diff
@@ -191,5 +192,21 @@ func syncConnection(ctx context.Context, sourceSpec specs.Source, destinationsSp
...
+		UserId: "test",
+		Event:  "sync",
```
erezrokah, line 197:
```suggestion
		Event:  "sync_source",
```

cli/cmd/sync.go
```diff
@@ -191,5 +192,21 @@ func syncConnection(ctx context.Context, sourceSpec specs.Source, destinationsSp
...
+			"source_path":    sourceSpec.Path,
+			"source_version": sourceSpec.Version,
+			"destinations":   sourceSpec.Destinations,
+			"resources":      totalResources,
+			"errors":         summary.Errors,
+			"panic":          summary.Panics,
+			"failed_writes":  failedWrites,
+		},
+	}); err != nil {
+		log.Error().Err(err).Msg("Failed to enqueue analytics event")
```
erezrokah, line 209: Maybe we'd want to make this a warning? Or even a debug? I'm not sure the user cares that much about a failed analytic invocation

yevgenypats:
After some more thought Im going to improve this a bit more and get rid of this rudder stuff. Just going to have a small go server at `analytics.cloudquery.io` that send directly to bigquery table via streaming api. this way we can easily change the backend if ever needed.

erezrokah:
> Just going to have a small go server at `analytics.cloudquery.io`

I like that. Having a proxy ensures we can improve our analytics without changing the CLI that often, maybe consider using a Serverless function for it hosted on Vercel

yevgenypats:
> > Just going to have a small go server at `analytics.cloudquery.io`
> 
> I like that. Having a proxy ensures we can improve our analytics without changing the CLI that often, maybe consider using a Serverless function for it hosted on Vercel

better to use google cloud run for this, especially as our database is there.

erezrokah:
@yevgenypats how about we close this PR for now?

cli/cmd/root.go
yevgenypats: This is just standard viper (https://github.com/spf13/viper#reading-config-from-ioreader) so users will be able to use 

```
telemetry-level: "off"
```

for flags that are passed to viper (like telemetry)

cli/cmd/sync.go
yevgenypats: I think we only have `sync` as there is no such thing as `sync_source` a sync is always source->to multiple dests

It's just a cli feature that can do those in a loop but for analytics purposes it's all the same event (there wont be a `sync` event and a `sync_source` event)

cli/cmd/sync.go
yevgenypats: done

yevgenypats:
> @yevgenypats how about we close this PR for now?

I don't think we should close it as it's ready for re-review and `analyticsv1.cloudquery.io` is deployed in prod

yevgenypats:
backend code - https://github.com/cloudquery/backend/tree/main/analytics

cli/cmd/root.go
yevgenypats: I also went further ahead and I removed viper library because it's not a great one, also we just use environment variable for some flags and if we use viper we loose the built-in cobra flag verification and need to reimplement. Moreover this also causes a confusion on when we should use viper and/or the variable bound via the flags because viper is only one way binding.

Now should be pretty simple and I just shaved up 1MB from our CLI because viper imports hcl,toml,yaml2,yaml3,ini. I think the big one was hcl as it's basically another language.

website/pages/docs/reference/cli/cloudquery.md
```diff
@@ -17,15 +17,15 @@ Find more information at:
...
-      --data-dir string        set persistent data directory (env: CQ_DATA_DIR) (default "./.cq")
-  -h, --help                   help for cloudquery
-      --log-console            enable console logging
-      --log-file-name string   Log filename (default "cloudquery.log")
-      --log-format string      Logging format (json, text) (default "text")
-      --log-level string       Logging level (default "info")
-      --no-log-file            Disable logging to file
-      --no-telemetry           disable telemetry collection
+      --color string             Enable colorized output (on, off, auto) (default "auto")
+      --data-dir string          set persistent data directory (env: CQ_DATA_DIR) (default "./.cq")
```
cqgaurav, line 21: Should we remove this too https://github.com/cloudquery/cloudquery/issues/2644 ?

yevgenypats: This is already handled here (This PR just needs a rebase) - https://github.com/cloudquery/cloudquery/pull/2796

cli/cmd/root.go
```diff
@@ -118,35 +85,11 @@ func NewCmdRoot() *cobra.Command {
...
-
-	hiddenFlags := []string{"telemetry-inspect", "telemetry-debug", "sentry-dsn"}
-	for _, f := range hiddenFlags {
-		err := cmd.PersistentFlags().MarkHidden(f)
-		if err != nil {
-			panic(err)
-		}
-	}
-	initViper()
+	cmd.PersistentFlags().Var(telemetryLevel, "telemetry-level", "Telemetry level (none, errors, all)")
```
hermanschaaf, line 88: There is some inconsistency here between `none, errors, all` and the enum on line 29 https://github.com/cloudquery/cloudquery/pull/2619/files#diff-eb154bbd601602eeae895e7af6a12b88cbdb7d585a2d32216c398f68cabcdff5R29 (`"trace", "debug", "info", "warn", "error"`). I assume we want the enum to be `none, errors, all`? I will push a fix for this

website/pages/docs/reference/cli/cloudquery.md
hermanschaaf: Rebased 👍

hermanschaaf (approved):
👍

Issue #2492, opened 2022-10-07
Remove the use of github.com/bxcodec/faker


yevgenypats:
Right now we use **heavly** github.com/bxcodec/faker in alot of our mock tests and in some of the new tests we use our own https://github.com/cloudquery/plugin-sdk/blob/main/faker/faker.go. 

We should move all of the mock tests to our own faker or at the very least disallow new code with `github.com/bxcodec/faker`.

Most of `github.com/bxcodec/faker` capabilities are mainly around generating random types of data like credit card and so on which we don't really need but the biggest issue is that this library wasn't built for production use and it is sharing all options across all objects instance which can cause race conditions and mock tests to fail because it is accessing maps concurrently.

https://github.com/bxcodec/faker/blob/master/faker.go#L243

This might also be a blocker to turn our mock tests into parallel with `t.Parallel` https://github.com/cloudquery/plugin-sdk/blob/main/plugins/source_testing.go#L30

bbernays:
A couple of questions:
1. Do we have any examples of the mock tests failing because of this library?
2. How much time or costs could be saved by running these tests in parallel?

erezrokah:
> 1\. Do we have any examples of the mock tests failing because of this library?

I experienced a couple. `bxcodec/faker` doesn't handle circular structs very well, and also inserts random data which makes it harder to test.


> 2\. How much time or costs could be saved by running these tests in parallel?

By default `go test` runs in parallel, so basically we've seen failures with `bxcodec/faker` just by running `go test` on the mock tests.

erezrokah:
This is listed as `fast follow` for v1 [here](https://www.notion.so/cloudquery/v2-Created-ed27199b013c4f6b82131afdcd6087fc#7c4573e7365441c6b780fd84f71dce54) (internal link)

yevgenypats:
I don't think they are running in parallel by default - https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently

erezrokah:
> I don't think they are running in parallel by default - https://stackoverflow.com/questions/24375966/does-go-test-run-unit-tests-concurrently

Ah good point. I did see `bxcodec/faker` fail on map write concurrency errors regardless, so probably due to:
![image](https://user-images.githubusercontent.com/26760571/195082052-cff31578-e126-4fe8-a268-64803bd50f47.png)

Issue #2474, opened 2022-10-06
bug(tests): Sporadically failing AWS test


hermanschaaf:
This error occurred in CI, but I was unable to reproduce it locally:

```
panic: runtime error: index out of range [0] with length 0

goroutine 20 [running]:
github.com/cloudquery/cloudquery/plugins/source/aws/resources/services/firehose.buildKinesisFirehoses(_, _)
	/home/runner/work/cloudquery/cloudquery/plugins/source/aws/resources/services/firehose/delivery_streams_mock_test.go:33 +0x5c7
github.com/cloudquery/cloudquery/plugins/source/aws/client.AwsMockTestHelper.func1({_, _}, {{0xc9cd60, 0xc000558bf0}, 0x0, {0x0, 0x0}, {0xc0005fe000, 0x1, 0x1f4}, ...}, ...)
	/home/runner/work/cloudquery/cloudquery/plugins/source/aws/client/testing.go:33 +0x222
github.com/cloudquery/plugin-sdk/plugins.(*SourcePlugin).Sync(_, {_, _}, {{0xc9cd60, 0xc000558bf0}, 0x0, {0x0, 0x0}, {0xc0005fe000, 0x1, ...}, ...}, ...)
	/home/runner/go/pkg/mod/github.com/cloudquery/plugin-sdk@v0.12.7/plugins/source.go:107 +0xbc
github.com/cloudquery/plugin-sdk/plugins.TestSourcePluginSync.func1()
	/home/runner/go/pkg/mod/github.com/cloudquery/plugin-sdk@v0.12.7/plugins/source_testing.go:37 +0xe7
created by github.com/cloudquery/plugin-sdk/plugins.TestSourcePluginSync
```

yevgenypats:
We shouldn't use `"github.com/bxcodec/faker"` anymore but our own which is available in the SDK and is not storing configuration parameters as global variables.

bbernays:
I am not sure I follow... What config params are you referring to?

yevgenypats:
Opened an issue - https://github.com/cloudquery/cloudquery/issues/2492

PR #2622, opened 2022-10-11
chore(tests): Remove use of github.com/bxcodec/faker

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/2492
Closes https://github.com/cloudquery/cloudquery/issues/2474


erezrokah (approved):

erezrokah:
I think this is technically a `chore` as it's not a user facing change

hermanschaaf:
@erezrokah Oh yeah, thanks 👍

yevgenypats (approved):
Awesome. Curious if we can also block this library from usage in one of the linters cc @erezrokah

PR #2627, opened 2022-10-11
fix!: Use JSON tag for column names

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/2189, extracted from https://github.com/cloudquery/cloudquery/pull/2623. Tests for code gen drift for the Cloudflare plugin.

This is a breaking change as it renames a few columns due to https://github.com/cloudquery/plugin-sdk/pull/112

PR #2629, opened 2022-10-11
fix!: Rename certificates `sha_1_fingerprint` to `sha1_fingerprint`

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/2627 for DigitalOcean.

Renamed `resources.go` to `main.go` for consistency with other plugins. Also a breaking change due to the column rename.


hermanschaaf:
We can probably override the column name in this case if we want to make the change non-breaking. Technically it's called `sha-1` (with a dash) so I think either with or without an underscore is fine. (Though our convention is generally to not have an underscore)

erezrokah:
> We can probably override the column name in this case if we want to make the change non-breaking. Technically it's called `SHA-1` (with a dash) so I think either with or without an underscore is fine. (Though our convention is generally to not have an underscore)

That's a good point. `sha1_fingerprint` is taken from the JSON tag on the struct that's the reason for the change. I'm not sure we have a case we override the JSON tag.

hermanschaaf:
@erezrokah We could override it in the recipe, right? Skip the field and add it manually with the name `sha_1_fingerprint ` to make it non-breaking.

hermanschaaf:
Not that it's a big deal. I'm good either way, if you'd prefer to make keep things consistent and make this a breaking change.

erezrokah:
> Not that it's a big deal. I'm good either way, if you'd prefer to make keep things consistent and make this a breaking change.

Yeah, I think we always use the JSON tag if exists. If it didn't have one we could keep the the `sha_1_`

hermanschaaf (approved):

Issue #2189, opened 2022-10-02
Add CI tests for detecting drifts between code and codegen resources


erezrokah:
We should run the code gen for each plugin a fail if resources change

erezrokah:
https://github.com/cloudquery/cloudquery/pull/2643 and https://github.com/cloudquery/cloudquery/pull/2633 are pending reviews, but this should be done for all plugins that have code gen

PR #2635, opened 2022-10-11
fix!: Rename domains `c_name` to `cname`

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/2189.

1. Added codegen drift detection
2. Moved generated resources under `../resources/services` for consistency with other plugins (there's a follow up here to put each namespace in a separate directory, but didn't want to do too many changes)
3. Moved `plugin/plugin.go` to `resources/plugin/plugin.go`. This is required so Go Releaser embeds the correct version of the plugin during release, see https://github.com/cloudquery/cloudquery/blob/6ab042925afaf7fe0a7fccbedd17f0915d40f7cf/plugins/.goreleaser.yaml#L12


.github/workflows/source_heroku.yml
```diff
@@ -40,3 +40,5 @@ jobs:
         run: make gen-docs
       - name: Fail if docs are changed
         run: test "$(git status -s ./docs/tables | wc -l)" -eq 0
```
hermanschaaf, line 42: I think this is missing the `gen-code` step

hermanschaaf (changes requested):
Missing gen-code, otherwise LGTM 👍

.github/workflows/source_heroku.yml
erezrokah: Ah, good catch, fixed in https://github.com/cloudquery/cloudquery/pull/2635/commits/4ec26f7e298e5c870e1d68bb66853b4bc178067b

hermanschaaf (approved):
Nice, thanks for doing the refactoring work as well 👍

erezrokah:
@hermanschaaf looks like another breaking change due to https://github.com/cloudquery/cloudquery/pull/2635/files#diff-f8e4debb63fb716d5f8829d87f8dbdc19564da7bc53b4d3481d6cade51593fab:
image

PR #2634, opened 2022-10-11
fix(gcp): Migrate Grafana dashboards

disq:
Just like #2621 and #2628

PR #2621, opened 2022-10-11
fix(aws): Migrate grafana dashboards

disq:
closes #2183

I couldn't get the `route_assoc` variable to work.

PR #2638, opened 2022-10-11
fix(k8s): Migrate Grafana dashboards

disq:
Just like #2634, #2621 and #2628 but this time completely untested.


erezrokah (approved):
So this is still much better that the current state (even untested), so I think we could merge this and fix any issues in a follow up PR

PR #2643, opened 2022-10-11
chore(ci): Fix Azure code gen

erezrokah:
#### Summary

Copy paste error fix


hermanschaaf (approved):

.github/workflows/source_azure.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: go run main.go doc docs/tables
```
hermanschaaf, line 42: Maybe this should use `make gen-docs`?

PR #2645, opened 2022-10-11
feat: Add Support for ECR Public Repositories

bbernays:
#### Summary

a few notes:
- `api.ecr-public` is not in the AWS manifest file we use for determining regional availability so had to update the generation file (don't love the solution but open to changing/replacing it with something else)
- Adds support for repositories (and images)... Once ECR Registries is merged we can add support for public registries as the gen + fetch + mock code is nearly identical


plugins/source/aws/client/data/partition_service_region.json
```diff
@@ -260,6 +260,12 @@
             "us-west-2": {}
           }
         },
+        "api.ecr-public": {
+          "regions": {
+            "us-east-1": {},
+            "us-west-2": {}
+          }
+        },
```
hermanschaaf, line 268: Hmm... assuming we got this file from somewhere, this can cause issues the next time we update it. Where do we get it from? Is it provided by AWS? Is ecr-public also not included in the latest version? 

Again assuming that this file was provided to us somehow, if we have to make a manual change, I think we should put it in a separate JSON file (called `paritition_service_region_overrides.json` or something like that) so we can be explicit about this being an addition that we made.

hermanschaaf: (See my later comment lower down - seems like my concerns were addressed)

hermanschaaf:
Looks good, but there's a conflict now: docs will need to be regenerated after merging main into this branch.

disq (approved):

PR #2651, opened 2022-10-12
fix: Flush destination plugins when sync finish

yevgenypats:
Seems like this bug was hanging there but now when we started using batch in our postgresql plugin we were hit by this.


erezrokah (approved):

hermanschaaf (approved):
Well spotted!

disq (approved):

PR #2655, opened 2022-10-12
chore(ci-cli): Check for docs drift

erezrokah:
#### Summary

The same we have for plugins, this workflow checks that the CLI docs are up to date

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -7,8 +7,8 @@ test:
 	go test ./...
 
 
-.PHONY: doc
-doc:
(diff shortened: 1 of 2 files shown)


cli/Makefile
```diff
@@ -7,8 +7,8 @@ test:
 	go test ./...
 
 
-.PHONY: doc
-doc:
+.PHONY: gen-docs
```
erezrokah, line 10: Renamed this for consistency with the plugins make file

PR #2675, opened 2022-10-12
doc: Change default write_mode to 'overwrite-delete-stale'

shimonp21:

### DIFF
--- a/website/pages/docs/reference/destination-spec.md
+++ b/website/pages/docs/reference/destination-spec.md
@@ -35,12 +35,13 @@ For example `name: postgresql` will resolve `path` to `cloudquery/postgresql`. I
 
 ### write_mode
 
-(`string`, optional, default: `overwrite`. Available: `overwrite`, `overwrite-delete-stale`, `append`)
+(`string`, optional, default: `overwrite-delete-stale`. Available: `overwrite-delete-stale`, `overwrite`, `append`)
 
 Specifies the update method to use when inserting rows. The exact semantics depend on the destination plugin, and all destinations don't support all options, so check the destination plugin documentation for details.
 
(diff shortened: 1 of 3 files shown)


hermanschaaf (approved):

disq (approved):

website/pages/docs/reference/destination-spec.md
```diff
@@ -35,12 +35,13 @@ For example `name: postgresql` will resolve `path` to `clouduquery/postgresql`.
...
 
-(`string`, optional, default: `overwrite`. Available: `overwrite`, `overwrite-delete-stale`, `append`)
+(`string`, optional, default: `overwrite-delete-stale`. Available: `overwrite-delete-stale`, `overwrite`, `append`)
 
 Specifies the update method to use when inserting rows. The exact semantics depend on the destination plugin, and all destinations don't support all options, so check the destination plugin documentation for details.
 
-- `overwrite`: Overwrites existing rows with the same primary key, but doesn't delete stale rows from previous syncs.
-- `overwrite-delete-stale`: Same as `overwrite`, but stale rows are also deleted.
+- `overwrite-delete-stale`: `sync`s overwrite existing rows with the same primary key, and delete rows that 
+                             are no longer present in the cloud.
```
disq, line 43: extra padded newline?

yevgenypats (changes requested):
Is it the default right now in the SDK? If not I suggest to wait before we make it a default, as Im not sure which default is better the one that doesn't delete data by default or the one that does without any warning.

website/pages/docs/reference/destination-spec.md
shimonp21: looks OK after MD does it's thing :)

yevgenypats (approved):

PR #2686, opened 2022-10-12
feat: Add API URL for many AWS resources

hermanschaaf:
This adds a URL to API reference as a table description for most tables from `accessanalyzer` through to `ec2`.

I can add more in a follow-up PR, I'm just mindful of the GitHub workflow limit of 300 files per PR.


erezrokah (approved):
🚀 I think it's ok to cross the 300 limit as long as the changes are only for a single plugin

hermanschaaf:
@erezrokah True, in this case it's probably okay. But maybe smaller PRs are better anyway, this one is already quite big :)

Issue #2682, opened 2022-10-12
chore(website): Replace subscribe serverless function with client side subscribe


erezrokah:
## Problem description

On our Website we have a form to subscribe to our newsletter.
It uses a Serverless Vercel function to handle the subscribe operation:
https://github.com/cloudquery/cloudquery/blob/10be7695e6164ed099c316b531ff6166724cc452/website/pages/api/subscribe.ts#L13

Since the function uses secrets we set those as environment variables in the Vercel project configuration.
As a result any PR from [a fork needs to be authorized by a team member.](https://vercel.com/docs/concepts/git#deploying-forks-of-public-git-repositories)

## TLDR

Having a subscribe function adds friction to every PR from a forked repository

## Solution

We don't need a function to handle the subscribe operation, as we can do it in the client side, see https://mailchimp.com/help/host-your-own-signup-forms/.

Once we do that we can remove the environment variables from the Vercel project and forked PRs should not require authorization

PR #2687, opened 2022-10-12
chore(website): Use client side submit

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/2682. Will test on the deploy preview, then move to ready to review


website/components/Footer.tsx
```diff
@@ -142,50 +142,36 @@ export function Footer() {
...
-      onSubmit={subscribe}
+      action="https://cloudquery.us1.list-manage.com/subscribe/post"
+      method="POST"
+      onSubmit={handleSubmit}
     >
       <label htmlFor="email-address" className="sr-only">
         Email address
       </label>
       <div>
+        <input type="hidden" name="u" value="5a433922a8db867a38c8a4a8e" />
```
erezrokah, line 164: Taken by following https://mailchimp.com/help/host-your-own-signup-forms/ (explained in the issue0

website/components/Footer.tsx
```diff
@@ -142,50 +142,36 @@ export function Footer() {
...
-      setMessage(error);
-
-      return;
-    }
-
     router.push("/confirm");
   };
 
   return (
     <form
```
erezrokah, line 153: This is now an uncontrolled component, so no need to track the email input and errors

website/components/Footer.tsx
```diff
@@ -145,21 +145,19 @@ function SubmitForm() {
...
+    const params = new URLSearchParams();
+    params.append("EMAIL", email);
+    params.append("u", "5a433922a8db867a38c8a4a8e");
+    params.append("id", "4a9aca3509");
+    params.append("c", "?");
 
-      return;
-    }
+    await fetch(
+      `https://cloudquery.us1.list-manage.com/subscribe/post-json?${params.toString()}`,
```
erezrokah, line 158: We need `post-json` and `no-cors` due to a Mailchimp limitation, which means we can't get the text response and check for errors.
If we want to handle errors we need to use https://github.com/camsong/fetch-jsonp to handle this `jsonp` request.
I don't expect this to fail so I think that's OK (it only failed for me when I tried to submit too many times due to rate limiting on Mailchimp side)

yevgenypats (approved):
very nice.

PR #2693, opened 2022-10-12
chore: Add Issue to public roadmap project

bbernays:


.github/workflows/issue_to_project.yml
```diff
@@ -23,3 +23,11 @@ jobs:
...
           resource_node_id: ${{ github.event.issue.node_id }}
           status_value: ${{ env.todo }} # Target status
+      - name: Add issue to Public Roadmap
+        uses: leonsteinhaeuser/project-beta-automations@v2.0.1
+        with:
+          gh_token: ${{ secrets.GH_CQ_BOT }}
+          organization: cloudquery
+          project_id: 7
+          resource_node_id: ${{ github.event.issue.node_id }}
+          status_value: No Status
```
erezrokah, line 33: I'm not sure `No Status` will work as it has to be one of:
https://github.com/orgs/cloudquery/projects/7/settings/fields/Status
image

Unless I'm missing something and there's another status field.
How about we drop `status_value` as it's not required

erezrokah (changes requested):
Looks good pending 1 comment. Also can we remove the other triaging step?

.github/workflows/issue_to_project.yml
bbernays: I am good with that

erezrokah:
@bbernays WDYT about removing the `- name: Move issue to ${{ env.todo }}` step? I'm not sure we need to add issues to 2 boards

bbernays:
> @bbernays WDYT about removing the `- name: Move issue to ${{ env.todo }}` step? I'm not sure we need to add issues to 2 boards

Sorry I had the changes local and forgot to push them

erezrokah (approved):

PR #2698, opened 2022-10-12
fix(GCP Policy): DNSSEC check gives incorrect results

SCKelemen:
The query for CIS 1.2.0 Section 3.3 gives incorrect results. 
All results fail, even though in some case the Cloud DNS API is not even enabled. This PR fixes the query.

- [ ] Update or add tests 🧪

Test Queries:
```sql
-- observe all checks fail
select * from gcp_policy_results
where check_id = '3.3';

-- observe that some cases now pass
select * from gcp_policy_results
where status = 'pass'
and project_id = 'cq-playground'
and check_id = '3.3';

-- observe that some cases now fail
select * from gcp_policy_results
where status = 'fail'
and project_id = 'cq-playground'
and check_id = '3.3';

-- find dns zones without dnssec enabled
select * from gcp_dns_managed_zones
where visibility != 'private'
and ((dnssec_config is null) or (dnssec_config->>'state' = 'off'));
```


yevgenypats (approved):
seems good to me. I'll let @disq take a look.

plugins/source/gcp/policies_v1/cis_v1.2.0/section_3.sql
```diff
@@ -10,7 +10,7 @@
 \ir ../queries/compute/legacy_network_exist.sql
 \set check_id '3.3'
 \echo "Executing check 3.3"
-\ir ../queries/compute/dnssec_disabled.sql
```
jsonpr, line 13: What are we planning on doing with the legacy query - are we planning on removing the file or leaving it?  Is it linked anywhere else?

plugins/source/gcp/policies_v1/queries/dns/zones_without_dnssec.sql
```diff
@@ -0,0 +1,20 @@
...
+-- where visibility != 'private'
+-- and ((dnssec_config is null) or (dnssec_config->>'state' = 'off'));
+
+
+INSERT INTO gcp_policy_results (resource_id, execution_time, framework, check_id, title, project_id, status)
+SELECT "id"                                                      AS resource_id,
+       :'execution_time'::timestamp                              AS execution_time,
+       :'framework'                                              AS framework,
+       :'check_id'                                               AS check_id,
+       'Ensure that DNSSEC is enabled for Cloud DNS (Automated)' AS title,
```
jsonpr, line 11: Agreed with naming - this gets weird as we could do:

zones_without_dnssec_enabled.

From a language perspective, we have:
* With or Without
* Enabled or Not Enabled (Disabled/Off)
* Then Fail or Pass.

plugins/source/gcp/policies_v1/cis_v1.2.0/section_3.sql
SCKelemen: I think I'll remove this query, since it doesn't appear to be useful at all.

plugins/source/gcp/policies_v1/queries/dns/zones_without_dnssec.sql
SCKelemen: I think it's probably best to say 
`zones_with_dnssec_disabled`, but technically,  the query is looking for zones where it is not enabled. The difference being that there could be more states than just enabled || disabled

SCKelemen:
automerge

erezrokah:
> automerge

@SCKelemen I can make this work if you'd like, but the way to do it at the moment is to add it as a label on the PR.
Having it as a PR comment can be cool 🤖

SCKelemen:
> > automerge
> 
> @SCKelemen I can make this work if you'd like, but the way to do it at the moment is to add it as a label on the PR. Having it as a PR comment can be cool 🤖

The label is fine. I guess that's easier for access control.  
Thanks!

PR #2726, opened 2022-10-12
chore(tests): Use other faker less

disq:
Around 300 files. Some got removed, some (k8s) didn't.

Continuation of https://github.com/cloudquery/cloudquery/pull/2622

To remove more old faker references, we need to implement the equivalent of `SetIgnoreInterface(bool)` ~and `SkipFields([]string)`~ as `faker.Option` in plugin-sdk/faker.


erezrokah (approved):

disq:
More removals after https://github.com/cloudquery/plugin-sdk/pull/287 or equivalent is merged and released.

erezrokah (approved):

PR #2749, opened 2022-10-12
fix: Improperly Configured Service Clients

bbernays:
#### Summary

Fixes panics in resources

### DIFF
--- a/plugins/source/aws/client/client.go
+++ b/plugins/source/aws/client/client.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/backup"
 	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
 	"github.com/aws/aws-sdk-go-v2/service/cloudfront"
+	"github.com/aws/aws-sdk-go-v2/service/cloudhsmv2"
 	"github.com/aws/aws-sdk-go-v2/service/cloudtrail"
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

erezrokah:
This should be caught by adding mock tests to these resources (I think)

bbernays:
> This should be caught by adding mock tests to these resources (I think)

I would have expected the same thing, but I don't think it is as all of these resources have mock tests:
https://github.com/cloudquery/cloudquery/blob/b448bc6cebdf7ddbcdcd9fedb6ebb5fa427821f2/plugins/source/aws/resources/services/cloudhsmv2/backups_mock_test.go

erezrokah:
> I would have expected the same thing, but I don't think it is as all of these resources have mock tests:
> https://github.com/cloudquery/cloudquery/blob/b448bc6cebdf7ddbcdcd9fedb6ebb5fa427821f2/plugins/source/aws/resources/services/cloudhsmv2/backups_mock_test.go

Yeah I was wrong. Mock tests won't catch this as we create mock services in the mock tests

yevgenypats:
Seems like a good place to either add test or maybe let the codegen output those same as in gcp plugin

bbernays:
> Seems like a good place to either add test or maybe let the codegen output those same as in gcp plugin

@yevgenypats - Created an issue to track this

Issue #2642, opened 2022-10-11
Bug(release): Some plugins don't have their version embedded correctly


erezrokah:
### Describe the Bug

In order for `<plugin> --version` command to work we need to embed the correct version via Go Releaser.
Go Releaser expects it to be under `resources/plugin.Version`:
https://github.com/cloudquery/cloudquery/blob/7b2c4aac79ca356efb41244c9ee3d8306c600dd2/plugins/.goreleaser.yaml#L12
That requires plugins to follow a specific directory and package convention.
Some don't like `k8s,okta,terraform`. We should fix it

### Expected Behavior

`<plugin> --version` should work for all plugins

erezrokah:
This is completed

yevgenypats:
Is it completed to destination plugins as well?

erezrokah:
Ah no, good catch, I'll re-open and fix thos

erezrokah:
I somehow thought we did this for destinations plugins when we introduced them 🙃

erezrokah:
Follow up in https://github.com/cloudquery/cloudquery/pull/2946 and https://github.com/cloudquery/cloudquery/pull/2947

PR #2761, opened 2022-10-13
fix(terraform): Structure plugin so version is embedded by Go Releaser

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/2642. Same as https://github.com/cloudquery/cloudquery/pull/2759 for Terraform


.gitignore
```diff
@@ -11,7 +11,6 @@ dist/
 vendor/
 bin/
 .release-env
-*.tfstate*
```
erezrokah, line 14: Removed this otherwise `plugins/source/terraform/resources/services/testdata/terraform.tfstate` is not committed.
I could make it more specific, but I don't think we need it since we removed the integration tests

yevgenypats (approved):

PR #2765, opened 2022-10-13
chore(website): Fix recipes versions

erezrokah:
#### Summary

In order for our `find/replace` to work the versions need to be quoted with `"`:
https://github.com/cloudquery/cloudquery/blob/e320875c26d221a6fe757d8b9df5a4f93829c225/.github/workflows/release_plugin.yml#L74


erezrokah:
We can also fix the regex to handle it, but this also makes sure we're consistent with the formatting of our YAML examples

yevgenypats (approved):
looks good. hopefully we will get to it one day via native Next/Nextra support.

PR #2788, opened 2022-10-13
chore(tests): Try parallel testing

disq:
Try parallel testing on all tests using mock test helpers.

### DIFF
--- a/plugins/source/aws/client/testing.go
+++ b/plugins/source/aws/client/testing.go
@@ -21,6 +21,7 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
 
 	table.IgnoreInTests = false
 	t.Helper()
+	t.Parallel()
 	ctrl := gomock.NewController(t)
 	l := zerolog.New(zerolog.NewTestWriter(t)).Output(
 		zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.StampMicro},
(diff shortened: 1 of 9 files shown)


erezrokah (approved):

erezrokah:
For AWS tests changed from `2m 6s` to `1m 48s`:
https://github.com/cloudquery/cloudquery/actions/runs/3243124043/jobs/5317379722
https://github.com/cloudquery/cloudquery/actions/runs/3243092646/jobs/5317306478

But that's only one run

yevgenypats:
> For AWS tests changed from `2m 6s` to `1m 48s`: https://github.com/cloudquery/cloudquery/actions/runs/3243124043/jobs/5317379722 https://github.com/cloudquery/cloudquery/actions/runs/3243092646/jobs/5317306478
> 
> But that's only one run

@erezrokah  maybe increasing `GOMAXPROCS ` to 4 or 8 will help ?

yevgenypats (changes requested):
I think this should into the SDK `TestSourcePluginSync`

disq:
> TestSourcePluginSync

but that method doesn't take an option, to override (disable) parallel testing?

yevgenypats:
> > TestSourcePluginSync
> 
> but that method doesn't take an option, to override (disable) parallel testing?

I think then it would be great case to introduce option and make it default enabled :)

erezrokah:
>@erezrokah maybe increasing GOMAXPROCS to 4 or 8 will help ?

Maybe GitHub machines are not strong enough to take advantage. Regardless this is probably good for local dev machiens

Issue #2644, opened 2022-10-11
--data-dir parameter ignored


mrichnu:
### Describe the Bug
When running `cloudquery sync` with the `--data-dir` flag to change the location where the plugin files are downloaded, cloudquery 1.0.7 is ignoring the value passed in and always saving plugins to the `./.cq` directory.

### Expected Behavior
cloudquery should save plugins in the directory specified by the `--data-dir` parameter.

### CloudQuery Version
cloudquery version 1.0.7

### Steps to Reproduce

Run cloudquery e.g.
```
cloudquery sync source.yml dest.yml \
    --data-dir /mnt/data \
    --log-console --color off --no-log-file --log-level warn \
    2>&1
```

### Additional Context

Note the `color` flag is being ignored as well; I always get colored output.

bbernays:
@mrichnu Thank you for this bug report! 


these 2 flags are not being properly used: https://github.com/cloudquery/cloudquery/blob/7b2c4aac79ca356efb41244c9ee3d8306c600dd2/cli/cmd/root.go#L109-L111

Instead of `PersistentFlags().String()` I think it should be some type of `Var` like all of the other flags

PR #2796, opened 2022-10-13
feat: Deprecate --data-dir in favor of --cq-dir

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery/issues/2644. Fixes the `--data-dir` flag.
Not sure if there's a better way to access the flag. Also we could considering making it a `sync` command flag instead of the root command.


cli/cmd/sync.go
```diff
@@ -37,6 +37,11 @@ func NewCmdSync() *cobra.Command {
 }
 
 func sync(cmd *cobra.Command, args []string) error {
+	dataDir, err := cmd.Flags().GetString("data-dir")
+	if err != nil {
+		return err
```
erezrokah, line 42: The error received is already quite descriptive

yevgenypats (changes requested):
maybe we should just name it `plugin` dir and make the default `./.cq/plugins` ?
I don't think we are going to save anything else there in the near future but if we will we can think about what is the right flag.

erezrokah:
> maybe we should just name it `plugin` dir and make the default `./.cq/plugins` ?
> I don't think we are going to save anything else there in the near future but if we will we can think about what is the right flag.

I'm not sure actually as I'd rather keep this flag generic without implementation details (e.g. plugins), as users shouldn't really care what we put in that directory.

Also this would require an SDK change https://github.com/cloudquery/plugin-sdk/blob/dc7fff126fcc4849f5f63725868cd7351391b714/clients/source.go#L97.
We could change the SDK API to `WithSourcePluginsDirectory` from `WithSourceDirectory`, but that seems a bit too much.

yevgenypats:
> > maybe we should just name it `plugin` dir and make the default `./.cq/plugins` ?
> > I don't think we are going to save anything else there in the near future but if we will we can think about what is the right flag.
> 
> I'm not sure actually as I'd rather keep this flag generic without implementation details (e.g. plugins), as users shouldn't really care what we put in that directory.
> 
> Also this would require an SDK change https://github.com/cloudquery/plugin-sdk/blob/dc7fff126fcc4849f5f63725868cd7351391b714/clients/source.go#L97. We could change the SDK API to `WithSourcePluginsDirectory` from `WithSourceDirectory`, but that seems a bit too much.

can we maybe change the name to something more descriptive (imo) ? Like `--cq-dir` i.e `this is where cloudquery store cloudquery related files like downloaded plugins` ... ?

erezrokah:
Yeah, `cq-dir` is better

erezrokah:
@yevgenypats followed up with https://github.com/cloudquery/cloudquery/pull/2796/commits/4b254d4587c7ab0ff21d64c88211ac616cdb2861.

I dropped the automatic binding for Viper otherwise, one would need to set `CQ_CQ_DIR` for that flag

cli/cmd/root.go
```diff
@@ -154,18 +153,30 @@ func NewCmdRoot() *cobra.Command {
...
+	initViper(cmd)
 	cmd.SetHelpCommand(&cobra.Command{Hidden: true})
 	cmd.AddCommand(NewCmdSync(), newCmdDoc())
 	cmd.CompletionOptions.HiddenDefaultCmd = true
 	cmd.DisableAutoGenTag = true
 	return cmd
 }
 
-func initViper() {
-	viper.AutomaticEnv()
```
erezrokah, line 166: Dropped this in favor of manual configuration. I think configuring it manually makes it much easier to understand which Env variable to use (and also allows us to customize the name)

yevgenypats: Great. See I did the same here - https://github.com/cloudquery/cloudquery/pull/2619

yevgenypats: EDIT: I it's better to drop viper completely it's a horrible library. Also, I don't think environment variable is needed for `--cq-dir` if someone will ask for it we can always add it.

Currently only added for telemetry as it's common to put it in the `.bashrc`

yevgenypats (changes requested):
Looks good. let's drop `viper` completely as in https://github.com/cloudquery/cloudquery/pull/2619

erezrokah:
> Looks good. let's drop `viper` completely as in #2619

Done in a separate PR https://github.com/cloudquery/cloudquery/pull/2904

cli/cmd/sync.go
```diff
@@ -36,7 +36,27 @@ func NewCmdSync() *cobra.Command {
 	return cmd
 }
 
+func getCQDir(cmd *cobra.Command) (string, error) {
+	// backwards compatibility
+	cqDir, err := cmd.Flags().GetString("data-dir") // old flag
```
yevgenypats, line 41: We are taking our backward compatability too far. This never worked. no need to make it work now as we deprecating it anyway - otherwise we just introduce more code to maintain for no reason.

yevgenypats: Once this is remove I think no need for this function anymore

yevgenypats (changes requested):
Looks good. One comment: Let's not introduce support for `--data-dir` given it never worked in v1 and we are deprecating it.

cli/cmd/sync.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/2796

yevgenypats (approved):

PR #2812, opened 2022-10-13
fix(cli): Deprecate color flag

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/2644.
Since the default is to have log color, it makes sense to have a `--no-color` flag instead of `--color`. I added one in a backwards compatible way. Also I think `auto` never really worked.


website/pages/docs/reference/cli/cloudquery.md
```diff
@@ -17,13 +17,13 @@ Find more information at:
 ### Options
 
 ```
-      --color string           Enable colorized output (on, off, auto) (default "auto")
```
erezrokah, line 20: Since we marked the flag as deprecated it's removed from the help, but still works

yevgenypats:
Giving it a bit more thought, I think we need to skip color all together and only do without colors.

ELT workloads produces tons of logs so no one is really going through them through the terminal.
By disabling color by default I think we will insure the correct behaviour out of the box and prevent confusion and more flags. 

I think our CLI is more of very slim worker rather then a UI or something like this so I think it makes sense to avoid colors completely, which causes issues in production environments. wdyt ?

yevgenypats (changes requested):
Giving it a bit more thought, I think we need to skip color all together and only do without colors.

ELT workloads produces tons of logs so no one is really going through them through the terminal.
By disabling color by default I think we will insure the correct behaviour out of the box and prevent confusion and more flags.

I think our CLI is more of very slim worker rather then a UI or something like this so I think it makes sense to avoid colors completely, which causes issues in production environments. wdyt ?

erezrokah:
> Giving it a bit more thought, I think we need to skip color all together and only do without colors.
> 
> ELT workloads produces tons of logs so no one is really going through them through the terminal. By disabling color by default I think we will insure the correct behaviour out of the box and prevent confusion and more flags.
> 
> I think our CLI is more of very slim worker rather then a UI or something like this so I think it makes sense to avoid colors completely, which causes issues in production environments. wdyt ?

Good point, I kept the deprecation with a message that console logs are always colorless

yevgenypats (approved):

PR #2826, opened 2022-10-13
fix: Improve IAM Policy sync performance

bbernays:
#### Summary

This PR adds in server side filtering so that the API doesn't try to get a lot of data that we never use.

In local tests I saw reduction in time from 10 seconds down to 2 seconds


yevgenypats (approved):

disq (approved):

jsonpr:
Another thought - we could use the ListPolicies AWS API call.  That would map closer to how we structure roles_fetch.go by using ListRoles and then we wouldn't need a filter and would avoid grabbing the Users, Groups, and Roles in the Authorization Details.  ListPolicies will reutrn both customer-defined managed policies and all AWS managed policies.

If there are reasons to continue using GetAccountAuthorizationDetails, I could see keeping this filter change, given how this belongs to the function fetchIamPolicies.  With the GetAccountAuthorizationDetails call, this will return Users, Groups, Roles, Local Managed Policies and AWS Managed Policies without any filter.  By adding the Local Managed Policy and AWS Managed Policies filters, this will only return policies which makes sense for the function.

jsonpr (changes requested):
Another thought - we could use the ListPolicies AWS API call. That would map closer to how we structure roles_fetch.go by using ListRoles and then we wouldn't need a filter and would avoid grabbing the Users, Groups, and Roles in the Authorization Details. ListPolicies will return both customer-defined managed policies and all AWS managed policies.

If there are reasons to continue using GetAccountAuthorizationDetails, I could see keeping this filter change, given how this belongs to the function fetchIamPolicies. With the GetAccountAuthorizationDetails call, this will return Users, Groups, Roles, Local Managed Policies and AWS Managed Policies without any filter. By adding the Local Managed Policy and AWS Managed Policies filters, this will only return policies which makes sense for the function.

No changes requested if this discussion can be resolved!

bbernays:
I agree that if we were to use `ListPolicies` it would better be aligned with our other resources... If we were to switch it would require a bunch more work:
1. Update client
2. Update mock tests
3. Ensure adequate performance
4. Release new major version (as the permissions required to fetch will change)

My suggestion is we merge this PR (as this seems to be a no downside performance improvement), and open an issue to discuss if the above work is worth it.

jsonpr:
Thanks for providing additional context.  I'm ok with this as a tactical fix and then revisiting a strategic solution of using ListPolicies. 

Regarding permissions, this is a complicated discussion as `GetAccountAuthorizationReport` is a Get permission compared to a `List` permission.  For example, the Amazon managed `ViewOnlyAccess` policy includes `iam:List*`, but not `iam:GetAccountAuthorizationReport`.  The good news is that the Amazon managed `ReadOnlyAccess` policy we typically use includes both `iam:List*` and `iam:Get*`, so the potential impact if any would be on CQ users who have done bespoke/custom IAM.  Long term, I prefer to start with `List` over `Get` permissions, but agree that customer impact should be considered before we make a larger change.

Agreed with testing and performance as good sanity checks.

hermanschaaf (approved):

PR #2837, opened 2022-10-14
fix!: More tag cleanup

### LINKED ISSUE #2836: Standardize Tags
### Describe the Bug

All resources should return tags in an object rather than a list

### Expected Behavior

The following tables all need to have their tag column adjusted:
```
"aws_ec2_internet_gateways"
"aws_ec2_network_acls" 
"aws_ec2_route_tables"
"aws_ec2_vpc_endpoint_services"
```

disq:
Closes #2836


yevgenypats (approved):

disq:
Technically this is not a breaking change schema-wise because it's all smushed into a `jsonb` column. But we change from an array of objects to a single map.

bbernays (approved):

PR #2845, opened 2022-10-14
fix!: Rollback: Remove the 'functions.Timeout' column

shimonp21:
The 'TypeInterval' type is broken right now.
Because of the new batching functionality, this failure completely breaks the GCP plugin (any failure in an INSERT causes all other INSERTS to also fail due to the batch).


cqgaurav:
For the uninitiated: why did the type break ?

yevgenypats (approved):

PR #2887, opened 2022-10-14
fix: Improve PG Write Speeds

bbernays:


yevgenypats (approved):
Amazing find!!! 🚀 

There bunch of unrelated fixes around the following but they will come in other PRs:

missing batches on delete filter, missing batches on migration, option to skip migrations, pool of go-routines sending batches, maybe changing the interface so it can work with unprepared statements and some more adventures with pgx and postgresql.

PR #2904, opened 2022-10-16
fix: Drop viper, make --no-telemetry work

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/2796#pullrequestreview-1143281167


cli/cmd/root.go
```diff
@@ -135,12 +137,7 @@ func NewCmdRoot() *cobra.Command {
...
 
 	// Telemtry (analytics) flags
-	cmd.PersistentFlags().Bool("no-telemetry", false, "disable telemetry collection")
-	// we dont need viper support for most flags as all can be used via command line for now (we can add in the future if really necessary)
-	// the only exception is the telemetry as people might want to put in a bash starter script
-	if err := viper.BindPFlag("no-telemetry", cmd.PersistentFlags().Lookup("no-telemetry")); err != nil {
-		panic(err)
-	}
+	cmd.PersistentFlags().BoolVar(&noTelemetry, "no-telemetry", false, "disable telemetry collection (env: CQ_NO_TELEMETRY)")
 	cmd.PersistentFlags().Bool("telemetry-inspect", false, "enable telemetry inspection")
```
yevgenypats, line 141: Unrelated to this PR but let's drop `telemetry-inspect` and `telemtry-debug` they are just unused flags which bloat our help and make it hard to understand what are the "real" options

yevgenypats (changes requested):
Looks good. one nit.

cli/cmd/root.go
erezrokah: Sure, done in https://github.com/cloudquery/cloudquery/pull/2904/commits/1470b1ce52dfb0d9234ff66b021e1c6508ac8480

cli/cmd/root.go
```diff
@@ -88,7 +87,10 @@ func NewCmdRoot() *cobra.Command {
 
 			mw := io.MultiWriter(writers...)
 			log.Logger = zerolog.New(mw).Level(zerologLevel).With().Str("module", "cli").Timestamp().Logger()
-			if sentryDsn != "" && Version != "development" {
+
+			noTelemetry = getEnvOrDefault("CQ_NO_TELEMETRY", "false") == "true" || noTelemetry
+			sentryEnabled := sentryDsn != "" && Version != "development" && !noTelemetry
```
yevgenypats, line 92: I think this is incorrect now.

we have two different things which are unrelated:
telemetry - we send anonymised usage data such as source, destination plugins and number of resources used.
sentry (error-reporting) - we send panics that inccur in our official plugins and/or cli

those two should be either seperate and/or as suggested before have one flag `telemetry` with `none,errors,all` both should work but if someone disable analytics they shouldn't automatically disable error reporting (but they still should have the option). Same as in vscode and any other software.

yevgenypats (changes requested):
Sorry maybe missed in previous review but the flags behaviour is now incorrect as no_telemtry disables both errors and analytics.

cli/cmd/root.go
erezrokah: Ah, ok so I thought we don't have telemetry and https://github.com/cloudquery/cloudquery/pull/2619 introduces it.
Maybe it's worth just closing this PR and doing everything in https://github.com/cloudquery/cloudquery/pull/2619?

Not sure we need the resolution of `none,errors,all`, but best to leave for https://github.com/cloudquery/cloudquery/pull/2619

yevgenypats: Sounds good. let's close this one.

yevgenypats:
Closing per comment here - https://github.com/cloudquery/cloudquery/pull/2904#discussion_r996454506

Issue #2757, opened 2022-10-13
chore(ci-release): Improve release workflow time


erezrokah:
Releasing some plugins (like AWS) can take quite a bit of time due to long build times.
For example to get the AWS plugin released we need to wait about ~20 minutes for the `validate-release` workflow that runs on the release PR:
https://github.com/cloudquery/cloudquery/actions/runs/3236336140/jobs/5302004441
![image](https://user-images.githubusercontent.com/26760571/195528746-b22f0e7f-0513-4e5a-9f2b-8795b422aa17.png)

Then we need to wait for the actually release workflow that runs on the tag push event:
https://github.com/cloudquery/cloudquery/actions/runs/3238001704/jobs/5305722318
![image](https://user-images.githubusercontent.com/26760571/195528668-6caaabf2-4bb0-462b-9851-9ba8ab67d90d.png)

### Possible solutions

1. ~~Adding go caching to these workflows. I didn't add it yet since I kind had the thought it's safer to do a clean slate when releasing to avoid any caching issues. I think we could try adding caching and use a cache that's specific for the releases to avoid mixing the cache with PRs~~ Looks like we already have caching for those but maybe it's not working correctly

2. Use stronger machines, per https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners. We've registered to the beta and waiting for a response

erezrokah:
This is done, we both have caching enabled and stronger machines

PR #2909, opened 2022-10-16
chore(release): Improve release workflow caching

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/2757.

At the moment all our workflows share the same cache key logic that is a composition of the OS, Go version, and a hash of the `go.sum` file of the specific component the workflow runs against.
This means that our release workflows (validate release and Go releaser) share the same cache as our testing workflows.
Which means a pull request can override the cache of the release workflows.

I think we would benefit from having a dedicated cache just for releases (which this PR does).

**TLDR**: This PR creates a dedicated cache for the `validate release` workflow and `release plugin/cli` workflows (a cache per component).

You can see an example here:
First run of AWS validate release is ~20 minutes:
https://github.com/cloudquery/cloudquery/actions/runs/3259211192/jobs/5351872391
image
image
Second run of AWS validate release is ~2 minutes (10x faster):
https://github.com/cloudquery/cloudquery/actions/runs/3259211192/jobs/5352039966
image

Still need to test this once merged to see if there's an improvement


yevgenypats (approved):

erezrokah:
PLEASE DON'T MERGE, I'll MERGE IT ONCE I'M AVAILABLE TO FIX ISSUES IF THIS BREAKS ANYTHING WITH RELEASES

PR #2949, opened 2022-10-18
chore(release): Revert caching changes

erezrokah:
#### Summary

This reverts https://github.com/cloudquery/cloudquery/pull/2909 and https://github.com/cloudquery/cloudquery/pull/2925, as caching in [GitHub Actions doesn't work as I expected](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache). 
TLDR: PRs can use the cache of the base branch, or their own branch, but the base branch doesn't share the cache of a PR branch.
This means we benefit from a cache only after merging a PR and building it on the `main` branch


erezrokah:
We should consider adding a workflow on `main` to prime the cache if we want to speed up releases/PRs

hermanschaaf (approved):

PR #2950, opened 2022-10-18
chore(pg): Refactor writing a bit

candiduslynx:


yevgenypats (changes requested):
Im in the middle of a refactor of destinations. so destinations mostly on freeze right now as otherwise I wont be able to solve conflicts.

plugins/destination/postgresql/client/write.go
```diff
@@ -31,52 +32,37 @@ func (c *Client) Write(ctx context.Context, table string, data map[string]interf
 	return nil
 }
 
-func insert(table string, data map[string]interface{}) (string, []interface{}) {
-	var sb strings.Builder
+func insert(table string, data map[string]interface{}) (string, []any) {
```
bbernays, line 35: I wouldn't switch `[]interface{}` to `[]any` unless we are going to make the change everywhere

bbernays (commented):
Overall looks good one nit

plugins/destination/postgresql/client/write.go
candiduslynx: for now, `any` is just an alias of `interface{}`, so in this case I propose to gradually convert to `any` (as it's a bit cleaner)

yevgenypats:
Closing as this will be changed completely by https://github.com/cloudquery/plugin-sdk/pull/298 and https://github.com/cloudquery/cloudquery/pull/3176

PR #2974, opened 2022-10-18
fix: A few fixes for analytics

hermanschaaf:
A few fixes for the upcoming anonymised analytics. Note that the base branch of this PR is `feat/analytics`.


yevgenypats (approved):

cli/cmd/root.go
```diff
@@ -26,7 +26,7 @@ Find more information at:
 func NewCmdRoot() *cobra.Command {
 	logLevel := enum.NewEnum([]string{"trace", "debug", "info", "warn", "error"}, "info")
 	logFormat := enum.NewEnum([]string{"text", "json"}, "text")
-	telemetryLevel := enum.NewEnum([]string{"trace", "debug", "info", "warn", "error"}, "info")
+	telemetryLevel := enum.NewEnum([]string{"none", "errors", "all"}, "all")
```
hermanschaaf, line 29: @yevgenypats I think we should reconsider the options here: some people may be okay with sending some anonymous usage data, but not okay with sending errors (since those could in theory contain some identifying information by accident).

Maybe `none`, `errors`, `stats`, `all`?

yevgenypats: sounds good.

PR #2983, opened 2022-10-18
doc: Add troubleshooting section about migration failures

shimonp21:


website/pages/docs/troubleshooting.md
```diff
@@ -22,3 +22,21 @@ There are a couple of ways to get help for any CloudQuery-related issues or ques
 ### Verbose Logging
 
 Usually the first step that will be needed to debug/resolve an issue is to run `cloudquery` with `--log-level debug` to enable verbose logging.
+
+### Error: "failed to migrate source"...
+
+If you see an error such as `failed to migrate source`, it means that, while upgrading a plugin, the migration of the SQL schema failed.
+CloudQuery makes a best-effort to automatically and transparently manage the schema of the plugins, but
```
hermanschaaf, line 29:
```suggestion
CloudQuery makes a best-effort attempt to automatically and transparently manage the schemas of plugins, but 
```

website/pages/docs/troubleshooting.md
```diff
@@ -22,3 +22,21 @@ There are a couple of ways to get help for any CloudQuery-related issues or ques
...
 
 Usually the first step that will be needed to debug/resolve an issue is to run `cloudquery` with `--log-level debug` to enable verbose logging.
+
+### Error: "failed to migrate source"...
+
+If you see an error such as `failed to migrate source`, it means that, while upgrading a plugin, the migration of the SQL schema failed.
+CloudQuery makes a best-effort to automatically and transparently manage the schema of the plugins, but 
+this can sometimes fail during version upgrades.
+
+The solution is to drop-and recreate the database or schema. (or all the plugins tables, such as `aws_*`).
```
hermanschaaf, line 32:
```suggestion
The easiest solution is to drop and recreate the database or schema (or all the plugins tables, such as `aws_*`).
```

website/pages/docs/troubleshooting.md
```diff
@@ -22,3 +22,21 @@ There are a couple of ways to get help for any CloudQuery-related issues or ques
...
+Some queries that may help you (handle with care!):
+
+```sql
+DROP SCHEMA public CASCADE;
+CREATE SCHEMA public;
+```
+
+```sql
+\dt aws_*
+```
```
hermanschaaf, line 42: We can maybe also mention here that there are other, less destructive, solutions. However, this will usually involve inspecting the schema changes (as described in the changelog) and manually migrating them.

hermanschaaf (approved):
Nice! Left a few suggestions

website/pages/docs/troubleshooting.md
shimonp21: Not sure we want to have that here - since if the user gets here it means that our auto-migration already failed (but only on some of the tables probably)- and something is broken in a non-trivial way..

hermanschaaf: Yeah, for sure, it will be non-trivial, and we'll try and make it as rare as possible. But perhaps it's a worthwhile option for some users who have a lot of triggers set up around their database.

website/pages/docs/troubleshooting.md
```diff
@@ -22,3 +22,21 @@ There are a couple of ways to get help for any CloudQuery-related issues or ques
...
```
bbernays, line 42: Is this SQL or `psql`?

website/pages/docs/troubleshooting.md
```diff
@@ -22,3 +22,21 @@ There are a couple of ways to get help for any CloudQuery-related issues or ques
...
+If you see an error such as `failed to migrate source`, it means that, while upgrading a plugin, the migration of the SQL schema failed.
+CloudQuery makes a best-effort attempt to automatically and transparently manage the schemas of plugins, but 
+this can sometimes fail during version upgrades.
+
+The easiest solution is to drop and recreate the database or schema (or all the plugins tables, such as `aws_*`).
+Some queries that may help you (handle with care!):
+
+```sql
+DROP SCHEMA public CASCADE;
+CREATE SCHEMA public;
```
bbernays, line 37: If you recreate the public schema don't you also need to regrant usage permissions?

PR #2990, opened 2022-10-18
Fix: Minor adjustments to analytics

hermanschaaf:
Quick follow-up to https://github.com/cloudquery/cloudquery/pull/2619

- Destination spec should be sending `Path`, not `Name`
- Add backwards-compatible support for `CQ_NO_TELEMETRY` in case someone is still relying on that from v0. We don't want to surprise anyone by sending telemetry even though they have that variable set, but `CQ_TELEMETRY_LEVEL` should be preferred from now on.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -36,7 +36,14 @@ func NewCmdRoot() *cobra.Command {
 	logFileName := "cloudquery.log"
 	sentryDsn := sentryDsnDefault
 
-	err := telemetryLevel.Set(getEnvOrDefault("CQ_TELEMETRY_LEVEL", telemetryLevel.Value))
+	// support legacy telemetry environment variable,
+	// but the newer CQ_TELEMETRY_LEVEL environment variable takes precedence
+	defaultTelemetryValue := telemetryLevel.Value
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):

cli/cmd/root.go
```diff
@@ -36,7 +36,14 @@ func NewCmdRoot() *cobra.Command {
 	logFileName := "cloudquery.log"
 	sentryDsn := sentryDsnDefault
 
-	err := telemetryLevel.Set(getEnvOrDefault("CQ_TELEMETRY_LEVEL", telemetryLevel.Value))
+	// support legacy telemetry environment variable,
+	// but the newer CQ_TELEMETRY_LEVEL environment variable takes precedence
+	defaultTelemetryValue := telemetryLevel.Value
+	legacyTelemetry := os.Getenv("CQ_NO_TELEMETRY")
+	if legacyTelemetry != "" {
+		defaultTelemetryValue = "none"
```
erezrokah, line 44: [Non blocking] Maybe add a warning message here to switch to the new env?

erezrokah (approved):

cli/cmd/root.go
hermanschaaf: Good idea, added 👍

PR #2991, opened 2022-10-18
feat: Add folder_ids to the GCP spec

shimonp21:


plugins/source/gcp/client/client.go
```diff
@@ -46,6 +50,12 @@ func (c *Client) Logger() *zerolog.Logger {
 }
 
 func New(ctx context.Context, logger zerolog.Logger, s specs.Source) (schema.ClientMeta, error) {
+	// We use a timeout to make sure that this method doesn't get stuck indefinitely.
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
yevgenypats, line 54: I think a minute is quite low. For example we saw users with more than 80k projects. 
Also I think there is no need for calling `cancel`

plugins/source/gcp/client/spec.go
```diff
@@ -4,4 +4,6 @@ package client
 type Spec struct {
 	ProjectIDs            []string `json:"project_ids"`
 	ServiceAccountKeyJSON string   `json:"service_account_key_json"`
+	FolderIDs             []string `json:"folder_ids"`
+	FolderRecursionDepth  int      `json:"folder_recursion_depth"`
```
yevgenypats, line 8: let's setup default for `FolderRecursionDepth` via `SetDefault` method

yevgenypats (changes requested):
Looks good. few comments

plugins/source/gcp/client/client.go
shimonp21: Removed. Will be added at the SDK level if required later.

plugins/source/gcp/client/spec.go
shimonp21: Changed to *int.

yevgenypats (approved):

plugins/source/gcp/client/client.go
```diff
@@ -60,6 +68,13 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source) (schema.Cli
 
 	serviceAccountKeyJSON := []byte(gcpSpec.ServiceAccountKeyJSON)
 
+	var folderRecursionDepth int
+	if gcpSpec.FolderRecursionDepth == nil {
```
yevgenypats, line 72: This should go as a method function though. i.e `gcpSpec.SetDefaults()`

Issue #1418, opened 2021-08-14
Support IdentityStore


yevgenypats:
IdentityStore is a key service in AWS at scale management. Add support for retrieving users and groups.

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/identitystore

jsonpr:
AWS released APIs to AWS Identity Center (formerly known as AWS SSO) that will permit us to pull the proper data from AWS Identity Center.

https://aws.amazon.com/blogs/security/announcing-new-aws-iam-identity-center-apis-to-manage-users-and-groups-at-scale/

jhudson10x:
I started working on this today. I anticipate having a PR open by the end of the week.

yevgenypats:
@jhudson10x Nice! Let us know if you have any questions/need help.

jhudson10x:
I've got the bulk of the code written but I'm running into an issue. The plugin compiles but the plugin sdk is not invoking the new table resolver. I can see that the table is registered because the table name validation step in the plugin is successful. I don't know why https://github.com/cloudquery/plugin-sdk/blob/051e24710b64672b4fa4eda1261e2558859cbc75/schema/table.go#L160 is not being called.

hermanschaaf:
@jhudson10x Were you able to figure it out? Your code looks good and I ran it locally to confirm that the resolver is now being called (I see the `*******FETCH********` and `*******RESPONSE********` print lines). One thing to note maybe is just that when you're developing on a plugin, the console output will go to the plugin process's terminal, not the cloudquery.log file.

Let us know if you need any further help!

jhudson10x:
Thank you @hermanschaaf. Yes, I was able to get past the issue. `cq-ben` in discord (not sure what their handle is here) pointed out that there is no `ssoadmin` key in `plugins/source/aws/client/data/partition_service_region.json`. I changed the region multiplexer to use the `identitystore` key and am unblocked now.

I have two more (non-blocking) questions now:

1. The service has to be configured in exactly one region. Right now it queries each of the 13 regions listed in identitystore, 12 of which always fail with a 400. Is there an elegant way to deal with this? I am not aware of one.
2. The service can't be configured in every AWS region, but it can be configured in more than just the 13 regions defined by the `identitystore` key. Should I add a new `ssoadmin` key with the correct list of regions?

bbernays:
[...]

PR #3005, opened 2022-10-18
feat(aws): Support identitystore and ssoadmin

jhudson10x:
This PR is ready for review for merge.There remains ssoadmin resources to be added. I can add them as part of this PR or I can open a new PR for those resources. Please advise.

#### Summary

Add support for the IAM Identity Center (Successor to AWS Single Sign-On) a.k.a `ssoadmin` and `identitystore` APIs


jhudson10x:
refs #1418

plugins/source/aws/resources/services/ssoadmin/instances_fetch.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ssoadmin"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchSsoadminInstances(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	fmt.Print("*******FETCH********\n")
```
disq, line 14: Forgotten print stmt, also two more below on line 20-21

plugins/source/aws/resources/services/ssoadmin/instances_fetch.go
```diff
@@ -0,0 +1,35 @@
...
+	config := ssoadmin.ListInstancesInput{}
+	fmt.Print(config)
+	for {
+		response, err := svc.ListInstances(ctx, &config)
+		fmt.Print("*******RESPONSE********\n")
+		fmt.Printf("%#v\n", response)
+		if err != nil {
+			return err
+		}
+		for _, i := range response.Instances {
```
disq, line 25: Use `res <- response.Instances`

plugins/source/aws/codegen/recipes/ssoadmin.go
```diff
@@ -0,0 +1,22 @@
...
+		{
+			SubService: "instances",
+			Struct:     &types.InstanceMetadata{},
+		},
+	}
+
+	// set default values
+	for _, r := range resources {
+		r.Service = "ssoadmin"
+		r.Multiplex = `client.ServiceAccountRegionMultiplexer("identitystore")`
```
erezrokah, line 19:
```suggestion
		r.Multiplex = `client.ServiceAccountRegionMultiplexer("sso")`
```

Should this be `sso`? I believe this should match https://github.com/cloudquery/cloudquery/blob/18c7a3224ba898451c1a29395acabc2e4395a40b/plugins/source/aws/client/data/partition_service_region.json#L5219

plugins/source/aws/resources/services/identitystore/users_fetch.go
```diff
@@ -0,0 +1,5 @@
+package identitystore
+
+func fetchIdentitystoreGroups() error {
```
erezrokah, line 3: Assuming this needs to be implemented?

erezrokah (changes requested):
Thanks @jhudson10x, added a few comments too?

What do you think about breaking this into multiple PRs per resource? Then it will be easier to merge the completed ones

plugins/source/aws/resources/services/identitystore/users_fetch.go
jhudson10x: Yes, not done yet. There is a question in my mind about best way to use the value returned by `ssoadmin` service (identity store id) in the `identitystore` service. I am thinking I will need to call that API twice, once in each service. Let me know if you have a better idea.

plugins/source/aws/codegen/recipes/ssoadmin.go
jhudson10x: No, actually, we need to make a new entry in that file for the `ssoadmin` service. There are 17 regions that it supports, though it can only be configured in one region at a time. @bbernays confirmed to me on Discord and in the related issue ticket that this will be ok.

jhudson10x:
> Thanks @jhudson10x, added a few comments too?
> 
> What do you think about breaking this into multiple PRs per resource? Then it will be easier to merge the completed ones

I am ok with breaking these into multiple PRs. I apologize for the delay in wrapping this one up. Last week was very busy with an onsite meeting and did not get a chance to do much.

plugins/source/aws/resources/services/identitystore/users_fetch.go
jhudson10x: I made it work by declaring the group memberships table as a Relation in the codegen recipe.

plugins/source/aws/resources/services/ssoadmin/instances_fetch.go
jhudson10x: I fixed this, thank you.

plugins/source/aws/resources/services/ssoadmin/instances_fetch.go
jhudson10x: Fixed, thanks. Wasn't forgotten at the time I committed it, was demonstrating a problem (note the WIP title at the time)

jhudson10x:
> > Thanks @jhudson10x, added a few comments too?
> > What do you think about breaking this into multiple PRs per resource? Then it will be easier to merge the completed ones
> 
> I am ok with breaking these into multiple PRs. I apologize for the delay in wrapping this one up. Last week was very busy with an onsite meeting and did not get a chance to do much.

Just to be clear in case it wasn't from context - the identitystore service depends on the ssoadmin instance resource. This PR is ready for review for merge. I'm going to be finishing the remaining ssoadmin resources today, but if you want to go ahead and wrap this one up I can open a new PR for those resources.

plugins/source/aws/codegen/recipes/ssoadmin.go
erezrokah: Thanks @jhudson10x for the added context, makes sense 👍

plugins/source/aws/resources/services/ssoadmin/instances_fetch.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/service/ssoadmin"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchSsoadminInstances(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	svc := meta.(*client.Client).Services().SSOAdmin
+	config := ssoadmin.ListInstancesInput{}
+	response, err := svc.ListInstances(ctx, &config)
```
erezrokah, line 14: Does this API needs to be paginated like the others? It seems it supports that functionality

erezrokah (changes requested):
This is great work @jhudson10x, sorry for the late re-review.

This looks great pending a few comments. Please let me know what you think

erezrokah:
Additionally to get the CI passing you'll need to run `make gen-docs` in the AWS plugin directory and push any changes to the docs.

Finally if it makes sense, you can add mock tests to verify the fetch functions work as expected, see example(s) in https://github.com/cloudquery/cloudquery/blob/9a313bb24316b02538d2abde79bd0873bf81fe03/plugins/source/aws/resources/services/elbv2/load_balancers_mock_test.go

plugins/source/aws/resources/plugin/tables.go
```diff
@@ -221,6 +223,9 @@ func tables() []*schema.Table {
 		iam.ServerCertificates(),
 		iam.Users(),
 		iam.VirtualMfaDevices(),
+		identitystore.Groups(),
+		identitystore.Users(),
+		identitystore.GroupMemberships(),
```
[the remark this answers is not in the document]
jhudson10x: Yes I think so. I will remove line 228 because it will be sync'd as relation to table on line 226. This makes sense.

plugins/source/aws/resources/services/ssoadmin/instances_fetch.go
jhudson10x: I think there is no possibility for more than 1 instance to exist. I can paginate it. It raises another question but we can look at that in next PR.

erezrokah:
/docs sha=0ef9afa7bcfef122e6b56a65057bac8a01f2c9ce plugin-aws

erezrokah:
/docs sha=0ef9afa7bcfef122e6b56a65057bac8a01f2c9ce plugin=aws

jhudson10x:
@erezrokah I thought we could delay answering the question "is there ever more than 1 SSO instance in an account" but maybe we should address it now because I have assumed this in the change, e.g.:

```
+func getIamInstance(ctx context.Context, meta schema.ClientMeta) (types.InstanceMetadata, error) {                                                                                                                 
+       svc := meta.(*client.Client).Services().SSOAdmin                                                                                                                                                            
+       config := ssoadmin.ListInstancesInput{}                                                                                                                                                                     
+       response, err := svc.ListInstances(ctx, &config)                                                                                                                                                            
+       if err == nil {                                                                                                                                                                                             
+               for _, i := range response.Instances {                                                                                                                                                              
+                       return i, err                                                                                                                                                                               
+               }                                                                                                                                                                                                   
+       }                                                                                                                                                                                                           
+       return types.InstanceMetadata{}, err                                                                                                                                                                        
+}                                                                                                                                                                                                                  
+                                                                                                                                                                                                                   
+func fetchIdentitystoreGroups(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {                                                                                
+       instance, err := getIamInstance(ctx, meta)                                                                                                                                                                  
+       if err != nil {                                                                                                                                                                                             
+               return err                                                                                                                                                                                          
+       }     
... // use the returned instance to do fetch
```

FWIW the AWS docs clearly state you can only have one: 


> You can have only one identity source per organization in AWS Organizations.
> https://docs.aws.amazon.com/singlesignon/latest/userguide/manage-your-identity-source.html

If there is a better way to access that identiy-store-id (this is what instance is needed for in above code) then I can change it. Maybe doing table relation? I do not know if that will work across services though.

erezrokah (approved):
Thanks for the follow up digging into AWS docs 🙃 

It all makes sense to me so I approved the PR. I'm planning to add mock tests for it today/tomorrow and then I'll merge it

bbernays (approved):

PR #3007, opened 2022-10-19
doc: Document that path is not optional

shimonp21:


hermanschaaf (approved):
One small suggestion, but otherwise looks good!

PR #3027, opened 2022-10-19
chore(ci): Add workflow to re-gen source plugins docs

erezrokah:
#### Summary

This adds a workflow so we can comment on a PR with:
`/docs sha=<sha> plugin=aws` and it will re-generate the docs a commit them back to the PR branch.

See example https://github.com/cloudquery/helm-charts/pull/137#issuecomment-1282648601

That should make it easier to handle SDK changes to docs generation

### DIFF
--- a/.github/workflows/manual_approve.yml
+++ b/.github/workflows/manual_approve.yml
@@ -1,0 +1,18 @@
+name: Manual Approve
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Source Plugin Docs Generation Command
+        uses: peter-evans/slash-command-dispatch@v3
+        with:
+          token: ${{ secrets.GH_CQ_BOT }}
+          reaction-token: ${{ secrets.GITHUB_TOKEN }}
+          issue-type: pull-request
+          commands: docs
+          permission: write
--- a/.github/workflows/manual_triggers.yml
+++ b/.github/workflows/manual_triggers.yml
@@ -1,0 +1,37 @@
+name: Manual Triggers
+
+on:
+  repository_dispatch:
+    types: [docs-command]
+
+jobs:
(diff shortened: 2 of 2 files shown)


hermanschaaf:
@erezrokah what should the `sha` be in this case?

erezrokah:
Usually the top of the PR branch (for this PR `599c43d31056f99f0a045045408ba6398c829ffa`)

erezrokah:
We can also just take the `head` of the PR but using the `sha` is more secure for forks.
The commit changes runs in an elevated context, so if we always take the head of the PR, someone can changes the content of the PR after a comment was added

.github/workflows/manual_approve.yml
```diff
@@ -0,0 +1,18 @@
+name: Manual Approve
```
hermanschaaf, line 1: Is this name right? 🤔 It's called Manual Approve but it looks like it's for triggering the docs command.

hermanschaaf (approved):

.github/workflows/manual_approve.yml
erezrokah: Ah, yeah. That's because we can add more commands, so each command has it's own name:
https://github.com/cloudquery/cloudquery/pull/3027/files#diff-37ad898e2510969cb639d6ff3de59e1dd6c1c5fe2c8f6ce28661c47ce3eddd21R11

For example https://github.com/cloudquery/helm-charts/blob/07954f7161c507fe3988e9354db99274d1707a74/.github/workflows/manual_approve.yml has 2 different commands.

hermanschaaf: Shouldn't it be called `Manual Commands` or something though?

erezrokah: Yes, that's a much better name. Fixed in https://github.com/cloudquery/cloudquery/pull/3027/commits/d876a7b9771217e3a5f1b3794edb6a18d3c760b5

PR #3030, opened 2022-10-19
fix(deps): Update plugin-sdk to v0.13.11

hermanschaaf:
This upgrades all plugins to plugin-sdk v0.13.11, which made changes to how docs are generated, so all docs are also regenerated.

I've also added a `make update-plugin-sdk` convenience command for if we want to do this again in the future.

### DIFF
--- a/.github/workflows/manual_commands.yml
+++ b/.github/workflows/manual_commands.yml
@@ -1,0 +1,18 @@
+name: Manual Commands
+
+on:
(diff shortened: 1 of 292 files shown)


hermanschaaf:
I don't know whether to classify this as a feature or a chore; it's vaguely user-facing but not really :)

erezrokah (approved):

PR #3056, opened 2022-10-19
chore(docs): AWS: Add missing configuration and example config

hermanschaaf:
This adds an example config to the top of the AWS config page to better place it in context. It also adds two options under `accounts` that were undocumented before.

Also, reverting the `Organization -> org` change as I think that was misread, the config option is called `org`, but the object is an `Organization` object.


.github/workflows/release_plugin.yml
```diff
@@ -65,6 +65,15 @@ jobs:
         if: steps.semver_parser.outputs.prerelease == ''
         run: 'echo "{ \"latest\": \"${{github.ref_name}}\" }" > ./sites/versions/v2/${{steps.split.outputs.plugin_type}}-${{steps.split.outputs.plugin_name}}.json'
 
+      - name: Update version in example config files
```
erezrokah, line 68: If this get a bit too much to maintain, maybe we can use the following approach:
https://github.com/cloudquery/.github/blob/ac27d588a9a9d492b7a3a2054638119d7c1df167/.github/renovate-terraform.json5#L19

We could move that config up to the default preset:
https://github.com/cloudquery/.github/blob/ac27d588a9a9d492b7a3a2054638119d7c1df167/.github/renovate-default.json5

Downside that we run renovate on a 1 hour schedule so it won't get updated immediately (we can change the schedule)

hermanschaaf: Yeah, or maybe we should fork gha-find-replace and add the ability to use multiple glob matches. If we did that, we could collapse all three these steps into one.

plugins/source/aws/docs/configuration.md
```diff
@@ -44,7 +67,16 @@ This is used to specify one or more accounts to extract information from.
...
+- `default_region` (string) (default: `us-east-1`)
+
+  If specified, this region will be used as the default region for the account.
+
+- `regions` (string)
+
+  Regions to use for this account. Defaults to global `regions` setting.
+
+
+## Organization
```
bbernays, line 79: The field name is `org` shouldn't the title of the section match the field name?

.github/workflows/release_plugin.yml
hermanschaaf: Okay no, it would still be 2 steps. But we could further widen the options to get it down to one.

plugins/source/aws/docs/configuration.md
hermanschaaf: @bbernays Yeah maybe that would be clearer, but it wasn't the original intention. It's documented as `org` on Line 38. This `Organization` section describes the Organization object, similar to how the setting is `accounts` but the heading is `Account` (singular, describing one Account object)

.github/workflows/release_plugin.yml
erezrokah: I think we might need do a follow up on this approach as some PRs are getting blocked due to conflicts:
https://github.com/cloudquery/cloudquery/pull/3052:
image

PR #3059, opened 2022-10-19
chore(ci): Move validate release workflow to each component so caching works

erezrokah:
#### Summary

This is similar to https://github.com/cloudquery/cloudquery/pull/2970, but for our releasing workflows (https://github.com/cloudquery/cloudquery/pull/2970 was for testing workflows).

At the moment we run Go releaser dry run on release PRs, then do another dry run + wet run on published tags.
Since we run the workflows on PRs & tags the caching doesn't work, per the explanation in https://github.com/cloudquery/cloudquery/pull/2970.

To get cache working we need to run Go releaser dry run on `push` events too. We currently use a single workflow to validate all components and use the PR branch name to parse the relevant component.
Since we don't have access to the PR branch name on push events we can:
1. Keep the current workflow and use an approach like https://github.com/8BitJonny/gh-get-current-pr to get the merged PR branch
2. Create a validate release workflow file per component
3. Add another job to validate releases in the current workflow files.

**This PRs implements number 3. as it seems the most consistent with our current CI approach.**

> Please note we don't want to run `validate-release` on each PR as it can be quite slow, and also it uses a Go releaser token secret which is not available to forked PRs


hermanschaaf (approved):
Oof, quite a bit of duplication 😄  But I guess you're right, this is in line with how we're doing things right now.

erezrokah:
> Oof, quite a bit of duplication 😄 But I guess you're right, this is in line with how we're doing things right now.

Yeah, there isn't a good way to re-use parts of workflow files in GitHub actions yet (other than triggering other actions)

PR #3073, opened 2022-10-19
feat: Added aws docdb clusters

amanenk:


erezrokah (approved):

plugins/source/aws/resources/services/docdb/clusters_fetch.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchDocdbClusters(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	svc := c.Services().DocDB
+
+	input := &docdb.DescribeDBClustersInput{}
+
+	for {
```
hermanschaaf, line 19: Suggestion: DBClusters supports a paginator, which should simplify our code a little since we wouldn't need to handle the markers:  https://docs.aws.amazon.com/sdk-for-go/api/service/docdb/#DocDB.DescribeDBClustersPages

plugins/source/aws/docs/tables/aws_docdb_cluster_snapshot_attributes.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|db_cluster_snapshot_attributes|JSON|
+|db_cluster_snapshot_identifier|String|
```
hermanschaaf, line 16: I think we need to add a `db_cluster_snapshot_arn` field here to serve as a foreign key to the parent table.

hermanschaaf (changes requested):
Looks great 👍  I think there are just a few small tweaks to make before we can merge

plugins/source/aws/docs/tables/aws_docdb_cluster_snapshot_attributes.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|db_cluster_snapshot_arn|String|
+|db_cluster_snapshot_attributes|JSON|
```
hermanschaaf, line 16: Sorry, looking at this again... I think if there's only this one column in this table, and this table maps to cluster snapshots one:one, then we should rather add `attributes` as a JSON column to `aws_docdb_cluster_snapshots`, similar to how we would add tags as a JSON column despite needing another SDK call.

hermanschaaf (changes requested):
One last thing about the attributes column, but I think once that's resolved we can merge 👍

hermanschaaf (approved):

yevgenypats (approved):

Issue #3072, opened 2022-10-19
Normalize Processor Architecture Names In Releases


nick123pig:
### Is your feature request related to a problem? Please describe.
The release architecture names for x86 processors are different in the cli vs the plugin releases.

For example, on this [this CLI release](https://github.com/cloudquery/cloudquery/releases/tag/cli-v1.3.4), the string `x86_64` is used, but on [this plugin release](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-aws-v3.4.1), the string `amd64` is used.

I am building cloudquery inside multi-arch docker containers. This is problematic since I have to create two different build args depending on which component I want to download

### Describe the solution you'd like
I would like a single processor architecture name for x86; I don't care which one it is (`x86_64` or `amd64`)

PR #3080, opened 2022-10-20
fix(cli): Update binary names to use Go releaser defaults

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3072

A follow up PR for the docs is coming soon, we should update the docs only after we release a new CLI


yevgenypats (approved):
Looks great. Can you update also the docs in this PR?

PR #3091, opened 2022-10-20
doc: Add information about plugin naming conventions

hermanschaaf:


website/pages/docs/developers/creating-new-plugin.md
```diff
@@ -9,4 +9,39 @@ More real world examples are:
 - [GCP Source Plugin](https://github.com/cloudquery/cloudquery/tree/main/plugins/source/gcp)
 - [PostgreSQL Destination Plugin](https://github.com/cloudquery/cloudquery/tree/main/plugins/destination/postgresql)
 
-All other official source and destination plugins can be found [here](https://github.com/cloudquery/cloudquery/tree/main/plugins)
\ No newline at end of file
+All other official source and destination plugins can be found [here](https://github.com/cloudquery/cloudquery/tree/main/plugins)
```
erezrokah, line 12: How about linking to https://www.cloudquery.io/docs/plugins/sources?

Our GitHub repo has some test plugins too

website/pages/docs/developers/creating-new-plugin.md
```diff
@@ -9,4 +9,39 @@ More real world examples are:
...
+
+## Naming Conventions
+
+Community plugins use the following Github repository naming conventions:
```
erezrokah, line 16:
```suggestion
Community plugins use the following GitHub repository naming conventions:
```

erezrokah (approved):
Looks great, added a few non blocking comments

website/pages/docs/developers/creating-new-plugin.md
```diff
@@ -9,4 +9,39 @@ More real world examples are:
...
+
+```yaml
+kind: destination
+spec:
+  path: org/name
+```
+
+for destination plugins. 
+
+Names generally contain no dashes or underscores. So for example, if you are developing a source plugin for a new cloud service called Cloud Widgets, you should create the plugin repository under `org/cq-source-cloudwidgets`.
```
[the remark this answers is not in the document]
hermanschaaf: Yeah, I don't think we need to enforce it, but I'm happy to change the wording to `should` 👍

website/pages/docs/developers/creating-new-plugin.md
hermanschaaf: I'd argue it's better to link to the code, since we are giving developers examples to reference. Even referencing the test plugins can be useful because of their simplicity (I know I have referenced them from time to time).

erezrokah: Ah good point, so maybe remove the word "official" from "All other official".

hermanschaaf: I rephrased it slightly 👍

erezrokah (approved):

PR #3099, opened 2022-10-20
feat: Server Side Filtering for Project Discovery

bbernays:
#### Summary

Enable users to use server side filtering to choose which projects to sync


bbernays:
> Thanks for the PR @bbernays, the code looks good.
> 
> Is it correct to say that users can use `project_ids` to filter projects too?
> 
> I get that this is a more dynamic way to do it, but users could also do a `curl` request to get the list and generate our config from it, is that correct?
> 
> I'm not sure where we draw the line of which API options we expose and which ones we don't

I don't think that `project_ids` can be used to filter projects, only add them

Yes you are correct that a user could theoretically use the `ListProjects` API to get the desired list of projects and then generate the config from that. 

I don't think we have a clear line for what we expose and what we don't. For me I am comfortable exposing anything that helps users fetch exactly what they want rather than having to over fetch.

plugins/source/gcp/client/client.go
```diff
@@ -142,16 +143,18 @@ func logProjectIds(logger *zerolog.Logger, projectIds []string) {
...
 	var (
 		projects []string
 	)
 	service, err := crmv1.NewService(ctx, options...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create cloudresourcemanager service: %w", err)
 	}
-
-	call := service.Projects.List().Filter("lifecycleState=ACTIVE").Context(ctx)
+	if filter == "" {
```
yevgenypats, line 154: Can we move the default logic to one place (I know not related to this PR specifcally). but let's do a 

func (*Spec) SetDefaults()

disq (approved):
LGTM

shimonp21:
I branched off a pull-request for this feature - very similar, but some small semantic difference:

- Even if a user specifies `project_filter`, still only grab AVAILABLE projects. grabbing UNAVAILABLE projects can cause cloudquery to get stuck, and is a bad experience.
- allow specifying `project_ids` in addition to filters, and add the projects from `project_ids` to them (with set-union).
- explictly disallow filters in conjunction with folder_ids.


That PR also fixes two unrelated bugs in the folder feature (oopsie), see the individual commits.

bbernays:
Closing in favor of #3170

PR #3124, opened 2022-10-20
chore(ci): Wait for correct deployment

erezrokah:
#### Summary

Fixes the broken links workflow to wait for the correct deployment (we have the Website and our versions site, so we need to wait for the Website deployment)

### DIFF
--- a/website/pages/confirm.mdx
+++ b/website/pages/confirm.mdx
@@ -2,7 +2,6 @@
 title: Confirm
 headeronly: true
 searchable: false
-
 ---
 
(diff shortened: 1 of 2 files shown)


website/pages/confirm.mdx
```diff
@@ -2,6 +2,7 @@
 title: Confirm
 headeronly: true
 searchable: false
+
```
hermanschaaf, line 5: Was this intentional?

erezrokah: It was to trigger a website build, I'll remove it

PR #3158, opened 2022-10-20
chore(ci-wait-for-all): Use `Ana06/get-changed-files` to compare PRs with `main`

erezrokah:
#### Summary

~~It seems this action fails to use the correct `sha` of the PR if the PR branch is not up to date. This should fix it~~

Use an action that uses the GitHub API as we need a three dots diff

### DIFF
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -12,10 +12,13 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          fetch-depth: 0
+          fetch-depth: 2
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@v32
+        with:
+          base_sha: ${{ github.event.pull_request.base.sha }}
+          sha: ${{ github.event.pull_request.head.sha }}
       - uses: actions/github-script@v6
         env:
           FILES: ${{ steps.changed-files.outputs.all_changed_files }}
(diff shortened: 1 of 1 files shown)


hermanschaaf (approved):

erezrokah:
Actually I think the issue is that this action uses a two dots diff instead of a three dots diff, see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons

erezrokah:
Issue https://github.com/tj-actions/changed-files/issues/702

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -11,14 +11,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@v32
+        uses: jitterbit/get-changed-files@v1
```
erezrokah, line 16: This action uses the GitHub API to do the diff, which should use a three dots diff.
We've used it before and removed it since it caused us to hit rate limits.
Since we're now on enterprise account our limit increased from 1000 to 15000 request per hour so I think we can try this again until https://github.com/tj-actions/changed-files/issues/702 is resolved.

Another benefit is that we can do a shallow clone which is faster

erezrokah: OK so https://github.com/jitterbit/get-changed-files/issues/19#issuecomment-854657729 works

erezrokah:
@hermanschaaf sorry for the back and forth this is ready now

hermanschaaf (approved):

PR #3160, opened 2022-10-20
feat: Add eventbridge resources

shimonp21:
#### Summary

Also contains some codegen (that works specifically for resources of this format).


hermanschaaf:
@shimonp21 Would you mind splitting the EventBridge and codegen features into two different PRs, so we can discuss them separately? That would probably involve removing the `generated by codegen` comment from the top of the files in the EventBridge PR, and removing the generated EventBridge code from the codegen PR. 

I have some concerns about introducing codegen for fetches and mocks, and I'd want to see that the approach can generalize before we commit to it, but don't necessarily want the EventBridge resource to be delayed because of that.

plugins/source/aws/resources/services/eventbridge/event_buses_fetch.go
```diff
@@ -68,3 +69,31 @@ func resolveEventBridgeTags(ctx context.Context, meta schema.ClientMeta, resourc
 	}
 	return resource.Set(c.Name, client.TagsToMap(output.Tags))
 }
+
+func resolveArchiveArn(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
hermanschaaf, line 73: I think you're cheating on the codegen a little here by putting some custom resolver code in another resource's file 😉

yevgenypats (approved):
Looks good. maybe worth splitting per @hermanschaaf suggestion but I know we also have a meeting about this tomorrow so we can discuss live

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -157,5 +185,77 @@ func (r *Resource) Generate() error {
...
+	if err := os.WriteFile(filePath, content, 0644); err != nil {
+		return fmt.Errorf("failed to write file %s: %w", filePath, err)
+	}
+
+	return nil
+}
+
+func (r *Resource) generateMockTest(dir string) error {
+	tpl, err := template.New("mock_test.go.tpl").Funcs(template.FuncMap{
+		"ToCamel": strcase.ToCamel,
```
hermanschaaf, line 225: Let's use our [own caser](https://github.com/cloudquery/plugin-sdk/blob/main/caser/caser.go#L144) here rather than github.com/iancoleman/strcase

plugins/source/aws/codegen/recipes/eventbridge.go
```diff
@@ -53,10 +53,112 @@ func EventbridgeResources() []*Resource {
...
+		r.CloudqueryServiceName = "EventBridge"
+	}
+
+	for _, r := range autogenGlobalResources {
+		r.Service = "eventbridge"
+		r.Multiplex = `client.AccountMultiplex`
+		r.ShouldGenerateResolverAndMockTest = true
+		r.ResolverAndMockTestTemplate = "list_resources_1"
+		r.CloudqueryServiceName = "EventBridge"
+	}
```
hermanschaaf, line 158: Nit: Probably personal taste, but these are similar enough that I'd keep it all in one list and one for loop, and explicitly mark the ones to be generated using `ShouldGenerateResolverAndMockTest: true`. The multiplex edge case can also be handled by doing:

```
// provide a default multiplexer
if r.Multiplex == "" {
     r.Multiplex = `client.ServiceAccountRegionMultiplexer("events")`
}
```

and then setting multiplex to `client.AccountMultiplex` in the few cases that need it.

plugins/source/aws/codegen/recipes/templates/resolver_and_mock_test/list_resources_1/README.md
```diff
@@ -0,0 +1,76 @@
+This folder contains templates for resolvers and mock-tests, for AWS apis of the form `List<Resource>s` (e.g. `eventbridge.ListArchives`)
+
+Example Resolver (`eventbridge.Archives`)
+```go
```
hermanschaaf, line 4: I like the idea of including an example, but this is going to subject to frequent change, maybe we can just link to the eventbridge resource from here?

hermanschaaf (approved):
LGTM 👍 Left a few nit comments, no blockers

plugins/source/aws/resources/services/eventbridge/event_buses_fetch.go
shimonp21: Moved to `arn_resolvers.go` file

plugins/source/aws/codegen/recipes/eventbridge.go
shimonp21: Done for the boolean - all resources need to explicitly set it to false/true (good, I think).
Left two lists for `globalResources` and `regionalResources`, because I don't think having a "default" is a good fit (is "global" or "regional" a more reasonable default?...).

plugins/source/aws/codegen/recipes/base.go
shimonp21: I simply copy-pasted from our schema-resolver :) Also, because the mock-test needs to reference the Table from the schema - important that they use the same function.

Changing this is out-of-scope, I think.

Issue #3168, opened 2022-10-21
Command-line failure should write to log


hermanschaaf:
In cases where the command-line returns an error, such as here: https://github.com/cloudquery/cloudquery/blob/main/cli/cmd/sync.go#L72, the error is reported to the console by cobra, but not written to the log. This means that the process can exit with a failure, but no indication of the reason stated in the logs. For environments relying on the log file, the logs should also have the error that the CLI exited with.

Issue #3077, opened 2022-10-19
Run Migration without a sync


bbernays:
Users that run syncs in parallel need a way to be able to run a migration as a one off command because otherwise when syncs are done in parallel the initial migration will fail

hermanschaaf:
I'd like to suggest the following interface:

 - add a `migrate` command that only runs the migration, nothing else
 - add a `--skip-migration` flag to `sync` to override the default behavior of performing a migration first

This will still require some synchronization on the user-side, however. For example, you would need an orchestrator that first runs `cloudquery migrate`, waits for it to complete, and then runs all the instances of `cloudquery sync --skip-migration` in parallel.

erezrokah:
Is there a way to lock the DB during the migration? Then we won't need an additional command (I vaguely remember we talked about this problem before)

bbernays:
I agree with @erezrokah that the best way to solve the underlying issue of coordinating a parallelized sync is to run the migration as a single transaction

hermanschaaf:
We discussed this further on a call, just summarizing the outcome here:
 - we will start by adding the `migrate` command and `--skip-migration` option
 - transactions may also work, but can also lead to some unforeseen bugs/complexities, and not every destination will support that kind of function.

A typical workflow might involve running many cloudquery workers in parallel. When it's time to upgrade, a separate worker can be started to run the migration. As long as the workers are still running the old version, it shouldn't cause any issues, as migrations are backwards-compatible. Once the migration is finished, the workers can then also be upgraded.

We're not saying we definitely won't explore transactions in the future, but for this case it seems less risky to give users a bit more control over when migrations run.

Issue #2922, opened 2022-10-17
Communicate Status of Migration


bbernays:
When initializing a source plugin CloudQuery runs a set of migrations to ensure the destination is up to date. During this time the output on the CLI is completely static and does not tell the user what is going on. This can be frustrating for users if they are using a remote DB as the migrations can take upwards of 30 seconds...

PR #3169, opened 2022-10-21
feat: Add migrate command and --no-migrate option to sync

hermanschaaf:
## Description

This change adds a couple of new features:
 - a `--no-migrate` flag to the `sync` command, to skip migrations when running a sync
 - a new `migrate` command, taking the same arguments as `sync`, to run migrations and nothing else
 - print lines to note when migrations start and end
 - log lines when the CLI exits with an error

## Example interactions

Migration command:

```
cloudquery migrate cloudquery.yml
Loading spec(s) from cloudquery.yml
Starting migration for: aws -> [postgresql]
Starting migration for: gcp -> [postgresql]
```

Migration before sync (default behavior):

```
cloudquery sync cloudquery.yml
Loading spec(s) from cloudquery.yml
Starting migration for: aws -> [postgresql]
Starting sync for: aws -> [postgresql]
Sync completed successfully.
Summary: resources: 7, errors: 0, panic: 0, failed_writes: 0, time: 6s
Starting migration for: gcp -> [postgresql]
Starting sync for: gcp -> [postgresql]
Sync completed successfully.
Summary: resources: 10, errors: 0, panic: 0, failed_writes: 0, time: 6s
```

Sync without migration using `--no-migrate` flag:

```
cloudquery sync --no-migrate cloudquery.yml
Loading spec(s) from cloudquery.yml
Starting sync for: aws -> [postgresql]
Sync completed successfully.
Summary: resources: 7, errors: 0, panic: 0, failed_writes: 0, time: 6s
Starting sync for: gcp -> [postgresql]
Sync completed successfully.
Summary: resources: 10, errors: 0, panic: 0, failed_writes: 0, time: 6s
```

## Related Issues

Closes https://github.com/cloudquery/cloudquery/issues/3077
Closes https://github.com/cloudquery/cloudquery/issues/2922
Closes https://github.com/cloudquery/cloudquery/issues/3168


website/pages/docs/reference/cli/cloudquery_sync.md
```diff
@@ -38,7 +39,7 @@ cloudquery sync ./directory ./aws.yml ./pg.yml
       --log-format string        Logging format (json, text) (default "text")
       --log-level string         Logging level (default "info")
       --no-log-file              Disable logging to file
-      --telemetry-level string   Telemetry level (none, errors, stats, all) (default "all")
+      --telemetry-level string   Telemetry level (none, errors, stats, all) (default "none")
```
hermanschaaf, line 42: Not sure why this changed 🤔

hermanschaaf: Oh, it's affected by having `CQ_NO_TELEMETRY` set locally. We'll need to fix this

disq: Weird.

disq (approved):
Reading through, it's a LGTM!

disq:
Should it be `--no-migration` or `--no-migrate`? (The opposite would be `--migrate` and not `--migration`?)

hermanschaaf:
@disq Yeah I was on the fence about it, let's make it `--no-migrate` 👍

cli/cmd/sync.go
```diff
@@ -28,12 +28,13 @@ cloudquery sync ./directory ./aws.yml ./pg.yml
...
 		Use:     "sync [files or directories]",
-		Short:   fetchShort,
-		Long:    fetchShort,
-		Example: fetchExample,
+		Short:   syncShort,
+		Long:    syncShort,
+		Example: syncExample,
 		Args:    cobra.MinimumNArgs(1),
-		RunE:    sync,
+		RunE:    logErrors(sync, "Sync failed with errors"),
```
yevgenypats, line 35: can we have this just in `main` this way we don't need to wrap every cobra command. also Im not sure it's super important to log this as if just exits then usually the container will exit with -1 and it will print the error but anyway I think it's fine to also print it to the logs just let's do it in one location.

hermanschaaf: I'm not that familiar with cobra, but I'm sure how we can do it from main, because the error from `runE` doesn't seem to get returned to us anywhere? Cobra just prints it out to stdout. But I can look into it some more

cli/cmd/sync.go
```diff
@@ -100,10 +106,9 @@ func syncConnection(ctx context.Context, cqDir string, sourceSpec specs.Source,
 		}
 	}()
 
-	destClients := make([]*clients.DestinationClient, len(sourceSpec.Destinations))
-	destSubscriptions := make([]chan []byte, len(sourceSpec.Destinations))
-	for i := range destSubscriptions {
-		destSubscriptions[i] = make(chan []byte)
+	destClients, err := initializeClients(ctx, sourceSpec, destinationsSpecs, cqDir)
```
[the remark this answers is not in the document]
hermanschaaf: Nice catch! Will look into this...

cli/cmd/sync.go
hermanschaaf: Ah okay nvm, I see now. Looks like it is being logged from `main.go` already, just without the context of what command was run. I'll remove it from this PR

website/pages/docs/reference/cli/cloudquery_sync.md
hermanschaaf: It's actually fine I think. Yes, it is affected by `CQ_NO_TELEMETRY`, but that environment variable is considered deprecated and developers should be using `CQ_TELEMETRY_LEVEL=none` locally now. If you do that, the default remains `all`. Unfortunately I don't see a way to support the legacy `CQ_NO_TELEMETRY` and not have it affect the default value. :/

cli/cmd/migrate.go
```diff
@@ -0,0 +1,137 @@
...
+			fmt.Println("failed to terminate source client: ", err)
+		}
+	}()
+
+	destClients, err := initializeDestinationClients(ctx, sourceSpec, destinationsSpecs, cqDir)
+	// we make sure to defer the closing of clients here before checking the error, otherwise
+	// if one of the clients returns an error, the clients opened before it may never get closed
+	defer func() {
+		for _, destClient := range destClients {
+			if destClient != nil {
```
disq, line 98: Wouldn't really work because if you look at `initializeDestinationClients`, partly-initialized slice isn't returned if there was an error.

disq (changes requested):

cli/cmd/migrate.go
hermanschaaf: Ah, nice catch. My plan was to return the slice from that function, but looks like I forgot to do that! 🤦

hermanschaaf: Updated, PTAL 🙇

cli/cmd/clients.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/cloudquery/plugin-sdk/clients"
+	"github.com/cloudquery/plugin-sdk/specs"
+	"github.com/rs/zerolog/log"
+)
+
+func initializeDestinationClients(ctx context.Context, sourceSpec specs.Source, destinationsSpecs []specs.Destination, cqDir string) ([]*clients.DestinationClient, error) {
+	var err error
```
yevgenypats, line 12: nit: let's also have the close logic here in case of an error. i.e this will return nil, err and close all clients in case one of them fails. + outside if it is not nil this will close all destinations - a bit of duplication but I think will make the code more streamlined/error-proof.

cli/cmd/migrate.go
```diff
@@ -0,0 +1,137 @@
...
+	for i, destinationSpec := range destinationsSpecs {
+		tables, err := sourceClient.GetTables(ctx)
+		if err != nil {
+			return count, fmt.Errorf("failed to get tables for source %s: %w", sourceSpec.Name, err)
+		}
+
+		if err := destClients[i].Migrate(ctx, tables); err != nil {
+			return count, fmt.Errorf("failed to migrate source %s on destination %s : %w", sourceSpec.Name, destinationSpec.Name, err)
+		}
+		count += len(tables)
```
yevgenypats, line 134: let's remove this count from here. it's not real stats/metrics because it is static. so we can just print the len of tables*detination outside and avoid another return parameter

yevgenypats (approved):
I think looks good two more nits.

cli/cmd/clients.go
hermanschaaf: Yeah, makes sense, I'll updated it 👍

cli/cmd/migrate.go
hermanschaaf: Removed, and refactored a bit because of it 👍

cli/cmd/clients.go
hermanschaaf: @yevgenypats I've refactored it a bit accordingly; rather than have the same code several times I felt it's a little nicer  to have a `destinationClients []*clients.DestinationClient` type with a `Close()` method.

hermanschaaf:
@yevgenypats the code changed a bit since your last review (but functionally still the same), so if you don't mind please have another look 🙇

yevgenypats (approved):
🚀

erezrokah (approved):
🎸

website/pages/docs/reference/cli/cloudquery_sync.md
erezrokah: It's a bit blunt but https://github.com/cloudquery/cloudquery/pull/3180 should help with this.
Not sure if there's a better way to do it via Cobra as I guess we always want to generate the docs without any env variables set

disq (approved):
🚀

PR #3170, opened 2022-10-21
feat: Project filter

shimonp21:
#### Summary

See individual commits for easier reviewing.

1) Fixes two previous bugs in the folder feature [forgotten recursion, forgotten set-union].
2) Adds support for the project-filter feature.

The project-filter is mutually exclusive with `folder_ids`, but `project_ids` that are specified will be added to the projects from the filter. More explicitly:
image

### DIFF
--- a/plugins/source/gcp/client/client.go
+++ b/plugins/source/gcp/client/client.go
@@ -188,7 +188,7 @@ func listFolders(ctx context.Context, folderClient *resourcemanagerv3.FoldersCli
 	})
 
 	for {
-		folder, err := it.Next()
+		child, err := it.Next()
 
 		if err == iterator.Done {
 			break
@@ -197,8 +197,12 @@ func listFolders(ctx context.Context, folderClient *resourcemanagerv3.FoldersCli
 			return nil, err
 		}
 
-		if folder.State == resourcemanagerv3pb.Folder_ACTIVE {
-			folders = append(folders, folder.Name)
+		if child.State == resourcemanagerv3pb.Folder_ACTIVE {
+			childFolders, err := listFolders(ctx, folderClient, child.Name, recursionDepth-1)
+			if err != nil {
+				return nil, err
+			}
+			folders = append(folders, childFolders...)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


plugins/source/gcp/client/client.go
```diff
@@ -173,6 +192,35 @@ func getProjectsV1(ctx context.Context, options ...option.ClientOption) ([]strin
...
+	for {
+		output, err := call.Do()
+		if err != nil {
+			return nil, err
+		}
+		for _, project := range output.Projects {
+			if project.LifecycleState == "ACTIVE" {
+				projects = append(projects, project.ProjectId)
+			}
+		}
```
bbernays, line 214: While I agree that this is a good protection for users, I think we need to explicitly warn users that we are skipping projects

plugins/source/gcp/client/client.go
```diff
@@ -105,14 +111,27 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source) (schema.Cli
...
 		c.logger.Info().Msg("listing folder projects..")
 		folderProjects, err := listProjectsInFolders(ctx, c.Services.ResourcemanagerProjectsClient, folderIds)
-		projects = append(projects, folderProjects...)
+		projects = setUnion(projects, folderProjects)
 		if err != nil {
 			return nil, fmt.Errorf("failed to list projects: %w", err)
 		}
+
+	case len(gcpSpec.ProjectFilter) > 0:
+		c.logger.Info().Msg("Listing projects with filter..")
```
disq, line 120: nit: ".." is a bit informal, would change to ellipsis ("...") (same for line 112)

plugins/source/gcp/client/client.go
```diff
@@ -105,14 +111,27 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source) (schema.Cli
...
+		projectsWithFilter, err := getProjectsV1WithFilter(ctx, gcpSpec.ProjectFilter, options...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get projects with filter: %s", err)
```
disq, line 123: nit: I would use `%w` for error
```suggestion
			return nil, fmt.Errorf("failed to get projects with filter: %w", err)
```

disq (approved):
Couple of nits.

plugins/source/gcp/client/client.go
shimonp21: Before this PR, we just skipped them (because they make cloudquery hang).
I think it's out-of-scope to change this behaviour.

bbernays: I would argue that it isn't out of scope because before this we didn't expose the filter option. By doing this without telling the user our results will not match the results of a user just running `gcp projects -filter...`.

PR #3180, opened 2022-10-23
chore(make): Unset envs before generating docs

erezrokah:
#### Summary

Not sure if this can be done via the Cobra `doc` command, but this PR should help avoid https://github.com/cloudquery/cloudquery/pull/3169#discussion_r1001608633


cli/Makefile
```diff
@@ -10,5 +10,5 @@ test:
 .PHONY: gen-docs
 gen-docs:
 	rm -rf ../website/pages/docs/reference/cli/*.md
-	go run main.go doc ../website/pages/docs/reference/cli/
+	unset CQ_NO_TELEMETRY && unset CQ_TELEMETRY_LEVEL && go run main.go doc ../website/pages/docs/reference/cli/
```
yevgenypats, line 13: why this is needed here? telemetry is only used in the sync command?

yevgenypats (changes requested):
Im not sure this is needed / has any effect

cli/Makefile
erezrokah: The variables are used at the root level at the moment:
https://github.com/cloudquery/cloudquery/blob/ff122b5194a7b817bb3f719a4f928fda22a41059/cli/cmd/root.go#L47

Same as the telemetry code (it's at the root level).

We can refactor that code to be only under `sync` but that's a bigger change.

You can run with `CQ_TELEMETRY_LEVEL=none gen-docs` without this fix to see the bug

yevgenypats: I see. ok that's fine even though we don't send any analytics other then in the sync command but sure it can keep the load on the server lighter.

yevgenypats: but actually now Im even more confused. why we are doing `unset` instead pf `CQ_TELEMETRY_EVEL=none`

erezrokah: > but actually now Im even more confused. why we are doing `unset` instead pf `CQ_TELEMETRY_EVEL=none`

Ah, I didn't explained this properly. It's not about not sending analytics for the docs command.
When cobra generates the docs, it uses whatever you set in `CQ_TELEMETRY_LEVEL` as the default, and it shows up in the markdown files.
See examples:

The following command
```bash
CQ_TELEMETRY_LEVEL=none go run main.go doc ../website/pages/docs/reference/cli
```
generates
```bash
--telemetry-level string   Telemetry level (none, errors, stats, all) (default "none")
```

The following command
```bash
CQ_TELEMETRY_LEVEL=errors go run main.go doc ../website/pages/docs/reference/cli
```
generates
```bash
--telemetry-level string   Telemetry level (none, errors, stats, all) (default "errors")
```

Hence we want to unset any environment variables that impact the defaults

erezrokah:
Closing in favor of https://github.com/cloudquery/cloudquery/pull/3458

PR #3181, opened 2022-10-23
fix(azure): Handle nil `VnetName`

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/453 (internal issue).
We need to check that `VnetName` is not `nil` too as we pass its value via `*`


plugins/source/azure/codegen/recipes/web.go
```diff
@@ -23,11 +23,12 @@ func Web() []Resource {
 	var vnetInfoResource = resourceDefinition{
 		azureStruct:  &web.VnetInfo{},
 		listFunction: "GetVnetConnection",
-		listFunctionArgsInit: []string{"site := parent.Item.(web.Site)", `if site.SiteConfig == nil {
-			               return nil
-			       }
-				   `},
-		listFunctionArgs: []string{"*site.ResourceGroup", "*site.Name", "*site.SiteConfig.VnetName"},
+		listFunctionArgsInit: []string{"site := parent.Item.(web.Site)", `vnetName := funk.Get(site, "SiteConfig.VnetName")
```
disq, line 26: why not just do `if site.SiteConfig == nil || site.SiteConfig.VnetName == nil { return nil }` ?

disq: @erezrokah yes but it also removes type safety a bit (you have to do `vnetname.(*string)` which might get broken in an azure "upgrade") keeping it basic/regular also adds readability in a sense that a maintainer doesn't have to know about funk.Get.

erezrokah: Good point, I'm counting on the mock tests to validate any changes to the Azure API.
Anyway I don't have a strong opinion on this, so I'll go ahead and update the code to your suggestion

yevgenypats (approved):
looks good though not a fan of the auto-generated fetcher :) 

The value for auto-generating the struct is clear to me as:
1) it is a lot of code
2) if the struct changes it will re-generate this

for the fetcher it the same amount of code and it is harder to maintain and contribute, if something changes in the struct it doesn't regenerate it as it just a static template.

PR #3195, opened 2022-10-25
feat: Add GH Workflow support

bbernays:
#### Summary

Add support Github Actions

To do:
- [x] Add unit test
- [x] Add Support For `GetWorkflowFile`


plugins/source/github/resources/services/actions/workflows_fetch.go
```diff
@@ -0,0 +1,39 @@
...
+				if opts.Page == resp.LastPage {
+					break
+				}
+			}
+		}
+		opts.Page = resp.NextPage
+		if opts.Page == resp.LastPage {
+			break
+		}
+	}
```
bbernays, line 37: This is not optimized as the only multiplexer that is available is an org multiplexer

erezrokah (approved):
Thanks @bbernays this works pending a way to multiplex repositories (unrelated to this PR).

Approved pending a mock test for it. Is `GetWorkflowFile` required? Looking the GitHub API we get the same data as the list operation:
https://docs.github.com/en/rest/actions/workflows#get-a-workflow

hermanschaaf (approved):

erezrokah:
@bbernays Should we bring this PR up to speed?

erezrokah:
So I brought this PR up to speed in https://github.com/cloudquery/cloudquery/pull/3195/commits/c7d7029e928264df0d2dcf376f57b33dabb04561 and https://github.com/cloudquery/cloudquery/pull/3195/commits/cbfe918a865d1e09ac1b6462f65252f9a9153923.

https://github.com/cloudquery/cloudquery/pull/3195/commits/cbfe918a865d1e09ac1b6462f65252f9a9153923 also fixes the `contents` column so:
1. It's a string and not a JSON. We can't guarantee that the workflow file is a valid YAML (it can have syntax errors, also GitHub has a custom parser for it so it's does not follow the spec). Better to keep the data raw
2. Uses `HTMLURL` as that also has the branch name, which will give us a better representation of the file's content
3. If we fail to set get the content, set the column value to `nil` instead of failing

PR #3196, opened 2022-10-25
fix(gcp): Skip unusable field PrivateKeyData

SCKelemen:
`ServiceAccountKey` in GCP IAM contains an unused field. The `PrivateKeyData` field is only returned once, at the time of key creation. Therefore this field will always be null.

- [ ] Update or add tests 🧪


erezrokah:
/docs sha=42feff17179976dad84f1fe38b6d4bbd60fb7901 plugin=gcp

erezrokah:
@SCKelemen You'll need to re-generate the code gen via `make gen-code`

SCKelemen:
This field should always be null, as this api won't ever return anything other than null. Should dropping this column be a breaking change?

yevgenypats (approved):

yevgenypats:
> This field should always be null, as this api won't ever return anything other than null. Should dropping this column be a breaking change?

dropping is not a breaking change as our auto-migrator wont drop columns by default

PR #3206, opened 2022-10-27
chore(docs): Update GCP version

chrisj-back2work:
Version string was missing the "v"; and referenced an older version


erezrokah (approved):
Thanks @chrisj-back2work 🚀

PR #3209, opened 2022-10-27
chore(website): Remove observability section from docs

yevgenypats:
observability via loki is going to be deprecated or at least not the first option. 

Metrics are in progress right now - https://github.com/cloudquery/plugin-sdk/ .

Removing as caused confusion for users given that's not in working state.


erezrokah (approved):
Looks good, can we also add a redirect in https://github.com/cloudquery/cloudquery/blob/454a62ae66e43b49448ddddefd797800c8db6085/website/vercel.json from `/docs/deployment/observability` to `/docs/deployment/overview` in case it is referenced from somewhere?

It's a small change and might save some people hitting a 404

erezrokah (approved):

Issue #3203, opened 2022-10-27
inspector findings taking long to load


nkannekanti132:
There are two tables for inspector_findings and cloudquery gets stuck for more than an hour for inspector findings
There is no data in inspector_findings table and data is present in inspector_findings_2 table

erezrokah:
Hi @nkannekanti132 and thank for reporting this.

The reason we have two versions of `inspector_findings` is because AWS has two versions of that service:
1. [Amazon Inspector Classic](https://docs.aws.amazon.com/inspector/v1/userguide/inspector_introduction.html) which maps out to the `aws_inspector_findings` table
2. [Amazon Inspector](https://docs.aws.amazon.com/inspector/latest/user/what-is-inspector.html) which maps out to the `aws_inspector2_findings` table

Can you confirm you're using the second service?

As for the time it takes to sync the data, I think there's something we can do to improve and will post a PR with a fix soon.

erezrokah:
A follow up on this, [I've increased the number of results we fetch per API call for Amazon Inspector Classic to 500](https://github.com/cloudquery/cloudquery/pull/3264) so that should speed things up for that service.

Unfortunately for the new Amazon Inspector we can only retrieve 100 findings at a time. Do you mind sharing how many you have? That could explain the time it take to sync those

erezrokah:
Also, can you share if you see any errors in `cloudquery.log`?

erezrokah:
Going to close this as stale pre my comments. @nkannekanti132 please respond if relevant and I'll re-open

PR #3264, opened 2022-10-27
feat(aws): Increase inspector v1 ListFindings MaxResults to maximum

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/3203

~~v1 inspector findings can have a `MaxResults` of 500 per the Go API docs and also [here](https://docs.aws.amazon.com/inspector/v1/APIReference/API_ListFindings.html#Inspector-ListFindings-request-maxResults)~~ Actually seems like this won't work, as `DescribeFindings` has a limit of 10 ARNs 🙃 , so we should actually decrease the number per https://docs.aws.amazon.com/inspector/v1/APIReference/API_DescribeFindings.html#Inspector-DescribeFindings-request-findingArns.

What's more confusing is that the docs say the response can have 100 items:
https://docs.aws.amazon.com/inspector/v1/APIReference/API_DescribeFindings.html#Inspector-DescribeFindings-response-findings

v2 Go API docs don't say what's the limit but [here](https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListFindings.html#inspector2-ListFindings-request-maxResults) is says 100, so needs more experimenting

### DIFF
--- a/plugins/source/aws/resources/services/inspector/findings_fetch.go
+++ b/plugins/source/aws/resources/services/inspector/findings_fetch.go
@@ -12,7 +12,7 @@ import (
 func fetchInspectorFindings(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
 	c := meta.(*client.Client)
 	svc := c.Services().Inspector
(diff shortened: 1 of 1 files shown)


erezrokah:
Going to close this until the user responds to the issue

Issue #1810, opened 2022-09-13
Integrate with Azure Web App Service Auth settings


kvish:
### Why do you need this resource or plugin? Please describe.
 https://docs.microsoft.com/en-us/rest/api/appservice/web-apps/get-auth-settings-v2-without-secrets
 
 USe web app v2 api to get auth settings without secrets

### Describe the solution you'd like

We want to track if auth is enabled on the web app service to ensure least privilege.

erezrokah:
Thanks for the issue @kvish, please let me know if https://github.com/cloudquery/cloudquery/pull/3269 is what you had in mind

PR #3269, opened 2022-10-27
feat(Azure): Add site auth settings v2

erezrokah:
#### Summary

Draft PR since needs some testing. Fixes https://github.com/cloudquery/cloudquery/issues/1810


plugins/source/azure/client/services/web.go
```diff
@@ -1,4 +1,4 @@
-//go:generate mockgen -destination=./mocks/web.go -package=mocks . WebAppsClient,WebSiteAuthSettingsClient,WebVnetConnectionsClient,WebPublishingProfilesClient
+//go:generate mockgen -destination=./mocks/web.go -package=mocks . WebAppsClient,WebSiteAuthSettingsClient,WebSiteAuthSettingsV2Client,WebVnetConnectionsClient,WebPublishingProfilesClient
```
erezrokah, line 1: These changes are manually generated (adding the client to the services package)

plugins/source/azure/codegen/recipes/base.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/Azure/go-autorest/autorest/date"
+	"github.com/cloudquery/plugin-sdk/caser"
```
erezrokah, line 12: These change are needed to properly handle the `v2` in the service name.

> There's some follow up work to do here to remove the usage of `strcase` altogether but I'll do it in a follow up PR

plugins/source/azure/codegen/recipes/web.go
```diff
@@ -20,6 +20,21 @@ func Web() []Resource {
 		mockListResult:           mockDirectResponse,
 	}
 
+	var authSettingsV2Resource = resourceDefinition{
```
erezrokah, line 23: These changes are manually generated (adding `SiteAuthSettingsV2` as a relation of `app`)

erezrokah:
/docs sha=509659b634c7af3616feeb5d60b28d56515571f3 plugin=azure

erezrokah:
> Draft PR since needs some testing. Fixes #1810

Tested, it works 👍

plugins/source/azure/codegen/recipes/web.go
```diff
@@ -20,6 +20,21 @@ func Web() []Resource {
...
+		azureStruct:          &web.SiteAuthSettingsV2{},
+		listFunction:         "GetAuthSettingsV2",
+		listFunctionArgsInit: []string{"site := parent.Item.(web.Site)"},
```
yevgenypats, line 26: Not necessarily related to this PR and I think brought this up in the past. Im not a big fan of this part which generates the fetch/resolver:

```
		listFunctionArgsInit: []string{"site := parent.Item.(web.Site)"},
		listFunctionArgs:     []string{"*site.ResourceGroup", "*site.Name"},
		listHandler: `if err != nil {
				return err
			}
			res <- response`,
		mockListFunctionArgsInit: []string{""},
```

hand-writing the resolver is the same amount of lines of code:

```
func fetchWebSiteAuthSettingsV2(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
	svc := meta.(*client.Client).Services().Web.SiteAuthSettingsV2

	site := parent.Item.(web.Site)
	response, err := svc.GetAuthSettingsV2(ctx, *site.ResourceGroup, *site.Name)
	if err != nil {
		return err
	}
	res <- response
	return nil
}
```

but it makes the following easier:
1) making changes with auto-complete and code editor. 
2) development cycle shorter as you don't need to go to the `string` that generates the same amount of code.

yevgenypats (changes requested):
Looks good. Not a blocker but I do think we need to discuss the resolver generation in Azure specifically.

yevgenypats (approved):

plugins/source/azure/codegen/recipes/web.go
erezrokah: > hand-writing the resolver is the same amount of lines of code

I think this misses the auto generation of relations, mocks and related hierarchy, specifically:
https://github.com/cloudquery/cloudquery/pull/3269/files#diff-ace9a63acb9a115db549f8e20ad70309187d445c0c5896f0c19e08ed4b8fb019R272
https://github.com/cloudquery/cloudquery/pull/3269/files#diff-fa2265cf35c6f5c185b76162db90ff3747eaf79d53a6c96ad5e470227db68880R31
https://github.com/cloudquery/cloudquery/pull/3269/files#diff-ed8636bbd866046d78e64dcc81ec0e14980f09b272fb0f51257ed4be760ccf2fR17

which are also auto generated.

I do think it need to be improved in a different way. Instead of manually writing resolvers that need custom code via the code gen, we should write that code under the service, e.g https://github.com/cloudquery/cloudquery/pull/3269/files#diff-d8dc88c32f2472a2227edefe1c0e386ffcff5b041ee83ef32f47de17497e77baR38 should be a wrapper on top of the Azure SDK that gets the `Site` then does any custom code inside it.

Then we'll have a separation between autogenerated code and manually written code

Issue #1410, opened 2022-03-24
Add Support for AWS::ServiceCatalog::*


bbernays:
Add support for the following (at the very least):

- [ ] AWS::ServiceCatalog::CloudFormationProduct
- [ ] AWS::ServiceCatalog::CloudFormationProvisionedProduct
- [ ] AWS::ServiceCatalog::Portfolio

PR #3275, opened 2022-10-28
feat(aws): AWS ServiceCatalog

disq:
Implements #1410


plugins/source/aws/Makefile
```diff
@@ -1,7 +1,7 @@
 # Generate mocks for mock/unit testing 
-.PHONY: generate-mocks
-generate-mocks:
-	go generate ./client/services/...
+.PHONY: gen-mocks
+gen-mocks:
```
disq, line 3: Since we have other "gen-*" targets it only made sense to rename this to match those.

plugins/source/aws/Makefile
```diff
@@ -1,7 +1,7 @@
...
+	go generate ./client/...
```
disq, line 4: Mocks are not in `./client/services/` any more so this was a no-op before.

plugins/source/aws/client/services.go
```diff
@@ -831,6 +833,18 @@ type SecretsManagerClient interface {
...
+//go:generate mockgen -package=mocks -destination=./mocks/mock_service_catalog.go . ServiceCatalogClient
+type ServiceCatalogClient interface {
+	ListPortfolios(ctx context.Context, input *servicecatalog.ListPortfoliosInput, optFns ...func(*servicecatalog.Options)) (*servicecatalog.ListPortfoliosOutput, error)
+	SearchProductsAsAdmin(ctx context.Context, params *servicecatalog.SearchProductsAsAdminInput, optFns ...func(*servicecatalog.Options)) (*servicecatalog.SearchProductsAsAdminOutput, error)
+	SearchProvisionedProducts(ctx context.Context, params *servicecatalog.SearchProvisionedProductsInput, optFns ...func(*servicecatalog.Options)) (*servicecatalog.SearchProvisionedProductsOutput, error)
+}
+
+//go:generate mockgen -package=mocks -destination=./mocks/mock_service_catalog_app_registry.go . ServiceCatalogAppRegistryClient
+type ServiceCatalogAppRegistryClient interface {
+	ListTagsForResource(ctx context.Context, params *servicecatalogappregistry.ListTagsForResourceInput, optFns ...func(*servicecatalogappregistry.Options)) (*servicecatalogappregistry.ListTagsForResourceOutput, error)
```
disq, line 845: This is used to get tags for `ListPortfolios` and `SearchProductsAsAdmin` results. Completely unsure if it will work or error out.

erezrokah (approved):
Looks good to me 🚀 

Optional - add a stack to https://github.com/cloudquery/deployments/tree/main/backup-plugins/aws/terraform so we can test it manually

PR #3286, opened 2022-10-30
feat: Update all plugins to SDK with metrics and DFS scheduler

yevgenypats:
Should be merged after this SDK PR https://github.com/cloudquery/plugin-sdk/pull/318


plugins/source/aws/resources/services/backup/global_settings_fetch.go
```diff
@@ -15,14 +15,6 @@ func fetchBackupGlobalSettings(ctx context.Context, meta schema.ClientMeta, pare
 
 	output, err := svc.DescribeGlobalSettings(ctx, &input)
 	if err != nil {
-		if client.IgnoreAccessDeniedServiceDisabled(err) || client.IsAWSError(err, "ERROR_9601") /* "Your account is not a member of an organization" */ {
```
erezrokah, line 18: How about we keep this change (and other removals of log messages/errors) to a separate PR.
Basically only do the `meta.Logger()` -> `meta.(*client.Client).Logger()` changes to keep the size of the PR small?

We can do a follow up to look into the error handling in various AWS resources and see how we want to handle them? I agree that it looks like we can remove most of them, but I can't be sure for a few

plugins/source/azure/Makefile
```diff
@@ -24,6 +24,10 @@ gen-docs:
 	rm -rf ./docs/tables/*
 	go run main.go doc ./docs/tables
 
+.PHONY: lint
+lint:
+	golangci-lint run
```
erezrokah, line 29:
```suggestion
	golangci-lint run --config ../../.golangci.yml 
```

We need to point to the shared config 2 levels up to match what we do in CI

plugins/source/cloudflare/Makefile
```diff
@@ -24,6 +24,10 @@ gen-docs:
 	rm -rf ./docs/tables/*
 	go run main.go doc ./docs/tables
 
+.PHONY: lint
+lint:
+	golangci-lint run
```
erezrokah, line 29:
```suggestion
	golangci-lint run --config ../../.golangci.yml 
```

plugins/source/github/Makefile
```diff
@@ -24,6 +24,10 @@ gen-docs:
 	rm -rf ./docs/tables/*
 	go run main.go doc ./docs/tables
 
+.PHONY: lint
+lint:
+	golangci-lint run
```
erezrokah, line 29:
```suggestion
	golangci-lint run --config ../../.golangci.yml 
```

plugins/source/k8s/Makefile
```diff
@@ -18,6 +18,10 @@ gen-docs:
 	rm -rf ./docs/tables/*
 	go run main.go doc ./docs/tables
 
+.PHONY: lint
+lint:
+	golangci-lint run
```
erezrokah, line 23:
```suggestion
	golangci-lint run --config ../../.golangci.yml 
```

plugins/source/okta/Makefile
```diff
@@ -19,6 +19,10 @@ install-tools:
 install-hooks:
 	pre-commit install
 
+.PHONY: lint
+lint:
+	golangci-lint run
```
erezrokah, line 24:
```suggestion
	golangci-lint run --config ../../.golangci.yml 
```

erezrokah (approved):
This looks 🚀  Added a few non blocking comments.

I'm assuming this needs to come with a separate PR for the CLI to call `GetMetrics`? Are we planning to release the CLI first, then plugins or the other way around?

plugins/source/aws/resources/services/backup/global_settings_fetch.go
yevgenypats: I think it's a nit, it's already split. Let's keep like this and address other issue with AWS later on with unneeded/un-normalized log messages and so on.

plugins/source/aws/client/client.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/service/docdb"
+
```
erezrokah, line 11:
```suggestion
```

Not sure about this change

erezrokah (approved):

Issue #3172, opened 2022-10-21
batch insert errors are not clear


amanenk:
### Describe the Bug

when batch size is 1000 and the one of the inserted item is too big postgresql destination plugin logs next error 
```
ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 24192 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_engine_versions`
```
If I set batch size to 1 there are next errors
```
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 24192 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 24192 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 19960 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 24192 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 21216 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 21216 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 19960 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
8:51PM ERR failed to write to destination error="failed to execute batch: ERROR: index row requires 21216 bytes, maximum size is 8191 (SQLSTATE 54000)" table=aws_docdb_cluster_parameter_groups
```
I assume that plugin takes the last table name used in batch insert.

The error feels a bit misleading because it shows the wrong table name.

yevgenypats:
[...]

PR #3323, opened 2022-10-30
feat: Migrate cli, plugins and destinations to new type system

yevgenypats:
This is instead of https://github.com/cloudquery/cloudquery/pull/3176 

SDK PRs:

https://github.com/cloudquery/plugin-sdk/pull/318
https://github.com/cloudquery/plugin-sdk/pull/320

Previous related CloudQuery PRs:

https://github.com/cloudquery/cloudquery/pull/3286


cli/cmd/sync2.go
```diff
@@ -0,0 +1,148 @@
...
+			clients.WithDestinationDirectory(cqDir),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to create destination plugin client for %s: %w", destinationSpec.Name, err)
+		}
+		if err := destClients[i].Initialize(ctx, destinationSpec); err != nil {
+			return fmt.Errorf("failed to initialize destination plugin client for %s: %w", destinationSpec.Name, err)
+		}
+
+		if err := destClients[i].Migrate(ctx, tables); err != nil {
```
hermanschaaf, line 58: @yevgenypats Maybe now is a good opportunity to change this so that we only migrate the tables specified in the config? This would also be nice for csv, since only the files for tables specified in the config will get created.

cli/cmd/sync2.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	_ = bar.Finish()
+
+	metrics, err := sourceClient.GetMetrics(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get metrics for source %s: %w", sourceSpec.Name, err)
+	}
+
+	fmt.Printf("Sync completed successfully. Resources: %d, Errors: %d, Panics: %d\n", metrics.TotalResources(), metrics.TotalErrors(), metrics.TotalPanics())
```
hermanschaaf, line 135: Looks like we're missing the time taken now, which I think was quite nice to have.

plugins/destination/postgresql/client/client.go
```diff
@@ -117,20 +113,11 @@ func New(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (pl
...
 func (c *Client) Close(ctx context.Context) error {
 	var err error
 	if c.conn == nil {
 		return fmt.Errorf("client already closed or not initialized")
 	}
-	if c.batch.Len() > 0 {
-		br := c.conn.SendBatch(ctx, c.batch)
-		err = br.Close()
-		c.batch = &pgx.Batch{}
-	}
```
hermanschaaf, line 133: Was this moved somewhere else?

plugins/destination/postgresql/client/client_test.go
```diff
@@ -6,38 +6,43 @@ import (
...
-		"interval_column":     "01:02:03",
-		"json_column":         map[string]interface{}{"1": float64(1), "test": "test"},
-		"uuid_array_column":   []interface{}{"1a6011b7-c5ee-4b55-95a6-37ce5e02a5a0", "9a6011b7-c5ee-4b55-95a6-37ce5e02a5a0"},
-		"inet_column":         "1.1.1.1",
-		"inet_array_column":   []interface{}{"8.8.8.8/0"},
-		"cidr_column":         "0.0.0.0/24",
-		"cidr_array_column":   []interface{}{"0.0.0.0/24", "0.0.0.0/16"},
-		"mac_addr_column":     "00:00:00:00:00:ab",
-	}
+	return nil
```
hermanschaaf, line 18: Looks like this still needs to be implemented

hermanschaaf (commented):
Leaving a few comments in the meantime, but I haven't quite read through it all yet; looks like there are still some moving parts here so I'll focus more on https://github.com/cloudquery/plugin-sdk/pull/320

plugins/source/aws/resources/services/lambda/functions_fetch.go
```diff
@@ -90,8 +91,8 @@ func resolvePolicyCodeSigningConfig(ctx context.Context, meta schema.ClientMeta,
 
 	// skip getting CodeSigningConfig since containerized lambda functions does not support this feature
 	// value can be nil if the caller doesn't have GetFunctionConfiguration permission and only has List*
-	lambdaType, ok := resource.Get("code_repository_type").(*string)
```
erezrokah, line 93: I'm not that happy about `cqtypes` leaking into resolvers. How would a plugin author know which type to use in the casting?

I'd expect us to maintain the original SDK type (I understand this is a problem we had before, but `cqtypes` makes it worse I think).

Maybe we can change this to `r.Code.RepositoryType` (with `nil` checks)

plugins/source/aws/resources/services/s3/buckets_fetch.go
```diff
@@ -136,11 +137,12 @@ func resolveS3BucketsAttributes(ctx context.Context, meta schema.ClientMeta, res
 func fetchS3BucketGrants(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
 	r := parent.Item.(*models.WrappedBucket)
 	svc := meta.(*client.Client).Services().S3
-	if parent.Get("region").(string) == "" {
```
erezrokah, line 139: Similar to before, maybe `r.Region` will work for these? Or we do some processing on it?

plugins/source/digitalocean/codegen/main.go
```diff
@@ -131,7 +131,7 @@ var Resources = []*recipes.Resource{
 		ExtraColumns: []codegen.ColumnDefinition{
 			{
 				Name:     "id",
-				Type:     schema.TypeString,
```
erezrokah, line 134: This seems correct, but I think the fix should be to drop these custom definitions and use a custom caser, for example https://github.com/cloudquery/cloudquery/blob/9d5959985b62ebb29f56df61efba31a4fed893cf/plugins/source/aws/codegen/recipes/base.go#L78

erezrokah (changes requested):
Added a few comments, also we need to add the CSV plugin to:
https://github.com/cloudquery/cloudquery/blob/9d5959985b62ebb29f56df61efba31a4fed893cf/release-please-config.json#L22
https://github.com/cloudquery/cloudquery/blob/9d5959985b62ebb29f56df61efba31a4fed893cf/.github/workflows/wait_for_required_workflows.yml#L51
https://github.com/cloudquery/cloudquery/blob/9d5959985b62ebb29f56df61efba31a4fed893cf/.github/pr_labeler.yml#L23

May I suggest to put the new CSV plugin in a separate PR?

Also my main issue is that `cqtypes` is leaking to the resolvers (we had this before actually when resolvers access the table data).
Not sure how to go about it, but maybe we can always access the struct item instead?

plugins/destination/postgresql/client/types.go
```diff
@@ -50,7 +52,7 @@ func (*Client) SchemaTypeToPg10(t schema.ValueType) string {
 	case schema.TypeIntArray:
 		return "bigint[]"
 	default:
-		return "text"
```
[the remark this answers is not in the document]
yevgenypats: Now we will just skip this (I need to add this logic). columns that are not supported will be just skipped and not inserted

plugins/destination/postgresql/client/write_test.go
```diff
@@ -1,144 +0,0 @@
-package client
```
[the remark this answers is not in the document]
yevgenypats: this moved to plugin-sdk as all destination have the same write/read test logic similar to how we do that in source plugins with a helper

plugins/source/aws/resources/services/iam/users_fetch.go
```diff
@@ -74,19 +73,9 @@ func fetchIamUserAccessKeys(ctx context.Context, meta schema.ClientMeta, parent
 		for i, key := range output.AccessKeyMetadata {
 			switch i {
 			case 0:
-				rotated := parent.Get("access_key_1_last_rotated")
```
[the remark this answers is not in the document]
yevgenypats: Yeah our CqTypes caught a bug because this doesn't exist probably it was there about a year ago in an old version

plugins/source/aws/resources/services/lambda/functions_fetch.go
yevgenypats: I think this is fine and we had the same behaviour before but actually even worth because you didn't really know what they type will be after it will be "resolved" but now it always the same per the definition of the column.

cli/cmd/sync2.go
yevgenypats: Agree. I think they are quite a few improvements that we can/need to do. for example also to spawn only one process per unique destination. but I suggest doing this in follow-up smaller PRs. should be also easier to review and easier/faster for me to develop

plugins/destination/postgresql/client/client.go
yevgenypats: yeah. batches are created on write as we don't want to reuse them between streams. and they are closed/send at the end of the stream rather then on close

plugins/destination/postgresql/client/client_test.go
yevgenypats: Moved it to SDK part

plugins/destination/postgresql/client/migrate_test.go.backup
```diff
@@ -0,0 +1,104 @@
+package client
```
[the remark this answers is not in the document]
yevgenypats: I'll do this in a follow-up PR and move migration testing to SDK as this needs to be the same for every destination.

plugins/source/digitalocean/codegen/main.go
yevgenypats: maybe we can do this in a follow-up PR as this PR is already quite huge

plugins/destination/postgresql/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Changelog
 
+<<<<<<< HEAD
```
erezrokah, line 3: You might want to fix this change

cli/cmd/sync2.go
yevgenypats: Added

plugins/source/aws/resources/services/s3/buckets_fetch.go
yevgenypats: I think I just improved the situation by knowing for sure what will be the type so if any changed I suggest to do follow-up PRs. before that we assumed it was string and it could fail in runtime. now we ensure it is always `*cqtypes.Text`

plugins/destination/postgresql/client/migrate.go
```diff
@@ -171,6 +174,9 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 	primaryKeys := []string{}
 	for i, col := range table.Columns {
 		pgType := c.SchemaTypeToPg(col.Type)
+		if pgType == "" {
+			c.logger.Warn().Str("table", table.Name).Str("column", col.Name).Msg("Column type is not supported, skipping")
```
hermanschaaf, line 178: Not sure what needs to happen here, maybe also a `continue`

plugins/destination/postgresql/client/write.go
```diff
@@ -3,127 +3,268 @@ package client
...
 	"github.com/cloudquery/plugin-sdk/specs"
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgtype"
 	"github.com/jackc/pgx/v4"
 )
 
-func (c *Client) Write(ctx context.Context, table string, data map[string]interface{}) error {
-	var sql string
-	var values []interface{}
+func CQTextArray(c *cqtypes.TextArray) pgtype.TextArray {
```
hermanschaaf, line 17: nit: two things about these functions:
 - these functions look like they don't need to be public (but I could be wrong there), and
 - I think the name should be e.g. `pgTextArray` or `toPgTextArray`, since it's converting from CQ to PG, but the current name would suggest the opposite.

plugins/destination/postgresql/client/write.go
```diff
@@ -3,127 +3,268 @@ package client
...
+				Bool:   t.Bool,
+				Status: pgtype.Status(t.Status),
+			}
+		case *cqtypes.Int8:
+			pgValues[i] = pgtype.Int8{
+				Int:    t.Int,
+				Status: pgtype.Status(t.Status),
+			}
+		case *cqtypes.Float8:
+			pgValues[i] = "0.0"
```
hermanschaaf, line 104: Maybe I'm missing something, but `Float8` looks like it still needs to be implemented

plugins/source/digitalocean/docs/tables/digitalocean_domain_records.md
```diff
@@ -12,7 +12,7 @@ The primary key for this table is **id**.
 |_cq_sync_time|Timestamp|
 |_cq_id|UUID|
 |_cq_parent_id|UUID|
-|id (PK)|String|
+|id (PK)|Int|
```
erezrokah, line 15: This is a breaking change in the DigitalOcean plugin correct?

plugins/source/azure/docs/tables/azure_cdn_custom_domains.md
```diff
@@ -4,6 +4,8 @@
 
 The primary key for this table is **id**.
 
+## Relations
+This table depends on [azure_cdn_endpoints](azure_cdn_endpoints.md).
```
erezrokah, line 8: Do we know what code change fixed this?

hermanschaaf (changes requested):

plugins/source/azure/docs/tables/azure_cdn_custom_domains.md
hermanschaaf: Yeah, there was a function in the SDK to calculate the parents of tables that wasn't recursive before, but is now. I can find a link if you're interested

plugins/destination/postgresql/client/migrate.go
```diff
@@ -79,6 +79,9 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table) erro
 	for _, col := range table.Columns {
 		columnName := pgx.Identifier{col.Name}.Sanitize()
 		columnType := c.SchemaTypeToPg(col.Type)
+		if columnType == "" {
+			c.logger.Warn().Str("table", table.Name).Str("column", col.Name).Msg("Column type not supported, skipping")
```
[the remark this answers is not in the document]
yevgenypats: Good catch

plugins/source/digitalocean/docs/tables/digitalocean_domain_records.md
yevgenypats: Yeah. though I dont think we have any usage for that plugin anyway per analytics.

hermanschaaf:
@yevgenypats LGTM, just one open question about float8 https://github.com/cloudquery/cloudquery/pull/3323#discussion_r1009644383

yevgenypats:
Closes https://github.com/cloudquery/cloudquery/issues/3172

plugins/destination/postgresql/client/write.go
yevgenypats: Good catch. Fixed

plugins/destination/postgresql/client/write.go
yevgenypats: Thanks! Changed to private and better Naming

PR #3327, opened 2022-10-31
chore: Use large runners only on `push`

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/3322.
Looks like we can't use large runners on PRs in a safe way.
I configured the following access rules:
image

But only the one on `main` works as PRs refs are based on the PR number, e.g. `refs/pull/<pr_number>/merge`.

This PR keeps the large runners only on `push` events. I'm hoping that should be sufficient to improve performance by priming the cache for all other PRs

### DIFF
--- a/.github/workflows/source_test.yml
+++ b/.github/workflows/source_test.yml
@@ -20,9 +20,9 @@ jobs:
     steps:
       - name: Check if should use large runner
         id: large-runner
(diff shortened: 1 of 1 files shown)


disq (approved):

erezrokah:
Open https://github.com/orgs/community/discussions/37892 as a follow up

PR #3329, opened 2022-10-31
doc: Add examples to configuration docs for the big plugins

shimonp21:
#### Summary

Docs for the long tail of small providers to follow.

### DIFF
--- a/plugins/source/aws/docs/configuration.md
+++ b/plugins/source/aws/docs/configuration.md
@@ -2,9 +2,9 @@
 
 ## Example
 
-This example connects a single AWS account in one region to a Postgres destination. The source spec section is described in the [Source Spec Reference](https://www.cloudquery.io/docs/reference/source-spec).
+This example connects a single AWS account in one region to a Postgres destination. The (top level) source spec section is described in the [Source Spec Reference](https://www.cloudquery.io/docs/reference/source-spec).
 
-```
+```yml
 kind: source
 spec:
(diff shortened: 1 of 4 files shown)


erezrokah (changes requested):
Maybe we could do this under [recipes](https://www.cloudquery.io/docs/recipes/aws-postgresql) and point to the readme files for all configuration properties?

The recipes are almost identical to what is suggested in this PR, and they also have the destination part so they can be easily copy pasted into a working configuration.

Additionally we could drop `registry` as it's optional

shimonp21:
I think the point is to reduce confusion caused by the nested specs - which is why I see a lot of value in having an example first-thing (We had some users on discord who put the properties in the wrong spec).

This was already done in AWS, I am just extending it to the rest of the plugins :)

I am Ok dropping the registry.

erezrokah (approved):
> I think the point is to reduce confusion caused by the nested specs - which is why I see a lot of value in having an example first-thing (We had some users on discord who put the properties in the wrong spec).

💯 Agree the current docs are confusing (see [issue](https://github.com/cloudquery/cloudquery/issues/2047) I opened a while back).

I'm not sure actually where we should put these examples. The GitHub docs are a bit hard to discover (you need to go to https://www.cloudquery.io/docs/plugins/sources, then click the plugin, then click configuration).

Giving this a second thought we can always move them if we see that we need to

PR #3330, opened 2022-10-31
fix(aws)!: Unified tag structure

### LINKED ISSUE #3280: Tags Columns in V1 aren't unified
### Describe the Bug
Some tables have tags as `list(dict)` and others are just type `dict`. ideally this would be unified as type `dict` for all tables. 

### Expected Behavior
`{"my_tag": "my_value"}` is best. Looks like the resolver looks like this in this situation: 
```
return resource.Set(c.Name, client.TagsToMap(output.Tags))
```

`[{"Key":"my_tag", "Value":"my_value"}]` is very clunky.

### CloudQuery Version
v1.3.4

### List of tables afflicted
source query: 
```
SELECT distinct cq_table, jsonb_typeof(tags) from aws_resources 
group by cq_table, tags
```

| cq_table | tag_type |
| -------- | -------- |
| aws_acm_certificates | array |
| aws_autoscaling_groups | array |
| aws_codebuild_projects | array |
| aws_dax_clusters | array |
| aws_ec2_customer_gateways | array |
| aws_ec2_ebs_snapshots | array |
| aws_ec2_ebs_volumes | array |
| aws_ec2_flow_logs | array |
| aws_ec2_hosts | array |
| aws_ec2_images | array |
| aws_ec2_instances | array |
| aws_ec2_key_pairs | array |
| aws_ec2_nat_gateways | array |
| aws_ec2_network_interfaces | array |
| aws_ec2_security_groups | array |
| aws_ec2_subnets | array |
| aws_ec2_vpc_endpoints | array |
| aws_ec2_vpc_endpoint_service_configurations | array |
| aws_ec2_vpc_peering_connections | array |
| aws_ec2_vpcs | array |
| aws_ec2_vpn_gateways | array |
| aws_efs_filesystems | array |
| aws_organizations_accounts | array |

disq:
Fixes #3280 

All tags (except one) should be JSON objects (and not JSON arrays) now.
The exception is `aws_autoscaling_groups`, which `tags` is defined as `[]TagDescription`, which is defined as:
```go
// Describes a tag for an Auto Scaling group.
type TagDescription struct {

	// The tag key.
	Key *string

	// Determines whether the tag is added to new instances as they are launched in the
	// group.
	PropagateAtLaunch *bool

	// The name of the group.
	ResourceId *string

	// The type of resource. The only supported value is auto-scaling-group.
	ResourceType *string

	// The tag value.
	Value *string

	noSmithyDocumentSerde
}
```


plugins/source/aws/codegen/recipes/iam.go
```diff
@@ -460,15 +440,10 @@ func IAMResources() []*Resource {
...
-					{
-						Name:     "tags",
-						Type:     schema.TypeJSON,
-						Resolver: `client.ResolveTags`,
-					},
 					{
 						Name:     "user_tags",
 						Type:     schema.TypeJSON,
-						Resolver: `client.ResolveTags`,
+						Resolver: `client.ResolveTags`, // FIXME this is wrong, add path
```
disq, line 446: This needs to be `client.ResolveTagField("User.Tags")` or something like that but I'm not sure about the syntax...

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -82,6 +82,26 @@ func awsNameTransformer(f reflect.StructField) (string, error) {
 	return c.ToSnake(f.Name), nil
 }
 
+func awsResolverTransformer(f reflect.StructField, path string) (string, error) {
+	if f.Type.String() == "[]types.Tag" {
```
candiduslynx, line 86: Could we stick to the reflect data instead of string representations? (basically, better option will be to define names for key/value fields and then transform the resolver accordingly)

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -82,6 +82,26 @@ func awsNameTransformer(f reflect.StructField) (string, error) {
...
+		if path == "Tags" {
+			return "client.ResolveTags", nil
+		}
+		return `client.ResolveTagField("` + path + `")`, nil
+	}
+
+	if path == "Tags" || path == "TagSet" {
+		switch f.Type.String() {
+		case "map[string]string", "map[string]*string", "map[string]interface {}", "[]types.TagDescription":
```
candiduslynx, line 95: 1. `"map[string]interface {}"` -> `"map[string]interface{}"` ?
2. maybe use reflect data instead?

candiduslynx (changes requested):

plugins/source/aws/codegen/recipes/iam.go
disq: Tested: User tags are not exposed in the `User` struct from `ListVirtualMFADevices`, so removing this column.

plugins/source/aws/codegen/recipes/base.go
disq: Each "types" will be different, so it won't be a blanket-catch-all unless we go and import all of those different `types` from different packages.

plugins/source/aws/codegen/recipes/base.go
disq: 1. This is what gives
2. It would work here as these are std types, and TagDescription is just from AutoScalingGroups... but is it necessary from a readability standpoint?

erezrokah (approved):
Nice touch with the custom resolver.

I think this should be a `fix!` as it's a breaking change (even without the column rename)

plugins/source/aws/docs/tables/aws_ec2_network_interfaces.md
```diff
@@ -40,4 +39,5 @@ The primary key for this table is **arn**.
 |source_dest_check|Bool|
 |status|String|
 |subnet_id|String|
+|tag_set|JSON|
```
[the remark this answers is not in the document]
disq: ooh yes, will fix

disq:
> I think this should be a `fix!` as it's a breaking change (even without the column rename)

JSON is JSON though... and we weren't intending these to be arrays.

erezrokah:
> JSON is JSON though... and we weren't intending these to be arrays.

Yeah, but the schema of the JSON is different correct? i.e. if someone has a query on it, that query will break?

plugins/source/aws/codegen/recipes/base.go
candiduslynx: that's right. what I meant was using the info that there're only a couple of exposed fields with well-known names (or belonging to well-known names lists) to be used as key-value pairs

disq:
> > JSON is JSON though... and we weren't intending these to be arrays.
> 
> Yeah, but the schema of the JSON is different correct? i.e. if someone has a query on it, that query will break?

yes but it will be better in the end

erezrokah:
> yes but it will be better in the end

Agree it's better but it's still breaking, and with a major version bump it's clear to users that they need to update their queries

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -82,6 +83,26 @@ func awsNameTransformer(f reflect.StructField) (string, error) {
...
+	if path == "Tags" || path == "TagSet" {
+		switch f.Type {
+		case reflect.TypeOf(map[string]string{}), reflect.TypeOf(map[string]*string{}), reflect.TypeOf(map[string]interface{}{}), reflect.TypeOf([]types.TagDescription{}):
+			// valid tag types
+		default:
+			return "", fmt.Errorf("%q field is not of type []types.Tag or acceptable map: %s", path, f.Type.String())
+		}
+	}
+
+	return `schema.PathResolver("` + path + `")`, nil
```
hermanschaaf, line 103: Maybe let's use `codegen.DefaultResolverTransformer` as the default here, so we automatically benefit from future SDK improvements of the default?

Barneyjm:
> Agree it's better but it's still breaking, and with a major version bump it's clear to users that they need to update their queries

FWIW the columns in some of these tables pre-V1 are type(Object) but in V1 are type(Array). This change would make it so that we _don't_ have breaking changes

plugins/source/aws/codegen/recipes/base.go
disq: Good idea, fixed in [`a953d78` (#3330)](https://github.com/cloudquery/cloudquery/pull/3330/commits/a953d78f26cdb5a4d433926a5e9112a18e58c112)

plugins/source/aws/docs/tables/aws_ec2_network_interfaces.md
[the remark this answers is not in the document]
disq: Fixed in  [`a953d78` (#3330)](https://github.com/cloudquery/cloudquery/pull/3330/commits/a953d78f26cdb5a4d433926a5e9112a18e58c112)

erezrokah:
> FWIW the columns in some of these tables pre-V1 are type(Object) but in V1 are type(Array). This change would make it so that we _don't_ have breaking changes

Thanks for the added context @Barneyjm. Agree this is not breaking compared to v0, but it can be considered breaking compared to the latest version of the latest AWS plugin.

We can make this clear with a note in the changelog, WDYT?

bbernays:
This doesn't help resources that have to make specific call to get tags like Neptune/DocDB/RDS: https://github.com/cloudquery/cloudquery/blob/776d849c3594f4646575ba44b3d764e3b13db1c2/plugins/source/aws/codegen/recipes/neptune.go#L26-L29

hermanschaaf (approved):

candiduslynx (approved):

yevgenypats (approved):

Issue #3344, opened 2022-10-31
chore(ci): Plugin version updates PR are often conflicted


erezrokah:
Since we update plugins versions in our docs after releasing them, the PRs to update the plugins are often conflicted:
![image](https://user-images.githubusercontent.com/26760571/199017662-f06659ad-5f83-42fa-957e-6346273803e6.png)

As a result we need to manually resolve the conflicts.

## Possible solution:
1. Structure the `sources.md` file in a way that prevents conflicts, see https://github.com/cloudquery/cloudquery/blob/0e9994fba7dd57e04bce9a93609f07d6db8f32e3/.release-please-manifest.json#L5
2. Use renovate to manage the updates as it can handle conflicts

Option 2. will create a delay as we run renovate once per hour

erezrokah:
Fixed in https://github.com/cloudquery/cloudquery/pull/3451

PR #3367, opened 2022-10-31
chore(website): Add sources table fillers

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3344 in a very ugly way (solution 1.)

The other option is to use renovate to do it via a regex and let it handle any conflicts

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -4,6 +4,8 @@
   "private": true,
   "description": "CloudQuery website and docs",
   "scripts": {
+    "predev": "sed -i '' -e ':a' -e 'N' -e '$!ba' -e 's/TABLE-FILLER\\n//g' 'pages/docs/plugins/sources.md'",
+    "prebuild": "sed -i -e ':a' -e 'N' -e '$!ba' -e 's/TABLE-FILLER\\n//g' 'pages/docs/plugins/sources.md'",
     "dev": "next",
     "start": "next start",
     "build": "next build",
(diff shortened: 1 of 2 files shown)


yevgenypats:
Yeah, that seems to build more tech debt on top of tech debt. Maybe we can use nextra parameters instead of changing the files on each release?

erezrokah:
> Maybe we can use nextra parameters instead of changing the files on each release?

`sources.md` needs to match https://github.com/cloudquery/cloudquery/blob/66e677727114f8764433006271aa891026b19f9f/.github/workflows/release_plugin.yml#L99

so we either keep the structure or find another (e.g. renovate) to do the updates

yevgenypats:
> > Maybe we can use nextra parameters instead of changing the files on each release?
> 
> `sources.md` needs to match
> 
> https://github.com/cloudquery/cloudquery/blob/66e677727114f8764433006271aa891026b19f9f/.github/workflows/release_plugin.yml#L99
> 
> so we either keep the structure or find another (e.g. renovate) to do the updates

Maybe Im missing something and worth doing short live session. but can't we use ${LATEST_AWS_VERSION} in the markdown and turn it into `.mdx`?

yevgenypats (changes requested):
I would like to discuss a solution without search and replace bash script. proposed here - https://github.com/cloudquery/cloudquery/pull/3451

erezrokah:
> I would like to discuss a solution without search and replace bash script. proposed here - #3451

Cool, that should work much better 🚀  Nice one 🎸 I'll comment on that PR

PR #3371, opened 2022-10-31
feat: Add CSV destination plugin

yevgenypats:
Add new CSV destination plugin. I used it both to battle test our new cloudquery type [system](https://github.com/cloudquery/plugin-sdk/pull/320) but also can be quite useful to scale out ETL workloads and then load CSV in batch to athen/bigquery and other datalakes/data warehouses


plugins/destination/csv/client/migrate.go
```diff
@@ -0,0 +1,77 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func (c *Client) migrate(tables schema.Tables) error {
+	for _, t := range tables {
+		filePath := path.Join(c.csvSpec.Directory, t.Name+".csv")
+		if _, err := os.Stat(filePath); err == nil {
```
disq, line 29: call/use as `migrateTable(t, filePath)` maybe? would solve the all the linting issues.

plugins/destination/csv/client/write.go
```diff
@@ -0,0 +1,114 @@
...
+				c.logger.Error().Err(err).Msgf("failed to close file %s", fileName)
+			}
+			delete(c.writers, t)
+		}
+	}
+}
+
+func (c *Client) startWrite(tables schema.Tables) error {
+	// we want to have the ability to rollback in case of an error so we dont have zombie files
+	var initializedTables []string
```
disq, line 31: why not keep `initializedTables` as a map inside `Client`? could be useful in `endWrite` too.

yevgenypats: this wont work as it can't be access concurrently and will introduce lock contention

yevgenypats:
Oh no. I ruined this PR.

PR #3434, opened 2022-11-01
feat: Add CSV Destination plugin

yevgenypats:
I rebased incorrectly previous PR.


plugins/destination/csv/client/migrate.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, t := range tables {
+		filePath := path.Join(c.csvSpec.Directory, t.Name+".csv")
+		//nolint:gocritic,revive
+		if _, err := os.Stat(filePath); err == nil {
+			f, err := os.Open(filePath)
+			if err != nil {
+				return err
+			}
+			//nolint:revive
+			defer f.Close()
```
disq, line 36: I would wrap the whole block inside a fn (anonymous or a separate function) and remove the `nolint`. No reason to keep files open more than necessary.

disq (changes requested):

plugins/destination/csv/client/client.go
```diff
@@ -0,0 +1,137 @@
...
+
+type tableWriter struct {
+	writer *csv.Writer
+	file   *os.File
+	count  uint64
+}
+
+func New(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (plugins.DestinationClient, error) {
+	if spec.WriteMode != specs.WriteModeAppend {
+		return nil, fmt.Errorf("csv destination only supports append mode only")
```
hermanschaaf, line 67:
```suggestion
		return nil, fmt.Errorf("csv destination only supports append mode")
```

plugins/destination/csv/client/client_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if client == nil {
+		t.Fatal("client is nil")
+	}
+	if err := client.Close(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = New(ctx, getTestLogger(t), specs.Destination{})
+	if err.Error() != "csv destination only supports append mode only" {
```
hermanschaaf, line 39:
```suggestion
	if err.Error() != "csv destination only supports append mode" {
```

plugins/destination/csv/main.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/cloudquery/cloudquery/plugins/destination/csv/client"
+	"github.com/cloudquery/cloudquery/plugins/destination/csv/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/plugins"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const (
+	// TODO: add sentry DSN
+	sentryDSN = ""
```
hermanschaaf: Let's remember to do this before merging ☝️

hermanschaaf (changes requested):

plugins/destination/csv/main.go
yevgenypats: Done

plugins/destination/csv/client/migrate.go
yevgenypats: Good idea.

hermanschaaf (approved):

Issue #1906, opened 2022-09-20
Support for AWS KMS Key Grants


jsonpr:
### Why do you need this resource or plugin? Please describe.
To add additional context and awareness of access to AWS KMS Keys.

### Describe the solution you'd like
An ability to query and find active KMS Key Grants on KMS Keys, as this is one possible mechanism for KMS Key Access.

https://docs.aws.amazon.com/kms/latest/developerguide/grants.html

disq:
Looks like `ListGrants`

Issue #3067, opened 2022-10-19
EC2 ARN issues


nick123pig:
### Describe the Bug
I submitted issue #2986 yesterday for EC2 instances in a rush, but I failed to inspect the full EC2 arns that are misconfigured. My results here are a result of manual inspection, but I wonder if there is a more programatic analysis that can be done.

The following ARNs seem to be out of alignment according to [the documentation](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2.html#amazonec2-resources-for-iam-policies)

> Name Issues

- `resolveFlowLogArn` in plugins/source/aws/resources/services/ec2/flow_logs_fetch.go the ARN Resource should be `vpc-flow-log`
- `resolveImageArn` in plugins/source/aws/resources/services/ec2/images_fetch.go the ARN Resource should be `image`
- `resolveInternetGatewayArn` in plugins/source/aws/resources/services/ec2/internet_gateways_fetch.go the ARN Resource should be `internet-gateway` 
- `resolveNatGatewayArn` in plugins/source/aws/resources/services/ec2/nat_gateways_fetch.go the ARN Resource should be `natgateway`
- `resolveRouteTableArn` in plugins/source/aws/resources/services/ec2/route_tables_fetch.go the ARN Resource should be  `route-table`
- `resolveSecurityGroupArn` in  plugins/source/aws/resources/services/ec2/security_groups_fetch.go the ARN Resource should be `security-group`
- `resolveNetworkAclArn` in plugins/source/aws/resources/services/ec2/network_acls_fetch.go the ARN Resource should be `network-acl`
- `resolveNetworkInterfaceArn` in plugins/source/aws/resources/services/ec2/network_interfaces_fetch.go the ARN Resource should be `network-interface`
- `resolveVpcEndpointServiceArn` in plugins/source/aws/resources/services/ec2/vpc_endpoint_services_fetch.go the ARN Resource should be `vpc-endpoint-service`
- `resolveVpcEndpointArn` in plugins/source/aws/resources/services/ec2/vpc_endpoints_fetch.go the ARN Resource should be `vpc-endpoint`
- `resolveVpcPeeringConnectionArn` in plugins/source/aws/resources/services/ec2/vpc_peering_connections_fetch.go the ARN Resource should be `vpc-peering-connection`
- `resolveVpnGatewayArn` in plugins/source/aws/resources/services/ec2/vpn_gateways_fetch.go the ARN Resource should be `vpn-gateway`

[...]

PR #3441, opened 2022-11-01
feat(aws): KMS Key Grants, better key fetch

disq:
Implements #1906


plugins/source/aws/codegen/recipes/kms.go
```diff
@@ -53,8 +53,23 @@ func KMSResources() []*Resource {
 						IgnoreInTests: true,
 					},
 				}...),
-		},
-	}
+			Relations: []string{"Grants()"},
+		}, {
+			SubService:  "grants",
```
disq, line 58: `grants` or `key_grants`? Not sure.

plugins/source/aws/resources/services/kms/grants_fetch.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+		res <- response.Grants
+	}
+	return nil
+}
+
+func resolveGrantARN(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	cl := meta.(*client.Client)
+	g := resource.Item.(types.GrantListEntry)
+	return resource.Set(c.Name, cl.ARN("kms", "grant", aws.ToString(g.KeyId), aws.ToString(g.GrantId)))
```
disq, line 33: Another made up ARN.

disq:
@jsonpr Please check if the PR satisfies the requirements (which were a little vague)

plugins/source/aws/codegen/recipes/kms.go
hermanschaaf: it's called `ListGrants` (https://docs.aws.amazon.com/kms/latest/APIReference/API_ListGrants.html) so let's go with `grants` 👍

plugins/source/aws/resources/services/kms/grants_fetch.go
hermanschaaf: Probably worth creating a grant in a test account just to check that the ARN matches

disq: The grants don't have arns so it's made up just like few others. Nothing to check in a test account.

hermanschaaf: If it doesn't match anything real on the AWS side, then I think we shouldn't be adding the ARN column? I wasn't aware we're doing this in other places as well

disq: @hermanschaaf check "Resource Issues" in [#3067 ](https://github.com/cloudquery/cloudquery/issues/3067)

hermanschaaf: Okay but most of those do exist, they're just wrong. And the ones that don't exist should be removed, let's not add more non-existent ones 🤷  If the issue is that we need a PK for upserting, we can use the separate columns as a composite PK, which I guess would be account_id, region_id, key_id, grant_id in this case.

disq: Look in the "Resource Issues" not the whole issue description. Instance statuses etc.

disq: removed in [`b68dd67` (#3441)](https://github.com/cloudquery/cloudquery/pull/3441/commits/b68dd671c66e3fd147891b4ded8a9d5178351e8c)

hermanschaaf: @disq I read the issue, and saw 3 instances of made-up ARNs discovered there. I don't think that's a good enough reason to add made-up ARNs as a rule now, so I'm not sure what you're saying. Let's rather remove any made-up ARNs that we become aware of. Unless there is some other reason to add it that I'm missing, that is. But even even if we do need a unique identifier for some reason, we shouldn't call it `arn` if it doesn't match something on the Amazon side.

disq: Also #3466

plugins/source/aws/codegen/recipes/kms.go
jsonpr: Agreed with `grants`

plugins/source/aws/resources/services/kms/grants_fetch.go
jsonpr: I don't see an ARN for KMS Grants either.  The GrantID looks to be a unique identifier as specified by AWS in their documentation - so we can use that as the unique identifier: https://docs.aws.amazon.com/kms/latest/APIReference/API_GrantListEntry.html.  

Will test later to check for an ARN.

plugins/source/aws/docs/tables/aws_kms_grants.md
```diff
@@ -0,0 +1,27 @@
...
+This table depends on [aws_kms_keys](aws_kms_keys.md).
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|account_id (PK)|String|
```
jsonpr, line 17: Will the account_id always match the issuing account?

plugins/source/aws/resources/services/kms/grants_fetch.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/service/kms"
+	"github.com/aws/aws-sdk-go-v2/service/kms/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchKmsGrants(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	k := parent.Item.(types.KeyMetadata)
+	config := kms.ListGrantsInput{KeyId: k.KeyId}
```
jsonpr, line 14: List Grants takes either Key ID or Key ARN - this should be ok as if we're listing grants within each AWS Account, the key ID will work.  The key ARN is only required if we're trying to find grants in a different AWS account.  

I don't think that's something we're going to do while listing grants for keys.

plugins/source/aws/docs/tables/aws_kms_grants.md
disq: @jsonpr Yes that is assumed, it's resolved from the `Client` and since we use the key ID in the `ListGrants` call...

disq:
This panics for me, investigating.

plugins/source/aws/codegen/recipes/kms.go
disq: Reopening this... Traditionally if a resource depends on another one we prefix it with the other's name, so `aws_kms_key_grants` makes more sense and helps communicate that it requires `aws_kms_keys`.

jsonpr: I'm ok with `aws_kms_key_grants` given the explanation above.  It does provide more clarity.

disq: Went with `aws_kms_key_grants` in [`e4416a1` (#3441)](https://github.com/cloudquery/cloudquery/pull/3441/commits/e4416a1d1c468b2811582a9aaf45bd9a9aae4129)

hermanschaaf: Sounds good 👍

hermanschaaf (approved):
👍

disq:
Looks like `KeyId` inside `types.GrantListEntry` is returned in an ARN format when queried from `ListGrants`. So KeyId is ARN of the key: `aws_kms_key_grants.key_id = aws_kms_keys.arn`. Weird. Should we address that?

PR #3447, opened 2022-11-01
doc: Add examples to readmes, and update stale docs

shimonp21:


plugins/source/heroku/client/spec.go
```diff
@@ -13,6 +13,8 @@ import (
 type Spec struct {
 	Token string `yaml:"token,omitempty" json:"token"`
 
+	// These don't actually work (throw a "json unknown field" error) - but not a high-piority to fix this,
```
hermanschaaf, line 16: I think it'll work if we change `yaml` to `json` in all the cases here. But yeah, not a high priority.

hermanschaaf: In fact, maybe we should just remove these options until further notice?

hermanschaaf (approved):
Very nice! ✨

plugins/source/heroku/client/spec.go
shimonp21: Will do in a follow up PR

hermanschaaf: Yup, not a blocker for this PR

PR #3449, opened 2022-11-01
feat: Add unique constraint on _cq_id column

hermanschaaf:
This adds a `unique not null` contraint on the internal `_cq_id` column, both when migrating an existing database and when creating a new database. 

It also fixes a bug where we didn't check if constraints changed before, if the column was new or if the column type changed.

### DIFF
--- a/plugins/destination/postgresql/client/client.go
+++ b/plugins/destination/postgresql/client/client.go
@@ -24,11 +24,6 @@ type Client struct {
 	batchSize           int
 }
 
-type pgTablePrimaryKeys struct {
-	name    string
-	columns []string
-}
-
 type pgColumn struct {
 	name string
 	typ  string
@@ -181,15 +176,6 @@ func (c *Client) getPgTableColumns(ctx context.Context, tableName string) (*pgTa
 		return nil, err
 	}
 	return &tc, nil
-}
-
-func (c *pgTablePrimaryKeys) columnExist(column string) bool {
-	for _, col := range c.columns {
-		if col == column {
-			return true
-		}
-	}
-	return false
 }
 
 func (c *pgTableColumns) getPgColumn(column string) *pgColumn {
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -21,16 +21,6 @@ JOIN   pg_attribute a ON a.attrelid = i.indrelid
 WHERE  i.indrelid = $1::regclass
 AND    i.indisprimary;
 `
-	// Modification of https://stackoverflow.com/a/27752061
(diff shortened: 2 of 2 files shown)


plugins/destination/postgresql/client/migrate.go
```diff
@@ -113,11 +129,24 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table) erro
 			if _, err := c.conn.Exec(ctx, sql); err != nil {
 				return fmt.Errorf("failed to add column %s on table %s: %w", col.Name, table.Name, err)
 			}
-		default:
```
hermanschaaf, line 116: I think this `default` case was a bug, and the PK check really should always be done, not only when there are no existence / type mismatches.

hermanschaaf:
> I think we can get away without checking if a column is unique. Just add unique to _cq_id if we have primary key. this will ensure less issues/race conditions and so on.

@yevgenypats Sorry, I'm not sure I follow... are you saying we shouldn't add a unique constraint to `_cq_id` on existing tables, only on new ones?

yevgenypats:
> > I think we can get away without checking if a column is unique. Just add unique to _cq_id if we have primary key. this will ensure less issues/race conditions and so on.
> 
> @yevgenypats Sorry, I'm not sure I follow... are you saying we shouldn't add a unique constraint to `_cq_id` on existing tables, only on new ones?

Yeah exactly. no need to add on existing. postgresql driver is complex enough.

hermanschaaf:
@yevgenypats Okay, I'll remove the code related to migrating tables, should be a lot simpler then.

plugins/destination/postgresql/client/client.go
```diff
@@ -24,11 +24,6 @@ type Client struct {
 	batchSize           int
 }
 
-type pgTablePrimaryKeys struct {
```
hermanschaaf, line 27: I replaced this with a map in the one function where we were using it, seems a bit simpler. We can add this back later if we need it.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -181,6 +183,10 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		}
 		columnName := pgx.Identifier{col.Name}.Sanitize()
 		fieldDef := columnName + " " + pgType
+		if col.Name == "_cq_id" {
+			// _cq_id column should always have a "unique not null" constraint
+			fieldDef += " UNIQUE NOT NULL"
```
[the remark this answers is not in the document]
hermanschaaf: Yes, you can have both

yevgenypats (approved):

PR #3450, opened 2022-11-01
feat(aws): AppRunner Resources: VPC Connector, VPC Ingress Connection And Autotscaling

bbernays:


disq (commented):
LGTM with single nit (`autos`)

disq (approved):

plugins/source/aws/resources/services/apprunner/vpc_ingress_connection.go
```diff
@@ -0,0 +1,73 @@
...
+				Resolver: schema.PathResolver("VpcIngressConnectionArn"),
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
+				},
+			},
+			{
+				Name:     "account_id",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("AccountId"),
+			},
```
bbernays, line 35: Do not merge until I verify what account this is and if it can be an outside account.

plugins/source/aws/codegen/recipes/apprunner.go
```diff
@@ -28,6 +28,62 @@ func ApprunnerResources() []*Resource {
...
+						Name:     "arn",
+						Type:     schema.TypeString,
+						Resolver: `schema.PathResolver("VpcConnectorArn")`,
+						Options:  schema.ColumnCreationOptions{PrimaryKey: true},
+					},
+				}...),
+		}, {
+			SubService:          "vpc_ingress_connection",
+			Struct:              &types.VpcIngressConnection{},
+			Description:         "https://docs.aws.amazon.com/apprunner/latest/api/API_VpcIngressConnection.html: account_id has been renamed to `source_account_id` to avoid conflict with the account_id column that indicates what account this was synced from.",
```
disq, line 68: Problem with description

disq (changes requested):

plugins/source/aws/codegen/recipes/apprunner.go
bbernays: What is the issue? Normally I would put this information in the column description field but we don't use that anywhere else

hermanschaaf: Yeah, I think it's fair to put it in the table description. Let's maybe just put the note in a separate paragraph, like:

```
`https://docs.aws.amazon.com/apprunner/latest/api/API_VpcIngressConnection.html

Notes:
 - `account_id` has been renamed to `source_account_id` to avoid conflict with the `account_id` column that indicates what account this was synced from.
```

WDYT?

disq (approved):

plugins/source/aws/codegen/recipes/apprunner.go
bbernays: Fixed, thanks for the suggestion... Only change is that you can't have a backtick in a multiline string 😢

plugins/source/aws/docs/tables/aws_apprunner_vpc_ingress_connection.md
```diff
@@ -0,0 +1,28 @@
+# Table: aws_apprunner_vpc_ingress_connection
+
+https://docs.aws.amazon.com/apprunner/latest/api/API_VpcIngressConnection.html
+
+			Notes:
```
hermanschaaf, line 5: The indentation is a little off here :)

hermanschaaf (approved):
LGTM 👍

hermanschaaf (approved):
👍 👍

PR #3451, opened 2022-11-01
chore: Add import versions instead of search/replace

yevgenypats:
@erezrokah I think we should try something like this PR instead of https://github.com/cloudquery/cloudquery/pull/3367 otherwise it will be very hard to control complexity with all the search and replace scripts.

Also suggest to use react-syntax-highlighter components instead of the search and replace - I think this works just need to small bit of css to looks nicer and similar to nextra built-in css with shiki library.

Hopefully this will solve the whole search and replace flaky thing


erezrokah:
@yevgenypats a follow up:
1. I cleaned up the PR a bit to make it more generic
2. I removed the changes to `yaml` configuration files as I find it a bit complex to write MDX every time we want to add an example
3. Added in the Vercel:
image

For 2. I think we could use renovate for that, see example:
https://github.com/cloudquery/setup-cloudquery/pull/86/files

I actually renovate will work better here as it can group all updates into a single PR. This is useful when we release a Postgres version that is not compatible with other plugins (like we just had)

erezrokah (approved):

PR #3454, opened 2022-11-02
doc: Remove duplicate source plugin entries

hermanschaaf:
Seems like there was maybe a bad merge at some point


yevgenypats:
Yeah, this should help avoid this in the future - https://github.com/cloudquery/cloudquery/pull/3451

erezrokah (approved):

erezrokah:
Let fix the duplicates first then handle the long term solution via https://github.com/cloudquery/cloudquery/pull/3451

PR #3455, opened 2022-11-02
fix: Delete broken spec parameters

shimonp21:

### DIFF
--- a/plugins/source/heroku/client/spec.go
+++ b/plugins/source/heroku/client/spec.go
@@ -1,8 +1,6 @@
 package client
 
 import (
-	"time"
-
 	"github.com/googleapis/gax-go/v2"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


plugins/source/heroku/client/spec.go
```diff
@@ -31,36 +20,18 @@ type BackoffSettings struct {
 
 func (c Spec) ClientOptions() []option.ClientOption {
```
erezrokah, line 21:
```suggestion
func (Spec) ClientOptions() []option.ClientOption {
```

erezrokah (approved):

hermanschaaf (approved):

yevgenypats (approved):

plugins/source/heroku/client/spec.go
shimonp21: removed the 'c' completely since '_' also complained about being unused :)

PR #3458, opened 2022-11-02
fix: Properly set default telemetry level in cobra

hermanschaaf:
Finally figured out how to set default values without having them be affected by environment variables. Should help avoid https://github.com/cloudquery/cloudquery/pull/3169#discussion_r1001608633

This means we won't need this change: https://github.com/cloudquery/cloudquery/pull/3180


disq (approved):

erezrokah (approved):
Nice 🚀

yevgenypats (approved):

PR #3471, opened 2022-11-02
feat: Add SQLite destination plugin

yevgenypats:
Fixes https://github.com/cloudquery/cloudquery/issues/4136

Still WIP. Im testing out our type system and destination API and adding some more  low hanging fruit destination plugins in the meanwhile that can open the door for convenient data exploration without database locally.


.github/styles/Vocab/Base/accept.txt
```diff
@@ -128,4 +128,5 @@ write_mode
 skip_tables
 subfolder
 recurse
-CCPA's
\ No newline at end of file
+CCPA's
+sqlite
```
erezrokah, line 132:
```suggestion
SQLite
```
These are case sensitive, so I think we want `SQLite` as that's the name of the product

plugins/destination/sqlite/README.md
```diff
@@ -0,0 +1,11 @@
+# CloudQuery SQLite Destination Plugin
+
+This destination plugin let's you sync data from a CloudQuery source to a sqlite. This can be useful for local data exploration as no db/service is required.
```
erezrokah, line 3:
```suggestion
This destination plugin let's you sync data from a CloudQuery source to an SQLite database. This can be useful for local data exploration as no db/service is required.
```

yevgenypats:
@erezrokah if you can help me take it from here (we can also have a short call on this if any issues).

I've tested now the `.goreleaser` file with [golang-cross docker](https://github.com/cloudquery/golang-cross/pkgs/container/golang-cross)

`docker pull ghcr.io/cloudquery/golang-cross:v10.0.0`

I tested it via `Makefile` but I guess we can use in github actions the `runs-on` so this should also keep workflows similar to each other. 

Important note - this docker only works on x86 machine (so it should work seamlessly on linux machine) i.e it wont be possible to test it locally on an arm64 mac (I just used a GCP instance to test it and make sure everything compiles and so on).

erezrokah:
> @erezrokah if you can help me take it from here (we can also have a short call on this if any issues).

Sure I'll take it from here

erezrokah (approved):

PR #3477, opened 2022-11-02
feat(gcp)!: Move GKE cluster reource to google.golang.org/api/container/v1

candiduslynx:
Basically, it moves the resource implementation from purely gRPC-based to the SDK-based instead.
This is a breaking change as `status` field changes type (now it's not `Int` but `String` instead).
Additionally, this PR adds `cost_management_config` field (`JSON`).

Column changes
| column | old type | new type |
| --- | --- | --- |
| `cost_management_config` | | `JSON` |
| `status` | `Int` | `String` |


disq (approved):

candiduslynx:
google.golang.org/api is an auto-generated code, closing this now (cloud.google.com/go/container is correct)

PR #3486, opened 2022-11-02
chore(ci): Log only pending actions instead of all actions

erezrokah:
#### Summary

Update the "wait-for-all" action to only print the pending actions on timeout to make it easier to understand what was missing.

### DIFF
--- a/plugins/source/azure/go.mod
+++ b/plugins/source/azure/go.mod
@@ -3,7 +3,7 @@ module github.com/cloudquery/cloudquery/plugins/source/azure
 go 1.19
 
 require (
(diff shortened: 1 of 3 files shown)


plugins/source/azure/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/cloudquery/cloudquery/plugins/source/azure
 go 1.19
 
 require (
-	github.com/Azure/azure-sdk-for-go v61.6.0+incompatible
+	github.com/Azure/azure-sdk-for-go v67.0.0+incompatible
```
erezrokah, line 6: I'll revert this, it's here only to the test the workflow

hermanschaaf (approved):

PR #3491, opened 2022-11-03
chore(docs): Align gcp+k8s recipe with other recipes

erezrokah:
#### Summary

Aligns this recipe with the format of others:
1. Add the `# latest version of` comment so the find replace works (for now we need it)
2. `connection_string: ${PG_CONNECTION_STRING}` instead of `CQ_PG_USER` and `CQ_PG_PASS`
3. Use `bash` for code block as that's what Nextra (our build framework) supports
4. Remove `write-mode`, not sure it's needed


website/pages/docs/recipes/gcp-k8s-postgresql.md
```diff
@@ -5,74 +5,82 @@ kind: source
...
+
+```bash
 gcloud components update
 ```
 
 Force credential update:
-```
+
+```bash
 gcloud container clusters get-credentials {$CLUSTER_NAME}
```
erezrokah, line 84: Not sure if this is correct or it should just be `$CLUSTER_NAME` or `<CLUSTER_NAME>`

shimonp21 (approved):

PR #3494, opened 2022-11-03
chore(aws): Autogenerate service interfaces and other common structures

hermanschaaf:
This removes several error-prone steps for contributing new AWS resources and replaces them with code generation. 

## Before
 - service interfaces had to be created manually by copy-pasting signatures from the AWS SDK
 - the `go:generate mockgen` comment had to be added manually
 - `go generate` had to be run separately to generate mocks
 - the service had to be added in `client/client.go`
 - the service had to be instantiated in `initServices`
 - the table had to be registered in `resources/plugin/tables.go`

## Now
 - a Client struct can be added to `codegen/services/clients.go`
 - `make gen` does all of the previously manual steps for you

## Notes for reviewers

This PR touches a lot of files, because I decided it would be better to generate consistent names than maintain a list of names just for Go-style capitalization. It's almost entirely renaming of variables, there are no functional changes.


erezrokah (approved):
This looks amazing 🚀

scripts/test-unit.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
yevgenypats, line 1: I think this file is not needed it's an old one and we don't use it anymore to detect what was changed as we use github built-in functionally

yevgenypats (changes requested):
Looks great! I think the naming change + the codegen change made it a bit hard to review but hopefully didn't miss anything important. There are couple of small improvements I think we should do, mainly to avoid git conflicts and readability.

scripts/test-unit.sh
hermanschaaf: I added it back because we still reference it in the makefile (`make test-unit`) and it is useful for running all the tests locally. If we have another way to run tests locally though then happy to replace it with that?

plugins/source/aws/client/services.go
```diff
@@ -82,913 +83,2607 @@ import (
...
+	Transfer                  TransferClient
+	Waf                       WafClient
+	Wafregional               WafregionalClient
+	Wafv2                     Wafv2Client
+	Workspaces                WorkspacesClient
+	Xray                      XrayClient
+}
+
+//go:generate mockgen -package=mocks -destination=./mocks/accessanalyzer.go . AccessanalyzerClient
+type AccessanalyzerClient interface {
```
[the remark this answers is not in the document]
hermanschaaf: Good idea, will do this in a follow-up shortly

yevgenypats (approved):

PR #3496, opened 2022-11-03
chore(ci): Exclude `mailchimp.com` from broken links checker

erezrokah:
#### Summary

Another flaky host


hermanschaaf (approved):
I feel like something might be wrong with our check, rather than these hosts 🤔

erezrokah:
I think most sites probably have some bot protection

PR #3546, opened 2022-11-03
fix(aws-code-gen): Prefix service name when filtering relations from table list

erezrokah:
…table list

#### Summary

We're missing a bunch of top level tables, as if there's a relation function named "Instances" it will also filter a top level "Instances" service.

See tables for v4.0.0:
https://github.com/cloudquery/cloudquery/blob/plugins-source-aws-v4.0.0/plugins/source/aws/resources/plugin/tables.go#L145

See tables for v4.2.1:
https://github.com/cloudquery/cloudquery/blob/plugins-source-aws-v4.2.1/plugins/source/aws/resources/plugin/tables.go#L146


hermanschaaf (approved):
Ah, nice catch! 🙇

PR #3602, opened 2022-11-04
feat: AppRunner add support for Connections

bbernays:


bbernays:
> LGTM; just wondering if any of these AppRunner resources can have tags? From the example here https://docs.aws.amazon.com/apprunner/latest/api/API_ListTagsForResource.html it seems like Services do, at least

Good catch... Need to look into this a bunch as it is not clear which resources support tags and which don't... Have created an issue to track this work (#3604)

hermanschaaf (approved):

PR #3497, opened 2022-11-03
chore: Improve docs recipes

yevgenypats:
Streamline our recipes. As our destination grows we cannot have destination paired with source recipes as it will grow exponentially.

PR #3606, opened 2022-11-04
chore: Split recipes to sources and destinations

yevgenypats:
Follow-up instead of https://github.com/cloudquery/cloudquery/pull/3497 due to conflicts


yevgenypats:
merging to avoid future conflicts. I think the broken link should be solved once this is merged due to redirects, if not will do follow-up PR

erezrokah:
> I think the broken link should be solved once this is merged due to redirects, if not will do follow-up PR

Follow up on https://github.com/cloudquery/cloudquery/pull/3611.

Vercel redirects should work for deploy previews so I believe the broken link check should reflect that as well

Issue #3628, opened 2022-11-06
feat(CLI): The CLI should verify all plugins use the same gRPC protocol version


erezrokah:
## Problem

Given the following configuration:

```yml
kind: source
spec:
  name: aws
  version: v3.8.0
  destinations: [postgresql]
  path: cloudquery/aws
---
kind: destination
spec:
  name: postgresql
  version: v1.6.0
  spec:
    connection_string: postgresql://postgres:pass@localhost:5432/postgres
```

The CLI fails with a confusing error:

```bash
Starting migration for: aws -> [postgresql]
Migration completed successfully.
Starting sync for: aws -> [postgresql]
Error: failed to write for aws->postgresql: failed to call Write.Send: EOF
exit status 1
```

The reason is that the AWS plugin in the config uses protocol version 1, while the PostgreSQL plugin uses version 2.

## Proposed solution

The CLI should verify all plugins use the same protocol version and if not prompt users to update the relevant ones

yevgenypats:
This is now will be automatically solved on gRPC side when we will upgrade sdk to the next version and update https://github.com/cloudquery/plugin-sdk/blob/main/internal/pb/source.proto#L3 protobuf package name

PR #3657, opened 2022-11-06
feat(cli): Validate plugins gRPC protocol version

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3628

Examples:
image

image


cli/cmd/sync.go
```diff
@@ -257,3 +261,31 @@ func syncConnectionV1(ctx context.Context, cqDir string, sourceClient *clients.S
...
+		destPluginVersion, err := destClient.Version(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get destination version: %w", err)
+		}
+
+		destProtocolVersion, err := destClient.GetProtocolVersion(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get protocol version from destination %q: %w", destPluginName+"@"+destPluginVersion, err)
+		}
+		if destProtocolVersion != expectedProtocolVersion {
```
erezrokah, line 282: Initially I thought about trying to be more specific (e.g. telling the user to update only the destination or only the source), but that probably has a ton of edge cases. For example what if we have protocol version 3, then we need to start to check ranges 🙃

disq (approved):

yevgenypats (changes requested):
I think let's wait with that. it should be done on the server side where there are still few culprits before we can take care of that.

Mostly we need to start versioning SDK with v1, v2 and making protobuf versioned as well (the package name in prtobouf) - this way v1 will never be able to connect to an uncompatible version.

erezrokah:
> I think let's wait with that. it should be done on the server side where there are still few culprits before we can take care of that.
> 
> Mostly we need to start versioning SDK with v1, v2 and making protobuf versioned as well (the package name in prtobouf) - this way v1 will never be able to connect to an uncompatible version.

We can always remove this logic when we have server side validation. This is mostly needed for users updating to the latest CLI version with the type system changes (we've seen reports on Discord with the error mentioned in the bug)

erezrokah:
Also the tests are blocked by https://github.com/cloudquery/plugin-sdk/issues/351 but we can comment them

erezrokah:
PR to fix the Windows bug https://github.com/cloudquery/plugin-sdk/pull/352

yevgenypats:
Im just working on this atm in my sdk pr https://github.com/cloudquery/plugin-sdk/pull/348 so this will create more work for me reviewing and then removing the logic. let's hold off

yevgenypats:
Let's close this?

PR #3727, opened 2022-11-07
fix: Update SDK to v0.13.23

yevgenypats:
update SDK to v0.13.23

Also, This defaults unset columns to NULL. we might also want to take care of that on the SDK side given I dont think any destination want to distinguish between null and unset as it's the same thing


hermanschaaf (approved):

erezrokah:
This should be merged as a `fix`

PR #3729, opened 2022-11-07
fix: Update CSV plugin to v0.13.23

yevgenypats:
Updating CSV to v0.13.23


hermanschaaf (approved):

erezrokah:
This should be merged as a `fix`

PR #3743, opened 2022-11-07
fix(deps): Update SDK to v0.13.23

erezrokah:


plugins/source/aws/client/helpers_test.go
```diff
@@ -66,11 +66,10 @@ func TestResolveARN(t *testing.T) {
 			col := schema.Column{Name: tt.columnName}
 			client := Client{Region: "region", Partition: "aws"}
 			err := resolver(context.Background(), &client, tt.resource, col)
-			expectedText := &schema.Text{}
-			_ = expectedText.Set(tt.want)
 
-			require.Equal(t, tt.resource.Get(tt.columnName), expectedText)
-			require.Equal(t, err != nil, tt.wantErr)
+			actual := tt.resource.Get(tt.columnName)
+			require.Equal(t, tt.want, actual)
```
erezrokah, line 71: `require.Equal(t require.TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{})` is the signature (expected should come first).

This is needed for the relevant error message to be printed when the assertion fails (e.g. wanted <x> but got <y>)

yevgenypats (approved):

PR #3841, opened 2022-11-07
chore(website): Update AWS Recipe

bbernays:


website/pages/docs/recipes/sources/aws.md
```diff
@@ -9,5 +9,22 @@ spec:
...
+  skip_tables:
+    - aws_ec2_vpc_endpoint_services # this resource includes services that are available from AWS as well as other AWS Accounts
+    - aws_docdb_cluster_parameter_groups
+    - aws_docdb_engine_versions
+    - aws_ec2_instance_types
+    - aws_elasticache_engine_versions
+    - aws_elasticache_parameter_groups
+    - aws_elasticache_reserved_cache_nodes_offerings
+    - aws_elasticache_service_updates
+    - aws_neptune_cluster_parameter_groups
```
disq, line 24: add `aws_lambda_runtimes` as well?

bbernays: I agree that it is a configuration rather than a resource, but it also doesn't require an API call

erezrokah (approved):
Doing this in the docs first makes sense.
If this works as a good default we could change it in the plugin's code

disq (approved):

yevgenypats (approved):

PR #3904, opened 2022-11-08
chore(codegen): Replace CloudQueryServiceName with function

hermanschaaf:
This removes one parameter from codegen for resolvers and mocks. 

Service name should now be consistent enough that capitalizing the first letter should work in 99% of cases. The service name should almost always be the same as the aws package name, which is where CloudQuery service name is derived from.


erezrokah (approved):

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -281,3 +277,8 @@ func (r *Resource) generateMockTest(dir string) error {
 func (r Resource) StructName() string {
 	return reflect.TypeOf(r.Struct).Elem().Name()
 }
+
+func (r Resource) CloudQueryServiceName() string {
```
hermanschaaf, line 281:
```suggestion

// CloudQueryServiceName is used for accessing 'client.Services().{{.CloudqueryServiceName}}' in templates
func (r Resource) CloudQueryServiceName() string {
```

Issue #1383, opened 2022-06-20
S3 Storage Lens


bbernays:
### Describe the resource.

S3 Storage lens aggregates metrics about S3 buckets in the account or in the org. This data is able to be dumped on a daily basis into a CSV or parquet file stored in an S3 bucket.

### Use Case

The metrics describe usage and provide drill downs by Region, storage class, bucket, and prefix. In large buckets there are no other ways to understand trends and usage other than listing every object, which is costly ($ and time)... 

### Additional context

If we didn't want to grab S3 Storage Lens we could also grab `NumberOfObjects` and `BucketSizeBytes` on a per bucket basis (note metric data is only available in the region where bucket is located)

cqgaurav:
Blocked by: https://github.com/aws/aws-sdk/issues/259

PR #3939, opened 2022-11-08
feat(aws): S3 storage lens configurations

disq:
Implements [#1383](https://github.com/cloudquery/cloudquery/issues/1383)

The S3 export configuration can be found under `aws_s3_lens_configurations.data_export->'S3BucketDestination'`... I don't think we should read the files and import the actual metrics? @bbernays 

**UPDATE**
I can't seem to get this return any data, even tried with `aws_debug` but `aws s3control list-storage-lens-configurations --account-id=XXXX` works for me with the same credentials. Tested the region thing etc. using the aws cli as well.

Related: https://github.com/aws/aws-sdk/issues/259
`aws_debug` wasn't enough so had to add `aws.LogResponseWithBody` to the client log mode, turns out the issue above matches our findings.


plugins/source/aws/codegen/recipes/s3.go
```diff
@@ -107,6 +108,27 @@ func S3Resources() []*Resource {
 					},
 				}...),
 		},
+		{
+			SubService:          "lens_configurations",
```
disq, line 112: `s3_storage_lens_configurations`? I think that's too long so I went for "lens*"

erezrokah (approved):

plugins/source/aws/codegen/recipes/s3.go
hermanschaaf: `s3_storage_lens_configurations` seems closer to the underlying SDK - AWS calls it `StorageLensConfigurations` so I think we should keep it as similar as possible, as long as we don't exceed table name length limits (which for `s3_storage_lens_configurations` we don't).

hermanschaaf (changes requested):
LGTM, let's just decide what the name should be

disq:
Also there's a 'Home Region' concept, I'm not sure if it would have an effect on our requests. Looks like it shouldn't, but I'll test it in a bit.

disq:
I can't seem to get this return any data, even tried with `aws_debug` (I don't see any data in the response either) but  `aws s3control list-storage-lens-configurations --account-id=XXXX` works.

Could be related: https://github.com/aws/aws-sdk/issues/259

erezrokah:
What is the next step for this PR? Contacting AWS support?

Hokwang:
@disq Hi, I didn't read all text but did you check https://github.com/aws/aws-sdk-go-v2 ?
I think https://github.com/aws/aws-sdk is not managed these days.

PR #3946, opened 2022-11-08
chore(docs): Remove stale destination plugins readme

erezrokah:
#### Summary

Removes some stale data from our docs


plugins/README.md
```diff
@@ -1,7 +1,6 @@
-CloudQuery Plugins
-==================================
+# CloudQuery Plugins
```
erezrokah, line 1: We use `##` a bit lower in the doc, so this keeps the headers format consistent

hermanschaaf (approved):

PR #3954, opened 2022-11-08
fix(codegen): Fix issues related to missing codegen for AWS

hermanschaaf:
Because the `gen-code` makefile step didn't first remove files before re-generating, there were some issues where files were left over from recipes that had been previously deleted. This PR fixes these issues and changes the `gen-code` step to delete files before regenerating, so that issues such as this will be picked up in CI in the future. 

The only functional change is that `aws_docdb_cluster_parameters` gets `account_id` and `region` columns that is was missing before (it uses account-region multiplexing).


plugins/source/aws/resources/services/docdb/cluster_snapshot_attributes.go
```diff
@@ -1,28 +0,0 @@
-// Code generated by codegen; DO NOT EDIT.
-
-package docdb
-
-import (
-	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
-	"github.com/cloudquery/plugin-sdk/schema"
-)
-
-func ClusterSnapshotAttributes() *schema.Table {
```
hermanschaaf, line 10: This function was never called.

plugins/source/aws/Makefile
```diff
@@ -30,6 +30,7 @@ lint:
 
 .PHONY: gen-code
 gen-code:
+	grep -rl '// Code generated by codegen; DO NOT EDIT.' resources/services/* | xargs rm
```
hermanschaaf, line 33: We should probably also do this for the codegen of other plugins

Issue #1984, opened 2022-09-22
AWS Resource Access Manager Support


jsonpr:
### Why do you need this resource or plugin? Please describe.
Cloudquery doesn't currently support AWS Resource Access Manager, a service used to share AWS resources with other AWS accounts.

### Describe the solution you'd like
Additional visibility into AWS RAM resource shares shared with an account/organization and shared from an account.

https://docs.aws.amazon.com/ram/latest/userguide/what-is.html

PR #3961, opened 2022-11-08
feat(aws): Add AWS RAM resources

candiduslynx:
Add AWS RAM resources (#1984)


plugins/source/aws/codegen/recipes/ram.go
```diff
@@ -0,0 +1,168 @@
...
+					Type:     schema.TypeString,
+					Resolver: `schema.PathResolver("ServiceName")`,
+					Options:  pk,
+				},
+			),
+		},
+	}
+	for _, r := range resources {
+		r.Service = "ram"
+		r.Description = "https://docs.aws.amazon.com/ram/latest/APIReference/API_" + r.StructName() + ".html"
```
disq, line 161: Why not get rid of `mx` above and put it as a single line here?

disq (commented):
LGTM with single nit

plugins/source/aws/codegen/recipes/ram.go
candiduslynx: `mx` is not used for the relations (naturally, you don't need to do multiplexing as the parent will reside in the same region).

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -277,9 +280,3 @@ func (r *Resource) generateMockTest(dir string) error {
 func (r Resource) StructName() string {
 	return reflect.TypeOf(r.Struct).Elem().Name()
 }
-
-// CloudQueryServiceName is used for accessing 'client.Services().{{.CloudqueryServiceName}}' in templates
-func (r Resource) CloudQueryServiceName() string {
-	csr := caser.New()
-	return csr.ToPascal(r.Service)
-}
```
hermanschaaf, line 285: Maybe a small thing, but I think I'd rather keep this function so the logic is in one place, rather than spread across all the templates. As we saw recently, this does change from time to time. :)

hermanschaaf (changes requested):
Looks good 👍  Just some minor nits / questions.

plugins/source/aws/docs/tables/aws_ram_principals.md
```diff
@@ -0,0 +1,21 @@
...
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|region|String|
+|id (PK)|String|
```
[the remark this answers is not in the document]
bbernays: Unless we can find documentation explicitly stating it is globally unique, we should assume that it is only unique within  region + account

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -39,6 +39,9 @@ type Resource struct {
 	// --------------------------------
 	ShouldGenerateResolverAndMockTest bool
 	ResolverAndMockTestTemplate       string
+
+	// To be used in list/paginator resolver
+	MaxResults int
```
bbernays, line 44: Do we want to split the codegen changes into a separate PR?

candiduslynx: I specifically added the new templates to gen the new resources, so maybe leave in single PR?
If you feel strongly it should be split I can do it, though.

hermanschaaf (approved):
Love the new PKColumns option 😍

plugins/source/aws/docs/tables/aws_ram_resource_share_invitations.md
```diff
@@ -0,0 +1,25 @@
+# Table: aws_ram_resource_share_invitations
+
+https://docs.aws.amazon.com/ram/latest/APIReference/API_ResourceShareInvitation.html
+
+The primary key for this table is **resource_share_invitation_arn**.
```
hermanschaaf, line 5: Could we rename the `resource_share_invitation_arn` column to `arn` please, to be consistent with other tables :bow:

hermanschaaf (changes requested):
Sorry, just noticed two ARNs column names that should ideally be consistent with other AWS tables.

plugins/source/aws/docs/tables/aws_ram_resource_shares.md
```diff
@@ -0,0 +1,30 @@
+# Table: aws_ram_resource_shares
+
+https://docs.aws.amazon.com/ram/latest/APIReference/API_ResourceShare.html
+
+The primary key for this table is **resource_share_arn**.
```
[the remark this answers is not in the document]
candiduslynx: will do

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -28,17 +28,24 @@ type Resource struct {
...
+	PKColumns             []string
 	Table                 *codegen.TableDefinition
 	Multiplex             string
 	PreResourceResolver   string
 	PostResourceResolver  string
 	Relations             []string
 	UnwrapEmbeddedStructs bool
 
+	// NameTransformer custom name transformer for resource
+	NameTransformer func(field reflect.StructField) (string, error)
```
hermanschaaf, line 40: I'm aware we have the same thing in the GCP plugin, but it seems a bit like overkill to have a generic transformer function here, when (I'm pretty sure) `map[string]string` would suffice for all cases we know of in AWS 😅 The use case in GCP is also slightly different: it's for overriding things like "CORS" with "cors" instead of "c_o_r_s" (and I think we actually have better mechanisms for handling this now)

Could we maybe make it something like `OverrideColumnNames: map[string]string{}` with the key being the struct field, and the value being the new column name? So then generally it would be used something like:

```
OverrideColumnNames: map[string]string{
    "ResourceShareArn": "arn",
}
```

This also has the advantage that you don't need to guess what the transformed column name will be, you can use the struct field name, which is certain.

What do you think?

hermanschaaf (approved):

bbernays (approved):

plugins/source/aws/codegen/recipes/base.go
candiduslynx: If we consider the field names as keys, only the name transformer option can help us at the moment.
We can do a simple map if we use the resulting column names instead, but for now I think it works better.

PR #3965, opened 2022-11-08
feat(aws): Update AWS packages

bbernays:
#### Summary

Generated by running: `go list -m all | grep github.com/aws/aws-sdk-go-v2 | awk '{print $1}' | xargs go get -u && go mod tidy && make gen`


hermanschaaf (approved):

erezrokah:
I believe this should be a `feat`. We should be able generate this by ticking any of the AWS boxes in https://github.com/cloudquery/cloudquery/issues/760.
Another option is to change our dep update schedule (monthly at the moment) to weekly for AWS updates

bbernays:
@erezrokah - It should definitely be more often than monthly. It might even be good to do it daily... There is no reason why we should not be right in lock step with AWS package releases now that we have the ability to autogenerate all of the services

disq (approved):

erezrokah:
> It might even be good to do it daily

Let's try with https://github.com/cloudquery/.github/pull/370/files

Issue #2301, opened 2022-10-04
AWS MSK


Barneyjm:
### Why do you need this resource or plugin? Please describe.
MSK is an expensive service we'd like to have more insight into

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/kafka

### Describe the solution you'd like
Add tables:

- clusters
- nodes
- configuration versions? (might be too many)

PR #3967, opened 2022-11-08
feat: Add MSK and codegen from AWS Client structs

hermanschaaf:
## Summary

This change adds:
 - MSK resources (called `kafka` to match the AWS SDK name)
 - A significant update to the codegen for resolvers and mock tests: The `List` and `Describe` function names are now determined automatically through reflection, and not inferred from the struct name. 
 - A new set of `List & Describe` templates are added to resolver + mock test codegen for resources that use `PreResourceResolver`.
 - Automatic support for parent-child relationships is built in. In this case, mock tests are updated to expect the parent table to make the calls, while child tables build the mock expectations for their own resource.
 - Removed paginated templates. Because Paginators are implemented inconsistently, it is less code to auto-generate from a Client struct without Paginators, because we'd always need to support both versions. Maybe later we can auto-detect the pagination Token field to use on output values, but that is not implemented here.

## Checklist

- [x] Codegen recipes for all Kafka resources
- [x] Fixes to existing resolver & mock codegen templates
- [x] Add Kafka resource tags

Closes https://github.com/cloudquery/cloudquery/issues/2301


plugins/source/aws/codegen/recipes/discover/discover.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return f[0], nil
+}
+
+func FindDescribeMethod(client, targetStruct interface{}) (DiscoveredMethod, error) {
+	return FindMethod(client, targetStruct, []string{"Describe", "Get"})
+}
+
+func FindListMethod(client, targetStruct interface{}) (DiscoveredMethod, error) {
+	return FindMethod(client, targetStruct, []string{"List"})
```
candiduslynx, line 128: Sometimes the list method [can be called](https://docs.aws.amazon.com/ram/latest/APIReference/API_GetResourceShareInvitations.html) `GetXXX`

plugins/source/aws/resources/services/ram/principals_fetch.go
```diff
@@ -11,15 +11,20 @@ import (
 	"github.com/cloudquery/plugin-sdk/schema"
 )
 
-func fetchRamPrincipals(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- interface{}) error {
-	input := &ram.ListPrincipalsInput{MaxResults: aws.Int32(500)}
-	paginator := ram.NewListPrincipalsPaginator(meta.(*client.Client).Services().Ram, input)
-	for paginator.HasMorePages() {
-		response, err := paginator.NextPage(ctx)
+func fetchRamPrincipals(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	var input ram.ListPrincipalsInput = listPrincipalsInput()
```
candiduslynx, line 15: so, dropping paginators altogether?

candiduslynx (commented):
I wonder, why do we drop paginators altogether?

plugins/source/aws/resources/services/ram/principals_fetch.go
hermanschaaf: This is explained in the PR description

plugins/source/aws/codegen/recipes/discover/discover.go
hermanschaaf: Hmm, yeah, good point... because that's the same prefix as `Describe`, the reflection may get confused... In this case you will need to specify an override using `ListMethodName`. So for the example you linked, you would set `ListMethodName` to `GetResourceShareInvitations`.

disq: Another one [here](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetInventorySchema.html)

candiduslynx (approved):

disq (approved):
LGTM with single nit

plugins/source/aws/codegen/main.go
```diff
@@ -93,6 +95,11 @@ func generateResources() ([]*recipes.Resource, error) {
 	resources = append(resources, recipes.WAFv2Resources()...)
 	resources = append(resources, recipes.WorkspacesResources()...)
 	resources = append(resources, recipes.XRayResources()...)
+
+	err := recipes.SetParentChildRelationships(resources)
+	if err != nil {
+		return nil, fmt.Errorf("failed to set parent-child relationships: %v", err)
```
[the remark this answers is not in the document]
hermanschaaf: Well-spotted, thanks!

Issue #3913, opened 2022-11-08
chore(ci): Test that our make files are working


erezrokah:
This came up from a user report on discord where they ran make commands (from the root of the repo) and the `build` one failed.

We should test those work to make it easier for people to contribute.

Since some of those commands are quite expensive and slow to run on each PR, we could have a scheduled workflow that runs once/twice a day to test them

PR #3982, opened 2022-11-09
chore: Delete root Makefile

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3913 that was experienced by a Discord user trying to contribute.
Since running all targets in the make file can be slow, this PR does it twice per day.

I also removed 2 targets:
1. `install-tools` was failing. I think it needs to be per module
2. `pre-commit` didn't do anything as we need to run go linter per module with the relevant config

Happy to open issues for ⬆️ if we feel like we need them


erezrokah:
Blocked by https://github.com/cloudquery/plugin-sdk/pull/373

.github/workflows/monorepo_makefile_test.yml
```diff
@@ -0,0 +1,39 @@
+name: Test monorepo Makefile
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  "0 12 * * *"
+  push:
```
erezrokah, line 7: This is here for testing purposes. I'll remove it before merging

yevgenypats:
Can we remove the root makefile at all? is it at all used by anyone? I only use a plugin specific makefile?

yevgenypats (approved):

Makefile
```diff
@@ -40,3 +29,8 @@ update-docs:
 .PHONY: update-plugin-sdk
 update-plugin-sdk:
 	./scripts/update-plugin-sdk.sh
+
+
+.PHONY: list
+list:
+	@grep '^[^#[:space:].].*$$' Makefile | sed 's/\(.*\):/"\1",/' | tr -d '\n' | sed 's/,$$//g'
```
disq: I know this is for CI only, but it doesn't work on mac unless you do `grep -E` at the start. It also seems to find some targets (from targets that invoke other targets) multiple times.

```
grep -E '^[^#[:space:].].*$$' Makefile | sed 's/\(.*\):/"\1",/' | tr -d '\n' | sed 's/,$$//g'
"gen-mocks","test","install-tools","install-hooks","gen-docs","lint","gen-code","gen", gen-code gen-mocks gen-docs
```

Issue #2605, opened 2022-10-10
Support for AWS File Cache (Amazon FSx)


jsonpr:
### Is your feature request related to a problem? Please describe.
https://aws.amazon.com/about-aws/whats-new/2022/09/amazon-file-cache-generally-available/

New release by Amazon for AWS File Cache (part of FSx).

### Describe the solution you'd like
CQ to support File Cache as a AWS resource.

### Additional context
https://docs.aws.amazon.com/fsx/latest/APIReference/API_DescribeFileCaches.html

PR #4091, opened 2022-11-09
feat(aws): Add support for FSX File Caches

disq:
Implements #2605


plugins/source/aws/resources/services/fsx/file_caches_fetch.go
```diff
@@ -0,0 +1,46 @@
...
+	cl := meta.(*client.Client)
+	svc := cl.Services().Fsx
+	input := fsx.DescribeFileCachesInput{MaxResults: aws.Int32(1000)}
+	paginator := fsx.NewDescribeFileCachesPaginator(svc, &input)
+	for paginator.HasMorePages() {
+		result, err := paginator.NextPage(ctx)
+		if err != nil {
+			var ae smithy.APIError
+			if errors.As(err, &ae) && ae.ErrorCode() == "BadRequest" && ae.ErrorMessage() == "The requested feature is not enabled for this AWS account." {
+				log.Warn().Err(err).Msg("skipping resource")
```
disq, line 26: warn the user? not warn the user? ⚖️

erezrokah: Is this feature enabled by default or disabled? If it's enabled by default we should throw an error, otherwise we can print a warning (I think)

erezrokah (approved):

erezrokah:
@disq can we merge this and maybe change the logic to return the error if we see that we need to?

PR #4106, opened 2022-11-09
fix!: Update SDK for Heroku, update code gen

erezrokah:
#### Summary

Replaces https://github.com/cloudquery/cloudquery/pull/4044. The new SDK changes cc @candiduslynx discovered a few empty tables for the Heroku plugin, so I'm fixing them in a separate PR.
Underlying issue is that we were referencing empty structs, for example: https://pkg.go.dev/github.com/davidji99/heroku-go/v5#AddOnWebhookDelivery

The solution has 2 parts:
1. Created a wrapper struct on top of the correct struct and use the `WithUnwrapAllEmbeddedStructs` option
2. When there's no correct struct (the result is only a list), embed the list result as a struct, for example https://pkg.go.dev/github.com/davidji99/heroku-go/v5#PipelineBuildListResult

This is a breaking change due to the changes in primary keys


plugins/source/heroku/codegen/recipes/recipes.go
```diff
@@ -31,6 +32,138 @@ type Resource struct {
...
+type AddOnWebhookEvent struct {
+	heroku.AddOnWebhookEventInfoResult
+}
+
+type AppWebhook struct {
+	heroku.AppWebhookInfoResult
+}
+
+// Element of PipelineBuildListResult
+type PipelineBuild struct {
```
erezrokah, line 48: If there's a better way to do this, please let me know

disq (approved):

plugins/source/heroku/codegen/recipes/recipes.go
candiduslynx: `make(PipelineBuildListResult, 1)[0]`?

candiduslynx:
updated by #4117

PR #4117, opened 2022-11-09
refactor(heroku): Gen using SDK structs

candiduslynx:
connected to https://github.com/cloudquery/plugin-sdk/pull/386


erezrokah (approved):
Nice one @candiduslynx, I renamed this to `refactor` has it doesn't have any user facing changes and should not trigger a changelog entry

PR #4119, opened 2022-11-09
fix(heroku): Fix data structures URLs in table descriptions

candiduslynx:


erezrokah (approved):
This is actually a `fix:` as it updates tables descriptions so it's user facing.

Thanks @candiduslynx!

PR #4143, opened 2022-11-10
chore(tests): Use temp dir for .cq and log file in cmd tests

hermanschaaf:
This updates the cli tests to use a temporary directory on every run, so that previous test runs cannot influence the outcome of the test (e.g. by already having the plugin downloaded locally).


disq (approved):

cli/cmd/migrate_test.go
```diff
@@ -8,21 +8,19 @@ import (
 )
 
 func TestMigrate(t *testing.T) {
-	t.Cleanup(func() {
-		os.RemoveAll("cloudquery.log")
```
disq, line 12: Meant to be removed? or `os.RemoveAll(logFileName)` I suppose?

cli/cmd/sync_test.go
```diff
@@ -8,21 +8,19 @@ import (
 )
 
 func TestSync(t *testing.T) {
-	t.Cleanup(func() {
-		os.RemoveAll("cloudquery.log")
```
disq, line 12: Same here

disq (changes requested):

cli/cmd/migrate_test.go
hermanschaaf: It's now written inside the temporary directory, so it should get removed along with the temporary directory at the end of the test.

cli/cmd/doc_test.go
```diff
@@ -12,17 +13,14 @@ var docFiles = []string{
...
-	tmpdir, tmpErr := os.MkdirTemp("", "docs_*")
-	if tmpErr != nil {
-		t.Fatalf("failed to create temporary directory: %v", tmpErr)
-	}
-	defer os.RemoveAll(tmpdir)
 	cmd := NewCmdRoot()
-	cmd.SetArgs([]string{"doc", tmpdir})
+	tmpDir := t.TempDir()
+	logFileName := path.Join(tmpDir, "cloudquery.log")
+	cmd.SetArgs([]string{"doc", tmpDir, "--cq-dir", tmpDir, "--log-file-name", logFileName})
```
hermanschaaf, line 19: Oh wait this is not right 😄 One sec...

hermanschaaf:
Not sure why the test is failing on Windows 🤔

erezrokah:
> Not sure why the test is failing on Windows 🤔

I think it's related to this bug https://github.com/cloudquery/plugin-sdk/issues/354
Failure https://github.com/cloudquery/cloudquery/actions/runs/3435864156/jobs/5728723449#step:7:32

hermanschaaf:
@erezrokah Ah right, okay. Maybe I'll disable this test on windows until we can figure out https://github.com/cloudquery/plugin-sdk/issues/354

erezrokah:
> @erezrokah Ah right, okay. Maybe I'll disable this test on windows until we can figure out [cloudquery/plugin-sdk#354](https://github.com/cloudquery/plugin-sdk/issues/354)

Problem is that we don't have that many CLI tests Windows. Maybe we can skip the cleanup on Windows?

disq (approved):

PR #4158, opened 2022-11-10
fix: Fix test destination plugin hang

hermanschaaf:
As best I can tell, the test destination plugin wasn't reading from the channel, causing a deadlock.

### DIFF
--- a/cli/cmd/testdata/dest.test.yml
+++ b/cli/cmd/testdata/dest.test.yml
@@ -1,6 +1,5 @@
 kind: "destination"
 spec:
   name: "test"
-  path: "localhost:7778"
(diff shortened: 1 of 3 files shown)


cli/cmd/testdata/dest.test.yml
```diff
@@ -1,5 +1,6 @@
 kind: "destination"
 spec:
   name: "test"
-  path: "cloudquery/test"
-  version: "v1.2.0" # latest version of destination test plugin
+  path: "localhost:7778"
```
hermanschaaf, line 4: Oops, will revert this

erezrokah (approved):

PR #4189, opened 2022-11-10
feat(azure): Add description to tables

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3086.

I looked at the Azure REST API and it has more data than the Go one. Also Azure has so many different versions of the same API so it seems the safest to point to the Go package docs


disq (approved):

erezrokah:
/gen sha=68cafe253b206f080586bf125978bae44a4f0815 plugin=azure

PR #4330, opened 2022-11-10
feat: Added IAM access advisor resources

### LINKED ISSUE #1402: IAM Access Advisor
### Describe the resource.

Generate a list of entities/actions for users, groups, roles, or managed policies that have been accessed in the last 400 days. 

https://docs.aws.amazon.com/cli/latest/reference/iam/generate-service-last-accessed-details.html
https://docs.aws.amazon.com/cli/latest/reference/iam/get-service-last-accessed-details.html
https://docs.aws.amazon.com/cli/latest/reference/iam/get-service-last-accessed-details-with-entities.html

- [ ] cloudquery/cloudquery#1389
- [ ] cloudquery/cloudquery#1388
- [ ] cloudquery/cloudquery#1387
- [ ] cloudquery/cloudquery#1386

### Use Case

Understanding historical usage

### Additional context

_No response_

amanenk:
#### Summary

closes #1402
The resources from this PR could be relations of iam_users, iam_policies, iam_groups, and iam_roles, but they are implemented as independent resources because they significantly affect fetch time, so users should be able to disable them.


plugins/source/aws/docs/tables/aws_iam_group_services_last_accessed.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|service_name|String|
```
hermanschaaf, line 16: Would account_id + service_name be a possible primary key here, or could there be cases where that is not unique? Is there any combination that will definitely be uniquely identifying?

plugins/source/aws/docs/tables/aws_iam_policy_services_last_accessed.md
```diff
@@ -0,0 +1,23 @@
...
+
+The primary key for this table is **_cq_id**.
+
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
```
hermanschaaf, line 13: Same question here and all the other tables

plugins/source/aws/resources/services/iam/access_details_fetch.go
```diff
@@ -0,0 +1,80 @@
...
+		details, err := svc.GetServiceLastAccessedDetails(ctx, &getDetails)
+		if err != nil {
+			return err
+		}
+
+		requestTime := 0
+		switch details.JobStatus {
+		case types.JobStatusTypeInProgress:
+			time.Sleep(time.Millisecond * 200)
+			requestTime += 200
```
hermanschaaf, line 37: It doesn't look like `requestTime` is being used for anything, it can probably be removed

plugins/source/aws/resources/services/iam/access_details_fetch.go
```diff
@@ -0,0 +1,80 @@
...
+			continue
+		case types.JobStatusTypeFailed:
+			return fmt.Errorf("failed to get last acessed details with error: %s - %s", *details.Error.Code, *details.Error.Message)
```
hermanschaaf, line 40:
```suggestion
			return fmt.Errorf("failed to get last accessed details with error: %s - %s", *details.Error.Code, *details.Error.Message)
```

plugins/source/aws/resources/services/iam/access_details_fetch.go
```diff
@@ -0,0 +1,80 @@
...
+	for {
+		details, err := svc.GetServiceLastAccessedDetails(ctx, &getDetails)
+		if err != nil {
+			return err
+		}
+
+		requestTime := 0
+		switch details.JobStatus {
+		case types.JobStatusTypeInProgress:
+			time.Sleep(time.Millisecond * 200)
```
hermanschaaf, line 36: How long do these reports take to generate on average? Seconds, minutes, hours? Just wondering if 200ms might be too frequent, maybe we should do every second or so.

plugins/source/aws/resources/services/iam/policy_services_last_accessed_fetch..go
```diff
@@ -0,0 +1,38 @@
...
+
+func fetchIamPolicyServicesLastAccessed(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	svc := c.Services().Iam
+	ch := make(chan interface{})
+
+	g, ctx := errgroup.WithContext(ctx)
+	g.Go(func() error {
+		if err := fetchIamPolicies(ctx, meta, nil, ch); err != nil {
+			return err
```
hermanschaaf, line 20: If this returns an error, I think we should still close the channel, otherwise the goroutine below will hang. (Same comment applies to the other similar functions)

hermanschaaf (changes requested):
Nice work! I'm leaving some feedback now, but because this deviates from the normal patterns a bit I'd like to check it and try it out locally a bit too.

hermanschaaf (changes requested):
@amanenk This looks good! I tested it a bit locally and found it to be a bit slow though (around 4 resources per second).

We may want to consider some ideas to speed it up:
 - have you tried using a PreResourceResolver step to fetch the details? This might/should enable the SDK to better leverage available goroutines
 - we may want to consider making these tables relations of `iam_users`, `iam_policies`, `iam_groups`, and `iam_roles` now, because it recently became possible to skip child relations in config. So if there is no avoiding them being slow, there is a way to skip them. This might help with the speed because we wouldn't need to list the users, policies, etc more than once.

bbernays:
> @amanenk This looks good! I tested it a bit locally and found it to be a bit slow though (around 4 resources per second).
> 
> We may want to consider some ideas to speed it up:
> 
> * have you tried using a PreResourceResolver step to fetch the details? This might/should enable the SDK to better leverage available goroutines
> * we may want to consider making these tables relations of `iam_users`, `iam_policies`, `iam_groups`, and `iam_roles` now, because it recently became possible to skip child relations in config. So if there is no avoiding them being slow, there is a way to skip them. This might help with the speed because we wouldn't need to list the users, policies, etc more than once.

I would actually say that these new resources should be standalone rather than relations... Happy to have a larger conversation about this though

cqgaurav:
Whats the next step here ?

hermanschaaf:
> Whats the next step here ?

@cqgaurav Good question... we're still debating this a bit internally - as things stand these resources would take too long to fetch, and it's currently hard to know which resources take very long, so we need to (at least) make sure that they aren't included in any config by default. So we should probably put this on hold a little bit until we find a solution to these other problems, then get back to this.

erezrokah:
Going to close the PR as stale. I'll take a look at https://github.com/cloudquery/cloudquery/issues/1402 to see what we can do regarding the performance issues

erezrokah:
As a side note, after looking at the AWS APIs, and this PR, I think it should be implemented as relations, e.g.
`users->jobs->details->entities` to make it as concurrent as possible. Unless I missed something this PR creates the jobs and gets the details sequentially https://github.com/cloudquery/cloudquery/pull/4330/files#diff-65c24d42cae8092f5dba5779744acf720c4f43ed7250390792832b5de062e7c9R30.

I'd expect to at create the jobs in parallel and also wait for them to complete in parallel

PR #4342, opened 2022-11-10
fix: Quote tables and columns for SQLite

yevgenypats:
This fixes sqlite destination and add documentation and recipe. 

Also, this added a small release blog so we can drop this tomorrow! :shipit:


plugins/destination/sqlite/client/migrate.go
```diff
@@ -88,7 +88,7 @@ func (c *Client) autoMigrateTable(_ context.Context, table *schema.Table) error
 		switch {
 		case sqliteColumn == nil:
 			c.logger.Debug().Str("table", table.Name).Str("column", col.Name).Msg("Column doesn't exist, creating")
-			sql := "alter table " + table.Name + " add column " + columnName + " " + columnType
+			sql := "alter table \"" + table.Name + "\" add column \"" + columnName + "\" \"" + columnType + `"`
```
disq, line 91: Why not use \` throughout and get rid of `"`
```suggestion
			sql := `alter table "` + table.Name + `" add column "` + columnName + `" "` + columnType + `"`
```

plugins/destination/sqlite/client/migrate.go
```diff
@@ -148,7 +148,7 @@ func (c *Client) createTableIfNotExist(_ context.Context, table *schema.Table) e
 	sb.WriteString(")")
 	_, err := c.db.Exec(sb.String())
 	if err != nil {
-		return fmt.Errorf("failed to create table %s: %w", table.Name, err)
+		return fmt.Errorf("failed to create table with '%s': %w", sb.String(), err)
```
disq, line 151: Would use `%q` instead of `'%s'` so that it's properly quoted
```suggestion
		return fmt.Errorf("failed to create table with %q: %w", sb.String(), err)
```

disq (approved):
LGTM with nits

PR #4376, opened 2022-11-11
fix(aws): Remove NotFound checks in PreResourceResolvers

### LINKED ISSUE #4374: Relations with PreResourceResolver parents are called with wrong parent if "Get" returns 404 and skipped
### Describe the Bug

If the parent has a `PreResourceResolver` defined, and that resolver checks for 404 and does a simple `return nil`, relations are called with the main `Resolver`'s output and not the "new item" from the PreResourceResolver.

### Expected Behavior

Relation resolvers shouldn't be called.

### References
First I thought it was an SDK issue, but it's not. https://github.com/cloudquery/plugin-sdk/issues/399

disq:
~Needs https://github.com/cloudquery/plugin-sdk/pull/401 to work.~

Fixes #4374


plugins/source/aws/go.mod
```diff
@@ -153,3 +153,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudquery/plugin-sdk => ../../../../plugin-sdk
```
candiduslynx, line 157: needs to be updated after https://github.com/cloudquery/plugin-sdk/pull/401

candiduslynx (dismissed):

yevgenypats (changes requested):
Why not just return the error (and remove the dead code checking for not-found) so errors will be logged in unified format?

disq:
> Why not just return the error (and remove the dead code checking for not-found) so errors will be logged in unified format?

Sure, done. All the regular resolvers still have notfound checks, maybe another, larger PR for those in the future as well.

erezrokah (approved):

erezrokah (approved):

PR #4378, opened 2022-11-11
feat: Add flag to force test destination plugin to return an error

hermanschaaf:
This adds an `error_on_write` option to the test destination plugin, which forces an error to be returned when Write is called. This can be used to test the plugin-sdk in e2e tests and make sure that the error is handled correctly.


erezrokah (approved):
🚀

Issue #1409, opened 2022-03-24
Add Support for AWS::SSM::*


bbernays:
- [ ] AWS::SSM::AssociationCompliance
- [ ] AWS::SSM::ManagedInstanceInventory
- [ ] AWS::SSM::PatchCompliance

disq:
About `AWS::SSM::AssociationCompliance`: Looks like [ListComplianceItems](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_ListComplianceItems.html) only supports the `ManagedInstance` resource type, not associations?

disq:
[ListInventoryEntries](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ssm#ListInventoryEntriesInput) require InstanceId and TypeName which gets a little hairy to query: We don't know if the ID returned from `GetInventory` is a node or an instance ID (not that there could be a difference), and for the `TypeName` we might have to run an API call for each item returned in the `GetInventorySchema` (15 types below), which could get too much.

```
aws ssm get-inventory-schema --region us-east-1 | jq -r .Schemas[].TypeName
AWS:AWSComponent
AWS:Application
AWS:ComplianceItem
AWS:File
AWS:InstanceDetailedInformation
AWS:InstanceInformation
AWS:Network
AWS:PatchCompliance
AWS:PatchSummary
AWS:ResourceGroup
AWS:Service
AWS:Tag
AWS:WindowsRegistry
AWS:WindowsRole
AWS:WindowsUpdate
```

PR #4381, opened 2022-11-11
feat(aws): More SSM resources

disq:
Hopefully implements #1409 

I'm not sure if this is enough or if we need more (`DescribeInstancePatchStates` for instance... which can take multiple InstanceIds as input, so challenges ahead)


hermanschaaf (approved):
LGTM 👍  I think we can always start here and then use the the `Describe...`-type calls later if it turns out they are required by some users.

erezrokah (approved):

disq:
Maybe merge for now and see it in the field vs. customer requirements/issues?

PR #4382, opened 2022-11-11
fix: RDS fixes

shimonp21:
#### Summary

See individual commits please :)


hermanschaaf:
@shimonp21 I haven't looked into this yet; what happened here, and how do you know this is dead code?

shimonp21:
Looked at the reference - There is no reason for the type-switch and for one of the functions, because `fetchRdsClusterParameters` is only called from one place.

shimonp21:
Also tested with logs and saw that one of the paths is never called

shimonp21:
@hermanschaaf, @bbernays  - Added one more fix here - PTAL

hermanschaaf (approved):

PR #4456, opened 2022-11-12
fix: Update link to gcp config docs

derhally:
#### Summary

Link to the gcp config docs was pointing to the wrong location


yevgenypats (approved):
Nice catch! Thanks! 🙏

PR #4468, opened 2022-11-13
test: Validate Go Releaser ZIP files

erezrokah:
#### Summary

Prevents https://github.com/cloudquery/cloudquery/pull/4286 from happening again


yevgenypats (changes requested):
Let's hold off a bit with this, I'd like to give a try to see if there is a way to simplify this as seems complexity is starting to grow around our workflows.

erezrokah:
Per @yevgenypats suggestion I'm going to see if we can move this validation to the Go Releaser configuration file

erezrokah:
@yevgenypats see different approach in https://github.com/cloudquery/cloudquery/pull/4468/commits/b9869b7208066ee3cfd2855ea6fed65d44be111b

plugins/.goreleaser.yaml
```diff
@@ -19,4 +19,18 @@ builds:
...
-        goarch: arm64
\ No newline at end of file
+        goarch: arm64
+after:
+  hooks:
+    - cmd: unzip -o {{ .Var.binary }}_linux_arm64.zip
+      dir: dist
+    - cmd: unzip -o {{ .Var.binary }}_darwin_arm64.zip
+      dir: dist
+    - cmd: unzip -o {{ .Var.binary }}_windows_amd64.zip
```
candiduslynx, line 29: Will we add windows_arm64 as well?

yevgenypats (approved):

PR #4469, opened 2022-11-13
chore(ci): Align gen command for source plugins

erezrokah:
#### Summary

This is required for the AWS `/gen` command to work as AWS also runs `make gen-mocks`


erezrokah:
Merging to unblock https://github.com/cloudquery/cloudquery/pull/4453

PR #4477, opened 2022-11-13
fix: Close log file and analytics client on errors

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/4479

First commit of this PR adds a test to reproduce an issue with the CLI not closing the log file and analytics client when there's an error (noticeable on Windows).
See error https://github.com/cloudquery/cloudquery/actions/runs/3455565203/jobs/5767678202#step:7:45
```
testing.go:1097: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestSyncshould_fail_with_missing_path_error_when_path_is_missing3395995344\001\cloudquery.log: The process cannot access the file because it is being used by another process.
```

Second commits fixes the CLI to close the log file and analytics client on error.

### Update

The way our CLI tests are set up doesn't allow us to test the fix properly as the code that closes the log needs to run outside the context of the Cobra command.
If we'd like to test it we'll need to run `main()` directly instead of executing the Cobra command


cli/cmd/root.go
```diff
@@ -95,14 +95,6 @@ func NewCmdRoot() *cobra.Command {
 
 			return nil
 		},
-		PersistentPostRun: func(cmd *cobra.Command, args []string) {
```
erezrokah, line 98: This doesn't run on errors

cli/cmd/root.go
```diff
@@ -137,6 +129,11 @@ func NewCmdRoot() *cobra.Command {
 	)
 	cmd.CompletionOptions.HiddenDefaultCmd = true
 	cmd.DisableAutoGenTag = true
+	cobra.OnFinalize(func() {
+		if analyticsClient != nil {
```
erezrokah, line 133: This is the way to clean up resources with Cobra. However we can't close the log file here since we use it later to print the error

cli/main.go
```diff
@@ -35,20 +35,23 @@ func executeRootCmdWithContext() error {
...
+		os.Exit(exitCode)
 	}()
+	defer cmd.CloseLogFile()
 
 	// This ensures we don't print anything until logging is configured
 	log.Logger = log.Level(zerolog.Disabled)
 	if err := executeRootCmdWithContext(); err != nil {
 		log.Error().Err(err).Msg("exiting with error")
-		//nolint:all This is fine if deferred is not called because there was no panic
-		os.Exit(1)
```
erezrokah, line 52: `os.Exit` skips any deferred functions hence if we want to close the log file we need to defer `os.Exit` too

cli/cmd/root.go
disq: Maybe it could be done with `runtime.SetFinalizer` but it would be hacky at best. There's a fairly new [runtime.addExitHook](https://go-review.googlesource.com/c/go/+/354790) discussion/patch which is merged but since it looks to be unexported I think it's not meant for outside use yet.

disq (approved):
All the hacky code from v0 cli are coming back slowly 😂 🤷

erezrokah:
Thanks for the review @disq, cc @yevgenypats in case there's a better way to do this

erezrokah:
Another option would be to never throw in Cobra and handle any failures in the `PostRun*` method

cli/main.go
candiduslynx: [runtime.Goexit()](https://pkg.go.dev/runtime#Goexit)?

erezrokah: @candiduslynx can you explain:

```
Calling Goexit from the main goroutine terminates that goroutine without func main returning. Since func main has not returned, the program continues execution of other goroutines. If all other goroutines exit, the program crashes.
```

erezrokah:
Going to merge this and we can do any follow up fixes in future PRs

Issue #2688, opened 2022-10-12
chore(ci): Warn when a PR touches more than 300 files for multiple plugins


erezrokah:
Due to a limitation in GitHub diffing if a PR changes more than 300 files, some of our workflows won't run:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

![image](https://user-images.githubusercontent.com/26760571/195340371-0a381b5d-1657-4ce1-a143-24d2c3454ba2.png)

We could add a check to detect when this happens and alert the PR author, as it's quite easy to miss.
It could be a part of the [wait for all](https://github.com/cloudquery/cloudquery/blob/b4647d6b27dc79da641b6a97d2e682c6052c3862/.github/workflows/wait_for_required_workflows.yml) workflow

PR #4483, opened 2022-11-13
chore(ci): Error when a PR changes too many files

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/2688

Related to https://github.com/cloudquery/cloudquery/pull/4460.

Not sure what else to do here, as our CI breaks completely when this happens


hermanschaaf (approved):
It would be nice if we could detect when multiple modules are being touched and only error then, but this is good for now, let's see if it becomes a problem 👍

PR #4484, opened 2022-11-13
chore: Simplify workflows

yevgenypats:


erezrokah (changes requested):
I think we should discuss this more as this will increase our CI time significantly.
Maybe we can schedule something?

yevgenypats:
I pushed to the wrong branch. I actually wanted to play with this and do various banchmarks and tries on a fork - https://github.com/cloudquery/cloudquery-test

PR #4486, opened 2022-11-13
fix: Error handling omissions

tockn:
#### Summary

Fix error handling omissions in `cli/cmd/sync.go`


erezrokah (approved):
Thanks @tockn!

PR #4487, opened 2022-11-13
fix: Error handling omissions in sync2

tockn:
#### Summary

I also found  error handling omissions in `cli/cmd/sync2.go` !
ref: https://github.com/cloudquery/cloudquery/pull/4486


erezrokah (approved):
🚀

PR #4495, opened 2022-11-13
feat: Regenerate k8s

yevgenypats:
Given multiple requests from number of key users around incompleteness  of k8s I updated our codegen to do both fetcher and mocks. k8s API is especially good and consistent so it was relatively straight forward to have 99% auto-generated source plugin apart of a few places with custom resolver or overrides (that are still done on the codegen level).

This doesn't have any breaking changes but things to consider in the future is how to handle v1,v1beta,v2 apis. We might want to introduce those into our tables as well but can be done in a future PR.

Once this is merged it will also help with the WIP policy for k8s.


plugins/source/k8s/client/client.go
```diff
@@ -52,14 +52,14 @@ func (c Client) WithContext(k8sContext string) *Client {
...
-	c.contexts = contexts
-}
+// func (c *Client) SetServices(s map[string]Services) {
+// 	c.services = s
+// 	contexts := make([]string, 0, len(s))
+// 	for k := range s {
+// 		contexts = append(contexts, k)
+// 	}
+// 	c.contexts = contexts
+// }
```
erezrokah, line 62:
```suggestion
```

Can we delete this if not used?

plugins/source/k8s/gen.sh
```diff
@@ -0,0 +1,6 @@
+mockgen --package=mocks --destination="client/mocks/mock.go" "k8s.io/client-go/kubernetes" Interface
```
erezrokah, line 1: Maybe we can put this in the Makefile under:
https://github.com/cloudquery/cloudquery/blob/43d5fcd77160c4803f88ddce55cd96969da66e21/plugins/source/k8s/Makefile#L25

or `gen-mocks:` and update `gen:`?

erezrokah (approved):
Nice 🚀  Added 2 non blocking comments

plugins/source/k8s/Makefile
```diff
@@ -1,7 +1,7 @@
 # Generate mocks for mock/unit testing 
 .PHONY: generate-mocks
-generate-mocks:
-	go generate ./client/services/...
+gen-mocks:
```
erezrokah, line 3: Can we add this to the `gen` target?

plugins/source/k8s/codegen/recipes/templates/mock.go.tpl
```diff
@@ -0,0 +1,51 @@
+// Code generated by codegen; DO NOT EDIT.
```
shimonp21, line 1: nit: rename file to `mock_test.go.tpl`

plugins/source/k8s/Makefile
yevgenypats: Added

erezrokah: Where :) ? I don't see a new commit

erezrokah:
@yevgenypats I added https://github.com/cloudquery/cloudquery/pull/4495/commits/09bbd5d5515e708fa790efcb1a642a47ac2fac7a per my comment. I hope that makes sense

PR #4548, opened 2022-11-14
fix(website): Point "Get Started" to Quickstart

erezrokah:
#### Summary

@hermanschaaf I think you mentioned this a while back


hermanschaaf (approved):
Yeay 🎉

Issue #1821, opened 2022-05-09
feat(getting-started): Auto select guide based on OS


erezrokah:
When navigating to https://docs.cloudquery.io/docs/getting-started/getting-started-with-aws (or any other getting started guide), under [download and install](https://docs.cloudquery.io/docs/getting-started/getting-started-with-aws#download-and-install) the selected command is for Linux.

It would be nice to default to the OS relevant guide

cqgaurav:
Perhaps even a copy button for the lazy folks out there

yevgenypats:
I think for starters we can do osx as default :)

yevgenypats:
From quick glance that can be non-trivial because of nextra/next/react so we will have to make sure this component also runs on the client as well which can have other weird effect like late loading (which might not be better then quick loading and just default for osx).

erezrokah:
> From quick glance that can be non-trivial because of nextra/next/react so we will have to make sure this component also runs on the client as well which can have other weird effect like late loading (which might not be better then quick loading and just default for osx).

Yeah, we'll need to see how it can be done in Nextra, or maybe there's a way to do it via Vercel (load a different version of the site based on user agent or something like that)

PR #4583, opened 2022-11-14
feat(website): Chose guide based on OS

erezrokah:
#### Summary

Fixes  https://github.com/cloudquery/cloudquery/issues/1821

One way to do dynamic content in a static generated site is to use edge logic. This uses Vercel's edge middleware to route to the correct QuickStart guide. As a result I had to split the guide into multiple guides, one per OS.
The good thing is due to JSX composition we don't need to duplicate the content.


yevgenypats (approved):
🚀

PR #4590, opened 2022-11-14
chore(gcp): Simplify iterator usage

candiduslynx:


disq (approved):
LGTM (as long as `it.Next()` is returns an unwrapped `iterator.Done` err)

PR #4617, opened 2022-11-15
fix(terraform): Update to new AWS SDK

candiduslynx:
Inspired by #4611


erezrokah (approved):
I changed this to `fix` so it triggers a release

plugins/source/k8s/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.19
 require (
 	github.com/cloudquery/plugin-sdk v1.5.2
 	github.com/golang/mock v1.6.0
+	github.com/google/gnostic v0.5.7-v3refs
```
erezrokah, line 8: This k8s change seems unrelated to this PR, can we do it in another one?

erezrokah (changes requested):
Thanks @candiduslynx, I think we should move the k8s change to another PR, WDYT?

plugins/source/terraform/client/backend.go
```diff
@@ -61,53 +62,49 @@ func parseAndValidate(reader io.Reader) (*TerraformData, error) {
...
 
-func NewS3TerraformBackend(config *BackendConfigBlock) (*TerraformBackend, error) {
+func NewS3TerraformBackend(ctx context.Context, cfg *BackendConfigBlock) (*TerraformBackend, error) {
 	var b S3BackendConfig
-	if err := json.Unmarshal(*config.Config, &b); err != nil {
+	if err := json.Unmarshal(*cfg.Config, &b); err != nil {
 		return nil, fmt.Errorf("cannot parse s3 backend config: %w", err)
 	}
 
+	awsConfig, err := config.LoadDefaultConfig(context.TODO())
```
disq, line 71: Not context.TODO :)
```suggestion
	awsConfig, err := config.LoadDefaultConfig(ctx)
```

disq (changes requested):
Single issue with a rogue context.TODO()

plugins/source/k8s/go.mod
candiduslynx: we can, I just ran `go mod tidy` in all of our plugins

erezrokah: Got it, but if we merge this PR a new changelog entry for k8s will be created with `Update to new AWS SDK` which is a bit confusing

erezrokah: I see that you reverted it, thanks!

erezrokah (approved):

disq (approved):

PR #4629, opened 2022-11-15
fix: Update k8s code gen

erezrokah:
#### Summary

This matches the `gen` Makefile target. I'll do all others plugins soon, k8s is done separately since it requires a code change


plugins/source/k8s/mocks/discovery/v1/client.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by MockGen. DO NOT EDIT.
-// Source: k8s.io/client-go/kubernetes/typed/discovery/v1 (interfaces: DiscoveryV1Interface)
+// Source: k8s.io/client-go/kubernetes/typed/events/v1 (interfaces: EventsV1Interface)
```
erezrokah, line 2: This doesn't seem right

plugins/source/k8s/Makefile
```diff
@@ -1,6 +1,7 @@
 # Generate mocks for mock/unit testing 
 .PHONY: generate-mocks
 gen-mocks:
+	go install github.com/golang/mock/mockgen@v1.6.0
```
erezrokah, line 4: We can do this only in the GitHub Action workflow, but seems useful to do it for everyone

erezrokah:
> also if something wasnt generated we will catch it in codegen as mocktests wont be able to work.

Should we run the tests then after the code generation to verify it? Or once before code gen, then another after?

It takes about ~2m to run the code gen (code, mocks and docs) with no cache:
image

And almost nothing with a cache:
image

yevgenypats:
> > also if something wasnt generated we will catch it in codegen as mocktests wont be able to work.
> 
> Should we run the tests then after the code generation to verify it? Or once before code gen, then another after?
> 
> It takes about ~2m to run the code gen (code, mocks and docs) with no cache: image
> 
> And almost nothing with a cache: image

code-gen and then tests.

erezrokah:
> code-gen and then tests.

That might make it hard to understand where failures are coming from, but we can try it out

erezrokah:
@yevgenypats can you take another look? I think this should be ready to merge

erezrokah:
Closing for https://github.com/cloudquery/cloudquery/pull/5252

PR #4667, opened 2022-11-15
feat(k8s-policy): Create CIS Kubernetes Policy v1.6.0

SCKelemen:
Implements the CIS Kubernetes Policy v1.6.0


yevgenypats (changes requested):
Let's remove all manual controls and all printf as it's better just to know that we implement specific sections that are accessible via the k8s API/client and link to the official CIS website to reference all other checks as otherwise we will just need to keep the text up to date all the time while not really providing the value.

Issue #3177, opened 2022-10-23
feat(docs): Bring Azure contributing guide up to speed


erezrokah:
We should update https://github.com/cloudquery/cloudquery/blob/bf4bcbc33a6c8cff75b3a9156052b6df4cf2a16e/plugins/source/azure/CONTRIBUTING.md to reference the new code gen, v1 terms, etc.

PR #4682, opened 2022-11-15
chore: Update Azure contributing guide

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3177. Initial new guide.
There's probably more we can improve both in the guide, and with the codegen, but this should be a good first step


plugins/source/azure/CONTRIBUTING.md
```diff
@@ -1,53 +1,114 @@
...
 
-## Setting up the service
+## 1. Generate Interfaces for the Azure SDK Function(s) that Fetch the Resource
+
+### Before you Start
+
+Inside the root of this repository, run:
+
+```shell
+make install-tools
```
hermanschaaf, line 22: I guess this should be updated now that the root makefile has been removed? Or shall we bring it back, at least just for this command?

hermanschaaf: Ah sorry, I see, the command is in the Azure makefile. Maybe we should just rephrase the above to say `in the azure directory` or something like that

plugins/source/azure/CONTRIBUTING.md
```diff
@@ -1,53 +1,114 @@
...
 
-A few important things to note when adding functions that call the Azure API:
+## General Tips
 
-- If possible, always use an API call that allows you to fetch many resources at once
-- Take pagination into account. Ensure you fetch **all** of the resources
+- Keep transformations to a minimum. As far as possible, we aim to deliver an accurate reflection of what the Azure API provides.
+- We generally only unroll structs one level deep. Nested structs should be transformed into JSON columns.
+- It's recommended to split each resource addition into a separate PR. This makes it easier to review and merge.
+- If the Azure SDK API is not consistent, it's recommended wrap it inside [`client/services`](client/services) to simplify the code generation process. See example [here](client/services/web.go#L61).
```
hermanschaaf, line 112:
```suggestion
- If the Azure SDK API is not consistent, it's recommended to wrap it inside [`client/services`](client/services) to simplify the code generation process. See example [here](client/services/web.go#L61).
```

hermanschaaf (approved):
Looks great! 🥇

erezrokah:
Ah, I'll follow up with another PR with the fixes as this got auto merged 🚤

PR #4691, opened 2022-11-15
chore: Remove IsErrorRegex code

yevgenypats:
Remove some more dead code and make logs consistent.

PR #4692, opened 2022-11-15
chore: Remove dead resolvers

yevgenypats:
remove dead resolvers code and json resolver as this is now being taken care of by CQ Type System


plugins/source/aws/resources/services/cloudformation/stacks_fetch.go
```diff
@@ -11,10 +10,6 @@ import (
 	"github.com/cloudquery/plugin-sdk/schema"
 )
 
-var (
-	validStackNotFoundRegex = regexp.MustCompile("Stack with id (.*) does not exist")
```
disq, line 15: This will conflict if #4691 is merged first

erezrokah (approved):
This looks good to me, but I'll merge with `fix:` so it triggers a release, and we can ship these refactors gradually

hermanschaaf (approved):
Looks good! ✨

PR #4693, opened 2022-11-15
feat(aws): Add Support for Eventbridge scheduler

bbernays:


erezrokah (approved):
You're on 🔥 @bbernays with all the new resources

Issue #4656, opened 2022-11-15
feat(aws): Improve error message when wrong `local_profile` is used


erezrokah:
Related to https://github.com/cloudquery/cloudquery/issues/4627.

If someone uses the wrong `local_profile`, e.g.
```yml
kind: source
spec:
  name: "aws"
  path: "cloudquery/aws"
  version: "v5.1.1" # latest version of aws plugin
  destinations: ["postgresql"]

  spec:
    accounts:
      - local_profile: "<does-not-exists>"
        id: "XXX"
```

They'll get the following error:

```
Loading spec(s) from aws.yml, postgresql.yml
Starting migration for: aws -> [postgresql]
Migration completed successfully.
Starting sync for:  aws -> [postgresql]
Error: failed to sync source aws: failed to fetch resources from stream: rpc error: code = DeadlineExceeded desc = failed to sync resources: failed to create execution client for source plugin aws: error retrieving credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, canceled, context deadline exceeded
```

We should either improve the docs or the error message

bbernays:
There is no difference in error based on the profile not being found. The way the SDK is implemented is that it checks a number of places for credentials and stops as soon as it finds some. So in the case when a profile doesn't exist it will fall back to check the Instance Metadata Service and then fail. In order to improve the error message CloudQuery needs to find the config/credentials files then parse them and find out if a profile with the specified name exists... Also this can be injected via Environment Variables `AWS_PROFILE` so it quickly gets complicated

https://github.com/cloudquery/cloudquery/blob/66e44450166addf87daec2c06231d1815563a5ae/plugins/source/aws/client/client.go#L322

erezrokah:
Thanks @bbernays I can see how validating the profile can get complex 🙃 

Maybe we can try https://github.com/cloudquery/cloudquery/pull/4708 for now

PR #4708, opened 2022-11-16
fix: Improve AWS credentials error message and local profile docs

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/4656


plugins/source/aws/client/client.go
```diff
@@ -321,7 +321,7 @@ func configureAwsClient(ctx context.Context, logger zerolog.Logger, awsConfig *S
 	// Test out retrieving credentials
 	if _, err := awsCfg.Credentials.Retrieve(ctx); err != nil {
 		logger.Error().Err(err).Msg("error retrieving credentials")
-		return awsCfg, fmt.Errorf("error retrieving credentials: %w", err)
+		return awsCfg, fmt.Errorf("error retrieving AWS credentials. Please verify your credentials and try again")
```
erezrokah, line 324: The full AWS error is still in the logs

plugins/source/aws/docs/configuration.md
```diff
@@ -53,7 +53,20 @@ This is used to specify one or more accounts to extract information from. Note t
 
 - `local_profile` (string) (default: will use current credentials)
 
-  [Local profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) to use to authenticate this account with
+  [Local profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) to use to authenticate this account with.
+  Please note this is the name of the profile, for example with the following credentials file:
```
hermanschaaf, line 57:
```suggestion
  Please note this should be set to the name of the profile. For example, with the following credentials file:
```

plugins/source/aws/client/client.go
hermanschaaf, line 324:
```suggestion
		return awsCfg, fmt.Errorf("error retrieving AWS credentials (see logs for details). Please verify your credentials and try again")
```

I'm thinking we should just point users to the log at least, since it's likely to be new users running into this error, and they may not be aware of the log file yet.

hermanschaaf: Yeah, the error is not great in the local profile case, but I guess it might be more enlightening in some other cases (depending on the underlying reason for failure)?

PR #4714, opened 2022-11-16
fix(aws): Remove ARN helper usage

### LINKED ISSUE #4689: refactor(aws): Remove 10 different types of creating ARNs
We need to remove all functional "dry" components creating `arn` as it makes the code hard to maintain and error prone. We only want to use `arn.ARN` without any wrapper as we prefer "KISS" over "DRY" especially when all the functional component dont really make the code shorter but rather creating a maze of functional components that create a string.

disq:
Fixes #4689


erezrokah (approved):
This looks good, but I'd do a `fix` so it triggers a release so we can ship this separately from other changes

plugins/source/aws/resources/services/apigateway/vpc_links_fetch.go
```diff
@@ -26,6 +29,11 @@ func fetchApigatewayVpcLinks(ctx context.Context, meta schema.ClientMeta, parent
...
 	cl := meta.(*client.Client)
 	link := resource.Item.(types.VpcLink)
-	arn := cl.RegionGlobalARN(client.ApigatewayService, "/vpclinks", *link.Id)
-	return resource.Set(c.Name, arn)
+	return resource.Set(c.Name, arn.ARN{
+		Partition: cl.Partition,
+		Service:   string(client.ApigatewayService),
+		Region:    cl.Region,
+		AccountID: "",
+		Resource:  fmt.Sprintf("/vpclinks/%s", aws.ToString(link.Id)),
```
hermanschaaf, line 37: This looks the same as before, but it's a little suspicious that it's the only `Resource` that starts with a forward slash :suspect:

hermanschaaf (approved):

yevgenypats (approved):

PR #4719, opened 2022-11-16
chore: Remove generated files before codegen for Azure

hermanschaaf:
This adds a step to the `gen-code` makefile command to remove all generated files first. This helps avoid issues where generated files are left in the repo despite no longer being part of any recipes.

I've also changed the comment at the top of the file to match that of the AWS source plugin. One reason is consistency, but another is that GoLand seems to use the `// Code generated by ...` string to determine when a file is auto-generated, and then warn the user when they edit it:

Screenshot 2022-11-16 at 11 03 15

I don't know if other IDEs support the same feature, but I think this is quite a nice thing to support.


erezrokah (approved):
🚀

PR #4720, opened 2022-11-16
fix: Remove codegen files before re-generating for GCP

hermanschaaf:
Similar to https://github.com/cloudquery/cloudquery/pull/4719, this adds a step to the gen-code makefile command for GCP to remove all generated files first. This helps avoid issues where generated files are left in the repo despite no longer being part of any recipes. 

There was one such issue which is fixed here. The mock file is different when generated, and it seems like manual edits have been made to it, so I've removed the `// Code generated by codegen; DO NOT EDIT.` comment from the top of the file.


erezrokah (approved):

hermanschaaf:
Changing this to `fix` because the linter is failing due to the use of some deprecated functions, and I'm updating it here to fix the issue.

PR #4721, opened 2022-11-16
fix!: DigitalOcean

shimonp21:
fix(digitalocean)!: This is only a breaking change if you did not use spaces, and start using spaces after this version. Before updating you'll need to drop your database.

- remove PK from `id` column in `digitalocean_space_cors` table
- some README updates.


plugins/source/digitalocean/README.md
```diff
@@ -28,7 +26,7 @@ The following example sets up the DigitalOcean plugin, and connects it to a post
 kind: source
 spec:
   # Source spec section
-  name: okta
+  name: digitalocean
```
erezrokah, line 29: 🚀

erezrokah (approved):
Looks good. Should this be a breaking change? Not sure actually, maybe is someone was relying on the uniqueness of ID?

Issue #2868, opened 2022-10-14
Document Running Plugins Locally


bbernays:
Users want to be able to execute plugins they build locally. our documentation should be clear on how to accomplish that

PR #4725, opened 2022-11-16
fix(website): Add running locally

disq:
Implements #2868


website/pages/docs/developers/running-locally.md
```diff
@@ -0,0 +1,54 @@
...
+/path/to/plugin serve
+```
+
+If you are running multiple plugins this way simultaneously, you will need to specify a different port for each one. You can do this with the `--address` flag:
+
+```bash
+/path/to/plugin serve --address localhost:7778
+```
+
+> 1:16PM INF Source plugin server listening address=127.0.0.1:7778
```
erezrokah, line 41:
```suggestion
> 1:16 PM INF Source plugin server listening address=127.0.0.1:7778
```

website/pages/docs/developers/running-locally.md
```diff
@@ -0,0 +1,54 @@
...
+/path/to/plugin serve
+```
+
+If you are running multiple plugins this way simultaneously, you will need to specify a different port for each one. You can do this with the `--address` flag:
+
+```bash
+/path/to/plugin serve --address localhost:7778
+```
+
+> 1:16 PM INF Source plugin server listening address=127.0.0.1:7778
```
erezrokah, line 41:
```suggestion
> `1:16PM INF Source plugin server listening address=127.0.0.1:7778`
```

website/pages/docs/developers/running-locally.md
disq: But that's not the logline says 😂 Happy to commit that to make Vale happy tho

erezrokah: Added another commit to put it in a code block, but also maybe fix the log line (probably not needed)

yevgenypats (approved):

Issue #4632, opened 2022-11-15
feat(website): Allow searching for supported resources/tables


erezrokah:
This should allow for better discoverability of supported resources than looking at https://github.com/cloudquery/cloudquery/blob/8006c2455f8ecd3255dececa159d2e6ff5c3d1d4/plugins/source/aws/docs/tables/README.md for example.

One way to do it is to create a custom search component instead of the default one (more information in https://github.com/shuding/nextra/issues/107).

https://github.com/cloudquery/plugin-sdk/pull/347 could be handy for this too to update search indices

erezrokah:
The built in Nextra search saves indices under `.next/static/chunks/nextra-data-en-US.json` https://github.com/shuding/nextra/blob/32cd3855cd0ae3c7687423bae6308a70cc28fd4c/packages/nextra-theme-docs/src/components/flexsearch.tsx#L68 so we could add entries to that file.

However I think that won't be that nice to mix site search with table search, so better to create a dedicated search component for tables

erezrokah:
Related to https://github.com/cloudquery/cloudquery/issues/5197

erezrokah:
Tables are now part of the Website, so I'll close this in favor of https://github.com/cloudquery/cloudquery/issues/5197 as I believe that issue should cover both. cc @cqgaurav

PR #4728, opened 2022-11-16
feat(website): Add source plugins tables

erezrokah:
#### Summary

~~DEFINITELY NOT READY YET. More a possible approach to add a tables search without rendering everything.~~

~~Related to https://github.com/cloudquery/cloudquery/issues/4632~~

~~Search is available in https://cloudquery-b47kv5cvg-cloudquery.vercel.app/docs/plugins/sources~~

Adds source plugins tables to our Website instead of redirecting to GitHub


website/components/TableSearch.tsx
```diff
@@ -0,0 +1,74 @@
...
+import Link from "next/link";
+import algoliasearch from "algoliasearch/lite";
+import {
+  InstantSearch,
+  connectSearchBox,
+  connectStateResults,
+} from "react-instantsearch-dom";
+
+const searchClient = algoliasearch(
+  "0OUG5EUZ6H",
```
erezrokah, line 12: I created this Algolia App on a personal account. We can create a CloudQuery one.
This is a client side API key, so safe to put in the code

website/components/TableSearch.tsx
```diff
@@ -0,0 +1,74 @@
...
+  "fa5e5811f7275c275157c74b46f11345"
+);
+
+const SearchBox = ({ refine }) => {
```
erezrokah, line 16: This needs CSS, we can probably reuse https://github.com/shuding/nextra/blob/32cd3855cd0ae3c7687423bae6308a70cc28fd4c/packages/nextra-theme-docs/src/components/search.tsx

website/pages/docs/plugins/sources.mdx
```diff
@@ -3,9 +3,13 @@ title: Sources
 ---
 
 import { getLatestVersion } from "../../../utils/versions";
+import { TableSearch } from "../../../components/TableSearch";
 
 # Source Plugins
 
+<TableSearch />
```
erezrokah, line 10: We cave move this to a better place

website/scripts/tables.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
```
erezrokah, line 2: Here as an example, should be generated using https://github.com/cloudquery/plugin-sdk/pull/347

website/scripts/tables.json
```diff
@@ -0,0 +1,10 @@
...
+    "name": "aws_accessanalyzer_analyzer_archive_rules",
+    "docs_link": "https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/docs/tables/aws_accessanalyzer_analyzer_archive_rules.md"
```
erezrokah, line 4: We'll need to mark `docs_link` as not searchable otherwise `github` will match everything etc.

erezrokah:
Closing as I think there's an easier way to do this

PR #4730, opened 2022-11-16
feat(aws): Add Support For Config Rules and Compliance

bbernays:


plugins/source/aws/codegen/recipes/config.go
```diff
@@ -60,6 +60,25 @@ func ConfigResources() []*Resource {
 					},
 				}...),
 		},
+		{
+			SubService: "config_rules",
+			Struct:     new(types.ConfigRule),
+			ExtraColumns: append(defaultRegionalColumns,
```
hermanschaaf, line 66: I think `ConfigRuleArn` needs to be skipped here so that it's not a duplicate of `arn`

plugins/source/aws/codegen/recipes/config.go
```diff
@@ -60,6 +60,25 @@ func ConfigResources() []*Resource {
...
+						Resolver: `schema.PathResolver("ConfigRuleArn")`,
+						Options:  schema.ColumnCreationOptions{PrimaryKey: true},
+					},
+				}...),
+			Relations: []string{"ConfigRuleCompliances()"},
+		},
+		{
+			SubService:   "config_rule_compliances",
+			Struct:       new(types.ComplianceByConfigRule),
+			ExtraColumns: defaultRegionalColumns,
```
hermanschaaf, line 80: I guess not, but are there any good primary key candidates we can use for this table?

hermanschaaf (changes requested):

plugins/source/aws/codegen/recipes/config.go
bbernays: I thought we didn't put PKs on relation tables

hermanschaaf: We do if there are good candidates, because it's more reliable than `_cq_id`. But if there are no good candidates, we don't.

plugins/source/aws/codegen/recipes/config.go
bbernays: fixed

hermanschaaf (approved):

PR #4773, opened 2022-11-17
fix: Fixed cloudfront policy queries with null json fields

### LINKED ISSUE #4334: GCP CIS 1.2.0 Logging Query Not Configured Across Services and Users
### Describe the Bug
The CIS 1.2.0 Logging Query Not Configured Across Services and Users returns a scalar issue.

Reported on Discord, thanks d0uble3l for reporting this!

`psql:./cis_v1.2.0/../queries/logging/not_configured_across_services_and_users.sql:47: ERROR:  cannot extract elements from a scalar`

### Expected Behavior
CIS 1.2.0 policy should have run successfully without the error.

### CloudQuery Version
GCP CIS Policies 1.2.0

https://github.com/cloudquery/cloudquery/blob/main/plugins/source/gcp/policies_v1/cis_v1.2.0/section_2.sql

amanenk:
#### Summary

closes #4334

Some queries were using `jsonb_array_elements` function that does not work with null `json` fields


disq:
`'null'` or `is null`? do some fields have the string `null` in them, instead of being `null`?

bbernays:
> `'null'` or `is null`? do some fields have the string `null` in them, instead of being `null`?

it is `'null'` it is a nested array... we should probably be putting an empty array in the json rather than a `null`


``` sql
select null::text 
union
select (distribution_config->'OriginGroups'->'Items')::text as ogs from aws_cloudfront_distributions acd

outputs:

"text"
"null"

**Comment by disq** [comment]
> > `'null'` or `is null`? do some fields have the string `null` in them, instead of being `null`?
> 
> it is `'null'` it is a nested array... we should probably be putting an empty array in the json rather than a `null`

Or use `->>` in the second level like so:
```sql
select null::text
union
select (distribution_config->'OriginGroups'->>'Items')::text as ogs from aws_cloudfront_distributions acd;

 text
------

(1 row)
```

```sql
-- with ->>
select (distribution_config->'OriginGroups'->>'Items') is null  as ogs from aws_cloudfront_distributions acd;
 ogs
-----
 t
 t
(2 rows)
```
```sql
-- with ->
select (distribution_config->'OriginGroups'->'Items') is null  as ogs from aws_cloudfront_distributions acd;
 ogs
-----
 f
 f
(2 rows)
```

bbernays:
@disq -I think this scenario is a bit different then that discussion because this CloudQuery making a change to a JSON object:
``` json
{
  "map_attribute": {
     "list1": ["item_1"],
     "empty_list": []     
  }
}
```

PG representation:


``` json
{
  "map_attribute": {
     "list1": ["item_1"],
     "empty_list": null
  }
}
```

disq (approved):

disq:
@bbernays yeah looks like the string match is the better option, `distribution_config->'OriginGroups'` gives me `{"Items": null, "Quantity": 0}` and `->'Items'` returns the string `null` for some reason.

bbernays (approved):

PR #4801, opened 2022-11-18
fix: Fix Name tag reference in grafana dashboard

cloudmatt:
#### Summary

When using the AWS grafana dashboard for that displays public and prviate EC2 instances the 'name' column wouldn't populate with the query provided so after playing around a bit I found an 'easier' way to get the name to appear in the column successfully.

**Current Dashboard Example:**
Screenshot 2022-11-18 at 08 03 51

---

**After applying this PR:**
Screenshot 2022-11-18 at 08 04 28

### DIFF
--- a/plugins/source/aws/dashboards/grafana/aws_ec2_public_private.json
+++ b/plugins/source/aws/dashboards/grafana/aws_ec2_public_private.json
@@ -147,7 +147,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


disq (approved):

disq:
Thanks for the PR! We recently overhauled the `tags` fields (to always be json objects and not arrays, if possible) but the dashboard queries were apparently missed.

hermanschaaf (approved):
Thanks for the fix @cloudmatt! Looks good 👍 

Like @disq mentioned, this was likely due to the recent work to consolidate the format of tags across AWS (https://github.com/cloudquery/cloudquery/pull/3330)

PR #4833, opened 2022-11-20
feat(aws): Add support for xray resource policy

bbernays:


bbernays:
This feature has been released in AWS SDKs but the feature has not been announced and associated documentation has not been released

disq (approved):

PR #4840, opened 2022-11-21
fix: Added password enabled column to `aws_iam_credential_reports`

### LINKED ISSUE #3145: aws_iam_credential_reports `password_status` column should be called `password_enabled`
`password_status` in the `aws_iam_credential_reports` table is called `password_enabled` in the original CSV. Changing it will be a breaking change, so just opening the issue to document it for now.

amanenk:
#### Summary

closes #3145


hermanschaaf (approved):
LGTM - let's maybe just leave comments on the `password_status` and `password_enabled` fields to explain the situation? We can say that `password_enabled` is an alias for the (now deprecated) `password_status` and link to this ticket: https://github.com/cloudquery/cloudquery/issues/3145

PR #4870, opened 2022-11-21
fix: Region Ordering Failures

bbernays:
#### Summary

If a user specifies a list of regions, we should try and get all Enabled regions from every specified region rather than just trying the first one...


hermanschaaf (approved):
🏃 ✨

plugins/source/aws/client/client.go
```diff
@@ -368,25 +368,24 @@ func Configure(ctx context.Context, logger zerolog.Logger, spec specs.Source) (s
...
-		if err != nil {
-			logger.Warn().Str("account", account.AccountName).Err(err).Msg("Failed to find disabled regions for account")
-			continue
+		for _, region := range regionsToCheck {
+			enabledRegions, err := getEnabledRegions(ctx, awsCfg, region)
+			if err != nil {
+				logger.Warn().Str("account", account.AccountName).Err(err).Msgf("Failed to find disabled regions for account when checking: %s", region)
+				continue
+			}
+			account.Regions = filterDisabledRegions(localRegions, enabledRegions)
```
hermanschaaf, line 387: Should we maybe break out of the loop here? Otherwise we'll do it for all regions even though only the last (working) result will count

bbernays: Fixed

yevgenypats (approved):

hermanschaaf (approved):

PR #4882, opened 2022-11-21
feat(aws): Validate Service Multiplexer

bbernays:
#### Summary

During codegen if a regional service multiplexer is used validate that it is a valid service name in `plugins/source/aws/client/data/partition_service_region.json`


plugins/source/aws/codegen/recipes/base.go
```diff
@@ -418,3 +425,42 @@ func CreateReplaceTransformer(replace map[string]string) func(field reflect.Stru
...
+func validateServiceMultiplex(multiplexerCall string) error {
+	re := regexp.MustCompile(`\"(.*?)\"`)
+	// Find the value of the service parameter
+	submatchAll := re.FindStringSubmatch(multiplexerCall)
+	t := client.ReadSupportedServiceRegions()
+
+	var services []string
+	for _, partition := range t.Partitions {
+		for service := range partition.Services {
+			services = appendIfMissing(services, service)
```
disq, line 438: Does using a map here make it more complex? Could then use some `funk` functs and get rid of the helpers?

bbernays: Good call! Didn't even need the `funk`... In the original implementation I had tried to use the generic functions in `go/map` and `go/slice` but in the end it devolved into recreating set functionality

bbernays: fixed

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -418,3 +424,27 @@ func CreateReplaceTransformer(replace map[string]string) func(field reflect.Stru
...
+	services := make(map[string]bool)
+
+	for _, partition := range t.Partitions {
+		for service := range partition.Services {
+			if _, ok := services[service]; !ok {
+				services[service] = true
+			}
+		}
+	}
+	if len(submatchAll) == 2 {
```
disq, line 443: For clarity why not run this check in line 432 (before the call to `ReadSupportedServiceRegions`) for a chance of early return? I don't think we actually have anything with no multiplexer set but it still makes sense: then here you don't need the `if` and it's all straightforward.

disq (changes requested):

plugins/source/aws/codegen/recipes/base.go
bbernays: Great call!... If a resource uses an Account multiplexer it doesn't have a service...

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -418,3 +424,30 @@ func CreateReplaceTransformer(replace map[string]string) func(field reflect.Stru
...
```
disq, line 446: You don't need this `if`

disq (changes requested):

plugins/source/aws/codegen/recipes/base.go
```diff
@@ -418,3 +424,28 @@ func CreateReplaceTransformer(replace map[string]string) func(field reflect.Stru
...
+			if _, ok := services[service]; !ok {
+				services[service] = true
+			}
+		}
+	}
+	if _, ok := services[submatchAll[1]]; !ok {
+		return fmt.Errorf("invalid partition: %s", submatchAll[1])
+	}
+	return nil
+
```
disq, line 450: nit: rogue newline (or move L449 to here and have the newline after the if block)

PR #4883, opened 2022-11-21
feat(aws): Add Support for Cloudwatch Logs Resource Policy

bbernays:


plugins/source/aws/resources/services/cloudwatchlogs/resource_policies.go
```diff
@@ -0,0 +1,53 @@
...
+			},
+			{
+				Name:     "last_updated_time",
+				Type:     schema.TypeInt,
+				Resolver: schema.PathResolver("LastUpdatedTime"),
+			},
+			{
+				Name:     "policy_document",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("PolicyDocument"),
```
disq, line 41: `PolicyDocument`s are sometimes `schema.TypeJSON` and not `schema.TypeString` (we both have it in different resources, not sure about the reason of the inconsistency) did/can you check if using TypeJSON would be better here or not?

hermanschaaf (approved):

plugins/source/aws/codegen/recipes/cloudwatchlogs.go
```diff
@@ -8,6 +8,13 @@ import (
 
 func CloudWatchLogsResources() []*Resource {
 	resources := []*Resource{
+		{
+			SubService:   "resource_policies",
+			Struct:       &types.ResourcePolicy{},
+			Description:  "https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_ResourcePolicy.html",
+			PKColumns:    []string{"policy_name"},
+			ExtraColumns: defaultRegionalColumnsPK,
```
hermanschaaf, line 16: Nit: I think using 

```
			PKColumns:    []string{"account_id", "region", "policy_name"},
			ExtraColumns: defaultRegionalColumns,
```

would make the actual PKs a bit clearer.

bbernays: fixed

disq (approved):

PR #4901, opened 2022-11-22
fix(aws): Typo in lambda query

altho1:
#### Summary

Replaced "ec2 instances" with "lambda functions" in plugins/source/aws/policies/queries/lambda/functions_with_public_egress.sql
Just to make it clear to the receivers of the data.


hermanschaaf (approved):
Thanks for the fix! 👍

disq (approved):

PR #4902, opened 2022-11-22
feat: New destination file plugin

yevgenypats:
Fixes https://github.com/cloudquery/cloudquery/issues/4983

New destination plugin `file` will support multiple format (csv, json) with multiple storage backends such as local, S3, GCS, Azure Blob Storage.

This is ready now for initial review.


plugins/destination/file/client/client.go
```diff
@@ -0,0 +1,112 @@
...
+
+	switch c.csvSpec.Backend {
+	case BackendTypeLocal:
+		if err := os.MkdirAll(c.csvSpec.Directory, 0755); err != nil {
+			return nil, fmt.Errorf("failed to create directory: %w", err)
+		}
+	case BackendTypeS3:
+		awsCfg, err := config.LoadDefaultConfig(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("unable to load AWS SDK config, %w", err)
```
disq, line 72:
```suggestion
			return nil, fmt.Errorf("unable to load AWS SDK config: %w", err)
```
err conversion is `wrapper text: %w`

plugins/destination/file/client/client.go
```diff
@@ -0,0 +1,112 @@
...
+			Key:    aws.String(c.path + "/cq-test-file"),
+			Body:   bytes.NewReader([]byte("test")),
+		}); err != nil {
+			return nil, fmt.Errorf("failed to write test file to S3: %w", err)
+		}
+	case BackendTypeGCS:
+		var err error
+		c.gcpStorageClient, err = storage.NewClient(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create GCP storage client %w", err)
```
disq, line 89:
```suggestion
			return nil, fmt.Errorf("failed to create GCP storage client: %w", err)
```

plugins/destination/file/client/client.go
```diff
@@ -0,0 +1,112 @@
...
+	case BackendTypeS3:
+		awsCfg, err := config.LoadDefaultConfig(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("unable to load AWS SDK config, %w", err)
+		}
+		awsClient := s3.NewFromConfig(awsCfg)
+		c.awsUploader = manager.NewUploader(awsClient)
+		c.awsDownloader = manager.NewDownloader(awsClient)
+
+		if _, err := c.awsUploader.Upload(ctx, &s3.PutObjectInput{
```
disq, line 78: Test objects shouldn't be part of a "serious" release but they can stay for now.

plugins/destination/file/client/json.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (c *Client) writeJSONResource(ctx context.Context, tableName string, resources <-chan []interface{}) error {
+	f, err := c.openAppendOnly(ctx, tableName+".json")
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	for r := range resources {
+		b, err := json.Marshal(r)
```
disq, line 22: In the future a drop in encoding/json replacement (eg. `fastjson`) can be used here.

plugins/destination/file/client/json.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+const maxJsonSize = 1024 * 1024 * 20
```
disq, line 12: This is only used for reading/scanner yet the name doesn't suggest that. (Could also move the const defn inside the read method)

plugins/destination/file/go.mod
```diff
@@ -0,0 +1,68 @@
...
+	github.com/aws/aws-sdk-go-v2 v1.17.1
+	github.com/aws/aws-sdk-go-v2/config v1.18.2
+	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.11.41
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.29.3
+	github.com/cloudquery/plugin-sdk v1.6.0
+	github.com/google/uuid v1.3.0
+	github.com/rs/zerolog v1.28.0
+)
+
+replace github.com/cloudquery/plugin-sdk => ../../../../plugin-sdk-split
```
disq, line 16: This needs to be removed before merge

plugins/destination/file/main.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/destination/file/client"
+	"github.com/cloudquery/cloudquery/plugins/destination/file/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/plugins"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const (
+	sentryDSN = "https://79d5e237dafe45e1a4ec0785bc528280@o1396617.ingest.sentry.io/4504083471335424"
```
disq, line 11: Consider replacing this before release (belongs to the CSV dest)

disq (changes requested):
Good start, few nits and remarks.

Should consider adding in-line gzip and combined-file support once it's reviewed/merged/refactored before that blog post though :-) As you'll soon see high number of tables will be an issue when using with data lakes.

yevgenypats:
Related Issue: https://github.com/cloudquery/cloudquery/issues/4983

plugins/destination/file/internal/backends/azure-blob-storage/file.go
```diff
@@ -0,0 +1,124 @@
+package azure_blob_storage
```
yevgenypats, line 1: can we keep the directory also `azure_blog_storage` instead of `azure-blog-storage` ?

hermanschaaf: Fixed

plugins/destination/file/README.md
```diff
@@ -0,0 +1,31 @@
...
+  name: "csv"
+  path: "cloudquery/csv"
+  version: "v1.0.1" # latest version of csv plugin
+  write_mode: "append" # CSV only supports 'append' mode
+
+  spec:
+    directory: './cq_csv_output'
+```
+
+## CSV Spec
```
erezrokah, line 25:
```suggestion
## File Spec
```

plugins/destination/file/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This is the (nested) spec used by the CSV destination Plugin.
+
+- `directory` (string) (optional, defaults to `./cq_csv_output`)
```
erezrokah, line 29:
```suggestion
- `directory` (string) (optional, defaults to `./cq_file_output`)
```

plugins/destination/file/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+  Directory where all CSV files will be written. A CSV file will be created per table.
```
erezrokah, line 31:
```suggestion
  Directory where all files will be written. A file will be created per table.
```

erezrokah (changes requested):
Added a few comments. I would also verify the steps in https://www.notion.so/cloudquery/Adding-a-new-plugin-to-the-monorepo-f216b653dbe648b2b3512fb8d59a8f89

plugins/destination/file/client/csv_transformer.go
```diff
@@ -0,0 +1,71 @@
+package client
+
+import "github.com/cloudquery/plugin-sdk/schema"
+
+func (*Client) transformCSVBool(v *schema.Bool) interface{} {
```
candiduslynx, line 5: `interface{}` -> `any` (it's alias, so do it for readability only)

plugins/destination/file/client/csv_transformer.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (*Client) transformCSVInetArray(v *schema.InetArray) interface{} {
+	return v.String()
+}
+
+func (*Client) transformCSVMacaddr(v *schema.Macaddr) interface{} {
+	return v.String()
+}
+
+func (*Client) transformCSVMacaddrArray(v *schema.MacaddrArray) interface{} {
```
candiduslynx, line 69: can it be just a single func?
```go
func (*Client) transform(v fmt.Stringer) any {
	return v.String()
]
```
(even more. so, make ret value a string in signature?

plugins/destination/file/client/csv_transformer.go
candiduslynx: I suspect, in this case it even can be ditched entirely

plugins/destination/file/client/deletestale.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (*Client) DeleteStale(ctx context.Context, tables schema.Tables, sourceName string, syncTime time.Time) error {
+	return fmt.Errorf("csv destination doesn't support overwrite-delete-stale mode. please use append mode")
```
candiduslynx, line 12:
```suggestion
	return fmt.Errorf("csv destination doesn't support overwrite-delete-stale mode, use append mode instead")
```

plugins/destination/file/client/json.go
candiduslynx: +1

plugins/destination/file/client/json.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (c *Client) writeJSONResource(ctx context.Context, table *schema.Table, resources <-chan []interface{}) error {
```
candiduslynx, line 14: interface{} -> any

plugins/destination/file/client/json.go
```diff
@@ -0,0 +1,67 @@
...
+	f, err := c.openAppendOnly(ctx, table.Name+".json")
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	for r := range resources {
+		jsonObj := make(map[string]interface{}, len(table.Columns))
+		for i := range r {
```
candiduslynx, line 23: use `for idx, val` form, maybe iterating on columns instead
```go
for idx, col := range table.Columns {
	jsonObj[col.Name] = r[idx]
}
```

plugins/destination/file/client/json.go
```diff
@@ -0,0 +1,67 @@
...
+		return fmt.Errorf("could not find column %s in table %s", schema.CqSourceNameColumn.Name, table.Name)
+	}
+	name := table.Name + ".json"
+	f, err := c.openReadOnly(ctx, name)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
```
candiduslynx, line 53: why not [json.Decoder](https://pkg.go.dev/encoding/json#Decoder)?

plugins/destination/file/client/json_transformer.go
```diff
@@ -0,0 +1,105 @@
...
+func (*Client) transformJSONInt8Array(v *schema.Int8Array) interface{} {
+	res := make([]string, len(v.Elements))
+	for i, e := range v.Elements {
+		res[i] = e.String()
+	}
+	return res
+}
+
+func (*Client) transformJSONJSON(v *schema.JSON) interface{} {
+	return map[string]interface{}{}
```
candiduslynx, line 32: needs impl?

plugins/destination/file/client/json_transformer.go
```diff
@@ -0,0 +1,105 @@
...
+	// 	return nil
+	// }
+	// var res map[string]interface{}
+	// if err := json.Unmarshal(v.Bytes, &res); err != nil {
+	// 	panic(err)
+	// }
+	// return res
+}
+
+func (*Client) transformJSONText(v *schema.Text) interface{} {
```
candiduslynx, line 43: interface{} -> any

plugins/destination/file/client/json_transformer.go
```diff
@@ -0,0 +1,105 @@
...
+	return res
+}
+
+func (*Client) transformJSONCIDR(v *schema.CIDR) interface{} {
+	return v.String()
+}
+
+func (*Client) transformJSONCIDRArray(v *schema.CIDRArray) interface{} {
+	res := make([]string, len(v.Elements))
+	for i, e := range v.Elements {
```
candiduslynx, line 77: this pattern can be extracted into a separate func (used 5+ times)

plugins/destination/file/client/transformer.go
```diff
@@ -0,0 +1,190 @@
+package client
+
+import "github.com/cloudquery/plugin-sdk/schema"
+
+func (c *Client) TransformBool(v *schema.Bool) interface{} {
```
candiduslynx, line 5: can these functions be only exported as `Transform(v schema.CQType)` that will do a type switch instead?

plugins/destination/file/client/write.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/cloudquery/plugin-sdk/plugins"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+type worker struct {
+	writeChan chan []interface{}
+}
+
+func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *plugins.ClientResource) error {
+	wg := sync.WaitGroup{}
```
candiduslynx, line 16:
```suggestion
	wg := new(sync.WaitGroup)
```

plugins/destination/file/client/write.go
```diff
@@ -0,0 +1,49 @@
...
+	workers := make(map[string]*worker, len(tables))
+	for _, t := range tables.FlattenTables() {
+		t := t
+		workers[t.Name] = &worker{
+			writeChan: make(chan []interface{}),
```
candiduslynx, line 21:
```suggestion
			writeChan: make(chan any),
```

plugins/destination/file/client/write.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		wg.Add(1)
+		switch c.csvSpec.Format {
+		case FormatTypeCSV:
+			go func() {
+				defer wg.Done()
+				c.writeCSVResource(ctx, t.Name, workers[t.Name].writeChan)
```
candiduslynx, line 28: don't do a lookup, just do
```go
worker := &worker{writeChan: make(chan any)}
workers[t.Name] = worker
```
above and then just use `worker`

plugins/destination/file/client/write.go
```diff
@@ -0,0 +1,49 @@
...
+				defer wg.Done()
+				c.writeJSONResource(ctx, t, workers[t.Name].writeChan)
+			}()
+		default:
+			panic("unknown format type")
+		}
+	}
+
+	for r := range res {
+		workers[r.TableName].writeChan <- r.Data
```
candiduslynx, line 41: shouldn't it be do with in parallel?

plugins/destination/file/go.mod
candiduslynx: +1

candiduslynx (changes requested):

yevgenypats:
Closing. Done here: https://github.com/cloudquery/cloudquery/pull/6096

Issue #4795, opened 2022-11-18
Gandi as a plugin


cqgaurav:
This has been requested by our customers, and widely supported.

disq:
https://api.gandi.net/docs/

PR #4940, opened 2022-11-22
feat(gandi): Gandi Source Plugin

disq:
Implements #4795 

Needs a new project in https://sentry.io/organizations/cloudquery-v2/projects/


plugins/source/gandi/codegen/recipes/certificates.go
```diff
@@ -0,0 +1,35 @@
...
+
+func CertificateResources() []Resource {
+	return []Resource{
+		{
+			DataStruct:   &certificate.CertificateType{},
+			PKColumns:    []string{"id"},
+			ExtraColumns: []codegen.ColumnDefinition{SharingIDColumn},
+			TableName:    "gandi_certificates",
+
+			Template:         "resource_manual",
```
hermanschaaf, line 16: Nit: I think we can call this template `resource` and have it be the default everywhere, since there are no other options right now anyway.

plugins/source/gandi/codegen/recipes/livedns.go
```diff
@@ -0,0 +1,43 @@
...
+			PKColumns:    []string{"fqdn"},
+			ExtraColumns: []codegen.ColumnDefinition{SharingIDColumn},
+			Relations:    []string{"DomainSnapshots()"},
+			TableName:    "gandi_livedns_domains",
+
+			Template:         "resource_manual",
+			Package:          "livedns",
+			TableFuncName:    "Domains",
+			Filename:         "domains.go",
+			ResolverFuncName: "fetchDomains",
```
hermanschaaf, line 22: I think most of these fields should ideally be inferred from a single entry and changed to the right case using `caser`, like `SubService` in the aws plugin case.

hermanschaaf (approved):
Nice, LGTM for a first version 👍 I left some non-blocking comments.

I think to release this we'll need some additional updates to the GitHub workflows, but I'm happy for that to be done in a separate PR (probably better that way too).

plugins/source/gandi/codegen/recipes/livedns.go
disq: I agree these were copied from the cloudflare plugin with minor modifications, when we have a more generic/basic codegen for small plugins we can replace these in all plugins.

plugins/source/gandi/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/gandi/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+// const sentryDSN = "https://6beffd303fbd4661a7b7c6c2d546b580@o1396617.ingest.sentry.io/6747634" TODO
```
disq, line 8: This requires updating once we create the new project.

plugins/source/gandi/codegen/main.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, c := range r.Table.Columns {
+		if strings.HasPrefix(c.Resolver, "client.") {
+			r.ImportClient = true
+		}
+	}
+
+	mainTemplate := r.Template + ".go.tpl"
+	tpl, err := template.New(mainTemplate).ParseFS(templatesFS, "templates/"+mainTemplate)
+	if err != nil {
+		log.Fatal(fmt.Errorf("failed to parse gandi templates: %w", err))
```
cqgaurav, line 74: Not blocking this PR, but is there some value in sharing this method (by parameterizing the name of the plugin) between all plugins ?

disq: @cqgaurav How we do things change between plugins especially with complex SDKs. It would mean lots of "flags" to understand and configure per plugin... Maybe a very limited set can be "copied" and configured per plugin (ie. this one was copied from the Cloudflare codegen and slightly modified). Eventually parts of it can be extracted as utilities and put into sdk the codegen package which provides main codegen functionality.

plugins/source/gandi/codegen/recipes/livedns.go
disq: Fixed

plugins/source/gandi/main.go
hermanschaaf, line 8:
```suggestion
const sentryDSN = "https://45b253f7a0794df2bba513eaa930121b@o1396617.ingest.sentry.io/4504214828875776"
```

Here is a new one for the Gandi plugin

hermanschaaf: See this comment https://github.com/cloudquery/cloudquery/pull/4940#discussion_r1031644219

plugins/source/gandi/client/spec.go
```diff
@@ -0,0 +1,10 @@
+package client
+
+type Spec struct {
+	ApiKey     string `json:"api_key,omitempty"`
```
hermanschaaf, line 4:
```suggestion
	APIKey     string `json:"api_key,omitempty"`
```

(nit)

plugins/source/gandi/codegen/main.go
```diff
@@ -0,0 +1,109 @@
...
+	var resources []*recipes.Resource
+	resources = append(resources, recipes.CertificateResources()...)
+	resources = append(resources, recipes.DomainResources()...)
+	resources = append(resources, recipes.LiveDNSResources()...)
+	resources = append(resources, recipes.SimpleHostingResources()...)
+
+	for _, r := range resources {
+		r.Infer()
+	}
+	if err := recipes.SetParentChildRelationships(resources); err != nil {
```
hermanschaaf, line 32: Unless we have resolver/mock templates that handle parent & child relations in this plugin (do we?), we can probably remove this call and the associated function.

plugins/source/gandi/docs/tables/gandi_certificate_packages.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|sharing_id|String|
```
hermanschaaf, line 15: Should `sharing_id` maybe be part of the PK here?

plugins/source/gandi/docs/tables/gandi_certificates.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+## Columns
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|sharing_id|String|
```
hermanschaaf, line 15: Should `sharing_id` be part of the PK here?

plugins/source/gandi/main.go
```diff
@@ -0,0 +1,12 @@
...
+
+func main() {
+	serve.Source(plugin.Plugin(), serve.WithSourceSentryDSN(""))
```
hermanschaaf, line 11: Also remember to use the constant value here ☝️

hermanschaaf (changes requested):
Now looks even better 👍 Just left a few comments for things to double-check

plugins/source/gandi/docs/tables/gandi_certificate_packages.md
disq: This is a "global" table with values like:
```
sharing_id      |
name            | cert_std_3_0_0
name_label      |
href            | https://api.gandi.net/v5/certificate/packages/cert_std_3_0_0
max_domains     | 3
type            |
type_label      |
wildcard        | f
```
it would be the same rows repeated if `sharing_id` was a key. The `sharing_id` column is provided because if the user sets it in the config it's passed down to the client automatically so filtering is probably done on the server side.

plugins/source/gandi/docs/tables/gandi_certificates.md
disq: I don't think certificates have clashing ids to warrant for an additional PK.

plugins/source/gandi/codegen/main.go
disq: It's used to properly generate child table names, completely optional, but gandi doesn't have 100% of resources covered (or the sdk has 100% implemented) so more resources will pop up and this could help us down the line (we'll eventually have more then a few child tables)

hermanschaaf (approved):
👍

yevgenypats (changes requested):
I didn't review the code but I think it is good for first version and @hermanschaaf already kindly reviewed. I think you can add to this PR https://github.com/cloudquery/cloudquery/blob/main/contributing/adding_a_new_plugin_to_cq_monorepo.md and then we can merge

disq:
@yevgenypats done

yevgenypats (approved):

PR #5065, opened 2022-11-24
chore: Try to fix/simplify snowflake ci

yevgenypats:
Im not sure what's happening  (maybe Github runners are down) but in any case I wanted to try and simplify the workflows which are starting to have a lot of know how, so going to try and play with a simple workflow for snowflake to see how it work and if it works well we can use it for all others.

### DIFF
--- a/.github/workflows/dest_snowflake.yml
+++ b/.github/workflows/dest_snowflake.yml
@@ -13,30 +13,10 @@ on:
       - ".github/workflows/dest_snowflake.yml"
 
 jobs:
-  resolve-runner:
-    timeout-minutes: 5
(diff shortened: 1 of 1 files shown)


yevgenypats:
Okay found out it wasn't because of anything on automation side (even though Ill do some more work to be able to run some steps locally and potentially simplify some of the tasks). 

This is related to https://github.com/mattn/go-ieproxy/issues/45 which needs CGO_ENABLED=1 on windows so I'll revert that for now in a follow-up PR and make it work on windows with CGO until this issue is resolved - https://github.com/snowflakedb/gosnowflake/issues/690

Issue #4964, opened 2022-11-23
bug(aws): Failed to refresh cached credentials, operation error STS: AssumeRole


hermanschaaf:
Originally reported [here](https://github.com/cloudquery/cloudquery/issues/4834#issuecomment-1324623524). This was about 6 hours into the run:

```
2022-11-23T06:07:14Z ERR column resolver finished with error error="operation error IoT: ListTagsForResource, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 4415c3fc-eb88-4b2d-b955-1dbe40693f78, api error ExpiredToken: The security token included in the request is expired" client=059487:eu-west-1 module=aws-src table=aws_iot_policies
2022-11-23T06:07:14Z ERR table resolver finished with error error="operation error IoT: GetPolicy, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 4415c3fc-eb88-4b2d-b955-1dbe40693f78, api error ExpiredToken: The security token included in the request is expired" client=059487:eu-west-1 module=aws-src table=aws_iot_policies
2022-11-23T06:07:14Z INF fetch table finished client=059487:eu-west-1 errors=2 module=aws-src resources=121169 table=aws_iot_policies
2022-11-23T06:07:14Z ERR column resolver finished with error error="operation error IoT: ListAttachedPolicies, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 35e60035-f79c-4449-9ee7-cf4a2931eafd, api error ExpiredToken: The security token included in the request is expired" client=059487:us-east-1 module=aws-src table=aws_iot_certificates
2022-11-23T06:07:14Z ERR table resolver finished with error error="operation error IoT: DescribeCertificate, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 35e60035-f79c-4449-9ee7-cf4a2931eafd, api error ExpiredToken: The security token included in the request is expired" client=059487:us-east-1 module=aws-src table=aws_iot_certificates
2022-11-23T06:07:14Z INF fetch table finished client=059487:us-east-1 errors=2 module=aws-src resources=198702 table=aws_iot_certificates
[...]

PR #5097, opened 2022-11-25
fix: Add refresh window and jitter for cached credentials

hermanschaaf:
This is meant to address https://github.com/cloudquery/cloudquery/issues/4964. It enables the credentials cache to refresh its credentials with the underlying provider before the hard timeout is reached.

The exact error reported in the issue was:

```
failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: 4415c3fc-eb88-4b2d-b955-1dbe40693f78, api error ExpiredToken: The security token included in the request is expired
```

With emphasis on the "failed to refresh cached credentials" part and the **ExpiredToken** error, this should address a potential race condition with expiring credentials that can cause the refresh request to fail unexpectedly due to ExpiredToken exceptions.

The underlying issue in the original ticket could be something else (I have so far been unable to reproduce it), but this change seems like a good thing to include in any case.

### DIFF
--- a/plugins/source/aws/client/client.go
+++ b/plugins/source/aws/client/client.go
@@ -249,13 +249,11 @@ func configureAwsClient(ctx context.Context, logger zerolog.Logger, awsConfig *S
 		config.WithDefaultRegion(defaultRegion),
 		// https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/
 		config.WithRetryer(func() aws.Retryer {
-			// return retry.NewAdaptiveMode()
 			return retry.NewStandard(func(so *retry.StandardOptions) {
 				so.MaxAttempts = maxAttempts
 				so.MaxBackoff = time.Duration(maxBackoff) * time.Second
 				so.RateLimiter = &NoRateLimiter{}
 			})
-			// return retry.AddWithMaxAttempts(retry.NewStandard(), 5)
 		}),
 	}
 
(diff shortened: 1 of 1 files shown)


plugins/source/aws/client/client.go
```diff
@@ -319,7 +331,7 @@ func Configure(ctx context.Context, logger zerolog.Logger, spec specs.Source) (s
 	client := NewAwsClient(logger)
 	var adminAccountSts AssumeRoleAPIClient
 	if awsConfig.Organization != nil && len(awsConfig.Accounts) > 0 {
-		return nil, errors.New("specifying accounts via both the Accounts and Org properties is not supported. If you want to do both, you should use multiple provider blocks")
+		return nil, errors.New("specifying accounts via both the Accounts and Org properties is not supported. To achieve both, use multiple source configurations")
```
hermanschaaf, line 334: Just updating some outdated wording from pre-v1 days

plugins/source/aws/client/client.go
```diff
@@ -249,13 +249,11 @@ func configureAwsClient(ctx context.Context, logger zerolog.Logger, awsConfig *S
 		config.WithDefaultRegion(defaultRegion),
 		// https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/
 		config.WithRetryer(func() aws.Retryer {
-			// return retry.NewAdaptiveMode()
 			return retry.NewStandard(func(so *retry.StandardOptions) {
 				so.MaxAttempts = maxAttempts
 				so.MaxBackoff = time.Duration(maxBackoff) * time.Second
 				so.RateLimiter = &NoRateLimiter{}
 			})
-			// return retry.AddWithMaxAttempts(retry.NewStandard(), 5)
```
hermanschaaf, line 258: Removing some unused code

yevgenypats (approved):

plugins/source/aws/client/client.go
```diff
@@ -287,12 +285,26 @@ func configureAwsClient(ctx context.Context, logger zerolog.Logger, awsConfig *S
...
+		awsCfg.Credentials = aws.NewCredentialsCache(provider, func(options *aws.CredentialsCacheOptions) {
+			// ExpiryWindow will allow the credentials to trigger refreshing prior to
+			// the credentials actually expiring. This is beneficial so race conditions
+			// with expiring credentials do not cause requests to fail unexpectedly
+			// due to ExpiredToken exceptions.
+			//
+			// An ExpiryWindow of 1 minute would cause calls to IsExpired() to return true
+			// 1 minute before the credentials are actually expired. This can cause an
+			// increased number of requests to refresh the credentials to occur.
+			options.ExpiryWindow = 1 * time.Minute
```
hermanschaaf, line 303:
```suggestion
			// ExpiryWindow will allow the credentials to trigger refreshing prior to
			// the credentials actually expiring. This is beneficial so race conditions
			// with expiring credentials do not cause requests to fail unexpectedly
			// due to ExpiredToken exceptions.
			//
			// An ExpiryWindow of 5 minute would cause calls to IsExpired() to return true
			// 5 minutes before the credentials are actually expired. This can cause an
			// increased number of requests to refresh the credentials to occur. We balance this with jitter.
			options.ExpiryWindow = 5 * time.Minute
```

Issue #4770, opened 2022-11-17
Cloudquery deletes data from other tables if provided list of tables in source config


rajeshwarakella1994:
By default CQ runs for all tables but when a configuration as mentioned below is provided, data is getting deleted from other tables as well.

This is the case with overwrite-delete-stale mode in postgres destination.
Sharing the config files below - 

**aws.yml**
```yml
kind: source
spec:
  name: "aws_667292283479"
  path: "cloudquery/aws"
  version: "v5.1.2"
  destinations: ["postgresql"]
  tables: 
  - aws_rds_engine_versions
  - aws_rds_cluster_parameters
```

**rds.yml**
```yml
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v1.7.10" 
  write_mode: "overwrite-delete-stale" # overwrite, overwrite-delete-stale, append
```

rajeshwarakella1994:
Discord thread here - https://discord.com/channels/872925471417962546/1042796460342775909

cloudmatt:
I _believe_ i'm seeing the same issue as well but with the GCP plugin. I have two 'instances' of the gcp plugin run (see config snippet below) but I only seem to be capturing a subset of the tables. I know it's not a permissions issue because this worked fine before I split out the two sources.

```yaml
---
kind: source
spec:
  name: gcp-all
  path: cloudquery/gcp
  version: "v3.0.5"
  tables: ["*"]
  skip_tables:
    - "gcp_compute_disk_types"
  destinations: ["postgresql"]
  spec:
    folder_ids:
     - "folders/123456789"
    project_ids:
      - projectA
      - projectB
---
kind: source
spec:
  name: gcp-bq-exception
  path: cloudquery/gcp
  version: "v3.0.5"
  tables: ["*"]
  skip_tables:
    - "gcp_compute_disk_types"
    - "gcp_bigquery_datasets"
    - "gcp_bigquery_tables"
  destinations: ["postgresql"]
  spec:
    project_ids:
      - projectC
      - projectD
---
kind: destination
spec:
  name: postgresql
  path: cloudquery/postgresql
  version: "v1.7.14"
  write_mode: overwrite-delete-stale
  spec:
    connection_string: ${CQ_DSN}
```

and the query
```sql
select count(_cq_table), _cq_table from gcp_resources group by _cq_table;
```

only returns
[...]

PR #5098, opened 2022-11-25
fix: Delete stale entries only for tables selected in config

hermanschaaf:
Depends on this plugin-sdk change: https://github.com/cloudquery/plugin-sdk/pull/440/

Fixes https://github.com/cloudquery/cloudquery/issues/4770


cli/cmd/sync2.go
```diff
@@ -29,13 +30,15 @@ func syncConnectionV2(ctx context.Context, cqDir string, sourceClient *clients.S
...
+	tables, err := sourceClient.GetTablesForSpec(ctx, &sourceSpec)
 	if err != nil {
 		return fmt.Errorf("failed to get tables for source %s: %w", sourceSpec.Name, err)
 	}
 
+	tableCount := len(schema.Tables(tables).FlattenTables())
+	fmt.Printf("Source %s will sync %d tables.\n", sourceSpec.Name, tableCount)
 	if !noMigrate {
-		fmt.Println("Starting migration for:", sourceSpec.Name, "->", sourceSpec.Destinations)
+		fmt.Printf("Starting migration for: %s -> %s\n", sourceSpec.Name, sourceSpec.Destinations)
```
disq, line 41: Just `\n`? Windows CLI may not look good

hermanschaaf: I can revert this since I ended up not really changing this line anyway, but I believe Go's implementation of `Println` also just appends `\n`, so we'd be in good company: https://github.com/golang/go/blob/master/src/fmt/print.go#L1225

cli/cmd/sync2.go
```diff
@@ -29,18 +30,39 @@ func syncConnectionV2(ctx context.Context, cqDir string, sourceClient *clients.S
 	}
 	defer destClients.Close()
 
-	tables, err := sourceClient.GetTables(ctx)
+	allTables, err := sourceClient.GetTables(ctx)
```
yevgenypats, line 33: I think this line is not needed because your call `GetTables` inside `getTablesForSpec`

cli/cmd/migrate.go
```diff
@@ -96,12 +97,15 @@ func migrateConnection(ctx context.Context, cqDir string, sourceSpec specs.Sourc
 	}
 	defer destClients.Close()
 
-	fmt.Println("Starting migration for:", sourceSpec.Name, "->", sourceSpec.Destinations)
-	log.Info().Str("source", sourceSpec.Name).Strs("destinations", sourceSpec.Destinations).Msg("Starting migration")
-	tables, err := sourceClient.GetTables(ctx)
+	tables, _, err := getTablesForSpec(ctx, sourceClient, sourceSpec)
```
yevgenypats, line 100: I think we don't want to call this here (yet).

yevgenypats (changes requested):
few comments

cli/cmd/sync2.go
hermanschaaf: We do need it because we still want Migrate to get the full list of tables. `getTablesForSpec` only returns all tables if the `GetTablesForSpec` method is not available.

yevgenypats (approved):

PR #5102, opened 2022-11-25
feat: BigQuery destination plugin

hermanschaaf:
This adds an initial version of a BigQuery destination plugin.

- [x] Support `append` write mode
- [x] Support time-based partitioning 
- [x] Documentation
- [x] Integration tests in CI


.github/workflows/dest_bigquery.yml
```diff
@@ -0,0 +1,95 @@
...
+      - ".github/workflows/dest_bigquery.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "plugins/destination/bigquery/**"
+      - ".github/workflows/dest_bigquery.yml"
+
+jobs:
+  plugins-destination-bigquery:
```
erezrokah, line 16: Not a blocker (we have the same problem with Snowflake) but this won't work for forked PRs, as probably the trust relationship is configured only for the `cloudquery/cloudquery` repo (as it should).
Also forks can't access `secrets`

erezrokah (approved):
This looks great from a first pass. Haven't tried to run it locally yet

contributing/adding_a_new_plugin_to_cq_monorepo.md
```diff
@@ -6,6 +6,6 @@ There are number of steps to follow in order to add all the relevant CI and rele
 
 1) Add your plugin to the `plugins/destination` or `plugins/source` folder.
 2) Add an entry to `.github/pr_labeler.yml`
-3) Add an entry to `./github/workflow/dest_your_plugin.yml` for running tests and linting.
+3) Add an entry to `./github/workflows/dest_your_plugin.yml` for running tests and linting.
```
disq, line 9: Nice catch

plugins/destination/bigquery/client/client.go
```diff
@@ -0,0 +1,59 @@
...
+
+func New(_ context.Context, logger zerolog.Logger, destSpec specs.Destination) (plugins.DestinationClient, error) {
+	if destSpec.WriteMode != specs.WriteModeAppend {
+		return nil, fmt.Errorf("bigquery destination only supports append mode")
+	}
+	c := &Client{
+		logger: logger.With().Str("module", "bq-dest").Logger(),
+	}
+	var spec Spec
+	c.spec = destSpec
```
disq, line 32: Move this assignment inside the `&Client{` initialization?

plugins/destination/bigquery/client/deletestale.go
```diff
@@ -0,0 +1,67 @@
...
+// DeleteStale for BigQuery deletes stale records from previous syncs. Because BigQuery uses a streaming buffer that
+// stores records for up to 90 minutes and disallows deletions during this time, it is important that syncs using the delete-stale
+// functionality be spaced at least 90 minutes apart. This shouldn't be a problem for most CloudQuery users, who typically
+// run syncs once a day. If more frequent syncs are needed, consider batch-loading from a GCS bucket.
+// See https://stackoverflow.com/questions/43085896/update-or-delete-tables-with-streaming-buffer-in-bigquery for more
+// details and further reading.
+// Note that although this function works, it is not currently used because the BigQuery plugin only supports "append" mode,
+// and it is not possible to overwrite rows in BigQuery. It might be used by other modes in the future.
+func (c *Client) DeleteStale(ctx context.Context, tables schema.Tables, source string, syncTime time.Time) error {
+	eg := errgroup.Group{}
```
disq, line 25: Why not:
```suggestion
	eg, ctx := errgroup.WithContext(ctx)
```

plugins/destination/bigquery/client/migrate.go
```diff
@@ -0,0 +1,115 @@
...
+
+const concurrentMigrations = 10
+
+// Migrate tables. It is the responsibility of the CLI of the client to lock before running migrations.
+func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
+	eg := errgroup.Group{}
+	eg.SetLimit(concurrentMigrations)
+	for _, table := range tables {
+		table := table
+		eg.Go(func() error {
```
disq, line 22: Again, why not:
```suggestion
	eg, ctx := errgroup.WithContext(ctx)
```

plugins/destination/bigquery/client/read.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+const (
+	readSQL = "SELECT * FROM `%s.%s.%s` WHERE `_cq_source_name` = @cq_source_name"
+)
+
+func (*Client) createResultsArray(table *schema.Table) []bigquery.Value {
+	results := make([]bigquery.Value, 0, len(table.Columns))
+	for _, col := range table.Columns {
+		switch col.Type {
```
disq, line 21: Maybe add a `default` case with a safety `panic()`, to blow up on unhandled/new types? Just like the one in `SchemaTypeToBigQuery`

.github/workflows/dest_bigquery.yml
hermanschaaf: Yeah, I believe this is a limitation right now. We'll need to think of ways to mitigate this at some point, but probably not high urgency.

website/pages/docs/plugins/destinations/bigquery/_meta.json
```diff
@@ -0,0 +1,3 @@
+{
+  "overview": "Overview"
```
disq, line 2: Why not also a configuration section and configuration.md?

plugins/destination/bigquery/client/write.go
```diff
@@ -0,0 +1,103 @@
...
+		workers[t.Name] = &worker{
+			writeChan: writeChan,
+		}
+		eg.Go(func() error {
+			return c.writeResource(ctx, t, writeChan)
+		})
+	}
+
+	for r := range res {
+		workers[r.TableName].writeChan <- r.Data
```
disq, line 96: Interesting. Nice.

website/pages/docs/plugins/destinations/bigquery/_meta.json
hermanschaaf: I think we can do that in a follow-up, but right now none of the destination plugins have a configuration.md, so let's at least keep it consistent.

yevgenypats (changes requested):
Looks great! Few comments and potential small bugs.

.github/workflows/dest_bigquery.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Test
+        run: make test
+        env:
+          BIGQUERY_PROJECT_ID: ${{ secrets.BIGQUERY_TEST_PROJECT_ID }}
+          BIGQUERY_DATASET_ID: ${{ secrets.BIGQUERY_TEST_DATASET_ID }}
+  validate-release:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/cloudquery/golang-cross:v10.0.0
```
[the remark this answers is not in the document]
hermanschaaf: Ah no, I'll remove this, was a copy-paste from snowflake probably.

plugins/destination/bigquery/client/client.go
```diff
@@ -0,0 +1,59 @@
...
+	if err := destSpec.UnmarshalSpec(&spec); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal bigquery spec: %w", err)
+	}
+	spec.SetDefaults()
+	if err := spec.Validate(); err != nil {
+		return nil, err
+	}
+	client, err := bigquery.NewClient(context.Background(), spec.ProjectID)
```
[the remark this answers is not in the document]
hermanschaaf: Using `ctx` doesn't work, as this context is used for all future queries (the `Query` command doesn't take a context). So by the time a query is made, the context has been cancelled. But maybe I should refactor this so that we create the client before every query, rather than at initialization?

plugins/destination/bigquery/client/write.go
```diff
@@ -0,0 +1,103 @@
...
+		c.logger.Debug().Interface("cols", saver.cols).Msg("got resource")
+		batch = append(batch, saver)
+		if len(batch) >= batchSize {
+			c.logger.Debug().Msg("Writing batch")
+			// we use a context with timeout here, because inserter.Put can retry indefinitely
+			// on retryable errors if not given a context timeout
+			timeoutCtx, cancel := context.WithTimeout(ctx, writeTimeout)
+			err := inserter.Put(timeoutCtx, batch)
+			if err != nil {
+				cancel()
```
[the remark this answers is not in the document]
hermanschaaf: The linter forced me to add it. 😅 They say you need to ensure that cancel is called on every possible path, otherwise there could be resource leakage.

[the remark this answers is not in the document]
hermanschaaf: The alternative is to use a `defer cancel()`, but then we should wrap the whole thing in a function so that we're not calling `defer` inside a for loop.

plugins/destination/bigquery/client/migrate.go
```diff
@@ -0,0 +1,115 @@
...
+	"google.golang.org/api/googleapi"
+)
+
+const concurrentMigrations = 10
+
+// Migrate tables. It is the responsibility of the CLI of the client to lock before running migrations.
+func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
+	eg := errgroup.Group{}
+	eg.SetLimit(concurrentMigrations)
+	for _, table := range tables {
```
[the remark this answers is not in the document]
hermanschaaf: Good catch 👍

plugins/destination/bigquery/client/client.go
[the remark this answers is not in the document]
hermanschaaf: I've updated it so that a new client (with attached context) is created on every `Write` / `Migrate` call. I tested this and it seems to work fine.

hermanschaaf:
@yevgenypats @disq Thanks for the reviews! Some great comments. I've updated based on all the things you highlighted now, PTAL 🙇

.github/workflows/dest_bigquery.yml
```diff
@@ -0,0 +1,111 @@
...
+      - ".github/workflows/dest_bigquery.yml"
+
+jobs:
+  resolve-runner:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    outputs:
+      runner: ${{ steps.resolve.outputs.runner }}
+    steps:
+      - name: Check if should use large runner
```
yevgenypats, line 22: let's remove this large runner as well. I don't think we need it works with the default runner fast enough given it's a small plugin otherwise it will block aws using big runner which I think it the only one that really needs that.

plugins/destination/bigquery/.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+        goarch: arm64
+      # linux arm64 has some issues with cross-compile
+      - goos: linux
+        goarch: arm64
+    overrides:
+      - goos: windows
+        goarch: amd64
+        goamd64: v1
+        env:
+          - CGO_ENABLED=1
```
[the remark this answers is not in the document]
yevgenypats: @hermanschaaf I think this is not address yet.

plugins/destination/bigquery/Makefile
```diff
@@ -0,0 +1,23 @@
...
+# we clean the cache to avoid scenarios when we change something in the db and we want to retest without noticing nothing run
+	go clean -testcache
+	go test -timeout 3m ./...
+
+.PHONY: lint
+lint:
+	@golangci-lint run --config ../../.golangci.yml --timeout 10m --verbose
+
+.PHONY: release-dry-run
+release-dry-run:
```
yevgenypats, line 14: THat's a copy/paste from snowflake. There is no need for `release-dry-run` for BigQuery given it just work (no CGO stuff)

yevgenypats (approved):
Looks good. few last nits.

plugins/destination/bigquery/.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+          - GO111MODULE=on
```
yevgenypats, line 45: Commenting again here as I don't think this is addressed

hermanschaaf: Ah thanks, sorry I thought this was the workflow file. Updating now

hermanschaaf: Should be fixed now!

PR #5103, opened 2022-11-25
chore(cloudflare): Better codegen

disq:
Get updates from the Gandi plugin and apply them, plus customizations


plugins/source/cloudflare/codegen/main.go
```diff
@@ -56,27 +75,21 @@ func generateTable(basedir string, r recipes.Resource) {
...
+		if strings.HasPrefix(c.Resolver, "client.") {
 			r.ImportClient = true
 		}
 	}
 
 	mainTemplate := r.Template + ".go.tpl"
 	tpl, err := template.New(mainTemplate).ParseFS(templatesFS, "templates/"+mainTemplate)
 	if err != nil {
-		log.Fatal(fmt.Errorf("failed to parse cf templates: %w", err))
+		log.Fatal(fmt.Errorf("failed to parse gandi templates: %w", err))
```
hermanschaaf, line 90:
```suggestion
		log.Fatal(fmt.Errorf("failed to parse templates: %w", err))
```

hermanschaaf (approved):
LGTM, nice improvement 👍 

If we're going to use this in a lot of plugins, we may want to move the functionality to the SDK so we don't copy some fairly complicated logic around name generation to a lot of different places. It seems like all plugins could benefit from the naming consistency offered by this approach as well.

disq:
> LGTM, nice improvement 👍
> 
> If we're going to use this in a lot of plugins, we may want to move the functionality to the SDK so we don't copy some fairly complicated logic around name generation to a lot of different places. It seems like all plugins could benefit from the naming consistency offered by this approach as well.

Yes, next step is to port it to Okta and circle back to Gandi to fit all changes or at least see differences.

erezrokah (approved):
🚀

PR #5116, opened 2022-11-28
fix(website): Remove GCP Terraform from deployment section

erezrokah:
#### Summary

The GCP repository is archived https://github.com/cloudquery/terraform-gcp-cloudquery


erezrokah:
Broken links should be fixed in https://github.com/cloudquery/cloudquery/pull/5117

yevgenypats (approved):

PR #5118, opened 2022-11-28
feat(okta)!: Upgrade SDK, use codegen, add mock tests

### LINKED ISSUE #3769: okta: users table has extended columns
`okta_users` table has extended columns that go into custom types, which is not consistent with the rest of cq source plugins.

`credentials` and `type` should be two JSON columns, but they expose 10 and 9 separate columns respectively.

disq:
Implements codegen using recent `gandi` and `cloudflare` changes as base.

Also fixes #3769

BEGIN_COMMIT_OVERRIDE
feat(okta):  Don't unwrap `okta_users` table `credentials` column
  BREAKING-CHANGE: Don't unwrap `okta_users` table `credentials` column. `credentials` column is now a JSON type column, instead of unwrapped individuals columns. Please use JSON query syntax to access the data.
  
feat(okta):  Don't unwrap `okta_users` table `type` column
  BREAKING-CHANGE: Don't unwrap `okta_users` table `type` column. `type` column is now a JSON type column, instead of unwrapped individuals columns. Please use JSON query syntax to access the data.
END_COMMIT_OVERRIDE


plugins/source/okta/docs/tables/okta_application_group_assignments.md
```diff
@@ -18,5 +18,4 @@ This table depends on [okta_applications](okta_applications.md).
 |app_id (PK)|String|
 |id (PK)|String|
 |last_updated|Timestamp|
-|priority|Int|
-|profile|JSON|
```
disq, line 22: I'll investigate what's up with this

plugins/source/okta/codegen/recipes/recipes.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if r.Service == "" {
+		log.Fatalf("Service is required for %s", r.SubService)
+	}
+	if r.SubService == "" {
+		r.SubService = csr.ToSnake(pluralizeClient.Singular(ds.Name()))
+	}
+}
+
+func (r *Resource) GenerateNames() {
```
yevgenypats, line 81: Maybe those should be set manually rather then a complex logic that generates them?

plugins/source/okta/codegen/recipes/recipes.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+
+	r.Filename = csr.ToSnake(r.TableName) + ".go"
+	r.TableFuncName = csr.ToPascal(r.TableName)
+	r.ResolverFuncName = "fetch" + r.TableFuncName
+}
+
+// SetParentChildRelationships calculates and sets the parent and children fields on resources.
+func SetParentChildRelationships(resources []*Resource) error {
```
yevgenypats, line 113: Same here?

yevgenypats (changes requested):
Just gave a quick look without trying to run this. I mostly worry where there is complex logic for autogenerating various stuff.

plugins/source/okta/codegen/recipes/recipes.go
disq: As you can see in the next line there's a name override option.

erezrokah (approved):
Looks great 🚀 

I think we can make the change log descriptive by using the following:

## PR title
```
feat(okta): Use codegen, add mock tests
```

## PR Body
```
feat(okta):  Don't unwrap `okta_users` table `credentials` column
  BREAKING-CHANGE: Don't unwrap `okta_users` table `credentials` column. `credentials` column is now a JSON type column, instead of unwrapped individuals columns. Please use JSON query syntax to access the data.
  
feat(okta):  Don't unwrap `okta_users` table `type` column
  BREAKING-CHANGE: Don't unwrap `okta_users` table `type` column. `type` column is now a JSON type column, instead of unwrapped individuals columns. Please use JSON query syntax to access the data.

**Comment by disq** [comment]
@erezrokah thanks for the review, is there a way to add issue-only comments in the PR body that won't go in the changelog? I still need to be able to add comments and references to the fixed issue etc.

**Comment by erezrokah** [comment]
> @erezrokah thanks for the review, is there a way to add issue-only comments in the PR body that won't go in the changelog? I still need to be able to add comments and references to the fixed issue etc.

I think you can do:

## PR Body
```
Put whatever you want here until the override string

BEGIN_COMMIT_OVERRIDE
feat(okta):  Don't unwrap `okta_users` table `credentials` column
  BREAKING-CHANGE: Don't unwrap `okta_users` table `credentials` column. `credentials` column is now a JSON type column, instead of unwrapped individuals columns. Please use JSON query syntax to access the data.
  
feat(okta):  Don't unwrap `okta_users` table `type` column
  BREAKING-CHANGE: Don't unwrap `okta_users` table `type` column. `type` column is now a JSON type column, instead of unwrapped individuals columns. Please use JSON query syntax to access the data.
END_COMMIT_OVERRIDE
```

Our release bot will use whatever is between `BEGIN_COMMIT_OVERRIDE` and `END_COMMIT_OVERRIDE` for the change log.

**Comment by erezrokah** [comment]
Also what matters is the commit body (if you don't use the override), but our auto merge bot uses the PR body, so if you want to customize the commit body, you'll need to merge manually and set it.

File: plugins/source/okta/codegen/recipes/recipes.go (hunk 3 above)
Comment (disq) [reply plugins/source/okta/codegen/recipes/recipes.go:106]: With [`d0dc60d` (#5118)](https://github.com/cloudquery/cloudquery/pull/5118/commits/d0dc60d64919d3719be14bbfa672b1af5896e712) it's a bit more simplified and overridable.

File: plugins/source/okta/codegen/recipes/applications.go [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+					Name:     "profile",
+					Type:     schema.TypeJSON,
+					Resolver: `schema.PathResolver("Profile")`,
+				},
+			},
+			Relations: []string{"ApplicationUsers()", "ApplicationGroupAssignments()"},
+		},
+		{
+			DataStruct: &okta.AppUser{},
+			PKColumns:  []string{"app_id", "id"},
```
candiduslynx, line 27: instead of adding `app_id` here you could add the PK option right in the extra col def

candiduslynx: better yet, utilize the following func
```go
func parentID(pfx string) codegen.ColumnDefinition {
	const parentIdResolver = `schema.ParentColumnResolver("id")`
	return codegen.ColumnDefinition{
		Name:     pfx + "_id",
		Type:     schema.TypeString,
		Resolver: parentIdResolver,
		Options:  schema.ColumnCreationOptions{PrimaryKey: true},
	}
}
```
and use it

plugins/source/okta/codegen/recipes/applications.go
```diff
@@ -0,0 +1,64 @@
+package recipes
+
+import (
+	"github.com/cloudquery/plugin-sdk/codegen"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/okta/okta-sdk-golang/v2/okta"
```
candiduslynx, line 6: https://pkg.go.dev/github.com/okta/okta-sdk-golang/v3/okta

plugins/source/okta/codegen/recipes/applications.go
```diff
@@ -0,0 +1,64 @@
...
+					Type:     schema.TypeJSON,
+					Resolver: `schema.PathResolver("Profile")`,
+				},
+			},
+			Relations: []string{"ApplicationUsers()", "ApplicationGroupAssignments()"},
+		},
+		{
+			DataStruct: &okta.AppUser{},
+			PKColumns:  []string{"app_id", "id"},
+			SkipFields: []string{"Profile"},
```
candiduslynx, line 28: Is it required? I thought by default `interface{} was transformed to a JSON field and the MarshalJSON would do the magic?

candiduslynx (changes requested):

plugins/source/okta/codegen/recipes/applications.go
disq: I find it more self contained this way, all PKs are in a single line.

plugins/source/okta/codegen/recipes/applications.go
disq: Profile was a useless field but I don't remember why.

plugins/source/okta/codegen/recipes/applications.go
disq: It doesn't look like v3 is released yet according to https://github.com/okta/okta-sdk-golang

plugins/source/okta/codegen/recipes/applications.go
candiduslynx: Yes, but the order will not be preserved, so you still need to be careful.
it might be a good thing to think about pk columns enforcing the pk order, though
@yevgenypats

plugins/source/okta/codegen/recipes/applications.go
candiduslynx: https://github.com/okta/okta-sdk-golang/releases/tag/v3.0.1

plugins/source/okta/codegen/recipes/applications.go
candiduslynx: You added it throught extra columns, so, not as useless?

candiduslynx (dismissed):

plugins/source/okta/codegen/recipes/applications.go
disq: tested: Removing `Profile` from here removes it from the generated code. So this looks to be necessary to override it to be JSON.

plugins/source/okta/codegen/recipes/applications.go
disq: They didn't update the readme 🤷

plugins/source/okta/codegen/recipes/applications.go
yevgenypats: I don't think it's important atm and don't have time to review this throughly for a small plugin with little usage so whatever @disq is going for.

plugins/source/okta/codegen/recipes/applications.go
disq: Fixed in [`91c66d2` (#5118)](https://github.com/cloudquery/cloudquery/pull/5118/commits/91c66d25ce8b8cfc281981ac1d79fe203c90ec33)

plugins/source/okta/client/client.go
```diff
@@ -8,16 +8,18 @@ import (
...
 )
 
 type Client struct {
 	// This is a client that you need to create and initialize in Configure
 	// It will be passed for each resource fetcher.
 	logger zerolog.Logger
-	Okta   *okta.Client
 	spec   specs.Source
+
+	Okta *okta.APIClient
```
candiduslynx, line 22: Curious: can it be embedded?

candiduslynx (approved):
1 nit: embed okta client into plugin client.

plugins/source/okta/client/client.go
disq: Done

yevgenypats (approved):

PR #5122, opened 2022-11-28
feat(crowdstrike): CrowdStrike source plugin

cqgaurav:
#### Summary

This PR adds a few services for a crowdstrike cloudquery plugin. It is intended to serve as a smaller example for adding a realisting CQ plugin.

Tested using the following source and destination configs

`cloudquery sync .
`
// source.yml
```
kind: source
spec:
  name: crowdstrike-0
  # registry: github
  # path: cloudquery/aws
  registry: grpc
  path: localhost:7777
  # version: v7.0.1
  tables: ['*']
  skip_tables: ['crowdstrike_incidents_crowdscore'] # skipped since this is pending on crowdstrike support.
  destinations: ["postgresql"]
  spec:
    client_id: "YOUR_CLIENT_ID"
    client_secret: "YOUR_CLIENT_SECRET"

```

```
kind: destination
spec:
  name: postgresql
  path: cloudquery/postgresql
  version: v1.7.9
  write_mode: append
  spec:
```

Then start a local postgre in docker using 
`docker exec -it cloudquery_postgres psql -U postgres
`
And test that the table was created.

```
postgres=# select * from crowdstrike_alerts_query;
 _cq_source_name |       _cq_sync_time        |                _cq_id                | _cq_parent
_id | errors |                                                                                   
               meta                                                                              
                    | resources 
-----------------+----------------------------+--------------------------------------+-----------
----+--------+-----------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------------
--------------------+-----------
 crowdstrike-0   | 2022-11-28 17:10:50.680223 | b0635155-2c56-41ee-8afb-59d7a0402ab2 |           
    | null   | {"writes": {"resources_affected": 0}, "trace_id": "173893ad-6264-4149-8a17-9c1e52a
8b5dc", "pagination": {"limit": 100, "total": 0, "offset": 0}, "powered_by": "detectsapi", "query
_time": 0.01502205} | {}
(1 row)
```

Note: I haven't been able to test this end to end because of the complexity of a test CrowdStrike setup combined with a non existent crowdstrike support, but hopefully it should work as an example.


plugins/source/crowdstrike/README.md
```diff
@@ -0,0 +1,37 @@
+# Crowdstrike Plugin
```
bbernays, line 1: I think it is actually `CrowdStrike`

plugins/source/crowdstrike/resources/services/incidents/crowdscore_fetch.go
```diff
@@ -0,0 +1,28 @@
...
+	retval, err := c.Services.Incidents.CrowdScore(&incidents.CrowdScoreParams{
+		Context: ctx,
+		Sort:    &desc,
+	})
+	if err != nil {
+		panic(err)
+	}
+	payload := retval.GetPayload()
+	fmt.Printf("As of %s your CrowdScore is %d.\n",
+		payload.Resources[0].Timestamp.String(), *payload.Resources[0].Score)
```
bbernays, line 25: I think this should be using the logger that is in the client

plugins/source/crowdstrike/codegen/recipes/alerts.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/crowdstrike/gofalcon/falcon/models"
+)
+
+func Alerts() []*Resource {
+	resources := []*Resource{
+		{
+			Service:    "alerts",
+			SubService: "query",
+			Struct:     &models.MsaQueryResponse{},
```
bbernays, line 12: It needs a PK field... Otherwise there is no way to do an update/upsert

cqgaurav: the api itself doesnt' ahve a PK field. Not sure what is the recommendation here.

plugins/source/crowdstrike/resources/services/incidents/crowdscore_fetch.go
cqgaurav: Removed

plugins/source/crowdstrike/README.md
cqgaurav: Done

plugins/source/crowdstrike/README.md
```diff
@@ -0,0 +1,38 @@
+# CrowdStrike Plugin
+
+This plugin pulls information from Crowdstrike and loads it into any supported CloudQuery destination (e.g. PostgreSQL).
```
disq, line 3: `CrowdStrike` :)

plugins/source/crowdstrike/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Links
+
+- [Tables](./docs/tables/README.md)
+
+## Authentication
+
+In order to fetch information from CrowdStrike, `cloudquery` needs to be authenticated. A ClientId and secret is required for authentication. Follow [these steps](https://www.crowdstrike.com/blog/tech-center/get-access-falcon-apis/) to set these up. Note that you will also need to enlist the client to have the appropriate scope for what you want to query.
```
disq, line 11: Maybe don't "leak" the "ClientId":
```suggestion
In order to fetch information from CrowdStrike, `cloudquery` needs to be authenticated. A client id and secret is required for authentication. Follow [these steps](https://www.crowdstrike.com/blog/tech-center/get-access-falcon-apis/) to set these up. Note that you will also need to enlist the client to have the appropriate scope for what you want to query.
```

plugins/source/crowdstrike/README.md
```diff
@@ -0,0 +1,38 @@
...
+  tables: ["*"]
+  destinations: ["postgresql"]
+  spec:
+    client_id: <CLIENT_ID>
+    client_secret: <CLIENT_SECRET>
+```
+
+## Example
+
+You can reduce alert fatigue by narrowing alerts from crowdstrike using fuzzy matching.
```
disq, line 34:
```suggestion
You can reduce alert fatigue by narrowing alerts from CrowdStrike using fuzzy matching.
```

Also, narrow _down_ or just narrow?

plugins/source/crowdstrike/client/client.go
```diff
@@ -0,0 +1,72 @@
...
+	Incidents Incidents
+	Alerts    Alerts
+}
+
+func (*Client) Logger() *zerolog.Logger {
+	return &log.Logger
+}
+
+func (*Client) ID() string {
+	return "CrowdStrikeClient"
```
disq, line 32:
```suggestion
	return c.spec.Name
```
(`func (c *Client)` obvs...)

plugins/source/crowdstrike/client/client.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func New(ctx context.Context, logger zerolog.Logger, s specs.Source) (schema.ClientMeta, error) {
+	crowdStrikeSpec := &Spec{}
+	if err := s.UnmarshalSpec(&crowdStrikeSpec); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal CrowdStrike spec: %w", err)
+	}
+	clientId, ok := os.LookupEnv("FALCON_CLIENT_ID")
```
disq, line 40: Are these env vars already used in existing CS tools? If not, remove them or make them clearer in naming (`CROWDSTRIKE` instead of `FALCON` maybe)

plugins/source/crowdstrike/resources/services/incidents/crowdscore_fetch.go
```diff
@@ -0,0 +1,25 @@
...
+
+func fetchCrowdscore(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	desc := "timestamp.desc"
+	retval, err := c.Services.Incidents.CrowdScore(&incidents.CrowdScoreParams{
+		Context: ctx,
+		Sort:    &desc,
+	})
+	if err != nil {
+		panic(err)
```
disq, line 20:
```suggestion
		return err
```

plugins/source/crowdstrike/resources/services/alerts/query_fetch.go
```diff
@@ -0,0 +1,25 @@
...
+
+func fetchQuery(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	desc := "timestamp.desc"
+	retval, err := c.Services.Alerts.GetQueriesAlertsV1(&alerts.GetQueriesAlertsV1Params{
+		Context: ctx,
+		Sort:    &desc,
+	})
+	if err != nil {
+		panic(err)
```
disq, line 20:
```suggestion
		return err
```

disq (changes requested):
Small things, also remove the `codegen` binary from `codegen/` :)

plugins/source/crowdstrike/codegen/recipes/alerts.go
bbernays: instead of using the `MsaQueryResponse` struct I think we should actually be storing the contents of 
`models.MsaQueryResponse.resources` which is of type `DomainEnvironmentScore` ...
We don't need the `errors` or `meta` from the request stored in the destination... This way we can also use the `id` column as a PK

cqgaurav: I am not sure if i am looking at the right place, but I don't see DomainEnvironmentScore as a type here https://github.com/CrowdStrike/gofalcon/blob/main/falcon/models/msa_query_response.go#L33

plugins/source/crowdstrike/README.md
cqgaurav: Done, i think.

plugins/source/crowdstrike/codegen/recipes/alerts.go
bbernays: I mistakenly put this comment in the wrong place it should be in `plugins/source/crowdstrike/codegen/recipes/incidents.go`

plugins/source/crowdstrike/codegen/recipes/incidents.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/crowdstrike/gofalcon/falcon/models"
+)
+
+func CrowdScore() []*Resource {
+	resources := []*Resource{
+		{
+			Service:    "incidents",
+			SubService: "crowdscore",
+			Struct:     &models.APIMsaEnvironmentScoreResponse{},
```
bbernays, line 12: instead of using the `APIMsaEnvironmentScoreResponse` struct I think we should actually be storing the contents of
`models.APIMsaEnvironmentScoreResponse.resources` which is of type DomainEnvironmentScore ...
We don't need the errors or meta from the request stored in the destination... This way we can also use the id column as a PK

plugins/source/crowdstrike/codegen/recipes/alerts.go
bbernays: For this resource do we need to be storing the `errors` and `meta` or could we have a single `text` column and put a PK on it?

erezrokah:
> @erezrokah Thanks for that link. Is it okay to break that into a separate pr to keep the example PR somewhat shorter ?
Maybe I should link/stack that from this one.

@cqgaurav I'm not sure. Is this PR here to serve as an example on how to add a community plugin or a plugin to the monorepo? If for a community plugin I think we should discuss this a bit more, as we've had a [template](https://github.com/cloudquery/cq-provider-template) and also considered a [scaffold command](https://github.com/cloudquery/cloudquery/pull/2055). If for the monorepo, I think it should include the parts that also release the plugin

cqgaurav:
@erezrokah Yes, I don't have a strong opinion, and its possible that the users don't necessarily care about differentiating between plugin and releasing it separately. I added the requested changes

plugins/source/crowdstrike/codegen/recipes/incidents.go
cqgaurav: Done

plugins/source/crowdstrike/README.md
cqgaurav: Done, thanks

plugins/source/crowdstrike/codegen/recipes/alerts.go
cqgaurav: For the alert fatigue usecase, we don't need errors and meta. Lemme know what you think, but i think putting a PK on the text field doesn't feel intuitive, it might be same value as cq_id.

.github/styles/Vocab/Base/accept.txt
```diff
@@ -135,4 +135,13 @@ bugfix
...
 npm
 Gandi
-CrowdStrike
\ No newline at end of file
+CrowdStrike
+parallelization
+hyperscale
+goroutines
+arn
+arns
```
disq, line 143: `arns`? why? Shouldn't it always be `ARNs`?

disq (approved):
👍 with a nit

.github/styles/Vocab/Base/accept.txt
cqgaurav: I think this was a merge artifact. I have removed it, lets see what breaks

erezrokah (approved):

erezrokah:
Do we want to add docs for this plugin? See example in https://github.com/cloudquery/cloudquery/tree/1c9755d87f3037bbb5eef1ab0de66b06da601ea8/website/pages/docs/plugins/sources/aws

PR #5138, opened 2022-11-29
feat: Azure migration diff

yevgenypats:
this is produced for some reason by running `make gen-code`


yevgenypats:
Ok seems because we do `os.Exec` `goimport` and in case `goimport` doesn't exist it will fail silently. Closing.

PR #5139, opened 2022-11-29
feat: Azure more differences

yevgenypats:
Seems having some more issues when running `make gen-code`

### DIFF
--- a/plugins/source/azure/resources/services/container/replications.go
+++ b/plugins/source/azure/resources/services/container/replications.go
@@ -5,9 +5,9 @@ package container
 import (
 	"context"
 
-	"github.com/Azure/azure-sdk-for-go/profiles/latest/containerregistry/mgmt/containerregistry"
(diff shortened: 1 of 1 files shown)


plugins/source/azure/resources/services/container/replications.go
```diff
@@ -5,9 +5,9 @@ package container
 import (
 	"context"
 
-	"github.com/Azure/azure-sdk-for-go/profiles/latest/containerregistry/mgmt/containerregistry"
 	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
 	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/yandex-cloud/go-genproto/yandex/cloud/containerregistry/v1"
```
erezrokah, line 10: I think this is wrong

erezrokah:
I think we can close this

PR #5143, opened 2022-11-29
fix: DeleteStale from relational tables as well

shimonp21:
#### Summary
Won't compile until https://github.com/cloudquery/plugin-sdk/pull/443 is merged.


yevgenypats (approved):

yevgenypats (approved):
This PR seems fine but also a test will be nice (in the SDK) so our destination testing framework will catch this

PR #5144, opened 2022-11-29
fix(gcp): ResourceManager folders listing (add OrgMultiplexing)

### LINKED ISSUE #2975: gcp_resourcemanager_folders table is always empty?
### Describe the Bug
Seems like we don't supply the required `parent` parameter to actually fetch folders. 
https://github.com/cloudquery/cloudquery/blob/352fb3fedbf69b715aa16c39b8545cc55f4f9d5c/plugins/source/gcp/resources/services/resourcemanager/folders.go#L73
The table is empty even when user has permissions for folders.

If this is an actual problem, the rest of `resourcemanager` resources also need to be inspected

disq:
Get the "parent" of the project (is either a `folder/1234` or an `organization/1234`) before querying the API.

Had to switch folder fetching and testing to manual resources to be able to add the extra API call. Test code uses the projects mock test constructor to get a `ResourcemanagerProjectsClient`.

I wasn't able to successfully test this due to `IAM_PERMISSION_DENIED`:
```
 Permission 'resourcemanager.folders.list' denied on resource '//cloudresourcemanager.googleapis.com/organizations/xxxx 4' (or it may not exist).\nerror details: name = ErrorInfo reason = IAM_PERMISSION_DENIED domain = cloudresourcemanager.googleapis.com metadata = map[permission:resourcemanager.folders.list resource:organizations/xxxx]
 PermissionDenied desc = Permission 'resourcemanager.folders.list' denied on resource '//cloudresourcemanager.googleapis.com/folders/xxxx' (or it may not exist).\nerror details: name = ErrorInfo reason = IAM_PERMISSION_DENIED domain = cloudresourcemanager.googleapis.com metadata = map[permission:resourcemanager.folders.list resource:folders/xxxx]
```

and such.

Fixes #2975 (hopefully)


plugins/source/gcp/codegen/recipes/resourcemanager.go
```diff
@@ -1,21 +1,22 @@
...
-		ListFunction:        (&pb.UnimplementedFoldersServer{}).ListFolders,
-		UnimplementedServer: &pb.UnimplementedFoldersServer{},
+		SubService: "folders",
+		Struct:     &pb.Folder{},
+		//NewFunction:         resourcemanager.NewFoldersClient,
+		//RequestStruct:       &pb.ListFoldersRequest{},
+		//ResponseStruct:      &pb.ListFoldersResponse{},
+		//RegisterServer:      pb.RegisterFoldersServer,
+		//ListFunction:        (&pb.UnimplementedFoldersServer{}).ListFolders,
+		//UnimplementedServer: &pb.UnimplementedFoldersServer{},
```
erezrokah, line 17:
```suggestion
```

Can we remove this commented out part?

erezrokah (approved):
Added one non blocking comment, but better if someone with more GCP plugin context review too

yevgenypats:
I think before merging we need to manually test it. I'll give you the right permissions.

disq:
Test results: it gives me the same folders for multiple projects, so we have dupes… since the folders are on an org level and not on a project level

```sql
select name, display_name, array_agg(project_id) from gcp_resourcemanager_folders  group by 1, 2;
```

```
-[ RECORD 1 ]+-------------------------------------------------------------------------------------------------------------------------------------
name         | folders/123
display_name | the-test-folder
array_agg    | {project1,project2,project3}
-[ RECORD 2 ]+-------------------------------------------------------------------------------------------------------------------------------------
name         | folders/456
display_name | system-gsuite
array_agg    | {project1,project2,project3}
```

disq:
New results after [`1846a61` (#5144)](https://github.com/cloudquery/cloudquery/pull/5144/commits/1846a61ca5a4fd6857a734d2e37806d2558b24d0)

```sql
select organization_id, name, parent, display_name, state from gcp_resourcemanager_folders;
```

```
-[ RECORD 1 ]---+---------------------------
organization_id | 12345
name            | folders/112233
parent          | organizations/12345
display_name    | system-gsuite
state           | ACTIVE
-[ RECORD 2 ]---+---------------------------
organization_id | 12345
name            | folders/445566
parent          | organizations/12345
display_name    | the-test-folder
state           | ACTIVE
```

erezrokah (approved):
Happy to approve again as this seems to be working.
Also this is a breaking change due to the column rename

disq:
This will get refactored a little bit after #5512 is merged

plugins/source/gcp/client/client.go
```diff
@@ -24,6 +24,9 @@ type Client struct {
 	ProjectId string
 	// Logger
 	logger zerolog.Logger
+
+	// This is used if we don't have a context in the call
+	ctx context.Context
```
disq, line 29: This is an unfortunate addition but we need some kind of context for the OrgMultiplexer, don't want to use context.BG

yevgenypats: Thx for pointing out. I'll take a look if there is a way to avoid it.

yevgenypats: I think in that case it should be called on `New` and then cached on the client level and the table should just get the cached data instead of reusing unrelated `ctx`

yevgenypats (changes requested):

plugins/source/gcp/client/client.go
disq: @yevgenypats what do you mean? ctx is already set on `New` and it's copied from the source client to the multiplexed clients if the OrgMultiplexer is used.

plugins/source/gcp/client/multiplex.go
```diff
@@ -13,3 +15,38 @@ func ProjectMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
...
+		return []schema.ClientMeta{client}
+	}
+
+	dupes := map[string]struct{}{}
+
+	l := make([]schema.ClientMeta, 0, len(client.projects))
+	for _, projectId := range client.projects {
+		resp, err := projectsClient.GetProject(client.ctx, &resourcemanagerpb.GetProjectRequest{
+			Name: "projects/" + projectId,
+		})
```
bbernays, line 34: What about retry logic? The default rate limit for getting projects is 600 req/minute... Not sure how fast this API is though. We might also want to document which quota users can raise if they want to increase performance of this section

plugins/source/gcp/client/multiplex.go
```diff
@@ -13,3 +15,38 @@ func ProjectMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
...
+
+	projectsClient, err := resourcemanager.NewProjectsClient(client.ctx, client.ClientOptions...)
+	if err != nil {
+		client.Logger().Error().Err(err).Msg("OrgMultiplex: Failed to create ProjectsClient")
+		return []schema.ClientMeta{client}
+	}
+
+	dupes := map[string]struct{}{}
+
+	l := make([]schema.ClientMeta, 0, len(client.projects))
```
bbernays, line 30: Do we want to provide some type of logging (even if it is just `debug`) so that users can be aware that CQ is not hanging?

disq: @bbernays erm, we have "skipping project" and "found project" messages on lines 43 and 47 already?

plugins/source/gcp/client/multiplex.go
disq: `NewProjectsClient` is already initialized with `client.ClientOptions...`, do we also need to put it in the `GetProject` call? Not sure

plugins/source/gcp/client/client.go
```diff
@@ -164,6 +179,15 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source) (schema.Cli
 	}
 
 	c.projects = projects
+
+	if len(c.orgs) == 0 {
```
yevgenypats, line 183: Can it be anything other then `0` I don't see it set before hand ?

yevgenypats (approved):
Nice! Had one nit

yevgenypats:
Seems something is failing now, not sure why?

PR #5145, opened 2022-11-29
refactor(website): Simplify Nextra versions replace code

erezrokah:
#### Summary

This PR refactors the code that does the versions replacement in our docs.
Main changes:
1. Use `fs.readdirSync` to read all files in one call instead of a while loop
2. Use the same object structure for the `cli` instead of a string, so we can make the replace iterative


website/next.config.mjs
```diff
@@ -1,68 +1,75 @@
...
-      if (version === undefined) {
-        throw new Error(`Could not find version for cli ${match}`)
-      }
-      node.value = node.value.replace(reCLI, version)
-    }
+    });
   }
   if (node.children !== undefined) {
-    node.children.map(replaceMdxCodeVersions)
+    node.children.forEach(replaceMdxCodeVersions);
```
erezrokah, line 69: `map` is indented to map the array into another array. I think `forEach` is what we want here

hermanschaaf (approved):

PR #5096, opened 2022-11-25
feat: Datadog plugin

amanenk:
#### Summary

closes #4794

PR #5170, opened 2022-11-30
fix(website): Don't error on missing plugin version for new plugins

erezrokah:
#### Summary

When adding a new plugin the version file doesn't exist yet, so the build fails.
This defaults to `v1.0.0` instead of failing the build, **only for new plugins**
See failure:
https://vercel.com/cloudquery/cloudquery-web/52G1znzrGcKCidgcvhCfKk5WZAhD
PR https://github.com/cloudquery/cloudquery/pull/5096


hermanschaaf:
I may not have the full picture, but erroring on missing version seems like a pretty good safeguard to have in place, so we don't run the risk of injecting incorrect versions if there is something wrong with the version file.

Could we maybe manually add the new plugin's version the first time instead?

erezrokah:
> I may not have the full picture, but erroring on missing version seems like a pretty good safeguard to have in place, so we don't run the risk of injecting incorrect versions if there is something wrong with the version file.
> 
> Could we maybe manually add the new plugin's version the first time instead?

Good point, however I think it will be quite easy to miss adding the version file and understanding where to add it for someone who's not familiar with the Website code.
I'm assuming we'll always have a `v1.0.0` of a plugin so even if something is wrong with the version files it will still point to an existing plugin.

erezrokah:
Maybe we can detect it's a new plugin somehow

website/next.config.mjs
```diff
@@ -24,6 +24,29 @@ function parseName(name) {
   return parts[0];
 }
 
+function isUnreleasedPlugin(key, name) {
+  if (key === "cli") {
+    return false;
+  }
+  const pluginDir = path.resolve("..", "plugins", key.slice(0, -1), name);
+  try {
+    // check that is this a plugin
```
hermanschaaf, line 33:
```suggestion
    // check that this is a plugin
```

(nit)

hermanschaaf (approved):
Nice, looks good

website/next.config.mjs
```diff
@@ -24,6 +24,29 @@ function parseName(name) {
...
+    return false;
+  }
+
+  const changelogPath = path.resolve(pluginDir, "CHANGELOG.md");
+  try {
+    const changelogContent = fs.readFileSync(changelogPath, "utf8");
+    const nonEmptyChangelog = changelogContent.includes("1.0.0");
+    return !nonEmptyChangelog;
+  } catch (err) {
+    // no ChangeLog, this is new plugin
```
erezrokah, line 45:
```suggestion
    // no Changelog, this is new plugin
```

PR #5194, opened 2022-11-30
chore: Add comment to section about closing destination clients

hermanschaaf:
~I believe for correctness we should be using the context from the `errgroup` here, but I'm not sure if it's caused any real-world problems.~ Just adding a comment to explain why `ctx` makes sense instead of `gctx`.


erezrokah (approved):

cli/cmd/sync2.go
```diff
@@ -117,7 +117,7 @@ func syncConnectionV2(ctx context.Context, cqDir string, sourceClient *clients.S
 			if err = destClients[i].Write2(gctx, selectedTables, sourceSpec.Name, syncTime, destSubscriptions[i]); err != nil {
 				return fmt.Errorf("failed to write for %s->%s: %w", sourceSpec.Name, destination, err)
 			}
-			if err := destClients[i].Close(ctx); err != nil {
+			if err := destClients[i].Close(gctx); err != nil {
```
hermanschaaf, line 120:
```suggestion
			// call Close on destination client using the outer context, so that it happens even if writes get cancelled 
			if err := destClients[i].Close(cctx); err != nil {
```

hermanschaaf:
After thinking about it some more, I think it was better before. I'm undoing the change but adding a comment about why it makes sense to use `ctx`, so others don't wonder about this in the future too.

erezrokah (approved):

PR #5249, opened 2022-12-01
fix: Handling of NULL bytes in Postgresql Text fields

shimonp21:
Required to fix batch-write-failures due to postgresql TEXT fields containing NULL bytes.

```
2022-11-30T12:00:11Z ERR failed to execute batch with pgerror error="ERROR: invalid byte sequence for encoding \"UTF8\": 0x00 (SQLSTATE 22021)" module=pg-dest
```


plugins/destination/postgresql/client/transformer.go
```diff
@@ -42,24 +45,28 @@ func (*Client) TransformInt8Array(v *schema.Int8Array) interface{} {
...
 		Bytes:  v.Bytes,
 		Status: cqStatusToPgStatus[v.Status],
-	}
+	}, nil
 }
 
-func (*Client) TransformText(v *schema.Text) interface{} {
+func (*Client) TransformText(v *schema.Text) (interface{}, error) {
+	if strings.Contains(v.Str, "\x00") {
+		return nil, fmt.Errorf("postgres TEXT field cannot contain null bytes")
```
yevgenypats, line 60: I don't think `TransformText` should return errors. just replace it with space or remove. That's the point of transformation. Lets' say we have a database were it doesn't support arrays inside jsons or whatever else then it should just transform and/or skip it to the best of it's ability. potentially you can log this as well via `c.log` but def no error

yevgenypats (changes requested):

plugins/destination/sqlite/client/transformer.go
```diff
@@ -27,11 +32,17 @@ func (*Client) TransformJSON(v *schema.JSON) interface{} {
 }
 
 func (*Client) TransformText(v *schema.Text) interface{} {
-	return v.String()
+	return stripNulls(v.String())
 }
 
 func (*Client) TransformTextArray(v *schema.TextArray) interface{} {
-	return v.String()
+	strippedStrings := make([]string, len(v.Elements))
```
yevgenypats, line 39: Nit: Given this function is heavily called, use string builder here instead of making two allocations. https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go and then calling `fmt.Sprintf`

yevgenypats (changes requested):
Some performance nit

PR #5252, opened 2022-12-01
chore(k8s): Fix k8s mocks targets

candiduslynx:


erezrokah:
@candiduslynx I think this is good, but I removed `automerge` so you can see the discussion in https://github.com/cloudquery/cloudquery/pull/4629.

We should also update the CI workflow to catch such issues in the future. See https://github.com/cloudquery/cloudquery/pull/4629/files#diff-e6ce7bc316a2e750a956a86aa73a9d0e59192339395eb4c0c1a70d5ff19a6df1

candiduslynx:
@erezrokah Thanks for the notes. I think I'll leave this PR scope to only manage mocks properly and all other automation/CI can be done in a different PR. WDYT?

erezrokah:
> @erezrokah Thanks for the notes. I think I'll leave this PR scope to only manage mocks properly and all other automation/CI can be done in a different PR. WDYT?

Yep this works

PR #5253, opened 2022-12-01
fix(website): Use `>=` instead of `>` for database versions

erezrokah:
#### Summary

We support versions v10 and v20.2 (inclusive)


disq (approved):
Yes please

PR #5259, opened 2022-12-01
feat(website): Add plugins tables

erezrokah:
#### Summary

A re-do of https://github.com/cloudquery/cloudquery/pull/4728.
I initially started with using the JSON output, but realized this is simpler as we just want to remove the links to individual tables from the readme.

**Other solutions**
 - Add a flag to the `doc` command to not output the links in the README markdown
 - Use `doc --format json` read, the JSON, create a markdown. Seems too much at the moment

~~Datadog CI will pass after https://github.com/cloudquery/cloudquery/pull/5258 is merged~~

Deploy preview in https://cloudquery-k4mb1p3j3-cloudquery.vercel.app/


hermanschaaf:
Couple of thoughts:
 - I think it would be useful if the tables still linked to their detail pages on Github, otherwise we might create the impression that there is no further information. I know this isn't necessarily easy to cleanly do with `sed`, which is why I abandoned the idea while back and built the json output first 😄
 - Should we update the "Tables" links on the overview page to link to these new pages?

erezrokah:
> * Should we update the "Tables" links on the overview page to link to these new pages?

Good point I'll fix that.

> * I think it would be useful if the tables still linked to their detail pages on Github, otherwise we might create the impression that there is no further information. I know this isn't necessarily easy to do cleanly with `sed`, which is why I abandoned the idea while back and built the json output first

I'm wondering if we should update the `doc` command to just output what we need, meaning link each table directly to GitHub using the full path instead of a relative link to the directory.

hermanschaaf:
> I'm wondering if we should update the doc command to just output what we need, meaning link each table directly to GitHub using the full path instead of a relative link to the directory.

@erezrokah Maybe, but then the doc command isn't correct (strictly speaking) when generating the docs locally, which I think is at least part of its appeal. I for one sometimes browse the docs locally from my IDE, and then the relative links are quite useful.

erezrokah:
> @erezrokah Maybe, but then the doc command isn't correct (strictly speaking) when generating the docs locally, which I think is at least part of its appeal. I for one sometimes browse the docs locally from my IDE, and then the relative links are quite useful.

Good point, I've change the `sed` so it embeds the links to the tables and updated the overview.
I think this should be a good start, and if we need more transformations we can go down the JSON path, WDYT?

hermanschaaf (approved):
LGTM 👍 

Nit: we may need to tweak the CSS styles a bit so that the child tables don't have an extra newline above them in the rendered result
Screenshot 2022-12-01 at 13 17 08

erezrokah:
> Nit: we may need to tweak the CSS styles a bit so that the child tables don't have an extra newline above them in the rendered result

Yeah, I saw that. Might needs to dive into how Nextra renders links in lists 😕 I'll do that in a follow up PR

Issue #5250, opened 2022-12-01
Bug(ci): The AWS endpoints sync workflow is faling


erezrokah:
See https://github.com/cloudquery/cloudquery/actions/runs/3586489256/jobs/6035740772

We should fix it

PR #5268, opened 2022-12-01
fix: AWS Endpoint Generation

bbernays:
#### Summary

ECR public API has been added to the endpoint list... we no longer need to hard code it.

Fixes https://github.com/cloudquery/cloudquery/issues/5250


hermanschaaf (approved):
Nice 💯

erezrokah (approved):

Issue #5197, opened 2022-11-30
feat(website): Better Website search


erezrokah:
Not the same as https://github.com/cloudquery/cloudquery/issues/4632 and only focuses on searching the docs.
The built-in Nextra search is quite basic and doesn't seem to have proper ranking.
For example, I'd expect that searching `AWS` will show the AWS source plugin overview first, instead it shows a link to a blog we published a few months ago:
![image](https://user-images.githubusercontent.com/26760571/204859744-61101daf-939b-4b02-9085-eee15f3b9732.png)

Also searching of `Overview` results in a confusing list as Nextra doesn't show the hierarchy:
![image](https://user-images.githubusercontent.com/26760571/204859967-f73327d1-c622-47b7-8925-b71aea9d35c8.png)

erezrokah:
PR to remove the search in https://github.com/cloudquery/cloudquery/pull/5270 until we make it usable

erezrokah:
The issue here https://github.com/shuding/nextra/issues/107 links to an [example](https://github.com/vercel/swr-site/blob/88d1b8affa7f1c344f3aaba229b51867b2cedc8d/components/docsearch.js) on how to use Algolia with Nextra

erezrokah:
Closing as won't fix for now (until our Website grows and lack of search makes it hard to find content)

PR #5270, opened 2022-12-01
fix(website): Remove search

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/5197 and https://github.com/cloudquery/cloudquery/issues/4632.

Our current search is not very useful, so I'm removing it


hermanschaaf (approved):
Hopefully to return soon in a new and more powerful shape! 💪

PR #5278, opened 2022-12-01
doc(website): Sort plugins

hermanschaaf:
As the list of plugins grows (great!), it's getting harder to find a specific plugin, especially if they are in a (seemingly to the user) random order. Let's sort them lexicographically?


erezrokah:
> As the list of plugins grows (great!), it's getting harder to find a specific plugin, especially if they are in a (seemingly to the user) random order. Let's sort them lexicographically?

I was wondering that too, but thinking at least GA plugins should be first?

hermanschaaf:
@erezrokah Ok, let's do that 👍

yevgenypats (approved):

yevgenypats:
> @erezrokah Ok, let's do that 👍

I think lexicography is better then GA first. otherwise it's not possible to search or understand where to find what

PR #5280, opened 2022-12-01
feat(aws): Add inline_policy field to permission_sets table

jhudson10x:
#### Summary

Add inline_policy to the SSOAdmin permission_sets


jhudson10x:
Whoops - TypeString needs to be TypeJSON.

plugins/source/aws/codegen/recipes/ssoadmin.go
```diff
@@ -19,6 +21,13 @@ func SSOAdminResources() []*Resource {
 			Struct:              &types.PermissionSet{},
 			Description:         "https://docs.aws.amazon.com/singlesignon/latest/APIReference/API_PermissionSet.html",
 			PreResourceResolver: "getSsoadminPermissionSet",
+			ExtraColumns: []codegen.ColumnDefinition{
+				{
+					Name:     "inline_policy",
+					Type:     schema.TypeString,
```
bbernays, line 27: If i recall correctly this polict is actually valid JSON... so I think the type should be `schema.TypeJSON`

plugins/source/aws/resources/services/ssoadmin/instances_mock_test.go
```diff
@@ -16,6 +16,7 @@ func buildInstances(t *testing.T, ctrl *gomock.Controller) client.Services {
 	im := types.InstanceMetadata{}
 	ps := types.PermissionSet{}
 	as := types.AccountAssignment{}
+	ip := ""
```
hermanschaaf, line 19: I think this will need to be valid JSON if the column type is to be JSON, and faking `&ip` lower down with a string object will probably lead to a random string (not valid JSON). 

Any valid JSON should do for the purposes of this test though, so maybe just instantiate this variable as `{"key": "value}` or similar

hermanschaaf (changes requested):
Thanks for adding this change! Looks good if we change the column type to JSON. I added one or two more thoughts as well

hermanschaaf (approved):
Great stuff, thanks for the contribution @jhudson10x! 👍 

Approving, but I'll give @bbernays a chance to take another look as well.

plugins/source/aws/resources/services/ssoadmin/instances_mock_test.go
```diff
@@ -16,6 +16,7 @@ func buildInstances(t *testing.T, ctrl *gomock.Controller) client.Services {
 	im := types.InstanceMetadata{}
 	ps := types.PermissionSet{}
 	as := types.AccountAssignment{}
+	ip := "{\"key\": \"value\"}"
```
hermanschaaf, line 19:
```suggestion
	ip := `{"key": "value"}`
```

Also a nit, but using backticks for the string allows you to do away with the backslashes

plugins/source/aws/resources/services/ssoadmin/permission_sets_fetch.go
```diff
@@ -10,6 +11,29 @@ import (
...
+	response, err := svc.GetInlinePolicyForPermissionSet(ctx, &config)
+	if err != nil {
+		return err
+	}
+
+	var inlinePolicy map[string]interface{}
+	err = json.Unmarshal([]byte(*response.InlinePolicy), &inlinePolicy)
+	if err != nil {
+		return err
+	}
```
bbernays, line 32:
```suggestion
```

You shouldn't need to this `Unmarshal` call... It should be done automatically by the SDK

bbernays (changes requested):
Other than 1 change it looks good!

plugins/source/aws/resources/services/ssoadmin/permission_sets_fetch.go
jhudson10x: This is the thing I am talking about in the discord thread - I don't see the contents of the policy show up in the table unless I do this

jhudson10x: @hermanschaaf @bbernays I tried it again and yes it works on my end too. I will merge the suggestion now thank you both.

plugins/source/aws/resources/services/ssoadmin/permission_sets_fetch.go
```diff
@@ -10,6 +11,24 @@ import (
...
+		PermissionSetArn: permissionSetARN,
+	}
+
+	response, err := svc.GetInlinePolicyForPermissionSet(ctx, &config)
+	if err != nil {
+		return err
+	}
+
+
+	return resource.Set(c.Name, inlinePolicy)
```
bbernays, line 29:
```suggestion
	return resource.Set(c.Name, response.InlinePolicy)
```

bbernays (approved):

PR #5282, opened 2022-12-01
refactor(website): Add PluginsTable component

erezrokah:
#### Summary

This PR adds a `PluginsTable` component to make it easier to add new plugins to the Website.
At first I tried making the table generic using MDX magic, but couldn't get it to render properly so this is the best I managed.

### DIFF
--- a/website/components/PluginsTable.tsx
+++ b/website/components/PluginsTable.tsx
@@ -1,0 +1,77 @@
+import { Table, Tr, Td, Th } from "nextra/components";
+import Link from "next/link";
+import { getLatestVersion } from "../utils/versions";
+
+const TableLink = ({ text, href, target = undefined }) => {
+  return (
+    <Link href={href}>
+      <a
+        className="nx-text-primary-500 nx-underline nx-decoration-from-font [text-underline-position:under]"
+        target={target}
+      >
+        {text}
+      </a>
+    </Link>
+  );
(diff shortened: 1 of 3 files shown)


website/components/PluginsTable.tsx
```diff
@@ -0,0 +1,77 @@
+import { Table, Tr, Td, Th } from "nextra/components";
+import Link from "next/link";
+import { getLatestVersion } from "../utils/versions";
+
+const TableLink = ({ text, href, target = undefined }) => {
+  return (
+    <Link href={href}>
+      <a
+        className="nx-text-primary-500 nx-underline nx-decoration-from-font [text-underline-position:under]"
```
erezrokah, line 9: Once https://github.com/shuding/nextra/issues/332 is fixed we should be able to not hard code the classes

website/components/PluginsTable.tsx
```diff
@@ -0,0 +1,77 @@
...
+          />
+        </Td>
+      )}
+      <Td>{stage}</Td>
+    </Tr>
+  );
+};
+
+export const PluginsTable = ({ plugins, type }) => {
+  const sortedPlugins = plugins.sort((a, b) => a.name.localeCompare(b.name));
```
erezrokah, line 56: Plugins are always sorted here

PR #4768, opened 2022-11-17
feat(azure)!: Move Azure source plugin to SDK v2

candiduslynx:
#### Summary

Upgrade Azure plugin to new SDK (#1358)

PR #5286, opened 2022-12-02
feat: Azure more new new resource manager SDK

yevgenypats:
This is a draft PR that helps me review https://github.com/cloudquery/cloudquery/pull/4768. I'll either apply fixes on top https://github.com/cloudquery/cloudquery/pull/4768 or will use this one directly if it will be much better.


plugins/source/azure/client/services.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return services, err
+	}
+	services.ArmfrontdoorDoors = ArmfrontdoorDoors
+
+	ArmkeyvaultVaults, err := armkeyvault.NewVaultsClient(subscriptionId, azCred, nil)
+	if err != nil {
+		return services, err
+	}
+	services.ArmkeyvaultVaults = ArmkeyvaultVaults
```
candiduslynx, line 127: a good idea to init in-place.
I'd go even further and do it like this:
```go
var err error
var services Services

services.ClientA, err = initClientA(...)
if err != nil {
    return services, err
}
...
```

Maybe I'll add it to my PR codegen as well

plugins/source/azure/resources/services/armcdn/endpoints_mock_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/julienschmidt/httprouter"
+)
+
+func createEndpoints() (*client.Services, error) {
+	var item armcdn.EndpointsClientListByProfileResponse
+	if err := faker.FakeObject(&item); err != nil {
+		return nil, err
+	}
+	emptyStr := ""
+	item.NextLink = &emptyStr
```
candiduslynx, line 27:
```suggestion
	item.NextLink = nil
```
OR
```suggestion
	item.NextLink = to.Ptr("")
```

plugins/source/azure/resources/services/armcdn/endpoints_mock_test.go
```diff
@@ -0,0 +1,56 @@
...
+	mux := httprouter.New()
+	mux.GET("/*filepath", func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
```
candiduslynx, line 29: I personally don't see a lot of difference between the generated interface + mock on it vs transport-based mock tests, but prefer the interface+mocks better.

plugins/source/azure/client/client.go
```diff
@@ -6,79 +6,44 @@ import (
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/subscription/armsubscription"
 	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/specs"
 	"github.com/rs/zerolog"
 )
 
+
+
 type Client struct {
 	subscriptions []string
```
candiduslynx, line 21: this info is already present in `services` as keys

candiduslynx (commented):
Not sure about http mocks at all

PR #5296, opened 2022-12-02
chore: Add policy documentation to website

hermanschaaf:
This adds a script to extract from policies:
 1. the tables used
 2. the query titles
and then write these to a `Policies` page on the website:
Screenshot 2022-12-02 at 17 04 46

I also ended up updating a number of policies. In some cases there were missing titles, and in some cases there were queries that had syntax errors and were not being used by any policy. I also added `as title` to many queries (and I think this aids readability), but ultimately realized there are too many instances of queries without it, so made the script handle cases where `as title` is not specified.


website/pages/docs/plugins/sources/aws/policies.md
```diff
@@ -0,0 +1,493 @@
+# Policies and Compliance Frameworks
```
erezrokah, line 1: We should exclude this file via https://github.com/cloudquery/cloudquery/blob/812241697c644bdb1ae202bbadcb3baae456f788/.github/workflows/lint_markdown.yml#L19

website/pages/docs/plugins/sources/aws/policies.md
```diff
@@ -0,0 +1,493 @@
...
+  - aws_elbv2_load_balancers
+  - aws_rds_instances
+  - aws_redshift_clusters
+```
+
+### Queries
+AWS Publicly Available performs the following checks:
+  - Find all API Gateway instances that are publicly accessible
+  - Find all API Gateway V2 instances (HTTP and Webhook) that are publicly accessible
+  - Find all Cloudfront distributions
```
erezrokah, line 414:
```suggestion
  - Find all CloudFront distributions
```

Not sure where this comes from but it should be CloudFront

website/pages/docs/plugins/sources/aws/policies.md
```diff
@@ -0,0 +1,493 @@
...
+  - aws_sns_topics
+```
+
+### Queries
+AWS Unused Resources performs the following checks:
+  - Unused ACM certificate
+  - Unused API Gateway API key
+  - Vaults with no recovery points
+  - Disabled CloudFront distribution
+  - Disabled Cloudwatch alarm
```
erezrokah, line 470:
```suggestion
  - Disabled CloudWatch alarm
```

website/pages/docs/plugins/sources/aws/policies.md
```diff
@@ -0,0 +1,493 @@
...
+  - Connections to Elasticsearch domains should be encrypted using TLS 1.2
+  - Classic Load Balancers with SSL/HTTPS listeners should use a certificate provided by AWS Certificate Manager
+  - Classic Load Balancer listeners should be configured with HTTPS or TLS termination
+  - Application load balancers should be configured to drop HTTP headers
+  - Application and Classic Load Balancers logging should be enabled
+  - Application Load Balancer deletion protection should be enabled
+  - Classic Load Balancers should have connection draining enabled
+  - Classic Load Balancers with HTTPS/SSL listeners should use a predefined security policy that has strong configuration
+  - Application Load Balancer should be configured to redirect all HTTP requests to HTTPS
+  - EMR clusters should not have public ip addresses
```
erezrokah, line 313:
```suggestion
  - EMR clusters should not have public IP addresses
```

website/pages/docs/plugins/sources/aws/policies.md
```diff
@@ -0,0 +1,493 @@
+# Policies and Compliance Frameworks
+
+CloudQuery SQL Policies for AWS. See the [Readme on Github](https://github.com/cloudquery/cloudquery/tree/main/plugins/source/aws/policies) for installation instructions.
```
erezrokah, line 3:
```suggestion
CloudQuery SQL Policies for AWS. See the [Readme on GitHub](https://github.com/cloudquery/cloudquery/tree/main/plugins/source/aws/policies) for installation instructions.
```

website/pages/docs/plugins/sources/aws/policies.md
```diff
@@ -0,0 +1,493 @@
...
+  - Ensure no root account access key exists (Scored)
+  - Ensure MFA is enabled for the "root" account
+  - Ensure hardware MFA is enabled for the "root" account (Scored)
+  - IAM users should not have IAM policies attached
+  - Ensure CloudTrail is enabled in all regions
+  - Ensure CloudTrail log file validation is enabled
+  - CloudTrail trails should be integrated with CloudWatch Logs
+  - Ensure S3 bucket access logging is enabled on the CloudTrail S3 bucket
+  - CloudTrail should have encryption at rest enabled
+  - Ensure rotation for customer created CMKs is enabled (Scored)
```
erezrokah, line 48:
```suggestion
  - Ensure rotation for customer created custom master keys is enabled (Scored)
```

Maybe it's better to use the full term instead of the acronym

erezrokah (approved):
I like this approach 🚀 
Not sure about the linting, it did find a few valid issues so we can fix them in the policies or disable it for the policies file.

plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -0,0 +1,167 @@
...
+	index, err := readIndex(dir)
+	if err != nil {
+		log.Fatalf("error reading index: %v", err)
+	}
+
+	var info []*PolicyInfo
+	for _, p := range index.Policies {
+		pi, err := getPolicyInfo(prefix, dir, p)
+		if err != nil {
+			log.Fatalf("error reading index: %v", err)
```
disq, line 159: Should fix the message to differ from the one in L152 :)

plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -0,0 +1,167 @@
...
+			if !strings.HasPrefix(m[1], prefix) {
+				continue
+			}
+			q.Tables = append(q.Tables, m[1])
+			q.Path = sqlPath
+		}
+
+		titleMatches := reTitle.FindAllStringSubmatch(content, -1)
+		if len(titleMatches) == 0 {
+			log.Printf("WARN: Failed to find title for query in %v", sqlPath)
```
disq, line 107: Nice log level :)

hermanschaaf: Haha, yeah, this is a bit WIP. I think we should probably fail the script if this happens, so that it needs to be fixed in the query first.

disq:
Nice, writing these things in Go (compared to bash) makes them extra maintainable and easy to understand.

PR #5345, opened 2022-12-04
docs: Update add plugin to monorepo doc

erezrokah:
#### Summary

Brings https://www.notion.so/cloudquery/Adding-a-new-plugin-to-the-monorepo-f216b653dbe648b2b3512fb8d59a8f89 (internal doc) to this repo.

At some point I think we can use this doc to automate some of the steps if we need to


yevgenypats (approved):
Looks great. I think no need to automate it for now given the pace of adding plugins is not that high so it's quite easy to follow those steps.

Issue #3557, opened 2022-11-03
chore(ci): Add workflow that adds a summary on changed/updated resources


erezrokah:
This could help catch https://github.com/cloudquery/cloudquery/pull/3546

hermanschaaf:
This change should enable us to compare the schemas of two plugin versions in CI: https://github.com/cloudquery/plugin-sdk/pull/347

Issue #2769, opened 2022-10-13
chore(ci): Detect when column types change and ensure we release it as a breaking change


erezrokah:
When we change a column type we should release it as a breaking change if it can't be migrated.
Since it's easy to miss by reviewers (🤚 me in https://github.com/cloudquery/cloudquery/pull/2763).

We should find a way to test for such changes, and alert developers.

### Possible solutions
1. Do it in our testing framework
2. Detect it via changes to the table doc files

bbernays:
I think we should also try to detect when a column gets renamed or deleted

disq:
Related #3557

erezrokah:
I think we can close this in favor of https://github.com/cloudquery/cloudquery/issues/3557

hermanschaaf:
An automated change recently removed a field, so I think we do need some check that ensures it's at least marked as a breaking change: [https://github.com/cloudquery/cloudquery/commit/e6b1e3b10b96c6a5ceb28389efc91da4d9aca590#diff-da3e874014b83b329fa368f[…]2445c406a63fa2c93fc02f1b6L28](https://github.com/cloudquery/cloudquery/commit/e6b1e3b10b96c6a5ceb28389efc91da4d9aca590#diff-da3e874014b83b329fa368f4009df201d465e882445c406a63fa2c93fc02f1b6L28)

erezrokah:
This is done via https://github.com/cloudquery/cloudquery/pull/5377. If https://github.com/cloudquery/cloudquery/pull/5377 works well for automated PRs we can consider enforcing it for manual PRs too (though I think the summary in https://github.com/cloudquery/cloudquery/pull/5347 should be enough)

PR #5347, opened 2022-12-04
chore(ci): Add comment with summary for changed tables

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/3557.
First step for https://github.com/cloudquery/cloudquery/issues/2769.

Example in https://github.com/cloudquery/cloudquery/pull/5375. I chose a flat list, sorted by tables, then for each table breaking changes come first.
I think a flat list can make it easier to add changelog entries to the PR, but we can always update the format of the comment.

I'll follow up next with a PR to fail renovate updates if there's a breaking change in a non major version (https://github.com/cloudquery/cloudquery/issues/2769).

Another possible solution is to use the plugins `doc --format json` command, but that requires checking out the code from `main`, running the command for each plugin, then doing the same for the PR branch, then comparing the JSON files.

> Possible improvement, detect order changes to PKs as breaking


.github/workflows/docs_changes_summary.yml
```diff
@@ -0,0 +1,69 @@
...
+jobs:
+  doc-changes:
+    defaults:
+      run:
+        working-directory: scripts/table_diff
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
hermanschaaf, line 18: Naive question: does this not check out the PR code when used on `pull_request_target`?

hermanschaaf (changes requested):
Looks great! Just want to check I'm not missing something about the safety of pull_request_target when it comes to checking out code

.github/workflows/docs_changes_summary.yml
erezrokah: `pull_request_target` runes in the context of the `main` branch so this checks out the `main` branch code.
In order to checkout the PR's code, one needs to pass the PR `sha` explicitly by doing:
```
- uses: actions/checkout@v2
      with:
        ref: ${{ github.event.pull_request.head.sha }}
```

more in https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

erezrokah:
> Just want to check I'm not missing something about the safety of pull_request_target when it comes to checking out code

Thanks for the review, and great point. Commented on the `pull_request_target` usage

.github/workflows/docs_changes_summary.yml
hermanschaaf: Ah okay, got it, thanks for the explanation!

PR #5377, opened 2022-12-05
chore(ci): Fail renovate updates with breaking changes

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/2769.
I think it's best to only enforce this check as a failure for renovate as an initial approach.
If that works well, we can enforce it on humans too.


.github/workflows/docs_changes_summary.yml
```diff
@@ -5,14 +5,20 @@ on:
   pull_request_target:
     branches:
       - main
-
+    types:
+      - opened
+      - edited
```
erezrokah, line 10: `edited` is needed so the workflow is triggered when a developer fixes the PR title to include a breaking change

.github/workflows/docs_changes_summary.yml
```diff
@@ -67,3 +73,42 @@ jobs:
...
+        if: steps.enforce.outputs.status == 'true'
+        run: npm install @conventional-commits/parser
+      - uses: actions/github-script@v6
+        if: steps.enforce.outputs.status == 'true'
+        with:
+          script: |
+            const {
+              parser,
+              toConventionalChangelogFormat,
+            } = require("@conventional-commits/parser");
```
erezrokah, line 99: I chose this package as this is what `release-please` uses https://github.com/googleapis/release-please/blob/f10f3b49a0cd048f84e2fcd399bbcee74bd3f47c/package.json#L70

.github/workflows/docs_changes_summary.yml
```diff
@@ -2,17 +2,23 @@ name: Summarize changes to source plugins docs
 
 on:
   # Using pull_request_target works on forked PRs too. This is safe since we don't checkout the PR code (we only use the diff)
-  pull_request_target:
+  pull_request:
```
erezrokah, line 5: I'll remove this before merging

.github/workflows/docs_changes_summary.yml
```diff
@@ -67,3 +73,42 @@ jobs:
...
+            try {
+              const ast = parser(title);
+              const { notes } = toConventionalChangelogFormat(ast);
+              const isBreaking = notes.some(({ title }) =>
+                title.includes("BREAKING CHANGE")
+              );
+              if (!isBreaking) {
+                const titleParts = title.split(":");
+                const expectedTitle = `${titleParts[0]}!:${titleParts[1]}`;
+                throw new Error(`PR title does not contain breaking change, please update the title to '${expectedTitle}'`);
```
erezrokah, line 110: If this works well we might want to just updated the title ourselves

PR #5378, opened 2022-12-05
fix(aws): Unused EIP policy - check associations not instance_ids

gavinclarkeuk:
#### Summary

The unused EIP policy check is generating a lot of false positives because it only checks for association to an EC2 instance. this misses various other valid associations (e.g. NAT Gateways, AWS Transfer servers).


erezrokah (approved):
Looks good to me, thanks @gavinclarkeuk! I've asked @bernays to review as well

bbernays (approved):
LGTM. Great catch! 🚢 🚀

hermanschaaf (approved):

PR #5445, opened 2022-12-06
chore(aws): Correct resource ARNs

candiduslynx:


erezrokah (commented):
Thanks for the PR @candiduslynx 🚀 If I got this right the idea is to use the auto generated constants from `client/partitions.go` to reference regions/service names in our code? Is that correct? If so can you show an example how these constants will be used?

candiduslynx:
decided to postpone

PR #5450, opened 2022-12-06
fix(cli): Flush sentry before exiting

erezrokah:
#### Summary

This is unrelated to the issue with plugins panics not showing up in sentry, but we still need to do it for the CLI.
I think this is the only place we actually send something to sentry in the CLI


erezrokah:
No need for this as @disq pointed out we're using `NewHTTPSyncTransport` which doesn't require a flush

PR #5452, opened 2022-12-06
fix: Prevent concurrent map read & write in snowflake write function

hermanschaaf:
This happens rarely, but sometimes the `workers` map gets accessed by the goroutine while the loop is still writing to the map, which causes a panic.


yevgenypats (approved):
Looks good. I think I need to update the destination plugin API to always batch by tables. This will simplify the implementation for all destination plugins. Also, this will solve the annoying thing on PostgreSQL and other destination where we can't know which table caused the error 😆

erezrokah (approved):

PR #5481, opened 2022-12-07
fix(website): Update CLI to latest version

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/pull/5480


yevgenypats (changes requested):
Where do we use this? Can we deprecate this and remove any workflow that updates that?

yevgenypats (approved):
Oh I see it's not the sites website.

PR #5482, opened 2022-12-07
feat(website): Document Homebrew specific version

erezrokah:
#### Summary

The convention in Homebrew is to use `@<version>` without the `v`, so this required a bit more changes.
Instead of adding to the existing code block, we can also use a callout for it to get more visibility (I'm not sure we want that)

### DIFF
--- a/website/components/mdx/_download-linux.mdx
+++ b/website/components/mdx/_download-linux.mdx
@@ -5,7 +5,7 @@ import { Tabs, Tab } from '../Tabs'
 <Tab>
 
 ```bash copy
-curl -L https://github.com/cloudquery/cloudquery/releases/download/VERSION_CLI/cloudquery_linux_amd64 -o cloudquery
+curl -L https://github.com/cloudquery/cloudquery/releases/download/VERSION_CLI_FULL/cloudquery_linux_amd64 -o cloudquery
 chmod a+x cloudquery
 ```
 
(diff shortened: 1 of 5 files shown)


yevgenypats (changes requested):
Why do we need full and trimmed? can we have just one?

erezrokah:
> Why do we need full and trimmed? can we have just one?

We use `v` in our tags https://github.com/cloudquery/cloudquery/releases/tag/cli-v2.0.25, and the naming convention in Homebrew is to not have `v`, so if we need to chose I think we'll need to change our tags

erezrokah:
We can also do something like `vVERSION_CLI` in the docs

website/components/mdx/_download-linux.mdx
```diff
@@ -5,7 +5,7 @@ import { Tabs, Tab } from '../Tabs'
 <Tab>
 
 ```bash copy
-curl -L https://github.com/cloudquery/cloudquery/releases/download/VERSION_CLI/cloudquery_linux_amd64 -o cloudquery
```
erezrokah, line 8: Not having `cli-` prefix here is a bug also on https://www.cloudquery.io/

erezrokah: Extracted this to a separate PR https://github.com/cloudquery/cloudquery/pull/5484

yevgenypats (approved):

Issue #5483, opened 2022-12-07
bug(website): Version replacement doesn't work when copy pasting code blocks


erezrokah:
To reproduce, copy paste https://www.cloudquery.io/docs/plugins/sources/aws/configuration#simple-example.

The result is:

```yml
kind: source
spec:
 # Source spec section
 name: aws
 path: cloudquery/aws
 version: "VERSION_SOURCE_AWS"
 tables: ["*"]
 destinations: ["postgresql"]
 spec: 
 # AWS Spec section described below
 regions: 
 - us-east-1
 accounts:
 - id: "account1"
 local_profile: "account1"
 aws_debug: false

```

PR #5485, opened 2022-12-07
fix(website): Remove copy from code blocks that have version placeholders

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5483


erezrokah:
Follow up should be done in https://github.com/cloudquery/cloudquery/issues/5486

yevgenypats (approved):

PR #5512, opened 2022-12-08
fix(gcp): Fix context to work when syncing

yevgenypats:
Seems `ctx` never worked given that `ctx` is passed on creation and not on call.

Also, this takes some learning from azure and makes code generation easier.

For context there is no need for `services` anymore and we can create and close the server in the `testing.go` helper function instead of doing this in every single mock function


.github/workflows/source_gcp.yml
```diff
@@ -65,6 +65,9 @@ jobs:
       - name: gen
         if: github.event_name == 'pull_request'
         run: make gen
+      - name: Debug git diff
```
hermanschaaf, line 68: Do we want to keep this?

erezrokah: I just added it since we're not sure why it's failing

plugins/source/gcp/client/client.go
```diff
@@ -19,9 +19,10 @@ const maxProjectIdsToLog int = 100
 
 type Client struct {
 	// plugin   *plugins.SourcePlugin
-	projects []string
+	projects      []string
+	ClientOptions []option.ClientOption
 	// All gcp services initialized by client
-	Services *Services
+	// Services *Services
```
hermanschaaf, line 25: Nit: let's completely remove this

hermanschaaf: Is the `Services` struct still used anywhere after this? Can it be removed? (I don't have the code checked out locally right now to check)

hermanschaaf (approved):
The fix makes sense to me, and I spot-checked a few resources - looks good, but I didn't go through it line by line :)

PR #5522, opened 2022-12-08
feat(gcp): Add retrier

### LINKED ISSUE #5514: GCP Retry

In the GCP plugin if an API is throttled it just returns an error.

Request that return with `5XX` or `429` should be retried

yevgenypats:
This adds an initial support for backoff retry for `ResourceExhausted` errors.

Closes #5514


plugins/source/gcp/client/retrier.go
```diff
@@ -0,0 +1,34 @@
...
+	if !ok {
+		return 0, false
+	}
+	c := st.Code()
+	for _, rc := range r.codes {
+		if c == rc {
+			if r.curRetrier >= r.maxRetries {
+				return 0, false
+			}
+			r.curRetrier++
```
disq, line 29: Can we log if it went into a retry, and due to what error etc? Otherwise it will be a quiet log while in the background we retry things

disq: [Example](https://github.com/cloudquery/cq-provider-gcp/blob/main/client/retry.go#L23)

yevgenypats: Good point. we should prob then add a way/configuration to log the grpc call as otherwise just having retry calls without everything else would be hard to understand.

yevgenypats: actually looking into that now it wont be too easy to pass which function it had but I'll add first the logging to the native grpc and it might be enough depending on how the logs are structured to see that a request is being retried. If it will become a big request then we can improve it even further to have the name of the function it is called from but it will be quite involved and "non-standard".

bbernays: Currently I am testing this out and having trouble inspecting what is going on with regards to validating that it is working...

yevgenypats: I see. in that case the logs in the retrier won't be enough. I'll take a look at adding the gcp logs to that PR as well

bbernays:
I found that `gcp_iam_roles` doesn't support this retry mechanism

erezrokah (approved):
Approving per other reviews

PR #5526, opened 2022-12-09
feat: Add more cases where DateTime types in CSV responses of AWS IAM…

CC11001100:
#### Summary

The deserialization of CSV by iam in resource under `cloudquery-aws-provider` does not cover all cases, which may lead to wrong data pull out. For details, we can read the official AWS documents, as mentioned in this document:
```
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_finding-unused.html
```
image

This is also mentioned in another document: 

```
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_getting-report.html
```

image

To sum up, we should add the case `no_information`: 

image

Please review my code, thank you


hermanschaaf (approved):

hermanschaaf:
Thanks for this fix @CC11001100 !

hermanschaaf:
It feels like we should be storing this column as a string (so that this information is not lost), but that's a bigger change, I'm happy with this for now.

yevgenypats (approved):
Thx!

Issue #5444, opened 2022-12-06
BigQuery destination: Sync can fail if new tables were created during migration


hermanschaaf:
It seems like it can take some time for information about BigQuery datasets (and perhaps tables?) to propagate to all read nodes. This can lead to a situation where we create new tables during the migration step, but then the sync fails later when it fails to find a table:

```
➜  cloudquery sync bigquery.yml
Loading spec(s) from bigquery.yml
Source aws will migrate and sync 59 tables.
Starting migration for: aws -> [bigquery]
Migration completed successfully.
Starting sync for: aws -> [bigquery]
Error: failed to write for aws->bigquery: failed to CloseAndRecv client: rpc error: code = Unknown desc = got EOF. failed to wait for plugin: failed to put item into BigQuery table aws_regions: googleapi: Error 404: Not found: Dataset cq-playground:cloudquery, notFound
```

This usually resolves itself within a minute or two, so it can be fixed by re-running the sync after waiting a bit. But this could be a poor first-time user experience, so we should maybe catch this 404 error and retry a few times with delay before erroring out.

(Also worth noting that this was done against a newly created dataset, which could be the issue as well)

PR #5527, opened 2022-12-09
fix: Fix BigQuery JSON credential loading and allow users to set dataset location in config

hermanschaaf:
This change:
 - fixes a bug with JSON credential loading that caused it to ignore the config option
 - allows users to set the dataset location in config, which fixes https://github.com/cloudquery/cloudquery/issues/5444

### DIFF
--- a/plugins/destination/bigquery/client/client.go
+++ b/plugins/destination/bigquery/client/client.go
@@ -17,8 +17,6 @@ type Client struct {
 	spec       specs.Destination
 	metrics    plugins.DestinationMetrics
 	pluginSpec Spec
-	projectID  string
-	datasetID  string
 }
 
 func New(ctx context.Context, logger zerolog.Logger, destSpec specs.Destination) (plugins.DestinationClient, error) {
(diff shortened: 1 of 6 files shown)


plugins/destination/bigquery/client/client.go
```diff
@@ -46,20 +46,23 @@ func New(ctx context.Context, logger zerolog.Logger, destSpec specs.Destination)
...
-	c.datasetID = spec.DatasetID
-	c.pluginSpec = spec
-
 	return c, nil
 }
 
 func (c *Client) bqClient(ctx context.Context) (*bigquery.Client, error) {
 	opts := []option.ClientOption{option.WithRequestReason("CloudQuery BigQuery destination")}
 	if len(c.pluginSpec.ServiceAccountKeyJSON) != 0 {
+		c.logger.Info().Msg("Using service account key JSON")
```
disq, line 55: Can this lead to excessive logging?

hermanschaaf: No, it'll be logged max 3 times per sync (once during initialization, once during migration, once during sync)

hermanschaaf: I can probably remove it now though since it doesn't provide much value.

disq (approved):

candiduslynx (approved):

yevgenypats (approved):

PR #5534, opened 2022-12-10
feat(gcp): Fetch Only enabled services

bbernays:
#### Summary

- [x] Creates a new multiplexer option that enables users to skip the resource if the service has not been determined that it is enabled `client.ProjectMultiplexEnabledServices`
- [x] Provide an opt-in method for filtering out resources/projects that are not enabled


shimonp21:
1) Why should this feature be opt-in? In what use case would a user want us to [unsuccesfully-try-to-grab-data-from-disabled-api's]?
  - making disabled API calls will only ever result in empty tables, right?
 2) IIRC we already had this feature once - but we reverted it because users ran into permissions errors for the "get-enabled-api's". I think we can add this, but need to make sure that permissions-errors don't stop fetches (i.e. best-effort).

bbernays:
> 1. Why should this feature be opt-in? In what use case would a user want us to [unsuccesfully-try-to-grab-data-from-disabled-api's]?
> 
> * making disabled API calls will only ever result in empty tables, right?
> 
> 2. IIRC we already had this feature once - but we reverted it because users ran into permissions errors for the "get-enabled-api's". I think we can add this, but need to make sure that permissions-errors don't stop fetches (i.e. best-effort).

In my view most optimizations and new features that add API calls should be released initially via opt-in mechanism until we are sure of the stability and impact on a larger range of environments. Overall I think we have to align on what is the use case and experience our defaults should be optimizing for

In this case the fact that we reverted the feature in the past is an exact reason to release it behind a an opt-in flag initially.

This feature can add significant latency to a sync if a user is syncing lots of projects (default rate limit is 600 req/minute)... The impact of syncing disabled APIs is:
1. No data for that project in a specific table
2. Lots of log messages saying the API has either been disabled or not used
3. Warnings will appear in user SIEM systems

shimonp21:
I see - If this feature is slow, than yeah - SGTM for making it opt-in.

out-of-scope of this PR: I would also consider exploring not-logging the "api-is-disabled" errors at all (or moving those to DEBUG log-level).

plugins/source/gcp/client/client.go
```diff
@@ -369,3 +385,51 @@ func setUnion(a []string, b []string) []string {
 	}
 	return union
 }
+
+func (c *Client) configureEnabledServices() error {
```
disq, line 389: Pass in `ctx` from Configure here so that it can be used below in `errgroup`.

plugins/source/gcp/client/client.go
```diff
@@ -30,9 +36,12 @@ type Client struct {
 
 	ClientOptions []option.ClientOption
 	CallOptions   []gax.CallOption
-	// this is set by table client multiplexer
+
+	EnabledServices map[string]map[string]bool
```
disq, line 40: I would use `struct{}` here instead of `bool`, sets are simpler to understand with struct{} as it can only have one value.

plugins/source/gcp/client/client.go
```diff
@@ -369,3 +385,51 @@ func setUnion(a []string, b []string) []string {
...
+		project := p
+		if err := goroutinesSem.Acquire(ctx, 1); err != nil {
+			return err
+		}
+		g.Go(func() error {
+			defer goroutinesSem.Release(1)
+			cl := c.withProject(project)
+			svc, err := cl.fetchEnabledServices(ctx)
+			esLock.Lock()
+			c.EnabledServices[project] = svc
```
disq, line 404: If `fetchEnabledServices` fails we set `nil` svc for the project here. If the `Configure` is turned into a warning later (instead of stopping execution) this will present an issue in the form of a panic.

disq (changes requested):
Looking good, small issues

plugins/source/gcp/client/client.go
bbernays: Why would it panic? we are checking both for not nil and for boolean == true:
`cl.EnabledServices[projectId] != nil && cl.EnabledServices[projectId][enabledService]`

disq: Oh ok didn't see that 👍

plugins/source/gcp/client/client.go
bbernays: Not sure I follow...

disq: The suggestion was to use `map[string]map[string]struct{}` so that it's clear that the `bool` value (of `false`) isn't used (so it's a set, rather than a map of "important" values). Not really a deal breaker but nice to have (it doesn't hinder readability that much)

plugins/source/gcp/resources/services/serviceusage/services_fetch.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugins/source/gcp/client"
+	"google.golang.org/api/iterator"
+)
+
+func fetchServices(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	if len(c.EnabledServices) > 0 {
+		for _, svc := range c.EnabledServices[c.ProjectId] {
+			res <- svc.(*pb.Service)
```
disq, line 17: Isn't  `svc` a `map[string]interface{}`?

bbernays: `svc` is just `interface{}` the skipped key value (`_`) is a `string`

plugins/source/gcp/client/services.go
```diff
@@ -0,0 +1,63 @@
+package client
+
+var GcpServices = map[string]bool{
+	"apikeys.googleapis.com":               true,
```
yevgenypats, line 4: Can we make those consts and then add this to the array. Similar how this is done https://github.com/cloudquery/cloudquery/blob/main/plugins/source/azure/client/namespaces.go

yevgenypats (changes requested):
LGTM. two minor comments.

plugins/source/gcp/client/services.go
bbernays: I don't think that `const` would make any sort of improvement because we are still just dealing with strings in the codegen recipes...

In a future PR I would like to autogenerate this whole map like we do with the AWS data json file

yevgenypats (approved):

Issue #5519, opened 2022-12-08
IAM SSH public keys are not covered


ukbea:
### Why do you need this resource or plugin? Please describe.
We need to scan the public ssh keys of an IAM user in order to identify potential security risks caused by public keys that are not rotated.

### Describe the solution you'd like
It should be straight forward to include the IAM resource model AwsIamSshPublicKey using ListSSHPublicKeys call from IAM service.
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/iam#Client.ListSSHPublicKeys

### Additional context
NA

erezrokah:
Hi @ukbea this would be a great addition, thank you for requesting it. Can you please let me know if https://github.com/cloudquery/cloudquery/pull/5538/files#diff-90a4f7a5d1118d22b2be21fe4dbeebbc7bbe0b9299b558858b4ff047b9527444 is what you're after? It will sync the keys metadata, but not the content. That would require another API call to https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/iam#GetSSHPublicKeyOutput.

I believe getting the metadata should be cover the issue you opened, but please let me know if that's not the case.

ukbea:
@erezrokah Sorry for delayed response. Yes we are not concerned about the content but rather metadata about the keys so we can figure out the status and the ıupload date of it.
Thanks for quick support and again sorry the delay.

erezrokah:
Thanks for the follow up and confirming @ukbea. This table is released in AWS `v7.4.0`. Please note you'll need to configure `aws_iam_users` to sync the keys data, as they keys are a child of `aws_iam_users`:
https://github.com/cloudquery/cloudquery/blob/f169b9a24a6ead4798296a3a4d7170192727d19e/plugins/source/aws/docs/tables/aws_iam_ssh_public_keys.md

PR #5538, opened 2022-12-11
feat(aws): Add IAM SSH Public Keys

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5519. In order to call `ListSSHPublicKeys` we need the user name, hence it being a relation of users.

This gets the key metadata, if we want the content too we need another API call to https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/iam#Client.GetSSHPublicKey


hermanschaaf (approved):
LGTM, though I think having the public key in a column would be useful in many applications, so we may want to add that too.

erezrokah:
> LGTM, though I think having the public key in a column would be useful in many applications, so we may want to add that too.

Thanks, I think we can always add it (won't be a breaking change). The use case from the issue is to enforce rotation, so I'm favoring for not doing another API call until someone requests it.

hermanschaaf:
@erezrokah Yeah, not a blocking comment, sounds good 👍

PR #5544, opened 2022-12-12
fix: BigQuery destination: wait for table migrations to complete

hermanschaaf:
This adds additional wait blocks to the BigQuery destination migration code to ensure that we do not return from `Migrate` before the tables are confirmed to be either created or matching the expected schema, depending on the case. This helps prevent an issue where BigQuery returns a 404 when a write immediately follows the creation of the table (as is usually the case when you run `cloudquery sync` for the first time).


yevgenypats:
Nice find. Is there an API in bigquery which can gurantee that when it completes the table is created?

hermanschaaf:
> Is there an API in bigquery which can gurantee that when it completes the table is created?

@yevgenypats Not as far as I know, or at least I couldn't find anything.

erezrokah (approved):

PR #5548, opened 2022-12-12
feat: Vercel source plugin

disq:
- [x] Read / auto-detect teams in Configure
- [x] Domains
- [x] Domain Records
- [x] Teams
- [x] Team Members
- [x] Deployments
- [x] Deployment Checks
- [x] Projects
- [x] Project Environments


plugins/source/vercel/.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
```
erezrokah, line 1: Let's remove this and the relevant make target, see https://github.com/cloudquery/cloudquery/pull/5559

plugins/source/vercel/Makefile
```diff
@@ -0,0 +1,39 @@
...
+.PHONY: install-tools
+install-tools:
+	@echo Installing tools from tools/tool.go
+	@cat tools/tool.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+
+# Install pre-commit hooks. This requires pre-commit to be installed (https://pre-commit.com/)
+.PHONY: install-hooks
+install-hooks:
+	pre-commit install
+
```
erezrokah, line 21:
```suggestion
```

See https://github.com/cloudquery/cloudquery/pull/5559

plugins/source/vercel/README.md
```diff
@@ -0,0 +1,7 @@
+# Vercel Plugin
```
erezrokah, line 1: We'll need to add `Vercel` to https://github.com/cloudquery/cloudquery/blob/51c68f8e934192a27ac6250a66df62e2b06d9797/.github/styles/Vocab/Base/accept.txt

erezrokah (approved):
🚀

plugins/source/vercel/internal/vercel/vercel.go
```diff
@@ -0,0 +1,91 @@
...
+		if res.StatusCode == http.StatusTooManyRequests {
+			val := res.Header.Get("X-Ratelimit-Reset")
+			if val != "" {
+				t, err := strconv.ParseInt(val, 10, 64)
+				if err == nil && t > 0 {
+					ts := time.Unix(t, 0)
+					val = ts.Format(time.RFC3339) + fmt.Sprintf(" (in %s)", time.Until(ts).Round(time.Second))
+				}
+
+				return nil, fmt.Errorf("request to %s failed: %s. Rate limit will reset at: %s", path, res.Status, val)
```
erezrokah, line 83: Probably a follow up, but we can add retry logic if the reset window is not large

plugins/source/vercel/README.md
disq: Yeah, it will end with merge conflicts and pain so I'm waiting until the review is over

yevgenypats (changes requested):
Looks good. One major comment and then I'll give it another round of reviews:

Can we avoid using mocks here and use the same approach we now use in GCP and Azure with a test http server and providing a test `baseURL` or a test `httpClient` ? This should reduce the amount of boilerplate, make it inline with all other plugins and give even better test coverage.

plugins/source/vercel/Makefile
```diff
@@ -0,0 +1,29 @@
+# Test unit
+.PHONY: test
+test:
+	go test -timeout 3m ./...
+
+# Install tools
+.PHONY: install-tools
+install-tools:
+	@echo Installing tools from tools/tool.go
```
yevgenypats, line 9: I think this is deprecated given mockgen is not used.

yevgenypats (approved):
Nice! Looks great. Had one nit.

PR #5550, opened 2022-12-12
fix: Fix deadlock in BigQuery destination if error occurs during write

hermanschaaf:
This fixes a deadlock that occurs if the BigQuery destination encounters an error that causes one of the writers to fail. Previously, the loop would have continued to try and send items to the channel, but these would not get processed. We now check that `gctx` is not done, and exit the loop either when `gctx` is done (so a worker error occurred) or `r, ok := <-res` has `ok` as false, which means the `res` channel was closed.

This issue might also happen in some other destinations, I will take a look there shortly.


plugins/destination/bigquery/client/write.go
```diff
@@ -95,8 +95,18 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *pl
 		})
 	}
 
-	for r := range res {
-		workers[r.TableName].writeChan <- r.Data
+	done := false
+	for !done {
```
candiduslynx, line 99:
```suggestion
	for more := true; more; {
		select {
		case r, ok := <-res:
			if !ok {
				more = false
				break
			}
			workers[r.TableName].writeChan <- r.Data
		case <-gctx.Done():
			more = false
		}
	}
```
Better not do extra flip

candiduslynx (changes requested):

plugins/destination/bigquery/client/write.go
hermanschaaf: `more` is not really accurate here though - we're exiting because something went wrong, not because there aren't any more entries

candiduslynx (approved):

PR #5552, opened 2022-12-12
feat(postgres): Log table name with pg errors

erezrokah:
#### Summary

The `Error() string` method of  `PgError` doesn't return the table name. This PR ensures we log the table name with the error reported:
https://github.com/jackc/pgconn/blob/4c048d40d859bbb702bff9b46cfefd44b89d82c1/errors.go#L52


plugins/destination/postgresql/client/write.go
```diff
@@ -20,6 +21,14 @@ var cqStatusToPgStatus = map[schema.Status]pgtype.Status{
 	schema.Present:   pgtype.Present,
 }
 
+type pgErrorWrap struct {
+	*pgconn.PgError
+}
+
+func (p *pgErrorWrap) Error() string {
+	return p.PgError.Error() + " (Table: " + p.PgError.TableName + ")"
```
erezrokah, line 29: We can also reimplement instead of calling the base `Error()`. Reimplementing might be more robust

plugins/destination/postgresql/client/write.go
```diff
@@ -39,12 +48,13 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *pl
...
 			if err := br.Close(); err != nil {
-				if _, ok := err.(*pgconn.PgError); !ok {
+				var pgErr *pgconn.PgError
+				if !errors.As(err, &pgErr) {
 					// not recoverable error
 					return fmt.Errorf("failed to execute batch: %w", err)
 				}
 				atomic.AddUint64(&c.metrics.Errors, 1)
-				c.logger.Error().Err(err).Msgf("failed to execute batch with pgerror")
+				c.logger.Error().Err(&pgErrorWrap{PgError: pgErr}).Msg("failed to execute batch with pgerror")
```
hermanschaaf, line 57:
```suggestion
				c.logger.Error().Err(err).Str("table", table.Name).Msg("failed to execute batch with pgerror")
```

Would something like this not achieve the goal of adding the table name to the error log?

erezrokah: I don't think so, as we only send the batch if `batch.Len() >= c.batchSize` so if we take the table name from the loop it will print the last table and not necessarily the one generated the error

hermanschaaf: Ah okay I see - but yeah having it as a log field sounds good 👍

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/5552/commits/0eecc4a68965feb8ab217d6788d2548c08772056

PR #5553, opened 2022-12-12
fix: Fix potential deadlock in Snowflake destination

hermanschaaf:
(Similar to: https://github.com/cloudquery/cloudquery/pull/5550)

If a worker returns an error, it will no longer be consuming from its `writeChan`, and in this case we should exit the loop early to avoid deadlock.


hermanschaaf:
@erezrokah Maybe I'm missing something, but the error should get returned by the call to `return eg.Wait()` at the bottom, so not sure what more we need to do. I haven't tested it for Snowflake specifically, but in the BigQuery case this error bubbled up and got reported on the CLI.

erezrokah:
> @erezrokah Maybe I'm missing something, but the error should get returned by the call to `return eg.Wait()` at the bottom, so not sure what more we need to do. I haven't tested it for Snowflake specifically, but in the BigQuery case this error bubbled up and got reported on the CLI.

You're right, I missed that. No need for a follow up

PR #5561, opened 2022-12-12
chore: Fix lint targets

erezrokah:
#### Summary

1. Some plugins did not use the shared linting configuration
2. Timeout is already configured in https://github.com/cloudquery/cloudquery/blob/51c68f8e934192a27ac6250a66df62e2b06d9797/plugins/.golangci.yml#L9
3. Verbose seems a bit too much, but I'm open to adding it to all plugins if it makes sense

### DIFF
--- a/plugins/destination/bigquery/Makefile
+++ b/plugins/destination/bigquery/Makefile
@@ -8,4 +8,4 @@ test:
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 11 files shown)


erezrokah:
Replaced by https://github.com/cloudquery/cloudquery/pull/5562 as we need to not run from a fork for some CI tests to pass

PR #5564, opened 2022-12-12
feat(azure): Add keyvaults  and network_gateways

yevgenypats:
Add more missing tables from previous sdk version


erezrokah (approved):
Looks good, added 1 non blocking comment

Issue #5571, opened 2022-12-13
Fix(postgres-plugin): Add support to use custom schema name in postgres plugin 


khannaabhi:
### Description 
The schema name used in Postgres plugin is hardcoded to "public" and cannot be override by `search_path` query param in the connection string

### Expected Behavior
On running cloudquery sync if `search_path` query param is in the connection string, tables should be created in the schema name provided and the data should be written in that schema.

### CloudQuery Version
cloudquery version 2.0.21

PR #5572, opened 2022-12-13
fix(postgres-plugin): added support to use custom schemaname in postr…

khannaabhi:
…es plugin

#### Summary
Cloudquery sync command was not able to migrate tables to a custom schema when we provided `search_path` in the connection string. The schema name "public" is hardcoded in postres plugin

Issue #5571

### DIFF
--- a/plugins/destination/postgresql/client/client.go
+++ b/plugins/destination/postgresql/client/client.go
@@ -100,7 +100,12 @@ func New(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (pl
 	if err != nil {
 		return nil, fmt.Errorf("failed to get current database: %w", err)
 	}
-	c.currentSchemaName = "public"
+
+	c.currentSchemaName, err = c.currentSchema(ctx)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


shimonp21:
@khannaabhi Thank you for opening this PR! But I believe this doesn't change any behavior actually: the `currentSchemaName` field is not actually used anywhere as far as I can tell.

In my tests, I was able to `sync` to a custom schema with the following config:
```
kind: destination
spec:
  name: "postgresql"
  registry: github
  path: "cloudquery/postgresql"
  version: "v1.8.0"

  spec:
    connection_string: "postgresql://postgres:pass@localhost:5432/postgres?search_path=shimon"
```

Can you provide the error-logs from the migration-failure you are seeing? Could it be that the schema doesn't exist when the `sync` starts? (we don't have code that creates a nonexistant schema).

khannaabhi:
Yes, it is working. I might not have created schema beforehand

shimonp21:
Happy to hear! I'm closing this pull request in that case :)

PR #5611, opened 2022-12-13
feat(azuredevops): Create client, fetch project ids

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/5511. This bootstrap the Azure client and manually adds a table with only one Id field.
The next PR will auto gen tables

Issue #5511, opened 2022-12-08
New Source Plugin: Azure DevOps


jsonpr:

erezrokah:
Marked this as `good first issue`. If someone wants to contribute https://github.com/cloudquery/cloudquery/blob/fee3b2e1df69f61e4bdfb5b0e599fb442a75f16f/plugins/source/azuredevops/codegen/recipes/core.go is a good example how to scaffold resources

bbernays:
I am going to close this because it seems like we now have a [azuredevops source plugin](https://github.com/cloudquery/cloudquery/pull/5558). 

If we are missing specific resources I suggest we open new issues for those resources

@erezrokah @jsonpr

erezrokah:
👍 We haven't added it to the Website though, but we can if someone requests it

bbernays:
I have created a new issue for [that work](https://github.com/cloudquery/cloudquery/issues/6585)...

PR #5618, opened 2022-12-13
feat: Add codegen

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/5611, related to https://github.com/cloudquery/cloudquery/issues/5511.

This adds basic code gen for AzureDevOps (only tables).
It also has `projects` generated, still with no primary key (not sure if `Id` is unique across the API, we'll need to check)


plugins/source/azuredevops/codegen/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"log"
+
+	"github.com/cloudquery/cloudquery/plugins/source/azuredevops/codegen/recipes"
+)
+
+func main() {
+	var resources []*recipes.Resource
+	resources = append(resources, recipes.Core()...)
```
candiduslynx, line 11: I'm not sure, but maybe we could do `All()` func in `recipes` package that will do append on its own?

plugins/source/azuredevops/codegen/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+	for _, resource := range resources {
+		if err := resource.Generate(); err != nil {
```
candiduslynx, line 14: use errgroup, like this:
```go
grp, _ := errgroup.WithCOntext(context.Background())
for _, resource := range resources {
	grp.Go(resource.Generate)
}
if err := grp.Wait(); err != nil {
	log.Fatal(err)
}
```

plugins/source/azuredevops/codegen/recipes/base.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/google/uuid"
+	"github.com/iancoleman/strcase"
+)
+
+//go:embed templates/*.go.tpl
+var templatesFS embed.FS
+
+type Resource struct {
+	Service    string
```
candiduslynx, line 24: This might come up later, but name override is almost always required.

plugins/source/azuredevops/codegen/recipes/base.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func renderTemplate(name string, filePath string, data interface{}) error {
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		return fmt.Errorf("failed to get caller information")
+	}
+	dir := path.Dir(filename)
+
+	tpl, err := template.New(name).Funcs(template.FuncMap{"ToCamel": strcase.ToCamel}).ParseFS(templatesFS, "templates/*.go.tpl")
```
candiduslynx, line 37: posibly extract this tpl parse code into a separate func? will come in handy for mocks.
See https://github.com/cloudquery/cloudquery/blob/main/plugins/source/tailscale/codegen/resources/template.go

plugins/source/azuredevops/codegen/recipes/base.go
```diff
@@ -0,0 +1,114 @@
...
+	if err := tpl.Execute(&buff, data); err != nil {
+		return fmt.Errorf("failed to execute template: %w", err)
+	}
+	outputPath := path.Join(dir, "../..", filePath)
+	outputDir := path.Dir(outputPath)
+	if err := os.MkdirAll(outputDir, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create directory %s: %w", outputDir, err)
+	}
+
+	content := buff.Bytes()
```
candiduslynx, line 56: extract write: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/tailscale/codegen/resources/write.go

plugins/source/azuredevops/codegen/recipes/base.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return fieldType == reflect.TypeOf(uuid.UUID{})
+}
+
+func typeTransformer(field reflect.StructField) (schema.ValueType, error) {
+	if isUUID(field.Type) {
+		return schema.TypeUUID, nil
+	}
+
+	return codegen.DefaultTypeTransformer(field)
```
candiduslynx, line 89: you can fallback by `schema.TypeInvalid, nil`
https://github.com/cloudquery/plugin-sdk/blob/fc9c7ab36ee60c29bd235c29cd2f9e3262f5cc17/codegen/column.go#L42

plugins/source/azuredevops/codegen/recipes/base.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (resource *Resource) Generate() error {
+	var err error
+	resource.Table, err = codegen.NewTableFromStruct(
+		fmt.Sprintf("azuredevops_%s_%s", resource.Service, resource.SubService),
```
candiduslynx, line 95: fmt would be overkill for 5 strings, maybe use `+` as bot concatenation of up to 5 strings it's optimized

plugins/source/azuredevops/codegen/recipes/base.go
```diff
@@ -0,0 +1,114 @@
...
+		resource.Struct,
+		codegen.WithTypeTransformer(typeTransformer),
```
candiduslynx, line 97: add PKColumns from start, too, ass it'll help in the future

plugins/source/azuredevops/codegen/recipes/core.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/microsoft/azure-devops-go-api/azuredevops/v6/core"
+)
+
+func Core() []*Resource {
+	resources := []*Resource{
+		{
+			SubService: "projects",
+			Struct:     core.TeamProjectReference{},
```
candiduslynx, line 11: curious, why not a pointer? VIa `new` or `&`?
(in other places we use pointers)

candiduslynx: Why aren't you using the proper type [TeamProject](https://pkg.go.dev/github.com/microsoft/azure-devops-go-api/azuredevops/v6/core#TeamProject) instead of reference (stripped down)?

I presume, list ill get the references, but then you'll need to gen preresolver and replace the item.

Basically, reference doesn't have properties and collections.

[GetProject](https://pkg.go.dev/github.com/microsoft/azure-devops-go-api/azuredevops/v6/core#ClientImpl.GetProject)
this, on the other hands, returns a [TeamProject](https://pkg.go.dev/github.com/microsoft/azure-devops-go-api/azuredevops/v6/core#TeamProject).
You'll still have to add an extra column for [GetProjectProperties](https://pkg.go.dev/github.com/microsoft/azure-devops-go-api/azuredevops/v6/core#ClientImpl.GetProjectProperties).
And a relation for collentions.

plugins/source/azuredevops/codegen/recipes/templates/resource.go.tpl
```diff
@@ -0,0 +1,11 @@
+// Code generated by codegen; DO NOT EDIT.
+
+package {{.Service}}
+
+import (
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func {{.SubService | ToCamel}}() *schema.Table {
```
candiduslynx, line 9: if you want to supposr relations starting from lower case, take a look at https://github.com/cloudquery/cloudquery/blob/main/plugins/source/tailscale/codegen/resources/templates/resource.go.tpl#L10-L14

candiduslynx (changes requested):

plugins/source/azuredevops/codegen/main.go
erezrokah: We're not very consistent with this pattern across plugins, AWS uses append, Azure uses static `init()` with a `var`.
I'll follow the Azure plugin as that's the newest version of code gen we have.

plugins/source/azuredevops/codegen/main.go
erezrokah: What is the impact of adding this code now? We only have 1 resource.

The purpose of the PR is to have basic code gen for AzureDevOps. I'd rather do any optimizations in follow up PRs to keep the PR small.

plugins/source/azuredevops/codegen/recipes/base.go
erezrokah: I think we could add this based on requirement WDYT? No need to add code that is not used

plugins/source/azuredevops/codegen/recipes/base.go
erezrokah: Would it be easier to do it when we actually need to re-use the function? This PR doesn't add mocks and this code is only used here.

plugins/source/azuredevops/codegen/recipes/base.go
erezrokah: Returning `schema.TypeInvalid` seems less clear than using the default transformer, in relaying the message that we want to fallback

plugins/source/azuredevops/codegen/main.go
yevgenypats: Yes, there is no need to add goroutines here.

plugins/source/azuredevops/codegen/recipes/base.go
erezrokah: This is a pattern we use across all plugins. Can you explain `as bot concatenation of up to 5 strings it's optimized`?
I find it easier to understand how the final string looks like when using the `fmt.Sprintf`

plugins/source/azuredevops/codegen/recipes/base.go
erezrokah: I'm not sure we need to add code that doesn't run. I'd rather add features to the code gen when we need them.
The purpose of this PR is not to have full code gen, but to split it to multiple PRs so we can add it incrementally.

plugins/source/azuredevops/codegen/recipes/templates/resource.go.tpl
erezrokah: Thanks for the reference. If we need to add relations we could use that

plugins/source/azuredevops/codegen/recipes/base.go
erezrokah: Happy to extract it to make the code more readable

plugins/source/azuredevops/codegen/recipes/core.go
erezrokah: > curious, why not a pointer?

Good point, I'll change it to a pointer for consistency.

> I presume, list ill get the references

Correct, the list operation returns the references in 1 API call.
Not sure we want to add the additional information unless someone requests it, as it requires additional API calls

plugins/source/azuredevops/codegen/recipes/base.go
candiduslynx: Go optimizes concat of up to 5 strings in compiler: `a + b + c + d + e` will be optimized, so it might be a bit faster way to do it.
As this is just codegen, no need for extra speed, though.

plugins/source/azuredevops/codegen/recipes/core.go
candiduslynx: At the very least [TeamProject](https://pkg.go.dev/github.com/microsoft/azure-devops-go-api/azuredevops/v6/core#TeamProject) has `Capabilities` field (also, `Links` field, but maybe we would want to skip it for now) that might be of interest, as well as `DefaultTeam`.

candiduslynx: decide to postpone till resources impl

candiduslynx (approved):

PR #5547, opened 2022-12-12
feat: Add PagerDuty plugin

shimonp21:
I skipped some resources that were hard to generate

- OnCall resources require, due to the fact that the IDs of all nested structs should be PKs. https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/on_call.go#L26
- [ResponsePlays](https://developer.pagerduty.com/api-reference/6aa5ad7adec98-list-response-plays) got 400 error from server...

PR #5624, opened 2022-12-13
feat: Slack source plugin

hermanschaaf:
This adds Slack as a source plugin. It includes a number of resources for this initial version, as well easy installation into a workspace using an app manifest.


plugins/source/slack/codegen/tables/templates/tables.go.tpl
```diff
@@ -0,0 +1,18 @@
+// Code generated by codegen; DO NOT EDIT.
+
+package plugin
+
+import (
+	"github.com/cloudquery/plugin-sdk/schema"
+{{- range $resource := . }}
```
candiduslynx, line 7: Drop extra variable, like this:
```
{{- range . }}
	"github.com/cloudquery/cloudquery/plugins/source/slack/resources/services/{{ .Service }}"
{{- end }}
```

plugins/source/slack/codegen/tables/templates/tables.go.tpl
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/slack/resources/services/{{ $resource.Service }}"
+{{- end }}
+)
+
+func tables() []*schema.Table {
+	return []*schema.Table{
+    {{- range $resource := . }}
```
candiduslynx, line 14: drop extra variable
```
   {{- range . }}
      {{ .Service }}.{{ .TableFuncName }}(),
   {{- end }}
```

plugins/source/slack/docs/tables/slack_conversation_bookmarks.md
```diff
@@ -0,0 +1,33 @@
...
+|_cq_parent_id|UUID|
+|team_id (PK)|String|
+|id (PK)|String|
+|channel_id (PK)|String|
+|title|String|
+|link|String|
+|emoji|String|
+|icon_url|String|
+|type|String|
+|date_created|Int|
```
candiduslynx, line 26: shoult be `schema.TypeTimestamp`, use type transformer for this

plugins/source/slack/go.mod
```diff
@@ -0,0 +1,43 @@
+module github.com/cloudquery/cloudquery/plugins/source/slack
+
+go 1.19
+
+require (
+	github.com/cloudquery/plugin-sdk v1.11.0
```
candiduslynx, line 6: update

plugins/source/slack/resources/plugin/plugin.go
```diff
@@ -0,0 +1,19 @@
+package plugin
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/slack/client"
+	"github.com/cloudquery/plugin-sdk/plugins"
+)
+
+var (
+	Version = "development"
```
candiduslynx, line 9:
```suggestion
	Version = "Development"
```

plugins/source/slack/resources/services/conversations/conversation_bookmarks_fetch.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+
+	"github.com/cloudquery/cloudquery/plugins/source/slack/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchConversationBookmarks(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	f := func() error {
+		channelID := parent.Get("id").String()
```
candiduslynx, line 13: nice, will need to remember it

plugins/source/slack/resources/services/conversations/conversation_bookmarks_fetch.go
```diff
@@ -0,0 +1,27 @@
...
+				// we do not expect to fetch bookmarks from channels the bot is not
+				// added to
+				return nil
+			}
+			return err
+		}
+		res <- bookmarks
+		return nil
+	}
+	return c.RetryOnRateLimitError("slack_conversation_bookmarks", f)
```
candiduslynx, line 26: should `f` be declared instead of declaring the func right as `RetryOnRateLimitError` param?

plugins/source/slack/resources/services/conversations/conversation_histories_mock_test.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func buildConversationHistoriesMock(t *testing.T, m *mocks.MockSlackClient) {
+	d := &slack.GetConversationHistoryResponse{}
+	err := faker.FakeObject(&d)
+	if err != nil {
+		t.Fatal(err)
+	}
+	d.HasMore = false
+	msgs := make([]slack.Message, 1)
```
candiduslynx, line 19: no need to manually create the slice of len 1 (faker does it for you?), just do `var []slack.Message`

website/custom.css
```diff
@@ -1,7 +1,15 @@
 code {
   @apply text-sm;
 }
-
+.btn {
```
candiduslynx, line 4: does it belong here?

website/data/slack/manifest.json
```diff
@@ -0,0 +1,49 @@
+{
```
candiduslynx, line 1: do we  need this file?
@erezrokah

candiduslynx (changes requested):

website/data/slack/manifest.json
hermanschaaf: Yes we do, it's specific to the Slack plugin to make installation into a workspace with all the necessary permissions easier

website/custom.css
hermanschaaf: Yes, it's for the Slack plugin docs - it renders the button here: https://cloudquery-nhehfbwxm-cloudquery.vercel.app/docs/plugins/sources/slack/overview
Screenshot 2022-12-19 at 10 06 00

plugins/source/slack/resources/services/conversations/conversation_histories_mock_test.go
hermanschaaf: I prefer being explicit about what the test is doing and the fact that we want exactly one message

plugins/source/slack/resources/services/conversations/conversation_bookmarks_fetch.go
hermanschaaf: Up to personal style, I suppose

plugins/source/slack/resources/services/conversations/conversation_bookmarks_fetch.go
hermanschaaf: I need to double-check this is right actually 😅

plugins/source/slack/resources/plugin/plugin.go
hermanschaaf: Sure, but it doesn't matter: https://github.com/cloudquery/plugin-sdk/blob/74c6672e9ebcc6e0e19f309b549be54519be81e1/serve/source.go#L124

plugins/source/slack/docs/tables/slack_conversation_bookmarks.md
hermanschaaf: Good call

plugins/source/slack/codegen/tables/templates/tables.go.tpl
hermanschaaf: I think variables serve a purpose for readability - it helps you know what you're working with, and I'd rather not remove it to save a few characters

plugins/source/slack/client/services/slack.go
```diff
@@ -0,0 +1,55 @@
...
+package services
+
+import (
+	"context"
+	"github.com/slack-go/slack"
+	"io"
+)
+
+//go:generate mockgen -package=mocks -destination=../mocks/slack.go -source=slack.go SlackClient
+type SlackClient interface {
```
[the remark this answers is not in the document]
hermanschaaf: I considered this approach, but I don't think we should do it here. It requires the tests to have internal knowledge of the underlying Slack APIs, when the Slack SDK is supposed to be a wrapper around that.

plugins/source/slack/Makefile
```diff
@@ -0,0 +1,40 @@
...
+	@cat tools/tool.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+
+# Install pre-commit hooks. This requires pre-commit to be installed (https://pre-commit.com/)
+.PHONY: install-hooks
+install-hooks:
+	pre-commit install
+
+.PHONY: lint
+lint:
+	@golangci-lint run --timeout 10m --verbose
```
erezrokah, line 19:
```suggestion
	golangci-lint run --config ../../.golangci.yml
```

See https://github.com/cloudquery/cloudquery/pull/5562

plugins/source/slack/Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+# Install tools
+.PHONY: install-tools
+install-tools:
+	@echo Installing tools from tools/tool.go
+	@cat tools/tool.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+
+# Install pre-commit hooks. This requires pre-commit to be installed (https://pre-commit.com/)
+.PHONY: install-hooks
+install-hooks:
```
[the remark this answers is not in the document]
erezrokah: Explanation in https://github.com/cloudquery/cloudquery/pull/5559

plugins/source/slack/README.md
```diff
@@ -0,0 +1,7 @@
+# Slack Source Plugin
+
+The Slack Source plugin for CloudQuery extracts configuration from a variety of Slack APIs and loads it into any supported CloudQuery destination (e.g. PostgreSQL).
+
+## Links
+
+- [User Guide](https://docs.cloudquery.io/docs/plugins/sources/slack/overview)
```
erezrokah, line 7:
```suggestion
- [User Guide](https://www.cloudquery.io/docs/plugins/sources/slack/overview)
```

We don't use the `docs` domain anymore

plugins/source/slack/client/testing.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/specs"
+	"github.com/golang/mock/gomock"
+	"github.com/rs/zerolog"
+	"github.com/slack-go/slack"
+)
+
+type TestOptions struct{}
+
+func MockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.T, *gomock.Controller) services.SlackClient, _ TestOptions) {
```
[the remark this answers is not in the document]
hermanschaaf: I don't think we should do this unless it makes sense for the specific case at hand.

plugins/source/slack/client/services/slack.go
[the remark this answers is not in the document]
erezrokah: Agree with @hermanschaaf, I opened a related pitfall in HTTP server tests in https://github.com/cloudquery/cloudquery/issues/5656 (where we use a wildcard to avoid needing to figure out the correct URL/arguments)

plugins/source/slack/codegen/recipes/conversations.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/cloudquery/plugin-sdk/codegen"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/slack-go/slack"
+)
+
+func ConversationResources() []*Resource {
+	resources := []*Resource{
+		{
+			SubService:  "conversations",
+			Description: "https://api.slack.com/methods/conversations.list",
```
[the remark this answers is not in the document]
hermanschaaf: 1. the API documentation should be the canonical source: it may include details and descriptions about the columns, limits and possible values not offered by SDK comments, as well as a number of examples
2. the field names in the Go SDK will be in camel case, not the same as the API
3. CloudQuery users are not necessarily familiar with Go, and shouldn't need to care whether our plugins are implemented in Go, Java, Python, or something else
4. for the above reasons we usually link to the API documentation in other plugins as well. There may be exceptions, but I think we should revisit that then.

plugins/source/slack/client/client.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/specs"
+	"github.com/rs/zerolog"
+	"github.com/slack-go/slack"
+)
+
+type Client struct {
+	logger zerolog.Logger
+	spec   specs.Source
+	Slack  services.SlackClient
```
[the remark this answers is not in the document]
hermanschaaf: I don't think we should embed this, for similar reasons @shimonp21 gave here https://github.com/cloudquery/cloudquery/pull/5547#discussion_r1052004850

plugins/source/slack/Makefile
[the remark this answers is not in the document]
hermanschaaf: Is it? I think pretty much all our plugins still have it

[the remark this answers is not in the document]
hermanschaaf: Oh, maybe this changed recently, looks like the version on `main` differs from what I checked against locally.

plugins/source/slack/client/services/slack.go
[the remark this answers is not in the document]
candiduslynx: We err on the side of testing via http router if we can do it via SDK options.
cc: @yevgenypats

plugins/source/slack/scripts/manifest/main.go
```diff
@@ -0,0 +1,13 @@
+package manifest
+
+import (
+	"log"
+	"os"
+)
+
+func main() {
```
erezrokah, line 8: Where do we use this script?

plugins/source/slack/tools/tool.go
```diff
@@ -0,0 +1,8 @@
+//go:build tools
+// +build tools
```
erezrokah, line 2:
```suggestion
```

Do we need `+build`? We're on `go@1.19`

erezrokah (approved):
This looks great 🚀 Added a few non blocking comments

plugins/source/slack/client/multiplexers.go
```diff
@@ -0,0 +1,12 @@
+package client
+
+import "github.com/cloudquery/plugin-sdk/schema"
+
+func TeamMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
+	var l = make([]schema.ClientMeta, 0)
```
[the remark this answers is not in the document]
hermanschaaf: updated 👍

plugins/source/slack/client/rate_limiting.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"math/rand"
+	"time"
+
+	"github.com/slack-go/slack"
+)
+
+// RetryOnRateLimitError will run the given resolver function and retry on rate limit exceeded errors
+// after waiting the specified amount of time.
+func (c *Client) RetryOnRateLimitError(tableName string, f func() error) error {
```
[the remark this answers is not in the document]
hermanschaaf: Thanks, updated 👍

plugins/source/slack/Makefile
```diff
@@ -0,0 +1,40 @@
+# Test unit
+.PHONY: test
+test:
+	go test -timeout 3m ./...
+
+# Install tools
+.PHONY: install-tools
+install-tools:
```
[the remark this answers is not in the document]
hermanschaaf: Updated, though since this was copy-pasted from another plugin, this will be an issue elsewhere as well. I'm seeing fixing this for other plugins as out-of-scope for this PR.

plugins/source/aws/codegen/services/generate.go
```diff
@@ -156,7 +156,10 @@ func Generate() error {
 			return fmt.Errorf("failed to execute template: %w", err)
 		}
 		filePath := path.Join(path.Dir(filename), fmt.Sprintf("../../client/services/%s.go", service.PackageName))
-		formatAndWriteFile(filePath, buff)
+		err := formatAndWriteFile(filePath, buff)
```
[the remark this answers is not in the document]
hermanschaaf: Oh yeah, probably accidentally included in this PR, thanks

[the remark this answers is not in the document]
hermanschaaf: Moved to this PR https://github.com/cloudquery/cloudquery/pull/5793

plugins/source/slack/tools/tool.go
hermanschaaf: No idea; copy-pasted from other plugins

plugins/source/slack/client/multiplexers.go
```diff
@@ -0,0 +1,12 @@
+package client
+
+import "github.com/cloudquery/plugin-sdk/schema"
+
+func TeamMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
+	client := meta.(*Client)
+	clients := make([]schema.ClientMeta, 0, len(client.Teams))
+	for i, team := range client.Teams {
+		clients[i] = client.withTeamID(team.ID)
```
candiduslynx, line 9: prealloc above makes the len = 0, so this will panic: https://go.dev/play/p/O5TBscjbVZU

candiduslynx:
```suggestion
		clients = append(clients, client.withTeamID(team.ID))
```

candiduslynx (changes requested):

plugins/source/slack/client/multiplexers.go
hermanschaaf: Updated 👍  I'll still do some more testing before merging anyway

plugins/source/slack/go.mod
hermanschaaf: Updated

candiduslynx (approved):

plugins/source/slack/resources/services/conversations/conversation_bookmarks_fetch.go
hermanschaaf: I think it's fine to do this as long as we don't plan on changing the behavior of the CQType `String()` methods. Currently they return the contents verbatim

plugins/source/slack/client/rate_limiting.go
```diff
@@ -0,0 +1,49 @@
...
+			continue
+		} else if !ok {
+			return err
+		}
+		randFloat64 := rand.Float64()
+		jitter := time.Duration(randFloat64 * float64(1*time.Second))
+		retryAfter := rateLimitedErr.RetryAfter + jitter
+		c.logger.Info().Str("table", tableName).Msgf("Rate limit exceeded, retrying in %.2fs", retryAfter.Seconds())
+		time.Sleep(retryAfter)
+		continue
```
candiduslynx, line 35:
```suggestion
		switch typed := err.(type) {
			case *slack.RateLimitedError:
				randFloat64 := rand.Float64()
				jitter := time.Duration(randFloat64 * float64(1*time.Second))
				retryAfter := typed.RetryAfter + jitter
				c.logger.Info().Str("table", tableName).Msgf("Rate limit exceeded, retrying in %.2fs", retryAfter.Seconds())
				time.Sleep(retryAfter)
			default:
				if !isRetryable(err) || retries >= maxRetries {
					return err
				}
				randFloat64 := rand.Float64()
				retryAfter := time.Duration(randFloat64 * float64(1*time.Second))
				time.Sleep(retryAfter)
				retries++
				c.logger.Info().Str("table", tableName).Msgf("Got retryable error (%v), retrying in %.2fs", err, retryAfter.Seconds())
		}
```

candiduslynx (approved):
only a comment about extra `continue` statement an the end of `for` loop body (not required)

plugins/source/slack/client/rate_limiting.go
hermanschaaf: Updated 👍

candiduslynx (approved):

plugins/source/slack/client/rate_limiting.go
```diff
@@ -0,0 +1,47 @@
...
+			time.Sleep(retryAfter)
+		default:
+			if !isRetryable(err) || retries >= c.maxRetries {
+				return err
+			}
+			randFloat64 := rand.Float64()
+			retryAfter := time.Duration(randFloat64 * float64(c.backoff))
+			retries++
+			c.logger.Info().Str("table", tableName).Msgf("Got retryable error (%v), retrying in %.2fs (%d/%d)", err, retryAfter.Seconds(), retries, c.maxRetries)
+			time.Sleep(retryAfter)
```
disq, line 32: Ctrl-C responsiveness here wouldn't be good, should get a `ctx` and do a `select` with `<-ctx.Done(): break/return` and `<-time.After(retryAfter)`

plugins/source/slack/client/rate_limiting.go
```diff
@@ -0,0 +1,47 @@
...
+func (c *Client) RetryOnError(tableName string, f func() error) error {
+	retries := 0
+	for err := f(); err != nil; err = f() {
+		switch typed := err.(type) {
+		case *slack.RateLimitedError:
+			randFloat64 := rand.Float64()
+			jitter := time.Duration(randFloat64 * float64(c.backoff))
+			retryAfter := typed.RetryAfter + jitter
+			c.logger.Info().Str("table", tableName).Msgf("Rate limit exceeded, retrying in %.2fs", retryAfter.Seconds())
+			time.Sleep(retryAfter)
```
disq, line 23: Same here.

plugins/source/slack/client/rate_limiting.go
hermanschaaf: I have not observed any issues with ctrl+c: are you sure?

plugins/source/slack/client/rate_limiting.go
```diff
@@ -0,0 +1,47 @@
...
+			c.logger.Info().Str("table", tableName).Msgf("Got retryable error (%v), retrying in %.2fs (%d/%d)", err, retryAfter.Seconds(), retries, c.maxRetries)
+			time.Sleep(retryAfter)
+		}
+	}
+	return nil
+}
+
+func isRetryable(err error) bool {
+	switch {
+	case strings.Contains(err.Error(), http.StatusText(http.StatusInternalServerError)):
```
disq, line 40: Not good form as the `StatusText` could be variable depending on server (it's not, but there's nothing preventing it to be?) so I would check if that's the case. Better way would be to unwrap until you get to a typed httperror and check response.Status int instead, but if the library doesn't do that (wrap the error response status codes in err) it would be just too complicated to add it on as an afterthought.

plugins/source/slack/client/rate_limiting.go
hermanschaaf: Oh wait, I see now - it doesn't take a long time in practice but there is an error in the logs

plugins/source/slack/codegen/recipes/recipes.go
```diff
@@ -0,0 +1,113 @@
...
+func init() {
+	pluralizeClient = pluralize.NewClient()
+	for _, s := range []string{"livedns", "simplehosting"} {
+		pluralizeClient.AddUncountableRule(s)
+	}
+
+	csr = caser.New()
+}
+
+func toSnake(word string) string {
```
disq, line 69: This can be removed and `csr.ToSnake` can be directly used (only in 2 places)

plugins/source/slack/client/rate_limiting.go
hermanschaaf: The library doesn't expose the http error afaict, so this is a best-effort based retry based on a few errors I have actually observed where the message did match the status text

plugins/source/slack/codegen/recipes/recipes.go
```diff
@@ -0,0 +1,113 @@
...
+		Resolver: "client.ResolveTeamID",
+	}
+
+	pluralizeClient *pluralize.Client
+	csr             *caser.Caser
+)
+
+func init() {
+	pluralizeClient = pluralize.NewClient()
+	for _, s := range []string{"livedns", "simplehosting"} {
```
disq, line 62: Don't need these.

website/utils/slack-app-link.ts
```diff
@@ -0,0 +1,6 @@
+
+export const getSlackAppLink = (type, name) => {
+    const manifest = require(`../data/slack/manifest.json`);
```
disq, line 3: Nice.

website/versions/source-slack.json
```diff
@@ -0,0 +1 @@
+{ "latest": "plugins-source-slack-v1.0.0" }
```
disq, line 1: Apparently this is unneeded and would be created by the our CI per @erezrokah 's comment [here](https://github.com/cloudquery/cloudquery/pull/5676#discussion_r1052118194).

hermanschaaf: Yeah, maybe something changed, but initially the lack of this file was causing the build to fail. I can remove it and check again

plugins/source/slack/client/rate_limiting.go
hermanschaaf: I changed it now and it didn't make a difference to the errors in the logs, so I don't think it really matters. But it's probably better to use the context in any case

plugins/source/slack/client/rate_limiting.go
hermanschaaf: Ok, I think I found a way to handle this better... will try updating

yevgenypats (approved):
Looks good! We can release this as first version and improve once we get some usage/feedback.

Issue #5634, opened 2022-12-14
bug(gcp): `gcp_resourcemanager_projects` is missing a primary key


erezrokah:
Probably a combination of `project_id` and `name should work

PR #5636, opened 2022-12-14
fix(gcp-resources): Set primary keys for `gcp_resourcemanager_projects`

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5634.

This is not a breaking change as auto migrate should work (it does I tested it)


erezrokah:
The comment https://github.com/cloudquery/cloudquery/pull/5636#issuecomment-1351123700 is wrong I think we can improve the table diff tool to tell it's a PK change

yevgenypats (approved):

PR #5654, opened 2022-12-14
fix(gcp): Remove error wrapping

erezrokah:
#### Summary

This is not needed anymore I think


yevgenypats (approved):
Let's also remove this from `go.mod`

PR #5655, opened 2022-12-14
fix(gcp-resources): Add `projects/` prefix when listing IAM roles

erezrokah:
#### Summary

Without this the `List` always errors out with `404`


erezrokah:
There's a thing to be said here about doing `/*filepath` in the tests. This doesn't validate we're using the correct arguments or even list function. For example `iamClient.Organizations.Roles.List("organizations/" + "test").PageToken(nextPageToken).Do()` also passes the test

plugins/source/gcp/resources/services/iam/roles_fetch.go
```diff
@@ -17,7 +17,7 @@ func fetchRoles(ctx context.Context, meta schema.ClientMeta, r *schema.Resource,
 		return err
 	}
 	for {
-		output, err := iamClient.Projects.Roles.List(c.ProjectId).PageSize(1000).PageToken(nextPageToken).Do()
+		output, err := iamClient.Projects.Roles.List("projects/" + c.ProjectId).PageSize(1000).PageToken(nextPageToken).Do()
```
candiduslynx, line 20: this works, too, but maybe we want to add `ProjectScope` func to `client` that will return `"projects/" + c.ProjectId` and use it? It this the only place this fix is required?

erezrokah: > It this the only place this fix is required?

This is the only resources that errors out right out of the box on our playground account with a 404. So it's the only one I could find that needs it.

The format differs between resources, some only need the `ProjectId`, and this is inlined at the code gen level, so I'm not sure there's a reason to put it under the client struct

candiduslynx (approved):

PR #5670, opened 2022-12-15
chore(website): Clean up blog meta.json file

erezrokah:
#### Summary

We don't need this as Nextra uses the `frontmatter` title field for the link text and the date for sorting.

`_meta.json` is useful when you want to have a different link text from the page title, add links that are not in the site files (like external ones) or manually sort the link.

See https://cloudquery-jvqt9pqf7-cloudquery.vercel.app/blog vs https://www.cloudquery.io/blog. The only changes in order are when blogs are posted on the same day.


erezrokah:
Need to verify https://github.com/cloudquery/cloudquery/pull/4836

erezrokah:
Our current Website still has a similar issue to https://github.com/cloudquery/cloudquery/pull/4836
https://www.cloudquery.io/blog/introducing-the-tailscale-source-plugin
image

The right link should be [Announcing the CloudQuery BigQuery Destination Plugin](https://www.cloudquery.io/blog/announcing-cloudquery-bigquery-destination) based on the blog list:

image

yevgenypats:
don't we want to control sometimes the url especially when there are things like weird characters in the title?

erezrokah:
> don't we want to control sometimes the url especially when there are things like weird characters in the title?

The URL is always the filename it's not related to the title, see `https://cloudquery-i6jznqe6o-cloudquery.vercel.app/blog/analysing-gcp-cost-with-bigquery-and-cq`

yevgenypats (approved):

PR #5676, opened 2022-12-15
feat(snyk): Snyk Source plugin

candiduslynx:


plugins/source/snyk/README.md
```diff
@@ -0,0 +1,7 @@
+# Tailscale Plugin
```
disq, line 1:
```suggestion
# Synk Plugin
```

plugins/source/snyk/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+The CloudQuery Tailscale plugin pulls configuration out of Tailscale and loads it into any supported CloudQuery destination (e.g. PostgreSQL).
```
disq, line 3: Synk

plugins/source/snyk/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+## Links
+
+- [User Guide](https://docs.cloudquery.io/docs/plugins/sources/tailscale/overview)
```
disq, line 7:
```suggestion
- [User Guide](https://docs.cloudquery.io/docs/plugins/sources/synk/overview)
```

plugins/source/snyk/client/client.go
```diff
@@ -0,0 +1,51 @@
...
+	OrganizationID string
+	organizations  []string
+
+	logger zerolog.Logger
+}
+
+var _ schema.ClientMeta = (*Client)(nil)
+
+func (c *Client) ID() string {
+	return "snyk/" + c.OrganizationID
```
disq, line 25:
```suggestion
	return c.OrganizationID
```
Unless OrgID can be empty...

plugins/source/snyk/client/spec.go
```diff
@@ -0,0 +1,40 @@
...
+	// EndpointURL is optional parameter to override the API URL for snyk.Client.
+	EndpointURL string `json:"endpoint_url,omitempty"`
+}
+
+const (
+	EnvSnykApiKey = "SNYK_API_KEY"
+)
+
+func (s *Spec) getClient(version string) (*snyk.Client, error) {
+	if len(s.APIKey) == 0 {
```
disq, line 27:
```suggestion
	if s.APIKey == "" {
```

Checking for empty string using len is a bit controversial

plugins/source/snyk/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/snyk/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const sentryDSN = "TODO"
```
disq, line 8:
```suggestion
const sentryDSN = "https://c2101fe3c5fd4f91a095b2b37dc6364a@o1396617.ingest.sentry.io/4504333793165313"
```

website/pages/docs/plugins/sources/overview.mdx
```diff
@@ -30,6 +30,7 @@ Official source plugins follow [release stages](#source-plugin-release-stages).
             { name: "Datadog", stage:"Preview" },
             { name: "Tailscale", stage:"Preview" },
             { name: "Vercel", stage:"Preview" },
+            { name: "Snyk", stage:"Preview" },
```
disq, line 33: Add it after Datadog so preview plugins are sorted as a subset?

disq (changes requested):

plugins/source/snyk/README.md
candiduslynx: Better yet, `Snyk` =)

plugins/source/snyk/client/spec.go
candiduslynx: well, len is built into the underlying struct, so it's even faster that comparing to empty string IMO

website/pages/docs/plugins/sources/overview.mdx
candiduslynx: then whole preview part still will require sorting (cloudflare w=needs to go on top for that part?)

plugins/source/snyk/.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
```
erezrokah, line 1: We can delete this file see https://github.com/cloudquery/cloudquery/pull/5559

plugins/source/snyk/client/client.go
disq: This is also tied to the spec.Name sometimes in some plugins.

candiduslynx: orgID shouldn't be empty (only initially?), so will change

plugins/source/snyk/main.go
disq, line 8: This is left as `TODO`, did you force push without pulling?

candiduslynx: ooh

candiduslynx: fixed

.github/pr_labeler.yml
```diff
@@ -1,3 +1,9 @@
+cli:
```
erezrokah, line 1: I'm not sure this change is needed. If we do this it creates the impression that you need to add the entry sorted/at the right place. Also doesn't seem related to this PR

website/versions/source-snyk.json
```diff
@@ -0,0 +1 @@
+{ "latest": "plugins-source-snyk-v1.0.0" }
```
erezrokah, line 1: I don't think we need to add this file, it will be generated by our CI

erezrokah (approved):
Looks good. Commented on the CI stuff the @disq already reviewed the code.

I do think we should not sort the `pr_labeler.yml` entries, as it only creates more work for developers. Also I think we don't need the comments `# destination plugins` and `# source plugins` as the entries have `plugins/source/*` or `plugins/destination/*` in the path so it's pretty clear what's the purpose of them

.github/pr_labeler.yml
candiduslynx: will revert then

website/versions/source-snyk.json
candiduslynx: removed

disq (approved):

plugins/source/snyk/client/spec.go
```diff
@@ -0,0 +1,40 @@
...
+		s.APIKey = os.Getenv(EnvSnykApiKey)
```
yevgenypats, line 28: Can we not read this via the env variable but only support it via `spec` as the spec reader already supports environment variable expansion and file expansion. This way the documentation is consistent with all other plugins.

yevgenypats (changes requested):
I don't think I found the `dependency` API. Can you point me to that? Without it the plugin doesn't have any use-case as far as Im aware.

plugins/source/snyk/client/spec.go
candiduslynx: Tailscale & GitHub plugins have this, too. Should we remove this from them as well?

candiduslynx: > I don't think I found the `dependency` API. Can you point me to that? Without it the plugin doesn't have any use-case as far as Im aware.

Currently there's no Snyk SDK that supports it.
We can either fork the one I'm using and implement it ourselves, or postpone.

yevgenypats: Yes, we should remove this from tailscale and github. then either fork or use directly but that's the only interesting API in snyk

yevgenypats (approved):

PR #5684, opened 2022-12-15
fix(bigquery): BQ batch size

bbernays:

### DIFF
--- a/plugins/destination/bigquery/client/client.go
+++ b/plugins/destination/bigquery/client/client.go
@@ -16,6 +16,7 @@ type Client struct {
 	logger     zerolog.Logger
 	spec       specs.Destination
 	metrics    plugins.DestinationMetrics
+	batchSize  int
 	pluginSpec Spec
 }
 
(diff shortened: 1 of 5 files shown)


disq (approved):

yevgenypats (approved):

plugins/destination/bigquery/client/client.go
```diff
@@ -37,7 +38,7 @@ func New(ctx context.Context, logger zerolog.Logger, destSpec specs.Destination)
 	}
 
 	c.pluginSpec = spec
-
+	c.batchSize = spec.BatchSize
```
candiduslynx, line 41: where is it used?

candiduslynx: specifically, as thi is a new fields, where is it passed to the write itself?

plugins/destination/bigquery/client/client.go
```diff
@@ -16,6 +16,7 @@ type Client struct {
 	logger     zerolog.Logger
 	spec       specs.Destination
 	metrics    plugins.DestinationMetrics
+	batchSize  int
```
candiduslynx, line 19: I see only initialization of this field.
Where is the propagation to the write?

candiduslynx (changes requested):

plugins/destination/bigquery/client/client.go
bbernays: https://github.com/cloudquery/cloudquery/pull/5681

PR #5733, opened 2022-12-17
fix(postgresql): Make pg error clear with table name

yevgenypats:
Sometimes we can have bugs in our type systems and/or destination which we need to find and fix as fast as we can. In PostgreSQL case we are using batch which makes it impossible to know which table and/or query caused it understand what is the bug. This finds the violating query/table before exiting and prints it so users can report us the issue.

### DIFF
--- a/plugins/destination/postgresql/client/write.go
+++ b/plugins/destination/postgresql/client/write.go
@@ -19,7 +19,6 @@ var cqStatusToPgStatus = map[schema.Status]pgtype.Status{
 	schema.Present:   pgtype.Present,
 }
 
-
 // Postgres best performance for CQ per our benchmark is achieved with batch that includes different tables
 // given we can't predict how many objects will be in each table.
 // This is why if we encounter an error we can't know which object/table caused and related the error to the object
@@ -39,8 +38,8 @@ func (c *Client) findErrorInBatch(ctx context.Context, items []*batchItem) error
 }
 
 type batchItem struct {
-	table string
-	sql     string
+	table     string
+	sql       string
 	arguments []interface{}
 }
 
@@ -48,7 +47,7 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *pl
 	var sql string
 	batch := &pgx.Batch{}
 	batchItems := make([]*batchItem, 0, c.batchSize)
-	// resources := 
+	// resources :=
 
 	for r := range res {
 		table := tables.Get(r.TableName)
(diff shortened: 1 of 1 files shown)


erezrokah (changes requested):
Thanks for the PR @yevgenypats can you explain how getting the error from the `pgError` instance doesn't work? We already log it in https://github.com/cloudquery/cloudquery/blob/f4502d45117559e6e5ea5271d7698e5eba47d659/plugins/destination/postgresql/client/write.go#L49

Also do we stop the sync on a batch failure? Before we kept syncing so at least you got some data. Stopping and reporting the offending table does makes sense as users can chose to skip the table and re-sync.

plugins/destination/postgresql/client/write.go
```diff
@@ -21,9 +19,35 @@ var cqStatusToPgStatus = map[schema.Status]pgtype.Status{
...
+// given we can't predict how many objects will be in each table.
+// This is why if we encounter an error we can't know which object/table caused and related the error to the object
+// we we need to find the error and exit/stop the sync.
+// Also, it is not advice to split the batch into different tables (for postgresql) as
+// this will mean increased number of goroutines and will mean more connections which PostgreSQL is not great with.
+// Also, I've benchmarked CopyFrom and batch and seems there is no significant difference in CQ case.
+// maybe in the future we can provide a flag to switch batching mechanism.
+func (c *Client) findErrorInBatch(ctx context.Context, items []*batchItem) error {
+	for _, item := range items {
+		_, err := c.conn.Exec(ctx, item.sql, item.arguments...)
```
erezrokah, line 32: Are queries guaranteed to behave the same the second time we execute them? I'm concerned this can generated unrelated errors

plugins/destination/postgresql/client/write.go
```diff
@@ -35,35 +59,38 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *pl
...
-				if !errors.As(err, &pgErr) {
-					// not recoverable error
-					return fmt.Errorf("failed to execute batch: %w", err)
+			if batchErr := br.Close(); batchErr != nil {
+				if err := c.findErrorInBatch(ctx, batchItems); err != nil {
+					return err
 				}
-				atomic.AddUint64(&c.metrics.Errors, 1)
-				c.logger.Error().Err(pgErr).Str("table", pgErr.TableName).Msg("failed to execute batch with pgerror")
+				return fmt.Errorf("failed to execute batch and was unable to pinpoint table: %w", batchErr)
```
erezrokah, line 74: Doesn't this mean everything was successfully inserted? Maybe log it and not return an error

yevgenypats:
@erezrokah Good catch! Seems you are correct. apparently there are two types of errors. One that is happening on Postgres side and for that we should usually have the `TableName`. Anotherone is happening on the client side while preparing the "Prepared statements" . I opened a a fix for `pgx` - https://github.com/jackc/pgx/pull/1441 .

In the meanwhile i'll close this one and open PR to update pg to `pgx v5` as the fix I opened is for `v5` (already have a PR in place)

PR #5757, opened 2022-12-18
feat(postgresql): Update pgx to v5

yevgenypats:
This is a follow-up to https://github.com/cloudquery/cloudquery/pull/5733 (see discussion there). and needed so we can incorporate this nice bit: https://github.com/jackc/pgx/pull/1441

The https://github.com/jackc/pgx/pull/1441 is not a pre-requirement for this to go through. even better we should prob ship this first and see that `v5` works for us and for our users and if not we can do the same fix for `v4` (we will just have to fork it as I don't think `v4` accepts anymore contributions)


plugins/destination/postgresql/client/write.go
```diff
@@ -45,8 +37,7 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *pl
 					// not recoverable error
 					return fmt.Errorf("failed to execute batch: %w", err)
 				}
-				atomic.AddUint64(&c.metrics.Errors, 1)
-				c.logger.Error().Err(pgErr).Str("table", pgErr.TableName).Msg("failed to execute batch with pgerror")
+				return fmt.Errorf("failed to execute batch with pgerror on table %s: %w", pgErr.TableName, err)
```
erezrokah, line 40: Since this stops the sync (correct me if I'm wrong), maybe add a call to action?
`Try skipping this table via 'skip_tables: ["<table_name>"]' and re-running the sync command`

erezrokah (approved):
Looks good overall. Let's merge this so we can battle test is on `main`

plugins/destination/postgresql/client/write.go
yevgenypats: I think no need, better we get the report and fix it asap. This is basically equivalent of panic, it shouldn't happen mostly as validation should happen in our type system.

Issue #5683, opened 2022-12-15
Azure all the tables are not getting resources.


Prateek-Rajpal:
### Describe the Bug
In Azure Plugin only 70 out of 225 tables are getting data from cloud. In cloudquery logs only 70 services are getting there resolver
client started. 

### Expected Behavior
All the services should start their resolver client and should fetch resources from the cloud .

### CloudQuery Version
I am using cloudquery version 2.0.21 

### Debug Output
Only these tables are starting their resolver clients
``` 
INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=0 module=azure-src resources=1 table=azure_subscription_tenants
2022-12-15T05:16:05Z ERR table resolver finished with error error="GET https://management.azure.com/global/providers/Microsoft.Authorization/roleDefinitions\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\nERROR CODE: MissingSubscription\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"MissingSubscription\",\n    \"message\": \"The request did not have a subscription or a valid tenant level resource provider.\"\n  }\n}\n--------------------------------------------------------------------------------\n" client=xxxx-xxxx-xxxx-xxxx module=azure-src table=azure_authorization_role_definitions
2022-12-15T05:16:05Z INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=1 module=azure-src resources=0 table=azure_authorization_role_definitions
2022-12-15T05:16:05Z INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=0 module=azure-src resources=0 table=azure_network_public_ip_addresses
2022-12-15T05:16:05Z INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=0 module=azure-src resources=0 table=azure_network_public_ip_prefixes
2022-12-15T05:16:05Z INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=0 module=azure-src resources=0 table=azure_network_virtual_wans
2022-12-15T05:16:05Z INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=0 module=azure-src resources=0 table=azure_keyvault_keyvault
2022-12-15T05:16:05Z INF fetch table finished client=xxxx-xxxx-xxxx-xxxx errors=0 module=azure-src resources=0 table=azure_network_application_security_groups
[...]

PR #5758, opened 2022-12-18
feat(azure): Add warning message when a namespace is not registered

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5683.
I'm not sure this is the correct solution (or implementation of this approach).

Another option is to remove the condition in https://github.com/cloudquery/cloudquery/blob/674ef9bbb7bb76fd8a20e6da2d0b29472e09ddd2/plugins/source/azure/client/client.go#L105 and let the plugin sync everything (I tried it, it doesn't generate lots of errors).

> The current state of Azure seems similar to what is implemented in https://github.com/cloudquery/cloudquery/pull/5534

### DIFF
--- a/plugins/source/azure/client/multiplexer.go
+++ b/plugins/source/azure/client/multiplexer.go
@@ -16,13 +16,15 @@ func SingleSubscriptionMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
 	}
 }
 
-func SubscriptionMultiplexRegisteredNamespace(namespace string) func(schema.ClientMeta) []schema.ClientMeta {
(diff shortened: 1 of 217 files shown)


yevgenypats:
I understand that change but I dont understand how this solves https://github.com/cloudquery/cloudquery/issues/5683 . What were the missing tables?

erezrokah:
> What were the missing tables?

They are not missing. We don't sync them because the subscription doesn't have these namespaces registered, see https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-providers-and-types#azure-portal

I think it's like enabled APIs in GCP

erezrokah:
Closing as stale

PR #5761, opened 2022-12-18
fix(azure): Filter out non supported namespaces - WIP - NOT A REAL FIX

erezrokah:
#### Summary

This is related to https://github.com/cloudquery/cloudquery/pull/5758, https://github.com/cloudquery/cloudquery/issues/5683 and https://github.com/cloudquery/cloudquery/issues/5760

It's meant to show we're auto generating resources with invalid namespaces.

Issue #5760, opened 2022-12-18
Bug(azure): Mismatch between auto generated namespaces and actual namespaces


erezrokah:
### Describe the Bug

Most resources in Azure use the `SubscriptionMultiplexRegisteredNamespace` which creates a client per namespace.
The namespaces are auto generated based on the Go packages and are written to:

https://github.com/cloudquery/cloudquery/blob/18003e9cf6ac036a33e88e6cdcd9b626a792c7de/plugins/source/azure/client/namespaces.go#L4

Looks like there are a bunch of non existent namespaces being generated. The supported list can be retrieved via `az provider list | jq -r '.[].namespace'` or by looking at https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers (`az provider list` returns a longer list which is a bit confusing so I'm not sure which one to use).

As a result these resources are always filtered out since they don't get returned there:
https://github.com/cloudquery/cloudquery/blob/18003e9cf6ac036a33e88e6cdcd9b626a792c7de/plugins/source/azure/client/client.go#L104

### Expected Behavior

Don't generate or use the `SubscriptionMultiplexRegisteredNamespace` for namespaces that don't exist

### Steps to Reproduce

Run a sync with the Azure v2 plugin on any account and see that the tables from offending namespaces are never synced (you don't see them in the logs).

**This is a different issue from https://github.com/cloudquery/cloudquery/pull/5758 where the namespace is valid but not registered on the subscription. The namespaces mentioned in this issue are not even returned by the `armresources.NewProvidersClient` API call(s).**

### Possible solution

I'm not sure we even need the `SubscriptionMultiplexRegisteredNamespace` as other than the existence check in https://github.com/cloudquery/cloudquery/blob/18003e9cf6ac036a33e88e6cdcd9b626a792c7de/plugins/source/azure/client/multiplexer.go#L24 we don't use it anywhere.
Maybe we can just not use that multiplexer, remove the namespaces resolving altogether and assume that if the Go API exists we can use it?

erezrokah:
Easier to see the issue in https://github.com/cloudquery/cloudquery/pull/5761

erezrokah:
Fixed by https://github.com/cloudquery/cloudquery/pull/5789

PR #5789, opened 2022-12-19
fix(azure): Use lowercase namespaces

yevgenypats:
This should fix some of the missing tables we were seeing in azure. apparently namespaces are not consistent and can be lowered case, camel cased, something else... So we normalize them all to lower-case (given they are case insensitive) on the client side.


yevgenypats:
Closes:
 https://github.com/cloudquery/cloudquery/pull/5761
 https://github.com/cloudquery/cloudquery/issues/5683

erezrokah (approved):
Looks good but I think this fixes https://github.com/cloudquery/cloudquery/issues/5760 and https://github.com/cloudquery/cloudquery/pull/5761, but not https://github.com/cloudquery/cloudquery/issues/5683 as for https://github.com/cloudquery/cloudquery/issues/5683 we still need https://github.com/cloudquery/cloudquery/pull/5758.

Also could we drop the namespace multiplexer altogether? We don't use the namespace in the List API calls. I tried it for a few API calls in https://github.com/cloudquery/cloudquery/pull/5758 and it doesn't generate a lot of errors (only one errored out with a "not registered service").

The last comment is more a product decision, dropping the namespace multiplexer will make the Azure plugin consistent with GCP's default behavior

yevgenypats:
> 

I don't think we need to drop the multiplexer because otherwise azure will hit disabled APIs and this will be (really) slow. we were actually trying to implement this in GCP but with no success so far due to some issues with their list disabled APIs. 

Regarding the issue - Im not sure what the customer is complaining about - is he not seeing anything at all? specific table. I think the issue needs more elaboration.

yevgenypats:
> ### This PR has the following changes to source plugin(s) tables:
> * ⚠️ BREAKING CHANGE: Table `azure_sql_deleted_servers` was removed

@erezrokah Can we not mark it as breaking. This table was never working.

yevgenypats:
> > ### This PR has the following changes to source plugin(s) tables:
> > 
> > * ⚠️ BREAKING CHANGE: Table `azure_sql_deleted_servers` was removed
> 
> @erezrokah Can we not mark it as breaking? This table was never working.

erezrokah:
@yevgenypats This comment is just FYI, as long as the PR title/body doesn't say it's breaking it won't bump the major version

erezrokah:
> I'm not sure what the customer is reporting about

There are 2 separate issues:
1. We don't provide any information when we filter the tables based on the provider being registered, so users can't know why the tables are not synced. This is fixed in https://github.com/cloudquery/cloudquery/pull/5758
2. We were using the wrong namespaces (https://github.com/cloudquery/cloudquery/issues/5760) which this PR closes

PR #5792, opened 2022-12-19
feat(aws): AWS SES resources: `aws_ses_active_receipt_rule_sets` & `aws_ses_custom_verification_email_templates`

### LINKED ISSUE #5626: AWS::SES::DescribeActiveReceiptRuleSet
### Why do you need this resource or plugin? Please describe.

Want to verify the TlsPolicy and Scan settings in SES.

A new table of SES Active Receipt Rule Sets

### Additional context
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ses#Client.DescribeActiveReceiptRuleSet
https://docs.aws.amazon.com/ses/latest/APIReference/API_DescribeActiveReceiptRuleSet.html
https://docs.aws.amazon.com/ses/latest/APIReference/API_ReceiptRule.html

candiduslynx:
#### Summary

* Added `aws_ses_active_receipt_rule_sets` (closes #5626)
* Added `aws_ses_custom_verification_email_templates`


plugins/source/aws/codegen/main.go
```diff
@@ -103,13 +104,12 @@ func generateResources() ([]*recipes.Resource, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to set parent-child relationships: %w", err)
 	}
+	grp := new(errgroup.Group)
```
erezrokah, line 107: This change seems unrelated to this PR. Can we extract it to a separate one? Also it would be useful to see some benchmarks if this improves the codegen performance

plugins/source/aws/resources/services/ses/configuration_sets_fetch.go
```diff
@@ -12,27 +12,35 @@ func fetchSesConfigurationSets(ctx context.Context, meta schema.ClientMeta, pare
...
 		return err
 	}
-	resource.Item = getOutput
+
+	resource.SetItem(getOutput)
+
 	return nil
 }
+
+func resolveConfigurationSetArn(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
erezrokah, line 42: Can we do this change in another PR? Adding resources using codegen already generates quite a lot of changes to review, so changing primary keys for resources not added makes it harder to review the PR

plugins/source/aws/resources/services/ses/configuration_sets_fetch.go
```diff
@@ -12,27 +12,35 @@ func fetchSesConfigurationSets(ctx context.Context, meta schema.ClientMeta, pare
...
 	}
+
 	return nil
 }
 
 func getConfigurationSet(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource) error {
 	c := meta.(*client.Client)
 	svc := c.Services().Sesv2
-	csName := resource.Item.(string)
+	csName := resource.GetItem().(string)
```
erezrokah, line 30: Other than styling preferences, is there a reason for this change? If we prefer one way or the other maybe we could remove the public `Item`?

plugins/source/aws/resources/services/ses/contact_lists_fetch.go
```diff
@@ -13,27 +13,29 @@ func fetchSesContactLists(ctx context.Context, meta schema.ClientMeta, parent *s
 	c := meta.(*client.Client)
 	svc := c.Services().Sesv2
 
-	config := sesv2.ListContactListsInput{}
-	p := sesv2.NewListContactListsPaginator(svc, &config)
+	p := sesv2.NewListContactListsPaginator(svc, nil)
```
erezrokah, line 16: Similar to other comments, this change seems to belong in another PR (assuming this one focuses on adding resources)

plugins/source/aws/codegen/recipes/ses.go
```diff
@@ -10,97 +11,122 @@ import (
...
+					Resolver: `schema.PathResolver("Metadata.CreatedTimestamp")`,
+				},
+			),
 		},
 		{
 			SubService:          "configuration_sets",
 			Struct:              &sesv2.GetConfigurationSetOutput{},
 			Description:         "https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_GetConfigurationSet.html",
-			SkipFields:          []string{"ConfigurationSetName", "ResultMetadata"},
+			SkipFields:          []string{"ResultMetadata"},
```
erezrokah, line 38: Can we split the refactoring changes (using primary keys instead of skip fields, using name transformer instead of skip fields, etc) from the code that adds the new resources?

Coupling multiple change in the same PR makes it hard to review the resources that were added

erezrokah (changes requested):
I think the changes looks good.
However can you split them into multiple PRs? Specifically:

1. Adding the new resources
2. Using error groups in code gen
3. Refactoring the code gen to use primary keys/name transformer
4. Changing the primary key for `aws_ses_identies`

plugins/source/aws/resources/services/ses/configuration_sets_fetch.go
candiduslynx: No, only style (+ IDK if we'd want to protect the underlying Item in any way @yevgenypats )

candiduslynx:
decoupled to #5802

erezrokah (approved):
🚀

PR #5798, opened 2022-12-19
chore: Use make target for root update docs script

erezrokah:
#### Summary

This is more correct as it also does the Website docs:
https://github.com/cloudquery/cloudquery/blob/91ec60b91e64b6a02b4fdd8f402ae1cf167051bf/plugins/source/aws/Makefile#L31


hermanschaaf (approved):
Ah okay I didn't see this PR before my comment on https://github.com/cloudquery/cloudquery/pull/5799 :)

PR #5799, opened 2022-12-19
chore: Add a script to codegen for all plugins

erezrokah:
#### Summary

Useful for https://github.com/cloudquery/cloudquery/pull/5790


hermanschaaf (approved):
Looks good, but I think this can probably replace `update-docs.sh` now. Care to remove that file while we're here?

PR #5804, opened 2022-12-19
feat(aws): Add Elastic Transcoder resources

### LINKED ISSUE #5627: AWS::ElasticTranscoder::ListPipelines
### Why do you need this resource or plugin? Please describe.
Verify KMS Key settings on Elastic Transcoder Pipelines.

A new table for elastic transcoder pipelines with various settings surfaced as columns including AwsKmsKeyArn.

### Additional context
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/elastictranscoder#Client.ListPipelines
https://docs.aws.amazon.com/elastictranscoder/latest/developerguide/list-pipelines.html

candiduslynx:
#### Summary

Closes #5627.


plugins/source/aws/resources/services/elastictranscoder/presets_fetch.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/elastictranscoder"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchElastictranscoderPresets(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	c := meta.(*client.Client)
+	svc := c.Services().Elastictranscoder
+
+	p := elastictranscoder.NewListPresetsPaginator(svc, nil)
```
hermanschaaf, line 15: I wonder why the AWS SDK supports a `StopOnDuplicateToken` option: should we perhaps be setting that to `true` to avoid infinite loops?

hermanschaaf (approved):

plugins/source/aws/resources/services/elastictranscoder/presets_fetch.go
candiduslynx: Well, the option itself isn't implemented as a custom option. It might be an overkill to do it manually for all pagers, and the `List` resolvers don't do such logic.

PR #5806, opened 2022-12-19
fix(aws)!: Fix PK in `aws_elasticsearch_domains`

candiduslynx:
BEGIN_COMMIT_OVERRIDE
feat(aws): Make `arn` primary key of `aws_elasticsearch_domains`. Rename `id` column to `domain_id`

BREAKING CHANGE: Rename `id` column to `domain_id` for `aws_elasticsearch_domains`

END_COMMIT_OVERRIDE

### DIFF
--- a/plugins/source/aws/resources/services/elasticsearch/tags_fetch.go
+++ b/plugins/source/aws/resources/services/elasticsearch/tags_fetch.go
@@ -1,0 +1,35 @@
+package elasticsearch
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/service/elasticsearchservice"
+	"github.com/aws/aws-sdk-go-v2/service/elasticsearchservice/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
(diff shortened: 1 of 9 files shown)


candiduslynx:
I can add name replacement for `domain_` to empty string, so there will be columns `id`, `name`, `endpoint-options`, but not sure

plugins/source/aws/resources/services/elasticsearch/tags_fetch.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/service/elasticsearchservice"
+	"github.com/aws/aws-sdk-go-v2/service/elasticsearchservice/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func resolveTags(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
erezrokah, line 13: Does this fix anything? Is it worth adding this to the changelog too?

erezrokah (approved):

plugins/source/aws/resources/services/elasticsearch/tags_fetch.go
candiduslynx: No, it is for other es resources

PR #5833, opened 2022-12-20
chore: Add missing code blocks languages, enforce via linter

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/5832 not from a fork.
This adds missing code blocks language to all our docs so the code has syntax highlighting. It also re-enables the markdown linter with only this single rule


website/pages/docs/plugins/destinations/snowflake/overview.md
```diff
@@ -38,7 +38,7 @@ This is the top level spec used by the Snowflake destination plugin.
 
   Example `connection_string`:
 
-  ```
+  ```bash
```
hermanschaaf, line 41: Should this be bash? I don't know what it should be, but bash doesn't use `//` for comments 🙂

hermanschaaf (approved):

website/pages/docs/plugins/destinations/snowflake/overview.md
erezrokah: Good point, change it to `yaml` as it's part of a spec example

Issue #5678, opened 2022-12-15
digitalocean_billing_history


bbernays:
### Describe the Bug

```
2022-12-15T15:38:29Z ERR failed to execute batch with pgerror error="ERROR: null value in column \"invoice_id\" of relation \"digitalocean_billing_history\" violates not-null constraint (SQLSTATE 23502)" module=pg-dest table=digitalocean_billing_history
```
### Expected Behavior

PK should be adjusted to align with API response

bbernays:
`invoice_id` is only present for `invoices` not for `payments`
example data from [docs](https://docs.digitalocean.com/reference/api/api-reference/#operation/billingHistory_list):
```
{
  "billing_history": [
    {
      "description": "Invoice for May 2018",
      "amount": "12.34",
      "invoice_id": "123",
      "invoice_uuid": "example-uuid",
      "date": "2018-06-01T08:44:38Z",
      "type": "Invoice"
    },
    {
      "description": "Payment (MC 2018)",
      "amount": "-12.34",
      "date": "2018-06-02T08:44:38Z",
      "type": "Payment"
    }
  ],
  "links": {
    "pages": {
      "next": "https://api.digitalocean.com/v2/customers/my/billing_history?page=2&per_page=2",
      "last": "https://api.digitalocean.com/v2/customers/my/billing_history?page=3&per_page=2"
    }
  },
  "meta": {
    "total": 5
  }
}
```

PR #5845, opened 2022-12-20
fix(digitalocean)!: Remove invoice_id from PK for billing history

hermanschaaf:
The `billing_history` table doesn't have a unique id we can use for payments, so we should fall back to the default `_cq_id` as a PK. I'm marking this as a breaking change because it changes the PKs for a table.

This is an example response from the API:

```
"billing_history": [
{
    "description": "Invoice for May 2018",
    "amount": "12.34",
    "invoice_id": "123",
    "invoice_uuid": "example-uuid",
    "date": "2018-06-01T08:44:38Z",
    "type": "Invoice"
},
{
    "description": "Payment (MC 2018)",
    "amount": "-12.34",
    "date": "2018-06-02T08:44:38Z",
    "type": "Payment"
}
],
```

Fixes https://github.com/cloudquery/cloudquery/issues/5678

BEGIN_COMMIT_OVERRIDE
fix(digitalocean): Remove invoice_id from PK for billing history

BREAKING CHANGE: Primary key column for `billing_history` changed from `invoice_id` to `_cq_id`

END_COMMIT_OVERRIDE


erezrokah (approved):
Looks great, one could say this is not a breaking change as we already have the data and we only change the constraint (it should be auto migratable I think), but better to play it safe in case someone was relying on the previous PK

yevgenypats (approved):

Issue #5788, opened 2022-12-19
Please add the following tables for aws cloud.


xi3007-prajpal:
### Why do you need this resource or plugin? Please describe.
Please add the following tables for aws:
    1. aws_elasticsearch_instance_types
    2. aws_elasticsearch_versions
    3. aws_managed_prometheus_workspaces
As,I want to use these resources in my project.

### Describe the solution you'd like
You can use the following apis and functions:
- For Elasticsearch instance types : 
https://docs.aws.amazon.com/sdk-for-go/api/service/elasticsearchservice/#ElasticsearchService.ListElasticsearchInstanceTypes 
- For ElasticsearchVersions: 
https://docs.aws.amazon.com/sdk-for-go/api/service/elasticsearchservice/#ElasticsearchService.ListElasticsearchVersions
- For ManagedPrometheus:
https://docs.aws.amazon.com/sdk-for-go/api/service/prometheusservice/#PrometheusService.ListWorkspaces

candiduslynx:
Will do Managed Prometheus in a separate PR

candiduslynx:
Hi @xi3007-prajpal!
I've implemented the required resources, however, the `instance_types` is not a separate resource but a column in `aws_elasticsearch_versions` table.

PR #5867, opened 2022-12-20
feat(aws): Add more Elastic Search resources

candiduslynx:
#### Summary

Implements part of #5788.


plugins/source/aws/codegen/recipes/elasticsearch.go
```diff
@@ -7,26 +7,65 @@ import (
...
+				},
+			),
+		},
+		{
+			Service:      "elasticsearch",
+			SubService:   "vpc_endpoints",
+			Struct:       new(types.VpcEndpoint),
+			Description:  "https://docs.aws.amazon.com/opensearch-service/latest/APIReference/API_VpcEndpoint.html",
+			Multiplex:    `client.ServiceAccountRegionMultiplexer("es")`,
+			PKColumns:    []string{"vpc_endpoint_id"},
```
hermanschaaf, line 67: In other cases like this one in AWS, we opted to rename the field to `id`. See for example [here](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/codegen/recipes/apigatewayv2.go#L24-L28) or [here](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/codegen/recipes/directconnect.go#L25-L30). Let's keep doing that for consistency.

The rule of thumb I follow is basically: keep column names the same as they are on the original struct, unless they are something like `<StructName>ID` or `<StructName>ARN`, in which case it should be renamed to `id` or `arn`--especially if it's part of the PK, since then joins will often need to be performed using these so it's better if it's easily guessable.

plugins/source/aws/resources/services/elasticsearch/versions_fetch.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func resolveInstanceTypes(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	svc := meta.(*client.Client).Services().Elasticsearchservice
+
+	var instanceTypes []types.ESPartitionInstanceType
+	p := elasticsearchservice.NewListElasticsearchInstanceTypesPaginator(svc,
+		&elasticsearchservice.ListElasticsearchInstanceTypesInput{
+			ElasticsearchVersion: aws.String(resource.Item.(string)),
+			MaxResults:           30,
```
hermanschaaf, line 42: Is 30 the max this API supports?

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func describeVPCEndpoints(ctx context.Context, svc services.ElasticsearchserviceClient, res chan<- any, vpcEndpointIDs []string) error {
+	const maxLen = 100
+	if l := len(vpcEndpointIDs); l > maxLen {
+		if err := describeVPCEndpoints(ctx, svc, res, vpcEndpointIDs[:l/2]); err != nil {
+			return err
+		}
+		return describeVPCEndpoints(ctx, svc, res, vpcEndpointIDs[l/2:])
+	}
```
hermanschaaf, line 45: Why is this doing division by 2? It's unlike other resolvers, so I think it deserves a comment in the code, at least

hermanschaaf (changes requested):

plugins/source/aws/resources/services/elasticsearch/versions_fetch.go
candiduslynx: yes

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
candiduslynx: they support amounts from 5 to 100, so we want to utilize as little calls as possible

plugins/source/aws/codegen/recipes/elasticsearch.go
candiduslynx: changed to `id`, also updated pk for `packages`

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/elasticsearchservice"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client/services"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchElasticsearchVpcEndpoints(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
```
hermanschaaf, line 12: Sorry, I thought I commented on this yesterday, but apparently not. Can the list / describe pattern here be implemented as a PreResourceResolver to better make use of the goroutine pool we have in the Plugin-SDK?

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
hermanschaaf: I see - I could be missing something, but I think it's unnecessary to reach for recursion here. Couldn't we call `DescribeVpcEndpoints` N / 100 times with the max number, and once with the remaining items? That would still be optimal in terms of the number of calls.

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
erezrokah: I think this falls under https://github.com/cloudquery/plugin-sdk/issues/338.
If we use `PreResourceResolver` we will only be able to call `DescribeVpcEndpoints` with a single `vpcEndpointID`

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
candiduslynx: We want to make as little API calls as possible, so we want to pass as many IDs to the describe calls as we can

candiduslynx: as for the `/100` - the minimum amount stated is 5 (IDK why, maybe they fixed it), so I'd better err on the safe side

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
hermanschaaf: Ah okay I see, thanks!

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
hermanschaaf: Ohh, I get it now, there is a _minimum_ number of items as well 😅 I missed that before, thanks!

hermanschaaf: I guess it could still be done without recursion in a variety of ways, but I understand why the code is like it is now, so I won't press the point :)

candiduslynx: rewriting for now, if we hit issues with AWS API -- we'll reconsider

plugins/source/aws/resources/services/elasticsearch/vpc_endpoints_fetch.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if len(vpcEndpointIDs) == 0 {
+		return nil
+	}
+
+	return describeVPCEndpoints(ctx, svc, res, vpcEndpointIDs)
+}
+
+func describeVPCEndpoints(ctx context.Context, svc services.ElasticsearchserviceClient, res chan<- any, vpcEndpointIDs []string) error {
+	// DescribeVpcEndpoints supports amounts [5, 100], so, if we have > 100 endpoints to fetch, split in halves
```
hermanschaaf, line 43:
```suggestion
	// DescribeVpcEndpoints supports amounts [5, 100], so, if we have > 100 endpoints to fetch, split in halves to ensure we always request 5 or more items.
```

hermanschaaf (approved):

PR #5892, opened 2022-12-21
fix(ci): Use forked version of `wait-for-vercel` with bug fix

erezrokah:


erezrokah:
This is not working (I tried using `ref` instead of `sha` for getting the deployments, so even if a commit is ignored by Vercel we still get the outdated link). Apparently listing deployments on `refs` (branch names) returns an empty array, assuming since Vercel only sets them on specific commits and not branches

PR #5896, opened 2022-12-21
feat: Pagerduty incident notes

shimonp21:


website/pages/blog/announcing-pagerduty-plugin.md
```diff
@@ -0,0 +1,39 @@
+---
```
hermanschaaf, line 1: Maybe let's separate the blog post into a separate PR, if that's okay?

shimonp21: Yeah obviously, I `checkout -b`ed from the wrong branch :)

shimonp21: adventures in
```bash
git rebase
git reset --mixed HEAD~
git reset --hard main
```

plugins/source/pagerduty/codegen/recipes/base.go
```diff
@@ -45,6 +45,7 @@ type Resource struct {
 	ResponseStructOverride        string
 	ResponseFieldOverride         string
 	RestPathOverride              string
+	ParentIsPointer               bool
```
hermanschaaf, line 48: I'm not a big fan of this... but since changing to something that doesn't require a field like this will be quite a lot of work, I'm willing to look the other way 😄

hermanschaaf (approved):

PR #5898, opened 2022-12-21
feat(azure): Cost management view queries, query views by subscription scope

disq:
**NOTICE** Let's keep this unmerged for now it as might need tweaking to make the blog post viable

`azure_cost_management_views` were coming up empty. Querying with the scope subscription we can get user's previously saved views (in [Cost Analysis](https://portal.azure.com/#view/Microsoft_Azure_CostManagement/Menu/~/costanalysis)) and some data.

Unfortunately the main resource only returns metadata about pivots and aggregations, to get the actual table data/metrics we still need to run the query. Not sure if this is the right way to do it in terms of performance.

Example "views" row:
```
subscription_id | 12345678-1234-1234-1234-123456789012
e_tag           | "fedcda987654321"
properties      | {"kpis": [{"id": "COST_NAVIGATOR.BUDGET_OPTIONS.NONE", "type": "Budget", "enabled": true}, {"type": "Forecast", "enabled": true}], "chart": "Table", "query": {"type": "ActualCost", "dataSet": {"filter": {"and": [{"dimensions": {"name": "PublisherType", "values": ["azure"], "operator": "In"}}, {"dimensions": {"name": "PublisherType", "values": ["azure"], "operator": "In"}}, {"dimensions": {"name": "PublisherType", "values": ["azure"], "operator": "In"}}]}, "sorting": [{"name": "UsageDate", "direction": "Ascending"}], "grouping": [{"name": "ResourceGroupName", "type": "Dimension"}], "aggregation": {"totalCost": {"name": "Cost", "function": "Sum"}, "totalCostUSD": {"name": "CostUSD", "function": "Sum"}}, "granularity": "Daily"}, "timeframe": "None", "includeMonetaryCommitment": false}, "scope": "subscriptions/12345678-1234-1234-1234-123456789012", "metric": "ActualCost", "pivots": [], "currency": "USD", "createdOn": "2022-12-21T14:06:33.1518576Z", "dateRange": "CurrentBillingPeriod", "modifiedOn": "2022-12-21T14:17:14.5524763Z", "accumulated": "False", "displayName": "test2"}
id              | /subscriptions/12345678-1234-1234-1234-123456789012/providers/Microsoft.CostManagement/views/test2
name            | test2
type            | Microsoft.CostManagement/Views
```

Example "view_queries" row:
```
properties      | {"rows": [[0.000203782, 0.000203782, 20221206, "er-function-app-resource-group-example-resources", "USD"], [0.374281415, 0.374281415, 20221206, "er-web-app-resource-group-example-resources", "USD"], [0.197516129032258, 0.197516129032258, 20221214, "abc-test-sg", "USD"], [0.526709677419355, 0.526709677419355, 20221215, "abc-test-sg", "USD"], [0.504763440860215, 0.504763440860215, 20221216, "abc-test-sg", "USD"], [0.526709677419355, 0.526709677419355, 20221217, "abc-test-sg", "USD"], [0.131677419354839, 0.131677419354839, 20221218, "abc-test-sg", "USD"]], "columns": [{"name": "Cost", "type": "Number"}, {"name": "CostUSD", "type": "Number"}, {"name": "UsageDate", "type": "Number"}, {"name": "ResourceGroupName", "type": "String"}, {"name": "Currency", "type": "String"}]}
e_tag           |
id              | subscriptions/12345678-1234-1234-1234-123456789012/providers/Microsoft.CostManagement/query/12345678-1234-5678-90ab-123456789012
location        |
name            | 12345678-1234-5678-90ab-123456789012
sku             |
tags            | {}
type            | Microsoft.CostManagement/query
```

There's a ResourceGuid option which doesn't match anything we have, so we use the "Resource" field which will give us a full id of the resource (`/subscriptions/<uuid>/resourcegroups/<groupname>/providers/<provider>/<type>/<name>` format) so we'll also need to include that id ("full_id" I'm calling it) in `azure_resources`.

The query to join all of them to see cost vs. res is:
```sql
WITH cost_by_guid AS (
SELECT (r->>(cid.pos::int - 1))::text AS res_id, (r->>(ccost.pos::int - 1))::numeric AS cost_usd
FROM azure_costmanagement_view_queries q
JOIN JSONB_ARRAY_ELEMENTS(q.properties->'rows') AS r ON TRUE
JOIN JSONB_ARRAY_ELEMENTS(q.properties->'columns') WITH ORDINALITY cid(col, pos) ON cid.col->>'name'='ResourceId'
JOIN JSONB_ARRAY_ELEMENTS(q.properties->'columns') WITH ORDINALITY ccost(col, pos) ON ccost.col->>'name'='CostUSD'
JOIN azure_costmanagement_views v ON v._cq_id=q._cq_parent_id
WHERE v.name='test3' -- <-- this is the name of your saved view with "resource" field
)
SELECT
	c.*,
	r.name,
	r.kind,
	r.location
FROM cost_by_guid c JOIN azure_resources r ON lower(r.full_id)=lower(c.res_id);
```


plugins/source/azure/resources/services/costmanagement/view_queries.go
```diff
@@ -0,0 +1,58 @@
...
+		Name:     "azure_costmanagement_view_queries",
+		Resolver: fetchViewQueries,
+		Columns: []schema.Column{
+			{
+				Name:     "properties",
+				Type:     schema.TypeJSON,
+				Resolver: schema.PathResolver("Properties"),
+			},
+			{
+				Name:     "e_tag",
```
disq, line 19: This `ETag` to `e_tag` is another issue with the current codegen, I'm not addressing it in this PR.

plugins/source/azure/resources/services/costmanagement/view_queries_fetch.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return err
+	}
+
+	b, err := json.Marshal(item.Properties.Query)
+	if err != nil {
+		return err
+	}
+	var qd armcostmanagement.QueryDefinition
+	if err := json.Unmarshal(b, &qd); err != nil {
```
disq, line 29: Types are not compatible (`ReportConfigDefinition` vs. `QueryDefinition`) so instead of doing a deep conversion we'll use their marshal/unmarshal methods instead.

erezrokah (approved):
Code looks good, I'm assuming this is the data we want for cost analysis

erezrokah (approved):

PR #5900, opened 2022-12-21
fix: Improve linting perf via caching

bbernays:


bbernays:
I have opened a PR with this functionality in the `golangci-lint-action` repo... Once that PR is merged we can go back to using that instead of `bbernays/golangci-lint-action`

[Link to upstream PR](https://github.com/golangci/golangci-lint-action/pull/629)

.github/workflows/source_aws.yml
```diff
@@ -145,14 +145,17 @@ jobs:
           --health-timeout 5s
           --health-retries 5
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
```
erezrokah, line 150: What's the reason we need `fetch-depth` here?

.github/workflows/source_aws.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: bbernays/golangci-lint-action@58bce521b66ed125524e24eeec9522d765a8c2a7
```
erezrokah, line 57: I would put the fork under `cloudquery/` and also use the branch name directly `cloudquery/golangci-lint-action@master` so if we follow up with fixes we don't have to update this `sha`

.github/workflows/source_aws.yml
```diff
@@ -145,14 +145,17 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
       - name: Set up Go 1.x
         uses: actions/setup-go@v3
         with:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
+      - name: Get dependencies
```
erezrokah, line 157: Won't `go build` download dependencies? Maybe we should remove it from all workflows

erezrokah (changes requested):
Thanks @bbernays, added a few comments

.github/workflows/source_aws.yml
bbernays: I made changes to the `policies` workflow so it matched 100% the other workflow... when investigating why the build was taking so long... I can revert all of the changes to that workflow

erezrokah: How about we do it in another PR? One that aligns all our workflows?

erezrokah (approved):
This looks good, we can follow up with our own forked version later on if they don't accept the fix

erezrokah (approved):

PR #5901, opened 2022-12-21
feat: Add PagerDuty IncidentLogs resource

shimonp21:
for IncidentLogEntries, the primary-key (id) is twice-embedded:

https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/log_entry.go#L43
https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/log_entry.go#L30
https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/client.go#L49.

Because we don't support recursive unwrapping, adding this table manually.


hermanschaaf (approved):

plugins/source/pagerduty/codegen/recipes/base.go
```diff
@@ -250,7 +247,8 @@ func SetParentChildRelationships(resources []*Resource) error {
 			name := csr.ToSnake(strings.TrimSuffix(ch, "()"))
 			v, ok := m[name]
 			if !ok {
-				return errors.New("child not found for " + r.SubService + " : " + name)
+				fmt.Println("WARNING: child not found for " + r.SubService + " : " + name)
```
hermanschaaf, line 250: Is this expected to always raise a warning now? Maybe we can add an exception for the manually created table so that it doesn't confuse new contributors?

plugins/source/pagerduty/codegen/recipes/base.go
```diff
@@ -21,18 +20,19 @@ import (
...
+	Description                      string
+	ExtraColumns                     []codegen.ColumnDefinition
+	PKColumns                        []string
+	Table                            *codegen.TableDefinition
+	Multiplex                        string
+	PreResourceResolver              string
+	PostResourceResolver             string
+	Relations                        []string
+	UnwrapEmbeddedStructs            bool
+	UnwrapEmbeddedStructsRecursively bool
```
hermanschaaf, line 35: I think this can be removed now

PR #5910, opened 2022-12-22
chore(go): Do not enforce `gofmt -s`

candiduslynx:


erezrokah (changes requested):
I'm not sure, didn't we decide we want this? https://pkg.go.dev/cmd/gofmt#hdr-The_simplify_command all looks like good things to enforce

PR #5913, opened 2022-12-22
refactor(gcp-codegen): Infer `RequestStructName`, `UnimplementedServerName` and `ResponseStructName`

erezrokah:
#### Summary

`RequestStructName` can be inferred from the `ListFunction`.
`UnimplementedServerName` can be inferred from `RegisterServer`
`ResponseStructName` can be inferred from `RegisterServer` or `StructName + ListFunctionName` (in most cases)


plugins/source/gcp/codegen/recipes/compute.go
```diff
@@ -77,135 +65,109 @@ var computeResourcesAggList = []*Resource{
...
 		ListFunction:    (&compute.NetworksClient{}).List,
 		NameTransformer: CreateReplaceTransformer(map[string]string{"ipv_6": "ipv6", "i_pv4": "ipv4"}),
 	},
 	{
 		SubService:     "ssl_policies",
 		Struct:         &pb.SslPolicy{},
 		NewFunction:    compute.NewSslPoliciesRESTClient,
-		RequestStruct:  &pb.ListSslPoliciesRequest{},
 		ResponseStruct: &pb.SslPoliciesList{},
-		ListFunction:   (&compute.InterconnectsClient{}).List,
```
erezrokah, line 182: This is wrong, we should be using `SslPoliciesClient`. It doesn't impact the current codegen as it only uses the function name `List`

disq (approved):

plugins/source/gcp/codegen/recipes/compute.go
disq: But why is it wrong...

PR #5918, opened 2022-12-22
fix(crowdstrike): Change timestamp column type and add documentation

hermanschaaf:
This:
 - changes the timestamp column to be of timestamp type (it's actually an aliased Go timestamp type)
 - adds back the Crowdstrike documentation removed before

Leaving in draft because I'd like to understand some potential use cases first (maybe by checking some example data)--it looks like `errors` and `metadata` may be part of the API response metadata and should be excluded, and maybe then we need a few more tables / columns before releasing. But I'm not familiar with CrowdStrike, so could be missing something

### DIFF
--- a/plugins/source/crowdstrike/codegen/recipes/incidents.go
+++ b/plugins/source/crowdstrike/codegen/recipes/incidents.go
@@ -1,6 +1,8 @@
 package recipes
 
 import (
+	"github.com/cloudquery/plugin-sdk/codegen"
+	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/crowdstrike/gofalcon/falcon/models"
 )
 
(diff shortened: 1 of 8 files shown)


plugins/source/crowdstrike/codegen/recipes/incidents.go
```diff
@@ -11,6 +13,14 @@ func CrowdScore() []*Resource {
 			SubService: "crowdscore",
 			Struct:     &models.DomainEnvironmentScore{},
 			PKColumns:  []string{"id"},
+			SkipFields: []string{"Timestamp"},
```
erezrokah, line 16: I know this is still in draft, but can we use a custom type transformer instead of re-defining the columns?
Example in https://github.com/cloudquery/cloudquery/blob/a59f2f26a50c997beeb52c4415a49ac589825f99/plugins/source/vercel/codegen/main.go#L51

hermanschaaf:
I chatted with @cqgaurav, we decided to close this until we get access to an account we can query for some example data again.

PR #5941, opened 2022-12-23
feat: Add instance_view to azure_compute_virtual_machines table

yevgenypats:
apparently `az vm list` always returns instance_views column as empty so to get that data we need additional call.


plugins/source/azure/codegen2/recipes/compute_e.go
```diff
@@ -26,6 +26,48 @@ func init() {
...
+			// 		SkipFetch:      true,
+			// 		ExtraColumns: []codegen.ColumnDefinition{
+			// 			{
+			// 				Name:     "id",
+			// 				Type:     schema.TypeString,
+			// 				Resolver: `schema.PathResolver("ComputerName")`,
+			// 			},
+			// 		},
+			// 	},
+			// },
```
erezrokah, line 69:
```suggestion
```

plugins/source/azure/codegen2/main.go
```diff
@@ -162,7 +164,8 @@ func generateTable(parent *recipes.Table, r *recipes.Table) error {
 	}
 	var buff bytes.Buffer
 	if err := tpl.Execute(&buff, r); err != nil {
-		return fmt.Errorf("failed to execute template for %s: %w", r.Table.Name, err)
+		return fmt.Errorf("failed to execute template for  %w", err)
+		// return fmt.Errorf("failed to execute template for %s: %w", r.Table.Name, err)
```
erezrokah, line 168:
```suggestion
```

erezrokah (approved):
Looks good, but maybe we can remove the commented out code

PR #5942, opened 2022-12-23
feat(aws): Support Custom endpoints

bbernays:
#### Summary

This PR enables users to configure a single global endpoint that the AWS Plugin will use to sync resources. This enables users to point CloudQuery at 3rd party tools like LocalStack


yevgenypats (approved):
Awesome!

PR #5943, opened 2022-12-23
doc(website): Add how-to guides section to website

hermanschaaf:
This moves some of our how-to guide-style blog posts into their own section on the website, called "How-to Guides". I hope this will help with discoverability of the various interesting ways you can use CloudQuery data! It also helps to highlight where we still have some gaps in our documentation.

This is just a first attempt to make an iterative improvement. I imagine, as the section grows, we may need to add some different sections / tags / something else.

While looking over these I also decided to shorten some titles and do some light editing, so that it displays more nicely on the index page. Hopefully these minor edits are not too controversial.


hermanschaaf:
Preview: https://cloudquery-qqw4m00r0-cloudquery.vercel.app/

website/components/HowToGuideIndex.tsx
```diff
@@ -0,0 +1,45 @@
...
+  frontMatter?: any;
+}
+
+function sortByTitle(a: Page, b: Page) {
+  return (
+     b.frontMatter?.title?.toLowerCase() < a.frontMatter?.title?.toLowerCase() ? 1 : -1
+  );
+}
+
+export function HowToGuideIndex({ more = "Read the guide" }) {
```
yevgenypats, line 18: Is this one necessary? i.e can't we use the same thing built-in theme as for the blog?

yevgenypats (approved):
Looks great. Only nit is if possible to avoid some of the custom frontend code and use the defaults somehow but we can do this later as it's not much code (for now)

website/components/HowToGuideIndex.tsx
hermanschaaf: We probably can if we provide the blog index with some more options, but the ordering is different for how-to guides (they currently are ordered by title, not by date)

website/vercel.json
```diff
@@ -144,6 +144,46 @@
     {
       "source": "/docs/plugins/sources/aws/multi_account",
       "destination": "/docs/plugins/sources/aws/multi-account"
+    },
+    {
+      "source": "/blog/cloud-asset-inventory-cloudquery-apache-superset",
+      "destination":"c/how-to-guides/loud-asset-inventory-cloudquery-apache-superset"
```
erezrokah, line 150:
```suggestion
      "destination":"/how-to-guides/cloud-asset-inventory-cloudquery-apache-superset"
```

erezrokah: giphy (1)

website/pages/how-to-guides/_meta.json
```diff
@@ -0,0 +1,25 @@
+{
+  "*": {
+    "//": "Disable sidebar and search for these pages",
+    "theme": {
+      "sidebar": false,
+      "typesetting": "article"
+    },
+    "searchable": false
+  },
+  "cloud-asset-inventory-cloudquery-apache-superset": {
```
erezrokah, line 10: What's the reason we need to explicitly set these?

website/components/HowToGuideHeader.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export function HowToGuideHeader() {
+  const config = useConfig();
+  const meta = config.frontMatter as PostMeta;
+
+  if (!meta) {
+    return null;
+  }
+
+  const image = `https://www.cloudquery.io/og-image/${meta.title}`;
```
erezrokah, line 57:
```suggestion
  const image = `/og-image/${meta.title}`;
```

We might want to use a relative URL so deploy previews point to it under their domain

website/components/HowToGuideIndex.tsx
erezrokah: Can we remove the dates from the how to guides? I think that will ensure they get sorted by titles

website/pages/how-to-guides/_meta.json
hermanschaaf: The reason is that I wanted the titles to be slightly different in the index than on the actual post. We have 5 posts that all start with `How to Build an Open Source Cloud Asset Inventory with` and the long titles made the index page look both repetitive and badly spaced. So in the index I called it `How Visualize CloudQuery data with...` to make it shorter but keep the original titles + url paths for SEO reasons.

website/components/HowToGuideHeader.tsx
hermanschaaf: Cool, I'll update it here and in the Blog header where this was copied from

website/components/HowToGuideIndex.tsx
hermanschaaf: Removed 👍

erezrokah (approved):

PR #5953, opened 2022-12-25
Draft: Refactor codegen

shimonp21:
This PR comes together with [sdk-#531](https://github.com/cloudquery/plugin-sdk/pull/531)

- Use golang's struct-inheritance (embedded-struct) instead of writing custom methods to copy-paste properties from codegen "Resources" to codegen "TableDefinitions"
- Move the `GenerateSchema` method to the SDK - no need to copy-paste it for every new plugin.

### Pros
- Reduce code-duplication: no need to copy-paste `generateSchema`, and no need to do write "field-forwarding" methods.
   - Example of "field-forwarding": https://github.com/cloudquery/cloudquery/blob/d65d80fc24bf2ae1fcda502da5de744f86491afd/plugins/source/aws/codegen/recipes/base.go#L145-L198
   - Duplicate `generateSchema` code per plugin:
      https://github.com/cloudquery/cloudquery/blob/d65d80fc24bf2ae1fcda502da5de744f86491afd/plugins/source/aws/codegen/recipes/base.go#L212-L243
- Allow further removal of code-duplication, by moving creation of the table-lists to the SDK

### Cons
- requires recipe resource definitions to change, due to the way golang works (https://gist.github.com/shimonp21/8907b859d60ec40e932172c0f21fa636):
   ```
   Resource {
   TableDefinition {
     Service: "service",
     SubService: "subservice",
     Struct: SomeStruct{},
   },
   // plugin-specific stuff goes here (resolver-and-mock-test parameters, usually).
   }
   ```
   
 ### Methodology
 Apart from two specific recipes that were manually updated, the manual refactor was done with [refactor.py](https://gist.github.com/shimonp21/6fc9c201eb9d56cfc8c3f5e7433bd91d)  
 
 - Only code that needs to be reviewed is in `main.go` and `recipes/base.go`.
    All the rest was refactored with `refactor.py`.


yevgenypats:
Let's close this as we are going to take this approach most prob

PR #5954, opened 2022-12-25
feat(salesforce): New Salesforce source plugin

yevgenypats:
This is a new salesforce source plugin.

Right now this plugin uses one table and stays true to the API with a table that have an `id` (pk), `type` (pk), and `raw` (raw response from rest endpoint).

It is possible to turn it into multi-table but it might involve additional change in the SDK to allow dynamic tables as this is how the API behaves. If this will be need we can then "unwrap" the columns in each table.


bbernays:
This will be the 3rd plugin that requires dynamic tables/columns (csv, Google sheets)... so probably a valuable R&D effort

erezrokah (approved):

Issue #5957, opened 2022-12-26
feat(gcp-codegen): Don't hard code GCP APIs list


erezrokah:
We have the GCP services list hard coded in https://github.com/cloudquery/cloudquery/blob/7bb33a87ad42d98a091638888d40348edb76d9d8/plugins/source/gcp/client/services.go#L3

We should consider auto generating it, maybe using https://developers.google.com/discovery/v1/reference/apis/list

PR #5958, opened 2022-12-26
refactor(gcp-codegen): Extract generateTemplate method

erezrokah:
#### Summary

This PR extracts `generateTemplate` into a separate method to reduce code duplication.
This should make it easier to implement https://github.com/cloudquery/cloudquery/issues/5957

PR #5961, opened 2022-12-26
feat(gcp): Auto generate services list

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5957.

We need to put the services file in both the code gen module and the GCP client as they both use it. Previously code gen referenced the GCP client directly which prevents auto generation of the services file.

It should be easier to review this once https://github.com/cloudquery/cloudquery/pull/5958 is merged.

Added a new makefile target as we should first discover the services, then run the codegen


erezrokah:
We should first merge https://github.com/cloudquery/cloudquery/pull/5958 and then rebase, to make it easier to review this PR

plugins/source/gcp/codegen/client/discover.go
```diff
@@ -0,0 +1,44 @@
...
+		Items []Service `json:"items"`
+	}
+	err = json.Unmarshal(body, &services)
+	if err != nil {
+		return nil, err
+	}
+
+	servicesMap := make(map[string]bool)
+	for _, service := range services.Items {
+		servicesMap[service.Name+".googleapis.com"] = true
```
erezrokah, line 38: This is require since there are multiple versions of the same services with the same name (e.g. v1, v1Beta). The API has a different ID for each version

bbernays: Instead of generating the domain name, why not just parse the DiscoveryURL and extract the hostname:

```
u := url.Parse(service.DiscoveryRestUrl)
hostname := u.Hostname()
servicesMap[hostname] = true
```

erezrokah: I tried that but it doesn't work for some services (e.g. compute):
`"discoveryRestUrl": "https://www.googleapis.com/discovery/v1/apis/compute/alpha/rest"`

bbernays: Fair!

yevgenypats (approved):

Issue #5995, opened 2022-12-27
bug(aws-ses): `aws_ses_active_receipt_rule_sets` generates a lot of resolver errors


erezrokah:
Run a sync with the following config:

```yaml
kind: source
spec:
  name: aws
  version: v8.1.0
  destinations: [postgresql]
  path: cloudquery/aws
  tables: ["aws_ses_active_receipt_rule_sets"]
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v1.10.0"
  write_mode: "overwrite-delete-stale"
  spec:
    connection_string: postgresql://postgres:pass@localhost:5432/postgres
```

See multiple errors in the following format:
```
ERR table resolver finished with error error="operation error SES: DescribeActiveReceiptRuleSet, https response error StatusCode: 400, RequestID: ********, api error InvalidAction: Unavailable Operation" client=******:eu-west-3
```

Apparently only a few regions support that API call https://docs.aws.amazon.com/ses/latest/dg/regions.html#region-receive-email

erezrokah:
Also fails on AWS CLI the same way:
![image](https://user-images.githubusercontent.com/26760571/209644122-f75e8d6f-6c10-404f-9a8e-1b54b38886e3.png)

PR #5997, opened 2022-12-27
fix(aws_ses_active_receipt_rule_sets): Don't call DescribeActiveReceiptRuleSet on unsupported regions

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5995

However I'm also happy to not do this and report the errors.


hermanschaaf:
Hmm, interesting case. I think we should not report these errors, just like we don't report errors for unsupported regions for other services (usually based on https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/client/data/partition_service_region.json). At the same time, ideally we should be able to keep this list of supported regions up-to-date in some way.

Is there a way we can automatically check the list supported regions? If not, maybe we should make the requests but ignore the error if it is "unsupported region"? That way we don't create a maintenance burden for ourselves and the code will always be correct, at the cost of a few additional requests

erezrokah:
> Is there a way we can automatically check the list supported regions? If not, maybe we should make the requests but ignore the error if it is "unsupported region"? That way we don't create a maintenance burden for ourselves and the code will always be correct, at the cost of a few additional requests

Hmm, OK so I think the correct way to do this is to add it to the regions JSON file and add a panic when AWS adds it.
For example https://github.com/cloudquery/cloudquery/pull/5268

hermanschaaf:
@erezrokah Yeah, though I'm not sure what key it will be added under. `email`? That's already there; this seems specific to just this one endpoint?

erezrokah:
> @erezrokah Yeah, though I'm not sure what key it will be added under. `email`? That's already there; this seems specific to just this one endpoint?

You're right I'm just noticing that as well. It seems there isn't an automated way to get that list (other than parsing the AWS docs).

I think ignoring the error for now should work well 👍  I'll update the PR

erezrokah:
@hermanschaaf followed up with https://github.com/cloudquery/cloudquery/pull/5997/commits/ddc2e2e5b3a3f5ed288df13972dd9638ccc6ce05

hermanschaaf (approved):

PR #6021, opened 2022-12-27
chore(go): Make PK order be breaking

candiduslynx:


scripts/table_diff/changes/changes_test.go
```diff
@@ -67,52 +87,40 @@ func Test_getChanges(t *testing.T) {
 					Breaking: true,
 				},
 				{
-					Text:     "Table `azure_appservice_functions`: column added with name `site_id` and type `String`",
-					Breaking: false,
+					Text: "Table `azure_appservice_functions`: column added with name `site_id` and type `String`",
```
erezrokah, line 90: Let's not remove the setting of `Breaking` to `false`, I think it's better to be explicit with the test data, otherwise someone reading the test will have to go to the struct definition to understand what is being tested.
Also it's unrelated to the change in this PR and it makes it hard to review the new test

scripts/table_diff/changes/testdata/pr_6012_diff.txt
```diff
@@ -0,0 +1,1042 @@
+diff --git a/plugins/source/github/codegen/recipes/actions.go b/plugins/source/github/codegen/recipes/actions.go
```
erezrokah, line 1: We can keep only the diff related to the `.md` files to make it easier to understand this file

scripts/table_diff/changes/changes.go
```diff
@@ -66,40 +89,33 @@ func getColumnChanges(file *gitdiff.File, table string) (changes []change) {
...
-			}
-		} else {
+		addedColumn, ok := addedColumns[deletedName]
+		switch {
+		case !ok:
 			changes = append(changes, change{
 				Text:     fmt.Sprintf("Table %s: column removed %s from table", backtickStrings(table, deletedName)...),
 				Breaking: true,
 			})
+		case addedColumn.dataType != deletedColumn.dataType:
```
erezrokah, line 99: I don't think this covers the same cases as before, for example what if `dataType` changes and a PK is added on the same column?
We will only report the first change instead of both, WDYT?

Also, this change seems unrelated to the purpose of this PR, so probably better to separate it as it makes harder to review the PR

scripts/table_diff/changes/changes.go
```diff
@@ -47,13 +48,35 @@ func parseColumnChange(line string) (name string, dataType string, pk bool) {
 	return cleanName, result["dataType"], result["name"] != cleanName
 }
 
+func parsePKChange(line string) (names []string) {
+	for _, match := range pkRegex.FindAllStringSubmatch(line, -1) {
```
erezrokah, line 52: Are we expecting to match the regex multiple times in a single line? If not can we use `FindStringSubmatch`?

scripts/table_diff/changes/changes.go
```diff
@@ -47,13 +48,35 @@ func parseColumnChange(line string) (name string, dataType string, pk bool) {
...
+
 func getColumnChanges(file *gitdiff.File, table string) (changes []change) {
 	addedColumns := make(map[string]column)
 	deletedColumns := make(map[string]column)
+	var addedPK, deletedPK []string
 	for _, fragment := range file.TextFragments {
 		for _, line := range fragment.Lines {
 			name, dataType, pk := parseColumnChange(line.Line)
 			if name == "" || dataType == "" {
+				pkChanges := parsePKChange(line.Line)
```
erezrokah, line 71: Can't we move this up just at the beginning of the loop here:
https://github.com/cloudquery/cloudquery/pull/6021/files#diff-6910d6ee0a1eadb1d8f2122f804f6ab3ef6e4c6d040b631527d79e73b4678c50R69
to avoid the nesting?

erezrokah (changes requested):
Thanks for the PR @candiduslynx 🚀 

I think the approach in this PR will work well. Added a few comments

scripts/table_diff/changes/changes.go
candiduslynx: yes, sure!

erezrokah (approved):
🚀

candiduslynx:
Need to be revisited after discussion at #6043 is complete

PR #6023, opened 2022-12-27
feat: Add azure security pricings resource

amanenk:


erezrokah:
Hi @amanenk, FYI quite a lot has changed with Azure codegen. To manually add resources we prefix the recipe with `_e.go`, see example in https://github.com/cloudquery/cloudquery/pull/5952. You can run `make gen` to generate all the relevant files/docs once you're done (it will create the Website docs missing from your PR).

erezrokah (approved):

PR #6044, opened 2022-12-27
feat(gcp-resources): Add Bare Metal Solution

erezrokah:
#### Summary

API https://cloud.google.com/bare-metal/docs/reference/rest

Still needs to test on real resources


yevgenypats (approved):

erezrokah:
Looks like it's not that straight forward to create this resource https://cloud.google.com/bare-metal/docs. So I think we can rely on the mock tests + the API calls not failing

PR #6045, opened 2022-12-27
fix(github): Fix parent ID column resolvers

candiduslynx:


bbernays (approved):
Looks good, but I didn't test it

PR #6047, opened 2022-12-27
doc(gcp): Update Contributing Docs

bbernays:


plugins/source/gcp/codegen/main.go
```diff
@@ -179,10 +179,15 @@ func generateResource(r recipes.Resource, mock bool) {
 		log.Fatal(fmt.Errorf("failed to create table for %s: %w", r.StructName, err))
 	}
 	if r.Multiplex == nil {
-		if _, ok := client.GcpServices[r.ServiceDNS]; !ok {
-			panic("unknown service DNS: " + r.ServiceDNS)
+		if r.ServiceDNS == "" {
```
erezrokah, line 182: Doesn't this change means that if someone forgets to set `ServiceDNS` they'll use the wrong multiplexer? Instead the should set `Multiplex` directly if they need `ProjectMultiplex`.
If we want to make `ServiceDNS` optional we can use https://github.com/cloudquery/cloudquery/pull/5956/files

erezrokah (changes requested):
Added one comment for clarification on the code gen.
The contributing guide looks good

plugins/source/gcp/codegen/main.go
bbernays: If `ServiceDNS` is empty, then `ProjectMultiplex` is the only `Multiplex` option that will work

erezrokah: Cool but should we always set it? If someone forgets to set it that specific service won't be enforce via the enabled services feature.
If someone wants to opt out from the enabled services feature they should set the Multiplex explicitly no?

erezrokah: I mean if you want to use the `ProjectMultiplex` you can set that one directly instead of setting an empty `ServiceDNS`

plugins/source/gcp/CONTRIBUTING.md
```diff
@@ -34,6 +34,25 @@ The process to follow for adding a new recipe is:
...
+1. `ProjectMultiplex` (_default_): This is the most basic of multiplexers in that it will resolve the resource in each project that is being synced. 
+2. `ProjectMultiplexEnabledServices(serviceDNS string)`:  This multiplexer will only attempt to sync a resource if that project has the service enabled otherwise the resource will be skipped for that specific projectID. On top of this the user must also enable the feature via `enabled_services_only: true` in the spec. In order to use this multiplexer you must specify a valid `resource.ServiceDNS`
+3. `client.OrgMultiplex`: For resources that are unique across an entire Organization. In order to use this multiplexer you have to explicitly set the multiplexer `client.OrgMultiplex
+``` go
+var OrgMultiplex = "client.OrgMultiplex"
+Resource{
+	Multiplex:  &OrgMultiplex   
+}
+```
+
```
disq, line 51: nit: Too many newlines?

erezrokah (approved):

PR #6091, opened 2022-12-28
fix(postgres-spec)!: Move `batch_size` from the plugin spec to the top level spec

erezrokah:
#### Summary

**Not the same** as https://github.com/cloudquery/cloudquery/pull/6092.
The Postgres plugin doesn't use managed batching so its configuration still works. However since it's available via the top level spec we should use that one for consistency in our docs.


candiduslynx (approved):

website/pages/docs/reference/destination-spec.md
```diff
@@ -61,6 +61,14 @@ Specifies the update method to use when inserting rows. The exact semantics depe
 - `overwrite`: Same as `overwrite-delete-stale`, but doesn't delete stale rows from previous `sync`s.
 - `append`: Rows are never overwritten or deleted, only appended.
 
+<!-- vale off -->
+### batch_size
+<!-- vale on -->
+
+(`int`, optional, default: `10000`)
```
erezrokah, line 68: Please note this has a higher default

erezrokah:
Closing as we decided to keep this option at the plugin level

hermanschaaf (approved):

Issue #6070, opened 2022-12-28
Bug(destinations-BigQuery): Batch write failures with `Error 413 (Request Entity Too Large)`


erezrokah:
Could be related to https://discord.com/channels/872925471417962546/1057191097648033833/1057356930928082984 (waiting for further logs).

I was able to reproduce this error on our `cq-playground` account with the following configuration (I kept only the tables that generate the errors):

```yaml
kind: source
spec:
  name: aws
  version: v9.0.1
  destinations: [bigquery]
  path: cloudquery/aws
  tables: [aws_docdb_cluster_parameter_groups, aws_ec2_instance_types, aws_rds_engine_versions]
---
kind: destination
spec:
  name: bigquery
  path: cloudquery/bigquery
  version: "v1.3.1"
  write_mode: "append"
  spec:
    project_id: <project-id>
    dataset_id: <datatest-id>
```

Also, at some point it seems CloudQuery is hanging, the source plugin reports the sync finished, but the CLI is stuck on `Syncing resources`

hermanschaaf:
@erezrokah Hmm, interesting. We should be able to fix this by trying to ensure we don't send batches above the max size, but as a workaround, does it work if you set `batch_size` in the spec to something lower, like 1 or 10?

erezrokah:
Thanks @hermanschaaf I'll try with a smaller batch. I think there's another issue here as it seems the CLI deadlocks (I'm looking into it)

erezrokah:
OK so discovered something else, `batch_size` at the plugin spec doesn't work anymore and need to be set on the top level spec:
```yml
kind: destination
spec:
  name: bigquery
  path: cloudquery/bigquery
  version: "v1.3.1"
  write_mode: "append"
  batch_size: 1 # This works
  spec:
    project_id: cq-playground
    dataset_id: erez_test
    batch_size: 1 # This doesn't
```

I'll open a PR to update the docs, and remove the configuration from relevant plugins

erezrokah:
PR(s) to fix batching configs:
https://github.com/cloudquery/cloudquery/pull/6092
https://github.com/cloudquery/cloudquery/pull/6091

erezrokah:
@hermanschaaf looks like using 1000 as batch size (the default before we switched to managed batching via the SDK) works 🎉 

https://github.com/cloudquery/cloudquery/pull/6092 should resolve this issue

hermanschaaf:
[...]

PR #6092, opened 2022-12-28
fix(bigquery-spec)!: Move `batch_size` from the plugin spec to the top level spec

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/6090 not from a fork so tests can pass.

Fixes https://github.com/cloudquery/cloudquery/issues/6070

With the SDK level batching the configuration that "takes" is now at the top spec level. Setting `batch_size` at the plugin level has no impact


candiduslynx (approved):

erezrokah:
Closing as we decided to keep this option at the plugin level

hermanschaaf (approved):

disq (approved):

PR #6096, opened 2022-12-28
feat: New file destination plugin

yevgenypats:
Docs and tests passing.


.github/workflows/dest_file.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: cloudquery/golangci-lint-action@master
+        with:
+          version: v1.50.1
+          working-directory: plugins/destination/file
+          args: "--config ../../.golangci.yml"
+      - name: Get dependencies
+        run: go get -t -d ./...
+      - name: Build
+        run: go build .
+      - name: Test PG
```
hermanschaaf, line 45:
```suggestion
      - name: Test file plugin
```

plugins/destination/file/client/client.go
```diff
@@ -0,0 +1,57 @@
...
+
+	csvTransformer         *csv.Transformer
+	csvReverseTransformer  *csv.ReverseTransformer
+	jsonTransformer        *json.Transformer
+	jsonReverseTransformer *json.ReverseTransformer
+}
+
+func New(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (destination.Client, error) {
+	if spec.WriteMode != specs.WriteModeAppend {
+		return nil, fmt.Errorf("localfile destination only supports append mode")
```
hermanschaaf, line 29:
```suggestion
		return nil, fmt.Errorf("file destination only supports append mode")
```

hermanschaaf, line 29:
```suggestion
		return nil, fmt.Errorf("file destination only supports append mode")
```

plugins/destination/file/client/deletestale.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (*Client) DeleteStale(ctx context.Context, tables schema.Tables, sourceName string, syncTime time.Time) error {
+	return fmt.Errorf("csv destination doesn't support overwrite-delete-stale mode. please use append mode")
```
hermanschaaf, line 12:
```suggestion
	return fmt.Errorf("file destination doesn't support overwrite-delete-stale mode. please use append mode")
```

plugins/destination/file/client/migrate.go
```diff
@@ -0,0 +1,12 @@
+package client
+
+import (
+	"context"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (*Client) Migrate(ctx context.Context, tables schema.Tables) error {
+	// migrate is not needed in migrate mode
```
hermanschaaf, line 10: What is the behavior intended to be if the columns for a CSV file change between versions, and a sync is run? Do we always overwrite the file? (In which case `append` is a bit confusing) Or will it append values in the incorrect positions or with additional columns not matched by the headers?

plugins/destination/file/client/read.go
```diff
@@ -0,0 +1,48 @@
...
+		return c.csvReverseTransformer.ReverseTransformValues(table, values)
+	case FormatTypeJSON:
+		return c.jsonReverseTransformer.ReverseTransformValues(table, values)
+	default:
+		panic("unknown format " + c.pluginSpec.Format)
+	}
+}
+
+func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []interface{}) error {
+	name := fmt.Sprintf("%s/%s.%s.%s", c.pluginSpec.Directory, table.Name, c.pluginSpec.Format, uuid.NewString())
```
hermanschaaf, line 26: I don't see how this can read an existing file if it's generating a new UUID every time 🤔

plugins/destination/file/client/spec.go
```diff
@@ -0,0 +1,34 @@
+package client
+
+import (
+	"fmt"
+)
+
+type FormatType string
+
+type BackendType string
```
hermanschaaf, line 9: I think this is no longer used

plugins/destination/file/client/spec.go
```diff
@@ -0,0 +1,34 @@
...
+const (
+	FormatTypeCSV  = "csv"
+	FormatTypeJSON = "json"
+)
+
+type Spec struct {
+	Directory string     `json:"directory,omitempty"`
+	Format    FormatType `json:"format,omitempty"`
+	// This is used for debugging purposes only
+	NoRotate bool `json:"no_rotate,omitempty"`
```
hermanschaaf, line 20: Should it be removed, if it's only for debugging? What does it do?

plugins/filetypes/csv/write.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"encoding/csv"
+	"io"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+
+func WriteTableBatch(w io.Writer, table *schema.Table, resources [][]any) error {
```
hermanschaaf, line 11: Why is this so much simpler than it was in the CSV plugin? Is it because we're now using a managed writer?

plugins/filetypes/json/read.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+const maxJsonSize = 1024 * 1024 * 20
```
hermanschaaf, line 12: I'm assuming this is the biggest that a single JSON row is allowed to be? We should probably document this as a comment, or perhaps later make it configurable. 

We will also need to deal with the case when this size is exceeded, because it will happen pretty soon. We had the same problem when reading JSON logs. There we stopped using a `bufio.Scanner` and instead used `bufio.Reader`. Maybe we can find "inspiration" from that implementation: https://github.com/cloudquery/plugin-sdk/blob/main/clients/log_reader.go#L33

The `Read` function is only used in testing right now though, so not a blocker by any means.

hermanschaaf (changes requested):

plugins/destination/file/client/migrate.go
yevgenypats: we always overwrite. also we always write to `.format.uuid` unless `NoRotate` is specified (need to document that indeed). but yeah it is up to the user. implementing migration over files is quite hard

plugins/filetypes/csv/write.go
yevgenypats: Yes!

plugins/destination/file/client/spec.go
```diff
@@ -0,0 +1,34 @@
...
+type BackendType string
+
+const (
+	FormatTypeCSV  = "csv"
+	FormatTypeJSON = "json"
+)
+
+type Spec struct {
+	Directory string     `json:"directory,omitempty"`
+	Format    FormatType `json:"format,omitempty"`
```
hermanschaaf, line 18: Perhaps not for this first iteration, but I think we should also support:
 - a "max size" option so that files are guaranteed to be below that size, and a new file with a different suffix is created when the limit is reached
 - a setting to allow each table to be written to a directory of its own name. e.g. `aws_ec2_instances` should get written to `output/aws_ec2_instances/aws_ec2_instances.csv`, potentially with a way to specify a suffix so that different syncs don't overwrite each other (maybe sync time could be used). This will make it easier to upload to some destinations like S3/Athena (at least until we support those directly, but even then there will be other similar destinations we don't support)

Again, just some ideas while I'm thinking about this - it might require some more thought and design so doesn't need to be in this PR

plugins/destination/file/client/migrate.go
hermanschaaf: Maybe we should use sync time instead of uuid, so the files are at least sorted in increasing order?

plugins/destination/file/Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: test
+test:
+	go test -timeout 3m ./...
+
+.PHONY: lint
+lint:
+	@golangci-lint run --config ../../.golangci.yml --timeout 10m --verbose
```
erezrokah, line 7: We can remove the timeout and verbose flags (we have the timeout in the config file, and verbose generates too much information for it to be useful)

plugins/destination/file/client/spec.go
yevgenypats: It is also for user usage. Added documentation under `website`

plugins/destination/file/client/read.go
yevgenypats: Added documentation under website. we use `NoRotate` in the tests

plugins/destination/file/client/migrate.go
yevgenypats: we need uuid because if those are uploaded to remote storage it can cause issues and uuid is the only way to avoid collision if it is running in distributed mode.

bbernays: You could use a `ksuid`. Segment has a good golang library: https://github.com/segmentio/ksuid

plugins/destination/file/client/spec.go
yevgenypats: Yeah it can be good improvements but I think `max_size` should be on the SDK level at least for managed as they can choose either `size` or `batch_size` or `min(max_batch_size, max_size)`

plugins/destination/file/client/migrate.go
hermanschaaf: Nice, I was also going to suggest a Snowflake ID that @candiduslynx mentioned on another thread the other day: https://en.wikipedia.org/wiki/Snowflake_ID. Seems like an ideal fit for this case

yevgenypats: yes, we can use one of those time based IDs but for now I don't think it's yet needed?

erezrokah: I think there's a version of UUID that is sortable https://github.com/oklog/ulid (or we can just contact time+UUID)

plugins/destination/file/client/spec.go
hermanschaaf: We should update the comment above the line then, it currently says that it's `for debugging purposes only`

erezrokah (approved):
Looks good, still need to test it locally.
Also since we have only 2 file types maybe do it as a package inside the plugin, not sure, to avoid the monorepo complexities, and only move it outside when it's used somewhere else than this plugin.

For example we don't have CI that runs the tests/linting/etc. on this module, so we'll need to create a separate one (or do it inside the file plugin CI)

website/pages/docs/plugins/destinations/file/overview.md
```diff
@@ -0,0 +1,46 @@
...
+  Directory where all CSV files will be written. A CSV file will be created per table.
+
+- `format` (string) (required)
+
+  Format of the output file. `json` and `csv` are supported.
+
+- `no_rotate` (bool) (optional)
+
+  if set to true, the plugin will append to the same file for each table.
+  Otherwise, for every batch a new file will be created with `.uuid` suffix.
```
hermanschaaf, line 46:
```suggestion
  If set to true, the plugin will write to one file per table.
  Otherwise, for every batch a new file will be created with a different `.<UUID>` suffix.
```

plugins/destination/file/client/migrate.go
disq: ksuid is very nice (I've used it lots before) the only downside is they look a bit weird.

plugins/destination/file/client/spec.go
```diff
@@ -0,0 +1,32 @@
...
+	// This is used for debugging purposes only
+	NoRotate bool `json:"no_rotate,omitempty"`
```
hermanschaaf, line 18: Maybe we should call this `Rotate` and set the default to `true`? Maybe it's just me, but it takes me a bit of thinking to figure out whether `NoRotate=false` means it will rotate or not :) 

Also, but separately, I feel like `rotate` should actually default to false. If you just want a quick and dirty solution or have a small setup, you don't need the files to be rotated (which brings in sorting and merging problems). The defaults should fit these users; users that are productionizing their setups can spend a few more minutes to learn about the available options

plugins/destination/file/Makefile
yevgenypats: > Looks good, still need to test it locally. Also since we have only 2 file types maybe do it as a package inside the plugin, not sure, to avoid the monorepo complexities, and only move it outside when it's used somewhere else than this plugin.
> 
> For example we don't have CI that runs the tests/linting/etc. on this module, so we'll need to create a separate one (or do it inside the file plugin CI)

Im going to open a PR for s3, gcs and azure in a moment after this one is merged so we need it to be a package

plugins/destination/file/client/read.go
hermanschaaf: I see, okay. We'll need a different solution when we start supporting cursor-based syncing, but since Read is only used in testing for now it's fine

plugins/destination/file/client/spec.go
yevgenypats: maybe in `local` it makes sense but Im trying to keep it consistent with all our "file" plugins and in most of then it will default to rotate. so don't know what is better - `inconsistency` or one of better default. I would vote for consistency across all plugins (of the same type)

hermanschaaf: Fair enough, but can we still name it `rotate` with a documented default of `true`?

yevgenypats: golang has issues with `nil` values so it will mean I'll need to turn this into `*bool` this is why I made it `NoRotate`

hermanschaaf: Okay, yeah I see. Not a big deal, let's leave it like it is. At least the meaning of `no_rotate: true` is pretty clear, and that's the only time you'd be adding it to config (because otherwise you can use the default)

hermanschaaf (approved):
LGTM apart from the linters failing right now

disq (approved):
The filename format could use a hash of all columns in table order (esp. in CSV) but maybe later I guess

PR #6108, opened 2022-12-28
feat: Add S3 destination

yevgenypats:
Still need to add tests (with workload identity) and documentation.


plugins/destination/s3/client/client.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	c.s3Client = s3.NewFromConfig(cfg)
+	c.uploader = manager.NewUploader(c.s3Client)
+	c.downloader = manager.NewDownloader(c.s3Client)
+
+	// we want to run this test because we want it to fail early if the bucket is not accessible
+	if _, err := c.uploader.Upload(ctx, &s3.PutObjectInput{
+		Bucket: aws.String(c.pluginSpec.Bucket),
+		Key:    aws.String("cq-test-file"),
```
hermanschaaf, line 69: I wonder if this file might confuse whatever process users use to ingest the data from S3 (Glue Crawler, Databricks, etc). Is it really necessary? The error will surface as soon as the first batch is written anyway. 

If we keep it, we should try and give it some name or property that will make sure it gets ignored by other processes (not sure if that's possible). Or otherwise maybe we should try and delete it? The downside there is that our process may not have `s3:DeleteObject` permissions.

erezrokah (approved):

plugins/destination/s3/client/spec.go
```diff
@@ -0,0 +1,32 @@
...
+type Spec struct {
+	Bucket   string     `json:"bucket,omitempty"`
+	Key      string     `json:"key,omitempty"`
+	Format   FormatType `json:"format,omitempty"`
+	NoRotate bool       `json:"no_rotate,omitempty"`
+}
+
+func (*Spec) SetDefaults() {}
+
+func (s *Spec) Validate() error {
```
erezrokah, line 23: Should we validate `Key`? The docs say it's required

erezrokah (commented):
Added a single comment

website/pages/docs/plugins/destinations/s3/overview.md
```diff
@@ -0,0 +1,46 @@
...
+
+<Badge text={"Latest: " + getLatestVersion("destination", "s3")}/>
+
+This destination plugin lets you sync data from a CloudQuery source to remote S3 storage in various formats such as CSV, JSON.
+
+This is useful in various use-cases, especially in data lakes where you can query the data direct from Athena or load it to various data warehouses such as BigQuery, RedShift, Snowflake and others.
+
+## Example
+
+This example configures a CSV destination, to create CSV files in `s3://bucket_name/path/to/files`. Note that the CSV plugin only supports `append` write-mode.
```
hermanschaaf, line 14:
```suggestion
This example configures a CSV destination, to create CSV files in `s3://bucket_name/path/to/files`. Note that the S3 plugin only supports `append` write-mode.
```

.github/workflows/dest_s3.yml
```diff
@@ -0,0 +1,89 @@
...
+  plugins-destination-s3:
+    timeout-minutes: 30
+    name: "plugins/destination/s3"
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./plugins/destination/s3
+    permissions:
+      id-token: write
+      contents: read # This is required for actions/checkout@v2
```
erezrokah, line 27:
```suggestion
      contents: read # This is required for actions/checkout@v3
```

plugins/destination/s3/client/spec.go
yevgenypats: Yeah, Im also going to change it to `Path` to be consistent with all others destinations

website/pages/docs/plugins/destinations/s3/overview.md
```diff
@@ -0,0 +1,46 @@
...
+
+  Bucket where to sync the files.
+
+- `key` (string) (required)
+
+  Path to where the files will be uploaded in the above bucket.
+
+- `format` (string) (required)
+
+  Format of the output file. `json` and `csv` are supported.
```
hermanschaaf, line 46: I think we should also include a section on how to authenticate to S3, even if it's just a link to the AWS source plugin docs.

hermanschaaf (approved):
LGTM, just one or two comments :)

erezrokah (changes requested):
We'll need to add this to the wait for all list, see https://github.com/cloudquery/cloudquery/blob/549e2904070da2c86c745ab7db573fbaaf95570b/contributing/adding_a_new_plugin_to_cq_monorepo.md

plugins/destination/s3/client/client.go
yevgenypats: Yeah I thought about the deletion but we will have permissions issues for sure. maybe we can try do an empty file ? I don't think there is a good way to test it ?

hermanschaaf: Yeah maybe let's start with an empty file if that works, that should be less confusing to a crawler. Also it might help to have the name start with a `.`, like `.cq-test-file`. Let's start with that and see if it causes any issues

website/pages/docs/plugins/destinations/s3/overview.md
bbernays: I think we need examples...

plugins/destination/s3/client/spec.go
bbernays: Why not call it `prefix`?

PR #6153, opened 2022-12-30
feat(stripe): Stripe source plugin

disq:


plugins/source/stripe/resources/services/invoices/invoice_items_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/cloudquery/cloudquery/plugins/source/stripe/client"
+	"github.com/cloudquery/cloudquery/plugins/source/stripe/resources/services/invoices"
+)
+
+func TestInvoicesInvoiceItem(t *testing.T) {
+	tbl := invoices.InvoiceItems()
+	for i, c := range tbl.Columns {
+		if c.Name == "plan" {
+			tbl.Columns[i].IgnoreInTests = true
```
disq, line 14: We need to do this because the mock server doesn't return the column even though the fixture is populated correctly.

plugins/source/stripe/resources/services/products/products_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"github.com/cloudquery/cloudquery/plugins/source/stripe/client"
+	"github.com/cloudquery/cloudquery/plugins/source/stripe/resources/services/products"
+)
+
+func TestProductsProduct(t *testing.T) {
+	tbl := products.Products()
+	for i, c := range tbl.Columns {
+		if c.Name == "attributes" || c.Name == "deactivate_on" {
+			tbl.Columns[i].IgnoreInTests = true
```
disq, line 14: We need to do this because the mock server doesn't return the column even though the fixture is populated correctly.

plugins/source/stripe/Makefile
```diff
@@ -0,0 +1,29 @@
+# Test unit
+.PHONY: test
+test: gen-fixtures
+	wget -nc https://raw.githubusercontent.com/stripe/openapi/master/openapi/spec3.json -P ./resources/testdata
```
disq, line 4: This is a large spec so we don't include it in the repo.

yevgenypats (changes requested):
Looks good! Can you try migrate this PR to use the new "no-codegen" approach with this PR - https://github.com/cloudquery/plugin-sdk/pull/564

plugins/source/stripe/codegen/fixtures/main.go
```diff
@@ -25,7 +26,15 @@ func main() {
 		Resources: make(map[string]any),
 	}
 
-	extras := []any{
+	dataStructs := []any{
```
disq, line 29: These are required to generate the fixture file to use with stripe's mock server.

plugins/source/stripe/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# Changelog
```
erezrokah, line 1: We can remove this file (our automation generates it)

plugins/source/stripe/go.mod
```diff
@@ -0,0 +1,49 @@
+module github.com/cloudquery/cloudquery/plugins/source/stripe
+
+go 1.19
+
+require (
+	github.com/cloudquery/plugin-sdk v1.17.3-0.20230104090631-4b15afb2acc1
```
disq, line 6: Need to update this to a SDK release with the 🤖🤖🤖 package.

plugins/source/stripe/codegen/recipes/account.go
```diff
@@ -0,0 +1,14 @@
+package recipes
+
+import (
+	"github.com/stripe/stripe-go/v74"
+)
+
+func init() {
```
hermanschaaf, line 7: The use of `init` to append to a global `AllResources` is quite... creative? 😅 I don't know how I feel about it, but I do admit that it at least avoids needing to remember to add each resource to a slice in `codegen/main.go`, so maybe it's better than what we have. Possibly something to debate a bit, because I think we should stick to one convention or the other, at least

plugins/source/stripe/docs/tables/stripe_invoice_items.md
```diff
@@ -0,0 +1,38 @@
+# Table: stripe_invoice_items
+
+https://stripe.com/docs/api/invoiceitem
```
hermanschaaf, line 3: This link gives a 404

plugins/source/stripe/resources/services/accounts/accounts.go
```diff
@@ -0,0 +1,41 @@
+// Code generated by codegen; DO NOT EDIT.
+
```
hermanschaaf, line 2: Might be controversial, but with how little code is in here now, and how few tables there are, I don't think we're winning much by making it generated (unless we can generate 50+ more). Could we remove the codegen for this plugin entirely, to make it easier to understand for outside contributors?

plugins/source/stripe/resources/services/invoices/invoice_items_test.go
hermanschaaf: Why isn't the column getting returned - isn't that something we control?

website/pages/docs/plugins/sources/stripe/overview.md
```diff
@@ -0,0 +1,41 @@
...
+
+This is the (nested) spec used by the Stripe source plugin:
+
+- `api_key` (string, required):
+  Your secret API key from the Stripe Dashboard.
+
+- `max_retries` (integer, optional. Default: 2):
+  Number of retries if a request was rate limited.
+
+## Example
```
hermanschaaf, line 24: Nit: This example should probably move to the top of the page, as that's the thing you need first before you need further details.

hermanschaaf (changes requested):
Looks good, just a couple of comments--mainly I think now that it's using transformers, we should see whether we can take more codegen out of the development loop to make it easier to contribute to.

plugins/source/stripe/docs/tables/stripe_invoice_items.md
disq: Fixed, should've been plural

plugins/source/stripe/resources/services/accounts/accounts.go
yevgenypats: Agree

disq: We can generate 50+ more because go-stripe is pretty consistent (except for `stripe.InvoiceItem` being `invoiceitem`, but that's something to do with the API itself) and with Stripe since we don't have a base installation to actually test everything going the codegen way seems superior. Any outside contributor can add 2 lines to a recipe file and run `make gen` to automatically implement any root level resource including mock fixtures. We don't need to treat everything as if they are nails, even if we have a hammer at our disposal.

website/pages/docs/plugins/sources/stripe/overview.md
disq: Good nit, fixed

plugins/source/stripe/resources/services/invoices/invoice_items_test.go
disq: No - mock server is implemented in the https://github.com/stripe/stripe-mock package and we're merely utilizing it to handle our requests, rather than faking our own server.

plugins/source/stripe/resources/services/accounts/accounts.go
yevgenypats: If it's all easily generatable why not generate all the tables right now? I think the best practice should be generate everything if it's easily generatable, if not then just do a few important ones and don't use codegen.

disq: Using half codegen looks like an interesting compromise where we generate the resolvers and other table defs, but don't make it more complicated by adding all the codegen.Table* logic.

plugins/source/stripe/go.mod
```diff
@@ -0,0 +1,46 @@
+module github.com/cloudquery/cloudquery/plugins/source/stripe
+
+go 1.19
+
+require (
+	github.com/cloudquery/plugin-sdk v1.21.1-0.20230105235501-d97b5c9659fd
```
disq, line 6: needs an SDK release from https://github.com/cloudquery/plugin-sdk/pull/580

plugins/source/stripe/resources/services/accounts/accounts.go
```diff
@@ -0,0 +1,81 @@
...
+		},
+		IsIncremental: true,
+
+		Relations: []*schema.Table{
+			Capabilities(),
+		},
+	}
+}
+
+func fetchAccounts(tableName string) schema.TableResolver {
```
yevgenypats, line 48: Im not quite sure I understood why this is a function that returns a function instead of just a function.

yevgenypats (changes requested):
Just first round of review. Im not sure I understand the structure of the fetchers or the reason for that?

plugins/source/stripe/resources/services/accounts/accounts.go
disq: This is to be able to use the name of the table, to use in state save/load operations. Could also make it a const in the fn as well I guess, since all is generated anyway.

yevgenypats: I see. yeah let's use `const` please, I think it will be more clear that way.

disq: Done

yevgenypats (changes requested):
Few more comments:

Can we do a test via `getAccount` on init (as I think an API key should always have permissions to that api) and then return an error as otherwise the CLI just shows that all succesfuly finished with some errors.

Also, something Im not able to understand is how I get a different result everytime I run a sync:

```
Starting sync for: stripe -> [postgresql]
Sync completed successfully. Resources: 159, Errors: 25, Panics: 0, Time: 17s
yevgenyp@yevgenys-mbp cloudquery % 
yevgenyp@yevgenys-mbp cloudquery % cloudquery sync ./cli/examples/stripe.yml
Loading spec(s) from ./cli/examples/stripe.yml
Source stripe will migrate and sync 61 tables.
Starting migration for: stripe -> [postgresql]
Starting sync for: stripe -> [postgresql]
Sync completed successfully. Resources: 123, Errors: 29, Panics: 0, Time: 17s
``` 

and errors+resources also don't match ? 🤔

yevgenypats (approved):
Looks good to go. I would change the default limits to `20`, and `90` to be on the safe side.

hermanschaaf:
Not a blocker, but how hard would it be to add support for the Events endpoint? https://stripe.com/docs/api/events/object Seems like it would be quite useful / important

disq:
> Not a blocker, but how hard would it be to add support for the Events endpoint? [stripe.com/docs/api/events/object](https://stripe.com/docs/api/events/object) Seems like it would be quite useful / important

I didn't because it would end up pushing too much data around. Now that we have the state thing we can add it.

hermanschaaf:
@disq Nice! It's also cool because Stripe only keeps the data for 30 days, so you need a solution like ours if you want it for longer.

disq:
https://github.com/cloudquery/cloudquery/pull/6776

PR #6155, opened 2022-12-30
fix(cli): Update progress bar on secondly interval

hermanschaaf:
A small update to the CLI progress bar so that it doesn't (appear to) hang for long periods when it takes time for new resources to be found. This change updates the progress bar every second.

##  📹 Demo 

https://user-images.githubusercontent.com/1121616/210093868-d6881c2b-dca3-4046-8c2b-c8aa2fd50f66.mov


yevgenypats (approved):
🪄  Nice!

PR #6162, opened 2022-12-31
feat: Add Neo4j destination

yevgenypats:
This adds Neo4j destination and can be useful for various networking, security audits, blue-teams, red-teams use-cases.

Similar popular projects that are bind to a specific neo4j db are:
- https://github.com/lyft/cartography
- https://github.com/duo-labs/cloudmapper

As a follow-up task it will be interesting to do a blog/deep-dive on how to add enrichments to that data so you can traverse easily resources and visualize network connections as the plugin is just doing the first step of putting the data there (It works the same way in the above projects where there is an enrichment step that adds the connections between resources).


.github/workflows/dest_neo4j.yml
```diff
@@ -0,0 +1,109 @@
...
+        uses: cloudquery/golangci-lint-action@master
+        with:
+          version: v1.50.1
+          working-directory: plugins/destination/neo4j
+          args: "--config ../../.golangci.yml"
+      - name: Get dependencies
+        run: go get -t -d ./...
+      - name: Build
+        run: go build .
+      - name: Test PG
```
candiduslynx, line 64: -
```suggestion
      - name: Test Neo4j
```

.github/workflows/dest_neo4j.yml
```diff
@@ -0,0 +1,109 @@
...
+        run: make test
+      # cockroachdb doesn't support services right now
```
candiduslynx, line 66: Remove cocroach comments

plugins/destination/neo4j/client/read.go
```diff
@@ -0,0 +1,122 @@
...
+			results = append(results, &r)
+		case schema.TypeJSON:
+			r := node.Props[col.Name].(string)
+			results = append(results, &r)
+		case schema.TypeUUIDArray:
+			r := make([]string, len(node.Props[col.Name].([]any)))
+			for i, v := range node.Props[col.Name].([]any) {
+				r[i] = v.(string)
+			}
+			results = append(results, r)
```
candiduslynx, line 56: (Nit): append can be done once after switch

candiduslynx (approved):
Only non-blocking comments

PR #6223, opened 2023-01-01
test(azblob): Ensure container exists before tests

erezrokah:
#### Summary

Azure playground account resources get nuked, so we can't rely on them existing. [I re-created the storage account](https://portal.azure.com/#@ypcloudquery.onmicrosoft.com/resource/subscriptions/21e0d134-5da8-4754-8587-c58425d93de5/resourceGroups/cloudquery-ci/providers/Microsoft.Storage/storageAccounts/cqdestinationazblob/overview) and excluded it via https://github.com/cloudquery/deployments/pull/48/files (internal link).

This is still missing the required permissions for the storage account, as I'm not sure which user I should set them on

### DIFF
--- a/plugins/destination/azblob/client/client_test.go
+++ b/plugins/destination/azblob/client/client_test.go
@@ -1,20 +1,64 @@
 package client
 
 import (
+	"context"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


erezrokah:
I'll close this and open from a non forked repo so the PR has access to secrets

PR #6226, opened 2023-01-01
fix(aws): Remove account validation

erezrokah:
#### Summary

Leaving this open for tomorrow. A user on Discord reported that a previously working configuration stopped working on `v9.0.1`.

Relevant configuration:

```yaml
  kind: source
  spec:
    name: aws
    path: cloudquery/aws
    version: "v9.0.1"
    destinations: ["postgresql"]
    tables: ["*"]
    skip_tables:
** snipped for copy/paste **
    spec:
      regions:
        - us-east-1
        - us-east-2
        - us-west-1
        - us-west-2
      accounts:
        - id: "services"
          role_arn: "arn:aws:iam::XXXXXXXXXXXX:role/security/cloudquery-ro-role"
          role_session_name: "cloudquery_inventory"
        - id: "prod"
          role_arn: "arn:aws:iam::XXXXXXXXXXXX:role/security/cloudquery-ro-role"
          role_session_name: "cloudquery_inventory"
```

Error:

```
spec validation failed: invalid accounts: invalid account id: services (should be 12 digits)"
```

It seems users should be able to use the alias of an account in the configuration. Keeping open for discussion and so that users can subscribe to get updates on this.
Also not sure about the `skip_member_accounts` validation (I kept that one).


hermanschaaf (changes requested):
@erezrokah This looks good, thanks! I'd say let's remove the account validation function entirely (so for OU accounts as well). It wasn't necessary. We can follow up to make sure that both account numbers and names are accepted by that as well, and if they aren't we should make it work the same way as the top-level list in any case

hermanschaaf:
Oh and just to be clear, let's keep the OU name validation. I believe that is still valid.

erezrokah:
> Oh and just to be clear, let's keep the OU name validation. I believe that is still valid.

👍 Done in https://github.com/cloudquery/cloudquery/pull/6226/commits/56232887db3db39de3e2ae634dec009e5da92976.

Also added an account alias to the tests

hermanschaaf (approved):

Issue #6116, opened 2022-12-29
chore(ci): Fix broken links CI check


erezrokah:
The broken links CI check fails when a build is ignored on Vercel.
https://github.com/cloudquery/cloudquery/pull/5980 should have fixed it but it didn't.

PR #6227, opened 2023-01-01
chore: Ignore merge commits when diffing Website builds

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6116

Using `git diff` doesn't work well if someone merges the PR branch with `main` as then we detect changes on the merge commit. Hence we need to use `git log` and skip the merge commits.

Once this is working we should update the settings on Vercel to use `website/scripts/diff.sh`


erezrokah:
Once this is merged I'll update the `git ignore` command on Vercel to use the same diffing script.

hermanschaaf (approved):

PR #6238, opened 2023-01-02
chore(website): Deprecate CSV

yevgenypats:
Deprecate CSV destination plugin in the docs.

### DIFF
--- a/website/pages/docs/plugins/destinations/_meta.json
+++ b/website/pages/docs/plugins/destinations/_meta.json
@@ -2,7 +2,7 @@
   "overview": "Overview",
   "azblob": "Azure Blob Storage",
   "bigquery": "BigQuery",
-  "csv": "CSV",
+  "csv": "CSV (Deprecated)",
   "file": "File",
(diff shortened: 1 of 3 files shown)


erezrokah (approved):
Looks good but maybe we can move this plugin to the end of the list in `_meta.json` and also change the sorting function here https://github.com/cloudquery/cloudquery/blob/f4a6bb8b79ebc47138c88880a31df5e4f4135b73/website/components/PluginsTable.tsx#L55 to list deprecated plugins last?

yevgenypats:
> Looks good but maybe we can move this plugin to the end of the list in `_meta.json` and also change the sorting function here
> 
> https://github.com/cloudquery/cloudquery/blob/f4a6bb8b79ebc47138c88880a31df5e4f4135b73/website/components/PluginsTable.tsx#L55
> 
> to list deprecated plugins last?

I feel it is ok both ways. let's keep the sorting the same for now as don't have a strong opinion either way.

PR #6239, opened 2023-01-02
feat(website): Show deprecated plugins last

erezrokah:
#### Summary

This ensures deprecated plugins are last on the list of plugins


erezrokah:
Ah, just saw https://github.com/cloudquery/cloudquery/pull/6238#issuecomment-1368738752. Let's close this

PR #6249, opened 2023-01-02
chore(test): Use `-race` test flag

### LINKED ISSUE #4243: Heroku plugin: data race in custom pagination routine
`go test -race ./...` currently returns some errors when run against the Heroku plugin. I'm not sure if this can affect anything in real-world cases--it might be due to the way we run tests in parallel.

I won't paste the full output here, but here is part:

```
WARNING: DATA RACE
Read at 0x00c0004c2200 by goroutine 733:
  net/http.(*Transport).queueForDial()
      /opt/homebrew/Cellar/go/1.19.1/libexec/src/net/http/transport.go:1418 +0x60
  net/http.(*Transport).getConn()
      /opt/homebrew/Cellar/go/1.19.1/libexec/src/net/http/transport.go:1373 +0x610
  net/http.(*Transport).roundTrip()
      /opt/homebrew/Cellar/go/1.19.1/libexec/src/net/http/transport.go:582 +0xa0c
  net/http.(*Transport).RoundTrip()
      /opt/homebrew/Cellar/go/1.19.1/libexec/src/net/http/roundtrip.go:17 +0x30
  github.com/cloudquery/cloudquery/plugins/source/heroku/client.Paginator.RoundTrip()
      /Users/hermanschaaf/code/cloudquery/cloudquery/plugins/source/heroku/client/pagination.go:39 +0x6c
  github.com/cloudquery/cloudquery/plugins/source/heroku/client.(*Paginator).RoundTrip()
      <autogenerated>:1 +0x4c

Previous write at 0x00c0004c2200 by goroutine 725:
  net/http/httptest.(*Server).Start()
      /opt/homebrew/Cellar/go/1.19.1/libexec/src/net/http/httptest/server.go:130 +0x68
  net/http/httptest.NewServer()
      /opt/homebrew/Cellar/go/1.19.1/libexec/src/net/http/httptest/server.go:107 +0x1ac
  github.com/cloudquery/cloudquery/plugins/source/heroku/resources/services.createAddOnServices()
      /Users/hermanschaaf/code/cloudquery/cloudquery/plugins/source/heroku/resources/services/add_on_service_mock_test.go:34 +0x1e4
  github.com/cloudquery/cloudquery/plugins/source/heroku/client.MockTestHelper.func1()
      /Users/hermanschaaf/code/cloudquery/cloudquery/plugins/source/heroku/client/testing.go:30 +0x38
  github.com/cloudquery/plugin-sdk/plugins.(*SourcePlugin).Sync()
      /Users/hermanschaaf/go/pkg/mod/github.com/cloudquery/plugin-sdk@v1.3.2/plugins/source.go:159 +0x34c
  github.com/cloudquery/plugin-sdk/plugins.TestSourcePluginSync.func1()
      /Users/hermanschaaf/go/pkg/mod/github.com/cloudquery/plugin-sdk@v1.3.2/plugins/source_testing.go:29 +0xec

Goroutine 733 (running) created at:
  github.com/cloudquery/plugin-sdk/plugins.(*SourcePlugin).resolveTableDfs()
      /Users/hermanschaaf/go/pkg/mod/github.com/cloudquery/plugin-sdk@v1.3.2/plugins/source_scheduler_dfs.go:74 +0x678
  github.com/cloudquery/plugin-sdk/plugins.(*SourcePlugin).syncDfs.func1()
      /Users/hermanschaaf/go/pkg/mod/github.com/cloudquery/plugin-sdk@v1.3.2/plugins/source_scheduler_dfs.go:61 +0xfc
```

It looks like it is caused by something in the custom `client.Paginator.RoundTrip()` implementation that we had to do as a workaround for the Heroku Go SDK not supporting pagination out of the box. I remember seeing some warnings on the [`RoundTripper` interface](https://pkg.go.dev/net/http#RoundTripper) that request attributes should be read in a separate goroutine, but in our case this was not possible, so this might be the cause of the data race.

One option here might be contributing pagination to the upstream Heroku SDK (or forking it), as it seems like it had some recent activity.

candiduslynx:
closes #4243

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test -race ./...
+        run: make test
       - name: gen-docs
         if: matrix.os == 'ubuntu-latest'
         run: make gen-docs
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
       - name: gen-docs
         if: matrix.os == 'ubuntu-latest'
         run: make gen-docs
--- a/.github/workflows/dest_test.yml
+++ b/.github/workflows/dest_test.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test -race ./...
+        run: make test
   validate-release:
     timeout-minutes: 30
     runs-on: ubuntu-latest
--- a/.github/workflows/dest_test.yml
+++ b/.github/workflows/dest_test.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
   validate-release:
     timeout-minutes: 30
     runs-on: ubuntu-latest
--- a/.github/workflows/source_azure.yml
+++ b/.github/workflows/source_azure.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
       - name: gen
         if: github.event_name == 'pull_request'
         run: make gen
--- a/.github/workflows/source_azure.yml
+++ b/.github/workflows/source_azure.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test -race ./...
+        run: make test
       - name: gen
         if: github.event_name == 'pull_request'
         run: make gen
--- a/.github/workflows/source_heroku.yml
+++ b/.github/workflows/source_heroku.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
       - name: gen
         if: github.event_name == 'pull_request'
         run: make gen
--- a/.github/workflows/source_heroku.yml
+++ b/.github/workflows/source_heroku.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test -race ./...
+        run: make test
       - name: gen
         if: github.event_name == 'pull_request'
         run: make gen
--- a/contributing/development_environment.md
+++ b/contributing/development_environment.md
@@ -46,7 +46,7 @@ Note that plugin logs will be output to the plugin process terminal.
 To run tests all unit tests for a plugin, inside the plugin directory run:
 
 ```shell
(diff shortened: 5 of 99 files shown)


contributing/development_environment.md
```diff
@@ -46,7 +46,7 @@ Note that plugin logs will be output to the plugin process terminal.
 To run tests all unit tests for a plugin, inside the plugin directory run:
 
 ```shell
-make test-unit  # This runs go test ./...
+make test-unit  # This runs go test -race ./...
```
hermanschaaf, line 49: Actually I think this command should be `make test` now

.github/workflows/cli.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
```
erezrokah, line 67:
```suggestion
        run: make test
```

.github/workflows/dest_test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
```
erezrokah, line 44:
```suggestion
        run: make test
```

.github/workflows/source_azure.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
```
erezrokah, line 64:
```suggestion
        run: make test
```

.github/workflows/source_heroku.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: go test ./...
+        run: go test -race ./...
```
erezrokah, line 44:
```suggestion
        run: make test
```

erezrokah (changes requested):
Thanks for the PR @candiduslynx, should we align all workflows to use `make test`? We can add the missing make file targets.

That way we don't need to duplicate the flags (if we change them)

erezrokah:
Also, not sure if this is a bug fix in Heroku or not. If it's a bug fix we should do it in a separate PR with a type `fix:`. We need a separate PR otherwise it will generate releases for all plugins.

hermanschaaf (approved):
LGTM, let's just keep an eye on build times after this

hermanschaaf:
> Also, not sure if this is a bug fix in Heroku or not. If it's a bug fix we should do it in a separate PR with a type fix:. We need a separate PR otherwise it will generate releases for all plugins.

@erezrokah Yeah I also thought about this, but it seems to only affect the testing of Heroku, so we should be okay. Would be good if @candiduslynx can just confirm that

contributing/development_environment.md
candiduslynx, line 49:
```suggestion
make test  # This runs go test -race ./...
```

erezrokah (approved):

PR #6253, opened 2023-01-02
doc(website): Update more how to guides

jsonpr:
Moved 5 of the top blog pages that are how to guides to how to section with redirects from old blog links. I did not validate all of the content, some should be verified for correctness to ensure they're not out of date given recent cloud service provider (CSP) and other updates.


yevgenypats:
Closing per slack discussion. Keeping How-tos only for CQ related blogs/how-tos

PR #6262, opened 2023-01-03
fix: Return error if read is called when `no_rotate` is false

erezrokah:
#### Summary

We should return an error when trying to read if `no_rotate` is `false` as we can't read a randomly generated name


erezrokah:
Need to open this not from a fork

Issue #6248, opened 2023-01-02
Bug(postgres): When recreating primary key(s) constraint, `not null` is not removed from old PK(s) 


erezrokah:
Related to https://github.com/cloudquery/cloudquery/issues/6246

When we recreate primary keys during migration, we set new primary keys columns to `not null`:
https://github.com/cloudquery/cloudquery/blob/efca4a7c8092f306a41583e2d5debf80875b64ae/plugins/destination/postgresql/client/migrate.go#L128

However we don't remove `not null` enforcement on previous primary keys.
Since we also don't delete renamed columns, when we try to insert new data to the table, the old column is populated with `NULL` and the insert fails.

### Proposed solution(s)

- Solution 1 - Look for all `not null` columns and remove the enforcement when recreating primary keys
- Solution 2 - If after the migration the table has `not null` columns that are not part of the new schema, fail the migration

candiduslynx:
Solution 3: if the new schema isn't the subset of the schema after migration (there might be extra columns, but the constraints and PK are the same) – fail the migration

erezrokah:
> there might be extra columns, but the constraints and PK are the same

Aren't extra columns OK? The sync will work and add the missing data no?

candiduslynx:
Yes, extra columns in the resulting schema ore OK.
Basically, we want to check that the resulting schema contains all and every:
* column from the new schema
* constraint from the new schema
* the PK is exactly the same (including constraints)

erezrokah:
I might be misunderstanding, but we're already creating any missing column/constraint/primary key here:
https://github.com/cloudquery/cloudquery/blob/efca4a7c8092f306a41583e2d5debf80875b64ae/plugins/destination/postgresql/client/migrate.go#L81
so I'm not sure how something will be missing.

The issue describes a use cases where the auto migration succeeds, but the DB is in a state that any future sync fails.

candiduslynx:
Basically, my proposal is extension of solution 2 with more general rules. However, we can just implement the 2nd solution for a quick fix.

PR #6282, opened 2023-01-03
fix: Error if after the migration there are `not null` columns that are not part of the new schema

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6248 (implements ~~first~~ second solution from the issue)


plugins/destination/postgresql/client/client_test.go
```diff
@@ -24,3 +32,44 @@ func TestPgPlugin(t *testing.T) {
 		},
 		destination.PluginTestSuiteTests{})
 }
+
+func TestPgPluginPrimaryKeyRename(t *testing.T) {
+	tableName := fmt.Sprintf("cq_test_pk_rename_%d", rand.Intn(100))
```
erezrokah, line 37: This is so we don't need to drop the table before each test

disq: Might work or not work depending on random seed :)

erezrokah: Yeah, this is mostly for local dev, the CI always starts with a new DB

yevgenypats:
I would be cautions in dropping things as a data-integration tool. users can (and do) create their own constraints, views and so on.

erezrokah:
> I would be cautions in dropping things as a data-integration tool. users can (and do) create their own constraints, views and so on.

That's my concern too. We're dropping the constraint on columns that were primary keys before the migration, and are not after the migration.
Instead of dropping the constraint we can exit with an error guiding the user what to do.

The main issue is that the migration succeeds, but puts the DB in a state that any following sync fails. Exiting with an error should also work.

yevgenypats:
> > I would be cautions in dropping things as a data-integration tool. users can (and do) create their own constraints, views and so on.
> 
> That's my concern too. We're dropping the constraint on columns that were primary keys before the migration, and are not after the migration. Instead of dropping the constraint we can exit with an error guiding the user what to do.
> 
> The main issue is that the migration succeeds, but puts the DB in a state that any following sync fails. Exiting with an error should also work.

Yes error would be a better/safer solution.

candiduslynx (approved):

erezrokah:
> Yes error would be a better/safer solution.

Done in https://github.com/cloudquery/cloudquery/pull/6282/commits/14790b6f71ad3552a58d83c500406c392ed154d8

disq:
Could also generate `drop not null` statements for the user to run.

disq:
> Ah, I tried that at first. But the problem they only work after we drop the primary keys. And if we drop the primary keys we won't be able to detect the error the next time the migration runs

Need an `--allow-schema-change` option maybe? Which will auto-fix things by dropping PKs and remove constraints (and drop unused columns...)

For the user message you could also suggest the drop pk commands as well as drop not nulls.

erezrokah:
> For the user message you could also suggest the drop pk commands as well as drop not nulls.

I felt that would be too verbose, but let me try that

erezrokah:
> For the user message you could also suggest the drop pk commands as well as drop not nulls.

But then again, it brings the question - do we tell the user to drop the column? Drop only the constraint? Rename it?
The use case here is a primary key rename. We can't detect the rename just by looking at the DB

disq:
> We can't detect the rename just by looking at the DB

In the olden times we did: I remember writing some code to compare a `schema.Table` with an actual table and to report what's changed and to generate sql migrations for the user. It detected column renames by [checking if the types were equal and the names were similar](https://github.com/cloudquery/cq-provider-sdk/blob/v0.9.5/migration/table.go#L213).

Here we'll just suggest the user to drop any unused column, including renames. We'll take care of the rest in the next migration (by adding what's missing). We could also tell the user what the next steps are (from our migrations POV, ie. what we're going to do on the next run if the columns are dropped), so they can decide for themselves if it's a rename and take necessary action? Too much? Maybe.

erezrokah:
> In the olden times we did: I remember writing some code to compare a `schema.Table` with an actual table and to report what's changed and to generate sql migrations for the user. It detected column renames by [checking if the types were equal and the names were similar](https://github.com/cloudquery/cq-provider-sdk/blob/v0.9.5/migration/table.go#L213).
> 
> Here we'll just suggest the user to drop any unused column, including renames. We'll take care of the rest in the next migration (by adding what's missing). We could also tell the user what the next steps are (from our migrations POV, ie. what we're going to do on the next run if the columns are dropped), so they can decide for themselves if it's a rename and take necessary action? Too much? Maybe.

I could go either way. I'd rather give a generic error that's 100% correct, than have something wrong in our heuristics and provide the wrong recommendation.

Maybe dropping the primary key and the `not null` constraints is the safest option to recommend. I'll give it a try

erezrokah:
image

OK, so I think https://github.com/cloudquery/cloudquery/pull/6282/commits/371be8e4533796fc8e09bf0b32ec76dd4addf744 should work (see image)

disq (approved):

PR #6336, opened 2023-01-04
feat: Hacker News source plugin

hermanschaaf:
This adds a Hacker News source plugin that is able to download all items from the Hacker News API. It can be useful, but is mainly meant to be for instructional purposes, as this is the first example of a plugin making use of the state backend for incremental syncing being introduced in https://github.com/cloudquery/plugin-sdk/pull/569 (this PR depends on that one).


cli/go.mod
```diff
@@ -40,3 +40,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudquery/plugin-sdk => ../../plugin-sdk
```
disq, line 44: You could `go get` an untagged 'version' of the SDK (using the `@commithash` syntax) and remove this line

plugins/source/hackernews/client/errors.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+		return err
+	}
+	return nil
+}
+
+func isRetryable(err error) bool {
+	var httpErr hackernews.HTTPError
+	if errors.As(err, &httpErr) {
+		return httpErr.Code >= 500 || httpErr.Code == 429
```
disq, line 36:
```suggestion
		return httpErr.Code >= http.StatusInternalServerError || httpErr.Code == StatusTooManyRequests
```

plugins/source/hackernews/client/services/backend.go
```diff
@@ -0,0 +1,10 @@
+package services
+
+import "context"
+
+//go:generate mockgen -package=mocks -destination=../mocks/backend.go -source=backend.go BackendClient
+type Backend interface {
+	Set(ctx context.Context, table, key, value string) error
```
disq, line 7: You should embed the interface (`backend.Backend`) from the plugin SDK here

plugins/source/hackernews/codegen/main.go
```diff
@@ -0,0 +1,135 @@
...
+	err := services.Generate()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var resources []*recipes.Resource
+	resources = append(resources, recipes.ItemResources()...)
+
+	for _, r := range resources {
+		r.Infer()
```
disq, line 36: Since you're not doing any parent-child-things you could move this line to the loop below.

plugins/source/hackernews/resources/services/items/items_fetch.go
```diff
@@ -0,0 +1,158 @@
...
+					if fetchErr != nil {
+						return fetchErr
+					}
+					success <- itemID
+				}
+			}
+		})
+	}
+
+	// We use a min heap to keep track of the items that have finished fetching,
```
disq, line 85: You could try to put this in another file, grouped under worker/manager 'structs' and divided into multiple methods.

plugins/source/hackernews/client/services/backend.go
hermanschaaf: Ah yeah, good idea

cli/go.mod
hermanschaaf: I'd rather not accidentally merge that in :)

plugins/source/hackernews/client/errors.go
hermanschaaf: Thanks!

plugins/source/hackernews/codegen/main.go
hermanschaaf: Removed the codegen directory completely

website/pages/docs/plugins/sources/hackernews/overview.md
```diff
@@ -0,0 +1,33 @@
...
+
+The Hacker News Source plugin for CloudQuery extracts configuration from the [Hacker News API](https://github.com/HackerNews/API) and loads it into any supported CloudQuery destination (e.g. PostgreSQL).
+
+It can be used for real applications, but is mainly intended to serve as a reference for how to write an incremental CloudQuery Source plugin. 
+
+## Configuration
+
+The following configuration syncs from Hacker News to a Postgres destination. The (top level) source spec section is described in the [Source Spec Reference](https://www.cloudquery.io/docs/reference/source-spec). The `postgresql` destination is not shown here and needs to be separately defined.
+
+```yaml
```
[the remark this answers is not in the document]
hermanschaaf: Thanks, fixed!

plugins/source/hackernews/resources/services/items/items_fetch.go
disq: I'm saving this response to use in future code reviews on my own PRs.

plugins/source/hackernews/resources/services/items/items_fetch.go
```diff
@@ -0,0 +1,158 @@
...
+	work := make(chan int)
+	maxID, err := c.HackerNews.MaxItemID(ctx)
+	if err != nil {
+		return err
+	}
+	c.Logger().Info().Msg("Found max ID, reading up to " + strconv.Itoa(maxID))
+
+	// spin up worker goroutines that will fetch items concurrently by reading ids from
+	// the work channel
+	workers := c.Spec.ItemConcurrency
```
erezrokah, line 60: I understand this is the most efficient to do this. Another way that might simplify the implementation but is less efficient is to use batches, i.e.

- Start 100 (or whatever) Go routines to get the first batch of items
- If all went well update the cursor and move to the next 100 (or whatever is left)
- If the batch failed set the cursor based on the section of the batch that succeeded

With ⬆️ I don't think we need a heap, only to track the results of the batch

Assuming that API calls take about the same time I'm not sure the performance difference will be that significant

erezrokah (approved):
Looks good, added a non blocking comment.

plugins/source/hackernews/resources/services/items/items_fetch.go
hermanschaaf: Haha 😂  The thing is, there's a lot going on here. Breaking it up into smaller functions may make each function simpler, but the overall system harder to understand. I'll see what I can do though

hermanschaaf: I did a refactor now but I'm not sure it's better, and not sure it's still correct either. Haven't pushed it up yet, I'll need to verify it with fresh eyes tomorrow

plugins/source/hackernews/resources/plugin/plugin.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func Plugin() *source.Plugin {
+	tables := []*schema.Table{
+		items.Items(),
+	}
+	return source.NewPlugin(
+		"hackernews",
+		Version,
+		tables,
```
disq, line 21:
```suggestion
		[]*schema.Table{
			items.Items(),
		},
```
Why not just make it inline?

hermanschaaf: I considered it, but decided it's easier to tell someone in CONTRIBUTING.md to "add new tables to the `tables` variable in plugin.go" than to tell them "add new tables to the `[]*schema.Table` slice (third parameter of `NewPlugin`) in plugin.go". That was my reasoning for not inlining it.

hermanschaaf:
@yevgenypats @erezrokah @disq This is ready for another review :bow: Still depends on https://github.com/cloudquery/plugin-sdk/pull/569 which is why the tests are currently not passing, but I greatly simplified the resolver code using @erezrokah's suggestion to do it in batches. The resulting code is slightly less efficient, but still close enough that I think it's better to keep it this simple.

plugins/source/hackernews/resources/services/items/items_fetch.go
```diff
@@ -0,0 +1,114 @@
...
+	c.Logger().Info().Msg("Found max ID, reading up to " + strconv.Itoa(maxID))
+
+	// Fetch items in batches of (max) 1000.
+	// This is not necessarily the most efficient way of doing it, but this code
+	// is meant to be for instructional purposes as an example of updating cursors,
+	// so we're keeping the logic relatively simple.
+	// The important thing is that the state backend does not ensure that the cursor
+	// is strictly increasing--it is the responsibility of the resolver to ensure this.
+	for cursor < maxID {
+		endID := cursor + 1000
```
erezrokah, line 66:
```suggestion
		endID := cursor + c.Spec.ItemConcurrency
```

If we do this we won't have to `g.SetLimit(c.Spec.ItemConcurrency)`. Also we keep the batches as small as possible so lower chances of a batch to fail

erezrokah (approved):
Looks great, added a comment for your consideration (non blocking)

plugins/source/hackernews/resources/services/items/items_fetch.go
hermanschaaf: I tried this, but it has a pretty big impact on performance, even after playing with different values for `ItemConcurrency` (on my machine: ~500 items/s falls to ~300 items/s). I think let's leave it at 1000 for now

yevgenypats (approved):
Looks great! Good idea on this plugin for cursor testing and I think can be a great post for Show HN :) 

Maybe move https://github.com/hermanschaaf/hackernews to `cloudquery/hackernews` so anyone in the team can contribute easily if needed?

hermanschaaf:
@yevgenypats Sure, I can migrate it to the CloudQuery org, just wasn't sure whether we want to be the maintainers of it? (Personally I'm fine with being the maintainter 😃 )

PR #6337, opened 2023-01-04
feat: Move AWS to use built-in transformations (no codegen)

yevgenypats:
Similar to https://github.com/cloudquery/cloudquery/pull/6276 moving transformation to be part of our SDK https://github.com/cloudquery/plugin-sdk/pull/564 

This is related to https://github.com/cloudquery/plugin-sdk/pull/573 

Few notes on column changes:

most of changes are just additional columns and this should be fine as we did too much transformation. we want to keep the code simple, focus on extraction and provide and easy way of doing "dumb" one level transformation via reflection. The additional column are because we have in some places a duplicate where we add our own `arn` so it can be consistent but this is our decision and we still want to be as close as possible to the API so some duplicates is fine (and it makes the code simple and consistent).

breaking changes in `aws_docdb_pending_maintenance_actions ` those apparently never worked and those columns where always empty (they didn't have mock tests so we didn't catch those)


plugins/source/aws/resources/services/sns/subscriptions.go
```diff
@@ -33,6 +33,26 @@ func Subscriptions() *schema.Table {
 					PrimaryKey: true,
 				},
 			},
+			{
+				Name:     "delivery_policy",
```
disq, line 37: Instead of adding these, should unwrap fields in Transformer

plugins/source/aws/resources/services/sns/topics.go
```diff
@@ -38,6 +38,21 @@ func Topics() *schema.Table {
 				Type:     schema.TypeJSON,
 				Resolver: resolveSnsTopicTags,
 			},
+			{
+				Name:     "delivery_policy",
```
disq, line 42: Instead of adding these, should unwrap fields in Transformer

plugins/source/aws/resources/services/sqs/queues.go
```diff
@@ -33,6 +33,21 @@ func Queues() *schema.Table {
 					PrimaryKey: true,
 				},
 			},
+			{
```
disq, line 36: Instead of adding these, should unwrap fields in Transformer

plugins/source/aws/resources/services/sns/subscriptions.go
yevgenypats: why unwrap? those were `json` before. I think this is a good way to override just by specifying the column directly.

disq: oh it's the type and not the missing fields? ok I misunderstood.

hermanschaaf (approved):

hermanschaaf:
Bringing an offline conversation here: I think we should keep the codegen that generates service interfaces for mocks, so those don't need to be written by hand. However, we can now make it part of the `make gen-mocks` step, so it's not something you're going to need to run often (only when adding a completely new service).

We should also make sure the Contributing doc is up-to-date, because I think it still references codegen.

PR #6338, opened 2023-01-04
feat(gcp-resources): Add Cloud Identity

erezrokah:
#### Summary

Relevant API https://cloud.google.com/identity/docs/reference/rest

### Notes

- Needs the old Go API under `google.golang.org/api`
- Relations are not needed as the API supports wildcards, e.g. `devices/-/deviceUsers/-`
- Does not include https://cloud.google.com/identity/docs/reference/rest/v1/groups/list as it needs either `identity_source` or `customerId`. Maybe we can get them using an API call, or ask users to add to the spec. I chose not to implement this for now
- I get permissions error on our account so maybe @yevgenypats can test this


yevgenypats (approved):

erezrokah:
Going to skip this resource as it requires using a service account (not recommended for security reasons) and also enable domain wide delegation https://stackoverflow.com/questions/70583072/google-cloud-identity-api-to-list-or-get-devices

PR #6353, opened 2023-01-04
feat: Oracle plugin

shimonp21:
First version of oracle plugin. See more information below.

TODOs:
- Add descriptions with links to oracle's API
- Add mock tests.
- Add more resources
   - Understand why `mysql` and `nosql` resources make the plugin get stuck.
   - You can modify the `codegen/main.go` `servicesAllowlist` and `servicesDenylist` to generate more resources.
- Fix column names (e.g. `oracle_volumes->size_in_m_bs`)
- Add nested tables.

Codegen explanation:
- `codegen` generates table-definitions and fetcher-functions, based on `oracle_clients.go`.
- The `oracle_clients` was semi-manually generated by me using the script `generate_oracleClients_struct.py`.
- `generate_oracleClients_struct.py` requires cloning [https://github.com/oracle/oci-go-sdk] and running the [pkgreflect] tool on it.


yevgenypats (changes requested):
Let's remove the codegen and use built-in plugin-sdk transformations instead . will be easier for my to review 2,000 LOCs lines instead of 19,082 LOCs.

plugins/source/oracle/codegen/generate_OracleClients_struct.py
```diff
@@ -0,0 +1,166 @@
+# Outputs an `OracleClients` struct, and intialization code for the struct.
+# To be used with one of `--imports`, `--initialization`, or `--struct` flags.
```
hermanschaaf, line 2: I'm not sure what work this is doing exactly, but we may be able to adapt https://github.com/cloudquery/plugin-sdk/pull/590 to also fit this use case? Seems like it should be possible

yevgenypats: Or adapt this - https://github.com/cloudquery/cloudquery/tree/main/plugins/source/azure/codegen0. Def don't see a strong need for python as Go has very strong library for parsing source code which is done in azure. maybe we can outsource this utility to a seperate codegen library as it's not really part of the SDK

shimonp21: This creates the `oracle_clients.go` file, that is used by the later stages of codegen for discovery purposes.
I kinda went for the fastest possible way (so I used the `https://github.com/ungerik/pkgreflect` library, and this script that does regex on it's output).

I would be happy to not rewrite this (because this was kind of a one-off - the OracleClients struct doesn't really need to be regenerated).

But I can rewrite this in golang if oyu both feel strongly about htis.

shimonp21: Honestly, I'd prefer to just delete this script rather than rewrite it more generically right now, since I am not sure rewriting will bring any benefit (the `OracleClients` struct will remain the same)..

hermanschaaf: Yeah don't worry, I don't think it's a blocker. Deleting the script and updating the file by hand from now on sounds like a good option for now.

shimonp21: As per slack discussion:
- removed `// Code generated by codegen` lines
- changed `.github` presubmits to only test doc-changes, not code-changes.
- Removed the python script: you can find it at https://gist.github.com/shimonp21/70c0a3cceaeab3f054e76c2846810b3e

yevgenypats (approved):
It looks good. Caveat: I didn't have a chance to run this with a proper Oracle account and test if the spec and multiplexers make the most sense. If it get's more traction we can fine-tune it later.

PR #6358, opened 2023-01-04
chore(website): Fix incorrect type and example for backoff_retries on GCP source plugin

mark-adams:
#### Summary

This fixes two issues with the GCP source plugin documentation:

1. Fixes a mistake where the `backoff_delay`'s type and default were incorrect. The parameter is actually an `int` (not a `string`) and the default is `30` not `30s`. Clarifying text was also added to indicate the parameter is the maximum delay in seconds.
2. Fix indentation on `enabled_services_only`


yevgenypats (approved):
Awesome. Thank you!!

PR #6366, opened 2023-01-05
chore(ci-lint): Only enable lint caching

erezrokah:
#### Summary

Our linter caches are huge and also exceed the limits so impact other workflows based on caching:
image

On AWS it takes ~60 seconds to just download the cache:
image

This PR tries to improve the situation by only keeping the lint cache, and disabling the build+dependencies cache.
If this doesn't work, I'll revert to the original action and disable caching altogether

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -59,6 +59,8 @@ jobs:
         with:
           version: v1.50.1
           working-directory: cli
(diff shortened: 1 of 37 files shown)


erezrokah:
Will open again not from a fork

erezrokah:
Continued in https://github.com/cloudquery/cloudquery/pull/6367

PR #6369, opened 2023-01-05
feat(aws): Remove codegen, use built-in transformers (Part 3: starting from Z)

disq:


plugins/source/aws/client/helpers.go
```diff
@@ -396,3 +397,18 @@ func Sleep(ctx context.Context, dur time.Duration) error {
 		return nil
 	}
 }
+
+func CreateTrimPrefixTransformer(prefixes ...string) func(field reflect.StructField) (string, error) {
```
disq, line 401: This was added to help with getting prefixes of some columns removed

yevgenypats (approved):
💨  Nice!

PR #6381, opened 2023-01-05
fix(bigquery): Implement some mitigations for Bigquery data and metadata delays

hermanschaaf:
These changes make our tests pass more reliably, and is intended to help useres who are using the BigQuery plugin for the first time or migrating tables.

Changes:
 - Check schema metadata 3 times before accepting it as OK after a migration.
 - Retry writes if the returned error is that the table was not found. This will be done until the write timeout is reached (5 minutes) or some other error is returned.


erezrokah (approved):

yevgenypats (approved):
🪄

Issue #6365, opened 2023-01-05
feat(gcp): Move as many resources as possible to new client


erezrokah:
We have a bunch of resources using https://github.com/googleapis/google-api-go-client and not https://github.com/googleapis/google-cloud-go.

Using 2 different clients creates inconsistency in our code, for example only the new client [supports retries](https://discord.com/channels/872925471417962546/1060374442913579019/1060494955832950795).

For some resource we must use the old one (for example https://github.com/cloudquery/cloudquery/pull/5919), but we should try and migrate resources (based on demand).

erezrokah:
If someone wants to contribute see example PR in https://github.com/cloudquery/cloudquery/pull/6390

rafabelts:
Hello, I would like to contribute and work on the migration 😁

erezrokah:
Hi @rafabelts 👋 That's great, please let us know if something is missing from the issue description or comments to get you started.

rafabelts:
Hi👋, just to confirm, I have to migrate from https://github.com/googleapis/google-api-go-client to https://github.com/googleapis/google-cloud-go, right?, but , should some resources like #5919 stay with the old client or should they be moved to the new one?

erezrokah:
> Hi👋, just to confirm, I have to migrate from https://github.com/googleapis/google-api-go-client to https://github.com/googleapis/google-cloud-go, right?, but , should some resources like #5919 stay with the old client or should they be moved to the new one?

Yes, exactly. I recommend first commenting with the list of resources that still use https://github.com/googleapis/google-api-go-client, then we should be able to investigate each one

rafabelts:
Ok ok, well, just one more question, where can I find the resources that are still using the old gcp client?

erezrokah:
Search under this directory https://github.com/cloudquery/cloudquery/tree/b4918ff09f7dae4d20f1beda896d2ec71233716b/plugins/source/gcp/resources/services for resources using `google.golang.org/`. Some are only using `google.golang.org/api/iterator` which is file, but others might be using an old client

rafabelts:
I will check it and then start working, is there any deadline?

erezrokah:
> is there any deadline?

No deadline 😄 Thanks @rafabelts 🙌

rafabelts:
Alright 😄, so if you want you can assign me the issue, I will try to have it ready as soon as possible.

erezrokah:
[...]

PR #6384, opened 2023-01-05
feat(gcp): Remove codegen

erezrokah:
#### Summary

This PR removes code gen from GCP (it keeps only service discovery).
Blocked by https://github.com/cloudquery/plugin-sdk/pull/578 and https://github.com/cloudquery/plugin-sdk/pull/579.

While working on this PR I noticed a few cases where we configured the wrong PKs on tables. I'll fix those separately

PR #6390, opened 2023-01-05
feat(gcp): Move services/iam to use googleapis/google-cloud-go

mark-adams:
#### Summary
Currently, services/iam in the gcp source plugin uses the googleapis/google-api-go-client which does not have support for automatically retrying failed requests.

This change replaces all of the API client code in services/iam with the applicable clients from googleapis/google-cloud-go which does support retrying failed requests. Ideally, templates/newapi_list.go.tpl would be modified to clear the CallOptions on the client so we'd only be respecting backoff_delay and backoff_retries. However, this change would modify behavior in multiple other services so I opt'd to save it for a separate PR to keep this one more concise.

I also had to modify the associated codegen recipe since GRPC was now being used.

Related to #6365, cc: @erezrokah


erezrokah (approved):
Thanks for the PR @mark-adams 🚀 This looks great. I took the liberty to merge this PR with the changes in https://github.com/cloudquery/cloudquery/pull/6384.

I'll also release this as a major version bump due to the column type changes

PR #6391, opened 2023-01-05
fix(aws): Enable Partial Resolution of Lambda

bbernays:
#### Summary

This enables CQ to store the results of `ListFunctions` even if the `GetFunction` fails because of a permission error. This was a fix that we had implemented in V0


plugins/source/aws/resources/services/lambda/functions_fetch.go
```diff
@@ -47,6 +47,11 @@ func getFunction(ctx context.Context, meta schema.ClientMeta, resource *schema.R
 			}
 			return nil
 		}
+		if client.IsAWSError(err, "AccessDenied") {
```
disq, line 50: Why not put it as an `or` in L45?

bbernays: I was specifically leaving that out as we want to be clear that we are missing some information because of a permission issue

disq: Oh ok, so maybe a comment? I checked the sdk and apparently you're right and the resource will stay even in case of an error from the PreResourceResolver (albeit half baked)

bbernays: Added a comment in the code

hermanschaaf (approved):

yevgenypats (approved):

PR #6392, opened 2023-01-05
feat(gcp):  Improve Error messages for Rate Limited

bbernays:
#### Summary

When large orgs use `enabled_services` they may well hit rate limits. This pr helps with documentation and gives the user a specific error message when it occurs

### DIFF
--- a/website/pages/docs/plugins/sources/gcp/configuration.md
+++ b/website/pages/docs/plugins/sources/gcp/configuration.md
@@ -55,7 +55,7 @@ This is the (nested) spec used by GCP Source Plugin
   If specified APIs will be retried with exponential backoff if they are rate limited. This is the max number of retries.
 
 - `enabled_services_only` (bool) (default: false).
-If enabled CloudQuery will skip any resources that belong to a service that has been disabled or not been enabled. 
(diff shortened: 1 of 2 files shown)


disq (approved):

website/pages/docs/plugins/sources/gcp/configuration.md
```diff
@@ -55,7 +55,7 @@ This is the (nested) spec used by GCP Source Plugin
   If specified APIs will be retried with exponential backoff if they are rate limited. This is the max number of retries.
 
 - `enabled_services_only` (bool) (default: false).
-If enabled CloudQuery will skip any resources that belong to a service that has been disabled or not been enabled. 
+If enabled CloudQuery will skip any resources that belong to a service that has been disabled or not been enabled. If you use this option on a large organization (greater than 500 projects) you should also set the `backoff_retries` to a value greater than `0` otherwise the sync could fail because of rate limiting.
```
erezrokah, line 58:
```suggestion
If enabled CloudQuery will skip any resources that belong to a service that has been disabled or not been enabled. If you use this option on a large organization (with more than 500 projects) you should also set the `backoff_retries` to a value greater than `0` otherwise the sync could fail because of rate limiting.
```

erezrokah (approved):
Looks, good one comment on the docs

Issue #6145, opened 2022-12-30
Add SQL Server as a supported destination


rafaelrodrigues3092:
### Is your feature request related to a problem? Please describe.
We have existing client and application environment details hosted in SQL Server.
We would like to tie Azure infrastructure to these environments.

### Describe the solution you'd like
Have SQL Server as a supported destination

### Describe alternatives you've considered
Currently using Postgres. We have been thinking of setting an ETL job to load data from PG to SQL Server.

### Additional context
SQL Server is one of the most widely used RDBMS

candiduslynx:
Hi @rafaelrodrigues3092!
We've just merged the SQL Server destination that works both with local/dedicated SQL server as well as Azure SQL Server.
We'll be releasing tomorrow, so you can check it out soon.

rafaelrodrigues3092:
This is awesome!
Thank you for the quick turnaround!
I will give it a try once it's out.
Thanks again!

candiduslynx:
@rafaelrodrigues3092, the plugin is [released](https://github.com/cloudquery/cloudquery/releases/tag/plugins-destination-mssql-v1.0.0)!
https://www.cloudquery.io/docs/plugins/destinations/mssql/overview

rafaelrodrigues3092:
@candiduslynx 
I was able to test it out. I did a single run from Azure (single subscription) -> MSSQL
At first glance, it seems to be working fine. I did notice some Primary Key Violations in the log.

Let me know if you want me to add them here, or can create a separate ticket with all my (potential) findings once I have a little more time to do a more thorough check.

Thanks again for this!

candiduslynx:
@rafaelrodrigues3092 Please create a separate ticket for us to review. Thanks!

candiduslynx:
> I will try to set up an Azure AD authentication scenario.

One tricky thing is that you'll need to set up application ID for this, too. And don't forget to set `auth_mode` to `azure`.

rafaelrodrigues3092:
> > I will try to set up an Azure AD authentication scenario.
> 
> One tricky thing is that you'll need to set up application ID for this, too. And don't forget to set `auth_mode` to `azure`.

@candiduslynx I think this is the case - but just to double check -  does this mean that the Service Principal used to collect the infrastructure resources on the target Azure Subscriptions (as described here https://www.cloudquery.io/docs/plugins/sources/azure/overview#creating-a-service-principal) is also the entity used to authenticate with the database server?

[...]

PR #6417, opened 2023-01-06
feat(mssql): Microsoft SQL Server destination

candiduslynx:
Implements #6145.
Depends on:
* cloudquery/plugin-sdk/pull/585
* cloudquery/plugin-sdk/pull/606


candiduslynx:
This PR still required adding test workflow for destination testing + docs. Code is almost ready, though.

plugins/destination/mssql/client/write.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/cloudquery/plugin-sdk/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/specs"
+)
+
+func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resources [][]any) error {
```
yevgenypats, line 12: This shouldn't be implemented. Destination plugin can't be both managed and unmanaged. For mysql you most prob want unmanaged as mentioned on slack because for managed plugin this means it will have to create more connections which mssql is not great at just like pg.

yevgenypats (changes requested):
Right now the plugin incorrectly implements multiple interfaces and it needs to only implement the unmanaged plugin.

plugins/destination/mssql/client/write.go
candiduslynx: Thanks for pointing this out!

candiduslynx:
TODO: add test runner like in https://github.com/Microsoft/mssql-docker/issues/133#issuecomment-883348917

erezrokah:
Hi @candiduslynx can we make sure to incorporate the fix from https://github.com/cloudquery/cloudquery/pull/6519?

candiduslynx:
> Hi @candiduslynx can we make sure to incorporate the fix from https://github.com/cloudquery/cloudquery/pull/6519?

@erezrokah I think I already included those

plugins/destination/mssql/main.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/destination/mssql/client"
+	"github.com/cloudquery/cloudquery/plugins/destination/mssql/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const (
+	sentryDSN = "TODO"
```
disq, line 11:
```suggestion
	sentryDSN = "https://f745aa5eaa44407ca4003a9c77a1b240@o1396617.ingest.sentry.io/4504481164754944"
```

plugins/destination/mssql/client/write.go
```diff
@@ -0,0 +1,118 @@
...
+
+func insert(table *schema.Table) string {
+	var sb strings.Builder
+
+	sb.WriteString("insert into ")
+	sb.WriteString(sanitizeIdentifier(table.Name))
+	sb.WriteString("\n(\n\t")
+
+	columns := tableColumnsSanitized(table)
+	sb.WriteString(strings.Join(columns, ",\n\t"))
```
yevgenypats, line 51: why `\n\t` is need for sql commands?

plugins/destination/mssql/client/write.go
```diff
@@ -0,0 +1,118 @@
...
+		query = insert
+	}
+
+	batchSize := c.spec.BatchSize
+
+	batch := make([]*stmt, 0, batchSize)
+	for r := range resources {
+		table := tables.Get(r.TableName)
+		if table == nil {
+			panic(fmt.Errorf("table %s not found", r.TableName))
```
yevgenypats, line 26: I don't think you need to panic here. just return an error it will stop the sync

plugins/destination/mssql/client/write.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		batch = append(batch, &stmt{query: query(table), params: r.Data})
+		if len(batch) == batchSize {
+			if err := c.send(ctx, batch); err != nil {
+				return err
+			}
+
+			// just clip, capacity is left intact (opposed to batch[:0:0])
+			batch = batch[:0]
```
yevgenypats, line 36: why not `batch := make([]*stmt, 0, batchSize)` ? what are you optimising for?

yevgenypats (changes requested):
Lot's of code :)

Just first round of review. I wasn't able to connect to Azure MS SQL -I think it is worth documenting how to do it or if it is more involved then a HowTo guide.

I tried and im getting:

```
```

plugins/destination/mssql/client/write.go
candiduslynx: nott required, but allows to read queries and debug them easier

plugins/destination/mssql/client/write.go
candiduslynx: just reusing the same memory

plugins/destination/mssql/client/write.go
candiduslynx: I copy-pasted it from [postgres](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/write.go#L64), I can change it to just returning the err.

plugins/destination/mssql/client/write.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/specs"
+)
+
+func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data [][]any) (err error) {
+	if len(data) == 0 {
```
yevgenypats, line 11: I think this is not needed as `WriteTableBatch` wont be called with no data

yevgenypats (changes requested):
Very nice! I really like the bulk-insert. In overwrite I think we can do something similar otherwise it will be quite slow / open many goroutines and so on.

I think in overwrite we should use a temptable and then do a merge on the serverside - it's quite common approach in databases that don't have good built-in batch support.

https://blog.jonschneider.com/2016/08/bulk-upsert-into-sql-server.html
https://stackoverflow.com/questions/4889123/any-way-to-sqlbulkcopy-insert-or-update-if-exists

candiduslynx:
@yevgenypats I tried using temporary tables and the overhead was enormous (see https://github.com/cloudquery/cloudquery/pull/6417/commits/fae6b7af7905541a94c42a086bc987ce2b248a48).
The code was creating tmp table, doing bulk insert and only then performing merge.

plugins/destination/mssql/client/write.go
candiduslynx: removed

candiduslynx:
@erezrokah seems that some base links (not touched by this PR) were broken

candiduslynx:
@yevgenypats I used table value parameters along with stored procedure to achieve better results.

candiduslynx:
@yevgenypats thanks for the review!
There was an attempt to use temporary tables + bulk insert + merge (see my comment above: https://github.com/cloudquery/cloudquery/pull/6417#issuecomment-1382838724).
Overall, creating a temporary table added a huge overhead, so the only feasible way of performing the overwrite would be to use staging tables or do the TVP approach I implementetd.

.github/workflows/dest_mssql.yml
```diff
@@ -0,0 +1,109 @@
+name: Destination Plugin Microsoft SQl Server Workflow
```
disq, line 1: `name: Destination Plugin Microsoft SQL Server Workflow` (`SQl` => `SQL`)

Issue #6416, opened 2023-01-06
New Source Plugin: AliCloud


lshmouse:
### Why do you need this resource or plugin? Please describe.

AliCloud(https://hk.alibabacloud.com/) is the leading public cloud provider in China. We want to add this source plugin to sync the resource like billings, oss bucket stats, ecs stats, cloud metrics and so on from the cloud,  to support cost analysis and optimization, resource compliance.

### Describe the solution you'd like
Add alicloud source to support sync following data from the cloud
- billings
- oss resource
- ecs resource 
- cms resource

Also we want to add more policy about cost optimization and resource compliance in this plugin.

### Additional context

hermanschaaf:
Hi @lshmouse, thanks again for raising this ticket and contributing the [first PR](https://github.com/cloudquery/cloudquery/pull/6418) for this plugin. Good news: we just released the first official version of the alicloud plugin! https://www.cloudquery.io/docs/plugins/sources/alicloud/overview 🎉 

Right now it supports Billing, OSS and ECS (Instances). CMS resources are not yet supported like you originally requested, but I raised a separate ticket for it here, so anyone is free to pick it up: https://github.com/cloudquery/cloudquery/issues/6930.

I will close this issue now, but feel free to follow up with new issues if there is anything missing or out of place. Thanks again!

PR #6418, opened 2023-01-06
feat(alicloud): Add alicloud source plugin to sync bss and oss resources

lshmouse:
#### Summary

https://github.com/cloudquery/cloudquery/issues/6416

Add the alicloud source plugin to sync bss and oss resources


plugins/source/alicloud/client/resolvers.go
```diff
@@ -0,0 +1,20 @@
...
+
+func ResolveParentColumn(field string) schema.ColumnResolver {
+	return func(_ context.Context, _ schema.ClientMeta, r *schema.Resource, c schema.Column) error {
+		return r.Set(c.Name, funk.Get(r.Parent.Item, field))
+	}
+}
+
+func ResolveUpdateDate() schema.ColumnResolver {
+	return func(_ context.Context, _ schema.ClientMeta, r *schema.Resource, c schema.Column) error {
+		return r.Set(c.Name, time.Now().AddDate(0, 0, -1).Format("2006-01-02"))
```
hermanschaaf, line 18: Why is this column resolver needed, and why does it add `-1` days? Do you need information about exactly when a row was added? The `_cq_sync_time` column doesn't give exact row-level data, but it does tell you when a sync started, which is usually good enough.

plugins/source/alicloud/codegen/recipies/oss.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/aliyun/aliyun-oss-go-sdk/oss"
+	"github.com/cloudquery/plugin-sdk/codegen"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func OSS() []*Resource {
+	bucketName := codegen.ColumnDefinition{
+		Name:     "name",
+		Type:     schema.TypeString,
+		Resolver: `client.ResolveParentColumn("Name")`,
```
hermanschaaf, line 13: I think we should be able to use `schema.ParentColumnResolver` from the plugin-sdk library here (e.g. https://github.com/cloudquery/cloudquery/blob/ec1d132529824d16682a92a83ee326fcba939efc/plugins/source/aws/codegen/recipes/s3.go#LL61C2-L61C2), and remove the custom resolver from this plugin. I notice this pattern was copied from the Github plugin, we should probably fix it there too in a follow-up PR :)

plugins/source/alicloud/config/alicloud.yaml
```diff
@@ -0,0 +1,22 @@
...
+    secret_key: ""
+---
+kind: destination
+spec:
+  name: "csv"
+  path: "cloudquery/csv"
+  version: "v1.1.12"
+  write_mode: "append" # CSV only supports 'append' mode
+  spec:
+    directory: './cq_csv_output'
```
hermanschaaf, line 22: This is a good way for developers to test the plugin against real infrastructure for now, but eventually we'll probably remove it here and move it to the docs. 

For testing in CI we'll require mock tests for the resolvers. Here's [an example from the Github plugin](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/github/resources/services/external/groups_test.go). I can also help you write those if you like, they are fairly mechanical once you've created the mocks :)

plugins/source/alicloud/client/resolvers.go
hermanschaaf: Oh, after reading a bit more of the PR, I see that this column is part of the PK, probably to avoid overwriting rows from previous syncs. You can also achieve the same by leaving the PK empty, in which case it will default to `_cq_sync_id`. In `overwrite-delete-stale` mode CloudQuery will first sync all rows, then delete rows from previous syncs.

plugins/source/alicloud/go.mod
```diff
@@ -0,0 +1,47 @@
+module github.com/cloudquery/cloudquery/plugins/source/alicloud
+
+go 1.19
+
+require (
+	github.com/alibabacloud-go/tea v1.1.20
+	github.com/aliyun/alibaba-cloud-sdk-go v1.62.80
+	github.com/aliyun/aliyun-oss-go-sdk v2.2.6+incompatible
+	github.com/cloudquery/plugin-sdk v1.11.1
```
hermanschaaf, line 9: This plugin-sdk version is a little out-of-date, let's make sure to use the latest version (`v1.21.0`)

plugins/source/alicloud/resources/services/bss/fetch.go
```diff
@@ -0,0 +1,174 @@
...
+		Currency:              item.Currency,
+		CommodityCode:         item.CommodityCode,
+		BillAccountName:       item.BillAccountName,
+		AdjustAmount:          item.AdjustAmount,
+		CashAmount:            item.CashAmount,
+	}
+}
+
+/*
+* https://help.aliyun.com/document_detail/100392.html
```
hermanschaaf, line 81: These links are great - we should also add them to the `Description` field of each table so they show up in the docs :)

plugins/source/alicloud/resources/services/bss/fetch.go
```diff
@@ -0,0 +1,174 @@
...
+ */
+func fetchBssBill(_ context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
+	spec := meta.(*client.Client).Spec
+	cli, err := bssopenapi.NewClientWithAccessKey(spec.RegionID, spec.AccessKey, spec.SecretKey)
```
hermanschaaf, line 85: I don't know if it's possible in this case, but plugins generally initialize the client in `client.go`, and store it as a property on the `client.Client` struct, so we could then access it here like

```
c := meta.(*client.Client)
cli := c.Client
```

This has the added benefit of erroring out earlier if there is an issue initializing the client, before calling all the resolvers.

plugins/source/alicloud/client/spec.go
```diff
@@ -0,0 +1,8 @@
+package client
+
+type Spec struct {
+	RegionID    string `json:"region_id,omitempty"`
+	AccessKey   string `json:"access_key,omitempty"`
+	SecretKey   string `json:"secret_key,omitempty"`
+	BillHistory int    `json:"bill_history,omitempty"`
```
hermanschaaf, line 7: What does this option do? Looks like it either loads the billing history or only the latest bill? Should it be a boolean in that case? I think we'll probably soon have a better way of handling historical billing cycles (https://github.com/cloudquery/plugin-sdk/pull/569), so maybe we can update it to use that when it's ready.

plugins/source/alicloud/resources/services/bss/fetch.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	for _, billingCycle := range billingCycles {
+		request := bssopenapi.CreateQueryBillOverviewRequest()
+		request.BillingCycle = billingCycle
+		response, err := cli.QueryBillOverview(request)
+		if err != nil {
+			return err
+		}
+		if response.GetHttpStatus() != http.StatusOK {
+			return fmt.Errorf("httpcode %d", response.GetHttpStatus())
```
hermanschaaf, line 39: We may want to handle retryable errors (like 5xx and 429), either here or in a generalized function. We don't have a perfect pattern for this yet, but see the Slack plugin for an example: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/slack/resources/services/users/users_fetch.go#L20 and https://github.com/cloudquery/cloudquery/blob/main/plugins/source/slack/client/rate_limiting.go#L13

plugins/source/alicloud/codegen/models/bss.go
```diff
@@ -0,0 +1,64 @@
+package models
+
+type BillOverview struct {
```
hermanschaaf, line 3: Mostly out of curiosity, what is the difference between these structs and the ones provided by the Alicloud SDK? Just wondering why (or whether) we need it

hermanschaaf (changes requested):
This is amazing, thank you @lshmouse! Really excited to be adding support for Alicloud. I left a few comments after an initial review. I'll also try and test it out a bit soon so I can give more thorough feedback. Let me know if you're willing to add some mock tests or if you'd like some help with that :)

lshmouse:
@hermanschaaf Thanks very much for your careful review. I will update the pull request later, thanks~

hermanschaaf:
Hi @lshmouse :wave: Just wanted to update you: I spent a bit more time going through the code today, and started bringing it up-to-date with the latest plugin-sdk and best practices. I'll push the commits to your branch once they're ready and then we can merge, but there's still a bit more work to do which I'll pick up net week. My work-in-progress branch is here if you want to have a look or merge it into your branch sooner rather than later: https://github.com/cloudquery/cloudquery/tree/alicloud Hope you don't mind!

hermanschaaf (approved):
Thanks again @lshmouse for this great contribution!

lshmouse:
@hermanschaaf I was blocked by something else in the past two weeks. Thanks very much for completing the patch~

PR #6424, opened 2023-01-06
feat(datadog): Migrate codegen to transformations

yevgenypats:


hermanschaaf:
@yevgenypats I think this PR included changes to the crowdstrike plugin by accident as well

disq:
`AdditionalProperties` was always skipped in codegen, maybe do it here as well?

yevgenypats:
> `AdditionalProperties` was always skipped in codegen, maybe do it here as well?

why was it skipped though?

disq:
> > `AdditionalProperties` was always skipped in codegen, maybe do it here as well?
> 
> why was it skipped though?

Not sure but it looks like an empty, undocumented map.

hermanschaaf (approved):

Issue #6433, opened 2023-01-06
bug: Cockroachdb tests failing


yevgenypats:
check why cockroachdb tests fail with new SDK https://github.com/cloudquery/cloudquery/pull/6421 (currently commented them out)

erezrokah:
The issue in the tests is described and fix in https://github.com/cloudquery/cloudquery/pull/6519#issue-1525200023

PR #6434, opened 2023-01-06
Revert "fix: Error if after the migration there are `not null` columns that are not part of the new schema"

yevgenypats:
I think this should be given additional take, potentially not doing any drops as we should avoid those like fire especially when running those in distributed mode and so on.


hermanschaaf (approved):

yevgenypats:
Also closes https://github.com/cloudquery/cloudquery/issues/6433

PR #6470, opened 2023-01-07
feat(fastly): Migrate codegen to transformations

yevgenypats:


plugins/source/fastly/codegen/services/generate.go
```diff
@@ -1,175 +0,0 @@
-package services
```
hermanschaaf, line 1: I think this file is useful but it can be made part of the `make gen-mocks` command. 

As an aside (not for this PR), I'm also thinking it would be good to move these to the SDK at some point, with it being used in quite a few plugins now I think we can see how it can be generalized. That would reduce the amount of code for this specific plugin as well.

yevgenypats: We can also discuss this live but I looked a bit at the services codegen again and it seems that it is not a lot of work to add a new interface for each table we develop. I think "codegen based development" is a real barrier to entry for developers and if it doesn't provide the 5,10x speed I would skip this completely.

hermanschaaf: I very strongly disagree here - most of the time a new plugin author needs to do it only once when setting up the plugin, and after that (if you used the automation) it rarely needs to be changed again. So it shouldn't be a barrier to adding new tables or contribution in general (in fact, the opposite). It's far, far less work to point to one struct and say "generate the interface for it" than it is to copy-paste every relevant function on that struct, modify all the import paths appropriately, and then make sure the mockgen comments for it are correct as well.

erezrokah: I would add that auto generating the interfaces ensures that when we update dependencies those get auto updated as well

hermanschaaf: I chatted with @yevgenypats about this; we agreed to move the interface generation functionality to the plugin SDK for those plugins that use mockgen tests and have complex enough clients to justify it.

hermanschaaf (approved):

PR #6471, opened 2023-01-07
feat(slack): Migrate codegen to transformations

yevgenypats:


hermanschaaf (approved):
Looks good; I'll fix the service interface generation bit after we merge

hermanschaaf:
Oh wait, there are some added columns that should remain skipped because the API doesn't actually ever return them, so it's confusing to keep them there

hermanschaaf:
The two columns are `has_2fa` and `presence`

PR #6472, opened 2023-01-07
feat(azuredevops): Migrate codegen to transformations

yevgenypats:


erezrokah (approved):
Did a small follow up to use `WithTypeTransformer` instead of skipping + redefining the `Id` field

PR #6530, opened 2023-01-09
fix(gcp): Etag to ByteArray

disq:
Fixes breaking change introduced in #6390 (was String, made IntArray, now ByteArray)

~~Not sure if we need a stringifying-resolver for the etag, will test.~~


erezrokah (approved):

disq:
We need a string resolver, will add.

disq:
String resolver behaves differently with data from `gcp_iam_service_accounts` vs. `gcp_iam_roles`.

For `gcp_iam_service_accounts` it returns `01021920`
For `gcp_iam_roles` it returns `\x07\x05Ƅ�)Sr` so it makes sense to use a hex encoder. (encoded: 0705c684e2295372)

If we encode the first one we come up with `3031303231393230` though... Endpoint discrepancy?

According to https://pkg.go.dev/cloud.google.com/go/iam@v0.10.0/admin/apiv1/adminpb#ServiceAccount `Etag` is deprecated in the ServiceAccount struct.

plugins/source/gcp/client/transformers.go
```diff
@@ -38,6 +38,10 @@ func replaceTransformer(field reflect.StructField) (string, error) {
 }
 
 func typeTransformer(field reflect.StructField) (schema.ValueType, error) {
+	if field.Name == "Etag" && field.Type == reflect.TypeOf([]byte{}) {
+		return schema.TypeString, nil
```
erezrokah, line 42:
```suggestion
		return schema.TypeByteArray, nil
```

Maybe we can do this until the SDK fix?

erezrokah (approved):

PR #6550, opened 2023-01-09
fix(gcp-resources): Remove `bucket_name` as PK of `gcp_storage_bucket_policies`

erezrokah:
#### Summary

`gcp_storage_bucket_policies` used the parent bucket's name as a PK. This means we could only get 1 policy per bucket in `overwrite*` modes.
Per a better field, this PR fixes it so we use `_cq_id`


yevgenypats:
Seems in that case the right pk is `version + parent_name`

erezrokah:
> Seems in that case the right pk is `version + parent_name`

The policy struct doesn't have a `version` (only `Etag` and an array of `Bindings`):
https://github.com/googleapis/google-cloud-go/blob/947c9ace8740e40abe0ca79acb894ea333655f88/iam/iam.go#L353

yevgenypats (approved):

PR #6551, opened 2023-01-09
fix(gcp-resources): Add `version` to PKs of `gcp_cloudiot_device_configs`

erezrokah:
#### Summary

`gcp_cloudiot_device_configs` used the parent resource `name` as PK (and `project_id`). Those are not good enough as it means we can only have 1 config per device. Since there isn't a good field to use for uniqueness it's better to use `_cq_id`


erezrokah:
> Good catch. Isn't the `version` + `device_name` is the right pk? per docs -

I was thinking about that too. But it doesn't say the version is unique per device (though I'd expect it to be). It only says it is set by the server

erezrokah:
> Good catch. Isn't the `version` + `device_name` is the right pk? per docs -

OK so let's try with `version`. If it's not unique we can drop the PKs

yevgenypats (approved):

PR #6552, opened 2023-01-09
fix(gcp-resources): Add `update_time` to PKs of `gcp_cloudiot_device_states`

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/6551


yevgenypats:
I think in that case the right pk is `device_name` + `updateTime`

erezrokah:
> I think in that case the right pk is `device_name` + `updateTime`

Is it guaranteed that `updateTime` is unique? I don't know actually. Maybe it's possible to update the device twice at the same time

erezrokah:
> I think in that case the right pk is `device_name` + `updateTime`

OK let's try with `update_time` and see how it goes

yevgenypats (approved):

erezrokah:
Something is up with the test failures, not sure why it fails the PK validation

erezrokah:
Blocked by https://github.com/cloudquery/plugin-sdk/pull/596, a bug in our type system

erezrokah:
The real blocker is https://github.com/cloudquery/plugin-sdk/pull/597

erezrokah:
Another correction, we should be handling this in https://github.com/cloudquery/cloudquery/blob/434fe4d15746277e903edd3ce5635fe2323ed413/plugins/source/gcp/client/transformers.go#L54. I'll check why it's not working

erezrokah:
> Another correction, we should be handling this in

OK so I just used the wrong resolver `Path` instead of `ResolveProtoTimestamp` when redefining the column 🤦

PR #6560, opened 2023-01-10
feat(hackernews): Add ability to start from given date

hermanschaaf:
This adds the ability to specify a start date to start the sync from. Hacker News data goes back many years, and many use cases will not require syncing it all (which can take a day or two). With this setting, a date like Jan 1st 2022 can be given to sync only the last year of data, which can be done in around two-three hours from a modest laptop on a home internet connection.

In terms of implementation, we binary search for the first post after the given date, and then start the cursor from there.

This also fixes the timestamp column type, which accidentally got reverted during the transformer transition.


plugins/source/hackernews/resources/services/items/items_fetch.go
```diff
@@ -35,6 +36,13 @@ func fetchItems(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource,
 		return fmt.Errorf("failed to retrieve state from backend: %w", err)
 	}
 
+	// find the max item ID from the Hacker News API
+	maxID, err := c.HackerNews.MaxItemID(ctx)
+	if err != nil {
+		return err
+	}
+	c.Logger().Info().Msg("Found max ID, reading up to " + strconv.Itoa(maxID))
```
disq, line 44: erm, structured logging?

plugins/source/hackernews/resources/services/items/items_fetch.go
```diff
@@ -49,12 +57,23 @@ func fetchItems(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource,
...
-		return err
+	// we allow the user to specify a start time for posts, so we need to find the first post after that time
+	if c.Spec.StartTime != "" {
+		startTime, _ := time.Parse(time.RFC3339, c.Spec.StartTime)
+		c.Logger().Info().Msgf("Finding first post after %s", startTime)
+		startItemID, err := findFirstPostAfter(ctx, c, startTime, maxID)
+		if err != nil {
+			return err
+		}
+		c.Logger().Info().Msgf("Found first post after %s with id %d", startTime, startItemID)
```
disq, line 68: more structured logging needed :) also in line 63

hermanschaaf: done 👍

plugins/source/hackernews/client/resolvers.go
```diff
@@ -0,0 +1,24 @@
...
+		t := funk.Get(r.Item, fieldName, funk.WithAllowZero())
+		switch v := t.(type) {
+		case int64:
+			return r.Set(col.Name, time.Unix(v, 0))
+		case int:
+			return r.Set(col.Name, time.Unix(int64(v), 0))
+		case uint64:
+			return r.Set(col.Name, time.Unix(int64(v), 0))
+		}
+		panic("unknown type for UnixTimeResolver")
```
disq, line 22:
```suggestion
		return fmt.Errorf("unhandled UnixTimeResolver of type %T", t)
```
Or thereabouts? Probably as a `default` case in the switch.

hermanschaaf: Updated 👍

disq (approved):

Issue #6380, opened 2023-01-05
Datadog: Permissions or documentation issue `datadog_user_organizations `


yevgenypats:
Users report having Permission Denied even with a wide-open service account for datadog. we need to either double test that we don't have issues that give that error . The table in question is `datadog_user_organizations `

Example config:

```
  ---
  kind: source
  spec:
    name: datadog
    path: cloudquery/datadog
    version: "v1.0.5"
    destinations: ["postgresql"]
    tables:
      - datadog_permissions
      - datadog_users
      - datadog_user_permissions
      - datadog_roles
      - datadog_role_permissions
      - datadog_role_users
    spec:
      accounts:
      - name: datadog
        api_key: ${DD_CLIENT_API_KEY}
        app_key: ${DD_CLIENT_APP_KEY}
```

cloudmatt:
Since I reported it, I guess I can give the actual error too. The error it gives is:
```
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
2023-01-08T18:05:00Z ERR table resolver finished with error error="403 Forbidden" client=datadog module=datadog-src table=datadog_user_organizations
```

shimonp21:
[...]

PR #6561, opened 2023-01-10
fix: Remove broken `datadog_user_organizations` table

shimonp21:
See context in https://github.com/cloudquery/cloudquery/issues/6380


erezrokah (approved):
I'll let others comment but if this never worked I wouldn't make it a breaking change

Issue #6559, opened 2023-01-10
Support for AWS KMS Key Policies


jsonpr:
https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/docs/tables/aws_kms_keys.md does not contain information about KMS Key Policies

PR #6562, opened 2023-01-10
feat(aws): KMS Key policies

disq:
Implements #6559


plugins/source/aws/resources/services/kms/key_policies.go
```diff
@@ -0,0 +1,82 @@
...
+			},
+		},
+	}
+}
+
+func fetchKeyPolicies(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	c := meta.(*client.Client)
+	svc := c.Services().Kms
+
+	const policyName = "default"
```
hermanschaaf, line 66: Quite weird that the only valid name is `default` 🤷

hermanschaaf (approved):
Nice! ✨

plugins/source/aws/resources/services/kms/key_policies.go
disq: That's what the doc says... We could've left it off to add it later if needed maybe (but then the PK change would've been breaking)

PR #6576, opened 2023-01-10
fix: Revert "fix(gcp-resources): Add `organization_id` and `name` as PK of `gcp_resourcemanager_folders`

erezrokah:
#### Summary

Looks like `organization_id` can be `nil`. `name` should be good enough:
https://github.com/cloudquery/cloudquery/blob/d90410764f8d3e3c0bed82db0a143885e0c90190/plugins/source/gcp/resources/services/resourcemanager/folders_fetch.go#L44

### DIFF
--- a/plugins/source/gcp/docs/tables/gcp_resourcemanager_folders.md
+++ b/plugins/source/gcp/docs/tables/gcp_resourcemanager_folders.md
@@ -2,7 +2,7 @@
 
 https://cloud.google.com/resource-manager/reference/rest/v3/folders#Folder
 
-The composite primary key for this table is (**organization_id**, **name**).
+The primary key for this table is **_cq_id**.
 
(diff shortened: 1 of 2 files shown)


disq:
Can it be, though? Maybe remove the check and see if any weird org ids pop up?

erezrokah:
> Can it be, though? Maybe remove the check and see if any weird org ids pop up?

I'd rather do it separately and fix this bug first as it fails our migration code

hermanschaaf:
@erezrokah What if `name` isn't unique enough though, couldn't that cause rows to be erroneously overwritten?

erezrokah:
Let's do a full revert of https://github.com/cloudquery/cloudquery/pull/6547 until we figure this out

disq (approved):

PR #6577, opened 2023-01-10
fix(gcp): Prevent gcp_resourcemanager_folders rows with nil org_id

disq:


erezrokah:
This looks good, but let's wait until we finish the SDK migration and do this after https://github.com/cloudquery/cloudquery/pull/6576 as we'll need to release the PK change as breaking regardless (the data is missing at the moment)

plugins/source/gcp/resources/services/resourcemanager/folders_fetch.go
```diff
@@ -41,7 +41,10 @@ func fetchFolders(ctx context.Context, meta schema.ClientMeta, parent *schema.Re
 func resolveOrganizationId(_ context.Context, meta schema.ClientMeta, r *schema.Resource, c schema.Column) error {
```
erezrokah, line 41: Shouldn't we always use `cl.OrgId`?

erezrokah (changes requested):
I'm not sure why we need to get the organization from the `Parent` field.
The docs for `ListFolders` say:
```
// ListFolders lists the folders that are direct descendants of supplied parent resource.
```

So it seems the folders will always we under `cl.OrgId `. Isn't that the case?

erezrokah (approved):

PR #6578, opened 2023-01-10
fix(website): Update URL fpr Azure Resources View to the appropriate repo

valdearg:
Old URL seems to have been archived and the query errors.


erezrokah (approved):
Great find @valdearg! Thanks for the PR 🎸

PR #6580, opened 2023-01-10
chore(digitalocean): Remove codegen remarks

disq:


erezrokah (approved):
Once you remove those the lint runs on them :)

PR #6588, opened 2023-01-10
fix(gitlab): Update to new SDK

erezrokah:
#### Summary

Updates to latest SDK.
Detailed changelog in a markdown comment of this PR ⬇️  (edit this description to see)


disq:
ip address types as int/byte arrays? that's weird when we have a dedicated ip type? the type in the struct is `*net.IP`.

candiduslynx (approved):

erezrokah:
> ip address types as int/byte arrays? that's weird when we have a dedicated ip type? the type in the struct is `*net.IP`.

Good catch, fixed in https://github.com/cloudquery/cloudquery/pull/6588/commits/0ccf23be4b556ad5d6f252cecc9e13df18fa251c.

I think this should be done in the SDK as `net.IP` is a built in type WDYT?

disq (approved):

PR #6609, opened 2023-01-10
fix: Correct error message in migration

erezrokah:
#### Summary

Fixes the wording in the error, also dropping the column should suffice


erezrokah:
Will re-open not from a fork

PR #6643, opened 2023-01-10
feat(bigquery): Add call to action to type change error message

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/6612, this PR adds a call to action when we change field types and make the error message more consistent with other destinations


erezrokah:
Will open from a non fork 🤦

PR #6644, opened 2023-01-10
feat(bigquery): Add call to action to type change error message

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/6643 not from a fork.

Similar to https://github.com/cloudquery/cloudquery/pull/6612, this PR adds a call to action when we change field types and makes the error message more consistent with other destinations


plugins/destination/bigquery/client/migrate.go
```diff
@@ -178,10 +178,10 @@ func mergeSchemas(haveSchema, wantSchema bigquery.Schema) (bigquery.Schema, erro
 	for _, f := range haveSchema {
 		if want, ok := wantMap[f.Name]; ok {
 			if want.Type != f.Type {
-				return nil, fmt.Errorf("field %v changed type from %v to %v", f.Name, f.Type, want.Type)
+				return nil, fmt.Errorf("column %v changed type from %v to %v. Try dropping the column and re-running", f.Name, f.Type, want.Type)
 			}
 		} else if f.Required {
```
erezrokah, line 183: Do we need this `required` check? cc @hermanschaaf

hermanschaaf: Yes, if a required column is not in our schema, all inserts will fail

erezrokah: Are we setting it or is this something the user does?

hermanschaaf: It's something a user might do, so we're basically checking that they didn't do that.

hermanschaaf (approved):

Issue #6634, opened 2023-01-10
Incorrect API Version- SQL Server Registrations


bbernays:
### Describe the Bug

```
ERR table resolver finished with error error="GET https://management.azure.com/subscriptions/<SUBSCRIPTION_ID>/providers/Microsoft.AzureData/sqlServerRegistrations\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\nERROR CODE: InvalidResourceType\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"InvalidResourceType\",\n    \"message\": \"The resource type 'sqlServerRegistrations' could not be found in the namespace 'Microsoft.AzureData' for api version '2019-07-24-preview'. The supported api-versions are '2019-05-10-preview'.\"\n  }\n}\n--------------------------------------------------------------------------------\n" client=<SUBSCRIPTION_ID> module=azure-src table=azure_azuredata_sql_server_registrations
```

### CloudQuery Version

Source Plugin: Azure v3.1.0

PR #6672, opened 2023-01-11
fix(azure-resources): Set APIVersion to "2019-05-10-preview" when fetching SQL Server Registrations

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6634


erezrokah:
Opened https://github.com/Azure/azure-sdk-for-go/issues/19789 as we're using the latest API

yevgenypats (approved):

Issue #6632, opened 2023-01-10
Incorrect API Version- Healthbots


bbernays:
### Describe the Bug

```
2023-01-10T16:08:21Z ERR table resolver finished with error error="GET https://management.azure.com/subscriptions/<SUBSCRIPTION_ID>/providers/Microsoft.HealthBot/healthBots\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\nERROR CODE: InvalidResourceType\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"InvalidResourceType\",\n    \"message\": \"The resource type 'healthBots' could not be found in the namespace 'Microsoft.HealthBot' for api version '2021-06-10'. The supported api-versions are '2020-12-08,2021-08-24,2022-08-08'.\"\n  }\n}\n--------------------------------------------------------------------------------\n" client=<SUBSCRIPTION_ID> module=azure-src table=azure_healthbot_bots

```

### CloudQuery Version

Source Plugin: Azure 3.1.0

PR #6673, opened 2023-01-11
fix(azure-resources): Set APIVersion to "2022-08-08" when fetching Healthbots

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/6672.
Fixes https://github.com/cloudquery/cloudquery/issues/6632


erezrokah:
Opened https://github.com/Azure/azure-sdk-for-go/issues/19790 as we're using the latest API

yevgenypats (approved):

PR #6677, opened 2023-01-11
feat: Azure resources for policy implementation v2

amanenk:


.github/workflows/source_azure.yml
```diff
@@ -69,7 +69,9 @@ jobs:
         run: make gen
       - name: Fail if generation updated files
         if: github.event_name == 'pull_request'
-        run: test "$(git status -s | wc -l)" -eq 0 || (git status -s; exit 1)
+        run: |
```
erezrokah, line 72: Do we need this change? We print the changed files on failure

plugins/source/azure/resources/services/resources/links.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armlinks"
+
+	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func Links() *schema.Table {
+	return &schema.Table{
+		Name:      "azure_resources_links",
```
erezrokah, line 14: Should this table be added to the `tables.go` file? I don't see it in the docs

plugins/source/azure/resources/services/sql/database_vulnerability_assesment_scans.go
```diff
@@ -0,0 +1,54 @@
+// Code generated by codegen; DO NOT EDIT.
```
erezrokah, line 1: Let's remove this `// Code generated by codegen; DO NOT EDIT.` header (also others if exist - I didn't look for others)

plugins/source/azure/resources/services/sql/database_vulnerability_assesment_scans.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/sql/armsql"
+	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+// todo add as a database relation
+func databaseVulnerabillityAssesmentScans() *schema.Table {
```
erezrokah, line 13: Should this be a relation of `azure_sql_database_vulnerability_assesments`?

plugins/source/azure/resources/services/sql/database_vulnerability_assesments.go
```diff
@@ -0,0 +1,52 @@
...
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
+				},
+			},
+		},
+	}
+}
+
+func fetchDatabaseVulnerabillityAssesments(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	pp := parent.Item.(*armsql.Server)
```
erezrokah, line 32: Should this be the parent of the parent?
Also should this be a relation of a databases table? I couldn't find that table anywhere

erezrokah (changes requested):
Thanks for the PR @amanenk, looks good. I think we're still missing some table definitions and relations

plugins/source/azure/resources/services/monitor/diagnostic_settings.go
```diff
@@ -0,0 +1,80 @@
...
+	ResourceId string
+}
+
+func DiagnosticSettings() *schema.Table {
+	return &schema.Table{
+		Name:        "azure_monitor_diagnostic_settings",
+		Resolver:    fetchDiagnosticSettings,
+		Description: "https://learn.microsoft.com/en-us/rest/api/monitor/diagnostic-settings/list?tabs=HTTP#diagnosticsettingsresource",
+		Multiplex:   client.SubscriptionMultiplexRegisteredNamespace("azure_monitor_diagnostic_settings", client.Namespacemicrosoft_insights),
+		Transform:   transformers.TransformWithStruct(&diagnosticSettingsWrapper{}, transformers.WithPrimaryKeys("ID")),
```
erezrokah, line 25:
```suggestion
		Transform:   transformers.TransformWithStruct(&diagnosticSettingsWrapper{}, transformers.WithPrimaryKeys("ID"), transformers.WithUnwrapStructFields("DiagnosticSettingsResource")),
```

Since we're wrapping the original struct, you'll need to unwrap it to get the properties

erezrokah (changes requested):
The new changes look good, added a few more comments

erezrokah (approved):
New changes look good but I think you'll need to re-gen the docs after the `azure_monitor_diagnostic_settings` and maybe fix some tests?

Issue #6626, opened 2023-01-10
Invalid Request Error


bbernays:
### Describe the Bug

Invalid request:
```
ERR table resolver finished with error error="GET https://management.azure.com/subscriptions/<SUBSCRIPTION_ID>/providers/Nginx.NginxPlus/nginxDeployments\n--------------------------------------------------------------------------------\nRESPONSE 400: 400 Bad Request\nERROR CODE: MissingApiVersionParameter\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"MissingApiVersionParameter\",\n    \"message\": \"The api-version query parameter (?api-version=) is required for all requests.\"\n  }\n}\n--------------------------------------------------------------------------------\n" client=<SUBSCRIPTION_ID> module=azure-src table=azure_nginx_deployments
```

### Expected Behavior

There should be no error

### CloudQuery Version

Source Plugin: Azure 3.1.0
Destination plugin: postgresql version 2.0.4
CloudQuery Cli version: 2.0.31

PR #6682, opened 2023-01-11
fix(azure-resources): Update `armnginx` to v2

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6626


plugins/source/azure/resources/services/nginx/deployments_fetch.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchDeployments(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	// This is required so we don't override mock test options
+	if cl.Options == nil {
+		cl.Options = &arm.ClientOptions{}
+	}
+	cl.Options.APIVersion = "2022-11-01-preview"
```
bbernays, line 18: In the SDK... it sets the API version to be `2022-08-01`

https://github.com/Azure/azure-sdk-for-go/blob/a97519174f72085debd57f0b1189aaa2b62f5f7b/sdk/resourcemanager/nginx/armnginx/deployments_client.go#L287


I am wondering if this is an issue we should raise with the SDK team?

erezrokah: OK so the the code you linked is for v2.0.0 and we're using `v1.0.1` I'll try to update first

bbernays: Yeah in `v1.0.1` it seems like they don't set the API version:
https://github.com/Azure/azure-sdk-for-go/blob/b11b29e0b4bf8ea6112ca3167574a4e7ee69a42a/sdk/resourcemanager/nginx/armnginx/deployments_client.go#LL272-L277C6

I assumed we were using the latest version of the SDK

erezrokah: So updating to `v2` fixes it 🚀 I checked the other PRs and for those resources we're on the latest version.
There's a bigger issue here that our codegen doesn't support versions greater than `v1` (we'll need to fix that)

yevgenypats (approved):

Issue #6625, opened 2023-01-10
Invalid Pks Azure Peering Service Providers


bbernays:
### Describe the Bug

Possible `null` value for PK column

```
ERR resource resolver finished with validation error error="missing primary key on columns: [id]" client=<SUBSCRIPTION_ID> module=azure-src table=azure_peering_service_providers
```

### Expected Behavior

No error

### CloudQuery Version
Source Plugin: Azure 3.1.0
Destination plugin:  postgresql version 2.0.4
CloudQuery Cli version: 2.0.31

PR #6694, opened 2023-01-11
fix(azure-resources): Make `subscription_id, name` PKs of `azure_peering_service_providers` instead of `id`

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6625.
`ID` is always `nil` for this resource. `Name` seems to work well


erezrokah:
> @erezrokah - I think we also need subscription id as a PK otherwise if the same service providers are enabled for multiple subscriptions then we will only see a single record

Good catch, I'll add it 🚀

yevgenypats (approved):

Issue #6696, opened 2023-01-11
feat(azure): Add log message when a table is skipped due to the provider not being registered


erezrokah:
Follow up to https://github.com/cloudquery/cloudquery/issues/5683 and related to https://github.com/cloudquery/cloudquery/pull/5758

When a provider is not registered on Azure we silently skip the table, and the user doesn't have a way of knowing why that resource is not synced:
https://github.com/cloudquery/cloudquery/blob/deba2b1e26455f0219310e9657bb81e3e6d9c181/plugins/source/azure/client/multiplexer.go#L24

We should at least add a log message explaining the table is skipped.

For example, we have a bunch of issues that we could have discovered with the Azure plugin:
![image](https://user-images.githubusercontent.com/26760571/211835735-bc7b7b65-a886-4fa5-83fa-88c37c2f7fce.png)
if those providers were enabled on our playground account

PR #6705, opened 2023-01-11
feat: Log info message when namespace not registered for subscription

hermanschaaf:
- Closes https://github.com/cloudquery/cloudquery/issues/6696


plugins/source/azure/client/multiplexer.go
```diff
@@ -23,6 +23,8 @@ func SubscriptionMultiplexRegisteredNamespace(namespace string) func(schema.Clie
 		for _, subId := range client.subscriptions {
 			if _, ok := client.registeredNamespaces[subId][namespace]; ok {
 				c = append(c, client.withSubscription(subId))
+			} else {
+				client.Logger().Info().Str("subscription_id", subId).Str("namespace", namespace).Msg("Skipping subscription, namespace not registered")
```
erezrokah, line 27:
```suggestion
				client.Logger().Info().Str("subscription_id", subId).Str("namespace", namespace).Msg("Skipping namespace, not registered for subscription")
```

erezrokah (approved):

plugins/source/azure/codegen0/templates/recipe.go.tpl
```diff
@@ -1,25 +0,0 @@
-// Code generated by codegen; DO NOT EDIT.
-package recipes
```
hermanschaaf, line 2: I deleted this file because it looks like it isn't used anywhere anymore

erezrokah (approved):

PR #6707, opened 2023-01-11
feat(shopify): Incremental updates

disq:
~still testing~ 🚀


plugins/source/shopify/docs/tables/shopify_abandoned_checkouts.md
```diff
@@ -1,6 +1,7 @@
 # Table: shopify_abandoned_checkouts
 
 The primary key for this table is **id**.
+It supports incremental syncs.
```
hermanschaaf, line 4: It's optional, but you can also use the `IncrementalKey` column creation option to indicate which column(s) are used as the cursor, and this will then get mentioned in the docs here as well.

hermanschaaf: I'm talking about `IncrementalKey`, a column-level option, which is different from `IsIncremental`, a table-level option.

hermanschaaf: For more context: `IsIncremental` is required, as `overwrite-delete-stale` relies on it to know not to delete stale entries. Column-level `IncrementalKey` is optional and only for documentation purposes right now.

disq: Thanks, added

plugins/source/shopify/internal/shopify/shopify.go
```diff
@@ -39,7 +40,7 @@ type HTTPDoer interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
-func New(log zerolog.Logger, hc HTTPDoer, apiKey, apiSecret, accessToken, shopURL string, maxRetries int64) (*Client, error) {
+func New(log zerolog.Logger, hc HTTPDoer, apiKey, apiSecret, accessToken, shopURL string, maxRetries, pageSize int64) (*Client, error) {
```
hermanschaaf, line 43: Nit: feels like this probably has enough arguments now that it should rather be turned into an options struct, which will help avoid passing the arguments in the wrong order.

plugins/source/shopify/resources/services/checkout/abandoned_checkouts_fetch.go
```diff
@@ -10,18 +13,52 @@ import (
...
+		value, err := cl.Backend.Get(ctx, key, cl.ID())
+		if err != nil {
+			return fmt.Errorf("failed to retrieve state from backend: %w", err)
+		}
+		if value != "" {
+			min, err = time.Parse(time.RFC3339, value)
+			if err != nil {
+				return fmt.Errorf("retrieved invalid state value: %q %w", value, err)
+			}
+			p.Set("updated_at_min", min.Format(time.RFC3339))
```
hermanschaaf, line 31: It looks like this key gets set but never read? Or maybe I'm just overlooking it on the GitHub interface, not sure

plugins/source/shopify/client/client.go
```diff
@@ -72,6 +75,9 @@ func getServiceClient(logger zerolog.Logger, spec *Spec) (*shopify.Client, error
 	if spec.MaxRetries < 1 {
 		spec.MaxRetries = 30
 	}
+	if spec.PageSize < 1 {
```
hermanschaaf, line 78: What is the benefit of having page size be part of the spec, rather than always choosing the largest possible page size?

hermanschaaf (changes requested):

plugins/source/shopify/resources/services/checkout/abandoned_checkouts_fetch.go
disq: It's read by the shopify api endpoint https://shopify.dev/api/admin-rest/2022-10/resources/abandoned-checkouts#get-checkouts?limit=1

plugins/source/shopify/client/client.go
disq: The benefit is actually documented in this PR:
> - `page_size` (integer, optional. Default: 50):
>  Maximum number of items queried each request. Find an optimum value to balance amount of data fetched and requests timing out. Maximum value 250.

Large page sizes tend to time out if you have a large account.

plugins/source/shopify/internal/shopify/shopify.go
disq: Fixed, not pretty but better.

plugins/source/shopify/internal/shopify/shopify.go
```diff
@@ -172,13 +177,11 @@ func (s *Client) retryableRequest(ctx context.Context, edge string) (*http.Respo
...
 
-func (s *Client) GetProducts(ctx context.Context, pageUrl string) (*GetProductsResponse, string, error) {
+func (s *Client) GetProducts(ctx context.Context, pageUrl string, params url.Values) (*GetProductsResponse, string, error) {
 	var ret GetProductsResponse
 
-	const pageSize = 250
-
 	if pageUrl == "" {
-		pageUrl = fmt.Sprintf("admin/api/%s/products.json?limit=%v", APIVersion, pageSize)
+		pageUrl = fmt.Sprintf("admin/api/%s/products.json?limit=%d&%s", APIVersion, s.pageSize, params.Encode())
```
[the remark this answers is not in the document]
disq: Fixed, slight refactor

plugins/source/shopify/internal/shopify/shopify.go
```diff
@@ -64,27 +66,30 @@ func New(log zerolog.Logger, hc HTTPDoer, apiKey, apiSecret, accessToken, shopUR
...
 			}
-			s.log.Debug().Msg("wait complete")
+			log.Debug().Msg("wait complete")
 		}
 
 		r, wait, err := s.retryableRequest(ctx, edge)
 		if err == nil {
 			return r, nil
 		}
+		log.Err(err).Msg("retryable request failed")
```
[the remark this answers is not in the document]
disq: Slight refactor

hermanschaaf (approved):
Nice! ❇️

plugins/source/shopify/client/client.go
hermanschaaf: Ah okay, gotcha 👍

Issue #6354, opened 2023-01-04
Azure Compute: VMs that are part of a scale set are invisible


maikzumstrull:
### Why do you need this resource or plugin? Please describe.
Being able to list _all_ virtual machines in the subscription is fundamental for a cloud inventory tool.

### Describe the solution you'd like
For each VirtualMachineScaleSet, use a VirtualMachineScaleSetVMsClient and retrieve all VMs that are part of that scale set using the "NewListPager - Gets a list of all virtual machines in a VM scale sets" call

The VMs found could either be added to the `azure_compute_virtual_machines` table (more intuitive) or or to a new separate and similar table for VMs that are part of a scale set (closer match to Azure API). Either way, the row for the VM should reference which scale set created it.

### Additional context
Currently, the VirtualMachinesClient's NewListAllPager call is used, which according to the documentation "Lists all of the virtual machines in the specified subscription." That's misleading – the actual behavior is to list only VMs that were directly created as VMs. VMs that are in a scale set can only be inventoried through the scale set that created them. They don't otherwise show up as VMs.

Current behavior (VMs in scale sets are ignored) confirmed on cloudquery `2.0.30` azure source plugin `3.1.0` using an Azure subscription that has both direct VMs and scale set VMs.

PR #6718, opened 2023-01-11
feat(azure-resources): Add Virtual Machine Scale Set VMs

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6354.

~~Still need to test on real resources~~


yevgenypats (approved):
I wonder if we can now delete `codegen0` completely as not really bringing value

erezrokah:
> I wonder if we can now delete `codegen0` completely as not really bringing value

I was thinking the same thing as it's a pain to exclude stuff and we're not adding new things via code gen.

erezrokah:
> I was thinking the same thing as it's a pain to exclude stuff and we're not adding new things via code gen.

I'll do it after this round of fixes so we can get a new release out quickly

PR #6741, opened 2023-01-11
fix(aws): Add PKs To IAM Resources

bbernays:


plugins/source/aws/docs/tables/aws_iam_user_access_keys.md
```diff
@@ -1,6 +1,6 @@
 # Table: aws_iam_user_access_keys
 
-The primary key for this table is **_cq_id**.
+The composite primary key for this table is (**account_id**, **user_arn**, **access_key_id**).
```
candiduslynx, line 3: Can't it be just access key id?

candiduslynx (approved):

yevgenypats (approved):

plugins/source/aws/docs/tables/aws_iam_user_access_keys.md
bbernays: I can't find anywhere that states the `access_key_id` is globally unique... If you can I am happy to change it to be just that

PR #6746, opened 2023-01-12
refactor(azure): Remove codegen

erezrokah:
#### Summary

Removes Azure code gen as the plugin is bootstrapped already and having the code gen only adds friction.

1. Removed all the `// Code generated by codegen; DO NOT EDIT.` comments
2. Merged the namespaces into a single file
3. Merged the tables into a single file (and sorted them)
4. The format changes and linter exclusion are needed as once you remove the `// Code generated by codegen; DO NOT EDIT.` comment the linter runs on those files

`Wait for all failure` is expected since this PR changes more than 300 files


erezrokah:
> Nice! Before we remove this can codegen help with `v2` upgrade or not really and we can do this without it?

I think we'll do that manually, there are only a few `v2`:
image

PR #6747, opened 2023-01-12
fix(aws): AWS S3 resources could have had duplicate entries

candiduslynx:


hermanschaaf:
Is this because these are child relations? I'd have to check the plugin-sdk code again to be sure, but I think we ignore the Multiplexer field on child relations, so there probably was no issue here?

candiduslynx:
Seems like the culprit was AWS (server side): https://github.com/aws/aws-sdk-go-v2/issues/1078

Issue #6600, opened 2023-01-10
Discussion: Align migration logic between destinations


erezrokah:
### Context

Prerequisite for https://github.com/cloudquery/cloudquery/issues/6584 and related to https://discord.com/channels/872925471417962546/1062341270955110501.

Currently our auto migration logic differs between destinations, which leads to confusion on what's breaking, what's auto migratable and what user action is needed on each major version bump or migration failure.

### Current state (based on https://github.com/cloudquery/cloudquery/issues/6043)

| Change type | Breaking? | Reason | Auto migratable
| :--- | :--- | :--- | :--- |
| Adding a column |  `false` | Doesn't break any data consistency | [Yes](https://www.cloudquery.io/docs/advanced-topics/migrations#addition)
| Adding a column to a PK |  `false` | Doesn't break any data consistency | Yes, assuming the data exists and not `null` [we recreate the constraint](https://github.com/cloudquery/cloudquery/blob/6bab35d0546d289bdc57c9c8d741cd046bb79c97/plugins/destination/postgresql/client/migrate.go#L157)
| Removing a column from a PK |  `false` | Assuming the new PK is unique this should be OK | [Yes, we recreate the constraint](https://github.com/cloudquery/cloudquery/blob/6bab35d0546d289bdc57c9c8d741cd046bb79c97/plugins/destination/postgresql/client/migrate.go#L157)
| Removing a column |  `true` | Column will be `null` in future syncs | [Yes](https://www.cloudquery.io/docs/advanced-topics/migrations#removal), we don't drop the column
| Changing composite PK columns order |  `false` | May change the partitioning, but otherwise, harmless | [Yes, we recreate the constraint](https://github.com/cloudquery/cloudquery/blob/6bab35d0546d289bdc57c9c8d741cd046bb79c97/plugins/destination/postgresql/client/migrate.go#L157)
[...]

PR #6749, opened 2023-01-12
fix(postgres): Error on new PK column

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/6600.

This changes the Postgres migration logic to fail when a new column is added and it's a PK.
This is the same as the SQLite plugin:
https://github.com/cloudquery/cloudquery/blob/bfd1494fa464f2cafa4b1d04bfd54ac4726319ad/plugins/destination/sqlite/client/migrate.go#L93

I think we can improve this as only error if the table exists and is not empty as for empty tables this shouldn't be an issue

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -123,7 +123,7 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table) erro
 
 			sql := "alter table " + tableName + " add column " + columnName + " " + columnType
 			if col.CreationOptions.PrimaryKey {
-				reCreatePrimaryKeys = true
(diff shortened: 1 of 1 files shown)


yevgenypats (changes requested):
The change looks good but before we do that I think we might want to introduce the `--force` on the SDK/destination plugin level as we do change PKs quite a lot and most of the time I think it worked so I would worry of adding a blocker without introducing the easy workaround that most people basically use now by default?

yevgenypats (approved):

erezrokah:
I'll merge this after we add force migration support like we did in SQLite https://github.com/cloudquery/cloudquery/pull/6763

erezrokah:
Going to close this and do it once I add `migrate_mode: forced`

PR #6750, opened 2023-01-12
fix(postgres): Error on column type change

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/6600.
This aligns Postgres with other destinations to fail on type changes instead of dropping and adding the column.

I'll open a separate PR to update the docs with https://github.com/cloudquery/cloudquery/pull/6749 in mind


erezrokah:
Before merging this we should support forced migration for Postgres like https://github.com/cloudquery/cloudquery/pull/6763

erezrokah:
Going to close this and do it once I add `migrate_mode: forced`

PR #6754, opened 2023-01-12
fix(aws)!: Add PK to `aws_s3_bucket_grants` table

candiduslynx:
BEGIN_COMMIT_OVERRIDE
fix!: introduce `aws_s3_bucket_grants` explicit primary key `(bucket_arn, grantee_id)`. If you've previously synced this table you'll need to drop it for the PK change (migration) to succeed.
END_COMMIT_OVERRIDE


hermanschaaf:
@candiduslynx Since this is a breaking change, could you please update the PR description to explain what this is fixing and what users will need to do to migrate? E.g. see this previous breaking change message https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/CHANGELOG.md#-breaking-changes-5 and the corresponding PR text: https://github.com/cloudquery/cloudquery/pull/4382#issue-1445504880

candiduslynx:
@hermanschaaf updated the desc

hermanschaaf (approved):

plugins/source/aws/resources/services/s3/bucket_cors_rules.go
```diff
@@ -13,7 +13,6 @@ func BucketCorsRules() *schema.Table {
 		Description: `https://docs.aws.amazon.com/AmazonS3/latest/API/API_CORSRule.html`,
 		Resolver:    fetchS3BucketCorsRules,
 		Transform:   transformers.TransformWithStruct(&types.CORSRule{}),
-		Multiplex:   client.AccountMultiplex,
```
erezrokah, line 16: What's the reason for removing the `AccountMultiplex`es? Is it related to the the bug fix?

hermanschaaf: I think it's from a previous hypothesis that this was causing duplicates, but in reality it doesn't matter: `Multiplex` is completely ignored for child tables

yevgenypats:
Adding context. This seems to be related to https://github.com/aws/aws-sdk-go-v2/issues/1078

PR #6759, opened 2023-01-12
feat(sqlite): Collect and report migration errors before starting the migration

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/6600. ~~Still in draft as I need to test it 😄~~ 
At the moment we migrate the tables and columns one by one and stop if we can't migrate due to potential data loss.
This has a few issues
- We can leave the DB in partial migration state
- We only report one error, so if there are multiple errors the user has to run the migration multiple times to know what to fix

### Notes
 
 - Didn't add the force flag to keep this PR small as possible. Once we add `--force` the change will be simply to ignore the errors and keep on migrating (e.g. drop/add columns/tables)
 - SQLite is missing the logic Postgres has to [recreate primary keys on existing tables](https://github.com/cloudquery/cloudquery/blob/59c32dc65c9df6870c4d973893a7e722c45188fd/plugins/destination/postgresql/client/migrate.go#L109). This is unrelated to this PR so didn't add it, but would be good to align that in a future PR
 - Might be worth moving some of the code to the SDK, but won't do that yet until I see how it looks in other plugins


plugins/destination/sqlite/client/migrate.go
```diff
@@ -38,70 +39,160 @@ func (i *tableInfo) getColumn(name string) *columnInfo {
...
 
-	for _, col := range table.Columns {
-		columnName := col.Name
-		columnType := c.SchemaTypeToSqlite(col.Type)
-		sqliteColumn := info.getColumn(columnName)
-
-		switch {
-		case sqliteColumn == nil:
-			c.logger.Debug().Str("table", table.Name).Str("column", col.Name).Msg("Column doesn't exist, creating")
-			sql := "alter table \"" + table.Name + "\" add column \"" + columnName + "\" \"" + columnType + `"`
```
erezrokah, line 91: We only modify columns when we add a new one and it's not a PK. I kept that logic

plugins/destination/sqlite/client/migrate.go
```diff
@@ -38,70 +39,160 @@ func (i *tableInfo) getColumn(name string) *columnInfo {
...
+func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
+	schemaChanges, err := c.getSchemaChanges(ctx, tables)
+	if err != nil {
+		return err
+	}
 
-	if info, err = c.getTableInfo(table.Name); err != nil {
-		return fmt.Errorf("failed to get table %s columns types: %w", table.Name, err)
+	migrationErrors := getMigrationErrors(schemaChanges)
+	if len(migrationErrors) > 0 {
```
erezrokah, line 162: When we add the `--force` flag, we'll log the errors and continue (the code to handle the errors, e.g. dropping a table is not here yet, I'll add that when I add the `--force` flag)

hermanschaaf (approved):

erezrokah:
Thanks for the review @hermanschaaf. I'll merge this as it's blocking another PR cc @yevgenypats if you have concerns I can follow up

PR #6763, opened 2023-01-12
feat(sqlite): Support force migration

erezrokah:
#### Summary

To be merged after https://github.com/cloudquery/cloudquery/pull/6759, and needs https://github.com/cloudquery/plugin-sdk/pull/604

### DIFF
--- a/plugins/destination/sqlite/client/migrate.go
+++ b/plugins/destination/sqlite/client/migrate.go
@@ -60,7 +60,7 @@ type migrationMessage struct {
 type migrationsMessages []migrationMessage
 
 func (m migrationsMessages) Errors() []string {
-	var errs []string
+	errs := make([]string, 0, len(m))
 	for _, msg := range m {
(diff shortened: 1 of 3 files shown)


plugins/destination/sqlite/client/migrate.go
```diff
@@ -38,70 +39,192 @@ func (i *tableInfo) getColumn(name string) *columnInfo {
...
+			}
+		}
+	}
+	return errors
+}
 
-	if info, err = c.getTableInfo(table.Name); err != nil {
-		return fmt.Errorf("failed to get table %s columns types: %w", table.Name, err)
+// This is the responsibility of the CLI of the client to lock before running migration
+func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
```
erezrokah, line 155: We could move this part to the SDK and pass in functions to drop/add columns/tables if we need to

PR #6772, opened 2023-01-12
fix(aws): Correctly fill in `grantee_id` column to `aws_s3_bucket_grants`

candiduslynx:
* Add `grantee_type` column to `aws_s3_bucket_grants`
* Correctly fill in `grantee_id` column to `aws_s3_bucket_grants` based on `grantee_type`


hermanschaaf (approved):
LGTM. We can release this as a patch, but maybe we should then also deprecate v11.0.0 (can do this tomorrow)

PR #6776, opened 2023-01-12
feat(stripe): Add events

disq:


plugins/source/stripe/resources/services/events/events.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			lp.Created = &vi
+		}
+	}
+
+	it := cl.Services.Events.List(lp)
+	for it.Next() {
+		data := it.Event()
+		lp.Created = client.MaxInt64(lp.Created, &data.Created)
+		res <- data
```
hermanschaaf, line 69: Ideally we should occasionally update the state while the sync is happening as well, in case it fails or stops halfway through

hermanschaaf (approved):
Nice, that was quick! 🏃

plugins/source/stripe/resources/services/events/events.go
bbernays: I am not sure we can do that... Just because the data was grabbed doesn't mean it was persisted in the datastore...

PR #6786, opened 2023-01-13
feat(gcp-resources): Add vision

shimonp21:
Caveats:
- Generated only most-imporant/top-level tables (skipped productsets).
- no programatic way to get list of supported locations. I copy-pasted them from the APIs error message:
```
2:56PM ERR table resolver finished with error error="rpc error: code = InvalidArgument desc = Location not supported: 'europe-west3'. Must be one of 'us-west1', 'us-east1', 'asia-east1', 'europe-west1'" client=cq-playground module=gcp-src table=gcp_vision_product_sets
```


plugins/source/gcp/resources/services/vision/reference_image.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/cloudquery/plugin-sdk/transformers"
+	"github.com/cloudquery/plugins/source/gcp/client"
+)
+
+func ReferenceImages() *schema.Table {
+	return &schema.Table{
+		Name:        "gcp_vision_product_reference_images",
+		Description: `https://cloud.google.com/vision/docs/reference/rest/v1/projects.locations.products.referenceImages`,
+		Resolver:    fetchReferenceImages,
+		// The list of locations was copied by me from the APIs error message. Yuck.
```
erezrokah, line 15:
```suggestion
```

erezrokah (approved):
Looks great, we can remove one code comment I think. I like the usage of a multiplexer here and will port https://github.com/cloudquery/cloudquery/blob/9ecc4d7dfb5767311cf79ee4c19dd6ecfc467f89/plugins/source/gcp/resources/services/cloudiot/device_registries_fetch.go#L29 once this is merged

PR #6789, opened 2023-01-13
feat: Add scaffold CLI

yevgenypats:
Right now this only supports `source` scaffolding and will support `destination` in a follow-up PR. This also runs some basic sanity tests the source scaffold compiles.


scaffold/cmd/root.go
```diff
@@ -0,0 +1,37 @@
...
+
+Open source data integration at scale.
+
+Find more information at:
+	https://www.cloudquery.io`
+)
+
+func NewCmdRoot() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "cloudquery",
```
hermanschaaf, line 20:
```suggestion
		Use:     "scaffold",
```

scaffold/cmd/templates/source/go.mod.tpl
```diff
@@ -0,0 +1,8 @@
+module github.com/{{.Org}}/cq-source-{{.Name}}
```
hermanschaaf, line 1: Potential future improvement: Maybe we should run `go mod tidy` on the directory after generating so that the `go.sum` gets generated as well. Otherwise you get a slightly confusing error the first time you try to run it

yevgenypats: we can, or we can bake `go.mod` in. The issue with running `go mod` that it might not be available and it's like a different step.

hermanschaaf (approved):
Nice, I like it a lot! ✨

Issue #6790, opened 2023-01-13
Fix: goreleaser for scaffold


yevgenypats:
I've just added a scaffold cli. Will need also to add a `.goreleaser` - https://github.com/cloudquery/cloudquery/pull/6789

PR #6801, opened 2023-01-15
chore(ci): Add scaffold Go releaser config

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6790.
Mostly the same as our CLI, but without the docker part. Binary name is `cq-scaffold` as `scaffold` is a bit generic.
Also kept the saving of the version file part as we might need it for the Website


yevgenypats (approved):
🪄

PR #6804, opened 2023-01-15
fix(website): Update link to TimescaleDB docs

erezrokah:
#### Summary

TimescaleDB changed the docs and broke one of our links to them. This PR fixes it


yevgenypats (changes requested):
I think let's deprecate this blog-post, it will be only confusing as we don't support Timescale and I don't think we are gonna support it anytime soon.

erezrokah:
> I think let's deprecate this blog-post, it will be only confusing as we don't support Timescale and I don't think we are gonna support it anytime soon.

It has a deprecation header https://github.com/cloudquery/cloudquery/pull/6804/files#diff-cfb6ce8fefb857cb15c8607729de69714b1d4a7bb016036b84d76c0570787abdL15.
We keep the olds posts to avoid 404s from external links

yevgenypats:
Maybe let's just redirect to destination pages: https://www.cloudquery.io/docs/plugins/destinations/overview as the content is completely stale so we don't need to maintain it and any links inside?

erezrokah:
> Maybe let's just redirect to destination pages: https://www.cloudquery.io/docs/plugins/destinations/overview as the content is completely stale so we don't need to maintain it and any links inside?

Good point. This PR is indented as a quick fix to our broken links CI.
Maybe what we do with deprecated blog posts is another discussion?

erezrokah:
I tried doing https://nextra.site/docs/docs-theme/page-configuration#hidden-routes but for some reason couldn't get it to work (maybe we need a dep update).

yevgenypats (approved):

disq (approved):

PR #6807, opened 2023-01-15
fix(gcp): Add `Location` to client `ID` when present

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/6786. We should add the location to the client ID as it has to be unique


plugins/source/gcp/client/client.go
```diff
@@ -88,6 +88,9 @@ func (c *Client) ID() string {
 	if c.OrgId != "" {
 		return "org:" + c.OrgId
 	}
+	if c.Location != "" {
+		return "project:" + c.ProjectId + ":location:" + c.Location
+	}
 	return c.ProjectId
```
yevgenypats, line 94: Nit: Can we do that the last option will be `project:` + c.ProjectId so at least we only have two types of IDs

yevgenypats (changes requested):
Small nit.

plugins/source/gcp/client/client.go
erezrokah: Sure, done in https://github.com/cloudquery/cloudquery/pull/6807/commits/3dff1260b044228bd2cc9c43d51f046c10c75904

yevgenypats (approved):

Issue #6753, opened 2023-01-12
bug(ci): GitHub Bot should not see Incremental Key changes as breaking


hermanschaaf:
Incremental table keys are indicated as such in the table docs, but a change in these should not be considered a breaking change. The bot that comments on the changed columns for each table should ignore changes in Incremental Keys.

See example PR here: https://github.com/cloudquery/cloudquery/pull/6707

![Screenshot 2023-01-12 at 10 38 45](https://user-images.githubusercontent.com/1121616/212045154-7f0863b2-df76-48f6-a5f3-3e060e420afb.png)

PR #6811, opened 2023-01-15
feat(table-diff): Add incrmental key support

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6753

### DIFF
--- a/scripts/table_diff/changes/changes.go
+++ b/scripts/table_diff/changes/changes.go
@@ -20,9 +20,24 @@ type change struct {
 	Breaking bool   `json:"breaking"`
 }
 
+type columnType int
+
+const (
+	columnTypePK columnType = 1 << iota
+	columnTypeIncremental
+)
+
 type column struct {
-	dataType string
-	pk       bool
+	dataType   string
+	columnType columnType
+}
+
+func (c column) pk() bool {
+	return c.columnType&columnTypePK != 0
+}
+
+func (c column) incremental() bool {
+	return c.columnType&columnTypeIncremental != 0
 }
 
 func backtickStrings(strings ...string) []interface{} {
@@ -33,10 +48,10 @@ func backtickStrings(strings ...string) []interface{} {
 	return backticked
 }
 
-func parseColumnChange(line string) (name string, dataType string, pk bool) {
+func parseColumnChange(line string) (name string, dataType string, columnType columnType) {
 	match := columnRegex.FindStringSubmatch(line)
 	if match == nil {
-		return "", "", false
+		return "", "", columnType
 	}
 	result := make(map[string]string)
 	for i, name := range columnRegex.SubexpNames() {
@@ -44,8 +59,14 @@ func parseColumnChange(line string) (name string, dataType string, pk bool) {
 			result[name] = match[i]
 		}
 	}
-	cleanName := strings.TrimSuffix(result["name"], " (PK)")
-	return cleanName, result["dataType"], result["name"] != cleanName
+	if strings.Contains(result["name"], " (PK)") {
+		columnType |= columnTypePK
+	}
+	if strings.Contains(result["name"], " (Incremental Key)") {
+		columnType |= columnTypeIncremental
+	}
+	cleanName := strings.Split(result["name"], " (")[0]
+	return cleanName, result["dataType"], columnType
 }
 
 func parsePKChange(line string) (names []string) {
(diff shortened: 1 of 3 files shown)


scripts/table_diff/changes/changes.go
```diff
@@ -20,9 +20,24 @@ type change struct {
 	Breaking bool   `json:"breaking"`
 }
 
+type columnType int
+
+const (
+	columnTypePK columnType = 1 << iota
```
erezrokah, line 26: Based on https://github.com/cloudquery/plugin-sdk/blob/37254f3ae99eb02bec824a211dd9e3b7869b55dd/plugins/source/templates/table.md.go.tpl#L41 columns can be both PK and Incremental (though I don't think we have those yet), so using a bit mask

scripts/table_diff/changes/changes_test.go
```diff
@@ -311,6 +313,40 @@ func Test_getChanges(t *testing.T) {
...
+			name:         "Should handle incremental column changes",
+			diffDataFile: "testdata/pr_6707_diff.txt",
+			wantChanges: []change{
+				{
+					Text:     "Table `shopify_abandoned_checkouts`: incremental sync support added to column `updated_at`",
+					Breaking: false,
+				},
+				{
+					Text:     "Table `shopify_customers`: incremental sync support removed from column `created_at`",
+					Breaking: true,
```
erezrokah, line 326: I fiddled with the git diff of the PR data to simulate a breaking change

scripts/table_diff/changes/changes.go
```diff
@@ -105,19 +126,33 @@ func getColumnChanges(file *gitdiff.File, table string) (changes []change) {
...
 
-			if !addedColumn.pk && deletedColumn.pk {
+			if !addedColumn.pk() && deletedColumn.pk() {
 				changes = append(changes, change{
 					Text:     fmt.Sprintf("Table %s: primary key constraint removed from column %s", backtickStrings(table, deletedName)...),
 					Breaking: false,
 				})
 			}
+
+			if addedColumn.incremental() && !deletedColumn.incremental() {
```
erezrokah, line 143: Please let me know if these messages make sense

scripts/table_diff/changes/changes.go
```diff
@@ -105,19 +126,33 @@ func getColumnChanges(file *gitdiff.File, table string) (changes []change) {
...
+				changes = append(changes, change{
+					Text:     fmt.Sprintf("Table %s: incremental sync support added to column %s", backtickStrings(table, deletedName)...),
```
hermanschaaf, line 145:
```suggestion
					Text:     fmt.Sprintf("Table %s: column %s added to cursor for incremental syncs", backtickStrings(table, deletedName)...),
```

I think it was fine, but this might be a little clearer

scripts/table_diff/changes/changes.go
```diff
@@ -105,19 +126,33 @@ func getColumnChanges(file *gitdiff.File, table string) (changes []change) {
...
+					Breaking: false,
+				})
+			}
+
+			if !addedColumn.incremental() && deletedColumn.incremental() {
+				changes = append(changes, change{
+					Text:     fmt.Sprintf("Table %s: incremental sync support removed from column %s", backtickStrings(table, deletedName)...),
```
hermanschaaf, line 152:
```suggestion
					Text:     fmt.Sprintf("Table %s: column %s removed from cursor for incremental syncs", backtickStrings(table, deletedName)...),
```

scripts/table_diff/changes/changes.go
```diff
@@ -105,19 +126,33 @@ func getColumnChanges(file *gitdiff.File, table string) (changes []change) {
...
 				changes = append(changes, change{
 					Text:     fmt.Sprintf("Table %s: primary key constraint removed from column %s", backtickStrings(table, deletedName)...),
 					Breaking: false,
 				})
 			}
+
+			if addedColumn.incremental() && !deletedColumn.incremental() {
+				changes = append(changes, change{
+					Text:     fmt.Sprintf("Table %s: incremental sync support added to column %s", backtickStrings(table, deletedName)...),
+					Breaking: false,
```
hermanschaaf, line 146: We may want to make this breaking, as there's a chance an update to the cursor or a full sync will be necessary on upgrade. Debatable, because it could simply be for documentation purposes or backwards-compatibility could be handled in the code. But maybe this comment can call it to attention and then we can decide on a case-by-case basis?

erezrokah: Sure, let's make it breaking and we can change it it doesn't make sense

PR #6834, opened 2023-01-16
fix: Remove dead link from deprecated blog post

shimonp21:


erezrokah:
See https://github.com/cloudquery/cloudquery/pull/6804 it uses the new correct link

PR #6840, opened 2023-01-16
chore(docs): Fix broken link

candiduslynx:


erezrokah:
See https://github.com/cloudquery/cloudquery/pull/6804

candiduslynx:
drop in favor of #6840

PR #6841, opened 2023-01-16
fix(k8s): Fix inet-type panic

shimonp21:
Make our resolvers ignore `None` strings.

The other way to fix is to just leave the incoming field as a `string`. This way is less breaking for users.


disq (approved):

plugins/source/k8s/client/helpers.go
```diff
@@ -52,34 +52,81 @@ func ResolveContext(_ context.Context, meta schema.ClientMeta, r *schema.Resourc
...
+func StringToInetArrayPathResolver(path string) schema.ColumnResolver {
+	return func(_ context.Context, meta schema.ClientMeta, r *schema.Resource, c schema.Column) error {
+		value := funk.Get(r.Item, path, funk.WithAllowZero())
+
+		stringArrayValue, ok := value.([]string)
+		if !ok {
+			return r.Set(c.Name, nil)
+		}
+
+		filteredStringArray := []string{}
```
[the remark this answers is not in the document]
shimonp21: Yeah, this solution doesn't mirror the API exactly. However, I believe this is "good enough".

- Semantically, `k8s` state that there may be a difference between `None` and `""`. https://github.com/kubernetes/api/blob/59fcd23597fd090dba6b7e903eb0a8c9e8efb0a6/core/v1/types.go#L4529-L4544

- for the "non-array" version, we can't distinguish between `None` and `""` - we only have one "null" to work with. Thus, not sure it's super-important to distinguish them in the array-version (if we don't do it everywhere).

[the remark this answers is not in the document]
shimonp21: @hermanschaaf No solution is perfect here I think (turning the `inets` into strings in order to be 100% compliant to the API is also breaking and annoying).

I could go either way honestly, but prefer the silent/non-breaking one..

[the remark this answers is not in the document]
hermanschaaf: I guess what I'm saying is that we seem to be dropping both the `""` and `"None"` IPs in the array version, rather than storing them as `nil` values in the array, which I believe our type system now allows. I'm not too worried about distinguishing between `""` and `"None"`

plugins/source/k8s/client/helpers.go
```diff
@@ -52,34 +52,85 @@ func ResolveContext(_ context.Context, meta schema.ClientMeta, r *schema.Resourc
...
+		stringArrayValue, ok := value.([]string)
+		if !ok {
+			return r.Set(c.Name, nil)
+		}
+
+		sanitized := []*string{}
+
+		for _, stringValue := range stringArrayValue {
+			if stringValue != "" && stringValue != "None" {
+				sanitized = append(sanitized, &stringValue)
```
candiduslynx, line 85: sanitizedValue has the same address, use idx & link to slice elem

plugins/source/k8s/client/helpers.go
```diff
@@ -52,34 +52,85 @@ func ResolveContext(_ context.Context, meta schema.ClientMeta, r *schema.Resourc
...
```
candiduslynx, line 124: same as prev comment

candiduslynx (changes requested):

hermanschaaf (approved):

yevgenypats:
Seems tests are failing

shimonp21:
Yeah, the failing tests was the linter complaining about the for-loop-address issue Alex found. Fixed now :)

plugins/source/k8s/client/helpers.go
```diff
@@ -52,34 +52,85 @@ func ResolveContext(_ context.Context, meta schema.ClientMeta, r *schema.Resourc
...
+
+		sanitized := []*string{}
+
+		for i := range stringArrayValue {
+			if stringArrayValue[i] != "" && stringArrayValue[i] != "None" {
+				sanitized = append(sanitized, &stringArrayValue[i])
+			} else {
+				sanitized = append(sanitized, nil)
+			}
+		}
```
candiduslynx, line 89: Actually, it could be fixed even easier: https://go.dev/play/p/p6YgBjZXixI
```suggestion
		for _, v := range stringArrayValue {
			v := v
			if v == "" || v == "None" {
				sanitized = append(sanitized, nil)
				continue
			}
			sanitized = append(sanitized, &v)
		}
```

plugins/source/k8s/client/helpers.go
```diff
@@ -52,34 +52,85 @@ func ResolveContext(_ context.Context, meta schema.ClientMeta, r *schema.Resourc
...
```
candiduslynx, line 128: Actually, it could be fixed even easier: https://go.dev/play/p/p6YgBjZXixI
```suggestion
		for _, v := range stringArrayValue {
			v := v
			if v == "" || v == "None" {
				sanitized = append(sanitized, nil)
				continue
			}
			sanitized = append(sanitized, &v)
		}
```

candiduslynx (approved):

PR #6842, opened 2023-01-16
feat(cli): Print and log sources and destinations versions

erezrokah:
#### Summary

To go with https://github.com/cloudquery/plugin-sdk/pull/609.
Examples 
image
image


cli/cmd/sync2.go
```diff
@@ -15,14 +15,14 @@ import (
 
 func syncConnectionV2(ctx context.Context, cqDir string, sourceClient *clients.SourceClient, sourceSpec specs.Source, destinationsSpecs []specs.Destination, uid string, noMigrate bool) error {
 	var err error
-	destinationNames := make([]string, len(destinationsSpecs))
+	destinationStrings := make([]string, len(destinationsSpecs))
```
erezrokah, line 18: zerolog supports `Stringers` and `Stringer`, however `Stringers` is not very useful as we need to create that array manually (we can't cast to an array directly)

hermanschaaf:
@erezrokah There were a couple of CI errors reported here, FYI

erezrokah:
> @erezrokah There were a couple of CI errors reported here, FYI

👍 Those should be fixed after https://github.com/cloudquery/plugin-sdk/pull/609 as we need the `String()` method on the specs

cli/cmd/sync2.go
```diff
@@ -44,50 +44,50 @@ func syncConnectionV2(ctx context.Context, cqDir string, sourceClient *clients.S
...
 		} else {
-			fmt.Printf("Source %s will migrate and sync %d %s.\n", sourceSpec.Name, tableCount, word)
+			fmt.Printf("Source %s will migrate and sync %d %s.\n", sourceSpec.VersionString(), tableCount, word)
 		}
 	}
 
 	if !noMigrate {
-		fmt.Println("Starting migration for:", sourceSpec.Name, "->", sourceSpec.Destinations)
-		log.Info().Str("source", sourceSpec.Name).Strs("destinations", sourceSpec.Destinations).Msg("Start migration")
+		fmt.Printf("Starting migration for:%s->%s\n", sourceSpec.VersionString(), destinationStrings)
```
erezrokah, line 54: `Printf` is less error prone as our linter errors if you use `%s` and don't pass a string

erezrokah:
@hermanschaaf this should be ready for another review after using the new SDK

cli/cmd/sync2.go
hermanschaaf, line 54:
```suggestion
		fmt.Printf("Starting migration for: %s -> %s\n", sourceSpec.VersionString(), destinationStrings)
```

hermanschaaf (approved):
Looks good, let's just fix the spacing in that one message before merging :)

PR #6844, opened 2023-01-16
fix(aws)!: Add region to `aws_backup_global_settings` primary key

disq:
~~Few things I don't understand:~~

~~- How is it global when it's region based?~~
~~- If it's not region based, just asks for a region to query the same settings from, then how come we have it as a region multiplexed resource?~~

https://docs.aws.amazon.com/aws-backup/latest/devguide/API_DescribeGlobalSettings.html


candiduslynx (approved):

erezrokah:
Please note this is a breaking change as `region` is missing on existing data

hermanschaaf (approved):

bbernays:
Looking at the data it is only about account level functionalities, and because accounts are regionless resources this should also be global

disq:
> Looking at the data it is only about account level functionalities, and because accounts are regionless resources this should also be global

Updated to use AccountMultiplex

candiduslynx (approved):

bbernays:
I think we have to test the change in multiplexer out because there are only certain regions where you can make this call...

disq:
> I think we have to test the change in multiplexer out because there are only certain regions where you can make this call...

Updated to use a supported region... Could also save the region. Also, currently there's no way to force it to a specific region?

plugins/source/aws/client/multiplexers.go
```diff
@@ -41,6 +41,26 @@ func AccountMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
...
+						continue
+					}
+					l = append(l, client.withPartitionAccountIDAndRegion(partition, accountID, region))
+					break
+				}
+			}
+		}
+		return l
+	}
+}
```
bbernays, line 62: What happens if user specifies a region in the config?

disq: Same thing that happens if the user specifies a region in the config with other region-aware multiplexers 🥲  This is a copy from the multiplexer below, with the single `break` statement thrown in.

bbernays:
After discussing with @candiduslynx, the most simple and straight forward way of ensuring we get the data is to add the `region` as a PK, any other solutions would need extensive documentation to ensure edge cases around regional availability and IAM permissions are understood

disq:
> After discussing with @candiduslynx, the most simple and straight forward way of ensuring we get the data is to add the `region` as a PK, any other solutions would need extensive documentation to ensure edge cases around regional availability and IAM permissions are understood

Reverted last 3 commits to get it where we want it.

candiduslynx (approved):

PR #6846, opened 2023-01-16
fix(AWS)!: Add region to zaws_docdb_engine_versions` primary key

candiduslynx:


bbernays:
`aws_rds_engine_versions`, `aws_elasticache_engine_versions`, and `aws_docdb_engine_versions` should all have the exact same structure

PR #6848, opened 2023-01-16
fix(sqlite-migrate): Use `int` when getting `pk` column information

erezrokah:
#### Summary

SQLite migration always fails on tables that have multiple PKs. The reason is that `pk` in `table_info` is either `0` for non PKs or the 1 based index of the column within the PKs (usually 1 for a single PK, but can be 2, 3, etc.)
See https://www.sqlite.org/pragma.html#pragma_table_info.

If you remove the fix you'll get the following error:
```
failed to get table table_1 columns types: sql: Scan error on column index 5, name "pk": sql/driver: couldn't convert 2 into type bool
```


shimonp21 (approved):
Tested manually - Approving :)

PR #6850, opened 2023-01-16
fix(aws): Add `region` to `aws_ram_resource_types` primary key

candiduslynx:
non-breaking as the field was already there


hermanschaaf (approved):

plugins/source/aws/docs/tables/aws_ram_resource_types.md
```diff
@@ -13,7 +13,7 @@ The composite primary key for this table is (**account_id**, **resource_type**,
 |_cq_id|UUID|
 |_cq_parent_id|UUID|
 |account_id (PK)|String|
-|region|String|
+|region (PK)|String|
+|resource_region_scope|String|
```
bbernays, line 17: I think this needs to be a PK...

plugins/source/aws/resources/services/ram/resource_types.go
```diff
@@ -12,37 +12,21 @@ func ResourceTypes() *schema.Table {
...
-				Name:     "service_name",
-				Type:     schema.TypeString,
-				Resolver: schema.PathResolver("ServiceName"),
-				CreationOptions: schema.ColumnCreationOptions{
-					PrimaryKey: true,
-				},
+				Name:            "region",
+				Type:            schema.TypeString,
+				Resolver:        client.ResolveAWSRegion,
+				CreationOptions: schema.ColumnCreationOptions{PrimaryKey: true},
```
bbernays, line 29: What if a service has a global + regional resources (WAF)?

bbernays (changes requested):

plugins/source/aws/docs/tables/aws_ram_resource_types.md
candiduslynx: It's the value, not the key IMO

plugins/source/aws/resources/services/ram/resource_types.go
bbernays: But both can be in the same region... So would create a duplicate

candiduslynx: I assume waf & waf regional will have different `service_name`

candiduslynx:
@bbernays I believe that the global vs regional is covered in this [doc](https://docs.aws.amazon.com/ram/latest/userguide/working-with-regional-vs-global.html).
Judging by the text, the resource type can be either regional or global, but not both.

bbernays:
@candiduslynx So based on that documentation, the region should be `nil`/`null` for global resources, while our implementation will **always** put a region because it is defined as being the region where the `ListResourceTypes` call was made

candiduslynx:
After a deeper discussion with @bbernays decided to only add `region` for now, as it's _currently_ correct. However, that _might_ change in the future if AWS wishes so.

bbernays (approved):

PR #6853, opened 2023-01-16
feat(aws): Remove more codegen

disq:
No more `"codegen"` in resources/ now.


erezrokah (approved):
Looks good, I think this is a `feat` as it adds new columns (I'm assuming we generated the tables from the wrong struct?)

disq:
> Looks good, I think this is a `feat` as it adds new columns (I'm assuming we generated the tables from the wrong struct?)

No, it's not a wrong struct, just that `alarm_arn` was in skip fields before, but now as a design decision we don't skip those fields and prefer to have duplicate arn columns (same with all other transformer generated resources).

yevgenypats (approved):

erezrokah:
> No, it's not a wrong struct, just that `alarm_arn` (and `security_profile_arn`) was in skip fields (in codegen) before, but now as a design decision we don't skip those fields and prefer to have duplicate arn columns (same with all other transformer generated resources).

Cool, thanks for the context. But should this still be a `feat`? It's a user facing change

PR #6887, opened 2023-01-17
fix(azure): Duplicate entries for `azure_subscription_subscriptions`

candiduslynx:


disq (approved):

plugins/source/azure/client/client.go
```diff
@@ -43,9 +43,10 @@ func (c *Client) discoverSubscriptions(ctx context.Context) error {
 			return err
 		}
 		for _, sub := range page.Value {
+			// we record all returned values, even disabled
+			c.SubscriptionsObjects = append(c.SubscriptionsObjects, sub)
```
yevgenypats, line 47: Im not sure we want disabled projects as it will increase the load on APIs significantly and will slow the syncing.

plugins/source/azure/resources/services/subscription/subscriptions_fetch.go
```diff
@@ -13,6 +13,7 @@ func fetchSubscriptions(ctx context.Context, meta schema.ClientMeta, parent *sch
 	// if we already fetched it then no need to run it again.
 	if len(cl.SubscriptionsObjects) != 0 {
 		res <- cl.SubscriptionsObjects
+		return nil
```
yevgenypats, line 16: Nice.

yevgenypats (changes requested):
Nice catch. One comment.

plugins/source/azure/client/client.go
candiduslynx: This object is used only in fetch. If this object is nil the fetch is still returning all values no matter what state the subscription is in.

plugins/source/azure/resources/services/subscription/subscriptions_fetch.go
candiduslynx: 100%

yevgenypats (approved):

PR #6900, opened 2023-01-17
chore(ci): Report if table is in code but not in docs

erezrokah:
#### Summary

Kind of follow up to https://github.com/cloudquery/cloudquery/pull/6677.
Since there's no code gen it's easy to miss creating new tables, but not adding to the tables list/relations.

This PR adds a script that looks for tables declarations in `resources/services` and checks they have matching docs (and vice versa). I had to refactor the test and terraform plugins a bit for this to work.

Example:
image


plugins/source/terraform/resources/services/tfdata.go
```diff
@@ -52,93 +50,7 @@ func TFData() *schema.Table {
 				Description: "The \"lineage\" is a unique ID assigned to a state when it is created",
 			},
 		},
-		Relations: []*schema.Table{
```
erezrokah, line 55: Changes here are moving the tables to dedicated files

scripts/missing_tables/main.go
```diff
@@ -0,0 +1,82 @@
...
+		content, err := ioutil.ReadFile("../../" + tableFile)
+		if err != nil {
+			panic(err)
+		}
+		contentString := string(content)
+		if !strings.Contains(contentString, "*schema.Table") {
+			continue
+		}
+
+		tableNameRegex := regexp.MustCompile(`Name\:.*?"(.*?)",`)
```
hermanschaaf, line 59: I think maybe this regex could also catch column names by accident?

hermanschaaf: To be absolutely sure we'd probably need to use the AST 😅  But another thing I've done that's a bit easier is to keep track of the curly brackets in a stack, and make sure the `Name` entry is at the correct level (right after the `schema.Table` definition)

hermanschaaf:
Nice! This issue has bitten me a few times already since we removed codegen

scripts/missing_tables/main.go
erezrokah: It doesn't have false positives at the moment. It works since it reports the first match.
I didn't want to AST 😄 as it can get a bit complex/slow.
I think I can make the regular expression a bit more strict (to match first name after table declaration) WDYT?

erezrokah: Something like https://github.com/cloudquery/cloudquery/pull/6900/commits/447fdfd77d1507926b2c5a7ec221601f30d9ca74

erezrokah:
@hermanschaaf did you mean to approve the PR? Your review is still pending 🙏

hermanschaaf (approved):
👍

PR #6904, opened 2023-01-17
chore(website): Update main README

yevgenypats:

PR #6905, opened 2023-01-17
feat(azure-resources): Add missing security tables

erezrokah:
#### Summary

Based on https://github.com/cloudquery/cloudquery/pull/6904#issuecomment-1385681261

cc @amanenk


erezrokah:
There's a bug in the missing table comment. I'll fix it tomorrow

erezrokah:
`azure_security_contacts` fails with:

```
unmarshalling type *armsecurity.ContactList: unmarshalling type *armsecurity.ContactList: json: cannot unmarshal array into Go value of type map[string]json.RawMessage
```

I'll open a bug report for the Azure SDK

hermanschaaf (approved):
Nice find!

PR #6913, opened 2023-01-17
fix: Rename 'aws_ec2_regional_configs' to 'aws_ec2_regional_config'

PeterMGuy:


bbernays:
@PeterMGuy - In version `v6.0.0` we changed that table

You can see the [release notes here](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-aws-v6.0.0)

PR #6919, opened 2023-01-18
chore(azure): Do a single append

candiduslynx:


erezrokah (commented):
Hi @candiduslynx, can you explain how this makes the code better?
Seems like we're splitting the logic of iterating over subscriptions

erezrokah (approved):

PR #6921, opened 2023-01-18
fix(hackernews): Fix HN plugin to still work when no backend is specified

hermanschaaf:


plugins/source/hackernews/client/backend.go
```diff
@@ -0,0 +1,17 @@
+package client
+
+import "context"
```
erezrokah, line 3: Maybe this should be the default SDK value?

erezrokah (approved):

plugins/source/hackernews/client/backend.go
hermanschaaf: The default SDK value is nil, causing the need for nil checks in all the incremental resolvers. I agree this is not ideal, so I raised https://github.com/cloudquery/plugin-sdk/pull/616 which we can use here if it's accepted.

PR #6935, opened 2023-01-18
feat(mixpanel): Mixpanel source plugin

disq:
`mixpanel_export_events` table is incremental, others are not.


.github/pr_labeler.yml
```diff
@@ -50,8 +54,6 @@ tailscale:
   - plugins/source/tailscale/**/*
 vercel:
   - plugins/source/vercel/**/*
-salesforce:
-  - plugins/source/salesforce/**/*
 azblob:
   - plugins/destination/azblob/**/*
 postgresql:
```
hermanschaaf, line 59: No good deed goes unpunished: because of the sorting this file now has some conflicts 😅

plugins/source/mixpanel/README.md
```diff
@@ -0,0 +1,7 @@
+# Mixpanel Plugin
+
+The Mixpanel Shopify plugin pulls data from Mixpanel and loads it into any supported CloudQuery destination (e.g. PostgreSQL).
```
hermanschaaf, line 3:
```suggestion
The Mixpanel plugin pulls data from Mixpanel and loads it into any supported CloudQuery destination (e.g. PostgreSQL).
```

plugins/source/mixpanel/client/client.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, errors.New("no credentials provided")
+	}
+	if spec.ProjectID < 1 {
+		return nil, errors.New("no project id provided")
+	}
+
+	const dateFormat = "2006-01-02"
+
+	if spec.StartDate == "" {
+		dt := time.Now().UTC().Add(-30 * 86400 * time.Second).Format(dateFormat)
```
hermanschaaf, line 72:
```suggestion
		dt := time.Now().UTC().AddDate(0, 0, -30).Format(dateFormat)
```

plugins/source/mixpanel/client/client.go
```diff
@@ -0,0 +1,116 @@
...
+		logger.Info().Str("start_date", dt).Msg("no start date provided, defaulting to 30 days ago")
+		spec.StartDate = dt
+	}
+	if spec.EndDate == "" {
+		dt := time.Now().UTC().Add(-86400 * time.Second).Format(dateFormat)
```
hermanschaaf, line 77:
```suggestion
		dt := time.Now().UTC().AddDate(0, 0, -1).Format(dateFormat)
```

plugins/source/mixpanel/client/client.go
```diff
@@ -0,0 +1,116 @@
...
+		logger.Info().Str("end_date", dt).Msg("no end date provided, defaulting to yesterday")
```
hermanschaaf, line 78: Why yesterday and not today?

plugins/source/mixpanel/client/client.go
```diff
@@ -0,0 +1,116 @@
...
+		// allow empty username for the deprecated project secret method: https://developer.mixpanel.com/reference/project-secret
+		return nil, errors.New("no credentials provided")
+	}
+	if spec.ProjectID < 1 {
+		return nil, errors.New("no project id provided")
+	}
+
+	const dateFormat = "2006-01-02"
+
+	if spec.StartDate == "" {
```
hermanschaaf, line 71: Can we move these defaults to a `SetDefaults` method on the Spec object instead of in this function? And probably a `Validate` method as well, which can be called from here. This is fairly standard in our plugins and I think it provides good separation of concerns.

plugins/source/mixpanel/client/spec.go
```diff
@@ -0,0 +1,17 @@
+package client
+
+type Spec struct {
+	Username string `json:"username,omitempty"`
+	Secret   string `json:"secret,omitempty"`
+
+	ProjectID   int64 `json:"project_id,omitempty"`
+	WorkspaceID int64 `json:"workspace_id,omitempty"`
+
+	StartDate string `json:"start_date,omitempty"`
```
hermanschaaf, line 10: I guess these are optional as well, because we provide defaults (comment below seems to indicate that they are not)

plugins/source/mixpanel/client/client.go
```diff
@@ -0,0 +1,116 @@
...
+	if spec.MaxRetries < 1 {
+		spec.MaxRetries = 5
+	}
+
+	rg, err := mixpanel.ParseRegion(spec.Region)
+	if err != nil {
+		return nil, err
+	}
+	if rg == mixpanel.RegionNone {
+		rg = mixpanel.RegionUS
```
hermanschaaf, line 101: What does region mean in the Mixpanel world? Does every workspace have a specific region? Could you miss datapoints if you extract from one region but actually multiple are in use?

plugins/source/mixpanel/docs/tables/mixpanel_cohorts.md
```diff
@@ -0,0 +1,24 @@
...
+## Columns
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|count|Int|
+|is_visible|Int|
```
hermanschaaf, line 19: Just asking: should this be a boolean?

plugins/source/mixpanel/docs/tables/mixpanel_export_events.md
```diff
@@ -0,0 +1,16 @@
+# Table: mixpanel_export_events
+
+The composite primary key for this table is (**project_id**, **time**).
```
hermanschaaf, line 3: Would definitely be nice to make this table incremental, it could be a _lot_ of data

plugins/source/mixpanel/docs/tables/mixpanel_export_events.md
```diff
@@ -0,0 +1,16 @@
...
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|project_id (PK)|Int|
+|time (PK)|Timestamp|
+|event|String|
```
hermanschaaf, line 15: I don't know what `event` contains, but I think either that or one of the properties will need to be part of the PK as well: relying on `time` wouldn't be good enough to differentiate events, especially if it's only at secondly granularity, but even if it is at millisecond. Unless Mixpanel guarantees the timestamps to be unique?

plugins/source/mixpanel/docs/tables/mixpanel_funnel_reports.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Columns
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|project_id|Int|
```
hermanschaaf, line 17: `project_id` should probably be part of the PK too

plugins/source/mixpanel/docs/tables/mixpanel_funnels.md
```diff
@@ -0,0 +1,20 @@
...
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|project_id|Int|
+|funnel_id|Int|
+|name|String|
```
hermanschaaf, line 20: Bit strange maybe, but these three keys together should be okay as a PK, right?

plugins/source/mixpanel/resources/services/export/export_events.go
```diff
@@ -0,0 +1,63 @@
...
+	e := r.Item.(mixpanel.ExportEvent)
+	ts, ok := e.Properties["time"]
+	if !ok {
+		return errors.New("event does not have a time property")
+	}
+	tf, ok := ts.(float64) // json.Number translates to float64
+	if !ok {
+		return fmt.Errorf("event time property is not a float: %T", ts)
+	}
+	return r.Set(c.Name, time.Unix(int64(tf), 0))
```
hermanschaaf, line 62: Yeah to my previous comment, looks like we're discarding any potential granularity of the timestamp here and only using unix seconds, which (if used as a PK) will probably cause events to be overwritten incorrectly? Not sure but if Mixpanel timestamps use floats as their PK then we can probably solve it by passing the microseconds in here as well, but we should be sure then that destination databases also store it at that granularity...

website/pages/docs/plugins/sources/mixpanel/overview.md
```diff
@@ -0,0 +1,63 @@
...
+  # Mixpanel specific configuration
+  spec:
+    username: "<YOUR_SERVICE_ACCOUNT_USER_HERE>"
+    secret: "<YOUR_SERVICE_ACCOUNT_SECRET_HERE>"
+    project_id: 12345 # Your project ID 
+```
+
+## Configuration Reference
+
+This is the (nested) spec used by the Stripe source plugin:
```
hermanschaaf, line 36:
```suggestion
This is the (nested) spec used by the Mixpanel source plugin:
```

website/pages/docs/plugins/sources/mixpanel/overview.md
```diff
@@ -0,0 +1,63 @@
...
+
+This example syncs from Mixpanel to a Postgres destination. The (top level) source spec section is described in the [Source Spec Reference](/docs/reference/source-spec).
+
+```yaml
+kind: source
+# Common source-plugin configuration
+spec:
+  name: mixpanel
+  path: cloudquery/mixpanel
+  version: "VERSION_SOURCE_STRIPE"
```
hermanschaaf, line 24:
```suggestion
  version: "VERSION_SOURCE_MIXPANEL"
```

hermanschaaf, line 24:
```suggestion
  version: "VERSION_SOURCE_MIXPANEL"
```

hermanschaaf (changes requested):
Looks good 🚀 Just commented on a few things to clarify / sort out before merging

.github/pr_labeler.yml
disq: yeah I know, I'll keep on top of it until we get it merged

plugins/source/mixpanel/client/client.go
disq: Usually (historically, based on previous multiple analytics and marketing data fetching experience) today's data is either incomplete (bad if you're using cursors and don't account for it) or doesn't exist... I didn't test this with mixpanel but it seemed like a good compromise. A way to specify relative dates could work here (maybe if it's an integer we can treat them as days to go back, etc...)

plugins/source/mixpanel/client/client.go
disq: Good shout, will do

plugins/source/mixpanel/client/client.go
disq: Your project lives in a specific region (EU residency data concerns). Multiple regions can't be in use in the same project. "documented" when creating a project :)

Screenshot 2023-01-23 at 11 05 41

plugins/source/mixpanel/docs/tables/mixpanel_cohorts.md
disq: The API returns an integer and not a boolean. Opted to keep it as-is for parity and code simplicity (would have to go for two separate structs or custom resolvers etc etc otherwise)

plugins/source/mixpanel/docs/tables/mixpanel_funnels.md
disq: `funnel_id` is supposed to be the PK but code has a bug so it's not applied. Will fix.

plugins/source/mixpanel/docs/tables/mixpanel_funnel_reports.md
disq: Not really, funnel id is supposed to be unique (again, not documented, but looks like it)

plugins/source/mixpanel/resources/services/export/export_events.go
disq: Time is seconds. It's probably not even unique including `event`.

plugins/source/mixpanel/client/client.go
hermanschaaf: I think syncing today's data should be fine, as on the next sync you will fill in the late-arriving datapoints. Incremental syncs should take care of this by fetching some overlapping window of events again on the next sync (probably about an hour), but it shouldn't be a concern for non-incremental tables.

plugins/source/mixpanel/resources/services/export/export_events.go
disq: Fixed in [`5e55b2e` (#6935)](https://github.com/cloudquery/cloudquery/pull/6935/commits/5e55b2e12611e093a7695bd69ca8f9df1477e987)

plugins/source/mixpanel/client/client.go
disq: Looks like the incremental cursor is going to be the `time` field (which is unix seconds). So we could move this to be today, if other resources don't fail if today is requested. If they do, we'll just fallback to the sane default of yesterday for those resources (can have a separate EndDateExceptToday field that gets auto calculated and so on)

disq: None of the resources complained, so now the end date is today.

hermanschaaf (approved):
✨

disq:
⌛

erezrokah:
@disq I'm looking after you:
image

PR #6951, opened 2023-01-18
feat: Add support for {{TABLE}} and {{UUID}} placeholders in S3 plugin path

hermanschaaf:


plugins/destination/s3/client/write.go
```diff
@@ -33,8 +48,8 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
 	default:
 		panic("unknown format " + c.pluginSpec.Format)
 	}
-	// we don't do that in parallel here because aws sdk moves the burden to the devleoper and
-	// we don't want to deal with that yet. in the future maybe we can run some benchmarks and see if adding parralization helps.
+	// we don't do upload in parallel here because AWS sdk moves the burden to the developer, and
```
disq, line 51:
```suggestion
	// we don't upload in parallel here because AWS sdk moves the burden to the developer, and
```

plugins/destination/s3/client/write.go
```diff
@@ -14,10 +15,24 @@ import (
...
+const (
+	PathVarTable = "{{TABLE}}"
+	PathVarUUID  = "{{UUID}}"
+)
+
 func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data [][]any) error {
-	name := fmt.Sprintf("%s/%s.%s.%s", c.pluginSpec.Path, table.Name, c.pluginSpec.Format, uuid.NewString())
-	if c.pluginSpec.NoRotate {
-		name = fmt.Sprintf("%s/%s.%s", c.pluginSpec.Path, table.Name, c.pluginSpec.Format)
+	path := c.pluginSpec.Path
```
candiduslynx, line 24: We should update `c.pluginSpec.Path` during initialization with new spec, so that we don't repeat the same work over and over in `WriteTableBatch`.
`WriteTableBatch` should be just filling in the path with:
```go
name := strings.ReplaceAll(path, PathVarTable, table.Name)
name = strings.ReplaceAll(path, PathVarUUID, uuid.NewString()) // don't need to check as was checked during init phase
```

candiduslynx (changes requested):

plugins/destination/s3/client/write.go
hermanschaaf: Good suggestion, thanks! Updated :+1:

yevgenypats (approved):

bbernays:
@hermanschaaf - Because Athena partitions on the path... Maybe we need a way of injecting a date into the object path?
`{{TABLE}}/{{YEAR}}/{{MONTH}}/{{DAY}}/{{HOUR}}/{{MINUTE}}/{{UUID}}

hermanschaaf:
@bbernays Yeah I think we will need something like this... though for now just wanted to keep it simple; we should be able to add support for dates in a backwards-compatible way.

plugins/destination/s3/client/spec_test.go
```diff
@@ -0,0 +1,47 @@
...
+	for _, tc := range cases {
+		err := tc.Give.Validate()
+		gotErr := err != nil
+		if diff := cmp.Diff(tc.WantErr, gotErr); diff != "" {
+			s := "nil"
+			if tc.WantErr {
+				s = "error"
+			}
+			t.Errorf("Validate(%v) got err = %v, want %v", tc.Give, err, s)
+		}
```
candiduslynx, line 45:
```suggestion
		err := tc.Give.Validate()
		switch tc.WantErr {
		case true:
			require.Error(t, err)
		default:
			require.NoError(t, err)
		}
```

plugins/destination/s3/client/write.go
```diff
@@ -14,10 +14,15 @@ import (
...
+	PathVarTable = "{{TABLE}}"
+	PathVarUUID  = "{{UUID}}"
+)
+
 func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data [][]any) error {
-	name := fmt.Sprintf("%s/%s.%s.%s", c.pluginSpec.Path, table.Name, c.pluginSpec.Format, uuid.NewString())
-	if c.pluginSpec.NoRotate {
-		name = fmt.Sprintf("%s/%s.%s", c.pluginSpec.Path, table.Name, c.pluginSpec.Format)
+	name := strings.ReplaceAll(c.pluginSpec.Path, PathVarTable, table.Name)
+	if !c.pluginSpec.NoRotate {
```
candiduslynx, line 24: no need to check, just update, as the set defaults + validate already did their job

candiduslynx (approved):

PR #6954, opened 2023-01-18
feat: Support Kinesis Firehose as a Destination

bbernays:
#### Summary

Supports Kinesis Firehose as a destination...

Code is working, but not sure how to test as Kinesis Firehose Doesn't have a capability to read from it. The closest we could do is set batching to the lowest setting and wait 5 minutes for data to appear in the destination. But there is no way to control for multiple tests running at the same time.


plugins/destination/firehose/client/client.go
```diff
@@ -0,0 +1,63 @@
...
+	destination.UnimplementedManagedWriter
+	destination.DefaultReverseTransformer
+	logger         zerolog.Logger
+	spec           specs.Destination
+	pluginSpec     Spec
+	metrics        destination.Metrics
+	firehoseClient *firehose.Client
+
+	jsonTransformer        *json.Transformer
+	jsonReverseTransformer *json.ReverseTransformer
```
bbernays, line 27: This needs to use the new FileTypes client...

hermanschaaf:
@bbernays The tests are failing here and the PR is still marked as draft, so I'll postpone review for now, but let me know when you're ready and I'd be happy to take a look

plugins/destination/firehose/client/write.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		c.logger.Error().Err(err).Msg("failed to write to firehose")
+		return err
+	}
+	retryRecords := getFailedRecords(recordsBatchInput, resp)
+	atomic.AddUint64(&c.metrics.Writes, uint64(len(recordsBatchInput.Records)-len(retryRecords.Records)))
+	return c.sendBatch(ctx, retryRecords, count+1)
+}
+
+func getFailedRecords(recordsBatchInput *firehose.PutRecordBatchInput, resp *firehose.PutRecordBatchOutput) *firehose.PutRecordBatchInput {
```
yevgenypats, line 116: Let's not do that. either a batch failed or succeeded and we log it. this will introduce a behaviour we had in v0 that we have no idea why something taking too long is it because we used `copyfrom`, `insertbatch` or `insert one by one` because the batch failed

yevgenypats: In general a failed batch should rarely happen so we shouldn't have a code handling that as most of the validation is hapening on the SDK side

yevgenypats (changes requested):
Looks. had on comment around the batching mechanism.

plugins/destination/firehose/client/write.go
bbernays: For Batch requests the SDK doesn't handle retries for individual records that are rate limited... This is the case for Kinesis Firehose, Kinesis Data streams, SQS, and DynamoDB... So it is actually quite easy to hit the default rate limit...

bbernays: The SDK will only retry if the request returns an retryable error... In the cases of Batch endpoints the request can return a 200 status code, but every single record could have been rate limited and therefore not inserted...

yevgenypats (approved):

hermanschaaf:
@bbernays I'll review the full PR shortly, but it would be really nice if we can test this somehow, even if the test (by necessity) takes 5 minutes. Could we use [localstack](https://docs.localstack.cloud/user-guide/aws/kinesis-firehose/) for testing this case? Maybe it doesn't have the same limitation

plugins/destination/firehose/client/client_test.go
```diff
@@ -0,0 +1,26 @@
...
+		Spec{
+			NoRotate:  true,
+			StreamARN: streamARN,
+		},
+		destination.PluginTestSuiteTests{
+			SkipOverwrite:     true,
+			SkipDeleteStale:   true,
+			SkipSecondAppend:  true,
+			SkipMigrateAppend: true,
+			SkipAppend:        true,
```
hermanschaaf, line 23: I think this is skipping.... everything? Except `MigrateOverwrite`, but I'm not sure why 😄

plugins/destination/firehose/client/read.go
```diff
@@ -0,0 +1,11 @@
+package client
+
+import (
+	"context"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (*Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []any) error {
+	panic("not implemented ")
```
hermanschaaf, line 10:
```suggestion
	panic("not implemented")
```

plugins/destination/firehose/client/write.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/aws/aws-sdk-go-v2/service/firehose"
+	"github.com/aws/aws-sdk-go-v2/service/firehose/types"
+	"github.com/cloudquery/plugin-sdk/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+const MAX_RECORD_SIZE_BYTES = 1024000
+const MAX_BATCH_RECORDS = 500
+const MAX_BATCH_SIZE_BYTES = 4194000
```
hermanschaaf, line 22:
```suggestion
const MaxRecordSizeBytes = 1024000
const MaxBatchRecords = 500
const MaxBatchSizeBytes = 4194000
```

(nit to make it match general Go style. The names will also have to be updated lower down)

plugins/destination/firehose/main.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/cloudquery/cloudquery/plugins/destination/firehose/client"
+	"github.com/cloudquery/cloudquery/plugins/destination/firehose/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const (
+	// TODO: Add Sentry DSN
+	sentryDSN = ""
```
hermanschaaf, line 12:
```suggestion
	sentryDSN = "https://03ef940ae6064ae8a69fa5e1ad74d3f2@o1396617.ingest.sentry.io/4504600056299520"
```

website/pages/docs/plugins/destinations/firehose/overview.md
```diff
@@ -0,0 +1,56 @@
...
+The (top level) spec section is described in the [Destination Spec Reference](/docs/reference/destination-spec).
+
+```yaml
+kind: destination
+spec:
+  name: "firehose"
+  path: "cloudquery/firehose"
+  version: "VERSION_DESTINATION_FIREHOSE"
+  write_mode: "append" # this plugin only supports 'append' mode
+  # batch_size: 10000 # optional
```
hermanschaaf, line 37:
```suggestion
  # batch_size: 500 # optional
```

plugins/destination/firehose/client/write.go
```diff
@@ -0,0 +1,133 @@
...
+		b, err := json.Marshal(jsonObj)
+		if err != nil {
+			return err
+		}
+		dst := &bytes.Buffer{}
+		err = json.Compact(dst, b)
+		if err != nil {
+			return err
+		}
+		if len(dst.Bytes()) > MAX_RECORD_SIZE_BYTES {
```
hermanschaaf, line 62:
```suggestion
		if len(dst.Bytes()) > MaxRecordSizeBytes {
```

plugins/destination/firehose/client/write.go
```diff
@@ -0,0 +1,133 @@
...
+			c.logger.Warn().Msgf("skipping record because it is too large: %s", string(b))
+			continue
+		}
+
+		// If adding this record would exceed the batch size, send the batch
+		if len(dst.Bytes())+batchSize > min(c.spec.BatchSizeBytes, MAX_BATCH_SIZE_BYTES) {
```
hermanschaaf, line 68:
```suggestion
		if len(dst.Bytes())+batchSize > min(c.spec.BatchSizeBytes, MaxBatchSizeBytes) {
```

plugins/destination/firehose/client/write.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		recordsBatchInput.Records = append(recordsBatchInput.Records, types.Record{
+			Data: dst.Bytes(),
+		})
+		// Store a running total of the batch size
+		batchSize += len(dst.Bytes())
+
+		// Send the batch if it is full
+		if len(recordsBatchInput.Records) == min(c.spec.BatchSize, MAX_BATCH_RECORDS) {
```
hermanschaaf, line 84:
```suggestion
		if len(recordsBatchInput.Records) == min(c.spec.BatchSize, MaxBatchRecords) {
```

hermanschaaf (approved):
LGTM, just left a few suggestions. I think Localstack testing is probably worth a shot as well :)

plugins/destination/firehose/client/client_test.go
bbernays: Because I didn't know that one existed... Will add it

PR #6958, opened 2023-01-18
feat(s3): Expose New Options

bbernays:
#### Summary

Users can now use the following options to change the destination behavior:
1. `include_header`: default is `false` when set to true first row will be the column headers
2. `delimiter`: default value is `,` but user can specify any single character

### DIFF
--- a/plugins/destination/s3/client/read.go
+++ b/plugins/destination/s3/client/read.go
@@ -45,11 +45,26 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
 
 	switch c.pluginSpec.Format {
 	case FormatTypeCSV:
-		if err := csv.Read(r, table, sourceName, res); err != nil {
+		opts := []csv.Options{
+			csv.WithDelimiter(c.pluginSpec.Delimiter),
+		}
+		if c.pluginSpec.IncludeHeaders {
+			opts = append(opts, csv.WithHeader())
+		}
+
+		client, err := csv.NewClient(opts...)
+		if err != nil {
+			return err
+		}
+		if err = client.Read(r, table, sourceName, res); err != nil {
 			return err
 		}
 	case FormatTypeJSON:
-		if err := json.Read(r, table, sourceName, res); err != nil {
+		client, err := json.NewClient()
+		if err != nil {
+			return err
+		}
+		if err := client.Read(r, table, sourceName, res); err != nil {
 			return err
 		}
 	default:
--- a/plugins/destination/s3/client/spec.go
+++ b/plugins/destination/s3/client/spec.go
@@ -12,13 +12,19 @@ const (
 )
 
 type Spec struct {
-	Bucket   string     `json:"bucket,omitempty"`
-	Path     string     `json:"path,omitempty"`
-	Format   FormatType `json:"format,omitempty"`
-	NoRotate bool       `json:"no_rotate,omitempty"`
+	Bucket         string     `json:"bucket,omitempty"`
+	Path           string     `json:"path,omitempty"`
+	Format         FormatType `json:"format,omitempty"`
+	IncludeHeaders bool       `json:"include_headers,omitempty"`
+	Delimiter      rune       `json:"delimiter,omitempty"`
+	NoRotate       bool       `json:"no_rotate,omitempty"`
 }
 
-func (*Spec) SetDefaults() {}
+func (s *Spec) SetDefaults() {
+	if s.Delimiter == 0 {
+		s.Delimiter = ','
+	}
+}
 
 func (s *Spec) Validate() error {
 	if s.Bucket == "" {
--- a/plugins/destination/s3/client/write.go
+++ b/plugins/destination/s3/client/write.go
@@ -23,11 +23,19 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data 
 	w := io.Writer(&b)
 	switch c.pluginSpec.Format {
 	case FormatTypeCSV:
-		if err := csv.WriteTableBatch(w, table, data); err != nil {
+		client, err := csv.NewClient()
+		if err != nil {
+			return err
+		}
+		if err := client.WriteTableBatch(w, table, data); err != nil {
 			return err
 		}
 	case FormatTypeJSON:
-		if err := json.WriteTableBatch(w, table, data); err != nil {
+		client, err := json.NewClient()
+		if err != nil {
+			return err
+		}
+		if err := client.WriteTableBatch(w, table, data); err != nil {
 			return err
 		}
 	default:
--- a/plugins/destination/s3/go.mod
+++ b/plugins/destination/s3/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.18.8
 	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.11.47
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.30.0
-	github.com/cloudquery/filetypes v1.0.6
+	github.com/cloudquery/filetypes v1.1.0
 	github.com/cloudquery/plugin-sdk v1.27.0
 	github.com/google/uuid v1.3.0
 	github.com/rs/zerolog v1.28.0
--- a/plugins/destination/s3/go.sum
+++ b/plugins/destination/s3/go.sum
@@ -78,8 +78,8 @@ github.com/chzyer/logex v1.1.10/go.mod h1:+Ywpsq7O8HXn0nuIou7OrIPyXbp3wmkHB+jjWR
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
-github.com/cloudquery/filetypes v1.0.6 h1:Jy93/hrSS7FEIQnAl9DSS18dh8xCeP7bUS/Ufrm27j0=
-github.com/cloudquery/filetypes v1.0.6/go.mod h1:J+aAJCqYnBsD0s5R350GORkL2od8rilOYu6vKBD5l8M=
+github.com/cloudquery/filetypes v1.1.0 h1:6OOogHzaBXPy0Lmxue0FZEYFI0ouY9EwzwIQPYSg95s=
+github.com/cloudquery/filetypes v1.1.0/go.mod h1:J+aAJCqYnBsD0s5R350GORkL2od8rilOYu6vKBD5l8M=
 github.com/cloudquery/plugin-sdk v1.27.0 h1:DXuvnBt1gOB98umBZU6jltZEV6oxfsdEBIAbQXFcIx4=
 github.com/cloudquery/plugin-sdk v1.27.0/go.mod h1:teMPyCON3uPdMsHvzpSiOg+IK2sOR5Tf9dYLreoURzI=
 github.com/cncf/udpa/go v0.0.0-20191209042840-269d4d468f6f/go.mod h1:M8M6+tZqaGXZJjfX53e64911xZQV5JYwmTeXPW+k8Sc=


plugins/destination/s3/client/write.go
```diff
@@ -23,11 +23,19 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
 	w := io.Writer(&b)
 	switch c.pluginSpec.Format {
 	case FormatTypeCSV:
-		if err := csv.WriteTableBatch(w, table, data); err != nil {
+		client, err := csv.NewClient()
```
candiduslynx, line 26: Shouldn't options be passed the same way as in read.go?
```suggestion
		opts := []csv.Options{
			csv.WithDelimiter(c.pluginSpec.Delimiter),
		}
		if c.pluginSpec.IncludeHeaders {
			opts = append(opts, csv.WithHeader())
		}

		client, err := csv.NewClient(opts...)
```

candiduslynx (changes requested):
I wonder if we could also add CLRF support to the csv plugin, as [encoding/csv.Writer](https://pkg.go.dev/encoding/csv#Writer) supports it, too. (Bound to another PR).

candiduslynx (approved):

plugins/destination/s3/client/spec.go
```diff
@@ -12,13 +12,19 @@ const (
...
 type Spec struct {
-	Bucket   string     `json:"bucket,omitempty"`
-	Path     string     `json:"path,omitempty"`
-	Format   FormatType `json:"format,omitempty"`
-	NoRotate bool       `json:"no_rotate,omitempty"`
+	Bucket         string     `json:"bucket,omitempty"`
+	Path           string     `json:"path,omitempty"`
+	Format         FormatType `json:"format,omitempty"`
+	IncludeHeaders bool       `json:"include_headers,omitempty"`
+	Delimiter      rune       `json:"delimiter,omitempty"`
```
erezrokah, line 19: Should this be a `string` instead of a `rune`? The docs say `string` as the type but explain it's a character.
Do we mind letting people use `, ` and not only `,`?

website/pages/docs/plugins/destinations/s3/overview.md
```diff
@@ -51,3 +51,11 @@ This is the (nested) spec used by the CSV destination Plugin.
 - `format` (string) (required)
 
   Format of the output file. `json` and `csv` are supported.
+
+- `include_headers` (boolean) (optional)
+
+  Whether the first line of the file should be the column headers. Default value is `false`.
```
hermanschaaf, line 57: What's the reasoning behind defaulting to false?

website/pages/docs/plugins/destinations/s3/overview.md
```diff
@@ -51,3 +51,11 @@ This is the (nested) spec used by the CSV destination Plugin.
...
+
+- `delimiter` (string) (optional)
+
+  Character that should be used as the delimiter. Default values is `,`.
```
hermanschaaf, line 61:
```suggestion
  Character that should be used as the delimiter. Default value is `,`.
```

plugins/destination/s3/client/spec.go
hermanschaaf: @erezrokah We can use a string here, but we'd always have to use only the first rune, because the Go CSV library only takes a rune https://pkg.go.dev/encoding/csv#Writer

erezrokah: Ah, so lets keep `rune` so if someone uses a string reading the config fails. Maybe update the docs to say `char` instead of string?

website/pages/docs/plugins/destinations/s3/overview.md
bbernays: Backwards compatibility...

hermanschaaf: Ah okay; if that's the main reason, I think we should consider doing a major bump instead. The data would be incredibly hard to use without headers 🤷

plugins/destination/s3/client/spec.go
```diff
@@ -12,13 +12,19 @@ const (
...
 
 type Spec struct {
-	Bucket   string     `json:"bucket,omitempty"`
-	Path     string     `json:"path,omitempty"`
-	Format   FormatType `json:"format,omitempty"`
-	NoRotate bool       `json:"no_rotate,omitempty"`
+	Bucket         string     `json:"bucket,omitempty"`
+	Path           string     `json:"path,omitempty"`
+	Format         FormatType `json:"format,omitempty"`
+	IncludeHeaders bool       `json:"include_headers,omitempty"`
```
hermanschaaf, line 18: This setting only applies to the CSV format: maybe we should introduce a `format_spec` or something along those lines to capture more settings in the future?

website/pages/docs/plugins/destinations/s3/overview.md
bbernays: I figured that by the time people realized they needed headers they would be comfortable enough to read the docs and make config changes to take advantage of this feature

plugins/destination/s3/client/spec.go
```diff
@@ -12,15 +12,19 @@ const (
...
 	Delimiter      rune       `json:"delimiter,omitempty"`
 	NoRotate       bool       `json:"no_rotate,omitempty"`
 }
 
+type Spec struct {
+	Bucket string `json:"bucket,omitempty"`
+	Path   string `json:"path,omitempty"`
+	FileSpec
+}
+
```
bbernays, line 27: @hermanschaaf @yevgenypats Is this what you were imagining?

In a future PR, we can move it to the filetypes package

bbernays: (Obviously this will require additional docs)

plugins/destination/s3/client/spec.go
```diff
@@ -29,6 +29,10 @@ func (s *Spec) SetDefaults() {
 	if s.Delimiter == 0 {
 		s.Delimiter = ','
 	}
+	if s.IncludeHeaders == nil {
+		headers := true
+		s.IncludeHeaders = &headers
```
candiduslynx, line 34: ptr.To(true) (https://pkg.go.dev/github.com/candiduslynx/ptr#To) 😄

candiduslynx (approved):

plugins/destination/azblob/client/read.go
```diff
@@ -4,20 +4,11 @@ import (
...
 func (c *Client) ReverseTransformValues(table *schema.Table, values []any) (schema.CQTypes, error) {
-	switch c.pluginSpec.Format {
-	case FormatTypeCSV:
-		return c.csvReverseTransformer.ReverseTransformValues(table, values)
-	case FormatTypeJSON:
-		return c.jsonReverseTransformer.ReverseTransformValues(table, values)
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
+	return c.filetype.ReverseTransformValues(table, values)
```
candiduslynx, line 11: can be omitted by embedding filetype client

plugins/destination/azblob/client/transformer.go
```diff
@@ -1,190 +1,73 @@
...
 func (c *Client) TransformBool(v *schema.Bool) any {
-	switch c.pluginSpec.Format {
-	case FormatTypeCSV:
-		return c.csvTransformer.TransformBool(v)
-	case FormatTypeJSON:
-		return c.jsonTransformer.TransformBool(v)
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
+	return c.filetype.TransformBool(v)
```
candiduslynx, line 8: all these funcs can be omitted by embedding filetype client

plugins/destination/azblob/client/write.go
```diff
@@ -19,17 +17,8 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
...
-		if err := csv.WriteTableBatch(w, table, data); err != nil {
-			return err
-		}
-	case FormatTypeJSON:
-		if err := json.WriteTableBatch(w, table, data); err != nil {
-			return err
-		}
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
+	if err := c.filetype.WriteTableBatch(w, table, data); err != nil {
```
candiduslynx, line 20: bytes.Buffer already implements io.Writer, no need to wrap it in the interface

plugins/destination/gcs/client/client.go
```diff
@@ -23,10 +22,7 @@ type Client struct {
 	gcsClient *storage.Client
 	bucket    *storage.BucketHandle
 
-	csvTransformer         *csv.Transformer
-	csvReverseTransformer  *csv.ReverseTransformer
-	jsonTransformer        *json.Transformer
-	jsonReverseTransformer *json.ReverseTransformer
+	filetype *filetypes.Client
```
candiduslynx, line 25: embed to drop a lot of extra code

plugins/destination/gcs/client/read.go
```diff
@@ -31,17 +22,6 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
-	case FormatTypeJSON:
-		if err := json.Read(r, table, sourceName, res); err != nil {
-			return err
-		}
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
-	return nil
+	return c.filetype.Read(r, table, sourceName, res)
+
```
candiduslynx:
```suggestion
```

plugins/destination/gcs/client/write.go
```diff
@@ -17,18 +15,8 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
...
-		if err := csv.WriteTableBatch(w, table, data); err != nil {
-			return err
-		}
-	case FormatTypeJSON:
-		if err := json.WriteTableBatch(w, table, data); err != nil {
-			return err
-		}
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
+	if err := c.filetype.WriteTableBatch(w, table, data); err != nil {
```
candiduslynx:
```suggestion
	return c.filetype.WriteTableBatch(w, table, data)
```

plugins/destination/kafka/client/client.go
```diff
@@ -25,23 +24,15 @@ type Client struct {
 	spec       specs.Destination
 	pluginSpec Spec
 	metrics    destination.Metrics
-
-	csvTransformer         *csv.Transformer
-	csvReverseTransformer  *csv.ReverseTransformer
-	jsonTransformer        *json.Transformer
-	jsonReverseTransformer *json.ReverseTransformer
+	filetype   *filetypes.Client
```
candiduslynx, line 27: embed

plugins/destination/s3/client/client.go
```diff
@@ -26,34 +25,32 @@ type Client struct {
 	s3Client   *s3.Client
 	uploader   *manager.Uploader
 	downloader *manager.Downloader
-
-	csvTransformer         *csv.Transformer
-	csvReverseTransformer  *csv.ReverseTransformer
-	jsonTransformer        *json.Transformer
-	jsonReverseTransformer *json.ReverseTransformer
+	filetype   *filetypes.Client
```
candiduslynx, line 28: embed filetype.Client

candiduslynx (changes requested):

plugins/destination/kafka/client/client.go
bbernays: Can't embed for kafka, as it uses `Write` rather than `BatchedWrite`

plugins/destination/s3/client/spec.go
bbernays: Yeah that could work, but overall it is the same number of LOC...

plugins/destination/s3/client/client.go
bbernays: Nice catch on that!

plugins/destination/kafka/client/client.go
candiduslynx: you can but will have to impl extra method

plugins/destination/azblob/client/read.go
```diff
@@ -32,17 +19,5 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
-		}
-	case FormatTypeJSON:
-		if err := json.Read(response.Body, table, sourceName, res); err != nil {
-			return err
-		}
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
-	return nil
+	return c.Client.Read(response.Body, table, sourceName, res)
```
candiduslynx, line 22: rm this func (embedded)

plugins/destination/gcs/client/read.go
```diff
@@ -31,17 +18,5 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
-		}
-	case FormatTypeJSON:
-		if err := json.Read(r, table, sourceName, res); err != nil {
-			return err
-		}
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
-	return nil
+	return c.Client.Read(r, table, sourceName, res)
```
candiduslynx, line 21: rm this func (use embedded). possibly move noRotate check to embedded client

plugins/destination/kafka/client/read.go
```diff
@@ -16,14 +14,7 @@ const (
...
 func (c *Client) ReverseTransformValues(table *schema.Table, values []any) (schema.CQTypes, error) {
-	switch c.pluginSpec.Format {
-	case FormatTypeCSV:
-		return c.csvReverseTransformer.ReverseTransformValues(table, values)
-	case FormatTypeJSON:
-		return c.jsonReverseTransformer.ReverseTransformValues(table, values)
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
+	return c.filetype.ReverseTransformValues(table, values)
```
candiduslynx, line 17: embed client, drop this func

plugins/destination/kafka/client/transformer.go
```diff
@@ -1,190 +1,73 @@
...
 func (c *Client) TransformBool(v *schema.Bool) any {
-	switch c.pluginSpec.Format {
-	case FormatTypeCSV:
-		return c.csvTransformer.TransformBool(v)
-	case FormatTypeJSON:
-		return c.jsonTransformer.TransformBool(v)
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
+	return c.filetype.TransformBool(v)
```
candiduslynx, line 8: embed & drop transformers

plugins/destination/s3/client/read.go
```diff
@@ -43,18 +30,5 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
-		}
-	case FormatTypeJSON:
-		if err := json.Read(r, table, sourceName, res); err != nil {
-			return err
-		}
-	default:
-		panic("unknown format " + c.pluginSpec.Format)
-	}
-	return nil
+	return c.Client.Read(r, table, sourceName, res)
```
candiduslynx, line 33: can noRotate check be moved to filetype plugin?
that way embedded client func will be enough

candiduslynx (changes requested):

plugins/destination/kafka/client/transformer.go
bbernays: Can't embed because there is no `BatchWrite` like the others

plugins/destination/kafka/client/read.go
bbernays: Can't embed because there is no `BatchWrite` like the other plugins

plugins/destination/gcs/client/read.go
bbernays: Read has a different function signature...

plugins/destination/azblob/client/read.go
bbernays: different function signature

plugins/destination/kafka/client/transformer.go
candiduslynx: why is this a problem?

plugins/destination/kafka/client/read.go
candiduslynx: So, if you embed you don't have to implement a lot of functions, just an extra `BatchWrite`, right?
So shy don't just embed and ditch all transformers then, by implementing a singe `BatchWrite`?

candiduslynx: additionally, plugin-sdk/plugins/destination have both unimplemented stubs for managed & unmanaged writers to embed for easier impl

plugins/destination/s3/client/read.go
bbernays: No, because the NoRotate is only used when the destination doesn't support appending, so it needs to be in the destination package rather than filetype package

candiduslynx: There're several reasons why I consider this:

- the spec that controls such behavior (has field NoRotate) is in filetypes package
- client is initialized using that spec
- client is also in that filetypes package

It seems natural to move noRotate check to the filetypes client impl then.

bbernays: I think I will just move the NoRotate out of filetypes

plugins/destination/kafka/client/write.go
```diff
@@ -82,3 +71,7 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *de
 
 	return nil
 }
+
+func (*Client) WriteTableBatch(ctx context.Context, table *schema.Table, data [][]any) error {
```
candiduslynx, line 75: maybe just embed [destination.UnimplementedManagedWriter](https://pkg.go.dev/github.com/cloudquery/plugin-sdk@v1.27.0/plugins/destination#UnimplementedManagedWriter)?

candiduslynx (commented):
a couple of comments left

plugins/destination/kafka/client/write.go
bbernays: It already is:
https://github.com/cloudquery/cloudquery/blob/24b0219ca14628da3f3188efb463915064bf9a0d/plugins/destination/kafka/client/client.go#L18-L33

candiduslynx: then explicit func isn't required

candiduslynx (approved):

candiduslynx (approved):

candiduslynx (approved):

candiduslynx (approved):

candiduslynx (approved):

candiduslynx (approved):

website/pages/docs/plugins/destinations/gcs/overview.mdx
```diff
@@ -47,3 +47,17 @@ This is the (nested) spec used by the CSV destination Plugin.
...
+
+## format_spec
+
+- `delimiter` (string) (optional) (default: `,`)
+
+  Character that will be used as want to use as the delimiter if the format type is `csv`
+
+- `include_headers` (bool) (optional) (default: true)
+
+  Specifies if the first header of a file should be the headers.
```
hermanschaaf, line 63:
```suggestion
  Specifies if the first line of a file should be the headers.
```

website/pages/docs/plugins/destinations/azblob/overview.mdx
```diff
@@ -55,3 +55,17 @@ This is the (nested) spec used by the Azure blob destination Plugin.
...
+
+## format_spec
+
+- `delimiter` (string) (optional) (default: `,`)
+
+  Character that will be used as want to use as the delimiter if the format type is `csv`
+
+- `include_headers` (bool) (optional) (default: true)
+
+  Specifies if the first header of a file should be the headers.
```
hermanschaaf, line 71:
```suggestion
  Specifies if the first line of a file should be the headers (when format is `csv`).
```

website/pages/docs/plugins/destinations/s3/overview.mdx
```diff
@@ -48,10 +48,24 @@ This is the (nested) spec used by the CSV destination Plugin.
...
+
+## format_spec
+
+- `delimiter` (string) (optional) (default: `,`)
+
+  Character that will be used as want to use as the delimiter if the format type is `csv`
+
+- `include_headers` (bool) (optional) (default: true)
+
+  Specifies if the first header of a file should be the headers.
```
hermanschaaf, line 71:
```suggestion
  Specifies if the first line of a file should be the headers (when format is `csv`).
```

hermanschaaf:
I haven't delved too deep into this, but is the default for `include_headers` really `true` like the docs say? Wouldn't we need to be using `*bool` type then?

bbernays:
> I haven't delved too deep into this, but is the default for `include_headers` really `true` like the docs say? Wouldn't we need to be using `*bool` type then?

Had a conversation... Changing the field to `no_headers` and default is false... This will force a breaking change for all 4 destinations

disq:
> > I haven't delved too deep into this, but is the default for `include_headers` really `true` like the docs say? Wouldn't we need to be using `*bool` type then?
> 
> Had a conversation... Changing the field to `no_headers` and default is false... This will force a breaking change for all 4 destinations

`skip_header`?

website/pages/docs/plugins/destinations/azblob/overview.mdx
```diff
@@ -55,3 +55,17 @@ This is the (nested) spec used by the Azure blob destination Plugin.
 - `format` (string) (required)
 
   Format of the output file. `json` and `csv` are supported.
+
+- `format_spec` (map [format_spec](#format_spec)) (default: current account)
```
hermanschaaf, line 59:
```suggestion
- `format_spec` (map [format_spec](#format_spec)) (optional)
```

website/pages/docs/plugins/destinations/gcs/overview.mdx
```diff
@@ -47,3 +47,17 @@ This is the (nested) spec used by the CSV destination Plugin.
 - `format` (string) (required)
 
   Format of the output file. `json` and `csv` are supported.
+
+- `format_spec` (map [format_spec](#format_spec)) (default: current account)
```
hermanschaaf, line 51:
```suggestion
- `format_spec` (map [format_spec](#format_spec)) (optional)
```

website/pages/docs/plugins/destinations/kafka/overview.mdx
```diff
@@ -48,4 +44,22 @@ This is the (nested) plugin spec
...
 
-  If true, the plugin will log all underlying Kafka client messages to the log.
\ No newline at end of file
+  If true, the plugin will log all underlying Kafka client messages to the log.
+
+- `format` (string) (required)
+
+  Format of the output file. `json` and `csv` are supported.
+
+- `format_spec` (map [format_spec](#format_spec)) (default: current account)
```
hermanschaaf, line 53:
```suggestion
- `format_spec` (map [format_spec](#format_spec)) (optional)
```

website/pages/docs/plugins/destinations/s3/overview.mdx
```diff
@@ -48,10 +48,24 @@ This is the (nested) spec used by the CSV destination Plugin.
...
+- `athena` (boolean) (optional, default `false`)
+
+  When `athena` is set to `true`, the S3 plugin will sanitize keys in JSON columns to be compatible with the Hive Metastore / Athena. This allows tables to be created with a Glue Crawler and then queried via Athena, without changes to the table schema.
+
 - `format` (string) (required)
 
   Format of the output file. `json` and `csv` are supported.
 
-- `athena` (boolean) (optional, default `false`)
+- `format_spec` (map [format_spec](#format_spec)) (default: current account)
```
hermanschaaf, line 59:
```suggestion
- `format_spec` (map [format_spec](#format_spec)) (optional)
```

hermanschaaf (approved):

disq:
Docs wise I don't think `format_spec` is clear, an example would've been nice.

bbernays:
> Docs wise (e.g. checking [here](https://cloudquery-hshvjps7n-cloudquery.vercel.app/docs/plugins/destinations/gcs/overview)) I don't think `format_spec` is clear, an example would've been nice.

Can we handle this in a follow up PR?

disq (approved):

yevgenypats (approved):

PR #6970, opened 2023-01-19
chore(mssql): Get Microsoft SQL Server ready for release

candiduslynx:


plugins/destination/mssql/client/migrate.go
```diff
@@ -17,16 +17,21 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
+		}
+	}()
 
 	if len(table.Columns) == 0 {
 		c.logger.Info().Str("table", table.Name).Msg("Table with no columns, skipping")
 		return nil
 	}
 
-	switch tableExist, err := c.tableExists(ctx, table.Name); {
+	switch tableExist, err := c.tableExists(ctx, table); {
```
hermanschaaf, line 34: I could be missing something, but it looks like `err` is being shadowed here (and inside the cases), so I don't think the `migrate`-scope `err` that is logged in the defer will ever be non-nil?

disq: The `err` in the defer is coming from the named return argument I think.

hermanschaaf: Right, but where is it getting set?

candiduslynx: yes, the err in defer is exactly from the named ret value

candiduslynx: on return

hermanschaaf: Oh okay, I didn't know defer worked that way with named return values, TIL

candiduslynx: yeah, basically, you can use defer to update named return values, see https://go.dev/blog/defer-panic-and-recover

.release-please-manifest.json
```diff
@@ -80,5 +80,6 @@
   "plugins/source/oracle": "1.0.1",
   "plugins/source/oracle+FILLER": "0.0.0",
   "plugins/source/alicloud": "1.0.0",
-  "plugins/source/alicloud+FILLER": "0.0.0"
+  "plugins/source/alicloud+FILLER": "0.0.0",
+  "plugins/destination/mssql": "1.0.0"
```
erezrokah, line 84:
```suggestion
  "plugins/source/alicloud+FILLER": "0.0.0"
```

https://github.com/cloudquery/cloudquery/blob/main/release-please-config.json is the file to update, not this one (this file is used by our bot to determine the next version so putting `1.0.0` here will generate a release PR for `1.0.1`)

yevgenypats (approved):

erezrokah (changes requested):

.release-please-manifest.json
candiduslynx: upd in #6973

PR #6971, opened 2023-01-19
chore: Add workflow to keep them in sync

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/blob/main/.github/workflows/endpoints.yml.
This PR creates a workflow to periodically update https://github.com/cloudquery/cloudquery/blob/796e82a30436e0df3a0d00fa82daad98cec015fb/plugins/source/gcp/client/services.go


plugins/source/gcp/client/services.go
```diff
@@ -149,7 +149,6 @@ var GcpServices = map[string]bool{
 	"iam.googleapis.com":                           true,
 	"iamcredentials.googleapis.com":                true,
 	"iap.googleapis.com":                           true,
-	"ideahub.googleapis.com":                       true,
```
erezrokah, line 152: We don't use this so it's not breaking

.github/workflows/gcp-services.yml
```diff
@@ -0,0 +1,36 @@
+name: Generate GCP Services
+on:
+  schedule:
+    - cron: "60 * * * *"
```
yevgenypats, line 4: is it every hours? can we do weekly basis?

erezrokah: This is actually wrong syntax 🤦 
I'll update it to run every Monday morning

erezrokah:
I'll go ahead and merge this as it's blocking another PR

PR #6978, opened 2023-01-19
Simplify file destinations

bbernays:


bbernays:
This pr relies on [#cloudquery/filetypes](https://github.com/cloudquery/filetypes/pull/35)

PR #6982, opened 2023-01-20
feat(clickhouse): ClickHouse destination

candiduslynx:
Implements #6254


plugins/destination/clickhouse/README.md
```diff
@@ -0,0 +1,7 @@
+# CloudQuery Microsoft SQL Server Destination Plugin
```
disq, line 1:
```suggestion
# CloudQuery ClickHouse Destination Plugin
```

plugins/destination/clickhouse/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+This destination plugin lets you sync data from any CloudQuery source to a Microsoft SQL Server database.
```
disq, line 3:
```suggestion
This destination plugin lets you sync data from any CloudQuery source to a ClickHouse database.
```

plugins/destination/clickhouse/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+## Links
+
+- [User Guide](https://cloudquery.io/docs/plugins/destinations/mssql/overview)
```
disq, line 7:
```suggestion
- [User Guide](https://cloudquery.io/docs/plugins/destinations/clickhouse/overview)
```

release-please-config.json
```diff
@@ -72,6 +72,9 @@
     "plugins/destination/mssql": {
       "component": "plugins-destination-mssql"
     },
+    "plugins/destination/clickhouse": {
+      "component": "plugins-destination-mssql"
```
disq, line 76:
```suggestion
      "component": "plugins-destination-clickhouse"
```

candiduslynx:
> Looks good! I did only a very quick walkthrough. Can you update the migration code to work similarly to the new way in my pg PR - #7819
> 
> This should make migration faster and easier to understand across plugins.

@yevgenypats 
I can update, but ClickHouse won't allow multi-statement execution unless configured in the database, so it'll still be better off to be a simple bunch of statements per table executed one by one.
The migration is still really fast, so it's not a big concern here.

yevgenypats:
> > Looks good! I did only a very quick walkthrough. Can you update the migration code to work similarly to the new way in my pg PR - #7819
> > This should make migration faster and easier to understand across plugins.
> 
> @yevgenypats I can update, but ClickHouse won't allow multi-statement execution unless configured in the database, so it'll still be better off to be a simple bunch of statements per table executed one by one. The migration is still really fast, so it's not a big concern here.

There is no need for multi-statement? Can you query the catalog all at once? In any case even if you do multiple statement, can you transform the result of each table definition to `Table` and use the new helper functions like `AddedColumns` `ChangedColumns` and so on.

candiduslynx:
The reason why we store JSON as string: https://github.com/ClickHouse/ClickHouse/issues/46190

candiduslynx:
> can you transform the result of each table definition to `Table` and use the new helper functions like `AddedColumns` `ChangedColumns` and so on.

@yevgenypats we have several types mapping to `String` column type, so currently there's no way to distinguish the original cq type.

It might've worked if we were passing the mapping func from cq type to db column type to the comparison, so that we just check if the column in the database corresponds to the proper type based on the table definition, but, for now, not possible.

yevgenypats (changes requested):
Same comment still stands. Any chance to normalize the ClickHouse table to the same CloudQuery schema tables instead of creating and maintaining the exact same struct and logic to detect schema changes?  See example - https://github.com/cloudquery/cloudquery/pull/7819

yevgenypats (changes requested):
The diff logic looks better now.

Next major comment - Let's please drop the use of templating it's an overkill and an un-necessary abstraction because we are not building a library/query-builder here but rather a single destination plugin so queries are only used in that plugin.

yevgenypats:
> The diff logic looks better now.
> 
> Next major comment - Let's please drop the use of templating it's an overkill and an un-necessary abstraction because we are not building a library/query-builder here but rather a single destination plugin so queries are only used in that plugin.

Moreover, templates are notoriously hard to debug as you can't single-step it, so unless there is a really good reason we don't really want to use it and this case there isn't really a reason to use this instead of regular string builder or string concatination.

plugins/destination/clickhouse/queries/test_data/col_add.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `table_name` ADD COLUMN `my_col` Int64
```
hermanschaaf, line 1: Let's rename this directory to the more standard `testdata` (no underscore) so that it gets ignored by the `go` tool.

website/pages/docs/plugins/destinations/clickhouse/configuration.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+import { getLatestVersion } from "../../../../../utils/versions";
+import { Badge } from "../../../../../components/Badge";
+import { Callout } from 'nextra-theme-docs'
+
+<Badge text={"Latest: " + getLatestVersion("destination", "clickhouse")}/>
+
+<Callout type="info">
+Make sure you use environment variable expansion in production instead of committing the credentials to the configuration file directly.
+</Callout>
```
hermanschaaf, line 11: This callout should probably be lower down (below the example), it doesn't need to be the first thing on the page

website/pages/docs/plugins/destinations/clickhouse/configuration.mdx
```diff
@@ -0,0 +1,67 @@
+# ClickHouse destination plugin configuration reference
```
hermanschaaf, line 1: Let's move the contents of this file to the overview page and delete this one, there's no need to split them at this stage

plugins/destination/clickhouse/queries/types.go
```diff
@@ -0,0 +1,163 @@
...
+	res := schema.Column{Name: name}
+
+	if strings.HasPrefix(typ, "Array(") {
+		switch typ {
+		case "Array(Nullable(Int64))":
+			res.Type = schema.TypeIntArray
+		case "Array(Nullable(String))":
+			res.Type = schema.TypeStringArray
+		case "Array(Nullable(UUID))":
+			res.Type = schema.TypeUUIDArray
```
hermanschaaf, line 74: We may want to have a default case here that panics, like in other switch statements

hermanschaaf (approved):
LGTM 👍  left some minor nice-to-have comments and a few questions

plugins/destination/clickhouse/client/unescape.go
```diff
@@ -0,0 +1,35 @@
+package client
+
+import (
+	"strconv"
+	"strings"
+)
+
+func unescape(str string) string {
```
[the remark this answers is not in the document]
candiduslynx: it was copied from MS SQL (that has tests for it), and it really should be a part of plugin-sdk per https://github.com/cloudquery/plugin-sdk/issues/622

plugins/destination/clickhouse/client/transform.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (*Client) TransformUUIDArray(v *schema.UUIDArray) any {
+	if v.Status != schema.Present {
+		return []*uuid.UUID(nil)
+	}
+
+	res := make([]uuid.UUID, len(v.Elements))
+	for i, el := range v.Elements {
+		res[i] = el.Bytes
+	}
```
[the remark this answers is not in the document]
candiduslynx: the driver still handles both types so no problem there.
However, I'll change the nil type for consistency

plugins/destination/clickhouse/client/unescape.go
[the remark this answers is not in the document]
hermanschaaf: Ah I see, but for now can we comment and copy the test here?

disq (approved):

yevgenypats (approved):
Great job!

PR #7010, opened 2023-01-23
fix(mssql)!: Fix JSON transformer & column types

### LINKED ISSUE #7001: Microsoft SQL Server destination escapes `&` symbol in data
`{"BillTo":"P\u0026T"}` instead of `{"BillTo":"P&T"}`. This works properly with Postgres destination.

### Expected Behavior
The data should be stored as `{"BillTo":"P&T"}`.

### CloudQuery Version
| Component | Version |
| --- | --- |
| `cloudquery` | `v2.2.0` |
| `azure` | `v3.4.0`|
| `mssql` |  `v1.0.0` |

### References
See #6984 for originally reported issue.

candiduslynx:
Fixes #7001

BEGIN_COMMIT_OVERRIDE
fix(mssql): Change column types from `varchar` to `nvarchar` to properly store Unicode characters
BREAKING-CHANGE: Change column types from `varchar` to `nvarchar` to properly store Unicode characters

fix(mssql): Fix JSON transformer to store unescaped data
END_COMMIT_OVERRIDE


plugins/destination/mssql/client/unicode.go
```diff
@@ -0,0 +1,25 @@
+package client
+
+import (
+	"strconv"
+)
+
+func unescape(str string) string {
```
erezrokah, line 7: Can we add some test cases for this logic? That would make it easier to understand what problem this code is solving

erezrokah (changes requested):
The change looks good, but can we add some tests to the unescape code?

plugins/destination/mssql/client/unicode.go
candiduslynx: done in 6e83ed0

erezrokah (approved):

Issue #7004, opened 2023-01-23
chore(deps): Update Nextra to the official version 


erezrokah:
There's a PR up for this but it's failing https://github.com/cloudquery/cloudquery/pull/6173.
We'll need to handle this manually.

Reasons to upgrade:
1. Project is out of beta (we're still using the beta)
2. We've been waiting on https://github.com/shuding/nextra/issues/709, which should improve build times for us and might allow us to add the tables + columns to our Website build
3. https://github.com/shuding/nextra/issues/852 should allow us to avoid manually configuring entries in `_meta.json` files
4. Don't think we need this now, but https://github.com/shuding/nextra/issues/181 might come in handy

erezrokah:
This is done

PR #7014, opened 2023-01-23
fix(website): Add Plausible Analytics to meta.json

erezrokah:
#### Summary

I think the plugin needs to be in the `meta.json` as a route, otherwise our versions find replace logic doesn't work


erezrokah:
I think number 3. point in https://github.com/cloudquery/cloudquery/issues/7004 should allow us to avoid manually editing the `meta.json` as we could dynamically load the plugins, sort them and get the correct title from the `overview.md`

yevgenypats (approved):

PR #7052, opened 2023-01-23
doc(plausible): Update plausible example to use sum instead of count

hermanschaaf:
Also fixes a missing backtick

### DIFF
--- a/website/pages/docs/plugins/sources/plausible/overview.md
+++ b/website/pages/docs/plugins/sources/plausible/overview.md
@@ -73,7 +73,7 @@ This is the (nested) spec used by this plugin:
 ### Sum all visitors since date
 
 ```sql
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

yevgenypats (approved):
Nice catch!

website/pages/docs/plugins/sources/plausible/overview.md
```diff
@@ -62,7 +62,7 @@ This is the (nested) spec used by this plugin:
 
 - `interval` (string array, optional, `date`):
     
-  Choose your reporting interval. Valid options are `date (always) and month (when specified period is longer than one calendar month).
+  Choose your reporting interval. Valid options are `date` (always) and month (when specified period is longer than one calendar month).
```
candiduslynx, line 65: Maybe also use buckticks around month?

PR #7053, opened 2023-01-23
fix(gcp-project-policies): Use correct API to get Policy v3, fix policy 2.1 query

erezrokah:
#### Summary

Reported on Discord https://discord.com/channels/872925471417962546/1065556442800734268/1065871250225909791

`projectsClient.GetIamPolicy` returns `v1.Policy` which has `log_type` as an enum, and as a result evaluated to an int in the destination database. Our query needs the strings, as they appear in `v3.Policy` (configured by the table).
To get the `v3.Policy` we need the protobuf API.

Additionally the check was wrong. We should verify all logging types are enabled for `allServices`.
This is how it looks in GCP:
image


yevgenypats (approved):

erezrokah:
I might do another follow to the query as I think we should get all projects, and verify each one has audit logs.
The current query won't catch projects that don't have any audit logs enabled

PR #7079, opened 2023-01-24
feat(s3): Add Athena flag to S3 destination

hermanschaaf:
This adds an `athena` flag to the S3 destination. When enabled, the destination will sanitize JSON keys to replace any invalid characters with `_`, which allows for any data to be queried using Athena.


yevgenypats (approved):

plugins/destination/s3/client/write.go
```diff
@@ -50,3 +64,24 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
 
 	return nil
 }
+
+// sanitizeJSONKeys replaces all invalid characters in JSON keys with underscores.
+// It does the replacement in-place, modifying the original object. This is required
+// for compatibility with Athena.
+func sanitizeJSONKeys(obj any) {
+	switch m := obj.(type) {
+	case map[string]any:
```
disq, line 73: This wouldn't match `map[string]string` etc, if we have those kinds of objects coming through here. Could be edge cases.

hermanschaaf: Hmm true, let me add some cases for this!

Issue #6984, opened 2023-01-21
MSSQL Destination PK violations & unicode characters


rafaelrodrigues3092:
### Describe the Bug
Observations after testing new MSSQL destination plugin:

1. Primary Key Violations
2. Unicode characters
I am using the Azure source plugin. In the tags column I am seeing issues with unicode characters
`{"BillTo":"P\u0026T"}` instead of `{"BillTo":"P&T"}`. This works properly with Postgres destination.

### Expected Behavior
No primary key violations, and data values with the actual special characters instead of the escaped unicode representations.

### CloudQuery Version
cloudquery version 2.2.0
azure (v3.4.0) -> mssql (v1.0.0)

### Debug Output

`
{"level":"error","module":"cli","class":14,"line_no":24,"message":"Violation of PRIMARY KEY constraint 'azure_network_azure_firewall_fqdn_tags_cqpk'. Cannot insert duplicate key in object 'dbo.azure_network_azure_firewall_fqdn_tags'. The duplicate key value is (/subscriptions//resourceGroups//providers/Microsoft.Network/azureFirewallFqdnTags/).","module":"dest-mssql","number":2627,"proc_name":"dbo.cq_proc_azure_network_azure_firewall_fqdn_tags","server_name":"<SQL_SERVER_INSTANCE>","state":1,"time":"2023-01-20T21:24:08Z"}
{"level":"error","module":"cli","duration":1765.3814,"error":"mssql: Violation of PRIMARY KEY constraint 'azure_network_azure_firewall_fqdn_tags_cqpk'. Cannot insert duplicate key in object 'dbo.azure_network_azure_firewall_fqdn_tags'. The duplicate key value is (/subscriptions//resourceGroups//providers/Microsoft.Network/azureFirewallFqdnTags/).","len":3,"message":"failed to write batch","table":"azure_network_azure_firewall_fqdn_tags","time":"2023-01-20T21:24:08Z"}
[...]

PR #7086, opened 2023-01-24
feat(mssql): Use `LEFT JOIN` for overwrite

candiduslynx:
As suggested [here](https://github.com/cloudquery/cloudquery/issues/6984#issuecomment-1401318407).
Preliminary testing shows up to 20% speedup for large batches.


disq (approved):
Good thing you used Go templating for the queries :)

Issue #7007, opened 2023-01-23
feat(gcp): Support SCC findings


erezrokah:
### Why do you need this resource or plugin? Please describe.

Requested by a user, see https://cloud.google.com/security-command-center/docs/reference/rest/v1/organizations.sources.findings/list

erezrokah:
https://github.com/cloudquery/cloudquery/pull/7089 added projects and organizations findings. What's missing is folders findings

PR #7089, opened 2023-01-24
feat(gcp-resources): Add Security Center projects and organizations findings

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery/issues/7007.
Relevant APIs:
- [Organizations](https://cloud.google.com/security-command-center/docs/reference/rest/v1/organizations.sources.findings/list)
- [Projects](https://cloud.google.com/security-command-center/docs/reference/rest/v1/projects.sources.findings/list)
- [Folders](https://cloud.google.com/security-command-center/docs/reference/rest/v1/folders.sources.findings/list)

This PR implements organizations and projects findings as we have multiplexers for those.
For folders we either need to:
1. List all projects like in https://github.com/cloudquery/cloudquery/blob/27696f9acd7f8d7e2828c160217962af4ce572fc/plugins/source/gcp/resources/services/resourcemanager/folders_fetch.go#L14 and make the findings a relation of folders
2. Have a folders multiplexer based on either the `folder_ids` config, or fetch all folders before syncing, see:
https://github.com/cloudquery/cloudquery/blob/27696f9acd7f8d7e2828c160217962af4ce572fc/plugins/source/gcp/client/client.go#L169

I'm currently leaning into 2. based on the config `folder_ids` as that is the quickest to implement and most consistent with the codebase.


plugins/source/gcp/resources/services/securitycenter/organization_findings.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/cloudquery/plugins/source/gcp/client"
+)
+
+func OrganizationFindings() *schema.Table {
+	return &schema.Table{
+		Name:        "gcp_securitycenter_organization_findings",
+		Description: `https://cloud.google.com/security-command-center/docs/reference/rest/v1/ListFindingsResponse#ListFindingsResult`,
+		Resolver:    fetchOrganizationFindings,
+		Multiplex:   client.OrgMultiplex,
+		Transform:   transformers.TransformWithStruct(&pb.ListFindingsResponse_ListFindingsResult{}, client.Options()...),
```
erezrokah, line 18: `ListFindingsResponse_ListFindingsResult` is a `struct` with `Finding`, `StateChange` and `Resource`.
`StateChange` is only returned if we pass `compareDuration` (see docs).
I thought about using the `Finding` struct directly and syncing only `response.Finding`, but it seems `Resource` has useful information.

To have a meaningful primary key I use a path resolver to access the finding name

disq (approved):

PR #7104, opened 2023-01-24
fix: Use reflection for sanitizeJSONKeys to cover more cases

hermanschaaf:
This covers more cases than before, such as `map[string]string` and `map[string]int`


disq (approved):
🪄

PR #7115, opened 2023-01-24
feat(scaffold): Update SDK, add auto deps updates

erezrokah:
#### Summary

This PR updates the plugin SDK for the scaffold CLI and also ensures it's gets dependency updates automatically going forward.
There's an existing test that runs the scaffold command and verifies the code builds, so that should ensure breaking changes are handled and not auto merged

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -11,4 +11,7 @@
     },
   ],
   ignorePaths: ["plugins/destination/csv/**"],
+  gomod: {
+    fileMatch: ["go\\.mod\\.tpl$"],
+  },
 }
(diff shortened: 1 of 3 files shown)


hermanschaaf (approved):
Nice! 💯

.github/renovate.json5
```diff
@@ -11,4 +11,7 @@
     },
   ],
   ignorePaths: ["plugins/destination/csv/**"],
+  gomod: {
+    fileMatch: ["go\\.mod\\.tpl$"],
```
erezrokah, line 15: I hope this works, as renovate also does some tidying after the update.
If it doesn't I'll need to find a way to exclude the template file from the tidying part

erezrokah: It works https://github.com/cloudquery/cloudquery/pull/7121/files#diff-24b7ec064fbad86a0563aef540f40e702d9797a62f53a834d126e797002f4665R6

PR #7118, opened 2023-01-24
feat(scaffold): Add Makefile and test workflow to scaffold, and flesh out readme

hermanschaaf:
This adds to the scaffold:
 - a unit test workflow
 - a makefile
 - more details to the readme


scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -0,0 +1,43 @@
+name: test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: [ main ]
```
erezrokah, line 8:
```suggestion
    branches: [main]
```

scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  test:
+    timeout-minutes: 30
+    name: "test"
```
erezrokah, line 13:
```suggestion
```

The default name is the id of the job, which is already `test`

scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: go.mod
+          cache: true
+          cache-dependency-path: go.sum
```
erezrokah, line 24:
```suggestion
```

I think that's the default already

scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -0,0 +1,43 @@
...
+      - name: golangci-lint
+        uses: cloudquery/golangci-lint-action@master
```
erezrokah, line 26: Let's use the official one? We only use our fork as caching is broken for the official one on monoreops

scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          version: v1.50.1
+          args: "--config .golangci.yml"
+          skip-pkg-cache: true
+          skip-build-cache: true
```
erezrokah, line 31:
```suggestion
```

I don't think we need those for non monorepos

scaffold/cmd/templates/source/Makefile.tpl
```diff
@@ -0,0 +1,16 @@
+.PHONY: test
+test:
+	go test -timeout 3m ./...
+
+.PHONY: lint
+lint:
+	@golangci-lint run --timeout 10m --verbose
```
erezrokah, line 7:
```suggestion
	@golangci-lint run --timeout 10m
```

Non blocking, I find that `--verbose` makes it very hard to find the actual errors reported by the linter

scaffold/cmd/templates/source/README.md.tpl
```diff
@@ -1 +1,51 @@
...
+
+```yaml
+kind: source
+spec:
+  name: "{{.Name}}"
+  path: "{{.Org}}/{{.Name}}"
+  version: "${VERSION}"
+  backend: "local" # remove this to disable incremental syncing
+  tables:
+    ["*"]
```
erezrokah, line 26:
```suggestion
```

Maybe we should only add the required fields? `backend` is in alpha and not needed for most plugins (yet)

scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
erezrokah: I think it's fine to use the defaults as linting rules can be quite subjective

PR #7174, opened 2023-01-25
doc: Change quickstart guide to fetch only some tables

shimonp21:


shimonp21:
Closing for now. Maybe we'll change this someday, but leaving as-is for now :)

PR #7177, opened 2023-01-25
fix(azure-resources): Use default API version of `armhealthbot` bots client

erezrokah:
#### Summary

The issue that required us to override the API version should be solved by the end of the week (see https://github.com/Azure/azure-sdk-for-go/issues/19790).

I tried testing the default API version and it still fails. I'll keep the PR open and test it again later on this week.


erezrokah:
> I tried testing the default API version and it still fails. I'll keep the PR open and test it again later on this week.

Tested again and it's working now. No need for the workaround so this PR is ready to be merged

hermanschaaf (approved):

PR #7179, opened 2023-01-25
feat(test): Add primary key violation check to test destination

### LINKED ISSUE #7175: Add primary key violation check to test destination

candiduslynx:
Implements #7175


yevgenypats (changes requested):
I don't think it should be done here. if we want this then it should be on the SDK side

candiduslynx:
I'll transfer this code to plugin-sdk for managed writers

PR #7184, opened 2023-01-25
fix: Fetch incidents with `dateRange="all"`, and add query examples to docs

shimonp21:
This fixes an issue where only the last month of incidents were fetched (see `since` parameter in https://developer.pagerduty.com/api-reference/9d0b4b12e36f9-list-incidents#Query-Parameters).

Also copied some query examples from blog to docs.


erezrokah (approved):
Unrelated to this PR, but we should remove all `// Code generated by codegen; DO NOT EDIT.` from the plugin

PR #7202, opened 2023-01-25
fix(aws)!: RDS Engine Version PKs

bbernays:
#### Summary

This is a unique resource... There is seemingly no set of fields that guarantee uniqueness... Also these values are not dependent on user input so they are 100% defined by AWS. For that reason I have just calculated a hash of the object and used that calculated hash to be the PK (along with region + account so users can determine what is available).


plugins/source/aws/resources/services/rds/engine_versions.go
```diff
@@ -32,25 +32,9 @@ func EngineVersions() *schema.Table {
 				},
 			},
 			{
-				Name:     "engine",
+				Name:     "_engine_version_hash",
```
bbernays, line 35: Don't love the name of this column... open to suggestions!

candiduslynx:
@bbernays I think the easiest way would be just adding `_cq_id` to the primary key.
So, the resulting pk should be: `account, region, engine, engine_version, _cq_id`

erezrokah:
> I think the easiest way would be just adding `_cq_id` to the primary key.

`_cq_id` changes on every sync so I'm not sure that would work in `overwrite` mode

candiduslynx:
> `_cq_id` changes on every sync so I'm not sure that would work in `overwrite` mode
Good catch!
Still, I want to keep engine & engine_version in PK, so that the only change is replacing `engine_version_description` with `_engine_version_hash`

plugins/source/aws/docs/tables/aws_rds_engine_versions.md
```diff
@@ -2,7 +2,7 @@
 
 https://docs.aws.amazon.com/documentdb/latest/developerguide/API_DBEngineVersion.html
 
-The composite primary key for this table is (**account_id**, **region**, **engine**, **engine_version**, **db_engine_version_description**).
+The composite primary key for this table is (**account_id**, **region**, **_engine_version_hash**).
```
candiduslynx, line 5: I think we'll be better off using `(account_id, region, engine, engine_version,  _engine_version_hash)`

plugins/source/aws/resources/services/rds/engine_versions.go
candiduslynx: i think it's fine for now

candiduslynx (changes requested):

hermanschaaf:
I don't understand: why is it better to use a hash of the engine version as the PK, rather than the engine version directly? (Or for that matter, if any set of fields could be hashed and make a decent PK, why don't we use that set of fields as the PK directly?)

candiduslynx:
@hermanschaaf 
> I don't understand: why is it better to use a hash of the engine version as the PK, rather than the engine version directly? (Or for that matter, if any set of fields could be hashed and make a decent PK, why don't we use that set of fields as the PK directly?)
`_engine_version_hash` is calculated based on the whole structure (not just a single field).

plugins/source/aws/resources/services/rds/engine_versions_fetch.go
```diff
@@ -24,3 +27,11 @@ func fetchRdsEngineVersions(ctx context.Context, meta schema.ClientMeta, _ *sche
 	}
 	return nil
 }
+func calculateUniqueHash(_ context.Context, _ schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	f := resource.Item.(types.DBEngineVersion)
+	hash, err := hashstructure.Hash(f, hashstructure.FormatV2, nil)
```
candiduslynx, line 32: does result differ if we pass just `resource.Item`?

candiduslynx (changes requested):

hermanschaaf:
> _engine_version_hash is calculated based on the whole structure (not just a single field).

Ah it's the whole item 🤦  Yeah sorry, I thought it was a single field.

plugins/source/aws/resources/services/rds/engine_versions_fetch.go
bbernays: I will look into it

erezrokah: My concern with this is that if a new field gets added to the struct, it will be added to the hash too.
For example a new `UpdateTime` field that will break the uniqueness.
On the other hand it could be considered a feature if the field does contribute to uniqueness 🙃 
Another use case is that AWS removes one field and adds another, then the hash changes.

Maybe we should hard code the list of fields that are a part of the hash and error if it changes?

candiduslynx: I think it's OK to leave full hash, as this is a despperate measure aimed solely at providing uniqueness to the resource PK.

bbernays: > does result differ if we pass just `resource.Item`?

Hash value is the same... So removed the casting

bbernays: I agree with @candiduslynx... Because this is fully defined by AWS and there is absolutely no guidance on uniqueness I think anything else that we do will leave us in a more brittle position

candiduslynx (approved):

plugins/source/aws/resources/services/rds/engine_versions_fetch.go
erezrokah: I think I didn't explain my suggestion properly.
It's very close to the existing solution.
I'm suggesting we have a hard coded list of all **current** fields of `DBEngineVersion` and use those as a hash.
Then if while creating the hash we detect a new/removed field we fail (our tests should catch this failure).
That way the current logic stays the same, but if even AWS changes the structure of `DBEngineVersion` we'll know.

candiduslynx: And **do** we want to know?

erezrokah: I can see 2 cases we'd like to know:
1. AWS adds a new field that can serve as a PK so we can drop this fix
2. AWS adds a new field that breaks uniqueness (like a timestamp) meaning this fix stops working

Regardless my comment is non blocking for the PR as the changes improve on the current state.

bbernays: Added an issue to track this... https://github.com/cloudquery/cloudquery/issues/7293

PR #7203, opened 2023-01-25
fix(aws): Built in Resolvers

bbernays:
#### Summary

Prior to this fix if a user tried to use `client.ResolveAWSAccount` on a column not on `account_id` it would fail because the resolver had the column name hard coded


erezrokah (changes requested):
Thanks for the PR @bbernays, can you share when do we want the account ID, region and namespace to be set on columns other than `account_id`, `region` and `namespace`?

I feel like I'm missing the use case here

hermanschaaf (approved):

bbernays:
> Thanks for the PR @bbernays, can you share when do we want the account ID, region and namespace to be set on columns other than `account_id`, `region` and `namespace`?
> 
> I feel like I'm missing the use case here

I ran into issues with Security Hub where the Region resolver was updating different columns from where it was called.

erezrokah (approved):

PR #7204, opened 2023-01-25
feat(aws): Add Support for Securityhub findings

bbernays:
#### Summary

The results from `GetFindings` include a `region` and `account_id` which identifies where the finding was actually generated NOT where it was pulled from. I have changed the name of the fields we add to `request_region` and `request_account_id` to disambiguate...


plugins/source/aws/resources/services/securityhub/findings.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func Findings() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_securityhub_findings",
+		Description: `https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_GetFindings.html`,
+		Resolver:    fetchFindings,
+		Transform:   transformers.TransformWithStruct(&types.AwsSecurityFinding{}, transformers.WithPrimaryKeys("AwsAccountId", "Region", "CreatedAt", "Description", "GeneratorId,Id", "ProductArn", "SchemaVersion", "Title")),
```
disq, line 15:
```suggestion
		Transform:   transformers.TransformWithStruct(&types.AwsSecurityFinding{}, transformers.WithPrimaryKeys("AwsAccountId", "Region", "CreatedAt", "Description", "GeneratorId", "Id", "ProductArn", "SchemaVersion", "Title")),
```

plugins/source/aws/resources/services/securityhub/findings.go
```diff
@@ -0,0 +1,30 @@
...
+func Findings() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_securityhub_findings",
+		Description: `https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_GetFindings.html`,
+		Resolver:    fetchFindings,
+		Transform:   transformers.TransformWithStruct(&types.AwsSecurityFinding{}, transformers.WithPrimaryKeys("AwsAccountId", "Region", "CreatedAt", "Description", "GeneratorId", "Id", "ProductArn", "SchemaVersion", "Title")),
+		Multiplex:   client.ServiceAccountRegionMultiplexer("securityhub"),
+		Columns: []schema.Column{
+			{
+				Name:     "request_account_id",
```
erezrokah, line 19: Can `AwsAccountId` and `Region` have different values from the ones in the multiplexer?
Might be confusing to have both so maybe keep only the struct ones (even if different)?

erezrokah (changes requested):
Looks great 🚀 One question also related to https://github.com/cloudquery/cloudquery/pull/7203

plugins/source/aws/resources/services/securityhub/findings.go
hermanschaaf: Yeah if I understood the PR description correctly, the values can be different from the one in the multiplexer, so I think this is a good compromise. @bbernays Maybe we can just add some details about this to the description of the table?

erezrokah: Ah 🤦 I missed that. Maybe we can drop the `request_` ones for now?

hermanschaaf: I think it depends: can security hub findings be different depending on the account/region the request is done from? If so, I think we should include the columns (and make them part of the PK?). If not, maybe we shouldn't be multiplexing on account and region

bbernays: Yes, Security Hub can aggregate findings from different regions based on user configuration

disq (approved):

plugins/source/aws/resources/services/securityhub/findings.go
bbernays: I will add a comment in the table description. Are there any other changes needed?

hermanschaaf (approved):
LGTM, will just need to run `make gen` again

plugins/source/aws/resources/services/securityhub/findings.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func Findings() *schema.Table {
+	return &schema.Table{
+		Name: "aws_securityhub_findings",
+		Description: `https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_GetFindings.html. 
+The 'request_account_id' and 'request_region' columns are added to show the account and region of where the request was made from. This is useful when multi region and account aggregation is enabled.`,
```
erezrokah, line 14: Nice touch 
giphy (2)

PR #7208, opened 2023-01-26
feat: New plugin-sdk features used for GCP

amanenk:
#### Summary

The pr changes the way of pk setting without overwriting columns in the structure.


erezrokah (approved):
Nice 🚀

Issue #7195, opened 2023-01-25
test: Add CLI test that runs on multiple sources


erezrokah:
To catch issues like https://github.com/cloudquery/cloudquery/pull/7192.

We should be able to add new test case here https://github.com/cloudquery/cloudquery/blob/5c8450154f9e93b2e71b7bffa359f4caa9976ad3/cli/cmd/sync_test.go

PR #7210, opened 2023-01-26
test(cli): Add multiple sources test

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7195


cli/cmd/sync_v0_2.go
```diff
@@ -114,7 +115,7 @@ func syncConnectionV0_2(ctx context.Context, cqDir string, sourceClient *source.
 			if err := destClients[i].Close(ctx); err != nil {
 				return fmt.Errorf("failed to close destination client for %s -> %s: %w", sourceSpec.VersionString(), destination.VersionString(), err)
 			}
-			failedWrites += destFailedWrites
+			atomic.AddUint64(&failedWrites, destFailedWrites)
```
erezrokah, line 118: So this test detected a data race 🙃  We can release as `fix` but seems not that important

erezrokah:
Apparently the data race was also in other versions of sync:
https://github.com/cloudquery/cloudquery/pull/7625/commits/723502ba359a2797b96c2a840de9f28223e743cc

Issue #6981, opened 2023-01-20
Pager Duty Source Plugin - Source spec config support for Team


rafaelrodrigues3092:
### Is your feature request related to a problem? Please describe.
When using the Pager Duty plugin I have encountered several API Throttling errors.
I was unable to get all the data I needed.

### Describe the solution you'd like
Introduce an optional **team** key in the source spec file config (https://www.cloudquery.io/docs/plugins/sources/pagerduty/overview).
This key would be used in the API calls to only retrieve records for this given team or teams:
This may help with:
- Large organizations where users may only want to know about incidents for certain teams
- API throttling, by potentially reducing the scope of the data pull

### Describe alternatives you've considered
I tried tunning the data pull as described here https://www.cloudquery.io/docs/advanced-topics/performance-tuning (including the experimental round-robin feature)

### Additional context
None

shimonp21:
@rafaelrodrigues3092 Hello, and thanks for reporting this issue!
I have created a PR (#7206) to add the `team_ids` option to the Spec. A few points I can think of in the meantime:

- The most common culprits for rate-limiting are the relational tables usually (because they require an API call **per resource**). A (not-ideal) workaround can be to skip them in your config:
  ```yaml
  skip_tables: [
  "pagerduty_incident_alerts",
  "pagerduty_incident_notes",
  "pagerduty_incident_log_entries",
  "pagerduty_ruleset_rules",
  "pagerduty_user_contact_methods",
  "pagerduty_user_notification_rules"]
  ```
  
- I believe the more robust solution here would be to add a rate-limiting mechanism in the plugin itself, so we don't exceed the ~1000 requests/minute allowed by pagerduty (https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTUz-rate-limiting#what-are-our-limits). This will of course mean slower syncs, but I believe that's better than being rate limited and not bringing the data at all.

rafaelrodrigues3092:
@shimonp21 
Thanks for this and for the pointers.
I will give it another try as soon as the new version is out.

rafaelrodrigues3092:
> @rafaelrodrigues3092 Hi, Released version `v1.3.0`, which has both fixes:
> 
> * `team_ids` filter in the Spec,
[...]

PR #7218, opened 2023-01-26
feat: PagerDuty Rate Limiting

shimonp21:
In my tests (toy account with 1000 incidents), this fixed the rate-limit errors that prevented cloudquery from getting data. Fetching '*' took 5m13s unfortunately.

I used the `RateLimitHttpClient` instead of just calling `limiter.Wait` before every call, since some calls are paginated by the SDK (i.e. the `for` loop is in the `pagerduty-go` sdk).

mention: https://github.com/cloudquery/cloudquery/issues/6981


website/pages/docs/plugins/sources/pagerduty/configuration.md
```diff
@@ -26,4 +26,10 @@ This is the (nested) spec used by the PagerDuty source plugin.
...
 
-  If specified, limits the sync to only resources related to the specified teams.
\ No newline at end of file
+  If specified, limits the sync to only resources related to the specified teams.
+
+- `max_requests_per_second` (int) (default: 10)
+  PagerDuty API is heavily rate-limited (900 requests/min = 15 requests/sec, across the entire organization). 
+  This option allows you to control the rate at which the plugin will make requests to the API. 
+  You can reduce this parameter in case you are still seeing rate limit errors (status code 429), or increase
+  it if your PagerDuty API quota is higher. See https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTUz-rate-limiting#what-are-our-limits for more info.
```
disq, line 35:
```suggestion
  it if your PagerDuty API quota is higher. See [PagerDuty rate limiting documentation](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTUz-rate-limiting#what-are-our-limits) for more info.
```

erezrokah (commented):
The code looks good @shimonp21. Have you considered using a retry with a backoff when the limit is reached (similar to [here](https://github.com/cloudquery/cloudquery/blob/888e6e201996f512ee277bab81fec2e60b51331c/plugins/source/digitalocean/client/helpers.go#L33))? Otherwise with the approach in this PR we can't guarantee we use the max available quota

shimonp21:
@erezrokah I considered it - but because the limit is **per org**  (https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTUz-rate-limiting#what-are-our-limits), I strongly want to avoid "using all available quota" :).
I really want to avoid locking up user's accounts because of "too many requests".

erezrokah (approved):

PR #7223, opened 2023-01-26
chore(website): Sort things

disq:


website/pages/docs/plugins/sources/overview.mdx
```diff
@@ -16,36 +16,36 @@ Official source plugins follow [release stages](#source-plugin-release-stages).
     type="source"
     plugins={
         [
-            { name: "Alibaba Cloud", stage: "Preview", id: `alicloud` },
             { name: "AWS", stage: "GA" },
```
erezrokah, line 19: These gets auto sorted already https://github.com/cloudquery/cloudquery/blob/d4d6fd1f7fae918a7eb70c4bc57de91829aa9e68/website/components/PluginsTable.tsx#L55, for `meta.json` I'm working on https://github.com/cloudquery/cloudquery/pull/6173 that should allow us to use a JS file for (e.g. `meta.js`) so we can sort automatically

disq: Would not sorting them on sources cause more merge conflicts? e.g. if everybody keeps adding to the end? Otherwise it's distributed to different parts of the file so it can hopefully auto merge? Maybe? If not I'll close this :)

erezrokah: We're not sorting so not to put a restriction that's not needed. You can add wherever.
I don't mind it being sorted, but I think it will quickly get unsorted :)

PR #7236, opened 2023-01-27
chore(website): Remove `title` from code blocks

candiduslynx:


candiduslynx:
Currently the code blocks with `copy` and title don't show properly.
@hermanschaaf @erezrokah 
See https://www.cloudquery.io/how-to-guides/how-to-load-infrastructure-data-into-athena

hermanschaaf:
Nice catch! I'm fairly sure this used to work... was it broken by a recent upgrade maybe?

hermanschaaf (approved):

hermanschaaf:
Let's remove it for now so the code blocks render at least

candiduslynx:
@erezrokah could you take a look into that? Additionally, do we need to update other docs as well?

hermanschaaf:
@candiduslynx It seems like it's mostly me who's been using this feature, so only some docs I touched have it. But some are quite important, like the AWS overview. So if it's not easy to fix we should remove it from there as well for the time being

hermanschaaf:
I'll raise a PR to remove it for the time being

PR #7262, opened 2023-01-27
fix(azure)!: Add `subscription_id` to `azure_support_services` PK

### LINKED ISSUE #7258: `azure_support_services` primary key violations

disq:
Fixes #7258


candiduslynx (approved):

erezrokah:
https://github.com/cloudquery/cloudquery/blob/729f0fa662f359ec4730379371bdc3e47ffcefde/plugins/source/azure/resources/services/support/services.go#L25

PR #7265, opened 2023-01-27
feat: HubSpot rate limiter

shimonp21:
Same as https://github.com/cloudquery/cloudquery/pull/7218


erezrokah (approved):

plugins/source/hubspot/client/spec.go
```diff
@@ -1,5 +1,19 @@
 package client
 
 type Spec struct {
-	// plugin spec goes here
+	MaxRequestsPerSecond *int `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
```
disq, line 4: Why ptr?

disq: Also yaml tag is not needed

PR #7266, opened 2023-01-27
chore(azure): Do less ad-hoc transformations

candiduslynx:


yevgenypats (approved):
Looks good from quick glance.

PR #7267, opened 2023-01-27
fix(azure)!: Change `azure_compute_skus` PK from `id` to `(subscription_id, family, kind, name)`

### LINKED ISSUE #7245: `azure_compute_skus` primary key violations
See #7183 for more details.

candiduslynx:
Fixes #7245


yevgenypats (changes requested):
Let's add id back as a pk to keep backward compatability

candiduslynx:
@yevgenypats 
> Let's add id back as a pk to keep backward compatibility

I don't think it's right, given that the table is most probably little used.
For starters, it's not present in any of the policies.

yevgenypats (approved):

erezrokah:
Trying to run this on our playground account generates quite a few errors for me:
image

Looks like the defined primary keys can be `nil`. Am I missing something? Can you please verify?

PR #7270, opened 2023-01-27
fix(azure)!: Add `subscription_id` to `azure_authorization_role_assignments`

### LINKED ISSUE #7242: `azure_authorization_role_assignments` primary key violations
See #7183 for more details.

candiduslynx:
Fixes #7242


disq (approved):

erezrokah:
Can you explain how this fixes PK violations for the table, Azure IDs already have the subscription id as a part of the id, so adding `subscription_id` should not prevent existing duplciates:
Screenshot 2023-01-28 at 18 24 14

I do think `subscription_id` should be a primary key to be consistent with other tables, but maybe we should investigate this more as I'm not sure how this can fix duplicates (if we see them)

PR #7271, opened 2023-01-27
fix(azure)!: Add `subscription_id` to `azure_cdn_edge_nodes` PK

### LINKED ISSUE #7243: `azure_cdn_edge_nodes` primary key violations
See #7183 for more details.

candiduslynx:
Fixes #7243


disq (approved):

erezrokah:
https://github.com/cloudquery/cloudquery/blob/729f0fa662f359ec4730379371bdc3e47ffcefde/plugins/source/azure/resources/services/cdn/edge_nodes.go#L25

PR #7280, opened 2023-01-27
fix(azure)!: Add `subscription_id` to `azure_security_topology` PK

### LINKED ISSUE #7257: `azure_security_topology` primary key violations

candiduslynx:
Fixes #7257


disq (approved):

erezrokah:
Can you explain how this fix prevents duplicates? Based on the example responses in https://learn.microsoft.com/en-us/rest/api/defenderforcloud/topology/list?tabs=HTTP the topology Id has the subscription Id as a part of it (e.g. `"id": "/subscriptions/3eeab341-f466-499c-a8be-85427e154bad/resourceGroups/myservers/providers/Microsoft.Security/locations/centralus/topologies/subnets"`).

While making the subscription Id a primary key makes sense for consistency with other tables, I'm not sure how this fixes any PK violations we have

PR #7284, opened 2023-01-27
fix(azure)!: Add `subscription_id` to `azure_reservations_reservation` PK

### LINKED ISSUE #7254: `azure_reservations_reservation` primary key violations

candiduslynx:
Fixes #7254


candiduslynx:
Given that the `id` should've been quite unique already, the feasible option is to add subscription id as well

disq (approved):

erezrokah:
https://github.com/cloudquery/cloudquery/blob/729f0fa662f359ec4730379371bdc3e47ffcefde/plugins/source/azure/resources/services/reservations/reservation.go#L25

PR #7296, opened 2023-01-28
feat(k8s): Support running cloudquery k8s plugin in k8s cluster using serviceaccount

lshmouse:
#### Summary

Support running cloudquery k8s plugin in k8s cluster using serviceaccount.

We are using argo workflow(https://argoproj.github.io/argo-workflows/) to schedule the cloudquery k8s resource syncing every day in the k8s pod.  Currently the client in k8s plugin only support kubeconfig env,  so we need the pack the kubeconfig into the docker image or put it in a configmap, which is not the k8s-way. 

So we want to add the support to create the client  in k8s plugin  using k8s serviceaccount.

The changes in the PR is a little hack. Suggestions are welcomed. Thanks~

### DIFF
--- a/plugins/source/k8s/client/client.go
+++ b/plugins/source/k8s/client/client.go
@@ -13,6 +13,7 @@ import (
 
 	// import all k8s auth options
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
 )
@@ -109,7 +110,7 @@ func Configure(ctx context.Context, logger zerolog.Logger, s specs.Source, _ sou
 
 	for _, ctxName := range contexts {
 		logger.Info().Str("context", ctxName).Msg("creating k8s client for context")
-		kClient, err := buildKubeClient(rawKubeConfig, ctxName)
+		kClient, err := buildKubeClient(logger, rawKubeConfig, ctxName)
 		if err != nil {
 			return nil, fmt.Errorf("failed to build k8s client for context %q: %w", ctxName, err)
 		}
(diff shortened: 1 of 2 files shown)


plugins/source/k8s/client/client.go
```diff
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"k8s.io/client-go/rest"
```
candiduslynx, line 6: please format with `goimports`

plugins/source/k8s/client/client.go
```diff
@@ -133,6 +134,10 @@ func buildKubeClient(kubeConfig api.Config, ctx string) (*kubernetes.Clientset,
 		&clientcmd.ClientConfigLoadingRules{},
 	)
 	restConfig, err := clientConfig.ClientConfig()
+	if err != nil {
+		logger.Warn().Msg("Failed to create k8s client, fallback to use the in-cluster config")
+	}
+	restConfig, err = rest.InClusterConfig()
```
candiduslynx, line 140: you're always overwriting `restConfig`

plugins/source/k8s/config/cloudquery-k8s-cluster-role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
candiduslynx, line 1: if that's a test file it should either be removed or it should be used in some test

candiduslynx (changes requested):
Hi @lshmouse!
Thanks for thee PR.
I think it needs a bit more work (se the comments to the PR).

lshmouse:
> Hi @lshmouse! Thanks for the PR. I think it needs a bit more work (se the comments to the PR).

Thanks for your reviewing, i will update the PR soon.

lshmouse:
Update for @candiduslynx 's review. Thanks~

candiduslynx (approved):

candiduslynx:
@lshmouse thank you for the contribution!

Issue #5486, opened 2022-12-07
feat(website): Find a way to add `copy` button to code blocks with version templates


erezrokah:
Follow up to https://github.com/cloudquery/cloudquery/issues/5483 and https://github.com/cloudquery/cloudquery/pull/5485.

The following doesn't work for code blocks copy pasting:
https://github.com/cloudquery/cloudquery/blob/f53f79c8975327f420a2339c07d1ed68013ec6d0/website/next.config.mjs#L85

erezrokah:
This issue was solved with this recent upgrade https://github.com/cloudquery/cloudquery/pull/6173 so I'll bring those back

PR #7303, opened 2023-01-29
fix(website): Add missing copy blocks

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5486. Brings back `copy` to code blocks with version template (and others), as this works in the new Nextra version, see https://github.com/cloudquery/cloudquery/issues/5486#issuecomment-1407641040


yevgenypats (approved):
🚀

PR #7339, opened 2023-01-30
feat(azure): Add single subscription multiplexing by registered namespace

candiduslynx:


erezrokah:
We verified this and APIs that don't need subscription id as a parameter don't require the service to be registered

PR #7347, opened 2023-01-30
fix(aws): PK for `aws_apigateway_domain_name_base_path_mappings`

bbernays:


hermanschaaf:
Looking at `resolveApigatewayDomainNameBasePathMappingArn`, it seems like `AccountID` is empty, even though we multiplex on account ID. Should we maybe keep account ID and region in the PK as well to be safe? Or do you think, because it's a domain name, it should be unique across accounts? 🤔

hermanschaaf (approved):

PR #7350, opened 2023-01-30
fix(aws)!: PK for `aws_apigateway_rest_api_authorizers`

bbernays:


plugins/source/aws/resources/services/apigateway/rest_api_authorizers.go
```diff
@@ -19,6 +19,9 @@ func RestApiAuthorizers() *schema.Table {
 				Name:     "account_id",
 				Type:     schema.TypeString,
 				Resolver: client.ResolveAWSAccount,
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
+				},
 			},
 			{
 				Name:     "region",
```
hermanschaaf, line 27: I think region should probably also be part of the PK, because we use the account-region multiplexer

hermanschaaf: Oh no, it's part of the ARN, nevermind

hermanschaaf (approved):

PR #7371, opened 2023-01-30
fix: Remove primary keys options for non existent fields

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/7334. GCP fails on the new SDK as it references non existing fields as PKs.
This PR removes that option (I'll see if we need to add PKs separately)


yevgenypats (approved):

website/pages/docs/plugins/sources/gcp/policies.md
```diff
@@ -8,7 +8,7 @@ This page documents the available CloudQuery SQL Policies for GCP. See the [read
 ### Requirements
 GCP CIS v1.2.0 requires the following tables to be synced before the policy is executed:
 
-```yaml copy
```
erezrokah, line 11: I added this in https://github.com/cloudquery/cloudquery/pull/7303, but didn't realize it's auto generated. I'll update the policy script separately

PR #7379, opened 2023-01-30
fix(aws)!: PK for `aws_apigateway_vpc_links`

bbernays:


hermanschaaf (approved):

plugins/source/aws/resources/services/apigateway/vpc_links.go
```diff
@@ -19,6 +19,9 @@ func VpcLinks() *schema.Table {
 				Name:     "account_id",
 				Type:     schema.TypeString,
 				Resolver: client.ResolveAWSAccount,
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
```
disq, line 23: Y this when `arn` is the PK?

hermanschaaf: The ARN doesn't include the account id

PR #7394, opened 2023-01-30
fix(aws)!: PK for `aws_eventbridge_event_bus_rules`

bbernays:


plugins/source/aws/resources/services/eventbridge/event_bus_rules.go
```diff
@@ -35,6 +35,13 @@ func EventBusRules() *schema.Table {
 				Type:     schema.TypeJSON,
 				Resolver: resolveEventbridgeEventBusRuleTags,
 			},
+			{
+				Name: "arn",
```
candiduslynx, line 39: use [transformers.WithPrimaryKeys](https://pkg.go.dev/github.com/cloudquery/plugin-sdk/transformers#WithPrimaryKeys) option and not the extra column.
Additionally, should drop the `event_bus_arn` column as duplicate

bbernays: `event_bus_arn` is the parent arn...

candiduslynx: Ah, sorry, then dismissing that part.
The transformation & ad-hoc columns still stand, though

candiduslynx (approved):

plugins/source/aws/resources/services/eventbridge/event_bus_rules.go
```diff
@@ -35,6 +35,13 @@ func EventBusRules() *schema.Table {
...
+				Type: schema.TypeString,
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
+				},
```
hermanschaaf, line 43: I think we'll need to specify the resolver here

hermanschaaf (changes requested):
Looks like there's a missing path resolver, but let me know if I misunderstood somethig

plugins/source/aws/resources/services/eventbridge/event_bus_rules.go
bbernays: If I needed the resolver, wouldn't the unit tests fail?

candiduslynx: [This](https://github.com/cloudquery/plugin-sdk/blob/main/transformers/struct.go#L258-L260) should skip adding from struct with default transformer.
So the following [code](https://github.com/cloudquery/plugin-sdk/blob/9f1e373b516be958f0594e84fbbbcd43951f14ad/plugins/source/scheduler.go#L117-L133) works instead (col resolver = `nil`).

I think it's better to:
1. Add resolver explicitly for explicitly defined columns
2. Add a check (in transform? or during plugin validate phase? but better in a static place) that all columns have resolvers defined.
3. Remove fallback for case with `nil` column resolver.

hermanschaaf (approved):

PR #7456, opened 2023-01-31
feat: Update file destination to latest filetype module

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/6958 for the file destination


erezrokah:
One thing I'm not sure about is if I should create a default file spec as in:
https://github.com/cloudquery/cloudquery/blob/87cc46bc244552a5abf4e88369d122f6866be3c5/plugins/destination/kafka/client/spec.go#L20

It seems it's not needed as the file types client does it.
Also other destinations don't do it:
https://github.com/cloudquery/cloudquery/blob/87cc46bc244552a5abf4e88369d122f6866be3c5/plugins/destination/gcs/client/spec.go#L16
https://github.com/cloudquery/cloudquery/blob/87cc46bc244552a5abf4e88369d122f6866be3c5/plugins/destination/s3/client/spec.go#L19

disq (approved):

erezrokah:
Replaced by https://github.com/cloudquery/cloudquery/pull/7466

PR #7465, opened 2023-01-31
draft: Onfleet plugin

shimonp21:
Draft PR.
Contains all the code/tables, but no documentation/release files (since those will be different, depending on whether this stays in the monorepo or moves to the `onfleet` github org).


plugins/source/onfleet/client/resolvers.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/thoas/go-funk"
+)
+
+// Resolves an unix-timestamp-with-millies field to a timestampo field
```
disq, line 12:
```suggestion
// Resolves an unix-timestamp-with-millies field to a timestamp field
```

plugins/source/onfleet/client/spec.go
```diff
@@ -0,0 +1,5 @@
+package client
+
+type Spec struct {
+	// plugin spec goes here
```
disq, line 4: No API key option in the spec? Only Env vars?

plugins/source/onfleet/resources/services/tasks.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func Tasks() *schema.Table {
+	return &schema.Table{
+		Name:        "onfleet_tasks",
+		Resolver:    fetchTasks,
+		Description: "https://docs.onfleet.com/reference/list-tasks",
+		Transform: transformers.TransformWithStruct(&onfleet.Task{},
+			transformers.WithPrimaryKeys("Id"),
+			transformers.WithTypeTransformer(client.OnfleetTypeTransformer),
```
disq, line 17: You can use the 'SharedTransformers' pattern here, moving WithType and WithResolver to a single function in `client/transformers.go` and just specifying primarykeys as the option

plugins/source/onfleet/client/spec.go
shimonp21: Putting plaintext API keys in plaintext files always seems not-great to me :) We also don't allow it in our big plugins AFAIK (aws, gcp
,...). 
It's a config option I'm happy to omit.

shimonp21: I.e. I think this feature can wait until a cutsomer requests it (hopefully they won't).

PR #7466, opened 2023-01-31
fix(destination-file): Update to new filetypes modules

erezrokah:
#### Summary

Replaces https://github.com/cloudquery/cloudquery/pull/7456

### DIFF
--- a/plugins/destination/file/client/client.go
+++ b/plugins/destination/file/client/client.go
@@ -19,11 +19,11 @@ type Client struct {
 	pluginSpec Spec
 
 	CSVClient              *csv.Client
-	CSVTransformer         csv.Transformer
-	CSVReverseTransformer  csv.ReverseTransformer
 	JSONClient             *json.Client
-	JSONTransformer        json.Transformer
-	JSONReverseTransformer json.ReverseTransformer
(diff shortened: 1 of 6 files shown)


plugins/destination/file/client/client.go
```diff
@@ -18,23 +18,21 @@ type Client struct {
...
-	csvTransformer         *csv.Transformer
-	csvReverseTransformer  *csv.ReverseTransformer
-	jsonTransformer        *json.Transformer
-	jsonReverseTransformer *json.ReverseTransformer
+	CSVClient              *csv.Client
+	CSVTransformer         csv.Transformer
+	CSVReverseTransformer  csv.ReverseTransformer
+	JSONClient             *json.Client
+	JSONTransformer        json.Transformer
+	JSONReverseTransformer json.ReverseTransformer
```
bbernays, line 26: Is there a reason why you needed to change the transformers?

erezrokah: Nice catch, no need to do it, left over from the old PR. Fixed in https://github.com/cloudquery/cloudquery/pull/7466/commits/2b3a8beabfdd5b780fd686b78ef46cdb8e23e759

PR #7468, opened 2023-01-31
fix(github-resources)!: Resolver Org Hooks request/response using GetHookDelivery

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/468.
We need to `Get` the hook to resolve the request and response

BEGIN_COMMIT_OVERRIDE
fix(github-resources): Resolver Org Hooks request/response using GetHookDelivery

BREAKING-CHANGE: The `request` and `response` columns for `github_hook_deliveries` are now of type `JSON` instead of `String`.
END_COMMIT_OVERRIDE


erezrokah:
@candiduslynx FYI made this a breaking change as the column types changed

PR #7608, opened 2023-02-01
feat: Add `.gitignore` template

erezrokah:
#### Summary

GoReleaser fails if git is in dirty state:
https://github.com/cloudquery/cq-source-plaid/actions/runs/4063182434/jobs/6995116392#step:6:27

This PR is based on https://github.com/github/gitignore/blob/main/Go.gitignore and adds the source binary and `dist` directory (GoReleaser output dir)


hermanschaaf:
Do we maybe want to mark this as `chore` @erezrokah ?

erezrokah:
> Do we maybe want to mark this as `chore` @erezrokah ?

I'm not sure I understand. This is a feature for the scaffold command, so when you run it, it will create `.gitignore` file for the newly generated plugin

scaffold/cmd/templates/source/.gitignore.tpl
```diff
@@ -0,0 +1,22 @@
...
+*.dylib
+
+# Test binary, built with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Dependency directories (remove the comment below to include it)
+# vendor/
```
candiduslynx, line 15: need to ignore vendor

candiduslynx (changes requested):

scaffold/cmd/templates/source/.gitignore.tpl
erezrokah: Can you please provide more context on your comment?
This is based on the GitHub template https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Go.gitignore#L18, and to my understanding subjective based on the use case.

Also should this be blocking for this PR?

hermanschaaf:
> I'm not sure I understand. This is a feature for the scaffold command, so when you run it, it will create .gitignore file for the newly generated plugin (it's not adding a .gitignore for the Scaffold module itself)

Ah sorry, I misunderstood! Makes sense 👍

hermanschaaf (approved):

scaffold/cmd/templates/source/.gitignore.tpl
erezrokah: I'll go ahead a merge this and we can do a follow up PR if it makes sense

candiduslynx: Basically not ignoring vendor means that you commit the dependencies into your repo. It's not the case here.

candiduslynx (approved):
I'm approving, but I really think we should add vendor to the gitignore as [here](https://github.com/cloudquery/cloudquery/blob/main/.gitignore#L11)

scaffold/cmd/templates/source/.gitignore.tpl
erezrokah: Got it, I think `vendor` is subjective (based on a quick research) as if you want to guarantee future reproducible builds you'll have to commit the vendor directory.
I tend to go with what ever the default is in GitHub mostly based on trust 🙃

PR #7617, opened 2023-02-01
chore(ci): Add workflow to set sentry alerts to Slack

erezrokah:
#### Summary

In order to send Sentry alerts to Slack we need to create one per project (i.e. plugin). That's a bit tedious so this workflow does it for us periodically so new plugins won't be missing alerts.
If the alert exists it skips it (so it won't override any manual configuration).


hermanschaaf (approved):
You and your army of bots 🤖 💯

PR #7618, opened 2023-02-01
chore(ci): Log matching runs and actions in waiting workflow

hermanschaaf:
This workflow is running forever, even though it says it's not waiting on anything. Not sure how that can happen, but hopefully this change will prevent such issues https://github.com/cloudquery/cloudquery/actions/runs/4066036975/jobs/7002516872


erezrokah:
Let's close this as I'm not sure this will work as we have runs with overlapping names (e.g `validate-release` and `policies-tests`)

hermanschaaf:
@erezrokah I updated it to log some more information so we can better debug next time this happens

erezrokah (approved):

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -135,7 +135,8 @@ jobs:
                 console.log("All required workflows have passed")
                 return
               }
-              
+              console.log(`Matching runs: ${matchingRuns}`)
```
erezrokah, line 138: Maybe log only the names here instead of the whole objects?

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -135,7 +135,8 @@ jobs:
...
+              console.log(`Actions: ${actions}`)
```
erezrokah, line 139: `actions` are a string array so this looks good

.github/workflows/wait_for_required_workflows.yml
hermanschaaf: Good idea

Issue #7596, opened 2023-02-01
Test: Postgres and CockroachDB tests are flaky 


erezrokah:
See https://github.com/cloudquery/cloudquery/actions/runs/4060635450/jobs/6989920149

We should fix it

PR #7620, opened 2023-02-01
fix: Sort PKs before reporting stale PKs error

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7596


plugins/destination/postgresql/client/migrate.go
```diff
@@ -63,8 +65,10 @@ func (c *Client) isTableExistSQL(ctx context.Context, table string) (bool, error
 
 func (c *Client) getStalePks(pgPKs map[string]bool, table *schema.Table) []string {
 	stalePks := []string{}
+	sortedPKs := maps.Keys(pgPKs)
+	sort.Strings(sortedPKs)
 	if c.enabledPks() {
```
hermanschaaf, line 70:
```suggestion
	if c.enabledPks() {
		sortedPKs := maps.Keys(pgPKs)
		sort.Strings(sortedPKs)
```

(nit, but I think this helps both efficiency and readability)

hermanschaaf (approved):

plugins/destination/postgresql/client/migrate.go
erezrokah: Nice one 🚀

disq (approved):

PR #7621, opened 2023-02-01
fix(aws): Update EC2 Tag structure

bbernays:


disq (approved):

hermanschaaf:
Hmm, I assume this was a regression - how did it happen? Was it during the transformer refactor?

hermanschaaf (approved):

bbernays:
> Hmm, I assume this was a regression - how did it happen? Was it during the transformer refactor?

Yeah it was during the transformer refactor... I am trying to figure out how I can verify that this was in fact the only miss... Any ideas?

hermanschaaf:
@bbernays I think we pretty much have to look for all `tags` columns that don't have a custom resolver. The best thing might be to write a test for AWS specifically that checks that every `tags` column is either using a custom resolver or maps to a `map[string]string`... might take some doing, but seems possible

candiduslynx:
Could we also add a resolver transformer that will check if the fields is a JSON field, and, in that case, verify that we're dealing with tags in form of kv pairs slice? This way it'll apply to all proper fields and we would need to do this fix once.

PR #7624, opened 2023-02-01
feat(gcp-resources): Add Security Center folders findings

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/7089.
Fixes https://github.com/cloudquery/cloudquery/issues/7007

It uses the spec folders ids, as the other alternative is to either:
1. Create a new `security_folders` table to list all folders and make this a relation of that
2. Use https://github.com/cloudquery/cloudquery/blob/f72091142a436fb13fad85d77a14c802b82c8f4f/plugins/source/gcp/resources/services/resourcemanager/folders.go#L12 and make this a relation of that table


erezrokah:
This should be ready to go (got the permissions to test it)

yevgenypats (approved):

Issue #7628, opened 2023-02-01
Azure Virtual Network Gateways and Virtual Network Gateway Connections


jsonpr:
Add CQ Support for Azure Virtual Network Gateways and Virtual Network Gateway Connections.

(https://learn.microsoft.com/en-us/rest/api/network-gateway/virtual-network-gateways)

https://learn.microsoft.com/en-us/rest/api/network-gateway/virtual-network-gateway-connections

PR #7636, opened 2023-02-02
feat(azure-resources): Add Virtual Network Gateways and Connections

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7628.

### Notes

1. Took a bit of time to figure out how to create these resources, but it's working on our playground account
2. Had to create a resource group multiplexer (and fix the code that retrieves those), as the API to list the new tables requires a resource group. Please note that these are different from other tables that we get the resource group from the parent, for example:
https://github.com/cloudquery/cloudquery/blob/8d9711567ed8eca5e814a1e8840d25923f190bc6/plugins/source/azure/resources/services/appservice/web_app_vnet_connections.go#L30, as here we need only a resource group and not a parent id for example.


plugins/source/azure/client/client.go
```diff
@@ -23,11 +23,13 @@ type Client struct {
 	SubscriptionsObjects []*armsubscription.Subscription
 	logger               zerolog.Logger
 	registeredNamespaces map[string]map[string]bool
-	resourceGroups       map[string][]*armresources.GenericResourceExpanded
+	resourceGroups       map[string][]string
```
erezrokah, line 26: We only need the group names, no reason to keep the whole object

plugins/source/azure/client/client.go
```diff
@@ -54,20 +56,25 @@ func (c *Client) discoverSubscriptions(ctx context.Context) error {
...
+	names := make([]string, len(resourceGroups))
+	for i, rg := range resourceGroups {
+		names[i] = *rg.Name
+	}
+	return names
+}
+
 func (c *Client) disocverResourceGroups(ctx context.Context) error {
-	c.resourceGroups = make(map[string][]*armresources.GenericResourceExpanded, len(c.subscriptions))
-	filter := "resourceType eq 'Microsoft.Resources/resourceGroups'"
```
erezrokah, line 59: This query wasn't working (returned 0 results). There's a dedicated client to get the resource groups so I used that one

plugins/source/azure/client/client.go
```diff
@@ -143,7 +150,10 @@ func (c *Client) Logger() *zerolog.Logger {
 }
 
 func (c *Client) ID() string {
-	return c.SubscriptionId
+	if c.ResourceGroup != "" {
```
erezrokah, line 153: Follows the Azure convention, see https://github.com/cloudquery/cloudquery/blob/8d9711567ed8eca5e814a1e8840d25923f190bc6/plugins/source/azure/client/helpers.go#L11

plugins/source/azure/client/multiplexer.go
```diff
@@ -35,6 +35,27 @@ func SubscriptionMultiplexRegisteredNamespace(table, namespace string) func(sche
 	}
 }
 
+func SubscriptionResourceGroupMultiplexRegisteredNamespace(table string, namespace string) func(schema.ClientMeta) []schema.ClientMeta {
+	return func(meta schema.ClientMeta) []schema.ClientMeta {
+		client := meta.(*Client)
+		var c = make([]schema.ClientMeta, 0)
+		for _, subId := range client.subscriptions {
+			if _, ok := client.registeredNamespaces[subId][namespace]; ok {
```
erezrokah, line 43: Kept the logic as `SubscriptionMultiplexRegisteredNamespace` to skip non registered namespaces

plugins/source/azure/resources/services/network/virtual_network_gateway_connections.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func fetchVirtualNetworkGatewayConnections(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	svc, err := armnetwork.NewVirtualNetworkGatewayConnectionsClient(cl.SubscriptionId, cl.Creds, cl.Options)
+	if err != nil {
+		return err
+	}
+
+	pager := svc.NewListPager(cl.ResourceGroup, nil)
```
erezrokah, line 30: See here that we only pass the resource group unlike other resources that pass other parent's fields

plugins/source/azure/resources/services/network/virtual_network_gateways.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func fetchVirtualNetworkGateways(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	svc, err := armnetwork.NewVirtualNetworkGatewaysClient(cl.SubscriptionId, cl.Creds, cl.Options)
+	if err != nil {
+		return err
+	}
+
+	pager := svc.NewListPager(cl.ResourceGroup, nil)
```
erezrokah, line 30: See here that we only pass the resource group unlike other resources that pass other parent's fields

yevgenypats (approved):

Issue #7637, opened 2023-02-02
feat(aws-resources): Add EBS `VolumeStatusEvent`


erezrokah:
See https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_VolumeStatusEvent.html

Looks like we need to sync https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_VolumeStatusItem.html to get those

PR #7638, opened 2023-02-02
feat(aws-resources): Add EC2 EBS Volume Statuses

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7637.
To get status events, we need `DescribeVolumeStatus`.

I tested syncing statues, but events is a bit hard as those are created by AWS, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-volume-status.html?icmpid=docs_ec2_console#monitoring-vol-events

### DIFF
--- a/plugins/source/aws/resources/services/ec2/ebs_volumes_fetch.go
+++ b/plugins/source/aws/resources/services/ec2/ebs_volumes_fetch.go
@@ -29,15 +29,20 @@ func fetchEc2EbsVolumes(ctx context.Context, meta schema.ClientMeta, _ *schema.R
 	return nil
 }
 
+func resolveVolumeARN(partition string, region string, accountID string, volumeId string) string {
+	a := arn.ARN{
+		Partition: partition,
+		Service:   "ec2",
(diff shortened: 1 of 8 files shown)


plugins/source/aws/resources/services/ec2/ebs_volumes_fetch.go
```diff
@@ -29,15 +29,20 @@ func fetchEc2EbsVolumes(ctx context.Context, meta schema.ClientMeta, _ *schema.R
 	return nil
 }
 
-func resolveEbsVolumeArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
-	cl := meta.(*client.Client)
-	volume := resource.Item.(types.Volume)
+func resolveVolumeARN(partition string, region string, accountID string, volumeId string) string {
```
erezrokah, line 32: So we can re-use this logic

hermanschaaf (approved):

PR #7651, opened 2023-02-05
feat(aws): Add New Validation for AWS Tags

bbernays:
#### Summary

This PR is blocked by new functionality added to the [Plugin SDK](https://github.com/cloudquery/plugin-sdk/pull/665) to validate all fields called `Tags`... 

Using this new functionality it identified a number of issues including
`cloudformation_stacks` (https://github.com/cloudquery/cloudquery/pull/7627)
`codebuild_projects` (https://github.com/cloudquery/cloudquery/pull/7627)
`ec2_directconnect` (https://github.com/cloudquery/cloudquery/pull/7627)
`customer_gateways` (https://github.com/cloudquery/cloudquery/pull/7627)
`redshift_clusters` (https://github.com/cloudquery/cloudquery/pull/7627)
`aws_autoscaling_groups`


plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,39 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
 		Version:      version,
 		Tables:       []string{table.Name},
 		Destinations: []string{"mock-destination"},
-	})
+	}, source.WithTestPluginAdditionalValidators(TagCheck))
+}
+
+func TagCheck(t *testing.T, tables schema.Tables, resources []*schema.Resource) {
```
disq, line 57: This doesn't need to be exported, could just be `tagCheck`?

plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,39 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
...
+						if resource.Table.Name != table.Name {
+							continue
+						}
+
+						for iResource, value := range resource.GetValues() {
+							if iResource != i {
+								continue
+							}
+							if value.Get() != nil && value.Get() != schema.Undefined {
+								_, ok := value.Get().(map[string]interface{})
```
disq, line 72:
```suggestion
								_, ok := value.Get().(map[string]any)
```

plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,39 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
...
+								if !ok {
+									t.Fatalf("unexpected value for tags column")
+								}
+							}
+						}
+					}
+
+					if column.Type != schema.TypeJSON {
+						t.Fatalf("tags column should be of type JSON")
+						fmt.Println(resources)
```
disq, line 82: Leftover print? Or use `t.Logf`? If might not work after `t.Fatalf` above, though. For good measure should include table name in the `t.Fatalf`

plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,39 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
...
 		Destinations: []string{"mock-destination"},
-	})
+	}, source.WithTestPluginAdditionalValidators(TagCheck))
+}
+
+func TagCheck(t *testing.T, tables schema.Tables, resources []*schema.Resource) {
+	for _, table := range tables {
+		t.Run(table.Name, func(t *testing.T) {
+			for i, column := range table.Columns {
+				if column.Name == "tags" {
```
hermanschaaf, line 61: Nit: change this to 

```
if column.Name != "tags" {
    continue
}
```

so that there's less nesting

plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,46 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
 		Version:      version,
 		Tables:       []string{table.Name},
 		Destinations: []string{"mock-destination"},
-	})
+	}, source.WithTestPluginAdditionalValidators(validateTagStructure))
+}
+func extractTables(tables schema.Tables) []*schema.Table {
```
disq, line 56:
```suggestion
func extractTables(tables schema.Tables) schema.Tables {
```
Maybe? Should be the same

plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,46 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
...
+				for _, resource := range resources {
+					if resource.Table.Name != table.Name {
+						continue
+					}
+
+					for iResource, value := range resource.GetValues() {
+						if iResource != i {
+							continue
+						}
+						if value.Get() != nil && value.Get() != schema.Undefined {
```
hermanschaaf, line 81: Could this happen? Isn't it a problem if the value is nil or undefined?

plugins/source/aws/resources/services/autoscaling/groups.go
```diff
@@ -43,6 +43,16 @@ func Groups() *schema.Table {
 					PrimaryKey: true,
 				},
 			},
+			{
+				Name:     "tags",
+				Type:     schema.TypeJSON,
+				Resolver: resolveCleanTags,
```
disq, line 49: why can't `client.ResolveTags` work on this? It just looks for a slice with `Key` and `Value` fields, but doesn't fail if the struct has any extras.

plugins/source/aws/client/testing.go
```diff
@@ -51,5 +51,46 @@ func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.
...
+							_, ok := value.Get().(map[string]any)
+							if !ok {
+								t.Fatalf("unexpected value for tags column")
```
hermanschaaf, line 84:
```suggestion
								t.Fatalf("unexpected type for tags column: got %v, want type map[string]any", value.Get())
```

plugins/source/aws/client/testing.go
hermanschaaf: Right, but I'm saying it seems like this test should fail if the tags are nil or undefined, not succeed?

plugins/source/aws/resources/services/autoscaling/groups.go
bbernays: That is cool! Didn't realize it could do that

plugins/source/aws/client/testing.go
bbernays: There are 100 other occurrences of us using `[]*schema.Table` rather than `schema.Tables` if that is a change we want to make we should consider how to enforce it and change it everywhere

plugins/source/aws/client/testing.go
bbernays: Fair... I have removed that conditional

plugins/source/aws/client/testing.go
hermanschaaf: I think @disq is just saying that this particular signature is inconsistent: it takes `schema.Tables` as an argument but then returns `[]*schema.Table`, which is actually the same thing - we may as well keep it consistent

disq (approved):

hermanschaaf (approved):

Issue #7684, opened 2023-02-06
feat(GCP): Compute routers + mapping info


erezrokah:
Requested on Discord https://discord.com/channels/872925471417962546/1072087046446915605/1072097218720440370

APIs:
https://cloud.google.com/compute/docs/reference/rest/v1/routers
https://cloud.google.com/compute/docs/reference/rest/v1/routers/getNatMappingInfo

PR #7691, opened 2023-02-06
feat(gcp-resources): Add Compute Routers and NAT Mappings

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7684


plugins/source/gcp/docs/tables/gcp_compute_router_nat_mapping_infos.md
```diff
@@ -0,0 +1,21 @@
...
+
+This table depends on [gcp_compute_routers](gcp_compute_routers.md).
+
+## Columns
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
```
erezrokah, line 17: Could not find a good PK to set here

plugins/source/gcp/resources/services/compute/router_nat_mapping_infos.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	it := gcpClient.GetNatMappingInfo(ctx, req, c.CallOptions...)
+	for {
+		resp, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+
+		if err != nil {
+			if strings.Contains(err.Error(), "No Nat mapping information is available in the given router.") {
```
erezrokah, line 53: I'm not sure if we should report this or not. Seems like this error is expected when there are no mappings so not sure we should fail

yevgenypats (approved):

PR #7695, opened 2023-02-06
feat(aws): Add support for Amplify Apps

bbernays:


plugins/source/aws/go.mod
```diff
@@ -113,6 +113,8 @@ require (
 	golang.org/x/sync v0.1.0
 )
 
+require github.com/aws/aws-sdk-go-v2/service/amplify v1.13.1 // indirect
```
erezrokah, line 116: Should this be a direct dependency? Maybe a `go mod tidy`?

erezrokah (approved):

plugins/source/aws/go.mod
bbernays: fixed

hermanschaaf (approved):

Issue #7604, opened 2023-02-01
bug(gcp): Security Center Findings fails to list organization findings is there are too many of those


erezrokah:
Reported on [Discord](https://discord.com/channels/872925471417962546/1069652274839830548).

Relevant code:
https://github.com/cloudquery/cloudquery/blob/c549c275077f1cdfb9df0b3f3c129cbf0b150552/plugins/source/gcp/resources/services/securitycenter/organization_findings.go#L43

⬆️ Works on our organization but fails to get all the data for the user's organization.

We might want to limit the amount of data we extract or look into making this an incremental table

PR #7696, opened 2023-02-06
feat(gcp-resources)!: Make Security Center Tables Incremental

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7604 by making security center findings incremental.
We had reports from 2 users that the amount of findings can be quite big, especially for organiazations.
From the docs:
https://pkg.go.dev/cloud.google.com/go/securitycenter@v1.18.1/apiv1/securitycenterpb#Finding./

Marked this as a breaking change due to the behavioral change

```
// The time the finding was first detected. If an existing finding is updated,
// then this is the time the update occurred.
// For example, if the finding represents an open firewall, this property
// captures the time the detector believes the firewall became open. The
// accuracy is determined by the detector. If the finding is later resolved,
// then this time reflects when the finding was resolved. This must not
// be set to a value greater than the current timestamp.
EventTime *[timestamppb](https://pkg.go.dev/google.golang.org/protobuf/types/known/timestamppb).[Timestamp](https://pkg.go.dev/google.golang.org/protobuf/types/known/timestamppb#Timestamp) `protobuf:"bytes,9,opt,name=event_time,json=eventTime,proto3" json:"event_time,omitempty"`
```

So I use that field as state.


erezrokah:
Made this a draft as I think I’ll need to ensure results are returned sorted by event time for this to work

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
```diff
@@ -2,28 +2,62 @@ package securitycenter
...
+	filter, err := c.Backend.Get(ctx, table, c.ID())
+	if err != nil {
+		c.Logger().Warn().Str("table", table).Msgf("failed to get filter %s", err.Error())
+	}
+	var req *pb.ListFindingsRequest
+	if filter == "" || err != nil {
+		req = &pb.ListFindingsRequest{
 			Parent:   parent,
-			PageSize: 1000,
+			OrderBy:  "event_time",
```
erezrokah, line 29: `OrderBy: "event_time"` is important as we use `event_time` as the cursor

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
```diff
@@ -2,28 +2,62 @@ package securitycenter
...
+			PageSize: pageSize,
+			Filter:   filter,
+		}
+	}
+	return req
+}
+
+func setBackendState(ctx context.Context, c *client.Client, table string, lastEventTime *timestamppb.Timestamp) {
+	err := c.Backend.Set(ctx, table, c.ID(), fmt.Sprintf(`event_time >= "%s"`, lastEventTime.AsTime().Format(time.RFC3339)))
```
erezrokah, line 44: I chose `>=` to have a bit of buffer. We could do `>` too but I think it's ok to have some overlap

disq (approved):

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
```diff
@@ -2,28 +2,62 @@ package securitycenter
...
+	if err != nil {
+		c.Logger().Warn().Str("table", table).Msgf("failed to set filter %s", err.Error())
```
disq, line 46: Traditionally (in other plugins) we fail if any of the backend methods fail.

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
```diff
@@ -2,28 +2,62 @@ package securitycenter
...
+const pageSize = 1000
+
+func getRequest(ctx context.Context, c *client.Client, table string, parent string) *pb.ListFindingsRequest {
+	filter, err := c.Backend.Get(ctx, table, c.ID())
+	if err != nil {
+		c.Logger().Warn().Str("table", table).Msgf("failed to get filter %s", err.Error())
+	}
+	var req *pb.ListFindingsRequest
+	if filter == "" || err != nil {
+		req = &pb.ListFindingsRequest{
```
disq, line 27: I would prepare a single `req` and then set `filter` if it's available. Also, since filter is a string and default zero value is already "" we don't need this if and "selective setting" at all.

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
erezrokah: I was wondering about that too. Should we fail? Not sure

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
erezrokah: Good point I'll update it

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
disq: Storing the filter string itself (and not the int/timestamp) is a unique approach and might make it hard to "port" old states later on (if the filter format changes in a newer `pb`)

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/7696/commits/88810fbf4e38c0df498fc02a3476410c904b0cdb

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
disq: If it errored and ran without filter it would suddenly take too long and fetch it all (which might have other implications API-limits wise)... Leaving the decision to the user (either remove the state file or remove backend config) is better I think.

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
erezrokah: It's actually intentional if someone wants to provide a different filter/state manually (was also a request).
I think if the filter format changes we can drop old states and do a full refresh

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
erezrokah: Got it, makes sense. I'll return an error instead

plugins/source/gcp/resources/services/securitycenter/findings_fetch.go
```diff
@@ -2,28 +2,59 @@ package securitycenter
...
+	filter, err := c.Backend.Get(ctx, table, c.ID())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get filter state %w for table %q", err, table)
+	}
+	req := &pb.ListFindingsRequest{
+		Parent:   parent,
+		OrderBy:  "event_time",
+		PageSize: pageSize,
+	}
+	if filter != "" && err == nil {
```
disq, line 30:
```suggestion
	if filter != "" {
```

disq (approved):

yevgenypats (approved):

PR #7699, opened 2023-02-06
feat(aws): Add IAM Signing Certificate

bbernays:


disq:
Should the PK include account_id or user_id? Because seems like people upload their own certificates using [UploadSigningCertificate](https://docs.aws.amazon.com/IAM/latest/APIReference/API_UploadSigningCertificate.html) so if the same cert can be used in multiple users/accounts could it have the same id?

erezrokah (approved):

bbernays:
> Should the PK include account_id or user_id? Because seems like people upload their own certificates using [UploadSigningCertificate](https://docs.aws.amazon.com/IAM/latest/APIReference/API_UploadSigningCertificate.html) so if the same cert can be used in multiple users/accounts could it have the same id?

I will add account as PK because I cannot find anything that says `CertificateId` is globally unique

disq (approved):

PR #7718, opened 2023-02-07
fix(azure): Fix Azure CIS Policies in Section 2

aruneko:
#### Summary

I found bugs in Azure CIS Benchmark 1.3.0 queries about MS Defender rules in Section 2.

Current queries check the `pricing_tier` (snake case) field in the `properties` column in the `azure_security_pricings` table. But the actual data format is below.

```json
{"subPlan": "PerTransaction", "pricingTier": "Standard", "freeTrialRemainingTime": "PT0S"}
```

That's why, those queries always return `fail` since the desired field is `pricingTier` (camel case).


erezrokah (approved):
Great find @aruneko, thanks for the PR 🎸

PR #7756, opened 2023-02-07
feat: Better handling of Ctrl+C, log failure reason when sync ends abruptly

hermanschaaf:
Before, when pressing ctrl+c, these messages would show in the CLI:

```
failed to terminate destination client:  destination plugin process exited with status signal: killed
failed to terminate source client:  source plugin process exited with status signal: killed
```

This PR fixes it so that these messages no longer appear. This is the main user-facing change. 

Other changes include:
 - sending an analytics message when the sync ends for any reason, rather than only on successful syncs
 - include an "exit reason" in the analytics message, so that we can better understand when and why syncs fail
 - introduction of an `CQ_ANALYTICS_HOST` environment variable that can override the default host. This is mainly useful for testing, so no code changes are necessary to send analytics to a test server. For normal operation, users should still use `--telemetry-level` or `CQ_TELEMETRY_LEVEL` to control the telemetry
 - send analytics not only for the GitHub registry, but also for path and grpc registries. This can help us better understand how plugins are being deployed. We strip off any potentially identifying information and send only the final part of the path (i.e. the name of the binary)

This change will be improved by https://github.com/cloudquery/plugin-sdk/pull/664 but doesn't depend on it to work.


cli/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/cloudquery/cloudquery/cli
 go 1.19
 
 require (
-	github.com/cloudquery/plugin-sdk v1.33.1
+	github.com/cloudquery/plugin-sdk v1.34.0
```
erezrokah, line 6: Our CLI is already using this version:
https://github.com/cloudquery/cloudquery/blob/3a25ce24c2b88a3c244c6312949b12e252644651/cli/go.mod#L6

Not sure why we see this diff

erezrokah (approved):

cli/cmd/sync_v1.go
```diff
@@ -42,18 +61,44 @@ func syncConnectionV1(ctx context.Context, cqDir string, sourceSpec specs.Source
 	log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Time("sync_time", syncTime).Msg("Start sync")
 	defer log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Time("sync_time", syncTime).Msg("End sync")
 
-	destClients, err := newDestinationClientsV0(ctx, sourceSpec, destinationsSpecs, cqDir)
+	destCtx, destCancel := context.WithCancel(context.Background())
```
disq, line 64: why not do a `defer destCancel()` here? Same for `sourceCancel`

hermanschaaf: Yeah good point, I think we can

hermanschaaf: Updated

cli/cmd/sync_v1.go
```diff
@@ -7,30 +7,51 @@ import (
...
-	//nolint:revive
 	defer func() {
+		if canceled {
+			return
+		}
 		if err := sourceClient.Terminate(); err != nil {
 			log.Error().Err(err).Msg("Failed to terminate source client")
 			fmt.Println("failed to terminate source client: ", err)
 		}
+		sourceCancel()
```
disq, line 54: No need for this now that we're deferring it before this defer

cli/cmd/sync_v1.go
```diff
@@ -42,18 +63,46 @@ func syncConnectionV1(ctx context.Context, cqDir string, sourceSpec specs.Source
 	log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Time("sync_time", syncTime).Msg("Start sync")
 	defer log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Time("sync_time", syncTime).Msg("End sync")
 
-	destClients, err := newDestinationClientsV0(ctx, sourceSpec, destinationsSpecs, cqDir)
+	destCtx, destCancel := context.WithCancel(context.Background())
+	defer destCancel()
+
+	destClients, err := newDestinationClientsV0(destCtx, sourceSpec, destinationsSpecs, cqDir)
 	if err != nil {
+		destCancel()
```
disq, line 71:
```suggestion
		
```

cli/cmd/sync_v1.go
```diff
@@ -42,18 +63,46 @@ func syncConnectionV1(ctx context.Context, cqDir string, sourceSpec specs.Source
...
+		exitReason = "failed to get destination plugin client"
 		return err
 	}
-	defer destClients.Close()
+	defer func() {
+		if canceled {
+			return
+		}
+		destClients.Close()
+		destCancel()
```
disq, line 80:
```suggestion
```

cli/cmd/sync_v1.go
disq, line 54:
```suggestion
```
No need for this now that we're deferring it before this defer

disq (approved):

cli/cmd/sync_v1.go
```diff
@@ -7,26 +7,46 @@ import (
...
+	}()
 	opts := []source.ClientOption{
 		source.WithLogger(log.Logger),
 		source.WithDirectory(cqDir),
 	}
 	if disableSentry {
 		opts = append(opts, source.WithNoSentry())
 	}
-	sourceClient, err := source.NewClient(ctx, sourceSpec.Registry, sourceSpec.Path, sourceSpec.Version, opts...)
+	sourceCtx, sourceCancel := context.WithCancel(context.Background())
```
yevgenypats, line 37: why background here?

cli/cmd/sync_v1.go
```diff
@@ -42,18 +62,44 @@ func syncConnectionV1(ctx context.Context, cqDir string, sourceSpec specs.Source
 	log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Time("sync_time", syncTime).Msg("Start sync")
 	defer log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Time("sync_time", syncTime).Msg("End sync")
 
-	destClients, err := newDestinationClientsV0(ctx, sourceSpec, destinationsSpecs, cqDir)
+	destCtx, destCancel := context.WithCancel(context.Background())
```
yevgenypats, line 65: Same here? why `context.Background` ?

yevgenypats (changes requested):
Few questions on the context. still need to give it another review as it's tricky code section.

cli/cmd/sync_v1.go
hermanschaaf: The idea here is that the CLI should manage the context for the plugins. If the CLI receives a signal and its context gets cancelled, we don't want that to bubble down to the plugins, because then we don't have time to fetch metrics first and send a shutdown signal of our own. Instead, the CLI should listen for when its context gets cancelled, do some clean-up, and then cancel the contexts of the plugins

hermanschaaf: To add to this: now that https://github.com/cloudquery/plugin-sdk/pull/664 has been merged, the plugins should run in a separate process group, so they won't receive an interrupt signal from the OS when ctrl+c is pressed. However, we still use a [command with context](https://github.com/cloudquery/plugin-sdk/blob/main/clients/source/v1/source.go#L129) which [sends a kill signal](https://pkg.go.dev/os/exec#CommandContext) as soon as the context is cancelled. To avoid this, we will either need to stop using CommandContext, or send a different context like the change in this PR is doing.

yevgenypats (approved):

erezrokah:
Is there anything blocking us from merging and releasing this?

hermanschaaf:
@erezrokah It's been a few weeks, so I'd like to just do a quick round of manual testing again to ensure that it all still works as expected. Some of these scenarios, like ctrl+c behavior, are not currently being tested in an automated way (though maybe that would be a good test to add)

hermanschaaf:
I did some more testing and it still looks good, let's merge it

Issue #7705, opened 2023-02-06
AWS EMR Security Configuration


jsonpr:
https://docs.aws.amazon.com/cli/latest/reference/emr/describe-security-configuration.html
https://docs.aws.amazon.com/cli/latest/reference/emr/list-security-configurations.html

PR #7812, opened 2023-02-08
feat(aws-resources): Add EMR Security Configuration

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7705

Couldn't find an ARN structure for this. AWS console doesn't let you create multiple configs with the same name


yevgenypats (approved):

hermanschaaf (approved):
👍

PR #7815, opened 2023-02-08
chore(mssql): Extend `unescape` to include quoted strings as well

candiduslynx:
Found during clickhouse plugin testing


plugins/destination/mssql/client/unicode.go
```diff
@@ -21,5 +22,11 @@ func unescape(str string) string {
 		str = str[1:]
 	}
 
+	// check for unquote, too
+	if strings.Contains(out, `\"`) || strings.Contains(out, `\n`) {
```
disq, line 26: I would negate this expression and do an early `return out` if it doesn't contain these. Also you can probably use `strings.ContainsAny`

candiduslynx: cannot use `ContainsAny` as we check for 2 characters (escaped sequence).
Moved to early exit

disq (approved):

PR #7819, opened 2023-02-08
feat(pg): Faster migrations

yevgenypats:
Blocked by https://github.com/cloudquery/plugin-sdk/pull/668

This streamlines migrations, introduce force option and in general should speed up migrations substantially as we only do now one query instead of query per table. 

Blocked by another take in SDK - https://github.com/cloudquery/plugin-sdk/pull/688


plugins/destination/postgresql/client/client.go
```diff
@@ -132,6 +135,19 @@ func (c *Client) currentDatabase(ctx context.Context) (string, error) {
 	return db, nil
 }
 
+func (c *Client) currentSchema(ctx context.Context) (string, error) {
+	var resetVal string
+	err := c.conn.QueryRow(ctx, "select reset_val from pg_settings where name='search_path'").Scan(&resetVal)
+	if err != nil {
+		return "", err
+	}
+	schemaTokens := strings.Split(resetVal, ",")
```
disq, line 144: This won't always work. Better to just do `select current_schema();`
https://stackoverflow.com/a/69841555/7506

yevgenypats: Good catch.

plugins/destination/postgresql/client/write.go
```diff
@@ -54,10 +54,27 @@ func pgErrToStr(err *pgconn.PgError) string {
...
 
+func (c *Client) populateConstraintNames(ctx context.Context, tables schema.Tables) error {
+	pgTables, err := c.listPgTables(ctx, tables)
+	if err != nil {
+		return err
+	}
+	for _, table := range tables.FlattenTables() {
+		pgTable := pgTables.Get(table.Name)
+		if pgTable == nil {
+			return fmt.Errorf("table %s not found in postgres. make sure to run migrate", table.Name)
```
candiduslynx, line 65:
```suggestion
			return fmt.Errorf("table %s not found in postgres (need to run migrate)", table.Name)
```
[nit]

plugins/destination/postgresql/client/write.go
```diff
@@ -66,7 +83,11 @@ func (c *Client) Write(ctx context.Context, tables schema.Tables, res <-chan *de
 		if c.spec.WriteMode == specs.WriteModeAppend {
 			sql = c.insert(table)
 		} else {
-			sql = c.upsert(table)
+			if len(table.PrimaryKeys()) > 0 {
```
candiduslynx, line 86: IMO it's always true (no PK -> use `_cq_id`, see [here](https://github.com/cloudquery/plugin-sdk/blob/main/plugins/source/plugin.go#L75))

plugins/destination/postgresql/client/types.go
```diff
@@ -94,3 +94,86 @@ func (*Client) SchemaTypeToCockroach(t schema.ValueType) string {
...
+	case "bigint[]":
+		return schema.TypeIntArray
+	default:
+		panic("unknown type " + t)
+	}
+}
+
+func (*Client) CockroachToSchemaType(t string) schema.ValueType {
+	switch t {
+	case "boolean":
```
candiduslynx, line 150: might be a good idea to have unexported maps and do a lookup instead

plugins/destination/postgresql/client/migrate.go
```diff
@@ -246,11 +261,6 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		sb.WriteString("_cqpk PRIMARY KEY (")
 		sb.WriteString(strings.Join(primaryKeys, ","))
 		sb.WriteString(")")
-	} else {
```
candiduslynx, line 249: This was part of the append mode workflow, as we use `_cq_id` instead of the real PK here.
Is it removed now?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,43 +3,125 @@ package client
...
 		if len(table.Columns) == 0 {
 			c.logger.Info().Str("table", table.Name).Msg("Table with no columns, skipping")
 			continue
 		}
-		tableExist, err := c.isTableExistSQL(ctx, table.Name)
-		if err != nil {
-			return fmt.Errorf("failed to check if table %s exists: %w", table.Name, err)
+
+		// In postgres if column is primary key, it can't be null
+		if c.enabledPks() {
```
candiduslynx, line 113: what about append mode (`_cq_id` must be marked `unique not null`)?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,43 +3,125 @@ package client
...
-AND    i.indisprimary;
+	// this returns the following table in sorted manner:
+	// +----------------+-------------+-------------+------------+---------------+-----------+---------------------+
+	// | ordinal_position | table_name | column_name | data_type | is_primary_key| not_null  | pk_constraint_name  |
+	// +----------------+-------------+-------------+------------+---------------+-----------+---------------------+
+	// |              1 | users       | id          | bigint     | YES           | true 		 | cq_users_pk 	  	   |
+	// |              2 | users       | name        | text       | NO            | false 	   | 					           |
+	// |              3 | users       | email       | text       | NO            | true 		 | cq_users_pk         |
+	// |              1 | posts       | id          | bigint     | YES           | true 		 | cq_posts_pk			   |
+	// |              2 | posts       | title       | text       | NO            | false 	   | 					           |
```
candiduslynx, line 22:
```suggestion
	// this returns the following table in sorted manner:
	// +----------------+-------------+-------------+------------+---------------+-----------+---------------------+
	// | ordinal_position | table_name | column_name | data_type | is_primary_key| not_null  | pk_constraint_name  |
	// +----------------+-------------+-------------+------------+---------------+-----------+---------------------+
	// |              1 | users       | id          | bigint     | YES           | true      | cq_users_pk 	  	   |
	// |              2 | users       | name        | text       | NO            | false     |                     |
	// |              3 | users       | email       | text       | NO            | true      | cq_users_pk         |
	// |              1 | posts       | id          | bigint     | YES           | true      | cq_posts_pk         |
	// |              2 | posts       | title       | text       | NO            | false     |                     |
```

candiduslynx (changes requested):
I sense that there are issues with the `append` mode now (`_cq_id` was used as PK there).

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +130,98 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
 
-	return nil
-}
-
-func (c *Client) setNotNullOnPks(ctx context.Context, table *schema.Table) error {
-	for _, col := range table.PrimaryKeys() {
-		sql := "alter table " + pgx.Identifier{table.Name}.Sanitize() + " alter column " + pgx.Identifier{col}.Sanitize() + " set not null"
-		if _, err := c.conn.Exec(ctx, sql); err != nil {
-			return fmt.Errorf("failed to set not null on column %s on table %s: %w", col, table.Name, err)
+	if c.enabledPks() && !table.IsPrimaryKeyEqual(pgTable) {
```
hermanschaaf, line 215: I think we may want to move this check to be above the `GetAddedColumns` check, so that no changes are made to the table if `migrate-forced` is false and a forced migration is necessary. Otherwise the table can be left in a half-migrated state

plugins/destination/postgresql/client/types.go
```diff
@@ -94,3 +94,86 @@ func (*Client) SchemaTypeToCockroach(t schema.ValueType) string {
...
+	default:
+		return c.Pg10ToSchemaType(t)
+	}
+}
+
+func (*Client) Pg10ToSchemaType(t string) schema.ValueType {
+	switch t {
+	case "boolean":
+		return schema.TypeBool
+	case "bigint", "integer":
```
hermanschaaf, line 111: Interesting, I guess there is some information lost in the conversion here because our types don't map 1:1 to postgres 🤔 So if you have a source table and write it back to postgres, the schemas won't match exactly (integers will become bigints). In this case it's fine, but hopefully we don't run into irreconcilable cases in other databases...

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,43 +3,125 @@ package client
...
+ AND pg_constraint.conkey[1] = pg_attribute.attnum
+ AND contype = 'p'
+ INNER JOIN
+ information_schema.columns ON columns.table_name = pg_class.relname AND columns.column_name = pg_attribute.attname AND columns.table_schema = pg_catalog.pg_namespace.nspname
+ WHERE
+ pg_attribute.attnum > 0
+ AND NOT pg_attribute.attisdropped
+ AND pg_catalog.pg_namespace.nspname = '%s'
+ ORDER BY
+ table_name ASC , ordinal_position ASC;
```
hermanschaaf, line 55: I think this query is not returning the expected results for tables with composite primary keys.

Here is a table I have locally (`\d aws_s3_bucket_grants`):

```
 aws_s3_bucket_grants
+-----------------+-----------------------------+-----------+
| Column          | Type                        | Modifiers |
|-----------------+-----------------------------+-----------|
| _cq_source_name | text                        |           |
| _cq_sync_time   | timestamp without time zone |           |
| _cq_id          | uuid                        |  not null |
| _cq_parent_id   | uuid                        |           |
| account_id      | text                        |           |
| bucket_arn      | text                        |  not null |
| grantee_type    | text                        |  not null |
| grantee_id      | text                        |  not null |
| grantee         | jsonb                       |           |
| permission      | text                        |  not null |
+-----------------+-----------------------------+-----------+
Indexes:
    "aws_s3_bucket_grants_cqpk" PRIMARY KEY, btree (bucket_arn, grantee_type, grantee_id, permission)
    "aws_s3_bucket_grants__cq_id_key" UNIQUE CONSTRAINT, btree (_cq_id)

Time: 0.040s
```

And here are the results for this table from this query:

```
+------------------+----------------------+-----------------+-----------------------------+----------------+----------+-----------------------------+
| ordinal_position | table_name           | column_name     | data_type                   | is_primary_key | not_null | primary_key_constraint_name |
|------------------+----------------------+-----------------+-----------------------------+----------------+----------+-----------------------------|
| 1                | aws_s3_bucket_grants | _cq_source_name | text                        | False          | False    |                             |
| 2                | aws_s3_bucket_grants | _cq_sync_time   | timestamp without time zone | False          | False    |                             |
| 3                | aws_s3_bucket_grants | _cq_id          | uuid                        | False          | True     |                             |
| 4                | aws_s3_bucket_grants | _cq_parent_id   | uuid                        | False          | False    |                             |
| 5                | aws_s3_bucket_grants | account_id      | text                        | False          | False    |                             |
| 6                | aws_s3_bucket_grants | bucket_arn      | text                        | True           | True     | aws_s3_bucket_grants_cqpk   |
| 7                | aws_s3_bucket_grants | grantee_type    | text                        | False          | True     |                             |
| 8                | aws_s3_bucket_grants | grantee_id      | text                        | False          | True     |                             |
| 9                | aws_s3_bucket_grants | grantee         | jsonb                       | False          | False    |                             |
| 10               | aws_s3_bucket_grants | permission      | text                        | False          | True     |                             |
+------------------+----------------------+-----------------+-----------------------------+----------------+----------+-----------------------------+
```

Notice that the `is_primary_key` column is true only for `bucket_arn`, even though it should be true for `grantee_type`, `grantee_id` and `permission` as well

hermanschaaf (changes requested):
It seems like the query to select all tables is not accounting for tables with composite PKs in the way we expect

plugins/destination/postgresql/client/types.go
yevgenypats: Good catch. I guess we need to transform the incoming tables to pgtables so we can compare them?

yevgenypats: Yeah I think in this case it's fine until we add BigInt type to our type system but if the information loss is the otherway around then we will need to transform first the Table -> DbTable and only then compare what we got from database table.

plugins/destination/postgresql/client/migrate.go
yevgenypats: why it's half baked? but I can't move it up because potentially the primary keys are on the columns that were just added.

yevgenypats: ok yeah you are right. I now moved that check up to fail early.

plugins/destination/postgresql/client/migrate.go
yevgenypats: good catch. fixed.

yevgenypats:
Finally! Both Postgres and Cockroach tests Passing with force/non-force migrations!

plugins/destination/postgresql/client/migrate.go
```diff
@@ -2,44 +2,164 @@ package client
...
+				NotNull:    notNull,
+			},
+			Type: c.PgToSchemaType(columnType),
+		})
+	}
+	return tables, nil
+}
+
+func (*Client) normalizeTableCockroach(table *schema.Table) *schema.Table {
+	for i := range table.Columns {
```
candiduslynx, line 102: Here we're modifying the original table spec instead of creating a copy.
This might give use some differences in the resulting data:
1. originally, column type was `cidr`
2. we've updated it to be `inet`
3. migrated (maybe even no-op)
4. writing data and marshaling it as `inet` instead of `cidr`

plugins/destination/postgresql/client/migrate.go
candiduslynx: this still stands

plugins/destination/postgresql/client/write.go
candiduslynx: now we add the `unique not null` to `_cq_id`. Do we want it to be just that (constraint on the column, no PK?)

candiduslynx (changes requested):
1. We have somewhat different data for `cocroachdb` when working with `cidr` columns (otherwise, we may as well make `cidr` be just alias for `inet` instead of its own type in our type system.
1. Need clarification on `_cq_id` column (we do add `unique not nul` but don't add it as PK?)

plugins/destination/postgresql/client/migrate.go
yevgenypats: cockroch doesn't support cidr hence the normalization so we can use built-in methods to compare tables.

plugins/destination/postgresql/README.md
```diff
@@ -1,6 +1,6 @@
 # CloudQuery PostgreSQL Destination Plugin
 
-This destination plugin lets you sync data from a CloudQuery source to a PostgreSQL compatible database.
+This destination plugin lets you sync data from a CloudQuery source to a PostgreSQL compatible database (CockroachDB).
```
erezrokah, line 3:
```suggestion
This destination plugin lets you sync data from a CloudQuery source to a PostgreSQL or CockroachDB database.
```

We haven't tested with others. Even CockroachDB differs a bit from PostgreSQL I think

plugins/destination/postgresql/client/migrate.go
```diff
@@ -2,44 +2,164 @@ package client
...
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var ordinalPosition int
+		var tableName, columnName, columnType, pkName string
+		var isPrimaryKey, notNull bool
+		if err := rows.Scan(&ordinalPosition, &tableName, &columnName, &columnType, &isPrimaryKey, &notNull, &pkName); err != nil {
+			return nil, err
+		}
+		// This is a small optimization to hold only the tables that are in the plugin
```
erezrokah, line 75:
```suggestion
		// We don't want to migrate tables that are not a part of the spec, or even non CloudQuery tables
```

erezrokah (changes requested):
Started testing the PR and noticed it fails to migrate with the following use case:
1. Run latest GCP with latest Postgres to create tables
2. Run latest GCP with Postgres from this PR
3. See migration failure:
```
postgres table gcp_aiplatform_job_locations has different types for columns [_cq_id:TypeUUID:NotNull] but schema wants [_cq_id:TypeUUID] , use --force to drop the column
```

Previously the Postgres destination set `_cq_id` as `UNIQUE NOT NULL`:
https://github.com/cloudquery/cloudquery/blob/aa7437154b5ffd89e2ae3c2aa49bda3ded6f99e9/plugins/destination/postgresql/client/migrate.go#L231

I think it's missing from this code, and thus fails the migration due to the diff in the `_cq_id` column.

> Maybe this should be done here https://github.com/cloudquery/plugin-sdk/blob/f3e2b1d982268ce9fa3c23a5cad5b853119c49e6/schema/meta.go#L17

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +168,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
+		return fmt.Errorf("failed to commit transaction to recreate primary keys: %w", err)
+	}
+	return nil
+}
 
-		tx, err := c.conn.Begin(ctx)
-		if err != nil {
-			return fmt.Errorf("failed to begin transaction to recreate primary keys: %w", err)
+func (c *Client) autoMigrateTable(ctx context.Context, pgTable *schema.Table, table *schema.Table) error {
+	changedColumns, changedPgColumns := table.GetChangedColumns(pgTable)
```
erezrokah, line 327: OK so might not be blocking but I think we should print all errors before suggesting `--force`.
Otherwise we might drop columns without explicitly warning the user about it when multiple columns require dropping

erezrokah (changes requested):
Added another comment

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +168,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
+	if c.spec.MigrateMode != specs.MigrateModeForced {
+		if changedColumns != nil {
+			return fmt.Errorf("postgres table %s has different types for columns %v but schema wants %v , use --force to drop the column", table.Name, changedPgColumns, changedColumns)
```
erezrokah, line 330:
```suggestion
			return fmt.Errorf("postgres table %s has different types for columns %v but schema wants %v , add `migrate_mode: forced` to the destination spec to drop the column", table.Name, changedPgColumns, changedColumns)
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +168,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
 		}
-		if _, err := tx.Exec(ctx, dropConstraintSQL); err != nil {
-			if err := tx.Rollback(ctx); err != nil {
-				c.logger.Error().Err(err).Msg("failed to rollback transaction")
-			}
-			return fmt.Errorf("failed to drop primary key constraint on table %s: %w", table.Name, err)
+		if c.enabledPks() && !table.IsPrimaryKeyEqual(pgTable) {
+			return fmt.Errorf("postgres table %s has different primary keys %v but schema wants %v , use --force to drop the column", table.Name, pgTable.PrimaryKeys(), table.PrimaryKeys())
```
erezrokah, line 333:
```suggestion
			return fmt.Errorf("postgres table %s has different primary keys %v but schema wants %v , add `migrate_mode: forced` to the destination spec to drop the column", table.Name, pgTable.PrimaryKeys(), table.PrimaryKeys())
```

erezrokah (changes requested):
A few more comments I missed

plugins/destination/postgresql/README.md
yevgenypats: We have tested on RDS for example and on Google Cloud SQL, so it is PostgreSQL compatible (maybe we need to adds RDS as well to the parentheses )

erezrokah: Nice, so maybe do `e.g. CockroachDB, RDS, Google Cloud SQL` like you said

plugins/destination/postgresql/client/migrate.go
yevgenypats: mm, I think let's do it in a follow-up as it can also be a pretty long error message which can be also confusing.

yevgenypats:
> Started testing the PR and noticed it fails to migrate with the following use case:
> 
> 1. Run latest GCP with latest Postgres to create tables
> 2. Run latest GCP with Postgres from this PR
> 3. See migration failure:
> 
> ```
> postgres table gcp_aiplatform_job_locations has different types for columns [_cq_id:TypeUUID:NotNull] but schema wants [_cq_id:TypeUUID] , use --force to drop the column
> ```
> 
> Previously the Postgres destination set `_cq_id` as `UNIQUE NOT NULL`:
> 
> https://github.com/cloudquery/cloudquery/blob/aa7437154b5ffd89e2ae3c2aa49bda3ded6f99e9/plugins/destination/postgresql/client/migrate.go#L231
> 
> I think it's missing from this code, and thus fails the migration due to the diff in the `_cq_id` column.
> 
> > Maybe this should be done here https://github.com/cloudquery/plugin-sdk/blob/f3e2b1d982268ce9fa3c23a5cad5b853119c49e6/schema/meta.go#L17 by adding column create options

Thanks. Good catch. Hopefully this should work now as it is fixed on the SDK level.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -2,44 +2,164 @@ package client
...
+				if table.Columns[i].CreationOptions.PrimaryKey {
+					table.Columns[i].CreationOptions.NotNull = true
+				}
+			}
+			// this is for backward compatibility as I believe this is as of right now defined on the source
+			if len(table.PrimaryKeys()) == 0 {
+				cqIdColumn := table.Columns.Get(schema.CqIDColumn.Name)
+				if cqIdColumn != nil {
+					cqIdColumn.CreationOptions.PrimaryKey = true
+					cqIdColumn.CreationOptions.NotNull = true
```
candiduslynx, line 154: can be dropped now

plugins/destination/postgresql/client/migrate.go
candiduslynx: resolved in https://github.com/cloudquery/cloudquery/pull/7819#issuecomment-1428288001

plugins/destination/postgresql/client/migrate.go
candiduslynx: resolved in https://github.com/cloudquery/cloudquery/pull/7819#issuecomment-1428288001

candiduslynx (approved):

disq (approved):

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +167,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
+	return nil
+}
 
-		tx, err := c.conn.Begin(ctx)
-		if err != nil {
-			return fmt.Errorf("failed to begin transaction to recreate primary keys: %w", err)
+func (c *Client) autoMigrateTable(ctx context.Context, pgTable *schema.Table, table *schema.Table) error {
+	changedColumns, changedPgColumns := table.GetChangedColumns(pgTable)
+	if c.spec.MigrateMode != specs.MigrateModeForced {
+		if changedColumns != nil {
```
erezrokah, line 328: Non blocking but I think we can change the order of the errors reported here (first report PK error), since if we add a PK to an existing column users will see:
```
postgres table aws_apigateway_domain_name_base_path_mappings has different types 
for columns [account_id:TypeString, arn:TypeString] but schema wants 
[account_id:TypeString:PK:NotNull, arn:TypeString:PK:NotNull] , 
add `migrate_mode: forced` to the destination spec to drop the column
```

So a PK addition to an exsiting column is reported as a column change (due to the `NotNull` addition).
If we change the order of how we report errors I think it will be more clear what what the cause of the failure, WDYT?

erezrokah (changes requested):
I'm getting an error to force migrate from AWS `v12.2.1` to `v13.3.0`.
Looking into it hence requesting changes

```
failed to add primary key constraint on table aws_apigateway_domain_name_base_path_mappings: ERROR: syntax error at or near "primary" (SQLSTATE 42601)
```

yevgenypats:
> I'm getting an error to force migrate from AWS `v12.2.1` to `v13.3.0`. Looking into it hence requesting changes
> 
> ```
> failed to add primary key constraint on table aws_apigateway_domain_name_base_path_mappings: ERROR: syntax error at or near "primary" (SQLSTATE 42601)
> ```

Great catch! I fixed now. We will slowly need to add more tests to the migration logic on the SDK side but I think we are almost there.

plugins/destination/postgresql/client/migrate.go
yevgenypats: I think it should be in this order because in postgres you first need to make sure there is `notnull` on the column before you add a pk and it can't be in the same transaction.

erezrokah: Ah, I think I didn't explain correctly. By different order I mean:
```go
		if c.enabledPks() && !table.IsPrimaryKeyEqual(pgTable) {
			return fmt.Errorf("postgres table %s has different primary keys %v but schema wants %v , add `migrate_mode: forced` to the destination spec to drop the column", table.Name, pgTable.PrimaryKeys(), table.PrimaryKeys())
		}
		if changedColumns != nil {
			return fmt.Errorf("postgres table %s has different types for columns %v but schema wants %v , add `migrate_mode: forced` to the destination spec to drop the column", table.Name, changedPgColumns, changedColumns)
		}
```

Instead of the current code, so the more specific error is printed first

erezrokah (changes requested):
PK migration seems to be fix, but now getting:
```
aws_apigateway_rest_api_documentation_parts.account_id contains null values, please drop the table manually via 'drop table aws_apigateway_rest_api_documentation_parts'
```

Shouldn't we force drop the table too?

yevgenypats:
> PK migration seems to be fix, but now getting:
> 
> ```
> aws_apigateway_rest_api_documentation_parts.account_id contains null values, please drop the table manually via 'drop table aws_apigateway_rest_api_documentation_parts'
> ```
> 
> Shouldn't we force drop the table too?

I don't think we should drop a table at least not for now as it's a more aggressive operation, maybe we can add this in the future if users are asking for that, but for now I would avoid this.

plugins/destination/postgresql/client/migrate.go
yevgenypats: I see. reordered.

erezrokah:
> I don't think we should drop a table at least not for now as it's a more aggressive operation, maybe we can add this in the future if users are asking for that, but for now I would avoid this.

So I think something is up with that logic:
image
```
column aws_iam_user_groups.user_id contains null values, please drop the table manually via 'drop table aws_iam_user_groups'
```

Taking a look

hermanschaaf:
> I don't think we should drop a table at least not for now as it's a more aggressive operation, maybe we can add this in the future if users are asking for that, but for now I would avoid this.

@yevgenypats I think we're going to run into the situation that requires a force drop of the table quite often (because of the frequent PK updates right now), and by using "force migrate" users opt into destructive behavior with the expectation that migrations will always succeed. Dropping a table seems fine from that perspective - they'll be running a sync to recreate the table with new data immediately after.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +167,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
-				c.logger.Info().Str("table", table.Name).Str("column", pk).Msg("Column exists with primary key but is not in the schema")
-				stalePks = append(stalePks, pk)
-			}
+func (c *Client) alterColumnCockroachDB(ctx context.Context, tableName string, column schema.Column) error {
+	columnName := pgx.Identifier{column.Name}.Sanitize()
+	columnType := c.SchemaTypeToPg(column.Type)
+	// try alter column in place first
+	sql := "alter table " + tableName + " alter column " + columnName + " type " + columnType
+	if column.CreationOptions.NotNull {
+		sql += " set not null"
```
erezrokah, line 199:
```suggestion
		sql += ", ALTER " + columnName + " set not null"
```

Otherwise this always fail. Also Should we return early when it's successful so we don't need to drop/add it later one?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +167,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
-	tableName := pgx.Identifier{table.Name}.Sanitize()
-	if pgColumns, err = c.getPgTableColumns(ctx, table.Name); err != nil {
-		return fmt.Errorf("failed to get table %s columns types: %w", table.Name, err)
+func (c *Client) alterColumnPg(ctx context.Context, tableName string, column schema.Column) error {
+	columnName := pgx.Identifier{column.Name}.Sanitize()
+	columnType := c.SchemaTypeToPg(column.Type)
+	// try alter column in place first
+	sql := "alter table " + tableName + " alter column " + columnName + " type " + columnType
+	if column.CreationOptions.NotNull {
+		sql += " set not null"
```
erezrokah, line 242:
```suggestion
		sql += ", ALTER " + columnName + " set not null"
```

Otherwise this always fail. Also Should we return early when it's successful so we don't need to drop/add it later one?

erezrokah (changes requested):
Added two more comments to fix the alter column statment

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +167,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
-	reCreatePrimaryKeys := false
-
-	for _, col := range table.Columns {
-		columnName := pgx.Identifier{col.Name}.Sanitize()
-		columnType := c.SchemaTypeToPg(col.Type)
-		if columnType == "" {
-			c.logger.Warn().Str("table", table.Name).Str("column", col.Name).Msg("Column type not supported, skipping")
-			continue
+	defer func() {
+		if err := tx.Rollback(ctx); err != nil {
```
erezrokah, line 252: I think Rollback will always fail if the Commit succeeds as it will try to rollback with a closed connection

plugins/destination/postgresql/client/migrate.go
```diff
@@ -48,165 +167,193 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
+	}
+	if _, err := tx.Exec(ctx, sql); err != nil {
+		var pgErr *pgconn.PgError
+		if !errors.As(err, &pgErr) {
+			return fmt.Errorf("failed to add column %s.%s: %w", tableName, column.Name, err)
+		}
+		// this means the column contains null value and we cannot recreate it without knowing what is the default
+		// This is usually the case when the column is a primary key
+		// the user will have to drop and recreate the table
+		if pgErr.Code == "23502" {
```
erezrokah, line 274: I don't understand this error report. If we drop and add a not null column on an existing table won't we always reach this code? I mean by dropiing the column we remove any existing values, so even if the column had all not null values we're basically removing them at the beginging of the transaction

erezrokah (changes requested):
Added anoter comment

plugins/destination/postgresql/client/migrate.go
yevgenypats: It is fine to call rollback after commit in pgx as long as we check if error is closed and then there is no need to log it. This way we ensure there is cleanup.

erezrokah: 👍 I'm good with always calling rollback, just at the moment it prints errors when it shouldn't so we should add a check there on the error type

plugins/destination/postgresql/client/migrate.go
yevgenypats: Indeed. maybe it's left overs. pretty strange it's here.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -164,7 +164,7 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
 	return nil
 }
 
-func (c *Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
+func (Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
```
candiduslynx, line 207:
```suggestion
func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
```

should not mix pointer & value receivers, just commit to one of those (in this case, pointer)

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,210 +3,326 @@ package client
...
-	reCreatePrimaryKeys := false
-
-	for _, col := range table.Columns {
-		columnName := pgx.Identifier{col.Name}.Sanitize()
-		columnType := c.SchemaTypeToPg(col.Type)
-		if columnType == "" {
-			c.logger.Warn().Str("table", table.Name).Str("column", col.Name).Msg("Column type not supported, skipping")
+func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, changes []schema.TableColumnChange) error {
+	for _, change := range changes {
+		switch change.Type {
```
erezrokah, line 193: Can we simplify this to:
```
if change.Type == schema.TableColumnChangeTypeAdd {
  if err := c.addColumn(ctx, table.Name, change.Current); err != nil {
    return err
  }
}
```

and drop the other cases? We already know the table is auto migratable when we call this method

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,210 +3,326 @@ package client
...
 
-			// if this column contains primary key we will need to recreate the primary key
-			if c.enabledPks() && col.CreationOptions.PrimaryKey {
-				reCreatePrimaryKeys = true
-			}
-			sql := "alter table " + tableName + " drop column " + columnName
-			// right now we will drop the column and re-create. in the future we will have an option to automigrate
-			if _, err := c.conn.Exec(ctx, sql); err != nil {
-				return fmt.Errorf("failed to drop column %s on table %s: %w", col.Name, table.Name, err)
+func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
```
erezrokah, line 207:
```suggestion
func canAutoMigrate(changes []schema.TableColumnChange) bool {
```

I'm not sure this needs to be a `Client` method

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,210 +3,326 @@ package client
...
+	result := true
+	for _, change := range changes {
+		switch change.Type {
+		case schema.TableColumnChangeTypeAdd:
+			if change.Current.CreationOptions.PrimaryKey || change.Current.CreationOptions.NotNull {
+				result = false
```
erezrokah, line 213: Can return early when on of these conditions happen?
```suggestion
				return false
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,210 +3,326 @@ package client
...
 			}
-			sql = "alter table " + tableName + " add column " + columnName + " " + columnType
-			if _, err := c.conn.Exec(ctx, sql); err != nil {
-				return fmt.Errorf("failed to add column %s on table %s: %w", col.Name, table.Name, err)
+		case schema.TableColumnChangeTypeRemove:
+			if change.Previous.CreationOptions.PrimaryKey || change.Previous.CreationOptions.NotNull {
+				result = false
```
erezrokah, line 217:
```suggestion
				return false
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,210 +3,326 @@ package client
...
 			}
+		case schema.TableColumnChangeTypeUpdate:
+			result = false
```
erezrokah, line 220:
```suggestion
			return false
```

erezrokah (approved):
I don't see anything blocking 🚀 I'm still testing, but if you'd like to merge I think that's fine and we can follow up with fixes if we have them

plugins/destination/postgresql/client/migrate.go
```diff
@@ -3,210 +3,326 @@ package client
...
-		if c.enabledPks() && pgPKs[col.Name] != col.CreationOptions.PrimaryKey {
-			c.logger.Info().Str("table", table.Name).Str("column", col.Name).Bool("pk", col.CreationOptions.PrimaryKey).Msg("Column exists with different primary keys")
-			reCreatePrimaryKeys = true
+// normalize the requested schema to be compatible with what Postgres supports
+func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
+	for _, table := range tables.FlattenTables() {
+		c.normalizeTable(table)
+	}
+	return tables
+}
```
erezrokah, line 234:
```suggestion
func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
	flattenedTables := tables.FlattenTables()
	for _, table := range flattenedTables {
		c.normalizeTable(table)
	}
	return flattenedTables
}
```

erezrokah (changes requested):
Sorry, found a bug

plugins/destination/postgresql/client/migrate.go
yevgenypats: I think it can in case we will want to add some logs.

plugins/destination/postgresql/client/migrate.go
yevgenypats: added.

plugins/destination/postgresql/client/migrate.go
yevgenypats: added

plugins/destination/postgresql/client/migrate.go
yevgenypats: added

plugins/destination/postgresql/client/migrate.go
yevgenypats: Good catch!

plugins/destination/postgresql/client/migrate.go
yevgenypats: Updated.

yevgenypats: It was more of a safe way to know if we missed anything.

erezrokah (approved):

hermanschaaf (approved):

plugins/destination/postgresql/client/types.go
```diff
@@ -94,3 +94,86 @@ func (*Client) SchemaTypeToCockroach(t schema.ValueType) string {
...
+		return schema.TypeFloat
+	case "uuid":
+		return schema.TypeUUID
+	case "text":
+		return schema.TypeString
+	case "bytea":
+		return schema.TypeByteArray
+	case "text[]":
+		return schema.TypeStringArray
+	case "timestamp without time zone":
```
hermanschaaf, line 123: This will only become a problem with CDC, but at some point we'll also need to support (or at least handle without panicking?) all the other postgres types, like `timestamptz`.

PR #7826, opened 2023-02-08
fix: Make Elasticsearch overwrite and delete-stale behavior consistent

hermanschaaf:
In the first iteration of the Elasticsearch destination, overwrite and delete-stale behavior were slightly inconsistent: overwrite would only overwrite in the current day's index, but delete-stale would stretch across all indexes. This PR fixes the behavior so that it acts consistently, but depends on the write mode. We also document the new index creation behavior.


disq (approved):

plugins/destination/elasticsearch/client/write.go
```diff
@@ -14,8 +14,15 @@ import (
...
 func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resources [][]any) error {
+	if len(resources) == 0 {
+		return nil
+	}
+
 	var buf bytes.Buffer
 	pks := pkIndexes(table) // do some work up front to avoid doing it for every resource
+	// get the sync time from the first resource in the batch (here we assume that all resources in the batch
+	// have the same sync time. At the moment this assumption holds.)
+	syncTime := resources[0][table.Columns.Index(schema.CqSyncTimeColumn.Name)].(time.Time)
```
hermanschaaf, line 25: I think this should be fine for the foreseeable future, but it would be nicer if `WriteTableBatch` received the sync time. An alternative (more complicated) solution could also be to actually write one batch per sync time received. I think for now this should be fine though.

PR #7898, opened 2023-02-10
chore(aws): Just send to chan

candiduslynx:


plugins/source/aws/resources/services/lambda/runtimes_fetch.go
```diff
@@ -9,12 +9,8 @@ import (
 )
 
 func fetchLambdaRuntimes(_ context.Context, _ schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
-	runtimes := make([]models.RuntimeWrapper, len(types.RuntimeProvidedal2.Values()))
-	for i, runtime := range types.RuntimeProvidedal2.Values() {
```
hermanschaaf, line 13: Why are we no longer using `RuntimeProvidedal2`?

candiduslynx: Moved to `typed.Runtime("")`

hermanschaaf (approved):

PR #7899, opened 2023-02-10
fix(aws): Fix `aws_iam_accounts` unmarshaling

KMConner:
#### Summary

Some properties in the AWS Account Summary were left at their default values (e.g. 0 or false).
Properties in `Account` struct were mapped from the map returned from the AWS API. (https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.GetAccountSummary)

While converting map to struct, the converter appear to be configured to refer json tag of each property.
The tags of the `Account` struct property has a json tag in the form of a snake case, but the AWS API return value is a Pascal case, so it is not correctly converted from a map to a structure.

I removed the json tags because it did not appear to be used elsewhere.


candiduslynx (approved):
Pretty neat!

https://go.dev/play/p/EOI81_gZpps
https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetAccountSummary.html

bbernays:
Great find! I think if we are removing the `json` tags we should also remove the `TagName: "json",` on this line: https://github.com/cloudquery/cloudquery/blob/f0f1e9faf3b53579daf9d4ae2509c6801e9243dd/plugins/source/aws/resources/services/iam/accounts_fetch.go#L22

bbernays (changes requested):
Great find! I think if we are removing the `json` tags we should also remove the `TagName: "json",` on this line: https://github.com/cloudquery/cloudquery/blob/f0f1e9faf3b53579daf9d4ae2509c6801e9243dd/plugins/source/aws/resources/services/iam/accounts_fetch.go#L22

plugins/source/aws/resources/services/iam/accounts_mock_test.go
```diff
@@ -15,32 +15,32 @@ func buildAccount(t *testing.T, ctrl *gomock.Controller) client.Services {
...
+		AttachedPoliciesPerRoleQuota      int32
+		AttachedPoliciesPerUserQuota      int32
+		Policies                          int32
+		PoliciesQuota                     int32
+		PolicySizeQuota                   int32
+		PolicyVersionsInUse               int32
+		PolicyVersionsInUseQuota          int32
+		VersionsPerPolicyQuota            int32
+		GlobalEndpointTokenVersion        int32
 	}{}
```
bbernays, line 44: I think this can be just replaced with `var accSummary models.Account`

KMConner:
@bbernays
I removed `TagName: "json"`.
Please review again!

erezrokah (approved):

erezrokah:
Thanks for the fix and quick follow up @KMConner I think this is ready to go 🚢

PR #7906, opened 2023-02-10
feat(aws): Add Support for Directconnect Locations

bbernays:


plugins/source/aws/resources/services/directconnect/locations.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/directconnect/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func Locations() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_directconnect_locations",
+		Description: `https://docs.aws.amazon.com/directconnect/latest/APIReference/API_Connection.html`,
```
erezrokah, line 13:
```suggestion
		Description: `https://docs.aws.amazon.com/directconnect/latest/APIReference/API_Location.html`,
```

plugins/source/aws/docs/tables/aws_directconnect_locations.md
```diff
@@ -0,0 +1,21 @@
...
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|region|String|
+|available_mac_sec_port_speeds|StringArray|
+|available_port_speeds|StringArray|
+|available_providers|StringArray|
+|location_code|String|
```
erezrokah, line 20: Do we know if the `account_id,region,location_code,location_name` combination is unique and can be used as a PK?

erezrokah (changes requested):
Thanks for the PR @bbernays added a few comments

plugins/source/aws/docs/tables/aws_directconnect_locations.md
bbernays: All of the other fields are attributes of a location... I would also be alright just hashing the entire object because it is 100% AWS defined without any user input

erezrokah (approved):

plugins/source/aws/docs/tables/aws_directconnect_locations.md
erezrokah: I think we can keep this for now and change based on users feedback

hermanschaaf: Seems like it's worth using `account_id,region,location_code`, if all the other fields are attributes of the location. Then this table will also work fine with `overwrite` mode

hermanschaaf (approved):

PR #7924, opened 2023-02-11
feat(aws): Add Support for Stack Sets

bbernays:


erezrokah:
Hi @bbernays, thank you for the PR 🚀 Can you explain a bit more what's missing from it?

bbernays:
@erezrokah Had to finish testing it with the different org permissions...

erezrokah (approved):

plugins/source/aws/resources/services/cloudformation/stacksets.go
```diff
@@ -0,0 +1,102 @@
...
+	// There is no way of determining if an account is a delegated admin or not. So just need to test it out and fail over to the other one
+	for _, callAs := range []types.CallAs{types.CallAsDelegatedAdmin, types.CallAsSelf} {
+		config := cloudformation.ListStackSetsInput{
+			CallAs: callAs,
+		}
+		paginator := cloudformation.NewListStackSetsPaginator(svc, &config)
+		for paginator.HasMorePages() {
+			page, err = paginator.NextPage(ctx)
+			if err != nil {
+				// Maybe try logging this error?
```
hermanschaaf, line 66: Yeah I think we should try and log this (but ignore the error that happens if called as the wrong type?)

PR #7925, opened 2023-02-11
feat(aws): Add New Elasticache Resources and Attributes

bbernays:


plugins/source/aws/docs/tables/aws_elasticache_events.md
```diff
@@ -0,0 +1,20 @@
...
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|region|String|
+|date (PK)|Timestamp|
```
erezrokah, line 17: Does this have enough resolution to be unique? Can two events happen at the same time?

erezrokah (changes requested):
Looks great 🚀  One comment on the PK

plugins/source/aws/docs/tables/aws_elasticache_events.md
bbernays: I don't know... All of the other fields can be empty... So not great candidates for PK... Maybe Hash the object unless you have another idea

hermanschaaf: Yeah let's rather hash the object if we're not certain 👍

erezrokah (approved):

PR #7937, opened 2023-02-12
fix: Just debugging error in pg

yevgenypats:
Trying to understand what is up with - https://github.com/cloudquery/cloudquery/actions/runs/4156256069/jobs/7189922011#step:8:182


yevgenypats:
Closing. Found the bug. Forgot to close a connection

PR #7942, opened 2023-02-12
feat(aws): Add Support for EC2 Managed Prefix List

bbernays:


plugins/source/aws/docs/tables/aws_ec2_managed_prefix_lists.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Columns
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|request_account_id (PK)|String|
```
erezrokah, line 16: What's the reason for the `request_` prefix? This table doesn't have any other `region` or `account columns

erezrokah (changes requested):
Thanks for the PR @bbernays, one question

plugins/source/aws/docs/tables/aws_ec2_managed_prefix_lists.md
bbernays: The `OwnerID` is actually the `account_id` that owns the prefix list... So didn't want users to get confused about the `account_id` field... And kept the `request_region` in there just to be consistent...

erezrokah (approved):

PR #7978, opened 2023-02-13
fix(aws)!: Add parent table ARNs as PKs to child tables for `aws_resiliencehub_*` tables

candiduslynx:
closes https://github.com/cloudquery/cloudquery-issues/issues/497
BEGIN_COMMIT_OVERRIDE
fix(aws): Add parent table ARNs as PKs to child tables for `aws_resiliencehub_*` resources
BREAKING-CHANGE: Add parent table ARNs as PKs to child tables for `aws_resiliencehub_*` resources

fix(aws): Remove `account_id` and `region` columns from `aws_resiliencehub_*` PKs
BREAKING-CHANGE: Remove `account_id` and `region` columns from `aws_resiliencehub_*` PKs

fix(aws): Use `physical_resource_identifier` instead of `physical_resource_id` in `aws_resiliencehub_app_version_resources` and `aws_resiliencehub_app_version_resource_mappings` PKs
BREAKING-CHANGE: Use `physical_resource_identifier` instead of `physical_resource_id` in `aws_resiliencehub_app_version_resources` and `aws_resiliencehub_app_version_resource_mappings` PKs
END_COMMIT_OVERRIDE


plugins/source/aws/docs/tables/aws_resiliencehub_app_assessments.md
```diff
@@ -24,13 +24,12 @@ The following tables depend on aws_resiliencehub_app_assessments:
 |_cq_sync_time|Timestamp|
 |_cq_id|UUID|
 |_cq_parent_id|UUID|
-|account_id|String|
-|region|String|
+|account_id (PK)|String|
+|region (PK)|String|
+|app_arn (PK)|String|
 |arn (PK)|String|
-|assessment_arn|String|
```
erezrokah, line 30: Was this a duplicate of `arn`?

erezrokah (approved):
Looks good, can we add a more detailed changelog on the changes?
I don't think we need to describe each table but write something like:
```
- Remove duplicate ARNs from `aws_resiliencehub_*` tables
- Add parent table ARNs as PKs to child tables for `aws_resiliencehub_*` tables
```

hermanschaaf (approved):
LGTM, just +1 to what Erez said

plugins/source/aws/resources/services/resiliencehub/columns.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func transformARN(path string) transformers.NameTransformer {
+	return func(field reflect.StructField) (string, error) {
+		if field.Name == path {
+			return "arn", nil
+		}
+		return transformers.DefaultNameTransformer(field)
+	}
+}
```
bbernays, line 44: When we create an `arn` field like this we are getting rid of the original column is that really what we want to do rather than just create a new column named `arn`?

hermanschaaf: IIRC in the past we used to rename to `arn`, but I looked at a number of examples now and since the transformer change I think we're leaving the originals in place and adding a new column called `arn`. That's also less of a breaking change, so I'd say let's do that here 👍

bbernays: I just did a quick search and I can't find any place where we use a `NameTransformer` aside from some fixes for `e_c_2` --> `EC2`

candiduslynx: Yeah. So, I'm either OK to leave the code as it now OR change the PK fields to the proper filed names from the structures (as it'll still be a breaking change), but not just adding a column with the same data.

So, there will be `app_arn` column in `apps` table, as an example of 2nd approach

hermanschaaf: It's actually important that the columns are called `arn`, as that gets used in the AWS resources view. Any tables without `arn` columns get ignored there, so we shouldn't call it `app_arn` if it's the primary `arn` for the table.

I'd say let's be consistent with other tables in the AWS plugin: if other tables are duplicating the data to two columns, which it looks like they are, let's do that here as well

candiduslynx: updated

plugins/source/aws/docs/tables/aws_resiliencehub_app_assessments.md
```diff
@@ -24,13 +24,13 @@ The following tables depend on aws_resiliencehub_app_assessments:
 |_cq_sync_time|Timestamp|
 |_cq_id|UUID|
 |_cq_parent_id|UUID|
-|account_id|String|
-|region|String|
+|account_id (PK)|String|
+|region (PK)|String|
+|app_arn (PK)|String|
 |arn (PK)|String|
```
bbernays, line 30: Is the `account_id` and `region` not in the ARN?

bbernays (changes requested):

plugins/source/aws/docs/tables/aws_resiliencehub_app_assessments.md
candiduslynx: I decided to make the PKs consistent for this service

bbernays: In the rest of the AWS plugin we try to use as few PK fields as possible... Why is diverging from the rest of the plugin worth it here?

candiduslynx: https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsresiliencehubservice.html#awsresiliencehubservice-resources-for-iam-policies
I'll update the code accordingly

plugins/source/aws/docs/tables/aws_resiliencehub_app_version_resource_mappings.md
```diff
@@ -18,8 +18,11 @@ This table depends on [aws_resiliencehub_app_versions](aws_resiliencehub_app_ver
 |_cq_parent_id|UUID|
 |account_id (PK)|String|
 |region (PK)|String|
+|app_arn (PK)|String|
```
bbernays, line 21: If we are adding `app_arn` doesn't that mean that `account_id` and `region` are duplicated?

plugins/source/aws/docs/tables/aws_resiliencehub_app_component_compliances.md
```diff
@@ -18,6 +18,8 @@ This table depends on [aws_resiliencehub_app_assessments](aws_resiliencehub_app_
 |_cq_parent_id|UUID|
 |account_id (PK)|String|
 |region (PK)|String|
+|app_arn (PK)|String|
```
bbernays, line 21: If we are adding `app_arn` doesn't that mean that `account_id` and `region` are duplicated?

plugins/source/aws/docs/tables/aws_resiliencehub_app_version_resources.md
```diff
@@ -18,8 +18,11 @@ This table depends on [aws_resiliencehub_app_versions](aws_resiliencehub_app_ver
 |_cq_parent_id|UUID|
 |account_id (PK)|String|
 |region (PK)|String|
+|app_arn (PK)|String|
```
bbernays, line 21: If we are adding `app_arn` doesn't that mean that `account_id` and `region` are duplicated?

plugins/source/aws/docs/tables/aws_resiliencehub_app_versions.md
```diff
@@ -22,4 +22,5 @@ The following tables depend on aws_resiliencehub_app_versions:
 |_cq_parent_id|UUID|
 |account_id (PK)|String|
 |region (PK)|String|
+|app_arn (PK)|String|
```
bbernays, line 25: If we are adding `app_arn` doesn't that mean that `account_id` and `region` are duplicated?

plugins/source/aws/docs/tables/aws_resiliencehub_component_recommendations.md
```diff
@@ -18,6 +18,8 @@ This table depends on [aws_resiliencehub_app_assessments](aws_resiliencehub_app_
 |_cq_parent_id|UUID|
 |account_id (PK)|String|
 |region (PK)|String|
+|app_arn (PK)|String|
+|assessment_arn (PK)|String|
```
bbernays, line 22: If we are adding `app_arn` doesn't that mean that `account_id` and `region` are duplicated?

bbernays (approved):

PR #7979, opened 2023-02-13
feat: Add 'gcp_projects' table

shimonp21:


hermanschaaf (approved):

erezrokah (changes requested):
Hi @shimonp21, what's the use case for this table?

I think we might want to have it contain all projects, while the multiplexer values are based on the spec (projects_filter, projects and folders)

erezrokah (approved):
@shimonp21 explained this to me on Slack, looks good

PR #7980, opened 2023-02-13
chore(docs): More Athena view creation Readme updates

hermanschaaf:
- Add instructions for Windows
- Add `GetBucketLocation` permission


yevgenypats (approved):
Looks good. Unrelated to this PR, In the S3 destination I think we should remove the "automatic" region finder to avoid more permission hell and let the user specify the region as they specify the bucket itself.

PR #7984, opened 2023-02-13
fix(aws)!: Add `account_id` and `region` to PK in `aws_directconnect_gateways`

disq:
Should hopefully fix the ARN PK issue.


hermanschaaf (approved):

plugins/source/aws/resources/services/directconnect/gateways_fetch.go
```diff
@@ -66,8 +68,15 @@ func fetchDirectconnectGatewayAttachments(ctx context.Context, meta schema.Clien
...
-	})
+func resolveGatewayARN(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	cl := meta.(*client.Client)
+	gw := resource.Item.(types.DirectConnectGateway)
+
+	return resource.Set(c.Name, arn.ARN{
+		Partition: cl.Partition,
+		Service:   "directconnect",
+		AccountID: *gw.OwnerAccount,
+		Region:    cl.Region,
```
disq, line 79: `cl.Region` here might be wrong maybe just add `Region` as one of the PKs and leave this empty.

candiduslynx: https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsdirectconnect.html#awsdirectconnect-resources-for-iam-policies
`dx-gateway` doesn't have the Region in ARN, so, I think, it'll be wrong to add it here.

candiduslynx (changes requested):

plugins/source/aws/resources/services/directconnect/gateways_fetch.go
disq: Good doc, fixed

candiduslynx (approved):

Issue #4383, opened 2022-11-11
Athena view for resources


jimr6007:
### Is your feature request related to a problem? Please describe.
may I ask for some help converting this to a Athena/Presto view:
https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/views/resources.sql

### Describe the solution you'd like

convert this Postgres SQL to Athena/Presto SQL
https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/views/resources.sql

bbernays:
@jimr6007 - Very interested in this request! How are you getting data into Athena/Presto in the first place though?

timoguin:
I'd love to have this too!

hermanschaaf:
Hey @jimr6007 and @timoguin, I just wanted to let you know that we are looking into this! 

As a first step, I've opened this related ticket to document how to load CloudQuery data into AWS Athena: https://github.com/cloudquery/cloudquery/pull/5943

While working on that ticket, like you, I was able to query the data from Athena by setting up a Glue Crawler. It seems to work! However, I found that the crawler gets the schema wrong for many tables, causing queries to fail or return incorrect results. It doesn't seem like there's any particular problem with our CSV output, and it even happens for small files with < 1000 rows, so I'm reaching out to AWS to see what we can do about this. If it doesn't work out, we may need to add support for Parquet as an output format to CloudQuery first, so Glue doesn't need to infer the schema in the first place :)

timoguin:
> Hey @jimr6007 and @timoguin, I just wanted to let you know that we are looking into this! 
> 
> As a first step, I've opened this related ticket to document how to load CloudQuery data into AWS Athena: https://github.com/cloudquery/cloudquery/pull/5943
> 
> While working on that ticket, like you, I was able to query the data from Athena by setting up a Glue Crawler. It seems to work! However, I found that the crawler gets the schema wrong for many tables, causing queries to fail or return incorrect results. It doesn't seem like there's any particular problem with our CSV output, and it even happens for small files with < 1000 rows, so I'm reaching out to AWS to see what we can do about this. If it doesn't work out, we may need to add support for Parquet as an output format to CloudQuery first, so Glue doesn't need to infer the schema in the first place :)

Awesome! I've been watching the output pretty intently!

Parquet would most certainly be the "righter" way, but great work on everything so far!

timoguin:
[...]

PR #7990, opened 2023-02-14
doc: Athena Resource View Standalone SQL Script

sfanous:
This adds a standalone SQL script that can be used to create an aws_resources view in Athena, similar to the Postgres one.

Provides an alternative option for https://github.com/cloudquery/cloudquery/issues/4383


hermanschaaf (approved):
Amazing, thanks for the contribution @sfanous !

yevgenypats (approved):

PR #7985, opened 2023-02-13
fix(aws)!: Add `resource_share_arn` and `region` to `aws_ram_principals` PK

disq:

PR #8010, opened 2023-02-14
fix(aws)!: Resource Access Manager (RAM) PK issues

disq:
The issue is the resource being shared so it's not located in the same account as it's owner. So if the user is fetching multiple accounts which have the resource share/share_perms the ARN clash is inevitable.

**Test Results**
- For `aws_ram_resources`: `arn` is the shared resource, `resource_share_arn` is the arn of the share itself, so both need to be in the PK, as well as AccountId and Region, because of the sharer/sharee being able to both view the same record.
- For `aws_ram_resource_shares`, the ARN's account_id/region is from the owner account and it has a UUID at the end of the ARN. So if you're sharing between two of your accounts the ARN clash is issue is real, adding current account/region from the querying account solves the problem.
- For the `aws_ram_resource_share_permissions`, the ARN is the arn of the permission itself, without any region or account or version, as in: `arn:aws:ram::aws:permission/AWSRAMDefaultPermissionGlueDatabase` and such (real value from test account) so adding the version and the account_id/region made sense.
- For `aws_ram_resource_share_associations`, there's no issue because the share "grantee" doesn't seem have access to the this list, it's only filled from the sharer account side.
- For `aws_ram_principals`, the previous PR #7985 seems to have fixed the issue, it's the same thing as the `aws_ram_resource_shares` issue.
- For `aws_ram_resource_share_invitations`, the ARN of the invitation is different from the resource share. But since you can share to multiple accounts or other entities (IAM users, etc) `ReceiverAccountId` and `ReceiverArn` needs to be added to the PK list. The problem is, `ReceiverArn` can be empty on some cases (if shared to account it's definitely empty) so it can't be a PK as-is. I added a new `receiver_combined` column to remedy the issue.


candiduslynx (approved):

erezrokah:
Made this a breaking change via the PR title

plugins/source/aws/resources/services/ram/resource_share_invitations.go
```diff
@@ -25,6 +25,14 @@ func ResourceShareInvitations() *schema.Table {
 					PrimaryKey: true,
 				},
 			},
+			{
+				Name:     "receiver_combined",
```
disq, line 29: Is this controversial? The reasoning is in the PR desc.

candiduslynx: It's OK, we did something similar in #6772

candiduslynx (approved):

disq:
Please hold off merging I'll take a look at a `aws_ram_resources` issue

candiduslynx (approved):

PR #8061, opened 2023-02-14
fix(aws)!: Use `log_group_arn` + `filter_name` for `aws_cloudwatchlogs_metric_filters` PK

disq:
Metric filters are defined under a log group so the ARN/PK should also include the log group name. Otherwise two of the equally named metric filters will clash.


plugins/source/aws/docs/tables/aws_cloudwatchlogs_metric_filters.md
```diff
@@ -2,7 +2,7 @@
 
 https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_MetricFilter.html
 
-The primary key for this table is **arn**.
+The composite primary key for this table is (**arn**, **log_group_name**).
```
disq, line 5: I don't like the custom/artificial `arn` here, maybe change it to be account+region+metric name+log group name?

disq: done in [`59b19a1` (#8061)](https://github.com/cloudquery/cloudquery/pull/8061/commits/59b19a18f618a7c2907736e77a5f3145c68da3bb)

plugins/source/aws/resources/services/cloudwatchlogs/metric_filters_fetch.go
```diff
@@ -31,12 +31,13 @@ func fetchCloudwatchlogsMetricFilters(ctx context.Context, meta schema.ClientMet
...
 func resolveMetricFilterArn(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	cl := meta.(*client.Client)
+	r := resource.Item.(types.MetricFilter)
 	a := arn.ARN{
 		Partition: cl.Partition,
 		Service:   "logs",
 		Region:    cl.Region,
 		AccountID: cl.AccountID,
-		Resource:  "metric_filter/" + aws.ToString(resource.Item.(types.MetricFilter).FilterName),
+		Resource:  "log-group:" + aws.ToString(r.LogGroupName) + ":metric_filter:" + aws.ToString(r.FilterName),
```
disq, line 40: inspired from https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazoncloudwatchlogs.html#amazoncloudwatchlogs-resources-for-iam-policies

The other "clean fix" is to remove this field and add a composite PK for account+region+log group+metric name.

candiduslynx: I think there's no such field as metric filter ARN, so I'm for removing this field and doing composite PK

candiduslynx (changes requested):

plugins/source/aws/resources/services/cloudwatchlogs/metric_filters_fetch.go
disq: Done in [`04e2744` (#8061)](https://github.com/cloudquery/cloudquery/pull/8061/commits/04e2744c90f228341eb50fa49da65c7f85c818c8)

plugins/source/aws/resources/services/cloudwatchlogs/metric_filters.go
```diff
@@ -13,18 +13,10 @@ func MetricFilters() *schema.Table {
...
-			{
-				Name:     "arn",
-				Type:     schema.TypeString,
-				Resolver: resolveMetricFilterArn,
-				CreationOptions: schema.ColumnCreationOptions{
-					PrimaryKey: true,
-				},
-			},
+			client.DefaultAccountIDColumn(true),
+			client.DefaultRegionColumn(true),
```
bbernays, line 19: Maybe convert the `LogGroupName` to an arn then we can use just those 2 columns (LogGroupARN and FilterName) for the PK... That way it also would make it easier to query and do joins between `aws_cloudwatchlogs_metric_filters` and `aws_cloudwatchlogs_log_groups`

disq: but yeah overall seems better, will do

bbernays: You would have to join on `account_id`, `region` and `log_group_name` because the names can repeat across regions... I have actually used that pattern in the past (putting the same log group name in every production region)

bbernays: It my view it is a potential "foot gun"... People won't always think about regional naming constraints and then they will blame CQ for duplicating data

plugins/source/aws/resources/services/cloudwatchlogs/metric_filters_fetch.go
```diff
@@ -28,15 +26,3 @@ func fetchCloudwatchlogsMetricFilters(ctx context.Context, meta schema.ClientMet
 	}
 	return nil
 }
```
bbernays, line 30: nit: I think we are trying to not have a separate file for `fetch` functions and just put it all in the same file where the table is defined

plugins/source/aws/resources/services/cloudwatchlogs/metric_filters.go
disq: Fixed in [`52376a1` (#8061)](https://github.com/cloudquery/cloudquery/pull/8061/commits/52376a1165d66143663baf209d2d740132cec8cc)

bbernays (approved):

PR #8070, opened 2023-02-14
fix(github)!: Change `github_organization_dependabot_alerts` PK from `(org,number)` to `(org,html_url)`

candiduslynx:
Closes https://github.com/cloudquery/cloudquery-issues/issues/482
Based on https://docs.github.com/en/rest/dependabot/alerts?apiVersion=2022-11-28#get-a-dependabot-alert.

### DIFF
--- a/plugins/source/github/client/columns.go
+++ b/plugins/source/github/client/columns.go
@@ -1,0 +1,13 @@
+package client
+
+import (
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+var OrgColumn = schema.Column{
+	Name:            "org",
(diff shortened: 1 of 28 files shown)


shimonp21 (approved):
LGTM 👍 
but maybe keep the relational-table names uppercase, as in the other plugins?

erezrokah:
> but maybe keep the relational-table names uppercase, as in the other plugins?

I think it's good to make them private so they are not mistakly added to the top level tables. We had a few cases before:
https://github.com/cloudquery/cloudquery/pull/7632

hermanschaaf (approved):

PR #8073, opened 2023-02-14
feat(aws): Add Support for CloudWatch Filter Subscriptions

bbernays:
#### Summary

Moved functions defined in `_fetch` files into the same file where the table is defined


erezrokah (approved):

plugins/source/aws/docs/tables/aws_cloudwatchlogs_log_group_subscription_filters.md
```diff
@@ -0,0 +1,27 @@
+# Table: aws_cloudwatchlogs_log_group_subscription_filters
+
+https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_SubscriptionFilter.html
+
+The composite primary key for this table is (**account_id**, **region**, **creation_time**, **filter_name**).
```
disq, line 5: No log group in the PKs? maybe add `log_group_arn` and remove account_id/region? (can't say anything about the creation time being in the PKs...)

plugins/source/aws/resources/services/cloudwatchlogs/subscription_filters.go
```diff
@@ -0,0 +1,39 @@
...
+		Multiplex:   client.ServiceAccountRegionMultiplexer("logs"),
+		Transform:   transformers.TransformWithStruct(&types.SubscriptionFilter{}, transformers.WithPrimaryKeys("FilterName", "CreationTime")),
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(true),
+			client.DefaultRegionColumn(true),
+		},
+	}
+}
+func fetchCloudwatchlogsSubscriptionFilters(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	var config cloudwatchlogs.DescribeSubscriptionFiltersInput
```
disq, line 27: Not using a log group from the parent as input, so this resource should be on the top level?

disq (changes requested):
Requesting changes for PKs and the subresourceness

plugins/source/aws/resources/services/cloudwatchlogs/subscription_filters.go
```diff
@@ -0,0 +1,47 @@
...
+		Multiplex:   client.ServiceAccountRegionMultiplexer("logs"),
+		Transform:   transformers.TransformWithStruct(&types.SubscriptionFilter{}, transformers.WithPrimaryKeys("FilterName", "CreationTime")),
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(false),
+			client.DefaultRegionColumn(false),
+			{
+				Name:        "log_group_arn",
+				Description: "The Amazon Resource Name (ARN) of the log group.",
+				Type:        schema.TypeString,
+				Resolver:    schema.ParentColumnResolver("arn"),
```
disq, line 27: Missing PK def here

plugins/source/aws/resources/services/cloudwatchlogs/subscription_filters.go
```diff
@@ -0,0 +1,47 @@
...
+			},
+		},
+	}
+}
+func fetchCloudwatchlogsSubscriptionFilters(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	config := cloudwatchlogs.DescribeSubscriptionFiltersInput{
+		LogGroupName: parent.Item.(types.LogGroup).LogGroupName,
```
disq, line 37: Having this here adds an extra call per log group. If the API allows loggroupless query (I'm inventing new words today) it would be better to do it that way to save on api calls, as with the metric filters. In that case the resource would need to be moved to the top level.

bbernays: It doesn't... That line had gotten removed while cleaning up the PR

disq (approved):

plugins/source/aws/resources/services/cloudwatchlogs/subscription_filters.go
bbernays: Screenshot 2023-02-14 at 10 15 28 AM

Issue #6247, opened 2023-01-02
[GitHub] Support GitHub Traffic API endpoints


hermanschaaf:
The GitHub API reports on repository traffic https://docs.github.com/en/rest/metrics/traffic?apiVersion=2022-11-28#about-the-repository-traffic-api but only retains data for 14 days. It would be useful if CloudQuery could be used to sync this data on a regular basis so that it can be retained somewhere for historical analysis.

PR #8074, opened 2023-02-14
feat(github): Add repositories multiplexer

erezrokah:
#### Summary

A first step for implementing https://github.com/cloudquery/cloudquery/issues/6247.
The APIs in that issue require passing a repository. We already list those twice for `github_workflows` and `github_repositories` so I think it's time to add a multiplexer


plugins/source/github/resources/services/actions/workflows_fetch.go
```diff
@@ -17,30 +17,18 @@ type Workflow struct {
...
-			actionOpts := &github.ListOptions{PerPage: 100}
-			for {
-				workflows, resp, err := c.Github.Actions.ListWorkflows(ctx, *repo.Owner.Login, *repo.Name, actionOpts)
-				if err != nil {
-					return err
-				}
-				for _, w := range workflows.Workflows {
-					res <- Workflow{Workflow: w, Repository: *repo.Name}
-				}
-				opts.Page = resp.NextPage
```
erezrokah, line 36: There was a bug here, this should be `actionOpts.Page`

plugins/source/github/resources/services/repositories/repositories_test.go
```diff
@@ -13,18 +13,6 @@ import (
 func buildRepositiories(t *testing.T, ctrl *gomock.Controller) client.GithubServices {
 	mock := mocks.NewMockRepositoriesService(ctrl)
 
-	var cs github.Repository
```
erezrokah, line 16: Moved to `GithubMockTestHelper` as needed by the multiplexer

plugins/source/github/client/client.go
```diff
@@ -74,6 +90,28 @@ func Configure(ctx context.Context, logger zerolog.Logger, s specs.Source, _ sou
...
 	}, nil
 }
+
+func discoverRepositories(ctx context.Context, client *github.Client, orgs []string) (map[string][]*github.Repository, error) {
+	opts := &github.RepositoryListByOrgOptions{ListOptions: github.ListOptions{PerPage: 100}}
+
+	orgRepos := make(map[string][]*github.Repository)
+	for _, org := range orgs {
+		for {
+			repos, resp, err := client.Repositories.ListByOrg(ctx, org, opts)
```
hermanschaaf, line 104: Since we're here.... it would be really nice if this also supported personal accounts, not just orgs 😄 

But happy to leave that for a future update also

erezrokah: Yeah, I think it's another feature, as our spec says `orgs` and we'll need to verify the API calls in all the resources (I think not all GitHub APIs are supported for personal accounts)

hermanschaaf (approved):
👍

PR #8075, opened 2023-02-14
chore: Remove crowdstrike plugin

bbernays:


website/versions/source-crowdstrike.json
```diff
@@ -1 +0,0 @@
-{ "latest": "plugins-source-crowdstrike-v1.1.4" }
```
erezrokah, line 1: I think you need to delete this file altogether

hermanschaaf (approved):

bbernays:
@hermanschaaf - This will need to be force merged I think

PR #8099, opened 2023-02-14
fix(azure): Ensure spec subscriptions are unique

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/491 (internal issue)
Fixes https://github.com/cloudquery/cloudquery-issues/issues/492 (internal issue)

After looking at the data and sample responses:
https://learn.microsoft.com/en-us/rest/api/firewall/azure-firewall-fqdn-tags/list-all?tabs=HTTP
https://learn.microsoft.com/en-us/rest/api/expressroute/express-route-service-providers/list?tabs=HTTP#expressrouteserviceprovider, I think the PKs for these tables are configured correctly `(subscription_id,id)` and can't cause duplicates as far as I can tell.

However I was able to reproduce the issue by specifying a duplicate subscription in the config, for example:

```yaml
kind: source
spec:
  name: azure
  version: v4.1.1
  destinations: [file]
  path: cloudquery/azure
  tables: [azure_network_express_route_service_providers, azure_network_azure_firewall_fqdn_tags]
  spec:
    subscriptions:
      - <sub_1>
      - <sub_1>
---
kind: destination
spec:
  name: file
  path: cloudquery/file
  version: "v1.1.0"
  write_mode: "append"
  spec:
    directory: "./cq_csv_output"
    format: "csv"
```


plugins/source/azure/client/client.go
```diff
@@ -116,9 +117,10 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source, _ source.Op
 		return nil, fmt.Errorf("failed to unmarshal gcp spec: %w", err)
 	}
 
+	uniqueSubscriptions := funk.Uniq(spec.Subscriptions).([]string)
```
candiduslynx, line 120: I'd rather use [slices.Compact](https://pkg.go.dev/golang.org/x/exp/slices#Compact) along with [slices.Sort](https://pkg.go.dev/golang.org/x/exp/slices#Sort)
```suggestion
	uniqueSubscriptions := slices.Compact(slices.Sort(spec.Subscriptions))
```

candiduslynx:
Do we want to ensure the same thing on other plugins? (e.g., AWS, GCP)

plugins/source/azure/client/client.go
candiduslynx: (because it's a part of Go tooling & they're bringing this package to main stdlib as well)

erezrokah: Nice one, but I'm not sure we should change the order of how subscriptions are defined in the spec, as I would expected CloudQuery to sync those in order

erezrokah:
> Do we want to ensure the same thing on other plugins? (e.g., AWS, GCP)

I think it's worth verify it

hermanschaaf (approved):

disq (approved):

PR #8106, opened 2023-02-14
refactor(github): Move fetch methods into table definitions files, use SDK resolvers instead of custom ones, use shared `RepositoryIdColumn`

candiduslynx:


erezrokah (changes requested):
Thanks for the PR @candiduslynx, can we scope the code in this PR to only do the change in the title?

It's hard to review with all the refactoring (which should be done in a separate `chore:` PR).
Also what's the reason we need `repository_id` as PK for `github_workflows`? Isn't `org` and `id` enough? Have we seen duplicate PK errors for this resource?

yevgenypats (changes requested):
Very hard to review. What does streamline means?

candiduslynx:
@yevgenypats 
1. Use `client.RepositoryIDColumn` where appropriate/possible
2. Move fetchers to the files defining tables (no extra `_fetch.go` file
3. Remove unnecessary resolvers:
  * `client.ResolveParentColumn` -> `schema.ParentColumnResolver`
  * drop `client.ResolveGrandParentColumn`)

erezrokah (approved):
Looks good, I updated the title to reflect the changes (from the title I think we can figure out that there are 2 distinct changes in this PR)

yevgenypats (approved):

PR #8124, opened 2023-02-15
chore(workflows): Run gen check prior to tests

candiduslynx:


erezrokah (commented):
I think `make gen` implicity builds as it runs `go run main.go doc` so needs to build the code?
Maybe keep `build` first? I don't have a strong opinon regardless

hermanschaaf (approved):
I would also probably move the build step to be before the `gen` step, but not a blocker for me 👍

Issue #7929, opened 2023-02-11
Azure Network Resources


jsonpr:
Network Interface IP Configurations
https://learn.microsoft.com/en-us/rest/api/virtualnetwork/network-interface-ip-configurations/list?tabs=HTTP

Network Subnets
https://learn.microsoft.com/en-us/rest/api/virtualnetwork/subnets/list?tabs=HTTP

PR #8126, opened 2023-02-15
feat(azure): Add network: interface_ip_configurations and virtual_network_subnets

disq:
Implements #7929


erezrokah (commented):
Looks great @disq, but I think we don't add a reference to the parent unless required for PK purposes

disq:
> Looks great @disq, but I think we don't add a reference to the parent unless required for PK purposes, isn't `ID` enough for that?

Yes `ID` includes the PK but wouldn't it be hard to match with parent as one would have to parse the ID and remove the last two parts (`/databases/<name>`) to get to the parent?

erezrokah:
> Yes `ID` includes the PK but wouldn't it be hard to match with parent as one would have to parse the ID and remove the last two parts (`/databases/<name>`) to get to the parent?

I agree. My comment intedened to keep the plugin consistent. If we add a parent ref we should do it for all relational tables and decide on a naming convention so maybe out of scope for this PR. cc @yevgenypats

erezrokah (approved):

Issue #7927, opened 2023-02-11
Azure ExpressRoute Resources


jsonpr:
Express Route Circuit Authorization
Express Route Circuit Peering
Express Route Filter and Filter Rules

PR #8128, opened 2023-02-15
feat(azure): Add network ExpressRoute circuit authorizations and peerings

disq:
Implements #7927


erezrokah (commented):
Same as https://github.com/cloudquery/cloudquery/pull/8126#pullrequestreview-1299315421

erezrokah (approved):

PR #8130, opened 2023-02-15
fix(aws)!: Step Function Executions And MapRuns

bbernays:
#### Summary

I have validated that data is able to be populated in all tables:
Screenshot 2023-02-15 at 9 18 19 AM

With no errors or panics
Screenshot 2023-02-15 at 9 18 53 AM


disq (approved):
PR prefix needs to be `fix(aws)!` because of the breaking (breaking what's already broken I know) change

Issue #7188, opened 2023-01-25
Support for AWS Trusted Advisor


jsonpr:
Support for AWS Trusted Advisor Recommendations (AWS Best Practices)

erezrokah:
Hi all 👋 Thanks for up voting the issue 🚀  I've opened a PR to support this resource and other support related resources.
We're having a bit of a discussion on the PR if to add data for each language or only for English. Feel free to comment on the PR

PR #8131, opened 2023-02-15
fix(aws): Use `credentialScope.region` if exists when generating regions data

erezrokah:
#### Summary

Required for https://github.com/cloudquery/cloudquery/issues/7188. I think our endpoints generation script has a bug.
We use endpoints as regions, but that's not a 1-1 mapping.
For example `access-analyzer` on region `ca-central-1` has more than one endpoint, the standard one and a FIPS one at `access-analyzer-fips.ca-central-1.amazonaws.com`, see:
https://github.com/aws/aws-sdk-go-v2/blob/1f9c50e881aca24ab2bf8ee0e8d15be7f19a056a/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen/endpoints.json#L114

Our current script adds `fips-ca-central-1` as a region, but it's not. It's just a different endpoint you can use for the REST API if you need to be FIPS compliant (as I far as I can undertand from https://aws.amazon.com/compliance/fips/).

Another issue, blocking https://github.com/cloudquery/cloudquery/issues/7188, is that because we use endpoints as regions we have in our JSON file the `aws-global` region for some services (like `support`, in the mentioned blocked issue). `aws-global` is not returned when we list enabled regions for an account, so any services that has it in our mapping files are simply skipped.

This PR gets the region from `credentialScope.region` if exists, otherwise uses the endpoint key. This is based on the logic in the Go SDK that resolves endpoints from regions (basically the inverted logic).
If you're really curious see the code that resolves the correct endpoint based on the region:
https://github.com/aws/aws-sdk-go-v2/blob/1f9c50e881aca24ab2bf8ee0e8d15be7f19a056a/service/support/endpoints.go#L95
https://github.com/aws/aws-sdk-go-v2/blob/1f9c50e881aca24ab2bf8ee0e8d15be7f19a056a/internal/endpoints/v2/endpoints.go#L91
https://github.com/aws/aws-sdk-go-v2/blob/1f9c50e881aca24ab2bf8ee0e8d15be7f19a056a/internal/endpoints/v2/endpoints.go#L77

Opening for feedback as I'm not 100% sure I got this right


bbernays:
So I am not seeing the impact of having more regions listed in the file. We validate regions input based on `ec2.DescribeRegions`. For the the `global-aws` Any resource that uses `aws-global` doesn't use a regional multiplexer.

erezrokah:
> For the the `global-aws` Any resource that uses `aws-global` doesn't use a regional multiplexer.

What multiplexer they use? I tried using the account one for https://github.com/cloudquery/cloudquery/issues/7188 but that doesn't work well, as we use the first account region which maps to an endpoint that doesn't support for that resource:
https://github.com/cloudquery/cloudquery/blob/2425cce25e642473d90bcdf84b64399a8ac0808d/plugins/source/aws/client/client.go#L388

bbernays (approved):
LGTM. Thanks for explaining it to me

erezrokah:
OK so I think this will also fix issues like:
image

See https://github.com/cloudquery/cloudquery/pull/7816/files and https://github.com/cloudquery/cloudquery/pull/7811/files. I guess AWS was experimenting with a new FIPS endpoint

PR #8134, opened 2023-02-15
feat(aws): Support for Org Policies, Organizational Units and Roots

bbernays:


plugins/source/aws/resources/services/organizations/organizational_units.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return nil
+}
+
+func getOUs(ctx context.Context, accountsApi services.OrganizationsClient, res chan<- any, parentID string) error {
+	q := []string{parentID}
+	var ou string
+	seenOUs := map[string]struct{}{}
+
+	for len(q) > 0 {
```
hermanschaaf, line 58: I thought this code looked familiar / quite like my own style, then I remembered I wrote it a while back for the AWS client. LGTM :)

bbernays: 100% I copied that straight from your work in the AWS Client...

hermanschaaf (approved):
LGTM 🚀

PR #8140, opened 2023-02-16
feat: Add Support resources (includes AWS Trusted Advisor)

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/7188 (trusted advisor is under `support package in the Go SDK).

So resources (e.g. severity) can be sync in different languages so added a multiplex for it (might be an overkill, and we can sync only english data, so happy to get feedback on this). 

Missing from this PR is https://docs.aws.amazon.com/awssupport/latest/APIReference/API_DescribeAttachment.html. Should I add it too?


hermanschaaf:
Are the Trusted Advisor results the same in all languages, just translated, or could they be different?

erezrokah:
> Are the Trusted Advisor results the same in all languages, just translated, or could they be different?

Specifally with `aws_support_trusted_advisor_checks` the checks are the same (same id different language), so the difference in the data is the translation. That's the reason for https://github.com/cloudquery/cloudquery/pull/8140/files#diff-b61b85903739a0f1122e5014a16c51fea232e93e6eeed0febaccdafb9bcd1981R31

plugins/source/aws/docs/tables/aws_support_trusted_advisor_check_results.md
```diff
@@ -0,0 +1,25 @@
...
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|category_specific_summary|JSON|
+|check_id|String|
```
hermanschaaf, line 21: Could `check_id` serve as a primary key here?

erezrokah: Yes! Good catch, I missed that (thought we get a results array). Fixed in https://github.com/cloudquery/cloudquery/pull/8140/commits/1a10e80ef72917c3b0f3528b58fe838ea5df5da7

erezrokah:
See also https://github.com/cloudquery/cloudquery/issues/7188#issuecomment-1432843947

plugins/source/aws/docs/tables/aws_support_trusted_advisor_check_summaries.md
```diff
@@ -0,0 +1,25 @@
...
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id (PK)|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|category_specific_summary|JSON|
+|check_id|String|
```
hermanschaaf, line 21: It looks like this table could also use `check_id` as PK 🤔 But I think we also need to add `account_id`, `region` and `language_code` as columns here then (all the parent PK parts, basically), so that it's guaranteed to be unique

plugins/source/aws/docs/tables/aws_support_trusted_advisor_check_results.md
```diff
@@ -0,0 +1,25 @@
...
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|category_specific_summary|JSON|
+|check_id (PK)|String|
```
hermanschaaf, line 21: same comment here (let me know if I got it wrong), but it looks like in addition to `check_id` we need `account_id`, `region` and `language_code`

plugins/source/aws/resources/services/support/trusted_advisor_check_summaries.go
```diff
@@ -0,0 +1,59 @@
...
+
+func fetchTrustedAdvisorCheckSummaries(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	c := meta.(*client.Client)
+	// No need to get the summary for each language, as those are the same have the same check id
+	if c.LanguageCode != "en" {
+		return nil
+	}
+	svc := c.Services().Support
+	check := parent.Item.(types.TrustedAdvisorCheckDescription)
+	input := support.DescribeTrustedAdvisorCheckSummariesInput{CheckIds: []string{aws.ToString(check.Id)}}
```
hermanschaaf, line 37: Looks like another case where https://github.com/cloudquery/plugin-sdk/issues/338 would have helped :)

hermanschaaf (changes requested):
Looks great! Let's just sort out those child table PKs if possible

plugins/source/aws/client/client.go
```diff
@@ -116,12 +117,16 @@ func (c *Client) Logger() *zerolog.Logger {
...
 func (c *Client) ID() string {
-	return strings.TrimRight(strings.Join([]string{
+	idStrings := []string{
 		c.AccountID,
 		c.Region,
 		c.AutoscalingNamespace,
 		string(c.WAFScope),
-	}, ":"), ":")
+	}
+	if c.LanguageCode != "" {
```
disq, line 126: Since we do a TrimRight(":") two lines below, we don't need this block. Should just add `c.LanguageCode` to the `[]string` defn above.

plugins/source/aws/resources/services/support/trusted_advisor_checks.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func TrustedAdvisorChecks() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_support_trusted_advisor_checks",
+		Description: `https://docs.aws.amazon.com/awssupport/latest/APIReference/API_DescribeTrustedAdvisorChecks.html`,
+		Resolver:    fetchTrustedAdvisorChecks,
+		Transform:   transformers.TransformWithStruct(&types.TrustedAdvisorCheckDescription{}, transformers.WithPrimaryKeys("Id")),
+		Multiplex:   client.ServiceAccountRegionsLanguageCodeMultiplex("support", []string{"en", "fr", "de", "id", "it", "ja", "ko", "pt_BR", "es", "zh", "zh_TW"}),
```
disq, line 20: Could extract the language list as a global []string here (unexported) so that it can be reused in the test as well. Otherwise the list being long could cause testing issues in the future (eg. test cases being forgotten for a newly added lang code)

plugins/source/aws/resources/services/support/trusted_advisor_check_summaries.go
erezrokah: 💯

plugins/source/aws/resources/services/support/trusted_advisor_checks.go
erezrokah: I think the test will fail it that case with a missing call (you need to mock each specific language code), but I'll do that anyways

plugins/source/aws/docs/tables/aws_support_trusted_advisor_check_results.md
erezrokah: `account_id` and `region` yes, but `language_code` no as we don't get this for each language:
https://github.com/cloudquery/cloudquery/pull/8140/files#diff-e2281beb0f75f74e18cbf10164be3cb5c89e063090181e2493b16dc915937e64R32

hermanschaaf: Yeah, but I would probably still include `language_code`, so it doesn't become a breaking change later (seeing as `language_code` is also part of the parent PK)

erezrokah (commented):
Thanks for the review @hermanschaaf and @disq. Good call on the PKs, followed up in https://github.com/cloudquery/cloudquery/pull/8140/commits/3fd3535ede405f0d2764a3db558a4a9629ee0b93 (with other fixes)

hermanschaaf (approved):

Issue #8137, opened 2023-02-15
aws_cloudformation_stack_resources omits stack_name and stack_id


jdg-journeyfront:
### Describe the Bug

Having successfully synced with AWS to my Postgres database, I'm inspecting my CloudFormation resources and it seems it's not possible to join the `aws_cloudformation_stacks` table with `aws_cloudformation_stack_resources` due to their being neither `stack_id` nor `stack_name` fields in that table.

### Expected Behavior

`aws_cloudformation_stack_resources` be written with `stack_id` and `stack_name` columns, especially as these fields are available from the API.

### CloudQuery Version

    $ cloudquery --version
    cloudquery version 2.3.9

Also with AWS plugin `v15.0.0`

### Steps to Reproduce

Here's my bare basic config file

``` yaml
---
kind: source
spec:
  name: aws
  path: cloudquery/aws
  version: v15.0.0
  destinations:
  - postgresql
  # spec: {}
---
kind: destination
spec:
  name: postgresql
  path: cloudquery/postgresql
  version: v2.0.10
  spec:
    connection_string: postgresql://postgres@172.17.0.2/postgres?sslmode=disable
```

SQL that I can't write.

``` sql
SELECT *
  FROM aws_cloudformation_stacks stack
  JOIN aws_cloudformation_stack_resources resource ON (stack.stack_id = resource.stack_id)
```

PR #8141, opened 2023-02-16
feat(aws): CloudFormation stack_resources to include stack id

disq:
Implements #8137 

~Having both stack name and stack id could be too much, maybe just keep stack_id?~


erezrokah (approved):
Looks, good, I'd use whatever is the PK for the parent

hermanschaaf (approved):

hermanschaaf:
The parent PK seems to be `arn`, so maybe we should include `stack_arn` as well? (We usually do this)

disq:
> The parent PK seems to be `arn`, so maybe we should include `stack_arn` as well? (We usually do this)

`id` and `arn` have the same value. We just took the api's exposed "id" and probably saw it's in the ARN format and created a duplicate `arn` column.

PR #8152, opened 2023-02-16
fix(aws): Fix error handling int `aws_alternate_contact` fetching

KMConner:
#### Summary

There are three types of AWS alternate contacts; BILLING, OPERATIONS, and SECURITY.
If one of the three contacts are not set, the plugin cannot fetch the other alternate contacts too.
This is because the fetch ends if an error, including `ResourceNotFoundException` occurrs.

So I fixed the error handling process so that `ResourceNotFoundException` will be ignored and other alternate contact information will be retrieved.


hermanschaaf (approved):
Looks good to me, thanks for the fix!

PR #8158, opened 2023-02-16
feat(aws): Use ServiceAccountRegion Multiplexer

bbernays:
#### Summary

Now that @erezrokah fixed the fact that the region in the data.json file doesn't have `aws-global` anymore we can now use the same multiplexer for (nearly) all resources

### DIFF
--- a/plugins/source/aws/resources/services/account/alternate_contacts.go
+++ b/plugins/source/aws/resources/services/account/alternate_contacts.go
@@ -12,7 +12,7 @@ func AlternateContacts() *schema.Table {
 		Name:        "aws_account_alternate_contacts",
 		Description: `https://docs.aws.amazon.com/accounts/latest/reference/API_AlternateContact.html`,
 		Resolver:    fetchAccountAlternateContacts,
-		Multiplex:   client.AccountMultiplex,
+		Multiplex:   client.ServiceAccountRegionMultiplexer("account"),
 		Transform:   transformers.TransformWithStruct(&types.AlternateContact{}),
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(true),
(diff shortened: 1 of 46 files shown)


erezrokah (approved):

plugins/source/aws/resources/services/account/alternate_contacts.go
```diff
@@ -12,7 +12,7 @@ func AlternateContacts() *schema.Table {
 		Name:        "aws_account_alternate_contacts",
 		Description: `https://docs.aws.amazon.com/accounts/latest/reference/API_AlternateContact.html`,
 		Resolver:    fetchAccountAlternateContacts,
-		Multiplex:   client.AccountMultiplex,
+		Multiplex:   client.ServiceAccountRegionMultiplexer("account"),
```
hermanschaaf, line 15: I'd have to check the most recent code, and I may very well be missing something, but  `client.AccountMultiplex` (multiplex over all accounts) seems quite different from `client.ServiceAccountRegionMultiplexer("account")` (multiplex over all accounts and regions for a service called `"account"`) - is this definitely right?

bbernays: It is only correct because in actuality there is only a single region (per partition) that supports this service

bbernays: [Here is how the `account` service is represented in that file now](https://github.com/erezrokah/cloudquery/blob/9397419590b9724d1ed5bd9743579ae742ec639f/plugins/source/aws/client/data/partition_service_region.json#L43-L47)

hermanschaaf: Ah cool, I see, thanks for the explanation!

hermanschaaf (approved):
🚀

PR #8162, opened 2023-02-16
fix(aws): Ignore Default ECR Errors

bbernays:
#### Summary

The following errors are thrown when a resource by default doesn't exist


erezrokah (approved):

hermanschaaf (approved):
🚀

PR #8164, opened 2023-02-16
fix(aws): Fix error for Empty Backup Notification

bbernays:
#### Summary

Fix is implemented based on suggestion in https://github.com/aws/aws-sdk-go-v2/issues/1885#issuecomment-1282663934


plugins/source/aws/resources/services/backup/vaults_fetch.go
```diff
@@ -98,8 +98,14 @@ func resolveVaultNotifications(ctx context.Context, meta schema.ClientMeta, reso
...
 	)
+
 	if err != nil {
-		return err
+		// This is a service/SDK issue.
+		// Workaround is suggested here https://github.com/aws/aws-sdk-go-v2/issues/1885#issuecomment-1282663934
+		if !strings.Contains(err.Error(), " Failed reading notifications from database for Backup vault ") {
+			return err
+		}
+		result = &backup.GetBackupVaultNotificationsOutput{}
```
hermanschaaf, line 108: should we maybe just return nil at this point, rather than set an empty column?

bbernays: I thought about that, but I figured this was the more accurate way of representing that error, because the error just means that there were no notifications found. But I could go either way...

PR #8165, opened 2023-02-16
chore(website): Update integration logos

hermanschaaf:
Before:

Screenshot 2023-02-16 at 16 51 33

After:

Screenshot 2023-02-16 at 16 52 22

And light mode:
Screenshot 2023-02-16 at 16 53 28

It took a bit of doing, and could probably use some more tweaking, but I think it's good to have an up-to-date list of integrations and feature them a bit more prominently.

One thing to note is that I've changed from svg's that were built into the code to `` tags. The new way will take a bit longer to load, but the files are very small, so I think it will be basically negligible, and be easier to maintain for us until we invest in some automation to build it into the static HTML again.


erezrokah (approved):

yevgenypats (approved):
Wow, looks nice 🚀 Real frontend expert 🪄

PR #8175, opened 2023-02-17
feat(test): Add testdata table

disq:
So that each cqtype can be used as source for destination testing/development.


plugins/source/test/resources/services/testdata.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+		table.Columns[i].CreationOptions = schema.ColumnCreationOptions{}
+		table.Columns[i].Resolver = schema.PathResolver(table.Columns[i].Name)
+	}
+
+	data := testdata.GenTestData(table)
+	dataAsMap := make(map[string]any)
+	for i, c := range table.Columns {
+		if data[i].GetStatus() == schema.Present {
+			dataAsMap[c.Name] = data[i].String()
```
disq, line 25: `data[i].Get()` fails with array types, but string seems to work with all.

plugins/source/test/resources/services/testdata.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/testdata"
+)
+
+func TestDataTable() *schema.Table {
+	table := testdata.TestSourceTable("test_testdata_table")
+	for i, c := range table.Columns {
+		if strings.HasPrefix(c.Name, "_cq_") {
+			table.Columns[i].Name = "test" + c.Name
```
disq, line 15: To reduce complexity, rather than removing the `_cq_id` and `_cq_parent_id` columns here I just add 'test' to the names.

plugins/source/test/resources/services/testdata.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+		table.Columns[i].CreationOptions = schema.ColumnCreationOptions{}
```
disq, line 17: This resets any primary key/not null definitions.

candiduslynx (approved):

hermanschaaf (approved):

PR #8228, opened 2023-02-19
fix(missing-tables): Skip test plugin

erezrokah:
#### Summary

The script logic started failing with https://github.com/cloudquery/cloudquery/pull/8175. The easiest solution is to skip the test plugin as it doesn't follow other plugins conventions


erezrokah:
Going to force this to fix the CI

PR #8230, opened 2023-02-19
feat(deps): Update BigQuery plugin-sdk to v1.39.0

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/8156.
The new SDK adds a new test:
1. Create initial scheme and write stuff
2. Add a new column
3. Migrate to new schema
4. Write more stuff
5. Validate new stuff was created

BiqQuery can have a few minutes delay until the schema is updated, so the new test fails when validating the new data (the data is missing the new column).
This PR adds a test configuration to wait additional time for schema propagation. Another option is to skip the test


plugins/destination/bigquery/client/spec.go
```diff
@@ -34,6 +34,9 @@ type Spec struct {
 	DatasetLocation       string                 `json:"dataset_location"`
 	TimePartitioning      TimePartitioningOption `json:"time_partitioning"`
 	ServiceAccountKeyJSON string                 `json:"service_account_key_json"`
+
+	// We override this in tests to ensure schema is ready
+	MinTriesSleepTimeSeconds int `json:"min_tries_sleep_time_seconds"`
```
candiduslynx, line 39: If that's a test-only field, don't export it & fill it manually

candiduslynx (changes requested):

yevgenypats:
Looks good but seems tests still fail, is it because of the timeout? I would prefer the delay instead of failing the tests

plugins/destination/bigquery/client/spec.go
erezrokah: I tried, couldn't find another way to pass configuration to plugins except the spec

erezrokah:
> Looks good but seems tests still fail, is it because of the timeout? I would prefer the delay instead of failing the tests

I'm taking another look, there might be an additional issue regarding comparison order similar to https://github.com/cloudquery/plugin-sdk/pull/565 and https://github.com/cloudquery/plugin-sdk/pull/561

plugins/destination/bigquery/client/spec.go
hermanschaaf: It's all in the same package, so think it can still be made private, so it doesn't become accessible to users in the yaml?

erezrokah: Sure I can do that (a global var), wasn't sure if it would be cleaner, but will give it a shot

hermanschaaf: It doesn't have to be a global var, it can be a private value in the spec, right?

erezrokah: I tried it but it doesn't propagate to the plugin as we do `UnmarshalSpec` during init so any non exported fields are missing after that

erezrokah: See relevant code https://github.com/cloudquery/cloudquery/blob/da5d2f8a4a0aae5f16905224e4901f64c733ca4d/plugins/destination/bigquery/client/client.go#L34.

If it's not exported in the struct via a JSON tag, we won't have that value in the plugin spec

hermanschaaf: Ah okay, I see, thanks! Maybe an unexported global var is the way to go then: even though that's not great either, at least it doesn't impact user-facing config then

erezrokah: Moved it to a global var but happy to do any other working solution.
Maybe another way to solve the delay is using an emulator (e.g. https://github.com/goccy/bigquery-emulator) instead of the real service for tests. Maybe then the schema update is instant

hermanschaaf: Ah yeah, I thought about that too but then saw that there was no official bigquery emulator. I didn't see this unofficial one, that looks like a great option!

erezrokah: Opened https://github.com/cloudquery/cloudquery/issues/8258 with some more context

disq (approved):

plugins/destination/bigquery/client/spec.go
erezrokah: @candiduslynx do you have any concerns with this PR? You review is blocking it

candiduslynx: approved

hermanschaaf:
Looks like the tests are now failing @erezrokah:

```

erezrokah:
Tried the emulator but ran into https://github.com/goccy/bigquery-emulator/issues/150

erezrokah:
I'll give the test delay one more go. If it doesn't work we'll have to disable it otherwise the tests will take too long

erezrokah:
Decided to skip the migrate test for now

PR #8255, opened 2023-02-20
feat(deps): Update ClickHouse plugin-sdk to v1.38.2

erezrokah:
#### Summary

Will need the ~~3~~ 2 fixes in https://github.com/cloudquery/plugin-sdk/pull/694 (multiple PRs coming soon)


plugins/destination/clickhouse/client/migrate.go
```diff
@@ -3,47 +3,84 @@ package client
...
 
 // Migrate relies on the CLI/client to lock before running migration.
 func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
-	current, err := c.getTableDefinitions(ctx)
+	currentSchema, err := c.getTableDefinitions(ctx)
 	if err != nil {
 		return err
 	}
 
-	const maxConcurrentMigrate = 10
```
erezrokah, line 20: Will bring this back in a bit, removed to make debugging easier

plugins/destination/clickhouse/client/client_test.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/cloudquery/plugin-sdk/plugins/destination"
 )
 
+var migrateStrategy = destination.MigrateStrategy{
+	AddColumn:           destination.DataLossNone,
+	AddColumnNotNull:    destination.DataLossTable,
+	RemoveColumn:        destination.DataLossNone,
```
candiduslynx: It's literally a column loss (not in a sense that we drop tada, but in a sense that no data will be filled in after the migration to new source schema).
```suggestion
	RemoveColumn:        destination.DataLossColumn,
```

candiduslynx: However, maybe we need to clarify what we expect to be represented. Is it the immediate impact?

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -3,47 +3,84 @@ package client
...
+			result = append(result, t.Name)
+			tableChanges = append(tableChanges, changes)
+		}
 	}
+	return result, tableChanges
+}
+
+func (*Client) canSafeMigrate(changes []schema.TableColumnChange) bool {
+	for _, change := range changes {
+		if change.Type == schema.TableColumnChangeTypeAdd && change.Current.CreationOptions.NotNull {
```
candiduslynx, line 79: I really think we need `migrate` or `changes` package in `plugin-sdk`.
These names get too long.

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -3,47 +3,84 @@ package client
...
+			return false
+		}
+
+		if change.Type == schema.TableColumnChangeTypeRemove && change.Previous.CreationOptions.NotNull {
```
candiduslynx, line 83: New/Current will be better that previous/current, as current should represent the current data schema, not the intended.
current = have in a have/want paradigm, basically.

plugins/destination/clickhouse/queries/tables.go
```diff
@@ -0,0 +1,41 @@
...
+	normalized := normalizeTable(table)
+	strBuilder := strings.Builder{}
+	strBuilder.WriteString("CREATE TABLE ")
+	strBuilder.WriteString(sanitizeID(normalized.Name))
+	strBuilder.WriteString(" (\n")
+	for _, col := range normalized.Columns {
+		strBuilder.WriteString(sanitizeID(col.Name))
+		strBuilder.WriteString(" ")
+		strBuilder.WriteString(chType(&col))
+		strBuilder.WriteString(",\n")
```
candiduslynx, line 29: will become
```
(
  a,
  b,
)
```

use join as was done in `definitions` func

plugins/destination/clickhouse/queries/testdata/create_table.sql
```diff
@@ -1,9 +1,9 @@
...
-  `extra_inet_arr_col` Array(Nullable(String))
-) ENGINE = MergeTree ORDER BY (`_cq_id`)
\ No newline at end of file
+`_cq_id` UUID,
+`_cq_parent_id` Nullable(UUID),
+`_cq_source_name` Nullable(String),
+`_cq_sync_time` Nullable(DateTime64(9)),
+`extra_col` Float64,
+`extra_inet_col` Nullable(String),
+`extra_inet_arr_col` Array(Nullable(String)),
```
candiduslynx, line 8: wrong, CH should fail on extra coma

plugins/destination/clickhouse/go.sum
```diff
@@ -48,6 +48,8 @@ github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMn
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
 github.com/cloudquery/plugin-sdk v1.37.1 h1:1/9cYMUSqFRkMqq4p+8ZThdPQ+09p+8Nlrt3ii9Fbf8=
 github.com/cloudquery/plugin-sdk v1.37.1/go.mod h1:anjM5gY7qgJDde9CCIlJbeKuFOxM4BBrFeX1jyU1dxo=
+github.com/cloudquery/plugin-sdk v1.38.1 h1:nyC6/rR1NqUEDfUYE4rMpc3arwhRf9d+gCu/R1IhkJ8=
```
candiduslynx: tidy?

plugins/destination/clickhouse/queries/tables_test.go
```diff
@@ -17,7 +17,7 @@ func TestCreateTable(t *testing.T) {
 			schema.Column{
 				Name:            "extra_col",
 				Type:            schema.TypeFloat,
-				CreationOptions: schema.ColumnCreationOptions{PrimaryKey: true},
+				CreationOptions: schema.ColumnCreationOptions{NotNull: true},
```
erezrokah, line 20: No need to test primary keys for a destination that supports only append mode.
We do however need a not null key as we using as a sorting key

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -52,27 +89,69 @@ func (c *Client) createTable(ctx context.Context, table *schema.Table) (err erro
...
+	for _, change := range changes {
+		switch {
+		case change.Type == schema.TableColumnChangeTypeAdd && !change.Current.CreationOptions.NotNull:
+			c.logger.Debug().Str("table", table.Name).Str("column", change.Current.Name).Msg("Adding new column")
+			err := c.conn.Exec(ctx, queries.AddColumn(table.Name, &change.Current))
+			if err != nil {
+				return err
+			}
+		case change.Type == schema.TableColumnChangeTypeUpdate && !change.Current.CreationOptions.NotNull:
+			c.logger.Debug().Str("table", table.Name).Str("column", change.Current.Name).Msg("dropping changed non null column")
```
candiduslynx:
```suggestion
			c.logger.Debug().Str("table", table.Name).Str("column", change.Current.Name).Msg("Recreating nullable column")
```

candiduslynx (changes requested):

plugins/destination/clickhouse/queries/testdata/create_table.sql
erezrokah: It doesn't (see the testing passing) and also https://clickhouse.com/docs/en/sql-reference/statements/create/table/

image

plugins/destination/clickhouse/client/migrate.go
erezrokah: Can you open an issue for it in the SDK? I don't think it's related to this PR

plugins/destination/clickhouse/client/migrate.go
erezrokah: Those are SDK structs, unrelated to this PR. Do you mind opening an SDK issue to improve the API?

plugins/destination/clickhouse/client/migrate.go
erezrokah: This was removed as we now recreate a table on column changes due to not going with https://github.com/cloudquery/plugin-sdk/pull/697

erezrokah (commented):
Thanks for the review @candiduslynx, I updated the PR as we decided not to go with https://github.com/cloudquery/plugin-sdk/pull/697 yet

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -12,67 +14,140 @@ import (
...
+			if current == nil {
 				return c.createTable(ctx, table)
 			}
 
-			return c.autoMigrateTable(ctx, curr, table)
+			err := c.autoMigrate(ctx, table, current)
+			if err != nil {
+				return err
+			}
+			c.logger.Err(err).Str("table", table.Name).Msg("Migrating table done")
```
candiduslynx, line 52: move back to defer

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -12,67 +14,140 @@ import (
...
+		changes := t.GetChanges(current)
+		if !c.canSafeMigrate(changes) {
+			result = append(result, t.Name)
+			tableChanges = append(tableChanges, changes)
+		}
+	}
+	return result, tableChanges
+}
+
+func (*Client) canSafeMigrate(changes []schema.TableColumnChange) bool {
```
candiduslynx, line 77:
```suggestion
func (*Client) canSafelyMigrate(changes []schema.TableColumnChange) bool {
```

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -12,67 +14,140 @@ import (
...
+			return false
+		}
+
+		if change.Type == schema.TableColumnChangeTypeRemove && change.Previous.CreationOptions.NotNull {
+			return false
+		}
+
+		if change.Type == schema.TableColumnChangeTypeUpdate {
+			return false
+		}
```
candiduslynx, line 89: we have 2 options for `change.Current.CreationOptions.NotNull` to exit early (`schema.TableColumnChangeTypeAdd` & `schema.TableColumnChangeTypeRemove`).
Maybe we can go with this code instead?
```suggestion
		switch change.Type {
		case schema.TableColumnChangeTypeUpdate:
			return false
		case schema.TableColumnChangeTypeAdd,
			schema.TableColumnChangeTypeRemove:
			if change.Current.CreationOptions.NotNull {
				return false
			}
		}
```

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -12,67 +14,140 @@ import (
...
+	c.logger.Debug().Str("table", table.Name).Msg("Dropping table")
+
+	return c.conn.Exec(ctx, queries.DropTable(table))
+}
 
-	have, want := current.GetChangedColumns(table)
-	if len(have) > 0 && c.spec.MigrateMode != specs.MigrateModeForced {
-		return fmt.Errorf("table %s has different types for columns %v but schema wants %v , add `migrate_mode: forced` to the destination spec to drop the columns", table.Name, have, want)
+func needsTableDrop(change schema.TableColumnChange) bool {
+	// We can safely add a nullable column without dropping the table
```
candiduslynx, line 107: See comment on the `canSafeMigrate` to transform to switch

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -12,67 +14,140 @@ import (
...
+		return false
+	}
+
+	return true
+}
+
+func changesSorter(a, b schema.TableColumnChange) bool {
+	aDropTable := needsTableDrop(a)
+	bDropTable := needsTableDrop(b)
+	return aDropTable && !bDropTable
```
candiduslynx, line 127: may skip internal variables, just use call results

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -12,67 +14,140 @@ import (
...
+}
+
+func (c *Client) autoMigrate(ctx context.Context, table *schema.Table, current *schema.Table) (err error) {
+	changes := table.GetChanges(current)
+	sort.SliceStable(changes, func(i, j int) bool {
```
candiduslynx, line 132: why do we want the sorted?

plugins/destination/clickhouse/queries/testdata/create_table.sql
```diff
@@ -1,9 +1,9 @@
...
-  `_cq_id` UUID,
-  `_cq_parent_id` Nullable(UUID),
-  `_cq_source_name` Nullable(String),
-  `_cq_sync_time` Nullable(DateTime64(9)),
-  `extra_col` Float64,
-  `extra_inet_col` Nullable(String),
-  `extra_inet_arr_col` Array(Nullable(String))
-) ENGINE = MergeTree ORDER BY (`_cq_id`)
\ No newline at end of file
+`_cq_id` UUID,
```
candiduslynx, line 2: could we leave the 2 spaces, though?
Less diff + easier query debug for those who look at the sdk logs

plugins/destination/clickhouse/queries/types.go
```diff
@@ -157,9 +157,8 @@ func normalizeColumn(column schema.Column) schema.Column {
 	}
 
 	res.CreationOptions = column.CreationOptions
-	if !res.CreationOptions.NotNull {
-		res.CreationOptions.NotNull = column.CreationOptions.PrimaryKey
-	}
+	// We only support append mode so we don't care about primary key changes to the schema
+	res.CreationOptions.PrimaryKey = false
```
candiduslynx, line 161: Does SDK make all PKs be not null?
In our case it would be beneficial to have this.

Now you're changing the behaviour, so I most certainly would mark this as a breaking change (consider the sync 1 where the col was a PK and was created as non-nullable, and then this version syncs and has to drop not null) (so, essentially, drops the data).

candiduslynx (changes requested):

plugins/destination/clickhouse/client/migrate.go
erezrokah: If a table has multiple changes, one for adding a column and another that requires a table drop+add, we should do the table drop+add first then exit early

plugins/destination/clickhouse/client/migrate.go
erezrokah: I actually find this easier to debug the code this way, but don't have a strong opinion

plugins/destination/clickhouse/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8255/commits/e28b6ba3187e8139537290393d372f8e6a2f2e54

plugins/destination/clickhouse/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8255/commits/04b2ff1a5a9c37308df778c10f46ae02b4b9effe

plugins/destination/clickhouse/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8255/commits/deb21b84f1b09ffa2ff78105209d3790e00bb6a3

plugins/destination/clickhouse/queries/testdata/create_table.sql
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8255/commits/6764c84e1bf8c5bb15145c7509cc4c8f7303f098

plugins/destination/clickhouse/client/migrate.go
erezrokah: This suggestion is not the same as the previous code. `TableColumnChangeTypeRemove` looks at the previous column and `TableColumnChangeTypeAdd` looks at the current column.

plugins/destination/clickhouse/client/migrate.go
erezrokah: Simplified in https://github.com/cloudquery/cloudquery/pull/8255/commits/6d1bdd9e45554de6f372aa0519e567350cbf5448

plugins/destination/clickhouse/queries/types.go
erezrokah: Great point! Re-added in https://github.com/cloudquery/cloudquery/pull/8255/commits/ba006d7c777bfc06f81a746522c449991393e481

candiduslynx (approved):

PR #8256, opened 2023-02-20
chore(file): Enable JSON:SkipMigrateAppend in testing

disq:
filetypes update ([PR](https://github.com/cloudquery/cloudquery/pull/8218)) fails otherwise.

### DIFF
--- a/plugins/destination/file/client/client_test.go
+++ b/plugins/destination/file/client/client_test.go
@@ -40,6 +40,7 @@ func TestPluginJSON(t *testing.T) {
 		destination.PluginTestSuiteTests{
 			SkipOverwrite:             true,
 			SkipDeleteStale:           true,
+			SkipMigrateAppend:         true,
 			SkipMigrateOverwrite:      true,
 			SkipMigrateOverwriteForce: true,
 			SkipMigrateAppendForce:    true,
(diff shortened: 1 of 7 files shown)


hermanschaaf (approved):

erezrokah (approved):
Looks good to unblock the PR, but let's update the title to reflect it does dep updates as the testing change is not relevant to users.

I think the "proper" fix should be to set a migrate strategy for the append tests, but I'll do that separately

candiduslynx (approved):

disq:
@erezrokah Renamed as `chore`. This only applies to the file plugin that's why it's `chore(file)` and not `fix(deps)`.

erezrokah:
> @erezrokah Renamed as `chore`. This only applies to the file plugin that's why it's `chore(file)` and not `fix(deps)`.

It updates dependencies which is a user facing change so it should be at least `fix`. For example this update includes https://github.com/cloudquery/filetypes/pull/71

disq:
Ah sorry it wasn't meant to update dependencies. Fixing

disq:
@erezrokah not updating dependencies any more, that will be done by the linked PR.

erezrokah (approved):

PR #8259, opened 2023-02-20
feat(deps): Update Elasticsearch plugin-sdk to v1.38.2

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/8156 as it requires an additional fix.
If the table doesn't have any PKs and in overwrite mode(s) we can only write a single resource per table as the doc ID will always be the same.
This PR changes it so if there are no PKs, we use all columns for the doc ID.

> Discovered by the `MigrateOverwrite` test


plugins/destination/elasticsearch/client/client_test.go
```diff
@@ -22,5 +31,8 @@ func TestPlugin(t *testing.T) {
 		destination.PluginTestSuiteTests{
 			SkipMigrateOverwriteForce: true,
 			SkipMigrateAppendForce:    true,
+
```
erezrokah, line 34: Once we add force migrate mode we can remove the `SkipMigrateOverwriteForce` and `SkipMigrateAppendForce`, but this is out of scope of this PR

plugins/destination/elasticsearch/client/write.go
```diff
@@ -62,6 +62,12 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resou
 
 func pkIndexes(table *schema.Table) []int {
 	pks := table.PrimaryKeys()
+	if len(pks) == 0 {
+		// if no PK is defined, use all columns for the ID which is based on the indices returned by this function
+		for _, col := range table.Columns {
+			pks = append(pks, col.Name)
+		}
```
candiduslynx, line 69:
```suggestion
		for _, col := range table.Columns {
			pks = table.Columns.Names()
		}
```

candiduslynx (changes requested):

plugins/destination/elasticsearch/client/write.go
erezrokah: Nice one 🚀

candiduslynx: 100%

candiduslynx (approved):
I'm approving but removing automerge bc the original author (@hermanschaaf) should take a look, too.

hermanschaaf (approved):
Nice catch!

PR #8262, opened 2023-02-20
feat(deps): Update File plugin-sdk to v1.38.2

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/8156.
For file destination we can only do force for every change (we might be able to do add and remove column safely for JSON as it's simply adding/removing a new field to the JSON object which should be fine. Out of scope for this PR).

### DIFF
--- a/plugins/destination/file/client/client_test.go
+++ b/plugins/destination/file/client/client_test.go
@@ -4,7 +4,16 @@ import (
 	"testing"
 
 	"github.com/cloudquery/plugin-sdk/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/specs"
 )
+
+var migrateStrategy = destination.MigrateStrategy{
+	AddColumn:           specs.MigrateModeForced,
+	AddColumnNotNull:    specs.MigrateModeForced,
+	RemoveColumn:        specs.MigrateModeForced,
+	RemoveColumnNotNull: specs.MigrateModeForced,
+	ChangeColumn:        specs.MigrateModeForced,
+}
 
 func TestPluginCSV(t *testing.T) {
 	destination.PluginTestSuiteRunner(t,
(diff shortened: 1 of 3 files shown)


disq (approved):

plugins/destination/file/client/client_test.go
```diff
@@ -62,10 +75,12 @@ func TestPluginParquet(t *testing.T) {
 			SkipOverwrite:             true,
 			SkipDeleteStale:           true,
 			SkipSecondAppend:          true,
-			SkipMigrateAppend:         true,
```
disq, line 65: Parquet format doesn't support any changes or appends once the file footer is written.

disq (changes requested):

plugins/destination/file/client/client_test.go
```diff
@@ -19,10 +28,12 @@ func TestPluginCSV(t *testing.T) {
 		destination.PluginTestSuiteTests{
 			SkipOverwrite:             true,
 			SkipDeleteStale:           true,
-			SkipMigrateAppend:         true,
 			SkipMigrateOverwrite:      true,
 			SkipMigrateOverwriteForce: true,
 			SkipMigrateAppendForce:    true,
```
erezrokah, line 33: We can remove `SkipMigrateOverwriteForce` and `SkipMigrateAppendForce` once we add support for forced mode, but out of scope for this PR

plugins/destination/file/client/client_test.go
erezrokah: OK so the tests API is a bit confusing but this will be skipped actually since the migrate strategy is forced:
https://github.com/cloudquery/plugin-sdk/blob/fac9ed3681d4f10800274a6bd35d2fa9cdcc732d/plugins/destination/plugin_testing_migrate.go#L91

Once we add support for forced mode (which will probably just delete the existing file), we'll need to remove `SkipMigrateAppendForce` and then it will verify the forced append work 🙃

disq (approved):

PR #8318, opened 2023-02-21
feat(deps)!: Update MSSQL plugin-sdk to v1.39.1

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/7819 for MSSQL.

This is a breaking change due to the type change from `datetimeoffset` to `datetime2` as we don't want to save time zone offset, see https://github.com/cloudquery/cloudquery/blob/ee4b2c01867b2140537ee4656d943d2e0cf2898b/plugins/destination/postgresql/client/types.go#L74

I suggest following the code in an IDE as it's hard to understand the changes from the diff.

BEGIN_COMMIT_OVERRIDE
feat(deps): Update MSSQL plugin-sdk to v1.38.2

BREAKING-CHANGE: Update MSSQL plugin-sdk to v1.38.2. You'll need to drop the database and start fresh due to a change in the schema for all timestamp columns from `datetimeoffset` to `datetime2`.
END_COMMIT_OVERRIDE


plugins/destination/mssql/client/columns.go
```diff
@@ -3,21 +3,19 @@ package client
...
 )
 
 func (c *Client) pkEnabled() bool {
 	return c.spec.WriteMode == specs.WriteModeOverwrite || c.spec.WriteMode == specs.WriteModeOverwriteDeleteStale
 }
 
-func (c *Client) getTableColumns(ctx context.Context, table *schema.Table) (queries.Definitions, error) {
-	query, params := queries.GetTableSchema(c.schemaName, table)
-	var tc queries.Definitions
+func (c *Client) getTableColumns(ctx context.Context, table *schema.Table) (schema.ColumnList, error) {
```
erezrokah, line 16: The changes here are related to me using `information_schema` to get the columns data instead of `sys.*`.
If you look at the changes to `plugins/destination/mssql/queries/table_info.go` you see the queries are shorter and easier to understand (in my opinion).
Also see https://www.mytecbits.com/microsoft/sql-server/information_schema-tables-vs-sys-tables

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+	var normalized schema.Tables
+	for _, table := range allTables {
+		tableCopy := table.Copy(table.Parent)
+		for i := range tableCopy.Columns {
+			// Since multiple schema types can map to the same MSSQL type we need to normalize them to avoid false positives when detecting schema changes
+			tableCopy.Columns[i].Type = queries.SchemaType(queries.SQLType(table.Columns[i].Type))
 		}
-	}()
+		// If there are no PKs, we use CqID as PK
+		pks := tableCopy.PrimaryKeys()
```
erezrokah, line 72: This logic was moved from `GetPKColumns`. We now do it at the beginning of the migration so no need to do it again

plugins/destination/mssql/client/table.go
```diff
@@ -1,20 +0,0 @@
...
-
-import (
-	"context"
-	"fmt"
-
-	"github.com/cloudquery/cloudquery/plugins/destination/mssql/queries"
-	"github.com/cloudquery/plugin-sdk/schema"
-)
-
-func (c *Client) tableExists(ctx context.Context, table *schema.Table) (bool, error) {
```
erezrokah, line 11: We don't need this as we get all tables in the database in a single query

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+				c.logger.Info().Str("table", table.Name).Msg("Table exists, force migration required")
+				if err := c.recreateTable(ctx, table); err != nil {
+					return err
+				}
+			}
 		}
 	}
 
-	return c.Migrate(ctx, table.Relations)
+	return nil
```
erezrokah, line 190: `normalizedTables` is a flat list of all tables so no need for the recursion

plugins/destination/mssql/queries/columns.go
```diff
@@ -16,28 +16,8 @@ func AddColumn(schemaName string, table *schema.Table, definition *Definition) s
 	})
 }
 
-func DropColumn(schemaName string, table *schema.Table, definition *Definition) string {
```
erezrokah, line 19: On forced migrate we only drop/add tables so no need for these anymore

plugins/destination/mssql/queries/columns_test.go
```diff
@@ -21,32 +21,3 @@ func TestAddColumn(t *testing.T) {
 
 	require.Equal(t, expected, query)
 }
-
-func TestDropColumn(t *testing.T) {
```
erezrokah, line 25: See https://github.com/cloudquery/cloudquery/pull/8318/files#r1112983724

plugins/destination/mssql/queries/definition.go
```diff
@@ -46,30 +46,11 @@ func (d *Definition) Nullable() *Definition {
...
 }
 
-func NewDefinition(name, typ string, nullable bool) *Definition {
-	d := &Definition{
-		Name:    strings.ToLower(name),
-		typ:     strings.ToLower(typ),
-		notNull: !nullable,
-	}
-	// add unique for _cq_id
-	d.unique = d.Name == schema.CqIDColumn.Name
```
erezrokah, line 56: I'm not sure why this is needed. As a general approach we should not rely on CqID existing, and only do whatever is  specified in `column.CreationOptions`

plugins/destination/mssql/queries/definition.go
```diff
@@ -46,30 +46,11 @@ func (d *Definition) Nullable() *Definition {
...
-		Name: column.Name,
-		typ:  SQLType(column.Type),
-	}
-
-	switch {
-	case column.Name == schema.CqIDColumn.Name:
-		// _cq_id column should always have a "UNIQUE NOT NULL" constraint
-		def.unique = true
-		def.notNull = true
-	case pkEnabled && column.CreationOptions.PrimaryKey:
```
erezrokah, line 71: I'll add this back for compatibility

plugins/destination/mssql/queries/definition.go
```diff
@@ -46,30 +46,11 @@ func (d *Definition) Nullable() *Definition {
 	}
 }
 
-func NewDefinition(name, typ string, nullable bool) *Definition {
```
erezrokah, line 49: Not used anymore due to the changes in `getTableColumns`

plugins/destination/mssql/queries/templates/drop_table.sql.tpl
```diff
@@ -0,0 +1 @@
+DROP TABLE {{.Table}};
```
erezrokah, line 1: We might need to re-consider some of these templates. I think adding a template each time you want to write a query is a bit much. Not sure what is the added value of having those instead of doing it in Go code directly

plugins/destination/mssql/queries/pk.go
```diff
@@ -20,7 +20,7 @@ func AddPK(schemaName string, table *schema.Table) string {
 	return execTemplate("pk_add.sql.tpl", &pkQueryBuilder{
 		Table:   sanitizeID(schemaName, table.Name),
 		Name:    pkConstraint(table),
-		Columns: GetPKColumns(table, true), // we call AddPK only for enabled
```
erezrokah, line 23: The logic that required passing `true` was moved to the code that normalized the tables before migration

plugins/destination/mssql/queries/tvp_test.go
```diff
@@ -69,7 +69,7 @@ func TestTVPAddType(t *testing.T) {
   [_cq_id] uniqueidentifier UNIQUE NOT NULL,
   [_cq_parent_id] uniqueidentifier,
   [_cq_source_name] nvarchar(4000),
-  [_cq_sync_time] datetimeoffset,
+  [_cq_sync_time] datetime2,
```
erezrokah, line 72: `datetimeoffset` saves the offset so if we write a `time.Time` and then read it the time zone information gets added and we get a different value than the original (see comment in PR description and https://github.com/cloudquery/cloudquery/blob/ee4b2c01867b2140537ee4656d943d2e0cf2898b/plugins/destination/postgresql/client/types.go#L74

plugins/destination/mssql/queries/definition.go
candiduslynx: we had `unique not null` in PostgreSQL previously. If we skip this now, that's fine

plugins/destination/mssql/queries/tvp_test.go
candiduslynx: Why don't we want the time zone?

plugins/destination/mssql/queries/table_info.go
```diff
@@ -2,64 +2,32 @@ package queries
...
+	return tableSchema, []any{sql.Named("tableName", table.Name), sql.Named("schemaName", schemaName)}
 }
 
-func TableExists(schemaName string, table *schema.Table) (query string, params []any) {
-	const tableExistsQuery = `SELECT COUNT(*)
-FROM sys.tables t
-    INNER JOIN sys.schemas s ON t.schema_id = s.schema_id
-WHERE s.[name] = @schemaName AND t.[name] = @tableName`
+func GetTablePK(schemaName string, databaseName string, table *schema.Table) (query string, params []any) {
+	pks := fmt.Sprintf(`SELECT Col.COLUMN_NAME from %s.INFORMATION_SCHEMA.TABLE_CONSTRAINTS Tab, %s.INFORMATION_SCHEMA.CONSTRAINT_COLUMN_USAGE Col
```
candiduslynx, line 24: see `sys` tables comments

plugins/destination/mssql/queries/table_info.go
```diff
@@ -2,64 +2,32 @@ package queries
...
+func AllTables(schemaName string) (query string, params []any) {
+	return `SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = @schemaName AND TABLE_TYPE = 'BASE TABLE';`, []any{sql.Named("schemaName", schemaName)}
+}
 
-	return sqlSelectPKColumns, []any{
-		sql.Named("tableName", table.Name),
-		sql.Named("schemaName", schemaName),
-	}
+func GetTableSchema(schemaName string, databaseName string, table *schema.Table) (query string, params []any) {
+	tableSchema := fmt.Sprintf(`SELECT COLUMN_NAME, DATA_TYPE, IS_NULLABLE, CHARACTER_MAXIMUM_LENGTH FROM %s.INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA = @schemaName AND TABLE_NAME = @tableName;`, databaseName)
```
candiduslynx, line 19: `%s.INFORMATION_SCHEMA.COLUMNS` should be replaces by `sanitizeID(databaseName, "INFORMATION_SCHEMA", "COLUMNS")` at the very least (imagine database named `select`)

plugins/destination/mssql/queries/tables_test.go
```diff
@@ -14,7 +14,7 @@ func TestCreateTable(t *testing.T) {
   [_cq_id] uniqueidentifier UNIQUE NOT NULL,
   [_cq_parent_id] uniqueidentifier,
   [_cq_source_name] nvarchar(4000),
-  [_cq_sync_time] datetimeoffset,
+  [_cq_sync_time] datetime2,
```
candiduslynx, line 17: requires https://github.com/cloudquery/plugin-sdk/issues/700

plugins/destination/mssql/queries/templates/drop_table.sql.tpl
candiduslynx: I'm OK with removing the templates altogether, as was done in CH dst (initially implemented with templates but then moved from them)

plugins/destination/mssql/client/tables.go
```diff
@@ -23,3 +23,28 @@ func (c *Client) createTable(ctx context.Context, table *schema.Table) (err erro
...
+}
+
+func (c *Client) recreateTable(ctx context.Context, table *schema.Table) (err error) {
+	err = c.dropTable(ctx, table)
+	if err != nil {
+		return err
+	}
+
+	return c.createTable(ctx, table)
+}
```
candiduslynx, line 50: we don't need named params here
```suggestion
func (c *Client) recreateTable(ctx context.Context, table *schema.Table) error {
	err := c.dropTable(ctx, table)
	if err != nil {
		return err
	}

	return c.createTable(ctx, table)
}
```

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+			c.logger.Info().Str("table", table.Name).Msg("Table with no columns, skipping")
+			continue
+		}
+		schemaTable := schemaTables.Get(table.Name)
+		if schemaTable == nil {
+			c.logger.Debug().Str("table", table.Name).Msg("Table doesn't exist, creating")
+			if err := c.createTable(ctx, table); err != nil {
+				return err
+			}
+		} else {
```
candiduslynx, line 174: I'd prefer `continue` statement in the `if` above, so that we don't have a lot of ident for `else`

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+			changes := table.GetChanges(schemaTable)
+			if c.canAutoMigrate(changes) {
+				c.logger.Info().Str("table", table.Name).Msg("Table exists, auto-migrating")
+				if err := c.autoMigrateTable(ctx, table, changes); err != nil {
+					return err
+				}
+			} else {
```
candiduslynx, line 181: `continue` statement in the `if` above, too

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+			result = append(result, t.Name)
+			tableChanges = append(tableChanges, changes)
+		}
+	}
+	return result, tableChanges
+}
+
+func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
+	for _, change := range changes {
+		if change.Type == schema.TableColumnChangeTypeAdd && (change.Current.CreationOptions.PrimaryKey || change.Current.CreationOptions.NotNull) {
```
candiduslynx, line 102: might be useful to extract a simple func
```go
func notNull(options schema.ColumnCreationOptions) bool {
	return options.NotNull || options.PrimaryKey
}
```
and use it here and in other places (normalization?)

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+
+		normalized = append(normalized, tableCopy)
+	}
+
+	return normalized
+}
 
-	if len(table.Columns) == 0 {
-		c.logger.Info().Str("table", table.Name).Msg("Table with no columns, skipping")
+func (c *Client) nonAutoMigrableTables(tables schema.Tables, schemaTables schema.Tables) ([]string, [][]schema.TableColumnChange) {
```
candiduslynx, line 83: use named return values here for the doc purposes `(names []string, changes [][]schema.TableColumnChange)`

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
 }
 
-func (c *Client) migrate(ctx context.Context, table *schema.Table) (err error) {
-	c.logger.Info().Str("table", table.Name).Msg("Migrating table started")
-	defer c.logger.Info().Str("table", table.Name).Msg("Migrating table done")
-	defer func() {
-		if err != nil {
-			c.logErr(err)
+func (c *Client) normalizedTables(tables schema.Tables) schema.Tables {
+	allTables := tables.FlattenTables()
```
candiduslynx, line 63: no need for extra var, nit

plugins/destination/mssql/client/migrate.go
```diff
@@ -2,48 +2,190 @@ package client
...
+	var normalized schema.Tables
+	for _, table := range allTables {
+		tableCopy := table.Copy(table.Parent)
```
candiduslynx, line 66: no need to do a copy as this is used in the migration only (won't affect the write?). See CH dst for this approach, too

plugins/destination/mssql/client/client.go
```diff
@@ -5,15 +5,17 @@ import (
...
 	mssql "github.com/microsoft/go-mssqldb"
 	"github.com/rs/zerolog"
 )
 
 type Client struct {
-	db         *sql.DB
-	schemaName string
+	db           *sql.DB
+	schemaName   string
+	databaseName string
```
candiduslynx, line 18: (applies to all such occasions).
Additionally, we don't need to specify the database name at all, as the conn string has to handle this itself.

For manual testing of queries you can use option to select db & schema.

candiduslynx (changes requested):

plugins/destination/mssql/queries/table_info.go
erezrokah: This specific query did not work without the database name. I can verify it again, but you can also test it yourself.
I'll sanitize it

plugins/destination/mssql/queries/types.go
```diff
@@ -20,7 +20,7 @@ func SQLType(t schema.ValueType) string {
 	case schema.TypeByteArray:
 		return "varbinary(max)"
 	case schema.TypeTimestamp:
-		return "datetimeoffset"
+		return "datetime2"
```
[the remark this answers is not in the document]
erezrokah: Related to https://github.com/cloudquery/plugin-sdk/issues/700

plugins/destination/mssql/queries/templates/drop_table.sql.tpl
erezrokah: Yeah, no need for it now. Maybe in a future refactor

plugins/destination/mssql/queries/table_info.go
erezrokah: OK so I tested this again and it works without the database name, so removed it in https://github.com/cloudquery/cloudquery/pull/8318/commits/95e6f22729951a878a81e0456b3b1f7831faedc2.

Not sure why it didn't work for me before

plugins/destination/mssql/client/migrate.go
erezrokah: Not sure that adds clarity as it's used only in this function so I don't see a reason to skip to another function to follow the logic.
Also the name `notNull` is not correct it should be something like `isOptionBreaking` or something like that. So a bit hard to explain out of the context of this function

plugins/destination/mssql/client/migrate.go
erezrokah: Actually if we don't copy it modifies the tables passed by the tests

plugins/destination/mssql/client/tables.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8318/commits/8a12efb7dd25a9ed54edde5666252f1b970d40a3

plugins/destination/mssql/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8318/commits/978a2215715f69c7766f31586e15262cbf08deeb

plugins/destination/mssql/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8318/commits/15175bd1d47fc03167b57e1c43acbfd1cc41fc1f

plugins/destination/mssql/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8318/commits/49a63d56af2c3c0206be397078e7cc963ac8442d

plugins/destination/mssql/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8318/commits/2c7a3a78928e8a7b80f5cf39ee4fcedd02553b88

plugins/destination/mssql/client/migrate.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/8318/commits/ba36505b15ae13a1e52bb28be4762b31cee43a35

candiduslynx (commented):
I'm OK with the changes, however, I think we need to address https://github.com/cloudquery/plugin-sdk/issues/700 prior to merging this

erezrokah:
@candiduslynx  small follow up in https://github.com/cloudquery/cloudquery/pull/8318/commits/b442896c312e5e01e38324ed2a82239500171b53 (still needs the UTC fix though)

plugins/destination/mssql/client/columns.go
```diff
@@ -38,7 +38,7 @@ func (c *Client) getTableColumns(ctx context.Context, table *schema.Table) (sche
 			if *charMaxLength == "-1" {
 				*charMaxLength = "max"
 			}
-			typ = "nvarchar(" + *charMaxLength + ")"
+			typ = fmt.Sprintf("%s(%s)", typ, *charMaxLength)
```
candiduslynx, line 41:
```suggestion
			typ += "(" + *charMaxLength + ")"
```

erezrokah:
@candiduslynx https://github.com/cloudquery/cloudquery/pull/8318/commits/99235be218d3d5aea43919c3c7416c16fd9ee987 is related to https://github.com/cloudquery/cloudquery/issues/8436

erezrokah:
@candiduslynx see https://github.com/cloudquery/plugin-sdk/pull/716. I think the impact of https://github.com/cloudquery/plugin-sdk/issues/700 is not that big

candiduslynx (approved):

PR #8326, opened 2023-02-21
fix(migrate): Make `_cq_id` unique

erezrokah:
#### Summary

Hot fix for Postgres. The real fix should be to add this option to the SDK, set it for the CQId column and use it in the destination

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -342,6 +342,9 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		}
 		columnName := pgx.Identifier{col.Name}.Sanitize()
 		fieldDef := columnName + " " + pgType
+		if col.Name == "_cq_id" {
+			fieldDef += " UNIQUE"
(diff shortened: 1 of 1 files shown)


candiduslynx (approved):

plugins/destination/postgresql/client/migrate.go
```diff
@@ -342,6 +342,9 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		}
 		columnName := pgx.Identifier{col.Name}.Sanitize()
 		fieldDef := columnName + " " + pgType
+		if col.Name == "_cq_id" {
```
candiduslynx, line 345: maybe we can use cq id col from schema pkg?
```suggestion
		if col.Name == schema.CqIdColumn.Name {
```

candiduslynx (approved):

plugins/destination/postgresql/client/migrate.go
erezrokah: Ah, shoot. I'll fix that in a bit

PR #8333, opened 2023-02-22
feat(aws): Add Support for CloudTrail Events With Incremental Table support

bbernays:


plugins/source/aws/resources/services/cloudtrail/events.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+func fetchCloudtrailEvents(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services().Cloudtrail
+
+	le := &cloudtrail.LookupEventsInput{}
+
+	const key = "cloudwatch_events"
```
hermanschaaf, line 49: Shouldn't this be `aws_cloudtrail_events`? (The table name)

hermanschaaf (changes requested):
Looks good, I just think the key name might be incorrect

plugins/source/aws/resources/services/cloudtrail/events.go
bbernays: So I agree it should definitely not be `cloudwatch` but should be `cloudtrail`... In the Stripe plugin it omits the plugin name `stripe_` so that is why I omitted the `aws_` from the key

https://github.com/cloudquery/cloudquery/blob/7cf277e837f88bea6622fadc9d7f682a5030d6d6/plugins/source/stripe/resources/services/topups/topups.go#L49-L63

hermanschaaf: Ok, it doesn't matter too much because internally we also use the source name to avoid naming collisions, but the backend interface is defined as `table`: https://github.com/cloudquery/plugin-sdk/blob/main/backend/backend.go#L7, and I think we should stick to that as far as possible. We can leave the Stripe plugin like it is, but let's use table name here

bbernays: Instead of defining a const, what if I were to just use `parent.Table.Name`?

hermanschaaf: It should be the current table's name, not the parent table. I would rather define a const and use it both here https://github.com/cloudquery/cloudquery/pull/8333/files#diff-3098e7cafab7f9f9a09ed33217c95a7a851f5ab3838299261d7e03ee325528deR17 and for the state key

bbernays: Done

hermanschaaf (approved):

kabachook:
Sync time goes brrrrrr

@bbernays @hermanschaaf @yevgenypats 
Hi, is that supposed to work only in incremental mode? 
I run sync every hour, every sync produces about 20k events in trail. Now I upgraded the plugin version, It started syncing the whole trail which has millions of events =)

hermanschaaf:
@kabachook The default behavior is to do a full sync of all the data every time. If you need the data in the `aws_cloudtrail_events` table but want it to be fetched incrementally, you will need to enable the local state backend as described here: https://www.cloudquery.io/docs/advanced-topics/managing-incremental-tables. Alternatively, if you don't need the data in the table, you can add it to the `skip_tables` section (or remove it from the `tables` section) in the source config.

If you have any follow-up questions, an issue would be a more appropriate place to discuss this, would you mind opening a new one or reaching out in Discord?

PR #8338, opened 2023-02-22
doc(website): Integration pages

hermanschaaf:


hermanschaaf:
Deployment preview: https://cloudquery-difsu2cw8-cloudquery.vercel.app/integrations

.github/workflows/deploy_website.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          node-version: 18.x
+      - name: Install Vercel CLI
+        run: npm install --global vercel@latest
+      - name: Pull Vercel Environment Information
+        run: vercel pull --yes --environment=production --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Build Project Artifacts
+        run: vercel build --prod --token=${{ secrets.VERCEL_TOKEN }}
+      - name: Deploy Project Artifacts to Vercel
+        run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
```
erezrokah, line 25:
```suggestion
        run: vercel deploy --prebuilt --prod --token=${{ secrets.VERCEL_TOKEN }}
```

Is `-prod` is the only difference from the preview workflow?
We can do a conditional step based on the branch name https://docs.github.com/en/actions/learn-github-actions/expressions#example-expression-in-an-if-conditional.

Also I'd also run on pushes to main and pull requests to main (not on every push to every branch), as that would increase our build usage significantly

website/components/pluginData.tsx
```diff
@@ -1,226 +1,283 @@
...
+    id: "snowflake",
+  },
+  {
+    name: "SQLite",
+    logo: "/images/logos/plugins/sqlite.svg",
+    id: "sqlite",
+  },
+];
+
+export const ALL_PLUGINS: Plugin[] = Array.prototype.concat(SOURCE_PLUGINS, DESTINATION_PLUGINS);
```
erezrokah, line 276:
```suggestion
export const ALL_PLUGINS: Plugin[] = [...SOURCE_PLUGINS, ...DESTINATION_PLUGINS];
```

Is a new shorter syntax for array concatenation

website/next.config.mjs
```diff
@@ -92,6 +112,8 @@ const withNextra = nextra({
 export default withNextra({
   reactStrictMode: true,
   experimental: {
+    workerThreads: false,
```
erezrokah, line 115: Is this needed with the large runners?

erezrokah (approved):
Looks great, nice usage of composition 🚀 

Nothing blocking on the code, but didn't test the preview yet

website/components/mdx/_download-macOS.mdx
```diff
@@ -5,12 +5,6 @@ import { Tabs, Tab } from '../Tabs'
 
 ```bash copy
 brew install cloudquery/tap/cloudquery
-
-# After initial install you can upgrade the version via:
-# brew upgrade cloudquery
-
-# To install a specific version use:
```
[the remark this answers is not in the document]
hermanschaaf: No, but I think this is a standard way to do it in homebrew, so we don't need to show it as part of the onboarding flow. Maybe I can add a page that documents it somewhere else though?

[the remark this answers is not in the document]
erezrokah: That works, as long as we have it documented I don't really care where (agree that putting it in the onboarding it a bit much)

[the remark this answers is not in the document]
hermanschaaf: Added in https://github.com/cloudquery/cloudquery/pull/8410

website/next.config.mjs
hermanschaaf: Ah no, we can remove this now, thanks!

.github/workflows/deploy_website.yml
hermanschaaf: Updated to run on pushes to main and pull request 👍  For the two file setup, I was mostly just following the recommendation here: https://vercel.com/guides/how-can-i-use-github-actions-with-vercel

All three steps have some small difference so I think it's probably okay to keep them separate, but if you want we can consolidate in a follow-up also

PR #8370, opened 2023-02-22
fix(migrate): Use unique column option instead of hard coding cq_id

erezrokah:
#### Summary

Needs https://github.com/cloudquery/plugin-sdk/pull/704 for it to work for with sources that don't use the latest SDK


yevgenypats (approved):

erezrokah:
Tested with https://github.com/cloudquery/plugin-sdk/pull/704 and this works

PR #8433, opened 2023-02-24
fix(aws)!: Missing partial PK for aws_ram_resource_share_permissions

disq:
Closes https://github.com/cloudquery/cloudquery-issues/issues/487


hermanschaaf:
I think this is a breaking change (and should be marked as such), because it's a new column that's being added to the PK.

hermanschaaf (changes requested):
Looks good, but since it needs to be a breaking change let's hold off on it for a bit to give us time to explore ways to make these transitions easier on users (https://github.com/cloudquery/cloudquery/issues/8474)

hermanschaaf (approved):

hermanschaaf:
Approving, because we now support [`deterministic_cq_id`](https://www.cloudquery.io/docs/reference/source-spec#deterministic_cq_id) and [`pk_mode`](https://www.cloudquery.io/docs/reference/destination-spec#pk_mode), which can be used to make these types of PK changes non-breaking. We will still mark it as breaking, but this gives users a nicer migration path if they want it.

PR #8434, opened 2023-02-24
fix(aws): EC2 Image Duplicate Records

bbernays:


plugins/source/aws/resources/services/ec2/images.go
```diff
@@ -33,3 +39,58 @@ func Images() *schema.Table {
...
+}
+
+func resolveImageArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	cl := meta.(*client.Client)
+	item := resource.Item.(types.Image)
+	a := arn.ARN{
+		Partition: cl.Partition,
+		Service:   "ec2",
+		Region:    cl.Region,
+		AccountID: aws.ToString(item.OwnerId),
```
bbernays, line 92: Changed this from 
```go
AccountID: cl.AccountID,
```

plugins/source/aws/resources/services/ec2/images.go
```diff
@@ -33,3 +39,58 @@ func Images() *schema.Table {
...
+		})
+		for pag.HasMorePages() {
+			resp, err := pag.NextPage(ctx)
+			if err != nil {
+				return err
+			}
+			for _, image := range resp.Images {
+				if aws.ToString(image.OwnerId) != c.AccountID {
+					res <- resp.Images
+				}
```
bbernays, line 76: Added filtering to exclude any Images that are owned by the same accountId

hermanschaaf: You mean "to exclude any images that are NOT owned by the same account ID", right?

hermanschaaf: Or no... maybe I'm not understanding this properly

hermanschaaf: Oh I see, first we list images we own, then images we can see (and filter out the ones we own)

hermanschaaf (approved):

plugins/source/aws/resources/services/ec2/images.go
bbernays: Exactly! There is no way to have multiple filters to get both the AMIs that the account owns as well as the AMIs that are shared with the account, so we have to have multiple calls

PR #8435, opened 2023-02-24
chore(website): Upgrade Nextra and re-enable Vercel

hermanschaaf:


erezrokah (approved):

erezrokah:
Nice one, we should also remove the large runner group

hermanschaaf:
@erezrokah Good point, thanks for the reminder. I'll do that as soon as this is merged

erezrokah:
@hermanschaaf I think we can ignore the wait for all failure as it is related to a bug in that script not handling removed workflow files

erezrokah:
I've also re-added `Vercel` to the branch protection rules

PR #8437, opened 2023-02-24
fix(aws)!: Missing partial/parent PK for `aws_ssm_instance_patches`

disq:
Also clean up SSM instances code structure


plugins/source/aws/resources/services/ssm/instance_compliance_items.go
```diff
@@ -28,11 +32,33 @@ func InstanceComplianceItems() *schema.Table {
 			{
 				Name:     "instance_arn",
 				Type:     schema.TypeString,
-				Resolver: resolveInstanceComplianceItemInstanceARN,
+				Resolver: schema.ParentColumnResolver("arn"),
```
disq, line 35: Change 2: Don't re-build ARN when the parent already has the same value

plugins/source/aws/resources/services/ssm/instance_patches.go
```diff
@@ -15,8 +18,16 @@ func InstancePatches() *schema.Table {
...
 		Multiplex:   client.ServiceAccountRegionMultiplexer("ssm"),
 		Columns: []schema.Column{
-			client.DefaultAccountIDColumn(true),
-			client.DefaultRegionColumn(true),
+			client.DefaultAccountIDColumn(false),
+			client.DefaultRegionColumn(false),
+			{
+				Name:     "instance_arn",
+				Type:     schema.TypeString,
+				Resolver: schema.ParentColumnResolver("arn"),
```
disq, line 26: Change 1: Add parent `arn` here as PK, drop account_id/region PKs

plugins/source/aws/resources/services/ssm/instance_compliance_items_fetch.go
```diff
@@ -1,24 +0,0 @@
...
-	"fmt"
-
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/aws/arn"
-	"github.com/aws/aws-sdk-go-v2/service/ssm/types"
-	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
-	"github.com/cloudquery/plugin-sdk/schema"
-)
-
-func resolveInstanceComplianceItemInstanceARN(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
disq, line 14: Part of change 2.

hermanschaaf (changes requested):
Looks good, but since it needs to be a breaking change let's hold off on it for a bit to give us time to explore ways to make these transitions easier on users (https://github.com/cloudquery/cloudquery/issues/8474)

Issue #6765, opened 2023-01-12
MySQL Destination


bbernays:
Add support for MySQL as a destination

PR #8438, opened 2023-02-24
feat: Add MySQL destination plugin

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/6765.

~~Still needs docs, cleanup and some testing on real source plugins, but the SDK tests are passing.~~


website/components/pluginData.tsx
```diff
@@ -246,6 +246,11 @@ export const DESTINATION_PLUGINS: Plugin[] = [
     logo: "/images/logos/plugins/mssql.svg",
     id: "mssql",
   },
+  {
+    name: "MySQL",
+    logo: "/images/logos/plugins/mysql.png",
```
erezrokah, line 251: See https://www.mysql.com/about/legal/logos.html for choice of logo

plugins/destination/mysql/client/schema.go
```diff
@@ -0,0 +1,145 @@
...
+		if column.CreationOptions.NotNull {
+			builder.WriteString(" NOT NULL")
+		}
+		if i < len(table.Columns)-1 {
+			builder.WriteString(",\n  ")
+		}
+	}
+	pks := table.PrimaryKeys()
+	if len(pks) > 0 {
+		builder.WriteString(",\n  ")
```
erezrokah, line 116: We don't set the constraint name to `table.Name + _cq_id` as for some tables it exceeded the 64 identifier limit.
Instead not setting a name defaults to `PRIMARY`

plugins/destination/mysql/Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: test
+test:
+	go test -race -timeout 3m ./...
```
candiduslynx, line 3: clean test cache as in other destination plugins

plugins/destination/mysql/client/client.go
```diff
@@ -0,0 +1,65 @@
...
+		dsn.Params = map[string]string{}
+	}
+	dsn.Params["parseTime"] = "true"
+	db, err := sql.Open("mysql", dsn.FormatDSN())
+	if err != nil {
+		return nil, fmt.Errorf("failed to open mysql connection: %w", err)
+	}
+
+	db.SetConnMaxLifetime(time.Minute * 3)
+	db.SetMaxOpenConns(10)
```
candiduslynx, line 53: maybe we can move this to env/spec?

plugins/destination/mysql/client/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+var migrateStrategy = destination.MigrateStrategy{
+	AddColumn:           specs.MigrateModeSafe,
+	AddColumnNotNull:    specs.MigrateModeForced,
+	RemoveColumn:        specs.MigrateModeSafe,
+	RemoveColumnNotNull: specs.MigrateModeForced,
+	ChangeColumn:        specs.MigrateModeForced,
+}
+
+func getConnectionString() string {
+	if testConn := os.Getenv("CQ_DEST_MY_SQL_TEST_CONNECTION_STRING"); len(testConn) > 0 {
```
candiduslynx, line 21:
```suggestion
	if testConn := os.Getenv("CQ_DEST_MYSQL_TEST_CONNECTION_STRING"); len(testConn) > 0 {
```

plugins/destination/mysql/client/migrate.go
```diff
@@ -0,0 +1,128 @@
...
+		if !c.pkEnabled() || len(pks) == 0 {
+			table.Columns.Get(schema.CqIDColumn.Name).CreationOptions.PrimaryKey = true
+		}
+
+		for i := range table.Columns {
+			if table.Columns[i].CreationOptions.PrimaryKey {
+				// Primary keys are implicitly set to not null in MySQL
+				table.Columns[i].CreationOptions.NotNull = true
+			}
+		}
```
candiduslynx, line 31:
```suggestion
		for _, col := range table.Columns {
			col.CreationOptions.NotNull = col.CreationOptions.NotNull || col.CreationOptions.PrimaryKey
		}
```

plugins/destination/mysql/client/migrate.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/specs"
+)
+
+func (c *Client) normalizedTables(tables schema.Tables) schema.Tables {
+	var normalized schema.Tables
+	for _, table := range tables.FlattenTables() {
+		table := table
```
candiduslynx, line 15: no need for this as no goroutines are used

plugins/destination/mysql/client/read.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (*Client) createResultsArray(table *schema.Table) []any {
+	results := make([]any, 0, len(table.Columns))
+	for _, col := range table.Columns {
+		switch col.Type {
+		case schema.TypeBool:
+			var r *bool
+			results = append(results, &r)
```
candiduslynx, line 18:
```suggestion
			results = append(results, new(*bool))
```

plugins/destination/mysql/client/read.go
```diff
@@ -0,0 +1,95 @@
...
+			builder.WriteString(", ")
+		}
+	}
+	builder.WriteString("FROM " + identifier(table.Name) + " WHERE _cq_source_name = ? ORDER BY _cq_sync_time ASC")
+	rows, err := c.db.QueryContext(ctx, builder.String(), sourceName)
+	if err != nil {
+		return err
+	}
+	for rows.Next() {
+		values := c.createResultsArray(table)
```
candiduslynx, line 87: doesn't it support the slice of pointers to `any`? as in [mssql](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/mssql/client/read.go#L35-L47)

plugins/destination/mysql/client/read.go
```diff
@@ -0,0 +1,95 @@
...
+		if err := rows.Scan(values...); err != nil {
+			return fmt.Errorf("failed to read from table %s: %w", table.Name, err)
+		}
+		res <- values
+	}
+	rows.Close()
```
candiduslynx, line 93: better do it via defer, as the rows scan err won't close rows otherwise

plugins/destination/mysql/client/schema.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func identifier(name string) string {
+	return fmt.Sprintf("`%s`", name)
```
candiduslynx, line 12:
```suggestion
	return "`" + name + "`"
```
for the sake of speed

plugins/destination/mysql/client/schema.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (c *Client) getTableColumns(ctx context.Context, table *schema.Table) (schema.ColumnList, error) {
+	query := `SELECT COLUMN_NAME, COLUMN_TYPE, IS_NULLABLE, CHARACTER_MAXIMUM_LENGTH, COLUMN_KEY FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = ?;`
+	var tc schema.ColumnList
+
+	rows, err := c.db.QueryContext(ctx, query, table.Name)
+	if err != nil {
+		return nil, err
+	}
```
candiduslynx, line 22:
```suggestion
	rows, err := c.db.QueryContext(ctx, query, table.Name)
	if err != nil {
		return nil, err
	}
	defer rows.Close()
```

plugins/destination/mysql/client/schema.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return tc, nil
+}
+
+func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema.Tables, error) {
+	query := `SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE';`
+	rows, err := c.db.QueryContext(ctx, query)
+	if err != nil {
+		return nil, err
+	}
```
candiduslynx, line 50:
```suggestion
	rows, err := c.db.QueryContext(ctx, query)
	if err != nil {
		return nil, err
	}
	defer rows.Close()
```

plugins/destination/mysql/client/transformer.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return v.Time
+}
+
+func (*Client) TransformJSON(v *schema.JSON) any {
+	if v.Status != schema.Present {
+		return nil
+	}
+
+	return string(v.Bytes)
```
candiduslynx, line 49: unescape logic check (see mssql implementation & required transformation + https://github.com/cloudquery/plugin-sdk/issues/622)

plugins/destination/mysql/client/transformer.go
```diff
@@ -0,0 +1,166 @@
...
+	if v.Status != schema.Present {
+		return nil
+	}
+
+	result := make([]string, len(v.Elements))
+	for i, elem := range v.Elements {
+		result[i] = elem.String()
+	}
+
+	return strings.Join(result, ",")
```
candiduslynx, line 72: You need just to use `v.String()` as return value, otherwise the default revere transformer would [fail](https://github.com/cloudquery/plugin-sdk/blob/main/schema/uuid_array.go#L62) on the `{` char absence

candiduslynx: Same goes for other array types except for bytes

plugins/destination/mysql/client/write.go
```diff
@@ -0,0 +1,77 @@
...
+			builder.WriteString(", ")
+		}
+	}
+	builder.WriteString(") VALUES (")
+	for i := range table.Columns {
+		builder.WriteString("?")
+		if i < len(table.Columns)-1 {
+			builder.WriteString(", ")
+		}
+	}
```
candiduslynx, line 28:
```suggestion
	builder.WriteString(strings.TrimSuffix(strings.Repeat("?,", len(table.Columns)), ","))
```

candiduslynx (changes requested):

plugins/destination/mysql/client/client.go
erezrokah: I would only add additional configuration based on users requests

plugins/destination/mysql/client/schema.go
erezrokah: Unless you have benchmarks that can show a significant improvement, I'd rather keep the current code for readability

plugins/destination/mysql/client/read.go
erezrokah: What's the added value of doing it this way? It won't be consistent with the rest of the code

plugins/destination/mysql/client/read.go
erezrokah: Tried this code and it doesn't work for all types (had to add special cases for bool, Inet, then stopped trying)

website/components/pluginData.tsx
hermanschaaf: They support a vector version (albeit in eps format, but we can convert it to svg), which would be nice to use. I'm happy to update it quickly?

hermanschaaf (approved):
Looks great!

plugins/destination/mysql/client/deletestale.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (c *Client) DeleteStale(ctx context.Context, tables schema.Tables, source string, syncTime time.Time) error {
```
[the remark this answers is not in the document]
erezrokah: Let me play with https://github.com/go-sql-driver/mysql#multistatements a bit to see if I can get it to work quickly

plugins/destination/mysql/client/transformer.go
erezrokah: Fixed in https://github.com/cloudquery/plugin-sdk/pull/714

candiduslynx (approved):

plugins/destination/mysql/client/deletestale.go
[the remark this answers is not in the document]
erezrokah: So we can't use placeholders `?` with `multiStatements`, see:
https://github.com/go-sql-driver/mysql#multistatements

so I'll need to format the parameters into the queries manually.

I'll keep this for now and we can re-visit batching if needed

PR #8441, opened 2023-02-26
fix(website): Update Gandi account links

erezrokah:
#### Summary

Existing links redirect to a new link now (which fails our broken links check).


yevgenypats (approved):

disq:
id.gandi.net is for logging in and account.gandi.net is for your account settings. The thing is the links don't work unless you're logged in.

erezrokah:
So something changed on their end as they do a temporary redirect to `id.gandi.net`, so no reason for the extra hop.
We can bring it back and exclude it from the check if that makes more sense

disq:
@erezrokah logging into `id.gandi.net` gets you to `https://admin.gandi.net/dashboard/<uuid>`, whereas `https://account.gandi.net/en` gets you to a special "Manage your user account & security settings" page, where you can click on "Authentication Options" (which gets you to `https://account.gandi.net/en/users/<your username>/security`) to be able to create/manage API keys (tucked away under the very bottom of the page, "Developer access" section)

So both links don't really work directly. With the old version is just one click away after login (provided login manages to get you back to the same page you visited in the first place)... from the dashboard (if you login via id.gandi) it's three more clicks (top-right your user name, click on "User settings", then move down and click on "Account and security" but I think it's still not extremely clear, just a little bit more convenient.

erezrokah:
Maybe we can link to their docs?

disq:
If only they had docs about how to create an api key. They have docs about API protocol (which includes sandbox access), [here](https://api.gandi.net/docs/) or generic info about them having an api, [here](https://docs.gandi.net/en/rest_api/index.html) but unfortunately none of these clearly document where to create an api key and so on.

erezrokah:
😮‍💨

PR #8442, opened 2023-02-26
fix: Vercel integration example broken link, history deprecation broken link

erezrokah:
#### Summary

More broken links fixes

### DIFF
--- a/website/pages/docs/plugins/sources/vercel/_configuration.mdx
+++ b/website/pages/docs/plugins/sources/vercel/_configuration.mdx
@@ -17,4 +17,4 @@ spec:
 ```
 
 The Vercel plugin supports incremental syncing. This means that only new data will be fetched from Vercel and loaded into your destination for supported tables (support depending on API endpoint). This is done by keeping track of the last `paginator` for each table and only fetching data that has been updated since then.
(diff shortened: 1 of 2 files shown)


website/pages/docs/plugins/sources/vercel/_configuration.mdx
```diff
@@ -17,4 +17,4 @@ spec:
 ```
 
 The Vercel plugin supports incremental syncing. This means that only new data will be fetched from Vercel and loaded into your destination for supported tables (support depending on API endpoint). This is done by keeping track of the last `paginator` for each table and only fetching data that has been updated since then.
-To enable this, `backend` option must be set in the spec (as shown in the [example](configuration#example)). This is documented in the [Managing Incremental Tables](/docs/advanced-topics/managing-incremental-tables) section.
+To enable this, `backend` option must be set in the spec (as shown in the [example](/docs/plugins/sources/vercel/configuration#example)). This is documented in the [Managing Incremental Tables](/docs/advanced-topics/managing-incremental-tables) section.
```
erezrokah, line 20: We need the full path from the root of the Website as this component is rendered both from the plugins page and the integrations page

hermanschaaf (approved):

PR #8473, opened 2023-02-27
fix(website): Use PostgreSQL instead of PostgresSQL in `cloud-asset-inventory-cloudquery-microsoft-power-bi.mdx`

eltociear:
#### Summary
PostgresSQL -> PostgreSQL


erezrokah (approved):
Thanks @eltociear, nice catch!

Issue #8436, opened 2023-02-24
feat: Improve invalid type error during migration


erezrokah:
When PostgresSQL destination migrates the DB, we read the current schema and try to convert the column types to types used by CloudQuery. If a user changes the column types for some reason (e.g. use another tool like Prisma to read and re-apply the schema), the migration fails with a hard to understand error:
https://github.com/cloudquery/cloudquery/blob/a186499219778f1b77b80f96cc50d53bf031ebdf/plugins/destination/postgresql/client/types.go#L149

We should:
- [x] Provide a better error. Example "failed to convert Postgres column type `<type>` to a supported CloudQuery type. Supported types are `<list-of-supported-types>`. Hint: Drop and add the table x to solve the issue."
- [x] Consider supporting more types (for example also `int`, etc.). Add a warning if the type is unexpected.
- [ ] If running on forced mode and we can't read the schema, we can drop and add the table automatically

erezrokah:
I'll skip the last one until someone requests it as I'd like to understand first the use case for changing the schema CloudQuery creates

PR #8475, opened 2023-02-27
fix(postgres): Better error message on unknown types, add missing float types

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8436.
It also adds missing types for `double precision`. The query we use returns it as `real` so the code failed when trying to read a table that has floats


plugins/destination/postgresql/client/types.go
```diff
@@ -97,89 +99,74 @@ func (*Client) SchemaTypeToCockroach(t schema.ValueType) string {
...
 	}
 }
 
-func (*Client) Pg10ToSchemaType(t string) schema.ValueType {
-	switch t {
-	case "boolean":
-		return schema.TypeBool
-	case "bigint", "integer":
-		return schema.TypeInt
-	case "double precision":
```
erezrokah, line 115: Our query that reads the current schema returns `TypeFloat` as PostgreSQL `real`, so this code always failed on a table that has floats columns. I added both `float` and `real` to the supported types

yevgenypats (approved):

PR #8478, opened 2023-02-27
doc(website): Add postgres source integration page

hermanschaaf:
Add a Postgres source integration page and tweaks wording on a couple of integration pages.


website/components/pluginData.tsx
```diff
@@ -5,6 +5,7 @@ export type Plugin = {
   description?: string;
   logo: string;
   logoDark?: string;
+  kind: "source" | "destination";
```
erezrokah, line 8: We do we use this property?

erezrokah (approved):

yevgenypats (approved):

website/components/pluginData.tsx
hermanschaaf: Removed ALL_PLUGINS and the extra field

Issue #8503, opened 2023-02-28
`aws_iam_policies` has escaped data in the `policy_version_list` column


candiduslynx:
Related to https://github.com/cloudquery/plugin-sdk/issues/622

erezrokah:
Thanks @candiduslynx, what are the steps to create the IAM policy that reproduces the issue?

candiduslynx:
Just sync the `aws_iam_policies` table and look for entries that have non-empty `policy_version_list`.

erezrokah:
Closing this as resolved. We discussed this internally and the document property of `policy_version_list` should be a JSON string (not an object), so this looks OK

PR #8522, opened 2023-02-28
fix(aws-resources): Use QueryUnescape instead of PathUnescape when decoding iam policy version list document

erezrokah:
#### Summary

Noticed this while working on https://github.com/cloudquery/cloudquery/issues/8503.
`PathUnescape` does not unescape `+` to ` `, while `QueryUnescape` does unescape it.
The AWS docs reference to `java.net.URLDecoder` which does unescape `+`, see:
https://docs.oracle.com/javase/7/docs/api/java/net/URLDecoder.html
image

This fix is unrelated to https://github.com/cloudquery/cloudquery/issues/8503. Another option is to not decode at all and keep the original data without transformations


disq (approved):

hermanschaaf (approved):

disq:
> Another option is to not decode at all and keep the original data without transformations

It's a hassle for some destinations to decode this to be able to parse/query it. URL functions often aren't included in database functionality or are missing important bits, so decoding it here is actually better.

PR #8556, opened 2023-02-28
feat(github): Add support for app-based authentication and filtering by repository

hermanschaaf:
- adds support for app-based authentication workflows, which are necessary for certain endpoints (like repo traffic stats)
- adds support for syncing only specified repositories

This implementation uses the appropriate app authentication for the given org, but if no org-specific authentication mechanism is available it falls back to either the personal access token or the first listed app auth.


erezrokah (approved):
Looks great, but can you link to the GitHub docs for the APIs that only work on GitHub apps?
I added repo traffic a while back and tested in with a personal access token

plugins/source/github/client/spec.go
```diff
@@ -1,6 +1,53 @@
...
+		}
+	}
+	if len(s.Orgs) == 0 && len(s.Repos) == 0 {
+		return fmt.Errorf("missing orgs or repos in configuration")
+	}
+	for _, repo := range s.Repos {
+		if err := validateRepo(repo); err != nil {
+			return err
+		}
+	}
```
bbernays, line 41: Do we want to validate that the repo is actually part of the Org?

plugins/source/github/client/client.go
```diff
@@ -68,52 +81,79 @@ func Configure(ctx context.Context, logger zerolog.Logger, s specs.Source, _ sou
...
-	logger.Info().Msg("Discovering organizations repositories")
-	orgRepositories, err := discoverRepositories(ctx, c, spec.Orgs)
+	c := &Client{
+		logger:      logger,
+		orgServices: ghServices,
+		orgs:        spec.Orgs,
+		repos:       spec.Repos,
+	}
+	c.
+		logger.Info().Msg("Discovering repositories")
```
bbernays, line 128:
```suggestion
	c.logger.Info().Msg("Discovering repositories")
```

plugins/source/github/client/multiplexer.go
```diff
@@ -16,7 +16,7 @@ func OrgMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
 func OrgRepositoryMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
 	var l = make([]schema.ClientMeta, 0)
 	client := meta.(*Client)
-	for _, o := range client.orgs {
+	for o := range client.orgRepositories {
```
candiduslynx, line 19:
```suggestion
	for o, repos := range client.orgRepositories {
		for _, repo := range repos {
			l = append(l, client.WithOrg(o).WithRepository(repo))
		}
	}
```

plugins/source/github/client/spec.go
```diff
@@ -1,6 +1,53 @@
...
+	for _, appAuth := range s.AppAuth {
+		if appAuth.Org == "" {
+			return fmt.Errorf("missing org in app auth configuration")
+		}
+		if appAuth.AppID != "" && appAuth.PrivateKeyPath == "" {
+			return fmt.Errorf("missing private key path in configuration")
+		}
+		if appAuth.AppID != "" && appAuth.InstallationID == "" {
+			return fmt.Errorf("missing installation id in configuration")
+		}
```
candiduslynx, line 32:
```suggestion
		if appAuth.AppID == "" {
			continue
		}
		if appAuth.PrivateKeyPath == "" {
			return fmt.Errorf("missing private key path in configuration")
		}
		if appAuth.InstallationID == "" {
			return fmt.Errorf("missing installation id in configuration")
		}
```

candiduslynx (approved):

plugins/source/github/client/spec.go
hermanschaaf: This would require one or more API calls, which will be done during initialization in any case (for multiplexing). I think some basic validation is fine here, we can't catch everything

erezrokah:
> It seems like they in fact mean that it works with both PATs and Apps? Not sure why our workflow failed with a Personal Access Token then though

I think it means both, see https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-organization-repositories
image

Maybe it was a permissions issue? I was using the old personal access token (not the new fined grained one)

hermanschaaf:
@erezrokah Yeah exactly, that's what I just realized as well. I was able to make it work just now using a classic PAT, but it depends on what your organization allows as well.

PR #8660, opened 2023-03-01
chore(ci-aws): Remove sanity sync test

erezrokah:
#### Summary

This sanity test was added in https://github.com/cloudquery/cloudquery/pull/5503 due to https://github.com/cloudquery/cloudquery-issues/issues/459.
However it has an increasing cost as we add more build minutes, and trying to catch SDK bugs using the AWS plugin is probably not the best approach.

> On the latest release PR this step too 27m and counting only to be killed by our build timeout, see https://github.com/cloudquery/cloudquery/actions/runs/4299344604/jobs/7494467760


yevgenypats (approved):

hermanschaaf:
Maybe we should just not sync all tables in this test, only one or two?

erezrokah:
> Maybe we should just not sync all tables in this test, only one or two?

I was thinking that too... Maybe just buckets? It will give us a bit of confidence. I can go either way

hermanschaaf:
@erezrokah Yeah I think this test is useful, but let's make it quick. I think any table will do, even if there are no resources.

bbernays:
Maybe also only sync a single region?

erezrokah:
Replaced by https://github.com/cloudquery/cloudquery/pull/8666

Issue #7916, opened 2023-02-10
AWS RDS Resources: Backtracks, Events, more


jsonpr:
- [ ] RDS BackTracks https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBClusterBacktracks.html
- [ ] Events: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeEvents.html
- [ ] Option Group (and Option Group Options): https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeOptionGroups.html
- [x] Reserved Instances: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeReservedDBInstances.html

Hokwang:
I need `reserved instances`.

bbernays:
@Hokwang - Thanks for the comment. I have opened a separate issue to track that request: #8261

hermanschaaf:
I've opened https://github.com/cloudquery/cloudquery/pull/8671 to add RDS Backtracks, Events and Option Groups

PR #8671, opened 2023-03-01
feat: Add RDS BackTracks, Events and Option Groups

hermanschaaf:
This adds three new tables:
 - RDS DB Cluster Backtracks (`aws_rds_cluster_backtracks`),
 - RDS Events (`aws_rds_events`), and 
 - RDS Option Groups (`aws_rds_option_groups`)

Closes [#7916](https://github.com/cloudquery/cloudquery/issues/7916)


yevgenypats (approved):

plugins/source/aws/resources/services/rds/events.go
```diff
@@ -0,0 +1,43 @@
...
+		Duration: &duration,
+	}
+	p := rds.NewDescribeEventsPaginator(svc, &config)
+	for p.HasMorePages() {
+		page, err := p.NextPage(ctx)
+		if err != nil {
+			return err
+		}
+		res <- page.Events
+	}
```
bbernays, line 41: Why not make this an incremental table?

hermanschaaf: Mostly because I wanted to keep it simple for the first iteration, and wasn't sure it was justified for the number of events you're likely to get for a single cluster (I think you're generally going to need one request anyway). We can always add it as an improvement later

Issue #8154, opened 2023-02-16
Make Plugin Docs searchable


bbernays:
### Is your feature request related to a problem? Please describe.

Right now if you google a CloudQuery table name you get results from other tools (pulumi, terraform) and even if you add cloudquery to the search Google still will not show results from our table documentation. The only way to get to the table docs is to navigate via source --> plugins --> <plugin> --> tables (edited) 

### Describe the solution you'd like

I don't really have a preference mainly because I don't fully understand the scope of the work but 2 ideas I had:
- Make Tables either on a [Github page](https://pages.github.com/)
- Make Tables part of our docs website

PR #8672, opened 2023-03-01
feat: Add tables search box

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8154.

The PR uses `getStaticProps` and `getStaticPaths` (the latter with an empty paths result) to load all tables on demand.
Once the table link is accessed it is generated and cached forever.

Based on https://github.com/shuding/nextra/blob/2f7f5516a66782ac3d6a02d427a34075504f4151/examples/swr-site/pages/remote/%5Bslug%5D.mdx?plain=1#L2.

Interesting bits are:
 -  `tables.js` to create the search data 
 - `Search.tsx` search component
 - `Makefile` changes to put the files in the right place and normalize the links to match the Website sturcture
 - `[table].mdx` files to generate the content on demand


website/pages/docs/plugins/sources/alicloud/tables.md
```diff
@@ -2,8 +2,8 @@
 
 ## Tables
 
-- [alicloud_bss_bill](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/docs/tables/alicloud_bss_bill.md)
-- [alicloud_bss_bill_overview](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/docs/tables/alicloud_bss_bill_overview.md)
-- [alicloud_ecs_instances](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/docs/tables/alicloud_ecs_instances.md)
-- [alicloud_oss_buckets](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/docs/tables/alicloud_oss_buckets.md)
-  - [alicloud_oss_bucket_stats](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/docs/tables/alicloud_oss_bucket_stats.md)
\ No newline at end of file
+- [alicloud_bss_bill](/docs/plugins/sources/alicloud/tables/alicloud_bss_bill)
```
candiduslynx, line 5: I wonder if it should be absolute path (should the path be prefixed with `/`)?

website/pages/docs/plugins/sources/gcp/tables/[table].mdx
```diff
@@ -0,0 +1,25 @@
+{/* <!-- vale off --> */}
```
Comment (candiduslynx) [inline website/pages/docs/plugins/sources/gcp/tables/[table].mdx:1]: I think that this code should somehow be extracted and called only with `gcp` param (as the table gen is repeated across all source plugins).

candiduslynx (changes requested):

yevgenypats:
Looks great! Do we want to do the same as we did with the docs and have the tables now only on our website, to save space and duplication workflows (we can do this also in follow-up)?

yevgenypats (approved):

erezrokah:
The `Summarize changes` workflow will be fixed posted merge (as it runs on `pull_request_target). See fix in https://github.com/cloudquery/cloudquery/pull/8672/files#diff-79fc9a700730d97b802a03cdd894962c175ff101b1f836c486eb66a9d746ec47

website/pages/docs/plugins/sources/alicloud/tables.md
erezrokah: Good point, I made those relative

website/pages/docs/plugins/sources/gcp/tables/[table].mdx
Comment (erezrokah) [reply website/pages/docs/plugins/sources/gcp/tables/[table].mdx:1]: I think https://github.com/cloudquery/cloudquery/pull/8672/commits/e63ff01d45778cf8287ba91c04e4563fbe3e3ab4 is the best we can do

erezrokah:
> Looks great! Do we want to do the same as we did with the docs and have the tables now only on our website, to save space and duplication workflows (we can do this also in follow-up)?

Great point, I removed the tables from GitHub and changed the GitHub `README.md` files to point to our Website

PR #8675, opened 2023-03-02
feat: Add AWS plugin support for EC2 Launch Templates

dpowley:
#### Summary

This PR adds support for [EC2 Launch Templates](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplate.html) and [Launch Template Versions](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateVersion.html) in the AWS source plugin. This implementation only collects the default launch template version (and not all versions). Could easily be tweaked to list all versions if needed.


plugins/source/aws/resources/services/ec2/launch_template_versions_fetch.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func fetchEc2LaunchTemplateVersions(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	var config ec2.DescribeLaunchTemplateVersionsInput
```
hermanschaaf: I think this was missing a config parameter to specify the launch template ID from the parent, so I pushed up a fix - let me know if you think that wasn't right

hermanschaaf (approved):
Looks great, thanks for the contribution @dpowley !  🚀

PR #8692, opened 2023-03-05
fix(website): Set plugins menu route to '#'

erezrokah:
#### Summary

Instead of https://github.com/cloudquery/cloudquery/pull/8670. This has the side effect of the top menu `plugins` to never be active (even after clicked), but I think this is still better that existing state.

What's we're doing is not strictly supported, see issue https://github.com/shuding/nextra/issues/1096


hermanschaaf (approved):
I'm not sure if it's better honestly, but happy to do it this way until we can fix the underlying bug :)

erezrokah:
> I'm not sure if it's better honestly, but happy to do it this way until we can fix the underlying bug :)

I'll try to take a look this week. I could not reproduce it on Nextra example docs site 😕

Issue #8369, opened 2023-02-22
Azure Resource - Monitor Subscription Diagnostic Settings


RyPeck:
Similar to the `azure_monitor_diagnostic_settings` table - requesting a new table `azure_monitor_subscription_diagnostic_settings`. Following the referenced resources in this table will allow look at the log settings.

https://learn.microsoft.com/en-us/rest/api/monitor/subscription-diagnostic-settings/list?tabs=HTTP

RyPeck:
https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/migrate-to-azure-storage-lifecycle-policy
`azure_monitor_diagnostic_settings` looks scheduled for deprecation.

PR #8693, opened 2023-03-05
feat(azure-resources): Add Subscription Diagnostic Settings

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8369

The Go SDK does not have a dedicated client/method to get the subscription diagnostic settings, but using the diagnostic settings client and passing a subscription Id works (see request format in https://learn.microsoft.com/en-us/rest/api/monitor/subscription-diagnostic-settings/list?tabs=HTTP)


plugins/source/azure/resources/services/monitor/subscription_diagnostic_settings.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func SubscriptionDiagnosticSettings() *schema.Table {
+	return &schema.Table{
+		Name:        "azure_monitor_subscription_diagnostic_settings",
+		Resolver:    fetchSubscriptionDiagnosticSettings,
+		Description: "https://learn.microsoft.com/en-us/rest/api/monitor/subscription-diagnostic-settings/list?tabs=HTTP#subscriptiondiagnosticsettingsresource",
+		Multiplex:   client.SubscriptionMultiplexRegisteredNamespace("azure_monitor_subscription_diagnostic_settings", client.Namespacemicrosoft_insights),
+		Transform:   transformers.TransformWithStruct(&armmonitor.DiagnosticSettingsResource{}, transformers.WithPrimaryKeys("ID")),
+		Columns:     schema.ColumnList{client.SubscriptionID},
```
disq, line 19: if `id` is the subscription id already, what's the point of adding `subscription_id`? `azure_subscription_subscriptions` is in a similar situation but it doesn't have the id duplication, so we can also give it a miss here?

erezrokah: Let me verify but I think the ID is of the diagnostic setting and not the subscription

erezrokah: Screenshot 2023-03-06 at 11 30 51

Yes, the ID is of the diagnostic setting

disq (approved):

PR #8695, opened 2023-03-05
feat(website): Add tables to sitemap

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/8672
Docs in https://github.com/iamvishnusankar/next-sitemap#additional-paths-function.
Since the tables are generated on demand we need to add them to the `sitemap.xml` explicitly.


website/utils/tables.ts
```diff
@@ -21,18 +21,4 @@ export const getStaticPaths = () => {
...
-    return {
-        plugin,
-        table,
-    }
-  })
-
-  tablesData.sort((a, b) => a.table.localeCompare(b.table));
-  return tablesData
-};
+export const getTablesData = getTablesDataJS;
```
erezrokah, line 24: `next-sitemap.config.js` does not support TypeScript, so moved this code to JS for re-use

candiduslynx (approved):

PR #8726, opened 2023-03-06
feat(aws-resources): Add IAM Access Advisor

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/1402
Fixes https://github.com/cloudquery/cloudquery/issues/1389
Fixes https://github.com/cloudquery/cloudquery/issues/1388
Fixes https://github.com/cloudquery/cloudquery/issues/1387
Fixes https://github.com/cloudquery/cloudquery/issues/1386

This PR implements https://docs.aws.amazon.com/cli/latest/reference/iam/get-service-last-accessed-details.html, but not https://docs.aws.amazon.com/cli/latest/reference/iam/get-service-last-accessed-details-with-entities.html as getting the entities is what makes this super slow.
I put the table definitions and fetchers in the same file as I think it's easier to follow the code that way, but happy to split into multiple files (one per table).

This follows a pattern of using relations to get maximum concurrency based on the spec setting.

Explanation about PKs:
- ~~We use an Arn to get a job, so 1-1 mapping. Hence we can use the Arn as the PK for the jobs table~~
- We use a job ID to get a list of `ServicesLastAccessed`. If you look at https://docs.aws.amazon.com/cli/latest/reference/iam/get-service-last-accessed-details-with-entities.html you'll see you need to pass a job ID and a service namespace to get the entities. Hence we should be able to use Arn (1-1 mapping to a job ID) and service namespace for uniqueness of  each `ServicesLastAccessed`.

Syncing these tables takes `2m21s` on our playground account

BEGIN_COMMIT_OVERRIDE
feat(aws-resources): Add IAM Access Advisor tables: `aws_iam_group_last_accessed_details`, `aws_iam_policy_last_accessed_details`, `aws_iam_role_last_accessed_details` and `aws_iam_user_last_accessed_details`. These might be slow to sync on some accounts. You can skip them if needed via `skip_tables: ["aws_iam_*_last_accessed_details"]`
END_COMMIT_OVERRIDE


bbernays:
How much concurrency does AWS support?

plugins/source/aws/resources/services/iam/last_accessed_jobs.go
```diff
@@ -0,0 +1,108 @@
...
+			client.DefaultAccountIDColumn(true),
+		},
+		Relations: []*schema.Table{userLastAccessedDetails()},
+	}
+}
+
+func roleLastAccessedJobs() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_iam_role_last_accessed_jobs",
+		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html#API_GenerateServiceLastAccessedDetails_Example_1_Response`,
```
hermanschaaf, line 35:
```suggestion
		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html`,
```

I guess we don't want to link to the example response specifically?

plugins/source/aws/resources/services/iam/last_accessed_jobs.go
```diff
@@ -0,0 +1,108 @@
...
+			client.DefaultAccountIDColumn(true),
+		},
+		Relations: []*schema.Table{roleLastAccessedDetails()},
+	}
+}
+
+func groupLastAccessedJobs() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_iam_group_last_accessed_jobs",
+		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html#API_GenerateServiceLastAccessedDetails_Example_1_Response`,
```
hermanschaaf, line 49:
```suggestion
		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html`,
```

plugins/source/aws/resources/services/iam/last_accessed_jobs.go
```diff
@@ -0,0 +1,108 @@
...
+			client.DefaultAccountIDColumn(true),
+		},
+		Relations: []*schema.Table{groupLastAccessedDetails()},
+	}
+}
+
+func policyLastAccessedJobs() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_iam_policy_last_accessed_jobs",
+		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html#API_GenerateServiceLastAccessedDetails_Example_1_Response`,
```
hermanschaaf, line 63:
```suggestion
		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html`,
```

plugins/source/aws/resources/services/iam/last_accessed_jobs.go
```diff
@@ -0,0 +1,108 @@
...
+
+type Job struct {
+	Arn   string `json:"arn"`
+	JobId string `json:"job_id"`
+}
+
+func userLastAccessedJobs() *schema.Table {
+	return &schema.Table{
+		Name:        "aws_iam_user_last_accessed_jobs",
+		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html#API_GenerateServiceLastAccessedDetails_Example_1_Response`,
```
hermanschaaf, line 21:
```suggestion
		Description: `https://docs.aws.amazon.com/IAM/latest/APIReference/API_GenerateServiceLastAccessedDetails.html`,
```

plugins/source/aws/resources/services/iam/last_accessed_jobs.go
erezrokah: Good point I missed that doc

erezrokah:
> How much concurrency does AWS support?

@bbernays can you explain your comment? Do you mean API limits? Concurrent report jobs limit?

I couldn't find anything about this in the AWS docs

bbernays:
> > How much concurrency does AWS support?
> 
> @bbernays can you explain your comment? Do you mean API limits? Concurrent report jobs limit?
> 
> I couldn't find anything about this in the AWS docs

Yeah I mean about concurrent reports jobs limit. I know the original PR had all of the jobs being created sequentially, so I wonder if there was a technical reason for that

erezrokah:
> Yeah I mean about concurrent reports jobs limit. I know the original PR had all of the jobs being created sequentially, so I wonder if there was a technical reason for that

I did not hit any limits on our playground account 😄 nor could find any docs about it

hermanschaaf:
I'd like to check my understanding of this code: 
 1. for every `group` / `policy` / `role`/ `user`, a `last_accessed_jobs` resolver will get called, because of the parent-child relationship with the corresponding `iam` table
 2. this child resolver creates a new job and writes the ARN and job ID to the corresponding `last_accessed_jobs` table. Then,
 3. a `last_accessed_details` resolver kicks off (again because of parent-child relationship), sleeping until the job either fails or succeeds. It then writes the details to the `last_accessed_details` table.

Is my understanding correct? This is a clever solution! I think it subtly changes the semantics of our tables though - especially the `last_accessed_jobs` table. Usually, a table is expected to fetch all available data for the particular resource. In the case of `last_accessed_jobs` though, it's more or less being employed as a queue--users won't get a full list of all "last accessed jobs" that have ever been created on their accounts, they will only get the last accessed jobs that have been generated by this particular sync. I think that's maybe not ideal? Is it necessary to have the `jobs` table at all? Why not have only a `details` table, where we both generate the job and wait for its results?

erezrokah:
> Is my understanding correct? This is a clever solution! I think it subtly changes the semantics of our tables though - especially the `last_accessed_jobs` table. Usually, a table is expected to fetch all available data for the particular resource. In the case of `last_accessed_jobs` though, it's more or less being employed as a queue--users won't get a full list of all "last accessed jobs" that have ever been created on their accounts, they will only get the last accessed jobs that have been generated by this particular sync. I think that's maybe not ideal? Is it necessary to have the `jobs` table at all? Why not have only a `details` table, where we both generate the job and wait for its results?

You correct! And you have a very good point. I think we can drop the jobs table as there's a 1-1 mapping between job and details. There will be a slight behavioral change (depending on the scheduler) as the details tables will take a bit longer and also get better concurrency.
The idea was to have as little API calls as possible per table, so the SDK can manage the concurrency of the resolvers.
Another thing to note is that a job ID is only valid within the same session:
image
so I didn't see a point in saving all of them.

TDLR: I'll drop the jobs tables

erezrokah:
@hermanschaaf cleaned up the job tables in https://github.com/cloudquery/cloudquery/pull/8726/commits/d3ed6b959696e98ecc22a93640ef3e0640b6fd7a

PR #8732, opened 2023-03-06
chore: Bring back plugins readme files

erezrokah:
#### Summary

Apparently some users are parsing these files to get the list of tables


hermanschaaf (approved):

hermanschaaf:
To reduce duplication, maybe we can copy the files to the website directory as part of the prebuild step?

erezrokah:
> To reduce duplication, maybe we can copy the files to the website directory as part of the prebuild step?

I'm not sure I follow this. Do you mean copy the `README.md` under `plugins/source/alicloud/docs/tables/README.md` to the Website in the prebuild step? If so we'll need to update our Vercel git ignore command to account for that (check if that file changes). Not sure what's better actually

PR #8728, opened 2023-03-06
Google Analytics source plugin

candiduslynx:
Implements #6779

PR #8734, opened 2023-03-06
chore(release): Sort release-please files

candiduslynx:
Extracted from #8728


erezrokah (changes requested):
Can you explain what problem this PR solves?

`.release-please-manifest.json` is auto generated by our bot so we should not edit it (we only due it due to https://github.com/googleapis/release-please/issues/1502 in specific cases).

The order of keys in `release-please-config.json` is not important and we should not assume/enforce it is sorted. The only value I see here is maybe avoiding git conflicts in `release-please-config.json` when different people add entries to it, and they both add it to the end, but I don't think we run into it often (or at all). And even if we do I think it's easier to resolve the conflict instead of always ensuring you add an entry in the right place (and that doesn't guarantee you'll be conflict free either).

candiduslynx:
@erezrokah 
> avoiding git conflicts in `release-please-config.json` when different people add entries to it

That's the precise reason, as several times when I was developing new plugin I got conflicts. Every time.
So, it can be at least OK to sort this now and ease the pain a bit for the next person.

As for the `.release-please-manifest.json` - I just sorted this, too, to keep in line with the config.

erezrokah:
> As for the `.release-please-manifest.json` - I just sorted this, too, to keep in line with the config.

If this doesn't have any impact can we undo this change? I don't see a reason to sort an auto generated file.

> That's the precise reason, as several times when I was developing new plugin I got conflicts. Every time.

This has a tradeoff of creating the impression of having to insert plugins in a sorted way to the config, which is unnecessary. Also it does not prevent conflicts when plugins are adjacent, so I'm not sure this is a better tradeoff than handling a conflict.
TDLR: the options are:
1. Handle conflicts once in a while which takes a second most of the time. This mostly happens if you append to the config which is not necessary. 
2. Insert items mostly in a sorted way + sort this file once in a while + still handle conflicts sometimes

Personally I think the second option is more time consuming, but happy to re-visit if other prefer the second option

candiduslynx:
> If this doesn't have any impact can we undo this change? I don't see a reason to sort an auto generated file.
It will still be sorted even being auto generated, as the update is made in-place

candiduslynx:
> it does not prevent conflicts when plugins are adjacent

The chances are really slim compared to adding the plugins at the end of the list, so I'd prefer the sorted list anyway

erezrokah:
Maybe we can add a workflow that sorts this file on a daily basis so we don't have to keep doing this manual work

erezrokah:
> > If this doesn't have any impact can we undo this change? I don't see a reason to sort an auto generated file.
> 
> It will still be sorted even being auto generated, as the update is made in-place

Can you explain what value we get from sorting the auto generated file? We already have the filler entries to prevent conflicts

candiduslynx:
> Can you explain what value we get from sorting the auto generated file? We already have the filler entries to prevent conflicts
Just to have this in sync. Can be safely reversed, if you insist.

erezrokah (approved):

PR #8735, opened 2023-03-06
fix(deps): Update ReadFile function from io/ioutil to os

jackgris:
Deprecated: As of Go 1.16, the same functionality is now provided by package os


erezrokah (approved):
Thanks @jackgris 🚀

PR #8758, opened 2023-03-07
feat: DuckDB destination

hermanschaaf:


.github/workflows/dest_duckdb.yml
```diff
@@ -0,0 +1,83 @@
...
+      - ".github/workflows/dest_duckdb.yml"
+  push:
+    branches:
+      - main
+    paths:
+      - "plugins/destination/duckdb/**"
+      - ".github/workflows/dest_duckdb.yml"
+
+jobs:
+  plugins-destination-duckdb:
```
candiduslynx, line 16: Do we have an image for duckdb to run against?

hermanschaaf: Oh yeah, I haven't updated the workflow yet, thanks

plugins/destination/duckdb/.goreleaser.yaml
```diff
@@ -0,0 +1,51 @@
...
+includes:
+  - from_file:
+      # Relative to the directory Go Releaser is run from (which is the root of the repository)
+      path: ./plugins/.goreleaser.template.yaml
+      
+builds:
+  - flags:
+      - -buildmode=exe
+    env:
+      - CGO_ENABLED=1
```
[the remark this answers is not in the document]
hermanschaaf: Not sure what you're asking? Similar to SQLite, the go-duckdb driver requires CGO

.github/workflows/dest_duckdb.yml
hermanschaaf: Actually, what am I saying 🤦 We don't need an image, because the tests run against an in-memory database

plugins/destination/duckdb/.goreleaser.yaml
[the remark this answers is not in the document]
candiduslynx: Yes, precisely that. (cgo produces non-static, if I'm not mistaken).
How do we distribute those, as dynamic executables?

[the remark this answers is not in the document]
hermanschaaf: Though it seems like maybe it won't work on Windows due to a bug: https://github.com/marcboeker/go-duckdb/issues/51

Guess we'll have to see and/or accept that as a limitation right now

plugins/destination/duckdb/client/migrate.go
```diff
@@ -0,0 +1,271 @@
...
+			// In DuckDB, a PK column must be NOT NULL, so we need to make sure that the schema we're comparing to has the same
+			// constraint.
+			if !c.enabledPks() {
+				tableCopy.Columns[i].CreationOptions.PrimaryKey = false
+			} else if tableCopy.Columns[i].CreationOptions.PrimaryKey {
+				tableCopy.Columns[i].CreationOptions.NotNull = true
+			}
+			// Since multiple schema types can map to the same duckdb type we need to normalize them to avoid false positives when detecting schema changes
+			tableCopy.Columns[i].Type = c.duckdbTypeToSchema(c.SchemaTypeToDuckDB(table.Columns[i].Type))
+		}
```
candiduslynx, line 74:
```suggestion
		for i, col := range tableCopy.Columns {
			// In DuckDB, a PK column must be NOT NULL, so we need to make sure that the schema we're comparing to has the same
			// constraint.
			switch {
				case !c.enabledPks():
					tableCopy.Columns[i].CreationOptions.PrimaryKey = false
				case col.CreationOptions.PrimaryKey:
					tableCopy.Columns[i].CreationOptions.NotNull = true
			}
			// Since multiple schema types can map to the same duckdb type we need to normalize them to avoid false positives when detecting schema changes
			tableCopy.Columns[i].Type = c.duckdbTypeToSchema(c.SchemaTypeToDuckDB(col.Type))
		}
```

plugins/destination/duckdb/client/migrate.go
```diff
@@ -0,0 +1,271 @@
...
+		return err
+	}
+
+	normalizedTables := c.normalizeColumns(tables)
+	for _, table := range normalizedTables {
+		c.logger.Warn().Msg("normalized table: " + table.Name + " columns: " + fmt.Sprint(table.Columns))
+	}
+	for _, table := range duckdbTables {
+		c.logger.Warn().Msg("duckdb     table: " + table.Name + " columns: " + fmt.Sprint(table.Columns))
+	}
```
candiduslynx, line 140:
```suggestion
```

plugins/destination/duckdb/client/migrate.go
```diff
@@ -0,0 +1,271 @@
...
+				if err := c.autoMigrateTable(table, changes); err != nil {
+					return err
+				}
+			} else {
+				c.logger.Info().Str("table", table.Name).Msg("Table exists, force migration required")
+				if err := c.recreateTable(table); err != nil {
+					return err
+				}
+			}
+		}
```
candiduslynx, line 173:
```suggestion
		if duckdb == nil {
			c.logger.Debug().Str("table", table.Name).Msg("Table doesn't exist, creating")
			if err := c.createTableIfNotExist(table); err != nil {
				return err
			}
			continue
		}

		changes := table.GetChanges(duckdb)
		if c.canAutoMigrate(changes) {
			c.logger.Info().Str("table", table.Name).Msg("Table exists, auto-migrating")
			if err := c.autoMigrateTable(table, changes); err != nil {
				return err
			}
			continue
		}

		c.logger.Info().Str("table", table.Name).Msg("Table exists, force migration required")
		if err := c.recreateTable(table); err != nil {
			return err
		}
```

plugins/destination/duckdb/client/read.go
```diff
@@ -0,0 +1,106 @@
...
+const (
+	readSQL = `SELECT %s FROM "%s" WHERE _cq_source_name = $1 order by _cq_sync_time asc`
+)
+
+func (*Client) createResultsArray(table *schema.Table) []any {
+	results := make([]any, 0, len(table.Columns))
+	for _, col := range table.Columns {
+		switch col.Type {
+		case schema.TypeBool:
+			var r *bool
```
candiduslynx, line 22: 1. group the options with the same `r` together
2. may be a bit simplified by `append(results, new(*bool))` code

plugins/destination/duckdb/client/read.go
```diff
@@ -0,0 +1,106 @@
...
+				values[i] = v.Get()
+			case *duckdb.Composite[[][]byte]:
+				values[i] = v.Get()
+			case *duckdb.Composite[[]int]:
+				values[i] = v.Get()
+			}
+		}
+		res <- values
+	}
+	rows.Close()
```
candiduslynx, line 104: use defer, as there's a path to return err on scan without closing rows

plugins/destination/duckdb/client/types.go
```diff
@@ -0,0 +1,102 @@
...
+	case schema.TypeUUIDArray:
+		return "uuid[]"
+	case schema.TypeCIDR:
+		return "text"
+	case schema.TypeCIDRArray:
+		return "text[]"
+	case schema.TypeMacAddr:
+		return "text"
+	case schema.TypeMacAddrArray:
+		return "text[]"
```
candiduslynx, line 38: group cases together (nit)

plugins/destination/duckdb/client/write.go
```diff
@@ -0,0 +1,111 @@
...
+	for i, c := range columns {
+		sb.WriteString(`"` + c.Name + `"`)
+		if i < columnsLen-1 {
+			sb.WriteString(",")
+		} else {
+			sb.WriteString(") values (")
+		}
+	}
+	counter := 0
+	for i := range columns {
```
candiduslynx, line 73:
```suggestion
	for i, col := range columns {
```

website/pages/docs/plugins/destinations/duckdb/overview.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+<Configuration />
+
+## DuckDB Spec
+
+This is the top level spec used by the DuckDB destination Plugin.
+
+- `connection_string` (string) (required)
+
+Absolute or relative path to a file, such as `./example.duckdb`
```
candiduslynx, line 24: ???

candiduslynx (changes requested):

website/pages/docs/plugins/destinations/duckdb/overview.mdx
hermanschaaf: I don't know what you are asking about here

candiduslynx: Is this line necessary or some leftover?

hermanschaaf: It's referring to the `connection_string` parameter; it is necessary

candiduslynx, line 24:
```suggestion
  Absolute or relative path to a file, such as `./example.duckdb`
```

candiduslynx (changes requested):

website/pages/docs/plugins/destinations/duckdb/overview.mdx
candiduslynx: OK, needs 2 spaces before it then

plugins/destination/duckdb/client/read.go
hermanschaaf: I don't think it really matters, and it's similar to most other destination plugins, so will keep it this way

plugins/destination/duckdb/client/read.go
hermanschaaf: Good catch!

candiduslynx (approved):

plugins/destination/duckdb/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM ghcr.io/gythialy/golang-cross:1.20.1-0
+
+# install goreleaser
+# https://github.com/goreleaser/goreleaser-pro/releases/download/v1.12.3-pro/goreleaser-pro_Linux_x86_64.tar.gz
+ARG GORELEASER_VERSION=1.12.3-pro
```
erezrokah, line 5: Opened https://github.com/cloudquery/cloudquery/pull/8833 so we can auto update this version

erezrokah (approved):
Looks great, don't have any comments on the code.
Will do a bit of testing

erezrokah:
> Will do a bit of testing

Looks good after a bit of testing

PR #8764, opened 2023-03-07
chore: Fix readme table links

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/8732 to fix the links in the README.md files (we could also bring the tables back altogether, but I'm hoping this is only temporary that we have the README under both the plugin and the Website)


hermanschaaf (approved):
LGTM, however the GCP workflow failed @erezrokah

PR #8789, opened 2023-03-07
feat: Facebook Marketing

shimonp21:
Facebook Marketing plugin.
See gist https://gist.github.com/shimonp21/a4a79c744d045cd95a4569b2fba2013f for the codegen for this plugin.


candiduslynx (dismissed):
ignore `plugins/source/facebookmarketing/facebookmarketing` and remove from PR

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -49,6 +49,7 @@ jobs:
                            "plugins/source/cloudflare",
                            "plugins/source/datadog",
                            "plugins/source/digitalocean",
+                           "plugins/source/facebookmarketing",
```
erezrokah, line 52: This is not needed anymore, see https://github.com/cloudquery/cloudquery/pull/8805

plugins/source/facebookmarketing/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The Facebook Marketing source plugin extracts information from the [Facebook Marketing API](https://developers.facebook.com/docs/marketing-api/reference/v16.0).
+
+## Links
+
+- [User Guide](https://cloudquery.io/docs/plugins/sources/hubspot/overview)
+
+## Development
+
+Much of the code here was "auto-generated" with https://gist.github.com/shimonp21/a4a79c744d045cd95a4569b2fba2013f. It requires cloning https://github.com/facebook/facebook-business-sdk-codegen as a prerequisite.
```
erezrokah, line 11:
```suggestion
Most of the code here was auto-generated with https://gist.github.com/shimonp21/a4a79c744d045cd95a4569b2fba2013f. It requires cloning https://github.com/facebook/facebook-business-sdk-codegen as a prerequisite.
```

plugins/source/facebookmarketing/README.md
```diff
@@ -0,0 +1,11 @@
+# Facebook Marketing Source Plugin
+
+The Facebook Marketing source plugin extracts information from the [Facebook Marketing API](https://developers.facebook.com/docs/marketing-api/reference/v16.0).
+
+## Links
+
+- [User Guide](https://cloudquery.io/docs/plugins/sources/hubspot/overview)
```
erezrokah, line 7:
```suggestion
- [User Guide](https://cloudquery.io/docs/plugins/sources/facebookmarketing/overview)
```

plugins/source/facebookmarketing/client/resolvers.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/thoas/go-funk"
+)
+
+// We can't use the default resolver because facebook's time format is not RFC3339.
+// '2023-03-05T16:46:23+0200' instead of RFC3339 '2023-03-05T16:46:23+02:00' (missing ':').
```
erezrokah, line 12: 😞

plugins/source/facebookmarketing/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Dependency directories (remove the comment below to include it)
+# vendor/
+
+
+.DS_Store
+
+facebookmarketing
```
erezrokah, line 20: Nice one of ignoring it, but you should also remove the binary as it's committed at the moment

plugins/source/facebookmarketing/go.mod
```diff
@@ -0,0 +1,37 @@
+module github.com/cloudquery/cloudquery/plugins/source/facebookmarketing
+
+go 1.19
+
+require (
+	github.com/cloudquery/plugin-sdk v1.40.0
```
erezrokah, line 6: The latest SDK is `v1.42.0`. Can we use the latest version?

plugins/source/facebookmarketing/resources/plugin/plugin.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+var (
+	Version = "development"
+)
+
+func Plugin() *source.Plugin {
+	return source.NewPlugin(
+		"cloudquery-facebookmarketing",
```
erezrokah, line 16:
```suggestion
		"facebookmarketing",
```

I don't think we prefix our own plugins

plugins/source/facebookmarketing/README.md
erezrokah: Maybe also add step by step docs on how to generate the code?

website/versions/source-facebookmarketing.json
```diff
@@ -0,0 +1 @@
+{ "latest": "plugins-source-facebookmarketing-v1.0.0" }
```
erezrokah, line 1: We don't need this file. The Website will show the plugin as "Unpublished" until we release it, and then the file will be created automatically.

erezrokah (approved):
Looks great, nothing blocking I think. Will do a bit of testing and add more comments

erezrokah (changes requested):
Sorry I missed something. Can you add a `[tables].mdx` file so the tables are rendered as a part of the Website, see https://github.com/cloudquery/cloudquery/blob/main/website/pages/docs/plugins/sources/alicloud/tables/%5Btable%5D.mdx

I'll update the contributing docs in a bit

website/pages/docs/plugins/sources/facebookmarketing/_configuration.mdx
```diff
@@ -0,0 +1,16 @@
...
+  # Source spec section
+  name: facebookmarketing
+  path: cloudquery/facebookmarketing
+  version: "VERSION_SOURCE_FACEBOOKMARKETING"
+  tables: ["*"]
+  destinations: ["DESTINATION_NAME"]
+
+  spec: 
+    # Facebook Marketing Spec section
+    ad_account_id: "AD_ACCOUNT_ID"
```
erezrokah, line 13: Maybe also explain where to find the account ID?

erezrokah:
Tested it a bit, looks good. It took me a bit of time to find the account ID, so maybe we can link to some Facebooks docs about it, WDYT?

plugins/source/facebookmarketing/README.md
yevgenypats: I would argue we can just drop this section as it's not something we can maintain re-generate once we push a first version, so users will need to manually edit this anyway.

erezrokah: That works too 🚀

plugins/source/facebookmarketing/client/resolvers.go
shimonp21: :)

plugins/source/facebookmarketing/README.md
shimonp21: I made the section shorter, but I think the script may be useful to future contributors (at least so they know where to find facebook's computer-readable API spec.

erezrokah:
Approved, just maybe ignore `codegen`. in our linter. See https://github.com/cloudquery/cloudquery/actions/runs/4373900846/jobs/7652659552#step:4:39

plugins/source/facebookmarketing/Makefile
```diff
@@ -0,0 +1,23 @@
...
+.PHONY: lint
+lint:
+	golangci-lint run --config ../../.golangci.yml
+
+.PHONY: gen-docs
+gen-docs:
+	go run main.go doc --format json docs
+	rm docs/__tables.json
+
+	rm -rf ../../../website/tables/facebookmarketing
```
candiduslynx, line 14: we use this code a lot.
@erezrokah maybe we can move it to root of source plugins and call with plugin dir param?
like `../../gen-docs facebookmarketing`
and contents like
```bash
#!/usr/bin/env bash
BASEDIR=$(dirname "$0")

rm -rf "${BASEDIR}/../../website/tables/${1}"

go run "${BASEDIR}/${1}/main.go" doc "${BASEDIR}/../../website/tables/${1}"
sed -i.bak -e 's_(\(.*\).md)_(tables/\1)_' "${BASEDIR}/../../website/tables/${1}/README.md"
mv "${BASEDIR}/../../website/tables/${1}/README.md" "${BASEDIR}/../../website/tables/${1}/README.md" "${BASEDIR}/../../website/pages/docs/plugins/sources/${1}/tables.md"
sed -i.bak -e 's_(\(.*\).md)_(\1)_' "${BASEDIR}/../../website/tables/${1}/*.md"
rm -rf "${BASEDIR}/../../website/tables/${1}/*.bak"
```

plugins/source/facebookmarketing/rest/ad.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+type Ad struct {
+	AccountId                         string         `json:"account_id"`
```
candiduslynx, line 12: `Id` -> `ID` per Go recommendations (here + resolvers)

plugins/source/facebookmarketing/Makefile
erezrokah: What do you mean by move to root? I think we'd still want to have the code duplicated, otherwise if we put the script under `plugins/source` for example we'll need to run the CI for every plugin when `plugins/source` changes

candiduslynx: Root = `plugins/source` dir.
We might want to run those, yes.
But, maybe not required initially, as every change to this file should also include manual generation of all source plugins?

plugins/source/facebookmarketing/rest/ad.go
shimonp21: I really dislike this golang recommendation (because it makes it  hard for me to parse an object-name into 'words').
Also, this was autogenerated by a script from a JSON file that is 'snake-case', and the script really has no way of knowing it needs to change 'Id' to 'ID' (without adding some sort of 'english dictionary' capability to the python script, which I don't want to do..).

plugins/source/facebookmarketing/client/resolvers.go
```diff
@@ -0,0 +1,33 @@
...
+
+func FacebookDatetimeResolver(path string) schema.ColumnResolver {
```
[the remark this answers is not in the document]
shimonp21: I don't see a big difference between these approaches, so I think it's OK to leave it as-is.

PR #8791, opened 2023-03-07
fix: Respect tables and skip_tables config in PostgreSQL source plugin

hermanschaaf:
Before this change, the PostgreSQL source plugin always synced all tables in the source database. This would lead to a panic if any table filters were applied, because the resources sent to the destination would be for tables that the destination does not have in its list of tables.

With this fix, users can choose which tables to sync using a source config, like usual. The default is still to sync all tables.


disq (approved):

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -52,11 +52,12 @@ func getTestConnectionString() string {
 	return testConn
 }
 
-func create_test_table(ctx context.Context, c *client.Client, table *schema.Table) error {
+func createTestTable(ctx context.Context, conn *pgxpool.Pool, table *schema.Table) error {
```
hermanschaaf, line 55: Most of these test changes were made so that we can pass in a separate connection (not the one managed by the plugin). I think it's better this way, but it's also because now the plugin will error with "pattern matching no tables" if a table is specified that doesn't exist yet

yevgenypats (approved):

erezrokah (approved):

PR #8799, opened 2023-03-07
fix(aws): Log Error On Skipped Multiplexer

bbernays:
#### Summary

Currently for the AWS plugin, if you specify a region and a table/resource that is not available in that region the user will have to have logs set to `trace` in order to get information that a specific region does not support the resource. This PR will:

- Have a single log message for all of the regions that are skipped on a per table basis (so a service that has multiple resources will have multiple log messages)
- If a resource is going to be fully skipped then it will generate a `error` log message, but if it has at least a single region then it will just generate a `warning` log message


erezrokah:
> Currently for the AWS plugin, if you specify a region and a table/resource that is not available in that region the user will have to have logs set to `trace` in order to get information that a specific region does not support the resource. This PR will make change the log level from `trace` to `warn`

How does this look like on our playground account? I remember there are quite a few tables that are not supported in all regions so I'm worried this will flood our logs with many warnings

disq (approved):
> This has the downside of printing warnings whenever a user uses table: ["*"].

I think that's OK

candiduslynx (approved):

erezrokah:
I got this 
image

With the following config:
```yaml
kind: source
spec:
  name: aws
  version: v7.3.2
  destinations: [postgresql]
  path: localhost:7777
  registry: grpc
  tables: ["aws_*_last_accessed_details"]
```

This seems a bit too much no?

candiduslynx:
Are we able to check that there's at least 1 match for account-region-service that we will be fetching resources for prior to printing the warn message?

bbernays:
> Are we able to check that there's at least 1 match for account-region-service that we will be fetching resources for prior to printing the warn message?

Not that I am aware of...

plugins/source/aws/client/multiplexers.go
```diff
@@ -32,7 +32,12 @@ func AccountMultiplex(meta schema.ClientMeta) []schema.ClientMeta {
 			region := getRegion(client.ServicesManager.services[partition][accountID])
 			// Ensure that the region is always set by a region that has been initialized
 			if region == "" {
-				meta.(*Client).Logger().Trace().Str("accountID", accountID).Str("partition", partition).Msg("no valid regions have been specified for this account")
+				if !client.allRegions {
```
erezrokah, line 35: Can we do `if client.allRegions` instead of inverted logic (`if !client.allRegions`)?

plugins/source/aws/client/multiplexers.go
```diff
@@ -119,17 +134,25 @@ func ServiceAccountRegionScopeMultiplexer(service string) func(meta schema.Clien
 				l = append(l, client.withPartitionAccountIDRegionAndScope(partition, accountID, cloudfrontScopeRegion, wafv2types.ScopeCloudfront))
 				for region := range client.ServicesManager.services[partition][accountID] {
 					if !isSupportedServiceForRegion(service, region) {
+
```
candiduslynx:
```suggestion
```

plugins/source/aws/client/multiplexers.go
```diff
@@ -119,17 +134,25 @@ func ServiceAccountRegionScopeMultiplexer(service string) func(meta schema.Clien
...
+							skippedRegions = append(skippedRegions, region)
 						}
+						client.Logger().Trace().Str("service", service).Str("region", region).Str("partition", partition).Msg("region is not supported for service")
 						continue
 					}
 					l = append(l, client.withPartitionAccountIDRegionAndScope(partition, accountID, region, wafv2types.ScopeRegional))
 				}
 			}
 		}
+		if len(skippedRegions) > 0 {
```
candiduslynx: extract into separate func for DRY sake?

candiduslynx (approved):

plugins/source/aws/client/multiplexers.go
bbernays: Changed the name so its not inverted

plugins/source/aws/client/multiplexers.go
```diff
@@ -106,13 +119,27 @@ func ServiceAccountRegionScopeMultiplexer(service string) func(meta schema.Clien
...
 }
+
+func generateLogMessages(client *Client, service string, skippedRegions []string, emptyMultiplexer bool) {
+	if len(skippedRegions) > 0 {
+		loggerEvent := client.Logger().Warn()
+		if emptyMultiplexer {
+			loggerEvent = client.Logger().Error()
+		}
+		loggerEvent.Str("service", service).Strs("regions", skippedRegions).Msg("specified regions are not supported for service")
+	}
```
candiduslynx, line 144: exit early (nit):
```suggestion
	if len(skippedRegions) == 0 {
		return
	}
	
	loggerEvent := client.Logger().Warn()
	if emptyMultiplexer {
		loggerEvent = client.Logger().Error()
	}
	loggerEvent.Str("service", service).Strs("regions", skippedRegions).Msg("specified regions are not supported for service")
```

candiduslynx (approved):

candiduslynx (approved):

erezrokah:
I think we'd still be getting duplicated log warnings, see:
image

Example config:
```yaml
kind: source
spec:
  name: aws
  version: v7.3.2
  destinations: [postgresql]
  path: localhost:7777
  registry: grpc
  tables: ["aws_iam*"]
  spec:
    regions: 
      - eu-central-1
```

Maybe we should either print the warning once per service, or once per table that uses the service (so users know which table caused the warning). Also can we print what regions are supported and suggest to use those in the configuration? That should allow users to self serve and fix the issue themselves. WDYT?

bbernays:
Here is the updated log message it includes the regions that are supported as well as the table name
Screenshot 2023-03-08 at 9 56 08 AM

plugins/source/aws/client/multiplexers.go
```diff
@@ -106,13 +123,31 @@ func ServiceAccountRegionScopeMultiplexer(service string) func(meta schema.Clien
...
+		generateLogMessages(client, table, service, skippedRegions, len(l) == 0)
 		return l
 	}
 }
+
+func generateLogMessages(client *Client, table, service string, skippedRegions []string, emptyMultiplexer bool) {
+	if len(skippedRegions) == 0 {
+		return
+	}
+	loggerEvent := client.Logger().Warn()
```
erezrokah, line 145: I still not sure this needs to be a warning. Let's say I use `tables: ["*"]` and specify a region, that is supported on some tables, and not supported on others. I'll get the warning, but the only way to fix it is to split the config correct? We don't have a way to skip regions for a specific table.
If that's the only way to fix the warning, I'd make this informational.

I agree that if we detect that a table is skipped altogether we should print an error

plugins/source/aws/client/client.go
```diff
@@ -39,6 +39,8 @@ type Client struct {
 	Partition            string
 	LanguageCode         string
 	Backend              backend.Backend
+	allRegions           bool
```
erezrokah, line 42:
```suggestion
```

I think this is unused

plugins/source/aws/client/multiplexers.go
bbernays: To me skipping a table is the same level as skipping an account... and we log a `warn` entry when an account is skipped

erezrokah: But we're not skipped a table altogether. We only skip it for some regions where

erezrokah: For 1. I don't think there's a way to fix it other than splitting the config. For 2. you can simply skip the table

bbernays: Changed it to `Info` and `Warn`

erezrokah (approved):

PR #8823, opened 2023-03-08
fix(website): Add canonical links

hermanschaaf:
I noticed that Google has been indexing some pages under different domains, e.g. `docs.cloudquery.io`, `v1.cloudquery.io` - this adds canonical links to point search engines to the main `www.` version of every page. I made use of the advice on this page https://rishimohan.me/blog/nextjs-canonical-tag to make sure it also works for ISR pages.


erezrokah (approved):
Nice 🚀

PR #8825, opened 2023-03-08
fix(neo4j): Logging and small fixes

disq:
- Enable logging and Bolt-logging (trace level)
- Upgrade neo4j driver to 5.6.0
- Log Cypher statements before execution
- Match `_cq_source_name` in `DeleteStale()`


plugins/destination/neo4j/client/deletestale.go
```diff
@@ -9,18 +9,16 @@ import (
...
 	defer session.Close(ctx)
 	for _, table := range tables {
 		stmt := fmt.Sprintf(deleteCypher, table.Name)
-		if _, err := session.Run(ctx, stmt, map[string]any{"_cq_sync_time": syncTime}); err != nil {
+		if _, err := session.Run(ctx, stmt, map[string]any{"$cq_source_name": source, "cq_sync_time": syncTime}); err != nil {
 			return err
 		}
 	}
-	return nil
+	return session.Close(ctx)
```
disq, line 23: To catch any potential errors from `session.Close`, we call it on return. The second call in `defer` is allowed and becomes a no-op.

erezrokah (approved):

Issue #8835, opened 2023-03-08
Azure Missing Tables From V0


bbernays:
### Describe the Bug
Following tables were present in V0, but the data is not available in V1...

The following tables need to be reimplemented:

`azure_servicebus_namespace_topics`
`azure_servicebus_namespace_private_endpoint_connections`
`azure_servicebus_namespace_topic_authorization_rules`

PR #8841, opened 2023-03-08
feat(azure-resources): Add Service Bus Namespace resources

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8835. ~~Still need to test on real resources~~. Tested on real resources.

This aligns Azure service bus to v1 (which is aligned with v0), see https://github.com/cloudquery/cloudquery/tree/plugins-source-azure-v1.0.0/plugins/source/azure/resources/services/servicebus.

### Notes
- We don't need `azure_servicebus_namespace_private_endpoint_connections` as that information is available on the namespace already. https://learn.microsoft.com/en-us/rest/api/servicebus/stable/namespaces/list?tabs=HTTP#sbnamespace
- There are also auth rules and keys at the namespace level. Did not add those yet (we can do based on deman).


plugins/source/azure/resources/services/servicebus/topic_authorization_rules.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
+	"github.com/cloudquery/plugin-sdk/faker"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+	"github.com/gorilla/mux"
+)
+
+func topicAuthorizationRules() *schema.Table {
+	return &schema.Table{
+		Name:        "azure_servicebus_namespace_topic_authorization_rules",
```
erezrokah, line 18: There are also auth rules at the namespace level (which are different) hence the prefixing

plugins/source/azure/resources/services/servicebus/topic_rules_access_keys.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
+	"github.com/cloudquery/plugin-sdk/faker"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+	"github.com/gorilla/mux"
+)
+
+func topicRuleAccessKeys() *schema.Table {
+	return &schema.Table{
+		Name:        "azure_servicebus_namespace_topic_rule_access_keys",
```
erezrokah, line 18: Had to trim `authorization_rule` to `rule` otherwise the table name is too long

website/tables/azure/azure_servicebus_namespace_topic_rule_access_keys.md
```diff
@@ -0,0 +1,27 @@
...
+| ------------- | ------------- |
+|_cq_source_name|String|
+|_cq_sync_time|Timestamp|
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|subscription_id|String|
+|rule_id (PK)|String|
+|alias_primary_connection_string|String|
+|alias_secondary_connection_string|String|
+|key_name (PK)|String|
```
erezrokah, line 23: Key names are unique based on my tests (create a key with the same name using the CLI overrides the previous one).

The rule ID has the subscription, namespace and topic built in, along with key name should be a good PK

bbernays (approved):

PR #8842, opened 2023-03-08
feat: Gremlin destination

disq:
- ~Authentication~ ~(AWS IAM auth can't be done for now, due to the package not providing an option to sign requests)~
- ~Documentation~
- ~CI~ (Using Gremlin server)

Since [Neptune](https://docs.aws.amazon.com/neptune/latest/userguide/intro.html) doesn't support lists all array types are stringified. It supports sets but then element order becomes an issue when running tests.

**UPDATE** Thanks to the nice people at TinkerPop discord IAM auth now works... but it's not auto-renewed so after 5 minutes it would expire

CosmosDB won't work, lots of compatibility issues [documented here](https://learn.microsoft.com/en-us/azure/cosmos-db/gremlin/support).


plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (*Client) TransformInt8Array(v *schema.Int8Array) any {
+	return v.String()
+	//res := make([]any, len(v.Elements))
+	//for i, e := range v.Elements {
+	//	res[i] = e.Int
+	//}
+	//return res
```
erezrokah, line 44:
```suggestion
```

plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (*Client) TransformTextArray(v *schema.TextArray) any {
+	return stripNulls(v.String())
+	//res := make([]any, len(v.Elements))
+	//for i, e := range v.Elements {
+	//	res[i] = stripNulls(e.String())
+	//}
+	//return res
```
erezrokah, line 67:
```suggestion
```

plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (*Client) TransformUUIDArray(v *schema.UUIDArray) any {
+	return v.String()
+	//res := make([]any, len(v.Elements))
+	//for i, e := range v.Elements {
+	//	res[i] = e.String()
+	//}
+	//return res
```
erezrokah, line 90:
```suggestion
```

plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (*Client) TransformCIDRArray(v *schema.CIDRArray) any {
+	return v.String()
+	//res := make([]any, len(v.Elements))
+	//for i, e := range v.Elements {
+	//	res[i] = e.String()
+	//}
+	//return res
```
erezrokah, line 106:
```suggestion
```

plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (*Client) TransformInetArray(v *schema.InetArray) any {
+	return v.String()
+	//res := make([]any, len(v.Elements))
+	//for i, e := range v.Elements {
+	//	res[i] = e.String()
+	//}
+	//return res
```
erezrokah, line 122:
```suggestion
```

plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func (*Client) TransformMacaddrArray(v *schema.MacaddrArray) any {
+	return v.String()
+	//res := make([]any, len(v.Elements))
+	//for i, e := range v.Elements {
+	//	res[i] = e.String()
+	//}
+	//return res
```
erezrokah, line 138:
```suggestion
```

plugins/destination/gremlin/client/write.go
```diff
@@ -0,0 +1,73 @@
...
+		)
+
+		for column := range nonPKs {
+			//if list, ok := rows[i][column].([]any); ok {
+			//	g = g.SideEffect(AnonT.Properties(column).Drop())
+			//	for _, item := range list {
+			//		g = g.Property(gremlingo.Cardinality.Set, column, item)
+			//	}
+			//	continue
+			//}
```
erezrokah, line 62:
```suggestion
```

erezrokah (approved):
Looks good, nothing blocking though I haven't tested it yet

plugins/destination/gremlin/client/read.go
```diff
@@ -0,0 +1,55 @@
...
+				if rowData[col.Name] == nil {
+					continue
+				}
+
+				data := rowData[col.Name].([]any)
+				if l := len(data); l == 1 {
+					madeRow[i] = data[0]
+				} else if l > 1 {
+					return fmt.Errorf("expected 1 value for %v, got %v", col.Name, l)
+					//madeRow[i] = data
```
candiduslynx, line 47:
```suggestion
```

plugins/destination/gremlin/client/client.go
```diff
@@ -0,0 +1,76 @@
...
+type Client struct {
+	destination.UnimplementedUnmanagedWriter
+	destination.DefaultReverseTransformer
+	logger     zerolog.Logger
+	spec       specs.Destination
+	pluginSpec Spec
+	client     *gremlingo.DriverRemoteConnection
+}
+
+var AnonT = gremlingo.T__
```
candiduslynx, line 23: why not using it directly?

disq: StackOverflow told me to. In previous incarnations this was used more than twice, now it's used just two times but it's easier to use/read still.

plugins/destination/gremlin/main.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/cloudquery/plugin-sdk/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const (
+	sentryDSN = "https://eabea8e3c07b4c44a298546306641da0@o1396617.ingest.sentry.io/4504809326641152"
+)
+
+func main() {
+	p := destination.NewPlugin("gremlin", plugin.Version, client.New, destination.WithManagedWriter(), destination.WithDefaultBatchSize(1000))
```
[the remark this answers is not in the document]
disq: We don't really do any batching and they're all upserted one by one for now. Sure let's start with 200.

plugins/destination/gremlin/client/write.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	gremlingo "github.com/apache/tinkerpop/gremlin-go/v3/driver"
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resources [][]any) error {
```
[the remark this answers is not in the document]
disq: It was, along with more than a kilobyte of chatgpt logs. I'll check again.

plugins/destination/gremlin/client/transformer.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (*Client) TransformCIDR(v *schema.CIDR) any {
+	if v.Status == schema.Present {
+		return v.String()
+	}
+	return nil
+}
+
+func (*Client) TransformCIDRArray(v *schema.CIDRArray) any {
+	return v.String()
```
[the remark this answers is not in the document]
disq: Sure. sqlite plugin still doesn't.

plugins/destination/gremlin/client/write.go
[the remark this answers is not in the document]
disq: fold-unfold is there, buried with the AnonT stuff.

plugins/destination/gremlin/main.go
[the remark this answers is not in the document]
disq: Batched upserts are now also added

disq:
TODO: Iterate() needs retries

> Iterate: E0502: error in read loop, error message '{code:244 message:{\"requestId\":\"xxxx\",\"code\":\"ConcurrentModificationException\",\"detailedMessage\":\"Failed to complete Insert operation for a Vertex due to conflicting concurrent operations. Please retry. 0 transactions are currently rolling back.\"} attributes:map[]}'. statusCode: 244

plugins/destination/gremlin/client/client.go
```diff
@@ -0,0 +1,66 @@
...
+	if err := spec.Validate(); err != nil {
+		return nil, err
+	}
+
+	c.pluginSpec = spec
+
+	c.client, err = gremlingo.NewDriverRemoteConnection(spec.Endpoint+"/gremlin",
+		func(settings *gremlingo.DriverRemoteConnectionSettings) {
+			settings.TraversalSource = "g"
+			settings.LogVerbosity = gremlingo.Debug
```
candiduslynx, line 45: IMO it should be controlled by spec

plugins/destination/gremlin/client/write.go
```diff
@@ -0,0 +1,65 @@
...
+		for column := range nonPKs {
+			g = g.Property(gremlingo.Cardinality.Single, column, rows[i][column])
+		}
+	}
+
+	if err := <-g.Iterate(); err != nil {
+		return fmt.Errorf("Iterate: %w", err)
+	}
+
+	return nil
```
candiduslynx, line 64:
```suggestion
	return <-g.Iterate()
```

website/pages/docs/plugins/destinations/gremlin/_configuration.mdx
```diff
@@ -0,0 +1,16 @@
...
+  name: "gremlin"
+  registry: "github"
+  path: "cloudquery/gremlin"
+  version: "VERSION_DESTINATION_GREMLIN"
+  # batch_size: 10000 # optional
+  # batch_size_bytes: 5242880 # optional
+  spec:
+    endpoint: "ws://localhost:8182"
+    username: "${USERNAME}"
+    password: "${PASSWORD}"
```
candiduslynx, line 15:
```suggestion
    username: "${GREMLIN_USERNAME}"
    password: "${GREMLIN_PASSWORD}"
```

candiduslynx (approved):

plugins/destination/gremlin/client/write.go
disq: Error messages can be a bit cryptic and we usually can't tell where they're coming from. That's why I wanted to wrap them.

plugins/destination/gremlin/client/client.go
disq: All gremlingo errors are logged as `TRACE` level in our logger (see logger.go) this is so that we don't miss any log messages and don't need to worry about converting between zerolog vs. gremlingo log levels.

website/pages/docs/plugins/destinations/gremlin/overview.mdx
```diff
@@ -0,0 +1,62 @@
...
+  - `"ws://localhost:8182"`
+  - `"wss://your-endpoint.cluster-id.your-region.neptune.amazonaws.com"`
+
+- `insecure` (boolean, optional)
+
+  Whether to skip TLS verification. Defaults to `false`. This should be set on a MacOS environment when connecting to an AWS Neptune endpoint.
+
+- `auth_mode` (either `basic` or `iam`, default: `basic`)
+
+  Authentication mode to use. `basic` uses static credentials, `iam` uses AWS IAM authentication.
```
bbernays, line 54: Maybe we want to be specific that `iam` is `aws-iam` because I think gcp has IAM as well...

plugins/destination/gremlin/client/client.go
candiduslynx: I mean that we should disable it by default and enable only if the user requests so

plugins/destination/gremlin/client/spec.go
```diff
@@ -26,11 +35,18 @@ func (s *Spec) SetDefaults() {
 		}
 		s.Endpoint = strings.Join(e, "://")
 	}
+
+	if s.AuthMode == "" {
+		s.AuthMode = authModeBasic
```
candiduslynx, line 40: I'd even set `s.AuthMode = strings.ToLower(s.AuthMode)` prior

website/pages/docs/plugins/destinations/gremlin/overview.mdx
candiduslynx: just `aws`?

candiduslynx (commented):
nits

plugins/destination/gremlin/client/spec.go
disq: Done

website/pages/docs/plugins/destinations/gremlin/overview.mdx
disq: Yeah made it `aws`, also added `none` and removed the "`basic` without credentials => no auth" case so it's more clear to use.

plugins/destination/gremlin/client/client.go
disq: Done

candiduslynx (approved):

candiduslynx (approved):

candiduslynx (approved):

candiduslynx:
Do we want to add label to repo?

plugins/destination/gremlin/client/client.go
```diff
@@ -0,0 +1,122 @@
...
+
+		req, err := http.NewRequest(http.MethodGet, baseURL, strings.NewReader(""))
+		if err != nil {
+			return nil, err
+		}
+
+		cfg, err := config.LoadDefaultConfig(ctx, config.WithDefaultRegion("us-east-1"))
+		if err != nil {
+			return nil, fmt.Errorf("unable to load AWS SDK config: %w", err)
+		}
```
bbernays, line 101: I don't think this should be `us-east-1`... I think it should be `cfg.Region` otherwise the SDK will use the us-east-1 sts endpoint rather than the local endpoint that is in the same region as the neptune db

website/pages/docs/plugins/destinations/gremlin/overview.mdx
```diff
@@ -0,0 +1,66 @@
...
+  - `"ws://localhost:8182"`
+  - `"wss://your-endpoint.cluster-id.your-region.neptune.amazonaws.com"`
+
+- `insecure` (boolean, optional)
+
+  Whether to skip TLS verification. Defaults to `false`. This should be set on a MacOS environment when connecting to an AWS Neptune endpoint.
+
+- `auth_mode` (either `none`, `basic` or `aws`, default: `none`)
+
+  Authentication mode to use. `basic` uses static credentials, `aws` uses AWS IAM authentication.
```
bbernays, line 54: If a user would use the AWS source plugin, how would they define a different set of credentials?

bbernays: Gonna resole this comment because I think this needs to be handled in a more general fashion: https://github.com/cloudquery/cloudquery/issues/8868

plugins/destination/gremlin/client/client.go
disq: "Default Region" is just that, if there's a `region` specified in the config (or an `AWS_REGION` env var) it would be used instead? That's my thinking but could be mistaken

bbernays: I agree with your thinking... sorry, I misunderstood/misread `cfg` ... I was thinking it came from the `spec` rather than the `cfg` for the aws sdk...

disq: I made it simpler (or easier/clearer) and put in an `aws_region` field in the spec.

candiduslynx (approved):

plugins/destination/gremlin/client/client.go
```diff
@@ -115,8 +116,18 @@ func (c *Client) getAuthInfo(ctx context.Context, baseURL string) (*gremlingo.Au
...
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	return c.client.CreateSession()
+
+	sess, err := c.client.CreateSession()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return sess, func() {
```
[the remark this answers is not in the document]
disq: It would be too generic and wouldn't match return values etc.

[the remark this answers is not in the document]
candiduslynx: I mean you don't need to return it at all, just do defer on the session, instead of `defer close()`

candiduslynx (approved):

plugins/destination/gremlin/client/client.go
```diff
@@ -59,8 +59,10 @@ func New(ctx context.Context, logger zerolog.Logger, destSpec specs.Destination)
 			settings.AuthInfo = au
 			settings.MaximumConcurrentConnections = spec.MaxConcurrentConnections
 
-			if logger.GetLevel() <= zerolog.DebugLevel {
+			if logger.GetLevel() <= zerolog.TraceLevel {
 				settings.LogVerbosity = gremlingo.Debug
+			} else {
+				settings.LogVerbosity = gremlingo.Off
```
disq, line 65: This was added because i18n in the logger (why??) is not thread safe

plugins/destination/gremlin/client/write.go
```diff
@@ -80,7 +80,7 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resou
 		}
 
 		nb := bo.NextBackOff()
-		c.logger.Warn().Err(err).Str("backoff_duration", nb.String()).Msg("Iterate failed, retrying")
+		c.logger.Debug().Err(err).Str("backoff_duration", nb.String()).Msg("Iterate failed, retrying")
```
disq, line 83: ConcurrectModificationException is a common occurance, let's not make a big deal of it.

candiduslynx (approved):

PR #8844, opened 2023-03-09
feat(clickhouse): Add table engine option

### LINKED ISSUE #8667: Support for ClickHouse ReplicatedMergeTree Table Engine

CloudQuery uses the ClickHouse `MergeTree` table engine. We're running ClickHouse with multiple replicas and `MergeTree` tables don't handle replication. But ClickHouse supports replication with the `ReplicatedMergeTree` table engine.

It would be great if additional support could be added for using the `ReplicatedMergeTree` table engine for table creation.

This could possibly be done by adding a `replicated` option to the spec. If the option is true, the `ReplicatedMergeTree` engine is used, and if false (or omitted), the `MergeTree` engine is used.

The `ReplicatedMergeTree` table engine has mandatory configuration parameters that would also need to be exposed in the spec if using that table engine. I'll add details to the additional context that go into more detail.

There are also other table engines in the `*MergeTree` family (also with `Replicated*MergeTree` counterparts), so this could be a more generic feature if you'd like, but as you already support the `MergeTree` table engine, having it's Replicated counterpart seems like a good place to start (I also don't personally have a use-case for the other engines at the moment).

### Describe alternatives you've considered

It's possible to manually convert a MergeTree table to a ReplicatedMergeTree table as documented [here](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication#converting-from-mergetree-to-replicatedmergetree). But would require manually managing the tables and moving data on the filesystem level which I'd like to avoid on a production setup.
### Additional context

As mentioned above, the `ReplicatedMergeTree` table engine has mandatory configuration parameters. It requires that the ZooKeeper path and replica name are specified. These can be provided as parameters to the Engine in the `CREATE TABLE` statement, or default values can be set in the ClickHouse configuration (so they wouldn't need to be provided to CloudQuery). The ClickHouse documentation explains that [here](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication#replicatedmergetree-parameters).

**Some Other Useful Links**
- [Overview of the *MergeTree family of table engines](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree)
- [Overview of replication with *MergeTree tables](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication)

candiduslynx:
Closes #8667


plugins/destination/clickhouse/queries/tables_test.go
```diff
@@ -43,7 +43,31 @@ func TestCreateTableCluster(t *testing.T) {
...
+				Name:            "extra_col",
+				Type:            schema.TypeFloat,
+				CreationOptions: schema.ColumnCreationOptions{NotNull: true},
+			},
+			schema.Column{Name: "extra_inet_col", Type: schema.TypeInet},
+			schema.Column{Name: "extra_inet_arr_col", Type: schema.TypeInetArray},
+		},
+	}, "", &Engine{
+		Name:       "ReplicatedMergeTree",
+		Parameters: []any{"a", "b", 1, 2, 3},
```
hermanschaaf, line 69: Should we maybe test some of the other cases as well, like float and bool? Or are they being tested elsewhere?

candiduslynx: No, didn't add tests for those. I'll add them soon

candiduslynx: done!

hermanschaaf (approved):

PR #8849, opened 2023-03-09
chore: Don't set canonical on non production builds

erezrokah:
#### Summary

~~Still testing this (hence the log message).~~
Deploy previews are not indexed by default, so we don't need the canonical. See https://vercel.com/guides/are-vercel-preview-deployment-indexed-by-search-engines.

If we keep the canonical, every time we add a new page to the Website our broken link check fails on the PR since the link uses the main domain, which doesn't have that page deployed yet

### DIFF
--- a/website/components/_app.tsx
+++ b/website/components/_app.tsx
@@ -26,7 +26,6 @@ declare global {
 }
 
 const getCanaonicalUrl = (path: string) => {
-  console.log(process.env)
   if (process.env.NEXT_PUBLIC_VERCEL_ENV !== "production") {
     return
   }
--- a/website/components/_app.tsx
+++ b/website/components/_app.tsx
@@ -25,9 +25,18 @@ declare global {
   }
 }
 
+const getCanaonicalUrl = (path: string) => {
+  console.log(process.env)
+  if (process.env.NEXT_PUBLIC_VERCEL_ENV !== "production") {
+    return
+  }
+
+  return (`https://www.cloudquery.io` + (path === "/" ? "": path)).split("?")[0]
+}
+
 export default function Nextra({ Component, pageProps }) {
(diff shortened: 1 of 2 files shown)


website/components/_app.tsx
```diff
@@ -25,9 +25,17 @@ declare global {
   }
 }
 
+const getCanaonicalUrl = (path: string) => {
+  if (process.env.NEXT_PUBLIC_VERCEL_ENV !== "production") {
```
hermanschaaf, line 29: Will `NEXT_PUBLIC_VERCEL_ENV` be set to `production` when we deploy to a custom domain, like `v1` or `docs`?

erezrokah: I think it's set to `production` on any build than runs on the default branch:
image

Maybe we should do `VERCEL_URL===https://www.cloudquery.io` or something, but for the sake of fixing previews I don't think it matters

hermanschaaf (approved):

website/components/_app.tsx
```diff
@@ -25,9 +25,17 @@ declare global {
...
+    return
+  }
+
+  return (`https://www.cloudquery.io` + (path === "/" ? "": path)).split("?")[0]
+}
+
 export default function Nextra({ Component, pageProps }) {
     const router = useRouter();
-    const canonicalUrl = (`https://www.cloudquery.io` + (router.asPath === "/" ? "": router.asPath)).split("?")[0];
+    const canonicalUrl = getCanaonicalUrl(router.asPath);
```
disq, line 38: `getCanaonicalUrl` should be `getCanonicalUrl`

erezrokah: Fixed in https://github.com/cloudquery/cloudquery/pull/8872

PR #8858, opened 2023-03-10
fix(gcp): Replace id into name on resource_id column

aruneko:
#### Summary
I have noticed that some Google Cloud CIS v1.2.0 policies results return raw resource id (such as just a number) as a resource_id. It is hard for me to identify the resource on the Google Cloud Console.

Instead of raw resource id, I put resource name as a resource_id. I know some policies have returned resource name as a resource_id already. To unify the format of the resource_id column helps users find failed resources easily.


erezrokah (approved):
Thanks @aruneko 🚀

Issue #8411, opened 2023-02-23
MySQL Source + CDC


yevgenypats:
Per user request it will be interesting to have a MySQL Source similar to the PostgreSQL source we have with CDC support.

Also, will be interesting to validate/test with AWS Aurora MySQL (though it should work the same way I believe).

PR #8881, opened 2023-03-12
feat: Add MySQL source

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8411

~~Still in draft as needs testing, CI, Sentry project and release configuration.~~
~~Tests are passing though~~


plugins/source/mysql/client/types.go
```diff
@@ -0,0 +1,65 @@
...
+		return schema.TypeJSON, nil
+	case "binary":
+		return schema.TypeByteArray, nil
+	case "tinyint", "smallint", "int", "mediumint", "bigint":
+		return schema.TypeInt, nil
+	case "blob":
+		return schema.TypeByteArray, nil
+	}
+
+	return schema.TypeString, nil
```
erezrokah, line 64: MySQL doesn't have fancy types so unless we try to parse values, we can use `TypeCIDR`, `TypeMacAddr`, etc. so we default to string. I'll clean up the other places too

plugins/source/mysql/client/list_tables.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cloudquery/plugin-sdk/schema"
+)
+
+func Identifier(name string) string {
+	return fmt.Sprintf("`%s`", name)
```
candiduslynx, line 11: As it's heavily used during sync
```suggestion
	return "`" + name + "`"
```

candiduslynx (approved):

plugins/source/mysql/client/client.go
```diff
@@ -0,0 +1,70 @@
...
+	metrics     *source.Metrics
+	Tables      schema.Tables
+	db          *sql.DB
+	tableSchema string
+}
+
+var _ schema.ClientMeta = (*Client)(nil)
+
+func (*Client) ID() string {
+	return "source-mysql"
```
disq, line 27: Don't we include `c.spec.Name` in case there are multiple sources going into the same destination?

disq (approved):
LGTM

plugins/source/mysql/client/client.go
erezrokah: I think the `ID` is only used to distinguish clients when multiplexing (which we don't have here):
https://github.com/cloudquery/plugin-sdk/blob/0f58331ea77ea07562a98edd50bd51b156685db9/plugins/source/metrics.go#L60

erezrokah: See also https://github.com/cloudquery/cloudquery/blob/34cbb8d99e0f5183bd885023dc8a138322b860ba/plugins/source/postgresql/client/client.go#L41

PR #8912, opened 2023-03-13
fix(aws)!: Replace the type of columns ending with `_at` to `Timestamp`

candiduslynx:
Caused by https://github.com/aws/aws-sdk/issues/492

BEGIN_COMMIT_OVERRIDE
fix(aws)!: Replace the type of columns ending with `_at` to `Timestamp` (#8912)

BREAKING-CHANGE: Replace the type of columns ending with `_at` to `Timestamp`. Initially it wasn't the case as the API markup of AWS doesn't include the proper annotations. See https://github.com/aws/aws-sdk/issues/492 for more details.
The following tables are affected:
* `aws_guardduty_detector_findings`: `created_at`, `updated_at`
* `aws_guardduty_detector_members`: `invited_at`, `updated_at`
* `aws_guardduty_detectors`: `created_at`, `updated_at`
* `aws_securityhub_findings`: `created_at`, `first_observed_at`, `last_observed_at`, `updated_at`

END_COMMIT_OVERRIDE


erezrokah (changes requested):
Great find! Let's wait for https://github.com/aws/aws-sdk-go-v2/issues/2052 so we can be positive about the new types before we introduce a breaking change

erezrokah:
Hi @candiduslynx, can we close this PR? We should probably wait until https://github.com/aws/aws-sdk/issues/492#issuecomment-1466775393 is resolved and then do this as a dependency update

candiduslynx:
@erezrokah I'm reopening this PR as https://github.com/aws/aws-sdk/issues/492 was denied proper resolution

plugins/source/aws/client/transformers.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/thoas/go-funk"
+)
+
+func TimestampTypeTransformer(field reflect.StructField) (schema.ValueType, error) {
+	if !strings.HasSuffix(field.Name, "At") {
+		return schema.TypeInvalid, nil // fallback
+	}
+
+	switch field.Type {
+	case reflect.TypeOf(""), reflect.TypeOf(new(string)): // nop
```
disq, line 20: You can do the `return schema.TypeTimestamp` here so it's easier to read

erezrokah (approved):
🚀

PR #8918, opened 2023-03-13
chore(transformers): Use `client.TransformWithStruct` instead of appending default options

candiduslynx:
Sometimes we have default transformer options (e.g., type & resolver transformers).
We use them like this:
```go
schema.Table{
  Transform: transformers.TransformWithStruct(&someStruct, append(client.DefaultOptions, a,b,c)...),
}
```

However, it’ll be much easier to do the following instead:
- in the client package define the following func
  ```go
  func TransformWithStruct(t any, opts ...transformers.StructTransformerOption) schema.Transform {
    return transformers.TransformWithStruct(t, append(options, opts...)...)
  }
  ```
- And replace resource transform calls to `client.TransformWithStruct(t, a, b, c)`

This PR does exactly this.

### DIFF
--- a/plugins/source/azuredevops/client/transformers.go
+++ b/plugins/source/azuredevops/client/transformers.go
@@ -25,10 +25,10 @@ func typeTransformer(field reflect.StructField) (schema.ValueType, error) {
 	return schema.TypeInvalid, nil
 }
 
-func Options() []transformers.StructTransformerOption {
-	options := []transformers.StructTransformerOption{
-		transformers.WithTypeTransformer(typeTransformer),
-	}
+var options = []transformers.StructTransformerOption{
+	transformers.WithTypeTransformer(typeTransformer),
+}
 
-	return options
+func TransformWithStruct(t any, opts ...transformers.StructTransformerOption) schema.Transform {
+	return transformers.TransformWithStruct(t, append(options, opts...)...)
 }
--- a/plugins/source/launchdarkly/client/transformers.go
+++ b/plugins/source/launchdarkly/client/transformers.go
@@ -4,13 +4,16 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/transformers"
 )
 
-func SharedTransformers(others ...transformers.StructTransformerOption) []transformers.StructTransformerOption {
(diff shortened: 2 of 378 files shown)


plugins/source/launchdarkly/client/transformers.go
```diff
@@ -4,13 +4,16 @@ import (
...
 
+	"github.com/cloudquery/plugin-sdk/schema"
 	"github.com/cloudquery/plugin-sdk/transformers"
 )
 
-func SharedTransformers(others ...transformers.StructTransformerOption) []transformers.StructTransformerOption {
-	return append([]transformers.StructTransformerOption{
-		transformers.WithNameTransformer(nameTransformer),
-	}, others...)
+var options = []transformers.StructTransformerOption{
```
erezrokah, line 11: Maybe drop this also as there's only a single option?

plugins/source/mixpanel/client/transformers.go
```diff
@@ -8,10 +8,12 @@ import (
 	"github.com/cloudquery/plugin-sdk/transformers"
 )
 
-func SharedTransformers(others ...transformers.StructTransformerOption) []transformers.StructTransformerOption {
-	return append([]transformers.StructTransformerOption{
-		transformers.WithTypeTransformer(typeTransformer),
-	}, others...)
+var options = []transformers.StructTransformerOption{
```
erezrokah, line 11: Same here

erezrokah (approved):

erezrokah:
Non blocking but maybe drop the options where there's a single one. Can be done in a follow up

plugins/source/azuredevops/client/transformers.go
```diff
@@ -25,10 +25,10 @@ func typeTransformer(field reflect.StructField) (schema.ValueType, error) {
 	return schema.TypeInvalid, nil
 }
 
-func Options() []transformers.StructTransformerOption {
-	options := []transformers.StructTransformerOption{
-		transformers.WithTypeTransformer(typeTransformer),
-	}
+var options = []transformers.StructTransformerOption{
```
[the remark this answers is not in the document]
candiduslynx: I don't have any opinion on that, but I made it universal for every plugin to be just a copy-paste funcs, and those work with slice for options.
Do you have a strong opinion on this?

PR #8921, opened 2023-03-13
fix(website): Handle source -> dest integration with the same id

erezrokah:
#### Summary

Fixes an edge case in our integrations pages where someone selects a source and destination with the same id (e.g. MySQL->MySQL or Postgres->Postgres)


hermanschaaf (approved):
Nice 👍

Issue #8922, opened 2023-03-13
Bug: PostgreSQL source


yevgenypats:
When syncing the PostgreSQL source to clickhouse with the following config:

```
kind: source
spec:
  name: postgresql
  path: cloudquery/postgresql
  version: "v1.0.4"
  spec:
    connection_string: "postgres://<usn>:<pass>@localhost:5432/<db>?sslmode=disable"
  destinations: ["clickhouse"]
---
kind: destination
spec:
  name:     "clickhouse"
  registry: "github"
  path:     "cloudquery/clickhouse"
  version:  "v1.3.0"
  write_mode: append
  spec:
    connection_string: "clickhouse://<usn>:<pass>@localhost:9000/<db>?debug=true"
```

Use reported the following error:

```
getting error

panic: unknown type character varying(100)

goroutine 23 [running]:
github.com/cloudquery/cloudquery/plugins/source/postgresql/client.(*Client).Pg10ToSchemaType(...)
```

PR #8926, opened 2023-03-13
fix(source-postgres): Support more Postgres and CockroachDB types

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8922.

Unlike the Postgres destination where we know the column types in advance, in the source plugin we need to account for types that are not part of the SDK schema.

Based on https://www.postgresql.org/docs/current/datatype.html#DATATYPE-TABLE and https://www.cockroachlabs.com/docs/stable/data-types.html#supported-types.

~~I'll update the tests to create a table with each possible type~~ Done


plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -52,20 +53,78 @@ func getTestConnectionString() string {
...
 	sb.WriteString("CREATE TABLE ")
-	sb.WriteString(table.Name)
+	sb.WriteString(tableName)
 	sb.WriteString(" (")
-	c := client.Client{}
-	for i, col := range table.Columns {
-		sb.WriteString(col.Name)
+	columns := getPgTypesToData()
+	for i, col := range columns {
+		sb.WriteString(pgx.Identifier{col.typeName + "_type"}.Sanitize())
```
candiduslynx, line 121: Why not use column names from DB?

candiduslynx: will create clash on several columns with the same type

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -76,33 +135,105 @@ func createTestTable(ctx context.Context, conn *pgxpool.Pool, table *schema.Tabl
...
-		if i < len(table.Columns)-1 {
-			sb.WriteString(", ")
+func insertTestTable(ctx context.Context, conn *pgxpool.Pool, tableName string, columns []pgTypeToValue) error {
+	var query = ""
+	query += "INSERT INTO " + tableName + " ("
+	for _, col := range columns {
+		if col.value == nil {
+			continue
 		}
+		query += pgx.Identifier{col.typeName + "_type"}.Sanitize() + ", "
```
candiduslynx, line 145: clash on columns with the same type

candiduslynx (changes requested):

plugins/source/postgresql/resources/plugin/plugin_test.go
erezrokah: Can you please explain your comment? We don't get the column names from the DB. `columns` have test data generated here https://github.com/cloudquery/cloudquery/pull/8926/files#diff-715416571f0e3a5047731c92c67565cf315764ec01e5b2d59969b00bc99a0b70R61

plugins/source/postgresql/resources/plugin/plugin_test.go
erezrokah: See previous column

plugins/source/postgresql/resources/plugin/plugin_test.go
candiduslynx: Ah, I reviewed and didn't notice that this is a test.

plugins/source/postgresql/resources/plugin/plugin_test.go
candiduslynx: Ah, I reviewed and didn't notice that this is a test.

candiduslynx (approved):
LGTM

PR #8939, opened 2023-03-13
fix: Use the correct path for test file

02strich:
#### Summary

When using the S3 destination plugin, cloudquery failed with an Access Denied error around the test file despite correct permissions.

Beforehand the code used the plugins path (i.e. cloudquery/s3). That means it doesn't actually test the correct permissions. It should use the output path that is provided by the user.


candiduslynx:
Hi @02strich!
Thanks for the PR!

I've tested it in #8940.
LGTM

candiduslynx (approved):

PR #8940, opened 2023-03-13
fix(s3): Use proper path

candiduslynx:


erezrokah:
Closing since https://github.com/cloudquery/cloudquery/pull/8939 is merged

Issue #8943, opened 2023-03-14
Feedback for “Quickstart - Linux”


srijavoona-debug:
after running curl command for downloading cloudquery it is showing command not found

yevgenypats:
Hey @srijavoona-debug can you please share the exact commands you run?

erezrokah:
@yevgenypats I think our linux guide shows `cloudquery sync ./cloudquery-config` where it should show `./cloudquery sync ./cloudquery-config` as it instructs downloading the local binary.

I'll take a look

PR #8945, opened 2023-03-14
fix(linux-quickstart): Update sync command to use local binary

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8943.

This seems like a lot of changes, but it mostly breaks `_configure.mdx` into multiple components so we can have `_configure.mdx` and `_configure-linux.mdx`. The latter uses a different sync command component that has `./cloudquery sync` instead of `cloudquery sync` (notice the `./` prefix).

There's some follow up to do here for MacOS quick start (and also for the integrations page), to switch from `cloudquery` to `./cloudquery` and back when someone clicks the corresponding table (i.e. `cloudquery` for Homebrew and `./cloudquery` for precompiled binary).

### DIFF
--- a/website/components/mdx/_configure-authenticate-provider.mdx
+++ b/website/components/mdx/_configure-authenticate-provider.mdx
@@ -1,0 +1,6 @@
+## Authenticate with the Cloud Provider
+
+CloudQuery uses similar authentication methods to the official SDKs of each cloud provider.
+The AWS authentication guide is located [here](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/README.md).
+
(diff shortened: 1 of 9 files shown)


hermanschaaf:
We could also consider recommending that the binary gets copied to `/usr/local/bin`? Then at least you can always run `cloudquery`

erezrokah:
> We could also consider recommending that the binary gets copied to `/usr/local/bin`? Then at least you can always run `cloudquery`

Yeah, `/usr/local/bin` might require elevated permissions on some environments. Maybe `${HOME}/bin`?

hermanschaaf (approved):

PR #9012, opened 2023-03-14
feat(aws): Add Tags column for Savings Plans

bbernays:


erezrokah (commented):
Hi @bbernays, it seems `tags already exists on this table. Is this changing the structure of the column?

PR #9021, opened 2023-03-14
feat(aws): Add Support for Sagemaker Apps

bbernays:
#### Summary

Adds support for AWS Sagemaker Apps. Also resolves app tags

~~NOTE: Has not been tested against any live resources yet...~~ 
Update: created apps in sagemaker and tested to ensure that both types (user and space) of apps could be found


hermanschaaf (approved):
LGTM from a code perspective; always good to test against real resources though if possible

disq (approved):

PR #8803, opened 2023-03-07
feat: Add command to generate sources tables

erezrokah:
#### Summary

Needs https://github.com/cloudquery/plugin-sdk/pull/726.
Fixes https://github.com/cloudquery/cloudquery/issues/8731

PR #9025, opened 2023-03-14
fix(cmd-tables): Call terminate client after docs generation

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/8803


hermanschaaf (approved):
💯

erezrokah:
Not sure why the tests are hanging in CI, looking into it

cli/cmd/tables.go
```diff
@@ -74,6 +74,11 @@ func tables(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("failed to generate docs. Error: %w", err)
 		}
+
+		err = sourceClient.Terminate()
+		if err != nil {
+			fmt.Println("Failed to terminate source client. Error: ", err)
```
erezrokah, line 80: I think we can keep generating docs for other plugins when this happens. No reason to exit early

erezrokah:
@hermanschaaf This ended up being a bug fix to the hang (not sure why it happens only when running `go test ./..`), so I requested another review

hermanschaaf (approved):
Makes sense 👍

Issue #8811, opened 2023-03-07
Primary Key Improvements


bbernays:
### Is your feature request related to a problem? Please describe.

Providers do not do consistent job of documenting what fields are guaranteed to uniquely define a response. As such CloudQuery is forced to release periodic breaking changes in order to fix Primary Key violations. 

### Describe the solution you'd like

Rather than having composite Primary Keys which require a breaking schema change in order to fix, we are looking into supporting using a deterministic hashing method to define a single unique column whose value would be changed by changing the columns that are used to calculate the hash. This solution would mean PK changes would be non-breaking for users that are using the `overwrite-delete-stale` mode, but if a user is using `overwrite` there is nothing that can be done to fix the duplicate PK issues as changing the hash will always result in a new row rather than an update

We have added functionality in the `plugin-sdk` to make the `_cq_id` deterministic based on the Primary Key fields (unless there is no PK defined, then the value is still random). Proposal to making this the default behavior:

- [x] Update SDK so that `_cq_id` is deterministic based on defined Primary Key values (feature is behind an opt in flag)
- [x] Update all source plugins to use new SDK
- [x] Update Postgres destination to not ignore `_cq_id` and `_parent_cq_id` during update
- [ ] Introduce new transformer `transformers.WholeItemAsPK` that enables developers to easily use all fields in the PK hashing https://github.com/cloudquery/plugin-sdk/pull/735
- [x] Update `plugin-sdk` to support an attribute in the destination spec that enables users to opt into using the deterministic `_cq_id` + `_parent_cq_id` as the only PK(s) https://github.com/cloudquery/cloudquery/issues/8474
- [x] ~~Update all destinations that support `overwrite` or `overwrite-delete-stale` to respect PK flag~~ Not needed as handled in the SDK

yevgenypats:
That's great summary! I would also add that "Primary Keys should not require a breaking schema change in order to fix." is not really possible at the end of the day at least in Postgres. i.e the workaround (which might be useful for some of our customers) will only make `_cq_id` as pk consisting of different columns which means when updating it might not break the schema but it will create inconsistent data with potential duplicates but different `_cq_id` because the hash is changed.

erezrokah:
Going to close this as resolved. If something is missing please re-open. Amazing work championing this @bbernays 🥇

PR #9043, opened 2023-03-14
feat(postgresql): Add Support For Different PK Types

bbernays:
#### Summary

depends on: https://github.com/cloudquery/plugin-sdk/pull/732

Initial implementation, still needs testing!

Related to: https://github.com/cloudquery/cloudquery/issues/8811

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -337,7 +337,6 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 	sb.WriteString(" (")
 	totalColumns := len(table.Columns)
 
-	primaryKeys := []string{}
 	for i, col := range table.Columns {
 		pgType := c.SchemaTypeToPg(col.Type)
 		if pgType == "" {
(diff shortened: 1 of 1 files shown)


plugins/destination/postgresql/client/migrate.go
```diff
@@ -377,6 +374,29 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
...
 }
 
+func (c *Client) listPKs(table *schema.Table) []string {
+	// If PKS are not needed for the writeMode then just return early
+	if !c.enabledPks() {
+		return []string{}
+	}
+	pks := []string{}
+	if c.spec.PKMode == specs.PKModeCQID {
+		pks = append(pks, schema.CqIDColumn.Name)
```
erezrokah, line 384: I'm not sure destinations should reference CqID directly. Update the table PKs detentions in the SDK based on the spec PK mode, WDYT?

bbernays: If we don't reference the `CqID` how do we enable users to be able to switch between `composite-keys` and `cq-ids`?

erezrokah: See my comment here https://github.com/cloudquery/plugin-sdk/pull/732#pullrequestreview-1339843474
We can update the column options here https://github.com/cloudquery/plugin-sdk/blob/ac55bec5f3c06b3cdd9a0409394d5bf5cf059137/plugins/source/plugin.go#L282 for example

erezrokah: If we do it on destinations, we can override the PK definitions here https://github.com/cloudquery/plugin-sdk/blob/ac55bec5f3c06b3cdd9a0409394d5bf5cf059137/plugins/destination/plugin.go#L187

bbernays: Great call! Thanks for the heads up

bbernays:
Closing this, as it can be done 100% handled in the SDK (thank you @erezrokah for that tip!)

PR #9057, opened 2023-03-14
doc: Add 'skip_tables' to quickstart guide

shimonp21:
On our empty account, `tables=['*']` takes >15m (not sure how long exactly because I was too impatient to find out). The `skip_tables` section reduces this to 2m04s.
I believe it is very important for a user's first experience with cloudquery to be performant and smooth (at the cost of having a longer config file). Twice in the last week I had to help users on discord to add this section to their config, and it's required for almost anyone using cloudquery with AWS.


disq (approved):

erezrokah (approved):
We've discussed this in the past. Maybe we should have `skip by default` option in the SDK as a follow up

bbernays (approved):
Love it!

PR #9079, opened 2023-03-14
fix(aws): AWS Password Policy Saved Query

dacamp:
#### Summary

The password policy should fail if the max_password_age is greater than (>) 90


hermanschaaf (approved):
Thanks for the fix @dacamp !

PR #9101, opened 2023-03-15
fix(aws): Fixed `aws_ecs_clusters` fetch to include cluster settings

KMConner:
#### Summary

There  exists `settings` field in `aws_ecs_clusters` table.
However, this field is always empty because settings field is not included in AWS API response by default.
(See https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeClusters.html for more information.)

So, I fixed API request parameters to include settings field.


hermanschaaf (approved):
Thanks @KMConner, great catch! I've also added the other missing types to the request, so hopefully we've got it all now

Issue #8929, opened 2023-03-13
Azure Source Enhancement: Consumption Data


tomtom215:
### Is your feature request related to a problem? Please describe.

We are testing CloudQuery in our Azure Enterprise subscription and would like to get more consumption data than just the cost report data that the current source plugin provides.

### Describe the solution you'd like

I would like to request the enhancement of the Azure Source plugin to include the possibility to include Azure Enterprise Consumption API data found here https://learn.microsoft.com/en-us/rest/api/consumption/

### Describe alternatives you've considered

Running CloudQuery Azure Source Plugin then manually running a batch of API requests to the Consumption API to enhance the CloudQuery data
### Additional context

I understand that this would require an Azure Enterprise subscription to test, so I'd be happy to try out any PR's

tomtom215:
Thanks for such a quick turn around @erezrokah ! We will be able to to test this early next week (this coming Monday/Tuesday) and will report any findings here. Please let us know if there's anything specific you'd like us to look into or try out

erezrokah:
Hi @tomtom215, thanks for the quick response! I marked https://github.com/cloudquery/cloudquery/pull/9117 as ready for review and followed up with a few updates based on my tests. Mostly I moved some APIs to use a billing profile scope as it that should work on all accounts, and also required to get the "Modern" responses.

As we don't have an enterprise account I couldn't validate all the data, but I did validate the API calls don't return an error (some return 404 or 204, but it appears that's expected based on the content of the error I got).

One thing to pay attention to is the `azure_consumption_billing_profile_reservation_recommendations`, `azure_consumption_billing_account_usage_details` and `azure_consumption_billing_account_charges` tables as those cast the response from an interface to a struct, see:
https://github.com/erezrokah/cloudquery/blob/c2664ed23a899733b2032ba04cd4b659586a76b7/plugins/source/azure/resources/services/consumption/billing_profile_reservation_recommendations.go#L43
https://github.com/erezrokah/cloudquery/blob/c2664ed23a899733b2032ba04cd4b659586a76b7/plugins/source/azure/resources/services/consumption/billing_account_usage_details.go#L35
https://github.com/erezrokah/cloudquery/blob/c2664ed23a899733b2032ba04cd4b659586a76b7/plugins/source/azure/resources/services/consumption/billing_account_charges.go#L33

The code is based on the docs, but I can't verify it without running on real resources

ethankelly99:
Hi @erezrokah! 
[...]

PR #9117, opened 2023-03-15
feat(azure-resources): Add Consumption Resources

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8929.

### Notes

- The consumption APIs require a `scope` argument that can be a subscription, billing account, billing account, enrollment account, and more. See example in https://learn.microsoft.com/en-us/rest/api/consumption/usage-details/list?tabs=HTTP#uri-parameters. We could support all scopes, but I'm starting with only billing account, profiles and subscriptions, based on the API so the result will have the "Modern" or "Legacy" struct returned (where applicable).
- Some APIs return 404 or 204  (no content) where there's no data, so I'm logging these errors instead of reporting them


erezrokah:
Moved back to draft, see https://github.com/cloudquery/cloudquery/issues/8929#issuecomment-1476544000

yevgenypats (approved):

erezrokah:
Hi @yevgenypats and thanks for the review. I updated this PR with 3 additional commits:
1. Multiplex over billing periods as required [here](https://github.com/cloudquery/cloudquery/issues/8929#issuecomment-1508724177). See https://github.com/cloudquery/cloudquery/pull/9117/commits/07789ef8449a9401f90d3102b203f6003c558433
2. Paginate over prices sheets as requested [here](https://github.com/cloudquery/cloudquery/issues/8929#issuecomment-1508724177). See https://github.com/cloudquery/cloudquery/pull/9117/commits/3424606ad47da5e567d2608440026e495e0d20da. I also opened https://github.com/Azure/azure-sdk-for-go/issues/20671 to see if there's a better way to do it.
3. Don't fail the sync if discovery of accounts and billing periods fail. See https://github.com/cloudquery/cloudquery/pull/9117/commits/3ad5f9f884e5cb372145afc11cf80e8388ed4dbb. Since we can't really test this thoroughly on all account types, I'd rather not fail the sync, as the discovery happens regardless if you configure the relevant tables or not.

I'll update the issue to get more feedback so I suggest we wait on reviewing the new changes until we receive it.

erezrokah:
Hi 👋 This issue was closed as we merged the consumption PR. It's released in [`v7.2.0`](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-azure-v7.2.0) of the Azure source plugin, as we wanted to push these new tables to more users and see if any other issues come up.

We might still need to do a bit of tuning to the resources, which we're happy to do in follow up PRs.
Please comment if the issue is not resolve and we can re-open it or open a new one with your requests.

PR #9123, opened 2023-03-15
fix(website): Support multiple authors in blog posts

hermanschaaf:
This allows us to specify e.g. `author: yevgenypats,hermanschaaf` in the metadata and get a result like this:

Screenshot 2023-03-15 at 16 55 16

I tested this locally because we don't actually have a blog post with multiple authors right now, but we may soon have.


erezrokah (approved):

website/components/BlogHeader.tsx
```diff
@@ -25,12 +25,13 @@ function Authors({ data }: { data: BlogPostMeta }) {
...
 
-  const author = CLOUDQUERY_TEAM[authorName];
+  const authorNames = authorName.split(",").map((name) => name.trim());
+  const authors = authorNames.map((name) => CLOUDQUERY_TEAM[name]);
 
   return (
     <div className="w-full border-b border-gray-400 authors border-opacity-20">
       <div className="flex flex-wrap justify-center py-8 mx-auto gap-7">
-        <Avatar {...author} />
+        {authors.map((author) => (<Avatar {...author} />))}
```
erezrokah, line 34:
```suggestion
        {authors.map((author) => (<Avatar key={author.name} {...author} />))}
```

I think that you need to specify a key when rendering arrays in React

erezrokah (changes requested):

website/components/BlogHeader.tsx
hermanschaaf: Ah yeah, good catch

Issue #8642, opened 2023-03-01
Google Ads Source plugin


bhavaniravi:
### Why do you need this resource or plugin? Please describe.
This is one of the most requested sources as a data freelancer.

Inspired from Airflow's version of GoogleAdstoGCS
https://airflow.apache.org/docs/apache-airflow-providers-google/stable/operators/ads.html

PR #9130, opened 2023-03-16
feat(googleads): Google Ads source plugin

candiduslynx:
Implements #8642


plugins/source/googleads/main.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/googleads/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/serve"
+)
+
+const sentryDSN = "" // TODO
```
disq, line 8: TODO found

disq (changes requested):
Few things around spec/init

plugins/source/googleads/client/hierarchy.go
```diff
@@ -0,0 +1,173 @@
...
+				res[k] = append(res[k], curr)
+			}
+		}
+	}
+	return res
+}
+
+// dedup makes sure that each value is saved only once
+func dedup(customers map[string][]string) map[string][]string {
+	saw := make(map[string]struct{})
```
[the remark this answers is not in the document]
candiduslynx: we can prealloc, but the real len will be based on the values, not the key amounts

plugins/source/googleads/main.go
candiduslynx: I'll do this in a bit

plugins/source/googleads/client/hierarchy.go
```diff
@@ -0,0 +1,173 @@
...
+
+	ids := make([]string, len(resp.ResourceNames))
+	for i, rn := range resp.ResourceNames {
+		ids[i] = strings.TrimPrefix(rn, "customers/")
+	}
+
+	return ids, nil
+}
+
+func limitTo(customers map[string][]string, only []string) map[string][]string {
```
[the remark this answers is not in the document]
candiduslynx: What we actually do here is not a dedup per se.

We have `customers` map representing the whole management tree accessible via credentials supplied (possibly limited by `Spec.LoginCustomerID`, but that's another story). This function traverses this tree to add all subtrees that have root matching `only` parameter (if it's empty we simply fetch everything).

The dedup func that is used afterwards is just making sure that we'll fetch the customer data only once (if we want to receive the data from 2 management accounts `A` and `B` and both of them have access to the customer `C` we'll add either `A:C` or `B:C` key-value pair, but not both).

plugins/source/googleads/client/context.go
```diff
@@ -13,21 +13,13 @@ const (
 
 func (c *Client) OutgoingContext(ctx context.Context) context.Context {
 	ctx = c.addDeveloperToken(ctx)
-	return c.addLoginCustomerID(ctx)
+	return addLoginCustomerID(ctx, c.ManagerID)
```
disq, line 16: Forgot to add `if len(c.ManagerID) > 0 { ...`

candiduslynx: I took a look at the logic and the manager ID is always present, so omitted intentionally.

candiduslynx: But I'll add it to be consistent

plugins/source/googleads/client/context.go
```diff
@@ -13,6 +13,9 @@ const (
 
 func (c *Client) OutgoingContext(ctx context.Context) context.Context {
 	ctx = c.addDeveloperToken(ctx)
+	if len(c.ManagerID) == 0 {
```
disq, line 16: If it's consistently there then why add it?

disq (approved):

plugins/source/googleads/client/context.go
candiduslynx: Well, it was me who implemented this initially.
If someone then updates the code without this knowledge, they might end up setting the `login-customer-id` to `""`, which, in its turn, may produce unexpected API response.
Better err on the safe side.

PR #9138, opened 2023-03-16
feat(website): Render integrations on demand on non production builds

erezrokah:
#### Summary

This PR's purpose is to improve build speed on non production builds, by rendering the integration pages on demand on such builds.
It does so by moving some of the code from the pre-build script into a catch all mdx component, similar to https://github.com/cloudquery/cloudquery/pull/8672 and `[table].mdx`.

This reduces our build time from 12m to 4m 🚀 

### Notes

- `[...slug].mdx` - The `...` means that it matches both `a` and `a/b`, etc. and `slug` will be an array of all path sections.
- `getStaticPaths` - returns an empty array on non production build, meaning all pages are rendered on demand for such builds. For production builds all paths are returned and will be rendered at build time.
- When rendering dynamic MDX in Nextra we need to pass any custom components to `RemoteContent`, hence the fancy dynamic imports there based on props.
- Also updated the site map configuration to account for these pages, which is why `integrations.js` needs to be in JS
- Instead of generating all the `_meta.json` files, it looks like the approach we need is to use `dynamicMeta` to inject those
- Technically we could probably move all the pre-build script to `[...slug].mdx`, but maybe save that for another PR


hermanschaaf (approved):
Amazing 🪄

PR #9140, opened 2023-03-17
feat(aws): Add `aws_iam_group_attached_policies`

tockn:
#### Summary

I am grateful for the amazing project.

Added `aws_iam_group_attached_policies` table of AWS plugin.
IAM groups can have not only inline policies but also managed policies attached. This PR implements the functionality to retrieve them.

- [ ] Run `make lint` to ensure the proposed changes follow the coding style 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))


tockn:
I don't know why `golangci-lint` is timeout 🤔

image

hermanschaaf:
Thank for this contribution @tockn ! Looks great, I just changed the `account_id` to not be part of the PK, because the Account ID is already included in the `group_arn`, similar to the `group_policies` table. Let me know if you don't think this should be the case.

hermanschaaf (approved):

tockn:
Thank you for the review and the change! I understand, and it seems right!

hermanschaaf:
Ah, wait a minute, the tests just alerted me to the fact that this information is already part of the `policies` column on the `aws_iam_groups` table, so we probably don't need the additional table? @tockn could you check if the information you require is already present there?

tockn:
I apologize for any confusion. As you said, the information for the attached managed policies already existed in the `aws_iam_group` table's `policies` column. I'll close this PR. Thank you for checking and providing feedback.

hermanschaaf:
@tockn No problem! Thanks for putting in the work, we could probably improve the documentation here so that it's easier to find in the first place. Let us know if there's anything else you need.

Issue #9141, opened 2023-03-17
bug: postgresql unknown type character varying


reedlxm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

as i want to sync from a postgresql to another postgresql,
take an error seams not support character varying type.

### Expected Behavior

i guess this is a bug or something goes wrong. that character varying is a very common type 

### CloudQuery (redacted) config

```
kind: source
spec:
  name: postgresql
  path: cloudquery/postgresql
  version: "v1.0.5"
  destinations: ["backup"]
  spec:
    connection_string: "postgres://"
---
kind: destination
spec:
  name: backup
  path: cloudquery/postgresql
  version: "v3.0.0"
  spec:
    connection_string: "postgres://"
```

### CloudQuery (redacted) logs

Starting migration with tables for: postgresql (v1.0.5) -> [backup (postgresql@v3.0.0)],
Error: failed to sync v1 source postgresql: failed to migrate source postgresql (v1.0.5) on destination backup (postgresql@v3.0.0) : failed to call Migrate: rpc error: code = Unknown desc = failed listing postgres tables: got unknown PostgreSQL type "character varying(150)" for column "" of table "" while trying to convert it to CloudQuery internal schema type. Supported PostgreSQL types are ["real" "macaddr[]" "inet" "macaddr" "inet[]" "float" "double precision" "text" "bytea" "uuid[]" "boolean" "bigint" "jsonb" "cidr[]" "bigint[]" "integer" "uuid" "text[]" "cidr" "timestamp"]

### CloudQuery version

docker leatest

### Pull request (optional)

- [ ] I can submit a pull request

hermanschaaf:
Thanks for reporting this @reedlxm - I think we recently had a similar issue https://github.com/cloudquery/cloudquery/issues/8922, but that fix got released in v1.0.5 that you are already using. I will take a look at this 👀

hermanschaaf:
I was able to reproduce the issue locally by doing the following:

```
cloudquery> create database source;
CREATE DATABASE
Time: 0.256s

cloudquery> \c source
You are now connected to database "source" as user "hermanschaaf"
Time: 0.014s

source> create table test_varchar3 (name varchar(150));
CREATE TABLE
Time: 0.019s

cloudquery> create database backup;
CREATE DATABASE
Time: 0.256s

cloudquery> \c backup
You are now connected to database "backup" as user "hermanschaaf"
Time: 0.014s

[...]

PR #9143, opened 2023-03-17
fix(postgresql): Normalize types in Postgres destination

hermanschaaf:
Fixes https://github.com/cloudquery/cloudquery/issues/9141. This will allow CloudQuery to sync to tables that had their schemas created by some external process, as long as the column can be normalized to a CloudQuery type. This is mostly useful for Postgres -> Postgres sync use cases right now. As we improve the internal type system, we should be able to do away with these training wheels and create the same column types on the destination as we find on the source for such cases.

The type normalization is copied from the [Postgres source](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/postgresql/client/types.go)

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -124,10 +124,7 @@ func (c *Client) listPgTables(ctx context.Context, pluginTables schema.Tables) (
 		if pkName != "" {
 			table.PkConstraintName = pkName
 		}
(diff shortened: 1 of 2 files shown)


candiduslynx (approved):

disq (approved):

erezrokah:
Looks great @hermanschaaf thank you for applying the fix 🚀

PR #9147, opened 2023-03-17
feat(file): Update to filetype v1.6.0

yevgenypats:


hermanschaaf:
@yevgenypats Could we maybe merge https://github.com/cloudquery/cloudquery/pull/9144 instead? It does the same as this one, but also adds support for CSV headers and delimeters by embedding the file format spec like the S3, GCS and Azure Blob destination plugins are doing.

Issue #9142, opened 2023-03-17
feat: Add resource groups to Azure source plugin


hermanschaaf:
### Describe the resource

Azure v1.4.0 supported a Azure Resource Groups table, but it was dropped in v2.0.0+. I believe we should be able to add it back: https://learn.microsoft.com/en-us/rest/api/resources/resource-groups

[Reported by a user on Discord](https://discord.com/channels/872925471417962546/873606591335759872/1086197834027630654)

### Use Case

List Azure resource groups

### Link to API documentation

https://learn.microsoft.com/en-us/rest/api/resources/resource-groups

### Pull request (optional)

- [ ] I can submit a pull request

hermanschaaf:
PR opened in https://github.com/cloudquery/cloudquery/pull/9158

PR #9158, opened 2023-03-17
feat: Add azure_resources_resource_groups table

hermanschaaf:
This adds a new `azure_resources_resource_groups` table alongside the existing `azure_resources_resources` table.

Closes https://github.com/cloudquery/cloudquery/issues/9142


erezrokah (commented):
Thanks for the PR @hermanschaaf, we already list the resource groups during initialization (since we have a resource that multiplexes over those), see https://github.com/cloudquery/cloudquery/blob/49d647730a85ea6fae51e97194ba61c0625d1331/plugins/source/azure/client/client.go#L76

At the moment we only keep the names, but we could keep the whole object and return it in the table resolver instead of doing the listing again, WDYT?

See example in https://github.com/cloudquery/cloudquery/blob/49d647730a85ea6fae51e97194ba61c0625d1331/plugins/source/azure/resources/services/subscription/subscriptions_fetch.go#L14

erezrokah (approved):

plugins/source/azure/client/helpers.go
```diff
@@ -8,7 +8,7 @@ import (
 // this is used to hook ParseResourceGroup and to have easier codegen
 var debug = false
 
-const resourceIDPatternText = `(?i)subscriptions/(.+)/resourceGroups/(.+)/providers/(.+?)/(.+?)/(.+)`
+const resourceIDPatternText = `(?i)subscriptions/(.+)/ResourceGroups/(.+)/providers/(.+?)/(.+?)/(.+)`
```
disq, line 11: Why this change? was it broken before? Tests set `debug` so this code doesn't really run without a real fetch.

Also we could remove the `const` and just have the string inside `regexp.MustCompile` below, it's not used anywhere else anyway.

hermanschaaf: Reverted

PR #9224, opened 2023-03-19
chore: Allocate more memory to Website build

erezrokah:
#### Summary

A test to see if this improves build speed, as the Vercel build container should have 8GB. See https://vercel.com/docs/concepts/limits/overview#build-container-resources


erezrokah:
We can close this, it has no consistent visible impact

Issue #7907, opened 2023-02-10
AWS SSM Resources


jsonpr:
AWS SSM Document Versions
AWS SSM Sessions

PR #9255, opened 2023-03-20
feat(aws): SSM Document Versions, SSM Sessions

disq:
Implements #7907


plugins/source/aws/resources/services/ssm/sessions.go
```diff
@@ -0,0 +1,50 @@
...
+		},
+	}
+}
+
+func fetchSsmSessions(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	c := meta.(*client.Client)
+	svc := c.Services().Ssm
+
+	params := ssm.DescribeSessionsInput{
+		State:   types.SessionStateActive,
```
disq, line 34: Only listing active SSM sessions.

website/tables/aws/aws_ssm_document_versions.md
```diff
@@ -0,0 +1,33 @@
+# Table: aws_ssm_document_versions
+
+This table shows data for Ssm Document Versions.
```
hermanschaaf, line 3: Could we add an exception for SSM here: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/resources/plugin/plugin.go#L16 and then rerun doc generation 🙇

plugins/source/aws/resources/services/ssm/sessions.go
hermanschaaf: I think that's probably okay, but maybe we should add this to the description of the table so that it's mentioned in the documentation?

hermanschaaf (approved):
LGTM, let's just document that we're only listing active sessions if you don't mind updating that

disq:
> LGTM, let's just document that we're only listing active sessions if you don't mind updating that

Good idea, done

disq:
> @disq Have you pushed the commit for the update? I don't see it 🤔

I have but @github

Issue #7921, opened 2023-02-11
AWS AutoScaling Plans and Application Autoscaling (Different from EC2 Autoscaling)


jsonpr:
https://awscli.amazonaws.com/v2/documentation/api/latest/reference/autoscaling-plans/index.html

Plan and Plan Resource

From a naming perspective, we don't have ec2 in the autoscaling resources (https://awscli.amazonaws.com/v2/documentation/api/latest/reference/autoscaling/index.html), which makes sense.  I'm open to naming suggestions or naming this `aws_autoscaling_plans_*`

Application Autoscaling: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/application-autoscaling/index.html

Scalable Targets, Scaling Activities, Scaling Policies, Scheduled Actions

PR #9261, opened 2023-03-20
feat(aws): More ApplicationAutoScaling resources

disq:
Some of #[7921](https://github.com/cloudquery/cloudquery/issues/7921)


plugins/source/aws/resources/services/applicationautoscaling/scalable_targets.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func ScalableTargets() *schema.Table {
+	tableName := "aws_applicationautoscaling_scalable_targets"
+	return &schema.Table{
+		Name:        tableName,
+		Description: `https://docs.aws.amazon.com/autoscaling/application/APIReference/API_ScalableTarget.html`,
+		Resolver:    fetchScalableTargets,
+		Multiplex:   client.ServiceAccountRegionNamespaceMultiplexer(tableName, "application-autoscaling"),
+		Transform:   transformers.TransformWithStruct(&types.ScalableTarget{}, transformers.WithPrimaryKeys("ResourceId")),
```
disq, line 21: `ResourceId` can be bunch of things, some ARN, some not. Refer to the URL in the description for details.

website/tables/aws/aws_applicationautoscaling_scheduled_actions.md
```diff
@@ -0,0 +1,30 @@
+# Table: aws_applicationautoscaling_scheduled_actions
+
+This table shows data for Applicationautoscaling Scheduled Actions.
```
hermanschaaf, line 3: Please add another exception for `Applicationautoscaling` to plugin.go for the docs generation 😄

plugins/source/aws/resources/services/applicationautoscaling/scheduled_actions_mock_test.go
```diff
@@ -0,0 +1,38 @@
...
+	m.EXPECT().DescribeScheduledActions(gomock.Any(), gomock.Any(), gomock.Any()).Return(
+		output,
+		nil,
+	)
+
+	return services
+}
+
+func TestScheduledActions(t *testing.T) {
+	client.AllNamespaces = []string{"test-namespace"} // Just one
```
hermanschaaf, line 35: Is there any way we could avoid changing this? Maybe it should be a property on the client instead.

As long as `AllNamespaces` is only used by this package and all the tests here set it to the same value, it should be okay, but as soon as one of those assumptions change we could have a fairly hard-to-find bug in our testing.

I see it was like this in the code before already, so consider this comment non-blocking.

website/tables/aws/aws_applicationautoscaling_scheduled_actions.md
disq: oof it reindented everything so I added it to the very top

hermanschaaf (approved):

Issue #7292, opened 2023-01-27
Test --cq-dir behavior


hermanschaaf:
We should test that when `--cq-dir` is set, it is respected by the plugin download process. This is an important feature for running multiple cloudquery instances in parallel from the same directory.

erezrokah:
We use it in our existing tests https://github.com/cloudquery/cloudquery/blob/b7a81763d61b6eaec6cad81636647afe85ac4cd9/cli/cmd/sync_test.go#L22 so maybe it would be easy to add an assertion there

PR #9262, opened 2023-03-20
chore(cli-testing): Test that --cq-dir parameter is used when caching plugins

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/7292


cli/cmd/sync_test.go
```diff
@@ -69,6 +75,27 @@ func TestSync(t *testing.T) {
...
+			if tc.wantDestinationPluginCache {
+				// check that destination plugin was downloaded to the cache using --cq-dir
+				p := path.Join(cqDir, "plugins", "destination")
+				files, err := os.ReadDir(p)
+				if err != nil {
+					t.Fatalf("failed to read cache directory %v: %v", p, err)
+				}
+				require.NotEmpty(t, files, "destination plugin not downloaded to cache")
+			}
+
```
erezrokah, line 98:
```suggestion
```

cli/cmd/sync_test.go
```diff
@@ -69,6 +75,27 @@ func TestSync(t *testing.T) {
 			} else {
 				require.Contains(t, commandError.Error(), tc.err)
 			}
+
+			if tc.wantSourcePluginCache {
+				// check that source plugin was downloaded to the cache using --cq-dir
+				p := path.Join(cqDir, "plugins", "source")
+				files, err := os.ReadDir(p)
```
erezrokah, line 82: Non blocking but we could use `require.FileExists` to test that the files in the directory match what we expect.
There's also `DirExists` but it doesn't check the content of a directory

erezrokah (approved):
Looks good. Some non blocking comments

cli/cmd/sync_test.go
hermanschaaf: Yeah, this will be a bit more work, because we'd have to parse the config to extract the version (the directory gets named that way). Not impossible, but I think it's probably not necessary for now, we just want to check that the correct cache directory got used in general.

erezrokah: Ah, yes but we could do a separate test with a dedicate config we know the names for

PR #9264, opened 2023-03-20
fix(aws): Resolve IoT certificates in parallel instead of serially

bbernays:
#### Summary

While investigating an issue, I refactored this resource to use our latest guidelines and style.

This pr has no functional differences


erezrokah (approved):
Looks good but I think this should be a fix called `fix: Resolve IOT certificates in parallel instead of serially` due to the `PreResourceResolver` change

hermanschaaf (approved):

Issue #9287, opened 2023-03-21
feat: Add identitystore Member Id to aws_identitystore_group_memberships in aws source plugin


castaples:
### Describe the resource

The identity store tables include users, groups and group memberships but this table (aws_identitystore_group_memberships) is missing the member id, so its impossible to link users to groups. See [ListGroupMemberships](https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_ListGroupMemberships.html#API_ListGroupMemberships_ResponseSyntax)

I was unsure if this is a bug or just not been implemented, so raised as a feature request assuming it was the latter.

### Use Case

The identity store tables have users, groups and group memberships but this table (aws_identitystore_group_memberships) is missing the member id, so its impossible to link users to groups.

### Link to API documentation

https://docs.aws.amazon.com/singlesignon/latest/IdentityStoreAPIReference/API_ListGroupMemberships.html#API_ListGroupMemberships_ResponseSyntax

### Pull request (optional)

- [ ] I can submit a pull request

hermanschaaf:
Hi @castaples, thanks for reporting this! It indeed looks like an omission for the `aws_identitystore_group_memberships` table. I'll see if I can quickly add the missing column

hermanschaaf:
This should be fixed by https://github.com/cloudquery/cloudquery/pull/9297. We'll aim to merge and release it some time today

PR #9297, opened 2023-03-21
feat: Add `member_id` to `aws_identitystore_group_memberships`

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/9287

This column was missed previously because it is implemented as an interface type in the AWS SDK. There is only one valid struct that implements this interface right now.


erezrokah (approved):

disq (approved):
Apart from PK being missing (probably `membership_id`, or all 3, or all 4?) LGTM

PR #9347, opened 2023-03-22
chore(googleanalytics): Add title transformer (currently not used)

candiduslynx:
Added the title transformer so that we won't forget it if we ever add prebuilt report tables.

### DIFF
--- a/plugins/source/googleanalytics/go.mod
+++ b/plugins/source/googleanalytics/go.mod
@@ -5,9 +5,9 @@ go 1.19
 require (
 	github.com/cloudquery/plugin-sdk v1.44.0
 	github.com/rs/zerolog v1.29.0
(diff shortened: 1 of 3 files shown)


candiduslynx:
> Hi @candiduslynx, I'm not sure what this PR solves and what's the reason for the `go.mod` and `go.sum` changes.
@erezrokah currently this code isn't doing much, but if we add the prebuilt tables then the titles will say `Googeanalytics X Y` or something like that.
The deps bump was required to have the title transformer option.

If we don't want to add this code in advance, this can be safely closed.

hermanschaaf:
I think let's close this for now; the problem will become apparent as soon as the docs are generated for the new tables, at which point we can add this bit of code (or override the titles manually)

Issue #8153, opened 2023-02-16
feat(azure): Add Virtual Machine Assess Patches


erezrokah:
Requested on Discord https://discord.com/channels/872925471417962546/974029056070795284/1075623719281639494

API - https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/assess-patches?tabs=HTTP

PR #9353, opened 2023-03-22
feat(azure): Add `azure_compute_virtual_machine_patch_assessments`

disq:
Implements #8153

First time using `PollUntilDone`

This looks to be running a brand new assessment each time (instead of reading results of last one) and can take a long time.

BEGIN_COMMIT_OVERRIDE
feat(azure-resources): Add Compute VM Patch Assessments table `azure_compute_virtual_machine_patch_assessments`. This might be slow to sync on some accounts and can be skipped if needed via `skip_tables: ["azure_compute_virtual_machine_patch_assessments"]`
END_COMMIT_OVERRIDE


website/tables/azure/azure_compute_virtual_machine_assess_patches.md
```diff
@@ -0,0 +1,29 @@
+# Table: azure_compute_virtual_machine_assess_patches
+
+This table shows data for Azure Compute Virtual Machine Assess Patches.
+
+https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/assess-patches?tabs=HTTP#virtualmachineassesspatchesresult
+
+The primary key for this table is **assessment_activity_id**.
```
disq, line 7: Not sure about this

erezrokah (approved):
Looks good, but do we know how much time it takes to poll?
Maybe add a change log note this is a "slow" resource (if true)

disq:
> Looks good, but do we know how much time it takes to poll? Maybe add a change log note this is a "slow" resource (if true)

I'll create a windows instance and see what happens

plugins/source/azure/resources/services/compute/virtual_machine_assess_patches.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+func virtualMachineAssessPatches() *schema.Table {
+	tableName := `azure_compute_virtual_machine_assess_patches`
+	return &schema.Table{
+		Name:     tableName,
+		Resolver: fetchVirtualMachineAssessPatches,
+		Description: `https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/assess-patches?tabs=HTTP#virtualmachineassesspatchesresult.
+
+This will begin patch assessments on available virtual machines and can take long to complete.
```
disq, line 21: Took a few (5?) seconds for a single VM, but can delay things a lot since it's done separately for each resource.

plugins/source/azure/resources/services/compute/virtual_machine_assess_patches.go
```diff
@@ -0,0 +1,72 @@
...
+
+Not available for all VMs. More at https://learn.microsoft.com/en-us/azure/virtual-machines/automatic-vm-guest-patching#requirements-for-enabling-automatic-vm-guest-patching
+`,
+		Multiplex: client.SubscriptionMultiplexRegisteredNamespace(tableName, client.Namespacemicrosoft_compute),
+		Transform: transformers.TransformWithStruct(&armcompute.VirtualMachineAssessPatchesResult{}, transformers.WithPrimaryKeys("AssessmentActivityID")),
+		Columns:   schema.ColumnList{client.SubscriptionID},
```
disq, line 27: Again, no reference to the actual VM id. `assessment_activity_id` is a uuid and not a resource id, the only way to match/find out which VM it belongs to is to use `_cq_parent_id` unless we add a `ParentColumnResolver("id")` here.

plugins/source/azure/resources/services/compute/virtual_machine_assess_patches.go
erezrokah: Yeah, so maybe we should add a commit message override similar to https://github.com/cloudquery/cloudquery/pull/8726 (see `BEGIN_COMMIT_OVERRIDE`)

plugins/source/azure/resources/services/compute/virtual_machine_assess_patches.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4"
+	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func virtualMachineAssessPatches() *schema.Table {
+	tableName := `azure_compute_virtual_machine_assess_patches`
```
disq, line 15: Or maybe `azure_compute_virtual_machine_patch_assessments`

Issue #9269, opened 2023-03-21
Investigation request: Investigate Azure plugin high memory usage


erezrokah:
We got reports of syncing on the `azure_compute_skus` compute table failing even with 8GB of memory, with reduced concurrency of `1000`.

We should investigate to see if this can be improved. One thing to note is the Azure client pre-initializes some data that is needed for multiplexing like `SubscriptionsObjects`, `ResourceGroups` and `registeredNamespaces`.
Example config:

```yaml
kind: source
spec:
  name: "azure"
  path: "cloudquery/azure"
  version: "v6.0.0"
  destinations: ["postgres"]
  tables: ["azure_compute_skus"]
```

erezrokah:
### Update 2

Using the second authentication method described here https://www.cloudquery.io/docs/plugins/sources/azure/overview#authentication-with-environment-variables seems to reduce memory usage significantly and allows the Azure plugin to finishing syncing with the default concurrency on our internal account. To people subscribed to this issue, it would be great if you could test it as well

tomtom215:
Hi @erezrokah our test run with all subscriptions but with excluding the SKU table is still running. So far memory consumption is much much lower (max so far is ~5.5GB). One additional table that might also be worth a check is "azure_policy_definitions". It was no where near the memory jump as the SKU table, but it was by far the largest ram spike in our test compared to the other tables, however this may be expected behaviour for that table. We'll keep you posted when our test completes

erezrokah:
> Hi @erezrokah our test run with all subscriptions but with excluding the SKU table is still running. So far memory consumption is much much lower (max so far is ~5.5GB). One additional table that might also be worth a check is "azure_policy_definitions". It was no where near the memory jump as the SKU table, but it was by far the largest ram spike in our test compared to the other tables, however this may be expected behaviour for that table. We'll keep you posted when our test completes

Thanks for testing @tomtom215! To confirm the only change you did to reduce memory consumption is skip the SKUs table? That's very helpful and I'll look into `azure_policy_definitions` as well

tomtom215:
Hi @erezrokah  - correct, the only change was to skip the SKUs table. I sent over our findings from the run separately as promised

erezrokah:
[...]

PR #9363, opened 2023-03-22
feat(azure-auth): Log `DefaultAzureCredential` credentials errors

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/9269.
When we create the Azure credentials via `NewDefaultAzureCredential` it creates a credential chain, currently env variables -> MS Identity -> CLI auth. CLI Auth is used last and can create performance and memory issues in Azure as it spawns the Azure CLI on each token request.

This PR logs any credentials creation errors by listening to the Azure Go SDK log. Those are only reported via the logger, so we need add a log listener:
https://github.com/Azure/azure-sdk-for-go/blob/97d5986a79730c2ae5c0c96ffec06293b75951d4/sdk/azidentity/default_azure_credential.go#L112

We log them as `Info` as we can't know which credentials the user intended to use


plugins/source/azure/client/client.go
```diff
@@ -154,6 +167,8 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source, _ source.Op
 		return nil, err
 	}
 
+	log.SetListener(nil)
```
disq, line 170: Removing the listener would also ignore any EventAuthentication error occuring after the initial setup, like a fail to refresh token etc.

erezrokah: Those are returned as errors from functions. We didn't set the listener before so we're not missing on anything here

disq (approved):

Issue #9366, opened 2023-03-22
bug: Azure `azure_subscription_subscription_locations` always fails with an error


erezrokah:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Trying to sync `azure_subscription_subscription_locations` always fails with an error:

```
GET https://management.azure.com/subscriptions//subscriptions/<id>/locations
--------------------------------------------------------------------------------
RESPONSE 400: 400 Bad Request
ERROR CODE: InvalidSubscriptionId
--------------------------------------------------------------------------------
{
  \"error\": {
    \"code\": \"InvalidSubscriptionId\",
    \"message\": \"The provided subscription identifier 'subscriptions' is malformed or invalid.\"
  }
}
--------------------------------------------------------------------------------

```

### Expected Behavior

Syncing the table should not fail

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: azure
  version: "v6.0.0"
  destinations: [postgresql]
  path: cloudquery/azure
  tables: [azure_subscription_subscription_locations]
---
kind: destination
spec:
  name: postgresql
  path: cloudquery/postgresql
  version: v3.0.2
  spec:
    connection_string: postgresql://postgres:pass@localhost:5432/postgres
```

### Steps To Reproduce

1. Run a sync with the supplied configuration
2. See error

### CloudQuery (redacted) logs

Provider above

### CloudQuery version

cloudquery version 2.5.2

### Pull request (optional)

- [X] I can submit a pull request

PR #9369, opened 2023-03-22
fix(azure-resources): Pass correct subscription ID when listing subscription locations

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/9366

### DIFF
--- a/plugins/source/azure/resources/services/subscription/subscription_mock_test.go
+++ b/plugins/source/azure/resources/services/subscription/subscription_mock_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/cloudquery/cloudquery/plugins/source/azure/client"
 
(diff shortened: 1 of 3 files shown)


plugins/source/azure/resources/services/subscription/subscription_mock_test.go
```diff
@@ -20,6 +21,8 @@ func createSubscriptions(router *mux.Router) error {
 
 	emptyStr := ""
 	item.NextLink = &emptyStr
+	item.Value[0].ID = to.Ptr("/subscriptions/sub-id")
```
erezrokah, line 24: The reason the tests didn't catch this is because we mock all strings to the same value of `test string` so it doesn't capture the different format between ID and SubscriptionID

hermanschaaf (approved):

PR #9375, opened 2023-03-23
fix(clickhouse): table is already exist

easayliu:
clickhouse migrate_mode: forced error,"table is already exist".
add alicloud daily bill table.

### DIFF
--- a/plugins/destination/clickhouse/queries/tables.go
+++ b/plugins/destination/clickhouse/queries/tables.go
@@ -19,7 +19,7 @@ func sortKeys(table *schema.Table) []string {
 func CreateTable(table *schema.Table, cluster string, engine *Engine) string {
 	normalized := normalizeTable(table)
 	strBuilder := strings.Builder{}
-	strBuilder.WriteString("CREATE TABLE ")
+	strBuilder.WriteString("CREATE TABLE IF NOT EXISTS")
 	strBuilder.WriteString(tableNamePart(normalized.Name, cluster))
 	strBuilder.WriteString(" (\n")
 	for i, col := range normalized.Columns {
--- a/plugins/destination/clickhouse/queries/tables.go
+++ b/plugins/destination/clickhouse/queries/tables.go
@@ -19,7 +19,7 @@ func sortKeys(table *schema.Table) []string {
 func CreateTable(table *schema.Table, cluster string, engine *Engine) string {
 	normalized := normalizeTable(table)
 	strBuilder := strings.Builder{}
-	strBuilder.WriteString("CREATE TABLE IF NOT EXISTS")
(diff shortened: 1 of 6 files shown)


candiduslynx:
Hi @easayliu!
Thanks for submitting the PR.
Could you please split the changes into 2 separate PRs? One for ClickHouse & the other for AliCloud?
This way the changes will perfectly visible in the change logs for the proper plugins.

plugins/destination/clickhouse/queries/tables.go
```diff
@@ -19,7 +19,7 @@ func sortKeys(table *schema.Table) []string {
 func CreateTable(table *schema.Table, cluster string, engine *Engine) string {
 	normalized := normalizeTable(table)
 	strBuilder := strings.Builder{}
-	strBuilder.WriteString("CREATE TABLE ")
+	strBuilder.WriteString("CREATE TABLE IF NOT EXISTS")
```
candiduslynx, line 22: Context: [migrate code](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/clickhouse/client/migrate.go#L115-L140).

This shouldn't be required as the migrate mode should drop the table.
I suspect, the real issue here is missing `ON CLUSTER` in the `DROP` statement. See #9377

plugins/source/alicloud/resources/services/bss/bill_des.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func Bill_Des() *schema.Table {
+	return &schema.Table{
+		Name:      "alicloud_bss_bill_des",
+		Resolver:  fetchBssDescribeinstanceBill,
+		Multiplex: client.AccountMultiplex,
+		Transform: transformers.TransformWithStruct(
+			&BillDesModel{},
+			//transformers.WithPrimaryKeys(
```
candiduslynx, line 16: Primary key should be added

candiduslynx (changes requested):

plugins/destination/clickhouse/queries/tables.go
easayliu: AGREE！ This is better

plugins/destination/clickhouse/queries/tables.go
```diff
@@ -19,7 +19,7 @@ func sortKeys(table *schema.Table) []string {
 func CreateTable(table *schema.Table, cluster string, engine *Engine) string {
 	normalized := normalizeTable(table)
 	strBuilder := strings.Builder{}
-	strBuilder.WriteString("CREATE TABLE ")
+	strBuilder.WriteString("CREATE TABLE")
```
candiduslynx, line 22: space is required as `tableNamePart` produces output without leading space

PR #9376, opened 2023-03-23
feat(aws): Add Support For DynamoDB Resources

bbernays:


plugins/source/aws/resources/services/dynamodb/backups.go
```diff
@@ -0,0 +1,71 @@
...
+		Transform:           transformers.TransformWithStruct(&types.BackupDescription{}),
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(false),
+			client.DefaultRegionColumn(false),
+			{
+				Name:     "arn",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("BackupDetails.BackupArn"),
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
```
candiduslynx, line 31: I wonder, do we want to start labeling the PK columns `NotNull` explicitly, or do a transition at some point?
cc: @yevgenypats

plugins/source/aws/resources/services/dynamodb/global_tables.go
```diff
@@ -0,0 +1,95 @@
...
+
+func fetchGlobalTables(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	c := meta.(*client.Client)
+	svc := c.Services().Dynamodb
+
+	config := dynamodb.ListGlobalTablesInput{
+		RegionName: aws.String(c.Region),
+	}
+	for {
+		output, err := svc.ListGlobalTables(ctx, &config)
```
candiduslynx, line 51: Isn't it [legacy](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_ListGlobalTables.html)?

Even more, I'd argue that we need to use [TableDescription.GlobalTableVersion](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TableDescription.html) field from [DescribeTable](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DescribeTable.html) output to determine whether we need to obtain global info for the table via [API_DescribeTableReplicaAutoScaling](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/DescribeTableReplicaAutoScaling.html) call.

So, there should be `aws_dynamodb_table_replica_auto_scaling` relation for `aws_dynamodb_tables` that would be populated for tables with `GlobalTableVersion != ""` (better yet, for `GlobalTableVersion == "2019.11.21"`).

candiduslynx (changes requested):

plugins/source/aws/resources/services/dynamodb/global_tables.go
bbernays: While I don't disagree we need to add that data, I think it falls outside the scope of this pr as the old GlobalTable is still a valid resource that users use

plugins/source/aws/resources/services/dynamodb/backups.go
candiduslynx: To set NotNull explicitly for all PKs

plugins/source/aws/resources/services/dynamodb/global_tables.go
candiduslynx: I suggest adding the same notice as AWS has (about this being only for legacy API) in the description

plugins/source/aws/resources/services/dynamodb/backups.go
bbernays: Why wouldn't we just do it in the SDK?

candiduslynx (approved):

PR #9380, opened 2023-03-23
feat(alicloud): Add daily bill table

easayliu:
#### Summary
add daily bill table


plugins/source/alicloud/resources/services/bss/bill_des_fetch.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	for _, billingCycle := range billingCycles {
+		billingDates := getDesBillingDates(billingCycle)
+		for _, billingDate := range billingDates {
+			request := bssopenapi.CreateQueryInstanceBillRequest()
+			log.Info().Str("data", billingDate)
+			request.BillingCycle = billingCycle
+			pageNum := 1
+			total := 0
+			DesmaxLimit := 100
```
candiduslynx, line 33:
```suggestion
			desmaxLimit := 100
```

plugins/source/alicloud/resources/services/bss/bill_des_fetch.go
```diff
@@ -0,0 +1,117 @@
...
+			request.PageNum = requests.NewInteger(pageNum)
+			request.BillingDate = billingDate
+			request.PageSize = requests.NewInteger(DesmaxLimit)
```
candiduslynx, line 36:
```suggestion
			request.PageSize = requests.NewInteger(desmaxLimit)
```

plugins/source/alicloud/resources/services/bss/bill_des_fetch.go
```diff
@@ -0,0 +1,117 @@
...
+						OutstandingAmount:     item.OutstandingAmount,
+						PipCode:               item.PipCode,
+						PretaxAmount:          item.PretaxAmount,
+						OwnerID:               item.OwnerID,
+						RecordID:              item.RecordID,
+						ResourceGroup:         item.ResourceGroup,
+						InstanceID:            item.InstanceID,
+						CashAmount:            item.CashAmount,
+					}
+
```
candiduslynx, line 83:
```suggestion
```

plugins/source/alicloud/resources/services/bss/bill_des.go
```diff
@@ -0,0 +1,57 @@
...
+
+func Bill_Des() *schema.Table {
+	return &schema.Table{
+		Name:      "alicloud_bss_bill_des",
+		Resolver:  fetchBssDescribeinstanceBill,
+		Multiplex: client.AccountMultiplex,
+		Transform: transformers.TransformWithStruct(
+			&BillDesModel{},
+			transformers.WithPrimaryKeys(
+				"BillingCycle",
```
candiduslynx, line 17: As you unwrap items from response, several of those will have the same `BillingCycle`.
Could you check if there's a better PK to be used?

plugins/source/alicloud/resources/services/bss/bill_des.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"github.com/cloudquery/cloudquery/plugins/source/alicloud/client"
+	"github.com/cloudquery/plugin-sdk/schema"
+	"github.com/cloudquery/plugin-sdk/transformers"
+)
+
+func Bill_Des() *schema.Table {
+	return &schema.Table{
+		Name:      "alicloud_bss_bill_des",
```
candiduslynx, line 11: What does `des` stand for?

candiduslynx (changes requested):

candiduslynx (changes requested):

candiduslynx (approved):

plugins/source/alicloud/resources/services/bss/bill_details_fetch.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+	return nil
+}
+
+func getDetailsBillingCycles() []string {
+	var months []string
+	curMonth := time.Now().Format("2006-01")
+	months = append(months, curMonth)
+	lastMonth := time.Now().AddDate(0, 0, -5).Format("2006-01")
```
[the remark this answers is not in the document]
hermanschaaf: I think this should match what we have in `bill_fetch.go`, which is using `(0, 0, -5)`: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/alicloud/resources/services/bss/bill_fetch.go#L93

Otherwise, bills and bill_details could return results for different time periods. I think the idea was to fetch the previous month for the first 5 days of a new month, when the user is not requesting historical data. It's fine if we want to change this to always fetch the previous month, but let's then also update `bill_fetch.go`. What do you think @easayliu ?

hermanschaaf:
I pushed a small change to update the title and description of the table, but otherwise looks good @easayliu 👍 Just one question left about whether we should use `(0, -1, 0)` or `(0, 0, -5)` when getting the last month. I think you have more context here so will let you decide what is best.

easayliu:
> I pushed a small change to update the title and description of the table, but otherwise looks good @easayliu 👍 Just one question left about whether we should use `(0, -1, 0)` or `(0, 0, -5)` when getting the last month. I think you have more context here so will let you decide what is best.

becauce The bill is updated around the 5th of each month,so i use `(0,0,-5)`. But now ,it is update recently, maybe it`s not a necessary。

candiduslynx (approved):

plugins/source/alicloud/resources/services/bss/bill_details_fetch.go
```diff
@@ -0,0 +1,113 @@
...
+	if lastMonth != curMonth {
+		months = append(months, lastMonth)
+	}
+	return months
+}
+
+func getDetailsBillingDates(month string) []string {
+	var days []string
+	firstDate := month + "-01"
+	first, _ := time.ParseInLocation("2006-01-02", firstDate, time.Local)
```
candiduslynx, line 108: Any specific reason to use time.ParseInLocation over time.Parse?

hermanschaaf (approved):

PR #9381, opened 2023-03-23
fix(gcp): Fix typo and condition in the CIS GCP Policy 6.3.6

aruneko:
#### Summary
The title, description and remediation of CIS GCP Policy v1.2.0 Section 6.3.6 has an error. It describes 3625 trace flag should be off, but in fact it should be on. The statement is fixed in CIS for GCP v1.3.0. In my opinion, for convenience, this fix should be backported in v1.2.0.


hermanschaaf:
I think this makes sense, but we should probably add a comment about why this differs from the policy document, otherwise we might get another PR by someone else in a few weeks changing it back :)

aruneko:
> otherwise we might get another PR by someone else in a few weeks changing it back :)

Great viewpoint!

I've add some comments for this changes ;)

hermanschaaf (approved):

hermanschaaf:
Thanks for the fix @aruneko !

PR #9382, opened 2023-03-23
feat(azure): Discover resource groups and namespaces in parallel

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/9269. I'm trying to reproduce https://github.com/cloudquery/cloudquery/issues/9269 with many subscriptions.
While doing so (I simulated a run on 100 subscriptions equivalent to our playground subscription) I noticed it takes a long time for the plugin to start syncing when I increase the subscriptions count.
The reason is that we discover resource groups and registered namespaces in serial.

This PR updates the code to do it in parallel. It improves the discovery time on 100 subscriptions from ~7m to ~30s.


plugins/source/azure/client/client.go
```diff
@@ -65,49 +67,89 @@ func (c *Client) discoverSubscriptions(ctx context.Context) error {
...
+
+	return groups, nil
+}
+
+func (c *Client) getRegisteredProvidersForSubscription(ctx context.Context, subscriptionId string) ([]*armresources.Provider, error) {
+	providerClient, err := armresources.NewProvidersClient(subscriptionId, c.Creds, c.Options)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create provider client: %w", err)
+	}
+	providers := make([]*armresources.Provider, 0)
```
disq, line 96:
```suggestion
	var providers []*armresources.Provider
```

plugins/source/azure/client/client.go
```diff
@@ -65,49 +67,89 @@ func (c *Client) discoverSubscriptions(ctx context.Context) error {
...
-					c.registeredNamespaces[subID][strings.ToLower(*p.Namespace)] = true
-				}
+
+			namespacesLock.Lock()
+			c.registeredNamespaces[subID] = make(map[string]bool)
+			for _, p := range providers {
+				c.registeredNamespaces[subID][strings.ToLower(*p.Namespace)] = true
 			}
-		}
+			namespacesLock.Unlock()
```
disq, line 147: I would move this to a defer statement just after the `.Lock()` call, to make sure it runs even after a panic

disq (approved):
LGTM with nits

Issue #9391, opened 2023-03-23
performance/bug: Dissallow hashstructure


yevgenypats:
we should disallow the use of [hashstructure](https://github.com/mitchellh/hashstructure) anywhere in our codebase for the following reasons:
1) It is unmaintained.
2) it has bad performance as it uses reflection. 
3) it gives inconsistent hash value that are depending on the go runtime instead of using consistent hashing of our type system.

PR #9393, opened 2023-03-23
fix(azure)!: Change `azure_compute_skus` PK to calc hash only on `locations` & `capabilities`

candiduslynx:


plugins/source/azure/resources/services/compute/skus.go
```diff
@@ -19,21 +19,27 @@ func SKUs() *schema.Table {
...
+				Name:            "locations_hash",
 				Type:            schema.TypeString,
-				Resolver:        calcEntryHash,
+				Resolver:        calcLocationsHash,
+				CreationOptions: schema.ColumnCreationOptions{PrimaryKey: true},
+			},
+			schema.Column{
+				Name:            "capabilities_hash",
+				Type:            schema.TypeString,
+				Resolver:        calcCapabilitiesHash,
```
bbernays, line 35: Is there documentation that justifies this change? And if this is about performance can we benchmark the memory savings and or CPU savings?

candiduslynx: No, I just had to inspect the `az vm list-skus` output.
Basically, there are a lot of duplicates around locations and different capabilities. However, this `jq` transformation lets us see the same amount of unique lines:
```bash
az vm list-skus \
  jq -c '.[] | {name, resourceType, locations, capabilities} | (.locations |= sort) | if .capabilities then . else . + {capabilities:[]} end | (.capabilities? |= sort)' \
  sort | wc -l
```

So, this is totally empirical change.

bbernays: I don't think we should be introducing breaking changes unless there is a specific benefit for users, and unless we can benchmark the memory improvements I don't think this change rises to that level

erezrokah: The work to investigate memory issues on Azure is still on going, see https://github.com/cloudquery/cloudquery/issues/9269.

So far removing `hashstructure` did not show any memory or performance improvements in my internal tests, so I'd put this PR hold for now.

We might want to remove `hashstructure` for other reasons to do it (1. and 3. described in https://github.com/cloudquery/cloudquery/issues/9391#issue-1637475685), but it shouldn't be tied to a specific table.

PR #9416, opened 2023-03-26
fix(release-summary): Only run the summary workflow if the PR has 1 commits

erezrokah:
#### Summary

https://github.com/cloudquery/cloudquery/pull/9260 broken the logic that detects if we already added a summary to a PR to avoid a recursive workflow.

This PR fixes the logic and makes it more robust by looking at number of commits instead of the content of the PR


erezrokah:
I'll force merge this as that's the only way to test it's working 100%

PR #9461, opened 2023-03-27
fix(github): Add pagination for dependabot resources

h1zhao:
add pagination for github dependabot source plugin


candiduslynx (approved):

plugins/source/github/resources/services/organizations/alerts.go
```diff
@@ -20,13 +20,19 @@ func alerts() *schema.Table {
 
 func fetchAlerts(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
 	c := meta.(*client.Client)
+	opts := &github.ListAlertsOptions{ListCursorOptions: github.ListCursorOptions{PerPage: 99}}
```
erezrokah, line 23: Is there a reason to chose 99? Looks like the API supports 100:
https://docs.github.com/en/rest/dependabot/alerts?apiVersion=2022-11-28#list-dependabot-alerts-for-an-organization

image

Same for the other table

erezrokah (commented):
Thanks for the PR @h1zhao and for the review @candiduslynx, added another comment for your consideration

plugins/source/github/resources/services/organizations/alerts.go
erezrokah: Yeah, let's do 100 I don't think we need to be concerned with hitting the limit as 100 is supported

PR #9465, opened 2023-03-28
fix(stripe): Expand fields on prices

tothandras:
#### Summary

Use the [expand parameter](https://stripe.com/docs/api/expanding_objects) to include fields on the `stripe_prices` table.
Other tables might also be affected, it might be worth considering exposing a config option.

### DIFF
--- a/plugins/source/stripe/codegen/templates/list.go.tpl
+++ b/plugins/source/stripe/codegen/templates/list.go.tpl
@@ -2,7 +2,7 @@ package {{.Service}}
 
 import (
 	"context"
-{{if .StateParamName}}
+{{- if .StateParamName}}
 	"fmt"
 	"strconv"
 {{end}}
@@ -68,6 +68,12 @@ func fetch{{.TableName | ToPascal}}(ctx context.Context, meta schema.ClientMeta,
 		lp := &stripe.{{.TableName | ToPascal | Singularize}}ListParams{
 {{.ListParams}}
 		}
+
+		{{- if len .ExpandFields}}
+		{{- range .ExpandFields}}
+		lp.AddExpand("{{.}}")
+		{{- end}}
+		{{- end}}
 
 {{if .StateParamName}}
 		const key = "{{.TableName}}"
--- a/plugins/source/stripe/codegen/templates/list.go.tpl
+++ b/plugins/source/stripe/codegen/templates/list.go.tpl
@@ -69,10 +69,8 @@ func fetch{{.TableName | ToPascal}}(ctx context.Context, meta schema.ClientMeta,
 {{.ListParams}}
 		}
 
-		{{- if len .ExpandFields}}
 		{{- range .ExpandFields}}
 		lp.AddExpand("{{.}}")
-		{{- end}}
 		{{- end}}
 
 {{if .StateParamName}}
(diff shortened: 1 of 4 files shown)


disq:
Thanks for the contribution! Since most of Stripe resource code is auto generated, the changes would get overwritten once `make gen-code` is ran (for example after a library update). But this pattern seems generic enough to include in the "codegen" section of the Stripe plugin, which is responsibe for generating the actual client code. I can help you with the changes if you like, or we can handle it internally within the CloudQuery team and it would get implemented fairly quickly. If you want to give it a try, the changes to the `codegen` package would be:

1. Add a new `[]string` field to the `Resource` struct (under `codegen/recipes/`) - the name of the new field should be self explanatory, something like `ExpandFields` could work
2. In `all.go` where all resource types are defined, find the element that defines `stripe.Price` and define the expanded fieldnames in the newly added field
3. In the template file used for listing resources, which is `codegen/templates/list.go.tpl`, add a `range` statement (see the [Go templating docs](https://pkg.go.dev/text/template) if unfamiliar) there to iterate over the newly added fields' values and call `lp.AddExpand` on them.
4. Run `make gen-code` and verify the generated code looks like intended. To avoid whitespace changes on other resources that use the same template, there's a [minus sign trick](https://pkg.go.dev/text/template#hdr-Text_and_spaces) which should help.

Let me know your thoughts and feel free to ask if I missed something! :-)

plugins/source/stripe/codegen/templates/list.go.tpl
```diff
@@ -69,6 +69,12 @@ func fetch{{.TableName | ToPascal}}(ctx context.Context, meta schema.ClientMeta,
 {{.ListParams}}
 		}
 
+		{{- if len .ExpandFields}}
```
disq, line 72:
```suggestion
```
This should not be needed as `range` wouldn't iterate of length is 0

plugins/source/stripe/codegen/templates/list.go.tpl
```diff
@@ -69,6 +69,12 @@ func fetch{{.TableName | ToPascal}}(ctx context.Context, meta schema.ClientMeta,
...
+		{{- range .ExpandFields}}
+		lp.AddExpand("{{.}}")
+		{{- end}}
+		{{- end}}
```
disq, line 76:
```suggestion
```
Matching end block for the removed `len` line above

disq (commented):
Looks good! Yeah, `ListParams` wouldn't be ideal, due to the way the struct is set up in the stripe client library (it's in an embedded struct and also needs string pointers...)

I've caught a single nit in the template, the `len` statement isn't necessary as without that the template still runs and generates the exact same code. If you can get it sorted (and given the tests pass) we can merge this and release an updated plugin soon.

tothandras:
Should I also check the other stripe resources and include their expanded fields?

disq (approved):
> Should I also check the other stripe resources and include their expanded fields?

I think it would be out of scope for this PR, but if you feel there are resources/fields that would benefit from the addition, feel free to add them.

tothandras:
> > Should I also check the other stripe resources and include their expanded fields?
> 
> I think it would be out of scope for this PR, but if you feel there are resources/fields that would benefit from the addition, feel free to add them.

Thanks, I'll open other PR's once we run into others.

PR #9467, opened 2023-03-28
chore(azure): Update major library versions

candiduslynx:
Perform initial major version update as sometimes doing first major bump via #760 gets stuck.
See:
- [ ] #9469
- [ ] #9470
- [ ] #9471


erezrokah (changes requested):
Hi @candiduslynx, is there a reason to do this update outside of our regular monthly 
schedule? It should happen in a few days in the beginning of the month.

Also, I'd rather if we work of each bot PR and fix it manually.
That should allow us to:
1. Do each change gradually as each PR is mapped to a single update
2. We get better change log entries as each PR is mapped to a corresponding change log (for example you classified this PR as `chore` which will not appear in the change log at all or trigger a release)
3. Each PR's description links to the module change log, so we can see what are the breaking changes and decide if they impact us.

Issue #7922, opened 2023-02-11
AWS ELB Listener Rules


jsonpr:
https://docs.aws.amazon.com/cli/latest/reference/elbv2/describe-rules.html

PR #9484, opened 2023-03-28
feat(aws): ELBv2 Listener Rules

disq:
Implements #7922 

First two commits are general clean up


erezrokah (approved):

website/tables/aws/aws_elbv2_listener_rules.md
```diff
@@ -0,0 +1,28 @@
...
+|_cq_id|UUID|
+|_cq_parent_id|UUID|
+|account_id|String|
+|region|String|
+|listener_arn|String|
+|actions|JSON|
+|conditions|JSON|
+|is_default|Bool|
+|priority|String|
+|rule_arn (PK)|String|
```
hermanschaaf, line 28: I think we should add an `arn` column (alias of `rule_arn`) and make that the PK instead, for consistency and so that the AWS resources view will work for this (it assumes the resource ARN is in a column called `arn`).

disq: good catch I'll open a follow up PR

PR #9495, opened 2023-03-29
feat(clickhouse)!: Read only plain `ca_cert`

candiduslynx:
BEGIN_COMMIT_OVERRIDE
feat(clickhouse)!: Read only plain `ca_cert` value

BREAKING-CHANGE: Stop reading `ca_cert` value as file path. See [file variable substitution](/docs/advanced-topics/environment-variable-substitution#file-variable-substitution-example) for how to read this value from a file.
END_COMMIT_OVERRIDE

### DIFF
--- a/plugins/destination/clickhouse/client/spec.go
+++ b/plugins/destination/clickhouse/client/spec.go
@@ -3,7 +3,6 @@ package client
 import (
 	"crypto/x509"
 	"fmt"
-	"os"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/cloudquery/cloudquery/plugins/destination/clickhouse/queries"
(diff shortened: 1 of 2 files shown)


erezrokah (commented):
Hi @candiduslynx, can you explain what is the bug here? Are we removing the file handling for using the built in SDK file expanding? Isn't this a breaking change? Or this didn't work at all?

Also, maybe provide some examples in `ca_cert` to how the value(s) should look like.

candiduslynx:
@erezrokah this PR has 2 changes (maybe they should be separate):
1. Check the `AppendCertsFromPEM` returned value to determine if the cert provided is valid
2. Remove the `read file` logic to use the built-in SDK file expansion

Possibly, the 2nd change should be expanded into a separate PR marked breaking (although I doubt anyone used this, as it required mounting the custom cert file to the container).

candiduslynx:
@erezrokah I've updated the PR & the description.
Kindly let me know if anything additional is required.

erezrokah (approved):
Thanks @candiduslynx, looks great 🚀

PR #9508, opened 2023-03-29
feat(aws): Explicitly Name Attached Policy Fields

bbernays:
#### Summary

One method for solving the issue of users not being able to find this data


erezrokah (commented):
We could move these to separate tables and do a breaking change (with proper change log), as we already have the user attached policies in a separate table https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_iam_user_attached_policies so better to be consistent.

WDYT?

bbernays:
I am fine if we want to close this PR and do it as a relation instead. I think we should plan our breaking changes ahead of time so that we can prioritize and batch our breaking change fixes (assuming they are not critical)

plugins/source/aws/resources/services/iam/roles.go
```diff
@@ -82,30 +77,6 @@ func getRole(ctx context.Context, meta schema.ClientMeta, resource *schema.Resou
...
-				return nil
-			}
-			return err
-		}
-		for _, p := range response.AttachedPolicies {
-			policies[*p.PolicyArn] = p.PolicyName
-		}
-	}
-	return resource.Set("policies", policies)
-}
```
bbernays, line 107: This is moved to `plugins/source/aws/resources/services/iam/role_attached_policies.go`

plugins/source/aws/resources/services/iam/users.go
```diff
@@ -21,11 +21,7 @@ func Users() *schema.Table {
 		Transform:           transformers.TransformWithStruct(&types.User{}),
 		Multiplex:           client.ServiceAccountRegionMultiplexer(tableName, "iam"),
 		Columns: []schema.Column{
-			{
-				Name:     "arn",
-				Type:     schema.TypeString,
-				Resolver: schema.PathResolver("Arn"),
-			},
```
bbernays, line 28: unneeded definition

plugins/source/aws/resources/services/iam/groups.go
```diff
@@ -46,33 +40,13 @@ func Groups() *schema.Table {
...
-	response, err := svc.ListAttachedGroupPolicies(ctx, &config)
-	if err != nil {
-		return err
-	}
-	policyMap := map[string]*string{}
-	for _, p := range response.AttachedPolicies {
-		policyMap[*p.PolicyArn] = p.PolicyName
-	}
-	return resource.Set(c.Name, policyMap)
-}
```
bbernays, line 78: Moved to `plugins/source/aws/resources/services/iam/group_attached_policies.go`

erezrokah (approved):
Looks great 🚀

erezrokah (changes requested):
Actually can we avoid the primary key order change?

erezrokah (approved):
Nice 🚀 Thanking for making it easier for me to review this PR

PR #9510, opened 2023-03-29
feat(k8s): Improve tables descriptions in docs

candiduslynx:


erezrokah (approved):

erezrokah (commented):
I think this should be a fix/feat as if someone runs the docs command they'll see different (better) results

hermanschaaf (approved):

Issue #8862, opened 2023-03-10
New Source: OracleDB


yevgenypats:
Similar to our [PostgreSQL source](https://www.cloudquery.io/docs/plugins/sources/postgresql/overview)

koltyakov:
Yes, please! This one would be really great to have. It's so often there is a legacy solution out there which you got to deal with. And it should be great to consolidate Oracle data in a decent databases.

PR #9511, opened 2023-03-29
feat: Add OracleDB source plugin

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8862

### References

- https://blogs.oracle.com/developers/post/connecting-a-go-application-to-oracle-database
- https://www.sqlines.com/mysql-to-oracle
- https://docs.oracle.com/database/121/SQLRF/sql_elements001.htm#SQLRF30020
- https://github.com/sijms/go-ora There's also https://github.com/godror/godror but that needs the [Oracle Client Libraries](https://github.com/godror/godror/tree/0040f4954e5bfabc081f2b1d58a28c4983a4712c#run-time-requirements) to be installed


erezrokah:
Marked this as ready for review but will probably need to tweak the CI once GitHub Actions [incident](https://www.githubstatus.com/incidents/z3c6q056q332) is resolved

koltyakov:
I'm testing this within an Oracle DB I have access to. A DB with lots of data and many schemas (owner namespaces). Hope some feedback can be helpful.

Probably the DB/connection is slow as I'm getting only 36 rps, while fetching similar data from a PostgreSQL DB within the same VPC gets 1700 rps. Too much difference, especially while a paginated chunk collection select is relatively identical from both DBs, and the target in both cases is the same SQLite on an M1 laptop.

I also got to modify USER_TAB_COLS to ALL_TAB_COLS and filter by specific table OWNER in listTables and updateTableConstraints to make it work. While `ALL_` could be connected with the account restrictions (in USER_TAB_COLS for my account is a void), working with OWNERs might be needed under the hood for all cases.

Data from a table is fetched in one chunk without pagination. Could it be the reason of a drastically slow RPS? However, in a PostgreSQL source plugin I can't see any pagination as well and it is not slow.

I also found that grabbing USER/ALL_TAB_COLS adds unrealistically long time to wait (minutes before even start) while only a couple of entities are intended being involve into sync. The filtering is happening not before but after fetching all tables.

erezrokah:
Thanks @koltyakov that's very helpful. Based on your feedback I'll add more tests targeted at performance (I'll create many different types of tables and resources to simulate), and improve based on those.

> UPD: Forgot to mentioned, a connection string without port number (default) couldn't be parsed.

I noticed that as well, seems to be a limitation of the Go module we use, but I'll see if there's anything to be done about that

erezrokah:
Hi @koltyakov, I did a couple of performance related fixes:
1. Dropped the `ORDER BY` from the `USER_TAB_COLS` query and now doing it in memory. It seems to have slow things down a bit, and doing it in Go seems to be faster. See https://github.com/cloudquery/cloudquery/pull/9511/commits/54d0b0a723d17f6636db7dc28b3409fdc82430cb
2. Remove the custom DB settings, and now using Go defaults. That should allow better connection pooling and creation. See https://github.com/cloudquery/cloudquery/pull/9511/commits/d29ad2ef09edf8d5fa49981ef39060ace65c5114
3. Use the concurrency setting when syncing tables, see https://github.com/cloudquery/cloudquery/pull/9511/commits/8c7bfc270fd331f3ff6e994266467015563b6093

I'll take a look at using a connection string without a port in a bit

erezrokah:
> I also got to modify USER_TAB_COLS to ALL_TAB_COLS and filter by specific table OWNER in listTables and updateTableConstraints to make it work. While ALL_ could be connected with the account restrictions (in USER_TAB_COLS for my account is a void), working with OWNERs might be needed under the hood for all cases.

Would it make sense to add an owner spec configuration item and filter by that?

erezrokah:
> UPD: Forgot to mentioned, a connection string without port number (default) couldn't be parsed.

See related issue. It seems to use a default port one needs to keep the `:`, see example in https://github.com/cloudquery/cloudquery/pull/9511/commits/63f96879b5a2de8c3754685e2e5d55f223036871

koltyakov:
Hi @erezrokah,

Thanks for the updates I'll check and share what performance improvements I could see with the same DB as I checked last time.

> Would it make sense to add an owner spec configuration item and filter by that?

Not sure, but probably. It also should be in queries to table data, for them to work I guess, as without a namespace before a table name a select won't necessarily work:

`Query 1 ERROR: ORA-00942: table or view does not exist` error when no owner is used before a table name.

In a scenario, when a user needs tables from different schemas/owners they could define multiple source yml's, so I'd say having owner as a spec parameter could be handy.

> Default port

It could be mentioned in docs that even with a default port used it should be explicitly provided, or to check if the connection url with without a port add :1521, maybe. Some software when exporting connection URL omits default port which could cause unnecessarily round trips to the docs.

```golang
func AddDefaultPort(connStr string) (string, error) {
	// Parse the URL to extract the host and port components
	parsedURL, err := url.Parse(connStr)
	if err != nil {
		return "", err
	}
	host, portString, _ := net.SplitHostPort(parsedURL.Host)
	if host == "" {
		host = parsedURL.Host
	}
	// If no port number found, add default port
	if portString == "" {
		parsedURL.Host = net.JoinHostPort(host, "1521")
	}
	return parsedURL.String(), nil
}
```

erezrokah:
> It could be mentioned in docs that even with a default port used it should be explicitly provided, or to check if the connection url with without a port add :1521, maybe. Some software when exporting connection URL omits default port which could cause unnecessarily round trips to the docs.

I'll make this more clear in the docs. I'd rather not do an custom parsing as that should be done in the Go module I believe, so I'd rather defer to it, at least in the initial release.

As for the owner and performance issues, we might need to get some more feedback so I lean into shipping this as is, and follow up later on based on the feedback we get. We can always add a owner option, but removing an existing one will be breaking for users. I'm testing on OrableDB Express Edition via a Docker image, so probably not the best performance indicator either.

erezrokah:
Followed up with https://github.com/cloudquery/cloudquery/pull/9511/commits/874ba6aac4885449c7a00ba320bedeb64c8d6f74 to remove a timezone related workaround fixed in https://github.com/sijms/go-ora/issues/338#issuecomment-1503615431

yevgenypats (approved):
Missed this one. Didn't review thoroughly but let's merge and then follow-up with the arrow PR.

erezrokah:
> Missed this one. Didn't review thoroughly but let's merge and then follow-up with the arrow PR.

Thanks, please not this is a source, not a destination so I think the Arrow PR should be similar to other sources

PR #9512, opened 2023-03-29
chore(website): Fail prebuild script on duplicate plugin ids

erezrokah:
#### Summary

I got bit by this in https://github.com/cloudquery/cloudquery/pull/9511, and apparently also `MySQL` but didn't notice.
If we have duplicate IDs the last one takes.

See current integration for MySQL:
https://www.cloudquery.io/integrations/mysql/file
image

It should show the example configuration, but it doesn't as it takes the values on the `unpublished` kind.


hermanschaaf (approved):
Ah, good catch!

erezrokah:
Going to force merge this to fix the Website as there's a GitHub Actions incident

Issue #9388, opened 2023-03-23
bug(aws,gcp,k8s,azure): Test caching in CI does not work


candiduslynx:
### Which problem is this feature request solving?

Every time we update huge plugins (like AWS) we retest all pf the code instead of only testing the affected parts.

### Describe the solution you'd like

Only testing the affected parts (need to make sure, though, that the libraries updates trigger the tests, too).
Possibly, this can be achieved by caching Go build cache & test cache (see [this](https://github.com/mvdan/github-actions-golang/blob/master/.github/workflows/test.yml) workflow as an example).

Relevant info:
https://pkg.go.dev/cmd/go#hdr-Build_and_test_caching
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows
https://github.com/marketplace/actions/cache

https://github.com/marketplace/actions/golang-with-cache
https://github.com/mvdan/github-actions-golang

### Pull request (optional)

- [x] I can submit a pull request

candiduslynx:
I'll be introducing [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) in #9518 to make the change easier.

candiduslynx:
# Desired outcome

Every plugin test workflow should utilize build & test cache.
This will enable us to significantly shorten the test time (for AWS it [allowed](https://github.com/cloudquery/cloudquery/actions/runs/4565948529/jobs/8057781114?pr=9515) running tests for 20s only instead of 10 minutes).

# Proposed solution
Add build & test caching for every plugin build we have.
The cache should be tied to the plugin, Go version & OS.

Cache should be restored every time we do something Go-related (build, test, lint), and saved after testing from the default (`main`) branch only (as it's inherited by other branches as well).
Noteworthy, the build & test cache should be saved regardless of changes if we're running in the default (`main`) branch.

## Additional steps

### Use [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)

We can utilize reusable workflows to extract the repeating parts from plugin workflows.
This allows us to:
1. Shorten the boilerplate code
2. Perform changes only in single place when we decide to make changes

### Update runner logic

Currently we use large runners for releases. The reason for this is to prime the build cache. However, this cache is only used during releases and `validate-release` workflows.

I want to utilize large runners on the default (`main`) branch only, as I propose to save the cache only from the default (`main`) branch.

erezrokah:
[...]

PR #9518, opened 2023-03-31
feat(workflows): Add reusable workflow files

candiduslynx:
Add [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) that will allow us to reuse the common job definitions:
* source plugin tests
  * adds ability to implement #9388 in a single place
* plugin release validation
* runner resolving

Once this part is in the `main` we can switch parts of the workflows to the reusable ones.

Example implementation: #9541 (aws source).

### DIFF
--- a/.github/workflows/reusable_resolve_runner.yml
+++ b/.github/workflows/reusable_resolve_runner.yml
@@ -1,0 +1,28 @@
+name: Resolve runner
+on:
+  workflow_call:
+    outputs:
+      runner:
+        description: 'Runner to be used (default: "ubuntu-latest")'
+        value:       ${{ jobs.resolve-runner.outputs.runner }}
+
+jobs:
+  resolve-runner:
+    timeout-minutes: 5
+    runs-on:         ubuntu-latest
+    outputs:
+      runner: ${{ steps.resolve.outputs.runner }}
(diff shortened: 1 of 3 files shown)


candiduslynx:
@erezrokah I think we might want to reconsider what runners we want to be used & when.
I propose using the large runners on `main` branch only or not use them at all.

The reason for them is simple: we would move from caching during release to caching from `main` branch. This enables:
1. Direct hits from other branches, as caches are inherited from the default branch (`main`)
2. Always storing the cache only from `main`

erezrokah:
> @erezrokah I think we might want to reconsider what runners we want to be used & when. I propose using the large runners on `main` branch only or not use them at all.
> 
> The reason for them is simple: we would move from caching during release to caching from `main` branch. This enables:
> 
> 1. Direct hits from other branches, as caches are inherited from the default branch (`main`)
> 2. Always storing the cache only from `main`

Can you open an issue describing the reasoning? I don't think a PR is the right place to discuss why we should do something nor do I see a downside with using large runners for PRs for big plugins.
Additionally we should discuss any caching changes in an issue as I'm not sure we'd like to mix between cache without properly discussing the impact

PR #9541, opened 2023-03-31
feat(aws): Update AWS source workflow

candiduslynx:
requires #9518 to be merged & change of reference to `@main` (or, better yet, to relative reference + add dependency in `on` workflow part)

### DIFF
--- a/.github/workflows/source_aws.yml
+++ b/.github/workflows/source_aws.yml
@@ -14,27 +14,16 @@ on:
 
 jobs:
   resolve-runner:
-    timeout-minutes: 5
-    runs-on: ubuntu-latest
-    outputs:
-      runner: ${{ steps.resolve.outputs.runner }}
-    steps:
-      - name: Check if should use large runner
-        id: large-runner
-        # We want to speed runs on the main branch which prime the cache
-        # We allow large runners only in this case to prevent forks from abusing them (it's enforced via runner groups access rules)
-        # IF YOU WANT TO USE A LARGE RUNNER YOU NEED TO ADD THE WORKFLOW TO THE `CloudQuery releases` GROUP IN https://github.com/organizations/cloudquery/settings/actions/runner-groups
-        if: github.event_name == 'push'
-        run: |
-          echo "runner=cloudquery-release-runner" >> $GITHUB_OUTPUT
-      - name: Resolve runner
-        id: resolve
-        run: |
-          RUNNER=${{ steps.large-runner.outputs.runner }}
-          echo "runner=${RUNNER:-"ubuntu-latest"}" >> $GITHUB_OUTPUT
-  plugins-source-aws:
+    uses: cloudquery/cloudquery/.github/workflows/reusable_resolve_runner.yml@feat/workflows/reusable
+  test:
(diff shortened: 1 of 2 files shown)


.github/workflows/source_aws.yml
```diff
@@ -44,86 +34,46 @@ jobs:
...
         uses: actions/setup-go@v3
         with:
-          go-version-file: plugins/source/aws/go.mod
-          cache: true
+          go-version-file:       plugins/source/aws/go.mod
+          cache:                 true
           cache-dependency-path: plugins/source/aws/go.sum
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+      - name: Restore build cache
```
erezrokah, line 44: We should first resolve https://github.com/cloudquery/cloudquery/pull/9656 before doing any caching related changes

.github/workflows/source_aws.yml
```diff
@@ -149,12 +99,23 @@ jobs:
...
+          cache:                 true
           cache-dependency-path: plugins/source/aws/go.sum
+      - name: Restore build cache
+        uses: actions/cache/restore@v3
+        with:
+          path: |
+                ~/.cache/go-build
+                ~/Library/Caches/go-build
+                ~\AppData\Local\go-build
+          key:  source-aws-${{ runner.os }}-go-${{ steps.set-up-go.outputs.go-version }}-${{ hashFiles('plugins/source/aws/go.sum') }}
```
erezrokah, line 116: We should first resolve https://github.com/cloudquery/cloudquery/pull/9656 before doing any caching related changes

.github/workflows/source_aws.yml
```diff
@@ -14,27 +14,17 @@ on:
...
-    timeout-minutes: 45
+    uses: cloudquery/cloudquery/.github/workflows/reusable_resolve_runner.yml@feat/workflows/reusable
+  test:
     name: "plugins/source/aws"
+    needs: [ resolve-runner ]
+    uses: cloudquery/cloudquery/.github/workflows/reusable_test_source.yml@feat/workflows/reusable
+    with:
+      plugin: aws
+      runner: ${{ needs.resolve-runner.outputs.runner }}
+  sanity-check:
```
erezrokah, line 25: Since the sanity check is now a separate job, we'll need to add it to the wait for all workflow, otherwise a failure of it won't be enforced

erezrokah (changes requested):
Thanks for the PR @candiduslynx, added a few comments. Also looks like the wait for all workflow is still broken, as with reusable workflows we need to account for each nested workflow name and not the top one

.github/workflows/source_aws.yml
candiduslynx: I thought the failure of the workflow job will trigger failure of the whole workflow, doesn't it?

erezrokah: Since it's not a required status check (that is controlled by the wait for all workflow) it will fail, but the PR will still be mergable

candiduslynx:
Drop in favor of #9780 and follow-up later

PR #9553, opened 2023-03-31
fix: MSSQL migrations

shimonp21:
Change normalization code to be analog to postgresql's migration code

Fixes failing any 2 consecutive links, even when there is no schema-change at all:

```
Loading spec(s) from .
Starting migration with 1 tables for: gcp (v8.4.2) -> [mssql (v3.0.3)]
Error: failed to sync v1 source gcp: failed to migrate source gcp (v8.4.2) on destination mssql (v3.0.3) : failed to call Migrate: rpc error: code = Unknown desc = tables gcp_compute_networks with changes [[column: self_link, type: update, current: self_link:TypeString:PK, previous: self_link:TypeString:PK:NotNull]] require force migration. use 'migrate_mode: forced'
```

### DIFF
--- a/plugins/destination/mssql/client/migrate.go
+++ b/plugins/destination/mssql/client/migrate.go
@@ -59,26 +59,28 @@ func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema
 	return schemaTables, nil
 }
 
-func (c *Client) normalizedTables(tables schema.Tables) schema.Tables {
-	var normalized schema.Tables
-	for _, table := range tables.FlattenTables() {
-		table := table
-		for i := range table.Columns {
-			// Since multiple schema types can map to the same MSSQL type we need to normalize them to avoid false positives when detecting schema changes
-			// This should never return an error
-			typ, _ := queries.SchemaType(table.Name, table.Columns[i].Name, queries.SQLType(table.Columns[i].Type))
-			table.Columns[i].Type = typ
-		}
-		// If there are no PKs, we use CqID as PK
-		pks := table.PrimaryKeys()
-		if !c.pkEnabled() || len(pks) == 0 {
-			table.Columns.Get(schema.CqIDColumn.Name).CreationOptions.PrimaryKey = true
-		}
+func (c *Client) normalizeTable(table *schema.Table) {
+	for i := range table.Columns {
+		// Since multiple schema types can map to the same MSSQL type we need to normalize them to avoid false positives when detecting schema changes
+		// This should never return an error
+		typ, _ := queries.SchemaType(table.Name, table.Columns[i].Name, queries.SQLType(table.Columns[i].Type))
+		table.Columns[i].Type = typ
 
-		normalized = append(normalized, table)
+		if !c.pkEnabled() {
+			table.Columns[i].CreationOptions.PrimaryKey = false
+		}
+		if table.Columns[i].CreationOptions.PrimaryKey {
+			table.Columns[i].CreationOptions.NotNull = true
+		}
 	}
+}
 
-	return normalized
+func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
+	flattenedTables := tables.FlattenTables()
+	for _, table := range flattenedTables {
+		c.normalizeTable(table)
+	}
+	return flattenedTables
 }
 
 func (c *Client) nonAutoMigratableTables(tables schema.Tables, schemaTables schema.Tables) (names []string, changes [][]schema.TableColumnChange) {
@@ -150,7 +152,7 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
 		return err
 	}
 
-	normalizedTables := c.normalizedTables(tables)
+	normalizedTables := c.normalizeTables(tables)
 
 	if c.spec.MigrateMode != specs.MigrateModeForced {
 		nonAutoMigratableTables, changes := c.nonAutoMigratableTables(normalizedTables, schemaTables)
--- a/plugins/destination/mssql/queries/templates/create_table.sql.tpl
+++ b/plugins/destination/mssql/queries/templates/create_table.sql.tpl
@@ -1,8 +1,10 @@
 CREATE TABLE {{.Table}} (
 {{with .Definitions}}{{template "col_defs.sql.tpl" .}}{{end}}
 {{- with .PrimaryKey}}
-  CONSTRAINT {{.Name}} PRIMARY KEY (
-{{template "col_names.sql.tpl" .Columns}}
+  {{- if ne (len .Columns) 0}}
+    CONSTRAINT {{.Name}} PRIMARY KEY (
+      {{template "col_names.sql.tpl" .Columns}}
+    )
+  {{- end}}
 {{- end}}
-  )
 );
--- a/plugins/destination/mssql/queries/tables_test.go
+++ b/plugins/destination/mssql/queries/tables_test.go
@@ -1,6 +1,7 @@
 package queries
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/cloudquery/plugin-sdk/schema"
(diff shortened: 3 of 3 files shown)


candiduslynx:
@shimonp21 
`self_link:TypeString:PK` -> `self_link:TypeString:PK:NotNull` is indeed a schema change that requires forced mode.
Maybe there was some other issue as well?

plugins/destination/mssql/client/migrate.go
```diff
@@ -59,26 +59,36 @@ func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema
...
+			cqIdColumn := table.Columns.Get(schema.CqIDColumn.Name)
+			if cqIdColumn != nil {
+				cqIdColumn.CreationOptions.PrimaryKey = true
+			}
+		}
 	}
+}
 
-	return normalized
+func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
```
candiduslynx, line 86:
```suggestion
func (c *Client) normalizedTables(tables schema.Tables) schema.Tables {
```

so that the call is unchanged

candiduslynx (changes requested):

plugins/destination/mssql/client/migrate.go
shimonp21: I copy-pasted the same name we have in postgresql

shimonp21:
@candiduslynx Yeah - it was already not-null in the MSSQL itself, but because the 'want/new' tables were not 'normalized' it tried to add a not-null constraint (even though one already existed...)

candiduslynx:
@shimonp21 thanks for the PR! I have only 1 outstanding issue with `_cq_id` setting done inside a loop, otherwise I'm OK with the changes (I guess we need to check ClickHouse as well)

plugins/destination/mssql/client/migrate.go
```diff
@@ -59,26 +59,36 @@ func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema
...
-		// If there are no PKs, we use CqID as PK
-		pks := table.PrimaryKeys()
-		if !c.pkEnabled() || len(pks) == 0 {
-			table.Columns.Get(schema.CqIDColumn.Name).CreationOptions.PrimaryKey = true
-		}
-
-		normalized = append(normalized, table)
+func (c *Client) normalizeTable(table *schema.Table) {
+	// this is for backward compatibility as I believe this is as of right now defined on the source
+	if c.pkEnabled() && len(table.PrimaryKeys()) == 0 {
```
candiduslynx, line 64: There's an issue with cq ID for cases where pk isn't enabled (append mode).

basically, you need to move this check after the loop and then do the update in the following case:
```go
if !c.pkEnabled() || len(table.PrimaryKeys()) == 0
```
this way you'll add PK to the cq ID if it's append mode or for overwrite modes if no PK defined

plugins/destination/mssql/client/migrate.go
```diff
@@ -150,7 +160,7 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
 		return err
 	}
 
-	normalizedTables := c.normalizedTables(tables)
+	normalizedTables := c.normalizeTables(tables)
```
candiduslynx, line 163: I'd rathe have it with the same name to have less diff (nit)

candiduslynx (changes requested):

plugins/destination/mssql/client/migrate.go
```diff
@@ -59,26 +59,36 @@ func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema
...
+	for i := range table.Columns {
+		// Since multiple schema types can map to the same MSSQL type we need to normalize them to avoid false positives when detecting schema changes
+		// This should never return an error
+		typ, _ := queries.SchemaType(table.Name, table.Columns[i].Name, queries.SQLType(table.Columns[i].Type))
+		table.Columns[i].Type = typ
+
+		if !c.pkEnabled() {
+			table.Columns[i].CreationOptions.PrimaryKey = false
+		}
+		if table.Columns[i].CreationOptions.PrimaryKey {
```
erezrokah, line 80: So this is the bulk of the fix? Setting `NotNull` in case the column is a PK?

plugins/destination/mssql/client/migrate.go
```diff
@@ -59,26 +59,36 @@ func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema
...
+	}
+
+	for i := range table.Columns {
+		// Since multiple schema types can map to the same MSSQL type we need to normalize them to avoid false positives when detecting schema changes
+		// This should never return an error
+		typ, _ := queries.SchemaType(table.Name, table.Columns[i].Name, queries.SQLType(table.Columns[i].Type))
+		table.Columns[i].Type = typ
+
+		if !c.pkEnabled() {
+			table.Columns[i].CreationOptions.PrimaryKey = false
```
candiduslynx, line 78: clears PK even on `_cq_id`, see [this comment](https://github.com/cloudquery/cloudquery/pull/9553/files#r1156030619)

plugins/destination/mssql/client/migrate.go
```diff
@@ -59,26 +59,36 @@ func (c *Client) schemaTables(ctx context.Context, tables schema.Tables) (schema
...
+		}
+		if table.Columns[i].CreationOptions.PrimaryKey {
+			table.Columns[i].CreationOptions.NotNull = true
+		}
+
+		// this is for backward compatibility as I believe this is as of right now defined on the source
+		if c.pkEnabled() && len(table.PrimaryKeys()) == 0 {
```
candiduslynx, line 77: move outside of loop (you set it every column otherwise)

candiduslynx (changes requested):

plugins/destination/mssql/client/migrate.go
shimonp21: I just remove this after talking with YP, this 'backward compatibility' hack is outdated and can be removed :)

plugins/destination/mssql/client/migrate.go
shimonp21: See comment on comment :)

plugins/destination/mssql/client/migrate.go
shimonp21: I don't believe that's correct (since I just copied the code from postgresql, which is our 'canonical destination').
https://github.com/cloudquery/cloudquery/blob/901472697dbd6370079659d0f08d8ef01865b987/plugins/destination/postgresql/client/migrate.go#L162

I believe that the correct behaviour is not to create primary keys at all for append mode (we don't have them for postgresql 'append' mode :) ).
re: "make _cq_id be PK if a PK is not defined" - that's handled by the source according to convo with YP - that was just a backward-compatibility hack that's six months outdated by now :) 

[above is corroborated by https://github.com/cloudquery/cloudquery/blob/901472697dbd6370079659d0f08d8ef01865b987/plugins/destination/postgresql/client/migrate.go#L361-L369]

shimonp21: See the second commit in this PR 'don't create PK in append mode' that should fix the failing tests :)

plugins/destination/mssql/queries/templates/create_table.sql.tpl
```diff
@@ -1,8 +1,10 @@
 CREATE TABLE {{.Table}} (
 {{with .Definitions}}{{template "col_defs.sql.tpl" .}}{{end}}
 {{- with .PrimaryKey}}
-  CONSTRAINT {{.Name}} PRIMARY KEY (
-{{template "col_names.sql.tpl" .Columns}}
+  {{- if ne (len .Columns) 0}}
```
candiduslynx, line 4: you can just use `{{with}}` as for slices of len = 0 (nil, too) it doesn't do anything

candiduslynx, line 4:
```suggestion
  {{- if .Columns}}
```
you can just use simple `{{if}}` as for slices of len = 0 (nil, too) it doesn't do anything

candiduslynx (changes requested):

shimonp21:
the test failures are due to a bug in the tests:
[the tables being migrated don't contain a PK, which is relied on by MSSQL plugin].
Tables coming from the source-plugins always contain a PK https://github.com/cloudquery/plugin-sdk/blob/bea4d00d6502a0a131abb2321685733af8de62c1/plugins/source/plugin.go#L85-L88.

Fix is, in the SDK, to change the 'id' columns to be PK, to better simulate the conditions of the real world (tables in our other tests always containt PK, just this migration-tests are missing them..).
https://github.com/cloudquery/plugin-sdk/blob/06fde4b0f4f4bf4bf07878f30d0cf6222e295642/plugins/destination/plugin_testing_migrate.go#L95

PR #9554, opened 2023-03-31
fix(github): Use cursor-based sync for `github_repository_dependabot_alerts`

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/530

Per [doc](https://docs.github.com/en/rest/dependabot/alerts?apiVersion=2022-11-28#list-dependabot-alerts-for-a-repository) this API supports cursors, so use them.


plugins/source/github/resources/services/repositories/alerts.go
```diff
@@ -30,8 +29,8 @@ func fetchAlerts(ctx context.Context, meta schema.ClientMeta, parent *schema.Res
 			return err
 		}
 		res <- alerts
-		opts.Page = strconv.FormatInt(int64(resp.NextPage), 10)
-		if resp.NextPage == 0 {
+		opts.After = resp.After
+		if len(resp.After) == 0 {
```
bbernays, line 33: LGTM as this is the same pattern we use in `resources/services/organizations/alerts.go`

bbernays (approved):

PR #9556, opened 2023-03-31
fix(aws)!: Remove `region` from `aws_neptune_global_clusters`

disq:
Global clusters don't have regions.

Neptune global cluster doesn't work with `ListTagsForResource` call: The endpoint requires a region. Giving it any region returns the `The specified resource name does not match an RDS resource in this region.` error. The ARN doesn't have any region itself. Neptune ARN docs [here](https://docs.aws.amazon.com/neptune/latest/userguide/tagging-arns.html) (no mention of global clusters)

Example Neptune global cluster arn: `arn:aws:rds::<account id>:global-cluster:<global cluster name>`

BEGIN_COMMIT_OVERRIDE
fix(aws)!: Remove `region` from `aws_neptune_global_clusters` (#9556)

BREAKING-CHANGE: Remove `region` from `aws_neptune_global_clusters` as global clusters are visible from many regions. Additionally, `aws_neptune_global_clusters` now uses account-level multiplexing instead of region-level, so only 1 entry would be returner when previously it could produce several entries.

fix(aws)!: Remove `tags` from `aws_neptune_global_clusters` (#9556)

BREAKING-CHANGE: Remove `tags` from `aws_neptune_global_clusters` as Neptune global cluster doesn't work with `ListTagsForResource`:
* This endpoint requires passing `region`
* Neptune global cluster doesn't have any region
* Passing any random region results in the `The specified resource name does not match an RDS resource in this region.`
END_COMMIT_OVERRIDE


disq:
> Hmm, interesting that they don't support calling for tags. Can you add tags on the cluster in the console?

There's no way to add tags to a global cluster in the console.

hermanschaaf (approved):

candiduslynx:
2 proposals:
1. split into 2 separate PRs
2. add commit override for the one with column removal

candiduslynx (approved):

PR #9557, opened 2023-03-31
fix(workflows): Wse empty tree as base for initial releases

candiduslynx:
Inspired by https://stackoverflow.com/a/54946189

### DIFF
--- a/.github/workflows/release_summary.yml
+++ b/.github/workflows/release_summary.yml
@@ -27,7 +27,13 @@ jobs:
       - name: Get latest tag
         id: tag
         run: |
-          tag=$(git describe --tags --match 'plugins-source-${{ steps.plugin.outputs.name }}*' --abbrev=0)
+          tag_prefix='plugins-source-${{ steps.plugin.outputs.name }}*'
+          if [[ $(git tag --list "${tag_prefix}" | wc -l) -gt 0 ]]; then
+             tag=$(git describe --tags --match "${tag_prefix}" --abbrev=0)
+          else
+             git tag --force empty-git-tree $(git hash-object -t tree /dev/null)
+             tag="empty-git-tree"
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/release_summary.yml
```diff
@@ -27,7 +27,13 @@ jobs:
       - name: Get latest tag
         id: tag
         run: |
-          tag=$(git describe --tags --match 'plugins-source-${{ steps.plugin.outputs.name }}*' --abbrev=0)
+          tag_prefix='plugins-source-${{ steps.plugin.outputs.name }}*'
+          if [[ $(git tag --list "${tag_prefix}" | wc -l) -gt 0 ]]; then
+             tag=$(git describe --tags --match "${tag_prefix}" --abbrev=0)
+          else
+             git tag --force empty-git-tree $(git hash-object -t tree /dev/null)
```
erezrokah, line 34: Nice, is creating the `tag` required? I'd prefer if we skip it as we `git push` to the repo at the end of the workflow, so I don't want us to push this tag back.

I can happen under some conditions https://github.com/stefanzweifel/git-auto-commit-action/blob/9cc0a1f55de2aa04a36721ca84463dd520f70f4c/entrypoint.sh#L172
not relevant for us at the moment, but we shouldn't make the `git tree` dirtier than needed if possible.

erezrokah (changes requested):
Thanks @candiduslynx, nice fix, added one comment for consideration

.github/workflows/release_summary.yml
candiduslynx: We return the tag to be the base of comparison.
If we squash the steps for tag determination & diff calc (basically, just do the inline diff calc) then we can skip the tag.
I'll try implementing this as a separate commit.

erezrokah (approved):
🚀

PR #9559, opened 2023-03-31
fix(aws)!: Add `cluster_arn` to `aws_ecs_cluster_services` primary key

disq:
The issue seems to be about the "old format" not having the parent (cluster) ARN in the service ARN... documented in [docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids)

> Current: arn:aws:ecs:region:aws_account_id:service/service-name
> New: arn:aws:ecs:region:aws_account_id:service/cluster-name/service-name

BEGIN_COMMIT_OVERRIDE
fix(aws)!: Add `cluster_arn` to `aws_ecs_cluster_services` primary key (#9559)

BREAKING-CHANGE: Add `cluster_arn` to `aws_ecs_cluster_services` primary key as cluster ID is absent in the "old format" ARNs. See [ecs-account-settings.html#ecs-resource-ids](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids) for more details.
END_COMMIT_OVERRIDE


candiduslynx (approved):

plugins/source/aws/resources/services/ecs/cluster_services.go
```diff
@@ -26,6 +26,14 @@ func ClusterServices() *schema.Table {
 					PrimaryKey: true,
 				},
 			},
+			{
```
candiduslynx, line 29: instead of this, use `transformers.WithPrimaryKeys("ClusterArn")`

candiduslynx (changes requested):

plugins/source/aws/resources/services/ecs/cluster_services.go
disq: Fixed

candiduslynx (approved):

PR #9611, opened 2023-04-01
fix(aws): Tag Pagination Properly Handled

bbernays:
#### Summary

Prior to this fix, cq would only resolve the first page of tags for sagemaker resources


plugins/source/aws/resources/services/sagemaker/training_jobs_fetch.go
```diff
@@ -1,64 +0,0 @@
...
-
-	c := meta.(*client.Client)
-	svc := c.Services().Sagemaker
-	config := sagemaker.ListTagsInput{
-		ResourceArn: r.TrainingJobArn,
-	}
-	response, err := svc.ListTags(ctx, &config)
-	if err != nil {
-		return err
-	}
```
bbernays, line 61: This was replaced with a paginator

disq (approved):

PR #9641, opened 2023-04-03
fix(clickhouse): Migration has to look only at `not null` option

candiduslynx:
Was inspired to check by #9553


bbernays (approved):

erezrokah (commented):
Does this PR fix anything? Looks like this should be a type `refactor:` and not appear in the change log

erezrokah:
So after discussing this with @candiduslynx the fix is to ignore the primary key setting when migrating tables

PR #9645, opened 2023-04-03
chore(aws): Pagination improvements

bbernays:
#### Summary

Replace manual pagination with the Paginator object

### DIFF
--- a/plugins/source/aws/resources/services/codebuild/projects.go
+++ b/plugins/source/aws/resources/services/codebuild/projects.go
@@ -3,7 +3,6 @@ package codebuild
 import (
 	"context"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/codebuild"
 	"github.com/aws/aws-sdk-go-v2/service/codebuild/types"
 	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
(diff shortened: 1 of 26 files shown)


plugins/source/aws/resources/services/codebuild/projects.go
```diff
@@ -43,24 +42,20 @@ func fetchCodebuildProjects(ctx context.Context, meta schema.ClientMeta, parent
...
+	paginator := codebuild.NewListProjectsPaginator(svc, &config)
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(ctx)
 		if err != nil {
 			return err
 		}
-		if len(response.Projects) == 0 {
-			break
+		if len(page.Projects) == 0 {
+			continue
```
disq, line 52:
```suggestion
			break
```

disq, line 52: This page being empty doesn't mean we don't have more pages I'm guessing... so we leave the actual check to `paginator.HasMorePages()`

bbernays: Exactly

disq: Also we can't use the PreResourceResolver pattern here because https://github.com/cloudquery/plugin-sdk/issues/338

plugins/source/aws/resources/services/transfer/servers.go
```diff
@@ -78,20 +73,17 @@ func resolveServersTags(ctx context.Context, meta schema.ClientMeta, resource *s
...
+	paginator := transfer.NewListTagsForResourcePaginator(svc, &input)
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(ctx)
 		if err != nil {
 			if cl.IsNotFoundError(err) {
 				continue
 			}
 			return err
 		}
-		client.TagsIntoMap(result.Tags, tags)
```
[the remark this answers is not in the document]
candiduslynx: they are most certainly:
* not duplicated
* a short list

so I'd leave as is

disq (approved):

PR #9651, opened 2023-04-03
fix(aws)!: Change Primary keys for `aws_dynamodb_global_tables` to include `region`

bbernays:
#### Summary

Global tables can span multiple regions, because the ARN does not include a region in it the singular value will be duplicated when resolving multiple regions.

By adding the region to the PK we can ensure the record is unique


disq:
Did you have a chance to test this for the _2017.11.29 (Legacy) Global Tables_? `aws_dynamodb_tables` contain the current non-legacy global tables.

candiduslynx (approved):

PR #9656, opened 2023-04-04
fix(workflows): Disable Go cache for non-test jobs in AWS workflow

candiduslynx:
Part of #9388


.github/workflows/wait_for_required_workflows.yml
```diff
@@ -81,7 +81,7 @@ jobs:
               const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
               console.log(`Got the following check runs: ${JSON.stringify(runs)}`)
               const matchingRuns = runs.filter(({ name }) => actions.includes(name))
-              const failedRuns = matchingRuns.filter(({ conclusion }) => conclusion !== 'success')
+              const failedRuns = matchingRuns.filter(({ conclusion }) => conclusion !== 'success' && conclusion !== 'skipped')
```
erezrokah, line 84: Is this change related to the caching change? Is so let's move it to a separate PR so we can test it separately and also revert it easily if it breaks something

.github/workflows/source_aws.yml
```diff
@@ -47,12 +47,26 @@ jobs:
...
           fetch-depth: 2
-      - name: Set up Go 1.x
+      - name: Set up Go
+        id:   set-up-go
         uses: actions/setup-go@v3
         with:
-          go-version-file: plugins/source/aws/go.mod
-          cache: true
+          go-version-file:       plugins/source/aws/go.mod
+          cache:                 true
```
erezrokah, line 55: Should we disable the `setup-go` caching if we're doing it manually? Also can we drop the indentation changes?

erezrokah (changes requested):
Thanks for the PR @candiduslynx, let's start by scoping this PR to a single change (the caching one).
Since CI changes can have a significant impact, we should make them as small as possible and test them one by one.

I'll try to test this change later on

.github/workflows/wait_for_required_workflows.yml
candiduslynx: it is related to skip in `validate-release`, I'll extract it

.github/workflows/source_aws.yml
candiduslynx: setup-go does the dependency (go mod) cache, which we omit in the build & test cache. Basically, both are required

.github/workflows/source_aws.yml
```diff
@@ -88,39 +100,49 @@ jobs:
...
+          cache-dependency-path: plugins/source/aws/go.sum
+      - name: Restore build cache
+        uses: actions/cache/restore@v3
+        with:
+          path:         |
+                        ~/.cache/go-build
+                        ~/Library/Caches/go-build
+                        ~\AppData\Local\go-build
```
erezrokah, line 135: Since we run only on Linux can we remove the `~/Library/Caches/go-build` and `~\AppData\Local\go-build`? See https://github.com/actions/cache/blob/04f198bf0b2a39f7230a4304bf07747a0bddf146/examples.md#linux-1

erezrokah (changes requested):

.github/workflows/source_aws.yml
```diff
@@ -88,39 +100,49 @@ jobs:
       - name: Run sync
         if: startsWith(github.head_ref, 'release-please--branches--main--components')
         run: cloudquery sync test/sanity.yml --log-console
+      - name: Save build cache
+        if:   github.ref == 'refs/heads/main'
```
erezrokah, line 104: What's the reason we only save the cache on the `main` branch?

.github/workflows/source_aws.yml
```diff
@@ -98,20 +119,23 @@ jobs:
...
+          cache:                 true
+          cache-dependency-path: plugins/source/aws/go.sum
+      - name: Restore build cache
+        uses: actions/cache/restore@v3
+        with:
+          path:         |
+                        ~/.cache/go-build
+                        ~/Library/Caches/go-build
+                        ~\AppData\Local\go-build
+          key:          source-aws-${{ runner.os }}-go-${{ steps.set-up-go.outputs.go-version }}-${{ hashFiles('plugins/source/aws/go.sum') }}
```
erezrokah, line 136: Let's also descope this change from the PR. Since this impacts release PRs and releases I'd rather do it separately (it changes caching so we use the same one for releases and test jobs, those run in parallel so I think we need to ensure there aren't race conditions here).
Also it seems the reason for this PR is to cache tests, which is unrelated to this change

erezrokah (changes requested):
Added a few more comments

.github/workflows/wait_for_required_workflows.yml
candiduslynx: Extracted to https://github.com/cloudquery/cloudquery/pull/9668

.github/workflows/source_aws.yml
candiduslynx: I used this just to be consistent. These directories don't exist on the runner & it doesn't fail the cache, however, if we ever decide to support non-linux machines it's be easier.

.github/workflows/source_aws.yml
erezrokah: I recommend reading https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache as PRs cannot modify the `main` branch cache anyway.
They can only access the `main` branch cache or their own

candiduslynx: It will be propagated to the other branches if it's created on `main`.
The other branches caches are only visible to the children branches, so of no use to the most of the PRs.

Butthey will occupy the cache space & we will overflow the 10GiB limit on the caches).

.github/workflows/source_aws.yml
candiduslynx: We want the build & test cache to be consistent. This change addresses this also by using the proper go version in the key file (see usage of setup-go outputs).
Additionally, there won't be any race conditions as the only place where we do save the cache is after push to `main` & testing (so that it has the most useful info for the future).

.github/workflows/source_aws.yml
erezrokah: What's the value of having separate caches? It's seems that having them separated means duplicate save/restore operations, also separate cache keys (we should verify that actually otherwise they'll override each other)

candiduslynx: It just requires additionally to track the go mod dir location & I'd rather leave it to the Go team =)

.github/workflows/source_aws.yml
erezrokah: But won't that mean PRs almost always use a dirty cache? Let's say someone merges an unrelated PR to `main`, so that means that if I push a new commit to my PR, I'll start using the cache from `main` instead of my PR's cache

erezrokah: Regarding cache limits, we're already way over https://github.com/cloudquery/cloudquery/actions/caches, but GitHub evicts the least recently used ones, so PR caches gets evicted fast

.github/workflows/source_aws.yml
erezrokah: > We want the build & test cache to be consistent. This change addresses this also by using the proper go version in the key file (see usage of setup-go outputs).

But it changes the cache key so the validate release and test workflows have the same key.
The reason for using this specific key in the validate release workflow is because it's used in the release workflow:
https://github.com/cloudquery/cloudquery/blob/33ebc306af2303174edc8bdc1a36e85cbcf88862/.github/workflows/release_plugin.yml#L44
https://github.com/cloudquery/cloudquery/blob/1621cf770f3c779335e2e6d38ab8e0d69ac92688/.github/workflows/source_aws.yml#L107

> so that it has the most useful info for the future

I'm not sure that's correct for releases. In releases the value is caching Go Releaser dry run, so next time we run it (the actual publish run) it's much faster, see example here:
https://github.com/cloudquery/cloudquery/actions/runs/4541185612/jobs/8003037262
image

erezrokah: Regardless I don't think it's the purpose of this PR so we should remove this change from it and focus the PR on caching tests

.github/workflows/source_aws.yml
erezrokah: Between the tradeoff of having two different caches I'd rather have the hard coded value as that's not likely to change and also shown in the cache action docs, see https://github.com/actions/cache/blob/04f198bf0b2a39f7230a4304bf07747a0bddf146/examples.md#linux-1

.github/workflows/source_aws.yml
```diff
@@ -47,12 +47,26 @@ jobs:
...
         uses: actions/setup-go@v3
         with:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
+      - name: Restore build cache
+        uses: actions/cache/restore@v3
+        with:
+          path:         |
+                        ~/.cache/go-build
```
erezrokah, line 61: Aren't the build artifacts already cached by `actions/setup-go`? See https://github.com/actions/setup-go/tree/41c2024c46acfe1d0b8c9b7f20e28406983e553b#caching-dependency-files-and-build-outputs and also https://github.com/actions/setup-go/blob/41c2024c46acfe1d0b8c9b7f20e28406983e553b/src/package-managers.ts#L13

Looks like it caches the result of `go env GOMODCACHE` (module cache) and `go env GOCACHE` (build cache).
`GOCACHE` seems to resolve to `~/.cache/go-build` in the GitHub action runner:
https://github.com/cloudquery/cloudquery/actions/runs/4607636489/jobs/8142407370#step:3:30
image

erezrokah (changes requested):
Added another comment as it seems `setup-go` should already cache build artifacts so we should probably research why it doesn't work before opting out the default logic

.github/workflows/source_aws.yml
erezrokah: Also 
image

.github/workflows/source_aws.yml
candiduslynx: I left only the setup-go cache for now

.github/workflows/source_aws.yml
candiduslynx: removed this for now

.github/workflows/source_aws.yml
candiduslynx: removed

.github/workflows/source_aws.yml
candiduslynx: I agree and I tried to remove the custom cache to use the setup-go cache now

.github/workflows/source_aws.yml
```diff
@@ -98,20 +100,21 @@ jobs:
...
           go-version-file: plugins/source/aws/go.mod
+          cache: false
+      - uses: actions/cache@v3
+        if: startsWith(github.head_ref, 'release-please--branches--main--components') || github.event_name == 'push'
+        with:
+          path: |
+                ~/.cache/go-build
+                ~/go/pkg/mod
+          key:          ${{ runner.os }}-go-1.19.5-release-cache-${{ hashFiles('plugins/source/aws/go.sum') }}
+          restore-keys: ${{ runner.os }}-go-1.19.5-release-cache-plugins-source-aws
```
candiduslynx, line 117: This should be addressed separately, as the go ver should be taken from setup-go output.
Additionally, we might want to change the naming convention there as well

candiduslynx:
@erezrokah I think this might actually work for the AWS test cache, and then we can fix it in other workflows as well

.github/workflows/source_aws.yml
```diff
@@ -112,6 +112,7 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version-file: plugins/source/aws/go.mod
+          cache: false
```
erezrokah, line 115:
```suggestion
```

This already the default so not needed. We can do it once we update to v4

.github/workflows/source_aws.yml
```diff
@@ -153,8 +154,7 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version-file: plugins/source/aws/go.mod
-          cache: true
-          cache-dependency-path: plugins/source/aws/go.sum
+          cache: false
```
erezrokah, line 157:
```suggestion
```

This already the default so not needed. We can do it once we update to v4

erezrokah (changes requested):
Added a few more comments. I don't think we should disable the cache for the test policies job.

We should do something like:

```yaml
      - uses: actions/cache@v3
        with:
          path: |
            ~/.cache/go-build
            ~/go/pkg/mod
          key: ${{ runner.os }}-go-1.19.5-policies-cache-${{ hashFiles('plugins/source/aws/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-go-1.19.5-policies-cache-plugins-source-aws
```

erezrokah:
Also related https://github.com/actions/setup-go/issues/358

.github/workflows/source_aws.yml
candiduslynx: I think it's better to leave this explicitly turned off as it doesn't affect the result now

.github/workflows/source_aws.yml
candiduslynx: I think it's better to leave this explicitly turned off as it doesn't affect the result now

candiduslynx:
> Added a few more comments. I don't think we should disable the cache for the test policies job.
I don't think so, too.
However, currently I'd like to skip this in test-policies as it only runs the build.

The proper solution IMO would be to disable `setup-go` caching and use manual logic for caching:

* Save only from `main` after tests and do it regardless of the cache hit (this way we update the cache with the latest test data available)
* Other jobs (including releases as well) should only perform cache restore

.github/workflows/source_aws.yml
erezrokah: What's the added value of setting a configuration to the default value?

erezrokah:
Also I think once https://github.com/actions/setup-go/issues/358 is resolved we'll revert back to using `setup-go` caching

candiduslynx:
> What the reasoning for not caching `test-policies`? Won't that make the CI slower as each time we'll do a full build, impacting the performance improvement we're trying to achieve?

We are mostly hit by the testing part, not the build one. So, instead of introducing another cache I'd like to scope caching in setup-go only to the testing job

erezrokah:
It takes ~6m build to AWS without cache:
image

And 8s with cache:
image

The results are from this PR.

Not sure we'd want to give that up

erezrokah:
> We are mostly hit by the testing part, not the build one

I think this is because caching didn't work for tests, and worked for builds

.github/workflows/source_aws.yml
```diff
@@ -112,6 +112,7 @@ jobs:
         uses: actions/setup-go@v3
```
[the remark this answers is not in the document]
candiduslynx: I think that we might be getting into another issue with this solution:
* the cache limit is 10 GiB
* each Go cache is about 100-300 MiB, let's use 150 MiB in the calculations to be on the conservative side (300 MiB likely comes from the huge plugins like AWS or Azure)
* that means we can store something like 50 caches to be used effectively before they start being evicted
* we already have 38 source plugins & 21 destination plugins (59 in total, add CLI to it to have 60)
* If we don't store a single cache but, instead, use 3 different caches, we effectively will store 150 (test) + 100 + 100 (2 x build & mod) = 350 MiB for a plugin (= 21 GB for all plugins + CLI, so we will be evicting a lot of data)

Looking at this, I suggest storing a single cache for plugin, that will include:
* mod cache
* build & test cache

To make most of it, we'll have to store it after the testing (and, better yet, from the `main` branch for it to be properly reused), and to restore in other builds related to a plugin (or CLI).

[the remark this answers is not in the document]
erezrokah: I suggested we first fix the bug in our CI, and do any cache optimization later on. Once we fix the bug we should be able so see if we're hitting any cache limits issues, the should be able to do benchmarks and see if that can be improved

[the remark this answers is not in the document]
candiduslynx: For the cache limits we can take a look here: https://github.com/cloudquery/cloudquery/actions/caches

I'll be updating the workflow files for the heavy sources (aws, azure, gcp, k8s) to work with go cache properly, we can discuss the different approach separately.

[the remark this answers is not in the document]
erezrokah: > we already have 38 source plugins & 21 destination plugins (59 in total, add CLI to it to have 60)

We only need this fix to 4 plugins, then ones that have policy tests

candiduslynx:
Drop this impl to start anew

PR #9668, opened 2023-04-04
feat(workflows): Move `validate-release` step checks to job level

candiduslynx:
Extracted from #9656.
Allows to skip `validate-release` job entirely (not using any runner) while not failing `waif-for-required-workflows`

### Note on `wait-for-required-workflows` change
Instead of skipping every step in `validate-release` job (that still will result in job running) I moved the `if` section to the `job` level, allowing to skip the entire job.
However, this requires `wait-for-required-workflows` to accept the `skipped` status as well.

### DIFF
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -81,7 +81,7 @@ jobs:
               const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
               console.log(`Got the following check runs: ${JSON.stringify(runs)}`)
               const matchingRuns = runs.filter(({ name }) => actions.includes(name))
-              const failedRuns = matchingRuns.filter(({ conclusion }) => conclusion !== 'success')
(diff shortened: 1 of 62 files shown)


.github/workflows/wait_for_required_workflows.yml
```diff
@@ -81,7 +83,7 @@ jobs:
               const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
               console.log(`Got the following check runs: ${JSON.stringify(runs)}`)
               const matchingRuns = runs.filter(({ name }) => actions.includes(name))
-              const failedRuns = matchingRuns.filter(({ conclusion }) => conclusion !== 'success')
+              const failedRuns = matchingRuns.filter(({ name, conclusion }) => !(conclusion == 'success' || (can_skip.includes(name) && conclusion == 'skipped')))
```
candiduslynx, line 86: I added an option to tell what jobs are OK to be skipped

candiduslynx:
This PR can be ditched in favor of #9735

PR #9708, opened 2023-04-04
chore: Fix Tests in table_diff

bbernays:
#### Summary
Fix tests in its own PR, so in the next pr #9709 I can safely change the logic


erezrokah (approved):
Ah, I think it's time we put this into our CI as this is not the first time we forgot to update the tests

PR #9709, opened 2023-04-04
chore: Change behavior for changing ordering

bbernays:
#### Summary

When a table has 2 Primary keys and we replace 1 of them the current implementation marks it as breaking because the order has changed. This PR validates that all of the pks before and after are exactly the same, but just in a different order.

### DIFF
--- a/scripts/table_diff/changes/changes_test.go
+++ b/scripts/table_diff/changes/changes_test.go
@@ -189,15 +189,15 @@ func Test_getChanges(t *testing.T) {
 			wantChanges: []change{
 				{
 					Text:     "Table `gcp_resourcemanager_projects`: primary key constraint added to column `name`",
-					Breaking: false,
+					Breaking: true,
 				},
 				{
 					Text:     "Table `gcp_resourcemanager_projects`: primary key constraint added to column `project_id`",
-					Breaking: false,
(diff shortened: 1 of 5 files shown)


erezrokah:
> When a table has 2 Primary keys and we replace 1 of them the current implementation marks it as breaking because the order has changed

I'm confused by this a bit. Isn't replacing a PK breaking?

erezrokah (commented):
OK so I think I get it now, it fixes the following report:

- :warning: BREAKING CHANGE: Table `aws_iam_users`: primary key order changed from `id, account_id` to `account_id, arn`

To check order change only if that's the only PK change

scripts/table_diff/changes/changes_test.go
```diff
@@ -223,6 +223,20 @@ func Test_getChanges(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:         "Should mark PK order change as breaking",
```
erezrokah, line 227:
```suggestion
			name:         "Should not mark report PK order change if they were other PK changes",
```

erezrokah:
Maybe we should keep that report? I mean a new PK was added + there's a new composite key

erezrokah (approved):

Issue #9655, opened 2023-04-04
feat: Reusable workflows


candiduslynx:
### Which problem is this feature request solving?

I would like to be able to reuse the workflows/jobs to reduce boilerplate code.

> **Warning**
The previous CI issue (#9388) took a couple of weeks to get resolved.

> **Note**
Currently only @erezrokah has the expertise required to review the changes.
Additionally, the changes have to be thoroughly tested.
This makes this issue a slow one to implement as well as review.

### Describe the solution you'd like

There's a nice feature GitHub provides called [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows).
This allows to extract the jobs that are copied many times into a separate workflows and instead of copying the workflow logic just call the reusable workflow.

This is extracted from #9388 to be implemented and discussed separately.

### Pull request (optional)

- [X] I can submit a pull request

candiduslynx:
AWS source: `validate-release` done in #9780

erezrokah:
Hi @candiduslynx, thanks for the PR. I think it might be worth bringing this issue to a wider discussion with the rest of the team. Not sure we would like to make the wait for all logic even more complex than it is now, just to reduce code duplication.

WDYT?

candiduslynx:
Hi @erezrokah, I'm fine with bringing it to a wider discussion.

The wait logic is there only because of the stuff we're missing from GitHub, so it will always be a hack.

I would still want us to reduce the code duplication and maintain a more easily upgradable approach.

erezrokah:
@bbernays I think this won't impact users writing their own plugins, as re-usable workflows will only be used in the monorepo (this repository). The workflows in the monorepo are already not usable in single plugin repos, and adding/not adding re-usable workflows won't change that I think

bbernays:
Thank you @erezrokah for clarifying that for me!

erezrokah:
> Beyond the user impact how much time and effort do we as a team spend editing or debugging github workflows?

On my end I would say I spend the most time when we need to update the wait for all workflow. Otherwise debugging a single plugin workflow (e.g. testing a new destination) is very much scoped to that plugin.

erezrokah:
Maybe it would be helpful to summarize where we would like to use re-suable workflows:

- `resolve-runner` - we have 5 of those
- `validate-release` - we have 62 of those. Of of the 62, 6 require `CGO_ENABLED: 1` so are different an will require a separate workflow
[...]

PR #9734, opened 2023-04-05
feat(workflows): Reusable `resolve-runner` workflow

candiduslynx:
Extracted from #9518.

Implemented using relative call + added dependency on the reusable workflow file changes.

Part of reusable workflows (#9655) implementation


erezrokah (commented):
This looks good I think, but as I don't want to make too many CI changes at the same time, I'll merge this on Sunday next week

erezrokah (approved):

PR #9735, opened 2023-04-05
feat(workflows): Reuse `validate-release`

candiduslynx:
This builds on top of ideas from #9668 and can safely replace it.

`wait-for-required-workflows` change is required to make it wait for skipped jobs, too

PR #9736, opened 2023-04-05
Release please  branches  main  components test workflow

candiduslynx:
Test #9735

### DIFF
--- a/.github/workflows/validate_release.yml
+++ b/.github/workflows/validate_release.yml
@@ -1,0 +1,69 @@
+name: Validate release
+
+on:
+  workflow_call:
+    inputs:
+      path:
+        description: 'Go mod directory (e.g. "plugins/source/aws", "plugins/destination/csv", "cli")'
+        type:        string
+        required:    true
+      runner:
+        description: 'Runner to be used (default: "ubuntu-latest")'
+        type:        string
+        required:    false
+        default:     'ubuntu-latest'
+      CGO_ENABLED:
+        description: 'CGO_ENABLED value to be used (default: 0)'
+        type:        number
+        required:    false
+        default:     0
+
+jobs:
+  validate-release:
+    # this if check is left here as a precautionary measure, as the proper way to do it is in the caller workflow
+    if:              startsWith(github.head_ref, 'release-please--branches--main--components') || github.event_name == 'push'
+    name:            Validate release
+    timeout-minutes: 60
+    runs-on:         ${{ inputs.runner }}
+    env:
+      CGO_ENABLED: 0
+    steps:
+    - id:  dash-path
+      run: |
+           echo "dashed=$(echo ${{ inputs.path }} | tr '/' '-')" >> $GITHUB_OUTPUT
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Set up Go
+      id:   set-up-go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file:       ${{ inputs.path }}/go.mod
+        cache:                 true
+        cache-dependency-path: ${{ inputs.path }}/go.sum
+    - uses: actions/cache@v3
+      with:
+        path:         |
+                      ~/.cache/go-build
+                      ~/go/pkg/mod
+        key:          ${{ runner.os }}-go-1.19.5-release-cache-${{ hashFiles(format('{0}/go.sum', inputs.path)) }}
+        # actually, this value is never ever saved, but left for consistency (renaming cache keys os out of scope)
+        restore-keys: |
+                      ${{ runner.os }}-go-1.19.5-release-cache-${{ steps.dash-path.outputs.dashed }}
+    - name: Install GoReleaser
+      uses: goreleaser/goreleaser-action@v3
+      with:
+        distribution: goreleaser-pro
+        version:      latest
+        install-only: true
+    - name: Run GoReleaser Dry-Run
+      run:  >
+            goreleaser release
+            --timeout 50m
+            --snapshot
+            --clean
+            --skip-validate
+            --skip-publish
+            --skip-sign
+            -f ${{ inputs.path }}/.goreleaser.yaml
+      env:
+        GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -18,10 +18,12 @@ jobs:
       - uses: actions/github-script@v6
         env:
           FILES: ${{ steps.changed-files.outputs.all }}
+          CAN_SKIP: validate-release
         with:
           script: |
             const fs = require("fs");
             const files = process.env.FILES.split(' ')
+            const can_skip = process.env.CAN_SKIP.split(' ')
             console.log(files)
             if (files.length >= 300) {
               // This is a GitHub limitation https://github.com/cloudquery/cloudquery/issues/2688
@@ -81,7 +83,7 @@ jobs:
               const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
               console.log(`Got the following check runs: ${JSON.stringify(runs)}`)
               const matchingRuns = runs.filter(({ name }) => actions.includes(name))
-              const failedRuns = matchingRuns.filter(({ conclusion }) => conclusion !== 'success')
+              const failedRuns = matchingRuns.filter(({ name, conclusion }) => !(conclusion == 'success' || (can_skip.includes(name) && conclusion == 'skipped')))
               if (failedRuns.length > 0) {
                 throw new Error(`The following required workflows failed: ${failedRuns.map(({ name }) => name).join(", ")}`)
               }
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -18,7 +18,8 @@ jobs:
       - uses: actions/github-script@v6
         env:
           FILES: ${{ steps.changed-files.outputs.all }}
-          CAN_SKIP: validate-release
+          CAN_SKIP: >-
+            validate-release
         with:
           script: |
             const fs = require("fs");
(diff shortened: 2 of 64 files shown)


candiduslynx:
@erezrokah I've updated the code to address merge conflict, it's ready for merge now

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -27,6 +29,10 @@ jobs:
               // This is a GitHub limitation https://github.com/cloudquery/cloudquery/issues/2688
               throw new Error('Too many files changed. Skipping check. Please split your PR into multiple ones.')
             }
+
+            const can_skip = process.env.CAN_SKIP.split("\n")
+            console.log(can_skip)
+
```
erezrokah, line 35:
```suggestion
```

I this log message needed? I think we can remove it now as it prints a constant value

erezrokah (changes requested):
I'm not sure this is working as expected.
For the skip test the [wait for all took](https://github.com/cloudquery/cloudquery/actions/runs/4649767311/jobs/8228345356?pr=9760) 15m 23s:
image
and the [AWS workflow took](https://github.com/cloudquery/cloudquery/actions/runs/4649767376?pr=9760) 16m 47s:
image

I'd expect the wait for all workflow to finish last.

For the non skip test (this PR):
The [wait for all took 7m 33s](https://github.com/cloudquery/cloudquery/actions/runs/4649748153/jobs/8228311891?pr=9736):
image

and the [AWS workflow took](https://github.com/cloudquery/cloudquery/actions/runs/4649748183?pr=9736) 14m10s:
image

so the gap is even higher.

I recommend scoping this change to a single plugin. That would make it easier to debug and review. If it works for a single plugin, we can add it to others. WDYT?

.github/workflows/wait_for_required_workflows.yml
candiduslynx: we need the `can_skip` variable though. I'll remove the print

candiduslynx:
@erezrokah the issue with timelines was caused by the merging one branch into the other (basically, last commit was the same) as it messed a bit with GH trying not to do the same work several times.

I've updated the branches and have them diverged a bit, it should help with the proper testing.

As for the fact that the AWS source job might take a bit `longer` than the wait -- it's all in the wait logs (we got the successful status from API, nothing else to wait). Maybe the time displayed for the job also includes some additional tine for artifacts/ something else?

candiduslynx:
@erezrokah see https://github.com/cloudquery/cloudquery/actions/runs/4649767311/jobs/8228345356?pr=9760#step:4:625 for the wait for the required workflows reasoning

erezrokah:
Should this PR be ready? I see the AWS still running but wait for all completed:
image

image

erezrokah:
As I mentioned, it might be easier to test this for a single plugin instead of all of them. Regardless I think we would like to introduce this change incrementally instead of all plugins as the same time.

candiduslynx:
> As I mentioned, it might be easier to test this for a single plugin instead of all of them. Regardless I think we would like to introduce this change incrementally instead of all plugins as the same time.

@erezrokah seems there's a bug in the logic as we see in [here](https://github.com/cloudquery/cloudquery/actions/runs/4650454094/jobs/8229467827?pr=9736#step:4:379) that the following entries triggered the workflows to be marked ready:
```json
{"name":"plugins/source/awspricing","conclusion":"success"},
{"name":"plugins/source/awspricing","conclusion":"success"}
```

So we saw a double entry for `awspricing` and no for `aws`, interesting...
I think fixing the logic for the `wait-for-required-workflows` may be out of scope of this PR.

I can do a separate PR introducing this change only to AWS source testing, and, if we're satisfied with the results, we can merge this and then do refactoring for the other plugin workflows, WDYT?

erezrokah:
> So we saw a double entry for awspricing and no for aws, interesting...
I think fixing the logic for the wait-for-required-workflows may be out of scope of this PR.

I haven't seen this issue before, so we should probably fix it before doing any other CI changes.
I'll try to reproduce in a separate PR, without this change and see if it does.

> I can do a separate PR introducing this change only to AWS source testing, and, if we're satisfied with the results, we can merge this and then do refactoring for the other plugin workflows, WDYT?

Than sounds good, but it will have to wait for the fix in the wait for all logic

erezrokah:
This SDK PR https://github.com/cloudquery/cloudquery/pull/9661 doesn't seem to have this issue

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       - uses: actions/github-script@v6
         env:
           FILES: ${{ steps.changed-files.outputs.all }}
+          CAN_SKIP: |-
+            validate-release / validate-release
```
erezrokah, line 22:
```suggestion
```

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -27,6 +29,9 @@ jobs:
               // This is a GitHub limitation https://github.com/cloudquery/cloudquery/issues/2688
               throw new Error('Too many files changed. Skipping check. Please split your PR into multiple ones.')
             }
+
+            const can_skip = process.env.CAN_SKIP.split("\n")
```
erezrokah, line 33:
```suggestion
            const canSkip = ['validate-release / validate-release']
```

Not sure why we need the env variable and use the casing used in JS (and in this script)

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -81,7 +87,7 @@ jobs:
               const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
               console.log(`Got the following check runs: ${JSON.stringify(runs)}`)
               const matchingRuns = runs.filter(({ name }) => actions.includes(name))
-              const failedRuns = matchingRuns.filter(({ conclusion }) => conclusion !== 'success')
+              const failedRuns = matchingRuns.filter(({ name, conclusion }) => !(conclusion == 'success' || (can_skip.includes(name) && conclusion == 'skipped')))
```
erezrokah, line 90:
```suggestion
              const failedRuns = matchingRuns.filter(({ name, conclusion }) => !(conclusion === 'success' || (canSkip.includes(name) && conclusion === 'skipped')))
```

`==` does type coercion in JS which we don't want

.github/workflows/validate_release.yml
```diff
@@ -0,0 +1,68 @@
...
+        description: 'Runner to be used (default: "ubuntu-latest")'
+        type:        string
+        required:    false
+        default:     'ubuntu-latest'
+      CGO_ENABLED:
+        description: 'CGO_ENABLED value to be used (default: 0)'
+        type:        number
+        required:    false
+        default:     0
+
```
erezrokah, line 20:
```suggestion
```

I don't think this is used anywhere, can you please verify?

erezrokah (changes requested):
Added a couple more comments on the code itself, though we should still ensure it works as expected per the previous comments

erezrokah:
Maybe we can add the reusable workflow without the skip logic at first to simplify things?

.github/workflows/validate_release.yml
candiduslynx: yep, leftover code

candiduslynx:
The issue is that the reusable workflow will still require the name `validate-release / validate-release` to be added, and the skip logic is really minor.
The logic issue we saw when a couple of runs with the same name (awspricing) were shown is due to the paging, I presume.

And the fact that we basically check for the following:
* any failed builds present? -> fail
* the len of filtered completed runs == the amount of runs we wait for -> success

In the run we had issue with we had the same entry counted twice, so that's why we skipped the wait and just went with it.

I will create a separate issue to fix the workflow behavior in the following way (I think it's non-blocking for now):
* account for the id of the run we see (so we skip all the runs we already processed)
* move from array to map (name -> amount) so that we properly check the runs for a specific action

candiduslynx:
@erezrokah I've extracted simple (no cgo) `valiate-release` for AWS Source plugin into #9780.
I'll be following up with cgo-releated ones after that one is merged, and then I'll create PRs to replace all others

erezrokah:
> I will create a separate issue to fix the workflow behavior in the following way (I think it's non-blocking for now):

I think if we have an issue with the wait for all logic, we should fix it first before making other CI changes, otherwise we won't be able to verify the changes we make.

Also I haven't seen that issue on other PRs, so we'll need to investigate.

candiduslynx:
> Also I haven't seen that issue on other PRs, so we'll need to investigate. We know what happens on this PR, but haven't confirmed the root cause, nor that it happens on all PRs

I think it might be caused by paging by 100 entries (as this is one of the PRs where we have a lot of actions running).
As 100 is [max](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference) we either need to verify that there's no issue with SDK or account for possible duplicates.

I really think that we can just do it in a separate issue and not make this blocking, as the issue was present even before (take any PR that updates some library for a lot of plugins).

erezrokah:
> There's no entry for plugins/source/aws being successful, but we possibly have some duplicate there because of the paging, that's why we just conclude successfully.

💯 Agree good catch, seems like there's a duplicate for `plugins/source/googleads`

> I really think that we can just do it in a separate issue and not make this blocking, as the issue was present even before (take any PR that updates some library for a lot of plugins).

Let's start with https://github.com/cloudquery/cloudquery/pull/9793 and log the IDs.

I don't think we should introduce non urgent (big) CI changes if we have a bug in our CI. The motivation for introducing re-usable workflows was to simplify the caching logic, but since then we discovered that we won't need to refactor it, and only fix 4 workflows.

So re-usable workflows seems to be more about removing code duplication (that we haven't had issues with before).
I don't see the urgency to push them

candiduslynx:
@erezrokah I've created #9797 to address the duplicates issue (possibly, just do the most obvious filter by ID) and make it working again.

erezrokah:
> @erezrokah I've created #9797 to address the duplicates issue (possibly, just do the most obvious filter by ID) and make it working again.

Can you spot duplicate IDs in https://github.com/cloudquery/cloudquery/actions/runs/4656875292/jobs/8240920196?pr=9736#step:4:381?

I did not find any

erezrokah:
I counted 60 validate release, 60 workflows (CLI, Scaffold and plugins), 4 test policies, which seems ok.

candiduslynx:
Drop for now, will take another shot after #9780

PR #9750, opened 2023-04-06
fix: Use aws.ToString(output.NextToken) when checking tokens

hermanschaaf:
This is a defensive change to ensure that we exit the loop regardless of whether the AWS SDK returns an empty string or a nil value for `NextToken`.


disq (approved):

plugins/source/aws/resources/services/ecs/cluster_container_instances.go
```diff
@@ -61,7 +62,7 @@ func fetchEcsClusterContainerInstances(ctx context.Context, meta schema.ClientMe
 
 		res <- describeContainerInstances.ContainerInstances
 
-		if listContainerInstances.NextToken == nil {
+		if aws.ToString(listContainerInstances.NextToken) == "" {
```
bbernays, line 65: I think ECS supports a paginator for `ListContainerInstances`

plugins/source/aws/resources/services/ecs/cluster_services.go
```diff
@@ -67,7 +68,7 @@ func fetchEcsClusterServices(ctx context.Context, meta schema.ClientMeta, parent
 
 		res <- describeServicesOutput.Services
 
-		if listServicesOutput.NextToken == nil {
+		if aws.ToString(listServicesOutput.NextToken) == "" {
```
bbernays, line 71: Maybe use the ECS `ListServices` paginator instead

hermanschaaf:
Merging the change for now - I will follow up later with changes to use paginators. I think there will be more cases where we can use paginators than just the services touched in this PR

PR #9751, opened 2023-04-06
feat(sqlite): Migrate to apache arrow type system

yevgenypats:
This still wait for another sdkv2 release for tests to pass but should be ready for initial review.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -120,5 +120,6 @@
   "plugins/source/googleads": "1.0.1",
   "plugins/source/googleads+FILLER": "0.0.0",
   "plugins/destination/meilisearch": "1.1.0",
-  "plugins/destination/meilisearch+FILLER": "1.0.0"
+  "plugins/destination/meilisearch+FILLER": "1.0.0",
+  "plugins/source/firestore": "1.0.0"
 }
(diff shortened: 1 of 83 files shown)


plugins/destination/sqlite/go.mod
```diff
@@ -9,13 +9,33 @@ require (
...
+	github.com/minio/c2goasm v0.0.0-20190812172519-36a3d3bbc4f3 // indirect
 	github.com/stretchr/testify v1.8.2 // indirect
+	github.com/zeebo/xxh3 v1.0.2 // indirect
 	golang.org/x/exp v0.0.0-20230321023759-10a507213a29 // indirect
+	golang.org/x/mod v0.8.0 // indirect
+	golang.org/x/tools v0.6.0 // indirect
+	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
 )
 
+replace github.com/cloudquery/plugin-sdk => ../../../../plugin-sdk
```
erezrokah, line 31: Can you maybe update this to point to a commit/branch so I can test?
Same with the other replace

erezrokah (commented):
Did a quick review of the code to make sure I get a general idea.
If you can update the `go.mod` to update to commit I can also test (or tell me which versions of SDK/Arrow to use).

erezrokah (approved):
Tested this on multiple sources (AWS, GCP and Azure) and upgrading between major versions using forced migration mode

PR #9754, opened 2023-04-06
fix(aws): Replace more manual pagination with Paginator objects

bbernays:
#### Summary

Rather than handling pagination manually, we should use the paginators that are part of the SDK


hermanschaaf (approved):
That's a lot of new paginators, nice work @bbernays !

PR #9757, opened 2023-04-06
fix: Fix case on arn attribute

gavinclarkeuk:
#### Summary

The MFA policy isn't linking any MFA devices to the root user because the case of the Arn attribute is incorrect.

Note that this query doesn't strictly enforce the hardware MFA policy, as it only checks virtual MFA devices. I don't believe there is anyway for CloudQuery to detect hardware MFA devices that are associated with the root user, as the AWS API requires you to be logged in as root to do that.


hermanschaaf (approved):
Thanks for the fix @gavinclarkeuk !

PR #9765, opened 2023-04-08
fix(aws): Replace Manually paginated calls with paginator where available

bbernays:
#### Summary

This PR uses the follow regex to find all paginated AWS calls: `aws.ToString\([A-Z a-z]*\.[A-Z a-z]*\) == ""`

Manually paginated calls were replaced with paginators. If a call doesn't support a paginator a comment was added


candiduslynx (approved):
Only a couple notes on places where using [`maps.Copy`](https://pkg.go.dev/golang.org/x/exp/maps#Copy) would make sense.
Can me left out for a follow-up PR.

erezrokah (approved):

Issue #9771, opened 2023-04-09
Cannot sync to Duckdb


pchaganti:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A sync from AWS to Duckdb fails with errors about more than one PK. Tried to skip the table, but this seems to be for  quite a few tables in my case.

```
Loading spec(s) from aws.yml, duckdb.yml
Starting migration with 526 tables for: aws (v15.7.0) -> [duckdb (v1.0.3)]
Error: failed to sync v1 source aws: failed to migrate source aws (v15.7.0) on destination duckdb (v1.0.3) : failed to call Migrate: rpc error: code = Unknown desc = failed to create table with 'CREATE TABLE IF NOT EXISTS "aws_apigateway_domain_name_base_path_mappings" ("_cq_source_name" text,"_cq_sync_time" timestamp,"_cq_id" uuid NOT NULL,"_cq_parent_id" uuid,"account_id" text PRIMARY KEY NOT NULL,"region" text,"domain_name_arn" text,"arn" text PRIMARY KEY NOT NULL,"base_path" text,"rest_api_id" text,"stage" text)': Parser Error: table "aws_apigateway_domain_name_base_path_mappings" has more than one primary key
```

It is happening for these tables too. Tried skipping thse as well. Tried out just syncing aws_s3_* and that fails as well.
```
    - aws_account_alternate_contacts
    - aws_apigateway_rest_api_deployments
    - aws_apigateway_rest_api_documentation_parts
    - aws_apigateway_rest_api_authorizers
    - aws_apigateway_domain_name_base_path_mappings
    - aws_apigateway_rest_api_documentation_versions
    - aws_apigateway_rest_api_gateway_responses
    - aws_apigateway_rest_api_models
    - aws_apigateway_rest_api_request_validators
    - aws_apigateway_rest_api_resources
                                                                                                                                                                                                                                                                                                                                                                                                              
```


### Expected Behavior

Completed sync.

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: aws
  path: cloudquery/aws
  version: "v15.7.0"
  tables: ["*"]
  skip_tables:
    - aws_ec2_vpc_endpoint_services # this resource includes services that are available from AWS as well as other AWS Accounts
     - aws_account_alternate_contacts
    - aws_apigateway_rest_api_deployments
    - aws_apigateway_rest_api_documentation_parts
    - aws_apigateway_rest_api_authorizers
[...]

PR #9772, opened 2023-04-09
fix(duckdb): Support multiple PKs

yevgenypats:
Closes https://github.com/cloudquery/cloudquery/issues/9771


plugins/destination/duckdb/client/migrate.go
```diff
@@ -215,7 +214,16 @@ func (c *Client) createTableIfNotExist(table *schema.Table) error {
 			sb.WriteString(",")
 		}
 	}
-
+	if len(pks) > 0 && c.enabledPks() {
```
erezrokah, line 217:
```suggestion
	if pks := table.PrimaryKeys(); len(pks) > 0 && c.enabledPks() {
```

No blocking but we could do something like this

erezrokah (approved):

PR #9780, opened 2023-04-09
feat(ci): Reuse `validate-release` in AWS Source plugin workflow

candiduslynx:
Extracted from #9736

### DIFF
--- a/.github/workflows/validate_release.yml
+++ b/.github/workflows/validate_release.yml
@@ -12,6 +12,11 @@ on:
         type:        string
         required:    false
         default:     'ubuntu-latest'
+    secrets:
+      GORELEASER_KEY:
+        description: 'GoReleaser key'
+        required:    true
+
 jobs:
   validate-release:
     # this if check is here as we check for 'validate-release / validate-release' run
--- a/.github/workflows/validate_release.yml
+++ b/.github/workflows/validate_release.yml
@@ -39,9 +39,7 @@ jobs:
       id:   set-up-go
       uses: actions/setup-go@v3
       with:
-        go-version-file:       ${{ inputs.path }}/go.mod
-        cache:                 true
-        cache-dependency-path: ${{ inputs.path }}/go.sum
+        go-version-file: ${{ inputs.path }}/go.mod
     - name: Install GoReleaser
       uses: goreleaser/goreleaser-action@v3
       with:
--- a/.github/workflows/validate_release.yml
+++ b/.github/workflows/validate_release.yml
@@ -4,7 +4,7 @@ on:
   workflow_call:
     inputs:
       path:
-        description: 'Go mod directory (e.g. "plugins/source/aws", "plugins/destination/csv", "cli")'
+        description: 'Go mod directory (e.g. "plugins/source/aws", "plugins/destination/file", "cli")'
         type:        string
         required:    true
       runner:
--- a/.github/workflows/validate_release.yml
+++ b/.github/workflows/validate_release.yml
@@ -1,0 +1,62 @@
+name: Validate release
+
+on:
+  workflow_call:
+    inputs:
+      path:
+        description: 'Go mod directory (e.g. "plugins/source/aws", "plugins/destination/csv", "cli")'
+        type:        string
+        required:    true
+      runner:
+        description: 'Runner to be used (default: "ubuntu-latest")'
+        type:        string
+        required:    false
+        default:     'ubuntu-latest'
+jobs:
+  validate-release:
+    # this if check is here as we check for 'validate-release / validate-release' run
+    if:              startsWith(github.head_ref, 'release-please--branches--main--components') || github.event_name == 'push'
+    timeout-minutes: 60
+    runs-on:         ${{ inputs.runner }}
+    env:
+      CGO_ENABLED: 0
+    steps:
+    - id:  dash-path
+      run: |
+           echo "dashed=$(echo ${{ inputs.path }} | tr '/' '-')" >> $GITHUB_OUTPUT
+    - name: Checkout
+      uses: actions/checkout@v3
+    - uses: actions/cache@v3
+      with:
+        path:         |
+                      ~/.cache/go-build
+                      ~/go/pkg/mod
+        key:          ${{ runner.os }}-go-1.19.5-release-cache-${{ hashFiles(format('{0}/go.sum', inputs.path)) }}
+        # actually, this value is never ever saved, but left for consistency (renaming cache keys os out of scope)
+        restore-keys: |
+                      ${{ runner.os }}-go-1.19.5-release-cache-${{ steps.dash-path.outputs.dashed }}
+    - name: Set up Go
+      id:   set-up-go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file:       ${{ inputs.path }}/go.mod
+        cache:                 true
+        cache-dependency-path: ${{ inputs.path }}/go.sum
+    - name: Install GoReleaser
+      uses: goreleaser/goreleaser-action@v3
+      with:
+        distribution: goreleaser-pro
+        version:      latest
+        install-only: true
+    - name: Run GoReleaser Dry-Run
+      run:  >
+            goreleaser release
+            --timeout 50m
+            --snapshot
+            --clean
+            --skip-validate
+            --skip-publish
+            --skip-sign
+            -f ${{ inputs.path }}/.goreleaser.yaml
+      env:
+        GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/wait_for_required_workflows.yml
```diff
@@ -51,7 +52,8 @@ jobs:
 
             // Most modules should have a 'validate-release' job
             for (const action of actions) {
-              actions = [...actions, 'validate-release']
+              // add 'validate-release / validate-release' as we use reusable with same job id
```
erezrokah, line 55: I believe this needs to be in a condition and only use `'validate-release / validate-release'` if it's the AWS plugin and use `'validate-release'` for others, WDYT?

.github/workflows/validate_release.yml
```diff
@@ -0,0 +1,62 @@
+name: Validate release
+
+on:
+  workflow_call:
+    inputs:
+      path:
+        description: 'Go mod directory (e.g. "plugins/source/aws", "plugins/destination/csv", "cli")'
```
erezrokah, line 7:
```suggestion
        description: 'Go mod directory (e.g. "plugins/source/aws", "plugins/destination/file", "cli")'
```

The CSV plugin is deprecated so let's not use it in an example

.github/workflows/validate_release.yml
```diff
@@ -0,0 +1,62 @@
...
+    # this if check is here as we check for 'validate-release / validate-release' run
+    if:              startsWith(github.head_ref, 'release-please--branches--main--components') || github.event_name == 'push'
+    timeout-minutes: 60
+    runs-on:         ${{ inputs.runner }}
+    env:
+      CGO_ENABLED: 0
+    steps:
+    - id:  dash-path
+      run: |
+           echo "dashed=$(echo ${{ inputs.path }} | tr '/' '-')" >> $GITHUB_OUTPUT
```
erezrokah, line 26:
```suggestion
```

I think we can drop this dashed logic and just used the value as is WDYT? It's used only for the fallback release cache key so we could simplify this logic

.github/workflows/validate_release.yml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Checkout
+      uses: actions/checkout@v3
+    - uses: actions/cache@v3
+      with:
+        path:         |
+                      ~/.cache/go-build
+                      ~/go/pkg/mod
+        key:          ${{ runner.os }}-go-1.19.5-release-cache-${{ hashFiles(format('{0}/go.sum', inputs.path)) }}
+        # actually, this value is never ever saved, but left for consistency (renaming cache keys os out of scope)
```
erezrokah, line 35: I don't understand this comment. When a cache is saved, it's saved both to `key` and each `restore-key`.
When the cache is restored, and there's a cache miss on `key` the `restore-keys` will be used in order.

.github/workflows/validate_release.yml
```diff
@@ -0,0 +1,62 @@
...
+        restore-keys: |
+                      ${{ runner.os }}-go-1.19.5-release-cache-${{ steps.dash-path.outputs.dashed }}
+    - name: Set up Go
+      id:   set-up-go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file:       ${{ inputs.path }}/go.mod
+        cache:                 true
+        cache-dependency-path: ${{ inputs.path }}/go.sum
```
erezrokah, line 44:
```suggestion
        go-version-file:       ${{ inputs.path }}/go.mod
```

We should not enable caching for `setup-go` if we're using `actions/cache`

.github/workflows/validate_release.yml
```diff
@@ -0,0 +1,62 @@
...
+            goreleaser release
+            --timeout 50m
+            --snapshot
+            --clean
+            --skip-validate
+            --skip-publish
+            --skip-sign
+            -f ${{ inputs.path }}/.goreleaser.yaml
+      env:
+        GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
```
erezrokah, line 62: I think you need to explicitly pass the secret to the re-usable workflow for this to work.
See it gets an empty value:
https://github.com/cloudquery/cloudquery/actions/runs/4651754791/jobs/8231547483?pr=9780#step:7:6
while in a workflow that works you see a hidden `***` value:
https://github.com/cloudquery/cloudquery/actions/runs/4653806372/jobs/8234992525?pr=9783#step:6:6

More in https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

erezrokah (changes requested):
Thanks for scoping down this to a single plugin @candiduslynx.

Added a few comments. I think this PR will break the wait for all logic for all other plugins, can you please take a look?

.github/workflows/wait_for_required_workflows.yml
candiduslynx: good catch!

.github/workflows/validate_release.yml
candiduslynx: I don't think the cache key can include `/` symbol (that's why the dashed logic is here).
Additionally, this step allows us to have essentially the same `restore-keys` that are now hardcoded (in plugin name part sense)

.github/workflows/validate_release.yml
candiduslynx: Cache isa saved _only_ for `key`.
`restore-keys` is useful to match (1 by 1 until the match is found) partially useful caches.
In this case, the `cache-hit` is `false` (as we didn't find the cache designated by `key`), so the cache will be saved in `post cache` step.

However, the `restore-keys` are just a list of possible caches that we might find a bit useful, and are usually listed from the most narrow scope to the most wide one.

Let's consider this example:
* we save cache by `key: a-b-c-d-hash(some-files)`
* we have somewhat related hashed that were saved with the same key pattern, but for different files (different file states)

Then, the following pattern can be used:

```yaml
key: a-b-c-d-hash(some-files)
restore-keys: |-
  a-b-c-d-
  a-b-c-
  a-b-
  a-
```

This allows us to restore the most matching cache entry in our store (if there's no `a-b-c-d-XYZ` entry, but we have `a-b-c-XYZ`, then this entry will be restore and the check process will stop).

.github/workflows/validate_release.yml
candiduslynx: I can just use `inherited` secrets, though.
Thanks for the catch!

candiduslynx: https://github.com/cloudquery/cloudquery/actions/runs/4656236159/jobs/8239666046?pr=9780#step:7:6

erezrokah: > I can just use inherited secrets, though.

Better not to implicitly pass all secrets. So https://github.com/cloudquery/cloudquery/pull/9780/commits/661c4800ea2c4e79be944bc34dee398d9e302a3f looks good

.github/workflows/validate_release.yml
erezrokah: Cool, thanks for the context. So let's remove it altogether? That should simplify the workflow and remove the need for the dash logic?

.github/workflows/validate_release.yml
erezrokah: See my comment here https://github.com/cloudquery/cloudquery/pull/9780#discussion_r1161604359

.github/workflows/validate_release.yml
candiduslynx: I just wanted to leave this consistent with the code I'm replacing (as all `validate-release` cache steps use `restore-keys`).

I can remove the restore part, however, the restore-keys has some benefits to be aware when we discuss cache strategy further (basically, having dirty cache is better that not to have any cache at all).

candiduslynx: So, should I remove this part for now and we're OK with this being inconsistent with the `validate-release` for now (although we know for sure that the `restore-keys` part doesn't have any hit due to the fact that we don't save any cache prefixed that way).
@erezrokah could you confirm?

.github/workflows/validate_release.yml
candiduslynx: removed in 07016b2

candiduslynx:
@erezrokah I think this is slimmed down as far as it can be, let's merge and then proceed with the following path:
* cgo-enabled plugins (duckdb, snowflake, ...)
* other validate-release replacements

erezrokah:
> @erezrokah I think this is slimmed down as far as it can be, let's merge and then proceed with the following path:

I think it looks good now, but as I mentioned here https://github.com/cloudquery/cloudquery/pull/9736#issuecomment-1501666150 let's fix the CI bug first. It shouldn't take than long to verify and fix since now we have the IDs logged.
So if it's a duplicate we can make the array unique, and if it's not a duplicate we should investigate more why we have duplicate runs

erezrokah (approved):

PR #9793, opened 2023-04-10
chore(ci): Log check run ID to detect duplicates

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/pull/9736#issuecomment-1501606356. We've seen cases were apparently the GitHub JS SDK returns duplicate check runs, so adding a log for the ID to see what's up.

### DIFF
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -78,7 +78,7 @@ jobs:
                 status: 'completed',
                 per_page: 100
               })
-              const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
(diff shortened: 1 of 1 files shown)


hermanschaaf (approved):

.github/workflows/wait_for_required_workflows.yml
```diff
@@ -78,7 +78,7 @@ jobs:
                 status: 'completed',
                 per_page: 100
               })
-              const runs = checkRuns.map(({ name, conclusion }) => ({ name, conclusion }))
+              const runs = checkRuns.map(({ id, name, conclusion }) => ({ id, name, conclusion }))
```
erezrokah, line 81: I wonder why the diff shows the first id with the space after the comma, and the second id with the space before 🤷

PR #9798, opened 2023-04-10
fix(ci): Filter out duplicates in `wait-for-required-workflows`

candiduslynx:
Follow-up from #9793

### DIFF
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -78,7 +78,11 @@ jobs:
                 status: 'completed',
                 per_page: 100
               })
-              const runs = checkRuns.map(({ id, name, conclusion }) => ({ id, name, conclusion }))
+
+              // filter as sometimes we can see duplicates (1st page fetch -> new check finished -> 2nd page fetch)
+              const rawRuns = checkRuns.map(({ id, name, conclusion }) => ({ id, name, conclusion }))
+              const runs = rawRuns.filter((element, index) => rawRuns.indexOf(element) === index)
+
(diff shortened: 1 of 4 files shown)


.github/workflows/wait_for_required_workflows.yml
```diff
@@ -78,7 +78,11 @@ jobs:
                 status: 'completed',
                 per_page: 100
               })
-              const runs = checkRuns.map(({ id, name, conclusion }) => ({ id, name, conclusion }))
+
+              // filter as sometimes we can see duplicates (1st page fetch -> new check finished -> 2nd page fetch)
+              const rawRuns = checkRuns.map(({ id, name, conclusion }) => ({ id, name, conclusion }))
+              const runs = rawRuns.filter((element, index) => rawRuns.indexOf(element) === index)
```
erezrokah, line 84: I don't think this logic works, as `indexOf` uses strict equality, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf#description
so it returns the same index only if the item points to the same object, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality#comparing_objects.

I suggest we first look at the logs when a duplicate occurs and see if the IDs are the same, then we can use that to remove duplicates. If the IDs are not the same, and the name is we should investigate further

erezrokah (changes requested):
Let's only do the workflow names fix for now

.github/workflows/wait_for_required_workflows.yml
candiduslynx: reverted

PR #9800, opened 2023-04-10
fix: Revert "fix(ci): Fix workflow names"

erezrokah:
Reverts cloudquery/cloudquery#9798


erezrokah:
Done in https://github.com/cloudquery/cloudquery/pull/9799/files

PR #9802, opened 2023-04-10
refactor(aws): Consolidate IoT Tag Resolvers

bbernays:
#### Summary

use single function to make API calls rather than reimplementing for each resource


erezrokah (approved):
Looks good, but should probably be a `refactor:` if doesn't have any user facing changes

disq (approved):

PR #9804, opened 2023-04-10
feat(aws): Parallelize resolution of `aws_iot_streams`

bbernays:
#### Summary

Summary
Use `PreResourceResolver` to call `DescribeStream` rather than calling sequentially

### DIFF
--- a/plugins/source/aws/resources/services/iot/streams.go
+++ b/plugins/source/aws/resources/services/iot/streams.go
@@ -14,11 +14,12 @@ import (
 func Streams() *schema.Table {
 	tableName := "aws_iot_streams"
 	return &schema.Table{
-		Name:        tableName,
-		Description: `https://docs.aws.amazon.com/iot/latest/apireference/API_StreamInfo.html`,
-		Resolver:    fetchIotStreams,
-		Transform:   transformers.TransformWithStruct(&types.StreamInfo{}),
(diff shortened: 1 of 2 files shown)


erezrokah (approved):

plugins/source/aws/resources/services/iot/streams.go
```diff
@@ -45,20 +46,21 @@ func fetchIotStreams(ctx context.Context, meta schema.ClientMeta, parent *schema
 		if err != nil {
 			return err
 		}
-		for _, s := range page.Streams {
-			// TODO: Handle resolution in parallel with PreResourceResolver
-			stream, err := svc.DescribeStream(ctx, &iot.DescribeStreamInput{
-				StreamId: s.StreamId,
-			}, func(options *iot.Options) {
-				options.Region = c.Region
```
erezrokah, line 53: Don't we need this function to set the region?

erezrokah (changes requested):
One question on the change

plugins/source/aws/resources/services/iot/streams.go
bbernays: No, these manual settings of region are a relic that never got cleaned up.

`c.Region` is always the same region that is set in the service initialization

erezrokah (approved):

PR #9812, opened 2023-04-10
fix(workflows): Properly use caches in AWS, Azure, GCP, K8s workflows

### LINKED ISSUE #9388: bug(aws,gcp,k8s,azure): Test caching in CI does not work

candiduslynx:
Short way instead of #9656.
Closes #9388

### DIFF
--- a/.github/workflows/source_aws.yml
+++ b/.github/workflows/source_aws.yml
@@ -141,7 +141,7 @@ jobs:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
-          cache-prefix: policies-cache-
+          cache-key-prefix: policies-cache-
       - name: Build
         run: go build .
       - name: Setup CloudQuery
--- a/.github/workflows/source_aws.yml
+++ b/.github/workflows/source_aws.yml
@@ -33,11 +33,12 @@ jobs:
         with:
           fetch-depth: 2
       - name: Set up Go 1.x
-        uses: actions/setup-go@v3
+        uses: erezrokah/setup-go@feat/add_cache_prefix
         with:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
+          cache-prefix: test-cache-
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
@@ -135,11 +136,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Go 1.x
-        uses: actions/setup-go@v3
+        uses: erezrokah/setup-go@feat/add_cache_prefix
         with:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
+          cache-prefix: policies-cache-
       - name: Build
         run: go build .
       - name: Setup CloudQuery
--- a/.github/workflows/source_aws.yml
+++ b/.github/workflows/source_aws.yml
@@ -38,7 +38,7 @@ jobs:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
-          cache-prefix: test-cache-
+          cache-key-prefix: test-cache-
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/source_aws.yml
```diff
@@ -135,11 +136,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Go 1.x
-        uses: actions/setup-go@v3
+        uses: erezrokah/setup-go@feat/add_cache_prefix
         with:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
+          cache-prefix: policies-cache-
```
erezrokah, line 144:
```suggestion
          cache-key-prefix: policies-cache-
```

.github/workflows/source_azure.yml
```diff
@@ -117,11 +118,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Go 1.x
-        uses: actions/setup-go@v3
+        uses: erezrokah/setup-go@feat/add_cache_prefix
         with:
           go-version-file: plugins/source/aws/go.mod
           cache: true
           cache-dependency-path: plugins/source/aws/go.sum
+          cache-prefix: policies-cache-
```
erezrokah, line 126:
```suggestion
          cache-key-prefix: policies-cache-
```

.github/workflows/source_k8s.yml
```diff
@@ -117,11 +118,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Set up Go 1.x
-        uses: actions/setup-go@v3
+        uses: erezrokah/setup-go@feat/add_cache_prefix
         with:
           go-version-file: plugins/source/k8s/go.mod
           cache: true
           cache-dependency-path: plugins/source/k8s/go.sum
+          cache-prefix: policies-cache-
```
erezrokah, line 126:
```suggestion
          cache-key-prefix: policies-cache-
```

erezrokah (changes requested):
I thinks something is not working with my cache key fix, see https://github.com/cloudquery/cloudquery/actions/runs/4659445116/jobs/8246354513?pr=9812#step:14:7

image

Maybe try directly with the cache action instead of my fork?

candiduslynx:
> I thinks something is not working with my cache key fix, see https://github.com/cloudquery/cloudquery/actions/runs/4659445116/jobs/8246354513?pr=9812#step:14:7
> 
> image
> 
> Maybe try directly with the cache action instead of my fork?

I think it'll be fixed by d685b03

erezrokah:
It happens on GCP too:
https://github.com/cloudquery/cloudquery/actions/runs/4659445149/jobs/8246354731#step:14:7
https://github.com/cloudquery/cloudquery/actions/runs/4659445149/jobs/8246353443#step:17:7
image

candiduslynx:
@erezrokah it's fine now?

erezrokah:
> @erezrokah it's [fine](https://github.com/cloudquery/cloudquery/actions/runs/4659711753/jobs/8246902366?pr=9812) now?

This shows a cache hit, so not the same test as it skips saving the cache. Can we find the workflow that saved that cache successfully to confirm? Or delete the cache manually and confirm each job saves it successfully? Another option to test is to add `v1` or something to the prefix to create a new key.

candiduslynx:
I'm removing the caches designated via [this](https://github.com/cloudquery/cloudquery/actions/caches?query=branch%3Arefs%2Fpull%2F9812%2Fmerge) branch and rerunning

candiduslynx:
Now triggering build

erezrokah:
Also https://github.com/cloudquery/cloudquery/pull/9812/commits/d4f2f23542a59284b155ef995f43c40d25dc8168. If it doesn't pass we can descope it from this PR

erezrokah (approved):

PR #9813, opened 2023-04-10
feat(workflows): Refactor `wait-for-required-workflows`

### LINKED ISSUE #9797: bug: Duplicates in `wait-for-required-workflows` resulting in false-positive success

### Current Behavior

When we check for the finished checks we use paging to get all of the runs.
In the situations where we need to fetch > 100 (page size) runs, we sometimes can encounter duplicates, and the workflow finished too fast.

### Expected Behavior

Even the duplicate ID shouldn't be a problem.

N/A

### Steps To Reproduce

_No response_

N/A

### CloudQuery version

N/A

### Additional Context

https://github.com/cloudquery/cloudquery/pull/9736#issuecomment-1501173716
https://github.com/cloudquery/cloudquery/pull/9736#issuecomment-1501606356
https://github.com/cloudquery/cloudquery/pull/9793

### Pull request (optional)

candiduslynx:
Closes #9797


erezrokah (changes requested):
See my comment here https://github.com/cloudquery/cloudquery/issues/9797#issuecomment-1502035204

I don't understand how we can submit a fix for this issue until we confirm we can de-duplicate using IDs

candiduslynx:
> See my comment here [#9797 (comment)](https://github.com/cloudquery/cloudquery/issues/9797#issuecomment-1502035204)

As we saw (without IDs logged, but fairly obvious due to the name uniqueness) the duplicated check run from paged API, the only suitable conclusion is the following:
* page 1 was fetched
* another check finished
  * vital: it had to fall into the 1st page
* we fetched the 2nd page, that would now start from the same entry the 1st one ended

> I don't understand how we can submit a fix for this issue until we confirm we can de-duplicate using IDs

See API [doc](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#get-a-check-run) for getting check run (ID is sufficient key).

erezrokah:
> * it tracks the workflows using maps and will fail if we ever encounter extra check (like, we wait for 50 `validate-release` and receive 51)

I think we can agree this code is already complex and error prone (as seen in the time spend on recent PRs) so I think it's better to identify the issue first, than do any code changes later

candiduslynx:
@erezrokah See https://github.com/cloudquery/cloudquery/issues/9797#issuecomment-1502662486 for the repro

erezrokah:
We discussed this in a Zoom call, and figured that once we add re-usable workflows we should be able to give each plugin validate release and test policies job a distinct name such as  `validate release / validate release aws` so we could do an array equals instead of a length check here:
https://github.com/cloudquery/cloudquery/blob/34b1569d17adeb14eb3d2694b7a4a4e9a5c52083/.github/workflows/wait_for_required_workflows.yml#L90

That should make the wait for all workflow more robust and also allow us to log and know which plugins are waiting for workflows with the same name

PR #9822, opened 2023-04-11
fix(ci): Remove duplicated check runs

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/9797. Follow up to https://github.com/cloudquery/cloudquery/pull/9793 and replaces https://github.com/cloudquery/cloudquery/pull/9813.

https://github.com/cloudquery/cloudquery/issues/9797#issuecomment-1502662486 Confirms the ID is duplicated and we can filter using it

### DIFF
--- a/.github/workflows/wait_for_required_workflows.yml
+++ b/.github/workflows/wait_for_required_workflows.yml
@@ -78,7 +78,8 @@ jobs:
                 status: 'completed',
                 per_page: 100
               })
-              const runs = checkRuns.map(({ id, name, conclusion }) => ({ id, name, conclusion }))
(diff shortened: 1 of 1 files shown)


erezrokah:
> I guess #9813 also addressed the issue fixed by #9798 (as we'll still fail if there's another unmatched workflow that triggers `validate-release` or the likes).
> So, I wonder if we want to address that issue, too, or just leave as is.

Since we haven't see that issue yet (duplicate names with the same ID), I'm not sure we should add code to fix it

candiduslynx (approved):

PR #9832, opened 2023-04-11
feat(ci): Check for dependencies to wait

candiduslynx:


.github/workflows/wait_for_required_workflows.yml
```diff
@@ -30,12 +30,28 @@ jobs:
...
+              return false
+            }
+
             const matchesWorkflow = (file, action) => {
               if (!file.startsWith('.github/workflows/')) {
                 return false
               }
               const contents = fs.readFileSync(file, 'utf8');
-              return contents.includes(`name: "${action}"`)
+              return contents.includes(`name: "${action}"`) || matchesByDependency(file, contents)
```
erezrokah, line 54: I'm not sure this will work, as `matchesWorkflow` iterates on files only changes by this PR.
To correctly find all dependent workflows we need to scan all workflow files in the repository

erezrokah (changes requested):
Let's start with the revert https://github.com/cloudquery/cloudquery/pull/9834.

I think it will take more than a few minutes to handle re-usable workflow dependencies in the wait for all logic

.github/workflows/wait_for_required_workflows.yml
candiduslynx: yep, already stumbled here

PR #9834, opened 2023-04-11
fix(ci): Revert "feat(ci): Reuse `validate-release` in AWS Source plugin workflow"

erezrokah:
Reverts cloudquery/cloudquery#9780

See https://github.com/cloudquery/cloudquery/issues/9655#issuecomment-1503122325. We should first take a look at that comment before using re-usable workflows


erezrokah:
CI is failing because we don't account for workflow file deletions

erezrokah:
> CI is failing because we don't account for workflow file deletions

Should be fixed by https://github.com/cloudquery/cloudquery/pull/9835

candiduslynx (approved):
to be rereverted after we fix the dependency wait

PR #9852, opened 2023-04-12
feat(meilisearch): Upgrade to `github.com/cloudquery/plugin-sdk/v2`

candiduslynx:
Implements #9849


plugins/destination/meilisearch/client/read.go
```diff
@@ -45,17 +57,127 @@ func (c *Client) Read(_ context.Context, table *schema.Table, sourceName string,
 	return nil
 }
 
-func unmap(table *schema.Table) func(hit any) ([]any, error) {
-	return func(hit any) ([]any, error) {
-		m, ok := hit.(map[string]any)
-		if !ok {
-			return nil, fmt.Errorf("unsupported format for doc: %T", hit)
-		}
+func unmap(sc *arrow.Schema) func(m map[string]any) (arrow.Record, error) {
```
yevgenypats, line 60: What is this function - `unmap` ? can we avoid using it and use `arrow.Field` directly without having a function that return a function?

yevgenypats (changes requested):

plugins/destination/meilisearch/client/read.go
candiduslynx: We read `map[string]any` (JSON) from Meilisearch and need to transform this to the comparable (record) format.

candiduslynx: Moved from closure to separate func in 8faf82a

plugins/destination/meilisearch/client/read.go
```diff
@@ -57,127 +56,125 @@ func (c *Client) Read(_ context.Context, sc *arrow.Schema, sourceName string, re
...
+		case *array.Float32Builder:
+			tBuilder.Append(float32(val.(float64)))
+		case *array.Float64Builder:
+			tBuilder.Append(val.(float64))
+		case *array.StringBuilder:
+			tBuilder.Append(val.(string))
+		case *array.BinaryBuilder:
+			val := val.(string)
+			if val == "null" {
+				tBuilder.Append(nil)
```
disq, line 98: Missing `continue` here

candiduslynx: nice find! Fixed by e5b39cd

plugins/destination/meilisearch/client/read.go
```diff
@@ -45,17 +56,124 @@ func (c *Client) Read(_ context.Context, table *schema.Table, sourceName string,
...
+				return nil, err
+			}
+			tBuilder.Append(mac)
+		case *types.UUIDBuilder:
+			uid, err := uuid.Parse(val.(string))
+			if err != nil {
+				return nil, err
+			}
+			tBuilder.Append(uid)
+		case *array.ListBuilder:
```
disq, line 131: Nested handling can be achieved by recursion, would need a bit of a refactor though

candiduslynx: Addressed in 6523871

plugins/destination/meilisearch/client/transform.go
```diff
@@ -2,179 +2,141 @@ package client
...
+		builder.Append(ts)
+	case *types.InetBuilder:
+		_, ipNet, err := net.ParseCIDR(val.(string))
+		if err != nil {
+			return err
+		}
+		builder.Append(*ipNet)
+	case *types.JSONBuilder:
+		builder.Append(val)
+	case *types.MacBuilder:
```
yevgenypats, line 112: This is not needed. you can just use `default` `AppendFromString`

plugins/destination/meilisearch/client/transform.go
```diff
@@ -2,179 +2,141 @@ package client
...
-		res[i] = t
+		builder.Append(data)
+	case *array.TimestampBuilder:
+		ts, err := arrow.TimestampFromString(val.(string), arrow.Microsecond)
+		if err != nil {
+			return err
+		}
+		builder.Append(ts)
+	case *types.InetBuilder:
+		_, ipNet, err := net.ParseCIDR(val.(string))
```
yevgenypats, line 105: This is not needed, can be safely deleted and `default` will take care of that

plugins/destination/meilisearch/client/transform.go
```diff
@@ -2,179 +2,141 @@ package client
...
+	case *types.JSONBuilder:
+		builder.Append(val)
+	case *types.MacBuilder:
+		mac, err := net.ParseMAC(val.(string))
+		if err != nil {
+			return err
+		}
+		builder.Append(mac)
+	case *types.UUIDBuilder:
+		uid, err := uuid.Parse(val.(string))
```
yevgenypats, line 119: Same. can be safely deleted as there is no special type for `uuid` in meilisearch and `default` case will take care of that

yevgenypats (changes requested):
Looks good. First round of review. found a few low hanging fruits.

plugins/destination/meilisearch/client/transform.go
candiduslynx: Addressed in cb52e9b

plugins/destination/meilisearch/client/transform.go
candiduslynx: Addressed in cb52e9b

plugins/destination/meilisearch/client/transform.go
candiduslynx: Addressed in cb52e9b

yevgenypats (approved):

PR #9923, opened 2023-04-13
fix(azure): Fix detecting conditions in CIS Section 8

aruneko:
#### Summary
I noticed that policies of CIS Azure v1.3.0 Section 8.1 and 8.2 always return `fail`. I add 3 changes the sql conditions mainly.

1. use sub-resoruce (keyvault_key and keyvault_secret) id as the `resource_id` column in `azure_policy_results` table.
    - That makes it easier to identify what does resource lack expiration date directly.
1. fix `fail` condition.
    - adjusting the original documentation of CIS Benchmark. It says that the failing condition is "enabled is true and expiration date is null".
1. use `JOIN` instead of `LEFT JOIN`.
    - If we use `LEFT JOIN` on an Azure environment which don't have key or secret in a key vault, we get null columns and it satisfies a failing condition.


erezrokah (approved):
Thanks @aruneko, this looks good to me 🚀  Much cleaner than before

PR #9947, opened 2023-04-13
chore(docs): Consolidate formatting for configs

hermanschaaf:
Removes some extra spacing present in a few plugin configs, bringing all configs into alignment.


erezrokah (approved):
Looks great, low priority follow up is to maybe do something like https://github.com/cloudquery/cloudquery/blob/main/.github/workflows/auto_sort_files.yml to align the configs once in a while (using prettier or any other formatter)

candiduslynx (approved):

PR #9972, opened 2023-04-13
feat(snyk): Rename `snyk_reporting_issues` to `snyk_reporting_latest_issues`, add new `snyk_reporting_issues` table

hermanschaaf:
This renames the previous `snyk_reporting_issues` table to `snyk_reporting_latest_issues` to better align with its name in the API, and adds an additional table called `snyk_reporting_issues` which allows for the configuration of time window for which issues should be fetched.

This would normally be a breaking change, but we haven't released the `snyk_reporting_issues` table yet, so I'm not marking it as such.


plugins/source/snyk/client/spec.go
```diff
@@ -21,11 +29,92 @@ type Spec struct {
...
+		if err != nil {
+			return fmt.Errorf("invalid to date: %w", err)
+		}
+	}
+	if t.SnykReportingIssues.Period != "" {
+		if !reValidPeriod.MatchString(t.SnykReportingIssues.Period) {
+			return fmt.Errorf("invalid period: %s", t.SnykReportingIssues.Period)
+		}
+	}
+	if t.SnykReportingIssues.To != "" && t.SnykReportingIssues.From == "" {
```
candiduslynx, line 58: Shouldn't it be flipped?
Judging by `SnykReportingIssuesOptions.FromTime`, `From == "" && Period != ""` is fine

hermanschaaf: I think it's correct; this case prevents setting `to` if `from` is not also set. We accept these combinations:
 - `from` + `to`
 - `from` (default `to` is now)
 - `period` (relative to now)
 
We do not accept `period` combined with either `to` or `from`, and we don't accept `to` without `from`.  Does that make sense? Do you still think there is something wrong here?

PR #9995, opened 2023-04-13
fix(website): Update Broken link

bbernays:
#### Summary

Title of section should be singular


website/pages/docs/plugins/sources/aws/configuration.md
```diff
@@ -130,7 +130,7 @@ This is the (nested) spec used by the AWS source plugin.
   The region that should be used for signing the request to the endpoint
 
 
-## accounts
+## account
```
bbernays, line 133: Changing this from plural to singular because all of the org references are to the singular

PR #9653, opened 2023-04-04
feat(firestore): Add firestore source

BubbaJoe:
#### Summary
Adds firebase source #9414 

- [x] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `make lint` to ensure the proposed changes follow the coding style 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Run `make test` to ensure the proposed changes pass the tests 🧪
- [x] If changing a source plugin run `make gen` to ensure docs are up to date 📝
- [x] Ensure the status checks below are successful ✅

PR #10007, opened 2023-04-13
feat(website): Add Firestore

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/9653. I'll need some help to confirm the docs, cc @BubbaJoe 

Docs in https://cloudquery-oo1n5e8ec-cloudquery.vercel.app/docs/plugins/sources/firestore/overview

### DIFF
--- a/plugins/source/firestore/client/client.go
+++ b/plugins/source/firestore/client/client.go
@@ -13,13 +13,13 @@ import (
 )
 
 type Client struct {
-	logger           zerolog.Logger
-	metrics          *source.Metrics
-	Tables           schema.Tables
-	client           *firestore.Client
-	maxBatchSize     int
-	orderByField     string
-	orderByDirection string
+	logger         zerolog.Logger
+	metrics        *source.Metrics
+	Tables         schema.Tables
+	client         *firestore.Client
+	maxBatchSize   int
+	orderBy        string
+	orderDirection string
 }
 
 var _ schema.ClientMeta = (*Client)(nil)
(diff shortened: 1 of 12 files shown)


BubbaJoe:
Thanks, this already looks really well done! Can you add docs for the following fields?

MaxBatchSize - when testing extract millions of rows. If the request took longer than a minute the batch would timeout. So I set the default to 50,000 IIRC. This might need to be configured depending on bandwidth + cpu.

Also OrderByField - this is optional but you can get your results ordered by a specific fields (yes, this can be more than 1, maybe I should have made it plural) of course this requires you to add an index on firebase which can only be done on the UI right now.


OrderByDirection - not used unless the above is field is specified but I think this is self explanatory.

BubbaJoe:
@erezrokah Forgot to tag you here ^

erezrokah:
Thanks @BubbaJoe, that's super helpful. I followed up with https://github.com/cloudquery/cloudquery/pull/10007/commits/7f7925f9cb09645d5ccf0c4b5976d47abdd1305a and renamed those to `order_by` and `order_direction`. Do that make sense?

Also it seems the spec values were not used so I fixed that, see https://github.com/cloudquery/cloudquery/pull/10007/commits/7f7925f9cb09645d5ccf0c4b5976d47abdd1305a#diff-2900389b353248fc48ac304a58d98219fa4c209abf43ccb456a43f4952db22f5R63

Please let me know if that looks OK.

BubbaJoe:
Thanks! Looks really good!

disq (approved):

erezrokah:
Officially released, thanks @BubbaJoe see https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-firestore-v1.0.0

PR #10012, opened 2023-04-14
fix(azure): Improve the query of CIS Azure v1.3.0 Section 7.7

aruneko:
#### Summary
Use azure_compute_disks table to extract virtual disks since it has path styled id column. It helps us to identify failed resources.


erezrokah (approved):
Thanks @aruneko! Sorry for the late review I completely missed this one

PR #10013, opened 2023-04-14
fix(Azure): Reduce duplicate results of Azure CIS v1.3.0 Section 9.1

aruneko:
#### Summary
I noticed that the current implementation of Azure CIS v1.3.0 Section 9.1 returns duplicate policy results. To reduce that, I add `GROUP BY` query into the SQL.


erezrokah (approved):
Thanks @aruneko, approving since I believe this solves the duplicates, but I'm wondering how we get those, as I'd expect us not to have them.

Can you share some redacted example data that creates them?

PR #10060, opened 2023-04-14
feat(source/mysql): Upgrade to `github.com/cloudquery/plugin-sdk/v2` 

### LINKED ISSUE #10034: feat: Migrate `plugin/source/mysql` to `github.com/cloudquery/plugin-sdk/v2`

Part of #9846

Upgrade the code.

### Pull request (optional)

candiduslynx:
Closes #10034.

It includes some nontrivial code for transformation (that is used solely for tests)


plugins/source/mysql/client/transformer.go
```diff
@@ -135,3 +140,41 @@ func (*Transformer) TransformTextArray(v *schema.TextArray) any {
...
+		}
+		// special handling for arrays
+		switch table.Columns[i].Type {
+		case schema.TypeStringArray,
+			schema.TypeIntArray,
+			schema.TypeUUIDArray,
+			schema.TypeInetArray,
+			schema.TypeCIDRArray,
+			schema.TypeMacAddrArray:
+			parts := strings.Split(strings.TrimSuffix(strings.TrimPrefix(val.(string), "["), "]"), ",")
```
disq, line 159: I wonder if there's a better way to read/deserialize array types from MySQL so we don't have to do this.

candiduslynx: me too, actually.
Still, I'd like to have @erezrokah to review this, too

yevgenypats (changes requested):
Given the upgrade is non-trivial I suggest waiting until we rolled out destinations so we can review destination first, learn from that and then move to sources.

plugins/source/mysql/client/transformer.go
candiduslynx: I think the issue is still caused only by tests, so the code is safe to go

candiduslynx:
@yevgenypats the non-trivial part is caused only by testing, so the mainline code update was, in fact, trivial.
I think that the testing updates are safe ATM, as the proper update for the types will be done later (proper source plugins Arrow migration).

plugins/source/mysql/client/transformer.go
```diff
@@ -135,3 +140,41 @@ func (*Transformer) TransformTextArray(v *schema.TextArray) any {
 
 	return v.String()
 }
+
+func (*Transformer) RecordToCQTypes(table *schema.Table, record arrow.Record) (schema.CQTypes, error) {
```
yevgenypats, line 144: Let's just bring back the old `testData` that generates CQTypes directly. this way this migration will be much easier (same as in oracleDB)

yevgenypats (changes requested):

plugins/source/mysql/client/transformer.go
candiduslynx: https://github.com/cloudquery/plugin-sdk/pull/811

candiduslynx: Done in a8c4fb63be7ed5a1c855e3125466a8cda2be0917

yevgenypats (approved):

erezrokah (approved):

Issue #9929, opened 2023-04-13
feat: Refactor `wait-for-required-workflows`


candiduslynx:
### Which problem is this feature request solving?

I think the `wait-for-required-workflows` script might require an overhaul to accommodate for #9655.

I wonder if we could move the script for wait from the workflow to the separate file (easier diff, dev): https://github.com/marketplace/actions/github-script#run-a-separate-file

Additionally, I think we could benefit from splitting the logic of the wait script into distinct parts (this can be doe either together with moving the code to separate file(s) or by splitting script step into several):
* define the list of actions to wait for
* actual waiting

> **Warning**
The previous CI issue (#9388) took a couple of weeks to get resolved.

> **Note**
Currently only @erezrokah has the expertise required to review the changes.
Additionally, the changes have to be thoroughly tested.
This makes this issue a slow one to implement as well as review.

### Describe the solution you'd like

1. Move the logic to a separate JavaScript file
2. Split the functionality from single script into several scoped scripts
3. Introduce `skip-is-ok` logic for reused steps as this allows not to use runner at all and thus, save some CPU cycles/money, too

### Pull request (optional)

- [X] I can submit a pull request

erezrokah:
Thanks for opening the issue @candiduslynx. The changes described in this issue makes sense to me, and they are low risk and should be fast to implement, specifically:

>Describe the solution you'd like
> 1. Move the logic to a separate JavaScript file
> 2. Split the functionality from single script into several scoped scripts
> 3. Introduce skip-is-ok logic for reused steps as this allows not to use runner at all and thus, save some CPU cycles/money, too

As for https://github.com/cloudquery/cloudquery/issues/9655, the bulk of that change (and the risky part) is described in https://github.com/cloudquery/cloudquery/issues/9655#issuecomment-1503122325. While the refactoring can help with making that change, I'm not sure it will reduce the risk or complexity of it. Regardless we don't need to couple those and can still do the refactoring, WDYT?

candiduslynx:
@erezrokah your summary is on point.
I implemented mging to a separate file & split in https://github.com/cloudquery/cloudquery/pull/10072, and I think it's ready to be merged (at the very least, reviewed).
As for the skip part - I'd like to add it in a separate PR to make it a smaller change.

yevgenypats:
Closing given low priority. if ever becomes a high priority we will re-open.

PR #10072, opened 2023-04-14
feat(ci): Move `wait-for-required-workflows` logic to a separate file

candiduslynx:
Part of #9929

Inspired by https://github.com/marketplace/actions/github-script#run-a-separate-file-with-an-async-function

The code is moved to a couple of separate js files & steps:
1. Define the required actions
2. Perform actual waiting

Tested #10120


erezrokah (changes requested):
Hi @candiduslynx, thanks for the PR.

Can we scope this change only to the first task of https://github.com/cloudquery/cloudquery/issues/9929?

That is:
> 1. Move the logic to a separate JavaScript file

That would make it easier to review, and also I think there might be a better way to split the code instead of using 2 action scripts I'd like to discuss, which seems more appropriate in a separate PR, so we don't have to block this one.

candiduslynx:
@erezrokah I reverted the split in b245f74

erezrokah (commented):
Thanks @candiduslynx, a few comments:

1. Can we put the new file under a `scripts/workflows/` directory? I think we put all common scripts there
2. Can you confirm the CWD directory when executing the script remains the same? I ask because we do  `fs.readFileSync(file, 'utf8');` on a relative file so I would to ensure that logic doesn't break (we catch any errors when reading files so they won't so up in the logs)

candiduslynx:
@erezrokah 
1. moved file
2. see the run in the test PR: [this line](https://github.com/cloudquery/cloudquery/actions/runs/4714552387/jobs/8360978870?pr=10120#step:4:15) prints CWD to be the root of the repo, as it should be

erezrokah (approved):

PR #10125, opened 2023-04-16
feat(pg)!: Update postgres to use arrow type system

yevgenypats:
This upgrades postgres destination to use Apache arrow type system with our v2 plugin-sdk

BEGIN_COMMIT_OVERRIDE
feat: Update to use arrow type system

END_COMMIT_OVERRIDE


plugins/destination/postgresql/client/migrate.go
```diff
@@ -326,50 +346,51 @@ func (c *Client) addColumn(ctx context.Context, tableName string, column schema.
...
+		if c.enabledPks() && schema.IsPk(col) {
 			primaryKeys = append(primaryKeys, col.Name)
 		}
 	}
 
 	if len(primaryKeys) > 0 {
 		// add composite PK constraint on primary key columns
 		sb.WriteString(", CONSTRAINT ")
-		sb.WriteString(table.Name)
+		sb.WriteString(tName)
```
erezrokah, line 385:
```suggestion
		sb.WriteString(tableName)
```

I think we want the sanitized version here

plugins/destination/postgresql/client/migrate.go
```diff
@@ -326,50 +346,51 @@ func (c *Client) addColumn(ctx context.Context, tableName string, column schema.
...
 		sb.WriteString("_cqpk PRIMARY KEY (")
 		sb.WriteString(strings.Join(primaryKeys, ","))
 		sb.WriteString(")")
 	}
 	sb.WriteString(")")
 	_, err := c.conn.Exec(ctx, sb.String())
 	if err != nil {
-		return fmt.Errorf("failed to create table %s: %w", table.Name, err)
+		return fmt.Errorf("failed to create table %s: %w", tableName, err)
```
erezrokah, line 393:
```suggestion
		return fmt.Errorf("failed to create table %s: %w", tName, err)
```

Maybe here we want the non sanitized name?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -326,50 +346,51 @@ func (c *Client) addColumn(ctx context.Context, tableName string, column schema.
...
 	sb.WriteString(tableName)
 	sb.WriteString(" (")
-	totalColumns := len(table.Columns)
+	totalColumns := len(table.Fields())
 
 	primaryKeys := []string{}
-	for i, col := range table.Columns {
+	for i, col := range table.Fields() {
 		pgType := c.SchemaTypeToPg(col.Type)
 		if pgType == "" {
```
erezrokah, line 361: Unrelated to this PR, but can `SchemaTypeToPg` return an empty string? Looks like we default to `text`

plugins/destination/postgresql/client/types.go
```diff
@@ -106,76 +106,73 @@ func (c *Client) PgToSchemaType(t string) schema.ValueType {
 	}
 }
 
-func (*Client) Pg10ToSchemaType(t string) schema.ValueType {
+func (c *Client) Pg10ToSchemaType(t string) arrow.DataType {
 	if strings.HasPrefix(t, "timestamp") {
-		return schema.TypeTimestamp
+		return arrow.FixedWidthTypes.Timestamp_us
```
erezrokah, line 111: Maybe add a comment that PostgresSQL permits up to microseconds (6 digits) precision, as for the reason for choosing micros seconds here

plugins/destination/postgresql/client/migrate.go
```diff
@@ -111,89 +115,103 @@ func (c *Client) listPgTables(ctx context.Context, pluginTables schema.Tables) (
...
-		case schema.TypeMacAddrArray:
-			table.Columns[i].Type = schema.TypeStringArray
-		}
-		if table.Columns[i].CreationOptions.PrimaryKey {
-			table.Columns[i].CreationOptions.NotNull = true
-		}
-	}
-	return table
-}
+// func (c *Client) normalizeTableCockroach(table *arrow.Schema) *arrow.Schema {
```
erezrokah, line 150: Can we delete this comment?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -111,89 +115,103 @@ func (c *Client) listPgTables(ctx context.Context, pluginTables schema.Tables) (
...
+	return arrow.NewSchema(fields, &md)
 }
 
-func (c *Client) normalizeTable(table *schema.Table) *schema.Table {
+func (c *Client) normalizeTable(table *arrow.Schema, pgTable *arrow.Schema) *arrow.Schema {
 	switch c.pgType {
-	case pgTypeCockroachDB:
-		return c.normalizeTableCockroach(table)
+	// case pgTypeCockroachDB:
+	// 	return c.normalizeTableCockroach(table)
```
erezrokah, line 201: Same here, we could delete this and even simplify to a single `if`

plugins/destination/postgresql/client/types.go
```diff
@@ -106,76 +106,73 @@ func (c *Client) PgToSchemaType(t string) schema.ValueType {
...
+	case "double precision":
+		return arrow.PrimitiveTypes.Float64
 	case "uuid":
-		return schema.TypeUUID
+		return types.ExtensionTypes.UUID
 	case "bytea":
-		return schema.TypeByteArray
-	case "text[]":
-		return schema.TypeStringArray
+		return arrow.BinaryTypes.Binary
```
erezrokah, line 133: Do we want to use `LargeBinary` here? I think we use it in SQLite.
Not sure if it makes a difference

plugins/destination/postgresql/client/types.go
```diff
@@ -106,76 +106,73 @@ func (c *Client) PgToSchemaType(t string) schema.ValueType {
...
 	case "inet":
-		return schema.TypeInet
-	case "inet[]":
-		return schema.TypeInetArray
-	case "bigint[]", "integer[]", "smallint[]", "bigserial[]", "smallserial[]", "serial[]":
-		return schema.TypeIntArray
+		return types.ExtensionTypes.Inet
 	default:
-		return schema.TypeString
+		return arrow.BinaryTypes.String
```
erezrokah, line 143: Similar here, do we want to use `LargeString` here?

erezrokah (approved):
Didn't see anything blocking, we just need to fix the CI

yevgenypats:
> Didn't see anything blocking, we just need to fix the CI

Thanks for the review. Im fixing cockroach and the CI and will give it another go and test with one of the big 3 plugins to see that we didn't miss anything.

plugins/destination/postgresql/client/client_test.go
```diff
@@ -4,14 +4,14 @@ import (
...
-	"github.com/cloudquery/plugin-sdk/specs"
+	"github.com/cloudquery/plugin-sdk/v2/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/v2/specs"
 )
 
 func getTestConnection() string {
 	testConn := os.Getenv("CQ_DEST_PG_TEST_CONN")
 	if testConn == "" {
-		return "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"
+		return "postgresql://root@localhost:26257/postgres?sslmode=disable"
```
disq, line 14:
```suggestion
		return "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"
```
I think this wasn't intended

plugins/destination/postgresql/client/migrate.go
yevgenypats: Fixed.

plugins/destination/postgresql/client/types_pg.go
```diff
@@ -0,0 +1,108 @@
...
+		return "boolean"
+	case *arrow.Int8Type, *arrow.Uint8Type:
+		return "smallint"
+	case *arrow.Int16Type, *arrow.Uint16Type:
+		return "smallint"
+	case *arrow.Int32Type, *arrow.Uint32Type:
+		return "integer"
+	case *arrow.Int64Type, *arrow.Uint64Type:
+		return "bigint"
+	case *arrow.Float32Type:
```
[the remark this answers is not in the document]
yevgenypats: There is no such thing as `Float16Type` in arrow

plugins/destination/postgresql/client/types_pg.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/cloudquery/plugin-sdk/v2/types"
+)
+
+func (c *Client) SchemaTypeToPg(t arrow.DataType) string {
+	switch c.pgType {
+	case pgTypeCockroachDB:
+		return c.SchemaTypeToCockroach(t)
+	default:
+		return c.SchemaTypeToPg10(t)
```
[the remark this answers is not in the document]
yevgenypats: I think we can keep this for now because it doesn't cost us anything. The moment it will diverge we can drop support for this.

plugins/destination/postgresql/client/types_pg.go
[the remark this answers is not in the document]
erezrokah: https://github.com/cloudquery/arrow/blob/9888ac36c1424840a5d95d1bf6a88cbb7671f0a3/go/arrow/datatype_fixedwidth.go#L507

plugins/destination/postgresql/client/types_pg.go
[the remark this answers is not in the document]
bbernays: Wouldn't dropping support for PG10 require another major version bump?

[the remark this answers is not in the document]
bbernays: Not saying we need to change the code, but I think we could change the documentation to reflect the fact that PG is EOL

plugins/destination/postgresql/client/types.go
yevgenypats: Added - https://www.postgresql.org/docs/current/datatype-datetime.html

PR #10147, opened 2023-04-17
fix: Return error instead of panicking

erezrokah:
#### Summary

The function returns an error so we could do that instead of panicking (unless there's a specific reason to panic I'm missing)


disq:
Could even wrap with `arrow.ErrNotImplemented` (it's a fetish of mine now)

yevgenypats (approved):

yevgenypats:
> Could even wrap with `arrow.ErrNotImplemented` (it's a fetish of mine now)

I think it's an error that shouldn't occur rather then not-implemented. That's why I used panic but ret error is also good. I think also `arrow.ErrNotImplemented` should only be used in arrow codebase (unless you are trolling me :) )

PR #10153, opened 2023-04-18
fix(destination/test): Arrow records memory leak

candiduslynx:
Follows #10144


hermanschaaf (approved):
Looks good, but I think we don't need the breaking change description on this PR, since it has already been migrated to Arrow

yevgenypats (approved):

PR #10154, opened 2023-04-18
fix(sqlite): Always release records

candiduslynx:
Follow-up for #10144

### DIFF
--- a/plugins/destination/sqlite/client/write.go
+++ b/plugins/destination/sqlite/client/write.go
@@ -10,7 +10,15 @@ import (
 	"github.com/cloudquery/plugin-sdk/v2/specs"
 )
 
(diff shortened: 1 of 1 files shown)


candiduslynx:
> That's not quite the correct solution. as this will explode in memory. it should be called inside the for loop.

@yevgenypats 
The release is still called inside the `for` loop. This is just a fallback for the premature exits.

hermanschaaf (approved):

yevgenypats (changes requested):
I don't think line 20 is needed. the channel is not a buffered channel and it's just adds un-necessary logic unless Im missing something.

yevgenypats (approved):

PR #10161, opened 2023-04-18
feat(snowflake)!: Update to SDK V2 (Arrow Support)

yevgenypats:


plugins/destination/snowflake/CHANGELOG.md
erezrokah, line 1: Seems like the changes in this file should be reverted?

plugins/destination/snowflake/client/client.go
```diff
@@ -38,11 +37,18 @@ func New(ctx context.Context, logger zerolog.Logger, destSpec specs.Destination)
 	if err := spec.Validate(); err != nil {
 		return nil, err
 	}
-	_, err := gosnowflake.ParseDSN(spec.ConnectionString)
+	cfg, err := gosnowflake.ParseDSN(spec.ConnectionString)
 	if err != nil {
 		return nil, err
 	}
-	db, err := sql.Open("snowflake", spec.ConnectionString)
+	binaryFormat := "BASE64"
```
erezrokah, line 44: This seems like an additional breaking change? Can we add it to the change log as well? See example https://github.com/cloudquery/cloudquery/pull/10169 for multiple breaking changes in a single PR

erezrokah (commented):
The code looks good, had 2 comments

plugins/destination/snowflake/client/client.go
disq: > This seems like an additional breaking change? Can we add it to the change log as well? See example #10169 for multiple breaking changes in a single PR

@erezrokah I think this is about how Arrow encodes binary columns.

yevgenypats: It's not really a breaking change because it's an encoding for the session only i.e for the arrow/json format when we insert.

erezrokah: Cool, I thought this impacts the format of the data saved in the DB https://docs.snowflake.com/en/user-guide/binary-input-output#base64

erezrokah (approved):

PR #10166, opened 2023-04-18
feat(oracledb): Upgrade to `github.com/cloudquery/plugin-sdk/v2`

### LINKED ISSUE #10165: feat: Migrate `plugin/source/oracledb` to `github.com/cloudquery/plugin-sdk/v2`

Part of https://github.com/cloudquery/cloudquery/issues/9846

Upgrade the code.

### Pull request (optional)

candiduslynx:
Closes #10165


plugins/source/oracledb/client/transformer.go
```diff
@@ -135,3 +140,41 @@ func (*Transformer) TransformTextArray(v *schema.TextArray) any {
 
 	return v.String()
 }
+
+func (*Transformer) RecordToCQTypes(table *schema.Table, record arrow.Record) (schema.CQTypes, error) {
```
yevgenypats, line 144: Can we just bring back the old `GenData` for this? this will make the migration simpler for now. once we migrate sources to arrow this would also be gone.

yevgenypats (changes requested):

plugins/source/oracledb/resources/plugin/plugin_test.go
```diff
@@ -47,6 +47,7 @@ func createTable(ctx context.Context, db *sql.DB, table *schema.Table) error {
 	builder.WriteString("CREATE TABLE ")
 	builder.WriteString(client.Identifier(table.Name))
 	builder.WriteString(" (\n  ")
+	pk := make([]string, 0, len(table.PrimaryKeys()))
```
erezrokah, line 50: Do we still need this change if we use the old test data code?

candiduslynx: Yes (as the gendata will gen the row data, but we still have the new table schema with multiple PK columns)

erezrokah: Update after syncing on Slack. This is needed since we still use a new version of the schema that has multiple PK supoprt

PR #10167, opened 2023-04-18
fix!: Upgrade S3 plugin-sdk and filetypes to v2

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/10112

BEGIN_COMMIT_OVERRIDE

BREAKING-CHANGE: This release introduces an internal change to our type system to use [Apache Arrow](https://arrow.apache.org/). This should not have any visible breaking changes for CSV or JSON output formats, however the Parquet output changes for UUID columns, which now have dashes, and timestamps, which now uses the default Arrow time format (e.g. `2023-01-02 12:23:45`). If you encounter an issue during the upgrade, please submit a [bug report](https://github.com/cloudquery/cloudquery/issues/new/choose).

END_COMMIT_OVERRIDE


plugins/destination/s3/client/write.go
```diff
@@ -66,10 +72,37 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
...
 
-// sanitizeJSONKeys replaces all invalid characters in JSON keys with underscores.
-// It does the replacement in-place, modifying the original object. This is required
-// for compatibility with Athena.
-func sanitizeJSONKeys(obj any) {
+// sanitizeRecordJSONKeys replaces all invalid characters in JSON keys with underscores. This is required
+// for compatibility with Athena. It returns a new record, and the old one is released.
+func sanitizeRecordJSONKeys(mem memory.Allocator, record arrow.Record) (arrow.Record, error) {
+	newRecordBuilder := array.NewRecordBuilder(mem, record.Schema())
+	defer record.Release()
```
yevgenypats, line 79: We can leave this out I think to make things consistent.

plugins/destination/s3/client/write.go
```diff
@@ -66,10 +72,37 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
...
+		col := record.Column(i)
+		if arrow.TypeEqual(col.DataType(), types.NewJSONType()) {
+			for r := 0; r < int(record.NumRows()); r++ {
+				obj := col.GetOneForMarshal(r)
+				sanitizeJSONKeysForObject(obj)
+				newRecordBuilder.Field(i).(*types.JSONBuilder).Append(obj)
+			}
+			continue
+		}
+		for r := 0; r < int(record.NumRows()); r++ {
```
yevgenypats, line 91: I think there is no need to use the `Marshal/Unmarshall` for the unchanged columns. You can just use `array.NewRecord` and pass the columns. This way it will be "zero-copy". and just use the builder for the JSONarray not the whole record.

hermanschaaf: Ah nice, will try that

hermanschaaf: Cool, that works, thanks for the suggestion!

plugins/destination/s3/client/write.go
```diff
@@ -66,10 +67,31 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
 	return nil
 }
 
-// sanitizeJSONKeys replaces all invalid characters in JSON keys with underscores.
-// It does the replacement in-place, modifying the original object. This is required
+// sanitizeRecordJSONKeys replaces all invalid characters in JSON keys with underscores. This is required
 // for compatibility with Athena.
-func sanitizeJSONKeys(obj any) {
+func sanitizeRecordJSONKeys(mem memory.Allocator, record arrow.Record) arrow.Record {
```
candiduslynx, line 72: rm mem param & use default allocator explicitly?

plugins/destination/s3/client/write.go
```diff
@@ -66,10 +67,31 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
...
+	cols := make([]arrow.Array, record.NumCols())
+	for i := 0; i < int(record.NumCols()); i++ {
+		col := record.Column(i)
```
candiduslynx, line 75: just iterate over `record.Columns()`?

candiduslynx (changes requested):

plugins/destination/s3/client/write.go
hermanschaaf: done

plugins/destination/s3/client/write.go
```diff
@@ -66,10 +66,31 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, data
...
+					continue
+				}
+				obj := col.GetOneForMarshal(r)
+				sanitizeJSONKeysForObject(obj)
+				b.Append(obj)
+			}
+			cols[i] = b.NewArray()
+			continue
+		}
+		cols[i] = col
```
candiduslynx, line 88: personal preference
```suggestion
		if !arrow.TypeEqual(col.DataType(), types.NewJSONType()) {
			cols[i] = col
			continue
		}

		b := types.NewJSONBuilder(array.NewExtensionBuilder(memory.DefaultAllocator, types.NewJSONType()))
		for r := 0; r < int(record.NumRows()); r++ {
			if col.IsNull(r) || !col.IsValid(r) {
				b.AppendNull()
				continue
			}

			obj := col.GetOneForMarshal(r)
			sanitizeJSONKeysForObject(obj)
			b.Append(obj)
		}

		cols[i] = b.NewArray()
	}		
```

candiduslynx (approved):
a small refactor for loop (nit)

plugins/destination/s3/client/write.go
hermanschaaf: Yeah, maybe slightly better but I think it's good enough as it is

plugins/destination/s3/client/write.go
hermanschaaf: Updated

plugins/destination/s3/client/write.go
candiduslynx: 100%, just wanted to use exit-early (continue) on the shorter path

yevgenypats (approved):

PR #10169, opened 2023-04-18
feat(mysql): Update to SDK v2 and Arrow

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/9930.
Still need to test a bit and I probably missed a bunch of stuff. Keeping as draft for now

BEGIN_COMMIT_OVERRIDE
   BREAKING-CHANGE: This release introduces an internal change to our type system to use [Apache Arrow](https://arrow.apache.org/). This should not have any visible breaking changes for CSV or JSON output formats, however the Parquet output changes for UUID columns, which now have dashes, and timestamps, which now uses the default Arrow time format (e.g. `2023-01-02 12:23:45`). If you encounter an issue during the upgrade, please submit a [bug report](https://github.com/cloudquery/cloudquery/issues/new/choose).

feat: Timestamp fields type changed from `datetime` to `datetime(6)` to avoid losing fractional data
  BREAKING-CHANGE: Timestamp fields type changed from `datetime` to `datetime(6)` to avoid losing fractional data

feat: Mac Address and Inet column type changed from `nvarchar(255)` to `text` as we cannot assume `nvarchar(255)` columns represent valid Mac Address or Inet 
  BREAKING-CHANGE: Mac Address and Inet column type changed from `nvarchar(255)` to `text` as we cannot assume `nvarchar(255)` columns represent valid Mac Address or Inet 
  
feat: Float fields type changed from `float` to `double` for increased precision 
  BREAKING-CHANGE: Float fields type changed from `float` to `double` for increased precision 

END_COMMIT_OVERRIDE


plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,66 @@ package client
...
+	case *arrow.Float16Type, *arrow.Float32Type, *arrow.Float64Type:
+		return "float"
+	case *arrow.StringType, *arrow.LargeStringType:
+		return "text"
+	case *arrow.BinaryType, *arrow.LargeBinaryType:
+		return "blob"
+	case *types.UUIDType:
+		return "binary(16)"
+	case *arrow.TimestampType:
+		return "datetime(6)"
```
erezrokah, line 57: We previously used `datetime` which does not save the fraction part (the tests caught this). I updated the type to `datetime(6)` and will add a breaking change log entry for it

yevgenypats (changes requested):
I think the list/nested handling is incorrect here. Take a look here - https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/transformer.go#L122

plugins/destination/mysql/client/read.go
```diff
@@ -2,83 +2,197 @@ package client
...
+				recordBuilder.Field(i).(*array.BinaryBuilder).Append(*val.(*[]byte))
+			}
+		case *arrow.TimestampType:
+			asTime := val.(**time.Time)
+			if *asTime == nil {
+				recordBuilder.Field(i).AppendNull()
+			} else {
+				recordBuilder.Field(i).(*array.TimestampBuilder).Append(arrow.Timestamp((*asTime).UnixMicro()))
+			}
+		case *types.UUIDType:
```
yevgenypats, line 159: This is not needed I believe as the `default` case should take care of that.

plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,66 @@ package client
...
+	case *arrow.Int8Type, *arrow.Uint8Type, *arrow.Int16Type, *arrow.Uint16Type, *arrow.Int32Type, *arrow.Uint32Type, *arrow.Int64Type, *arrow.Uint64Type:
+		return "bigint"
+	case *arrow.Float16Type, *arrow.Float32Type, *arrow.Float64Type:
+		return "float"
+	case *arrow.StringType, *arrow.LargeStringType:
+		return "text"
+	case *arrow.BinaryType, *arrow.LargeBinaryType:
+		return "blob"
+	case *types.UUIDType:
+		return "binary(16)"
```
yevgenypats, line 55: Let's use string I think.

plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,66 @@ package client
...
+	sort.Strings(supportedTypes)
+	return nil, fmt.Errorf("got unknown MySQL type %q for column %q of table %q while trying to convert it to CloudQuery internal schema type. Supported MySQL types are %q", sqlType, columnName, tableName, supportedTypes)
+}
+
+func arrowTypeToMySqlStr(t arrow.DataType) string {
+	switch t.(type) {
+	case *arrow.BooleanType:
+		return "bool"
+	case *arrow.Int8Type, *arrow.Uint8Type, *arrow.Int16Type, *arrow.Uint16Type, *arrow.Int32Type, *arrow.Uint32Type, *arrow.Int64Type, *arrow.Uint64Type:
+		return "bigint"
```
yevgenypats, line 47: why all the same? It has `tinyint,smallint,mediumint,int,bigint` this would be very useful for db->db replication when the type is important.

plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,66 @@ package client
...
+		return "blob"
+	case *types.UUIDType:
+		return "binary(16)"
+	case *arrow.TimestampType:
+		return "datetime(6)"
+	case *types.JSONType:
+		return "json"
+	case *arrow.StructType:
+		return "json"
+	case *types.InetType, *types.MacType:
```
yevgenypats, line 62: I think for this one we can also use the `default` case so no special handling will be needed in reverse transform as well.

plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,66 @@ package client
...
+		return "nvarchar(255)"
+	default:
```
yevgenypats, line 64: I would add another case when the type is a List type then we can also use `json` and then use `marshaljson` in the transformer for those

yevgenypats (changes requested):
Wasn't aware mysql doesn't have nested types. Updated my review accordingly.

plugins/destination/mysql/client/types.go
erezrokah: Good point, mostly since this is what we had before in the old type system. I'll update it

plugins/destination/mysql/client/types.go
erezrokah: Good point JSON for list types is a great solution

plugins/destination/mysql/client/types.go
erezrokah: There are a few options for UUID:
 - `binary(16)` is the most efficient
 - `VARCHAR(36)` or VARCHAR(255) which will show nicely in queries
 - `text` which is less efficient
 
 The thing is that when we query the DB during migration we can't 100% say that `VARCHAR(36)` or `VARCHAR(255)` is UUID unless we try to parse the data.
 
 Assuming `binary(16)` is UUID is safer (but also not 100%). So the choice for `binary(16)` was kind of a tradeoff to avoid using `text`.
 
 We have a similar issue with `bool` which is an alias for `tinyint(1)`, so we assume `tinyint(1)` is boolean

plugins/destination/mysql/client/types.go
erezrokah: Split those and the floats in https://github.com/cloudquery/cloudquery/pull/10169/commits/7eef56ab47a2152ede32b43231537f220f8e4335

plugins/destination/mysql/client/read.go
erezrokah: I think using the default only works if we use the string representation for UUID

plugins/destination/mysql/client/types.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/10169/commits/f9a1c30ba2420962227192325dbded2e919bbe6e

plugins/destination/mysql/client/types.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/10169/commits/f9a1c30ba2420962227192325dbded2e919bbe6e

plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,84 @@ package client
...
-		"binary(16)":    schema.TypeUUID,
-		"blob":          schema.TypeByteArray,
-		"datetime":      schema.TypeTimestamp,
-		"nvarchar(255)": schema.TypeInet,
-		"varchar(255)":  schema.TypeInet,
-		"text":          schema.TypeString,
-		"json":          schema.TypeJSON,
+	sqlTypeToSchemaType := map[string]arrow.DataType{
+		"bool":              arrow.FixedWidthTypes.Boolean,
+		"tinyint(1)":        arrow.FixedWidthTypes.Boolean,
```
yevgenypats, line 20: why this one is needed?

plugins/destination/mysql/client/types.go
```diff
@@ -2,63 +2,84 @@ package client
...
+		"tinyint":           arrow.PrimitiveTypes.Int8,
+		"smallint":          arrow.PrimitiveTypes.Int16,
+		"int":               arrow.PrimitiveTypes.Int32,
+		"bigint":            arrow.PrimitiveTypes.Int64,
+		"bigint(20)":        arrow.PrimitiveTypes.Int64,
```
yevgenypats, line 25: Why this one is needed?

yevgenypats (approved):
Looks good. had few non blocking questions.

plugins/destination/mysql/client/types.go
erezrokah: This is more a defensive measure I think in case someone is doing MySQL -> MySQL sync and defined `bool` as `tinyint(1)` (similar to https://github.com/cloudquery/cloudquery/pull/9143/files). I think we can drop it for now

plugins/destination/mysql/client/types.go
erezrokah: Fixed in https://github.com/cloudquery/cloudquery/pull/10169/commits/ef4fdd18e07a15a2cb98cd3d77b12b9c2fa23fb5

plugins/destination/mysql/client/types.go
erezrokah: OK so the correct reason was that we return `bool` a bit below so when we normalized we need to support `bool` too.
https://github.com/cloudquery/cloudquery/pull/10169/commits/ef4fdd18e07a15a2cb98cd3d77b12b9c2fa23fb5 should fix it.

See normalization code here https://github.com/cloudquery/cloudquery/pull/10169/files#diff-7a94fa4d0f7e4c8c80b1ad959ace684a43b44408b3d1d9cee3ea9ef2d87dd9a2R29

erezrokah:
I updated the description with all the breaking changes. I think the `int` changes should not be visible as we still map `schema.TypeInt` to `arrow.PrimitiveTypes.Int64` so the mapping will be the same.

PR #10177, opened 2023-04-18
feat(aws): Parallelize Initialization of Accounts after discovery

### LINKED ISSUE #10151: feat: Improve Initialization Performance for AWS Plugin

When initializing AWS client with LOTS of accounts (either via Org or hardcoded) it can take upwards of 30 minutes for the sync to begin

Make it faster...

### Pull request (optional)

bbernays:
#### Summary

closes #10151 

In our small `org` this change brought the initialization time down from ~5 seconds to ~1 second


plugins/source/aws/client/client.go
```diff
@@ -184,151 +184,7 @@ func (c *Client) withLanguageCode(code string) *Client {
 	return &newC
 }
 
-func verifyRegions(regions []string) error {
```
bbernays, line 187: moved to `client/account.go`

plugins/source/aws/client/client.go
```diff
@@ -184,151 +184,7 @@ func (c *Client) withLanguageCode(code string) *Client {
...
-			return errInvalidRegion
-		}
-		regionExist := availableRegions[region]
-		if !hasWildcard && !regionExist {
-			return errUnknownRegion(region)
-		}
-	}
-	return nil
-}
-func isAllRegions(regions []string) bool {
```
bbernays, line 212: moved to `client/account.go`

plugins/source/aws/client/client.go
```diff
@@ -184,151 +184,7 @@ func (c *Client) withLanguageCode(code string) *Client {
...
-	}
-	return wildcardAllRegions
-}
-
-func getAccountId(ctx context.Context, awsCfg aws.Config) (*sts.GetCallerIdentityOutput, error) {
-	svc := sts.NewFromConfig(awsCfg)
-	return svc.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
-}
-
-func configureAwsSDK(ctx context.Context, logger zerolog.Logger, awsPluginSpec *Spec, account Account, stsClient AssumeRoleAPIClient) (aws.Config, error) {
```
bbernays, line 231: moved to `client/aws_sdk.go`

plugins/source/aws/client/client.go
```diff
@@ -359,132 +217,33 @@ func Configure(ctx context.Context, logger zerolog.Logger, spec specs.Source, op
...
-		}
-		client.ServicesManager.InitServicesForPartitionAccountAndScope(iamArn.Partition, *output.Account, initServices(cloudfrontScopeRegion, awsCfg))
-	}
-	if len(client.ServicesManager.services) == 0 {
-		return nil, fmt.Errorf("no enabled accounts instantiated")
-	}
-	return &client, nil
-}
-
-func findEnabledRegions(ctx context.Context, logger zerolog.Logger, accountName string, ec2Client services.Ec2Client, localRegions []string, accountDefaultRegion string) []string {
```
bbernays, line 428: moved to `client/account.go`

plugins/source/aws/client/client.go
```diff
@@ -359,132 +217,33 @@ func Configure(ctx context.Context, logger zerolog.Logger, spec specs.Source, op
...
-	}
-	return []string{}
-}
-
-func getEnabledRegions(ctx context.Context, ec2Client services.Ec2Client, region string) ([]types.Region, error) {
-	res, err := ec2Client.DescribeRegions(ctx,
-		&ec2.DescribeRegionsInput{AllRegions: aws.Bool(false)},
-		func(o *ec2.Options) {
-			o.Region = region
+			defer initLock.Unlock()
```
disq, line 234: Would move `defer` right after the `Lock()`, so that if something panics the lock is still intact (could prevent hanging)

plugins/source/aws/client/spec.go
```diff
@@ -83,3 +84,9 @@ func validateOUs(ous []string) error {
 	}
 	return nil
 }
+
+func (s *Spec) SetDefaults() {
+	if s.InitializationConcurrency <= 0 {
+		s.InitializationConcurrency = 1
```
disq, line 90:
```suggestion
		s.InitializationConcurrency = 4
```
Maybe default to an 'harmless' value such as 4?

plugins/source/aws/client/client.go
bbernays: So this is a really weird diff that github created:
This is the actual code
```
		errorGroup.Go(func() error {
			svcsDetail, err := client.setupAWSAccount(gtx, logger, &awsPluginSpec, adminAccountSts, account)
			if err != nil {
				return err
			}
			initLock.Lock()
			for _, details := range svcsDetail {
				client.ServicesManager.InitServices(details)
			}
			defer initLock.Unlock()
			return nil
		})
```

plugins/source/aws/client/spec.go
bbernays: Isn't the most harmless value `1`?

plugins/source/aws/client/client.go
bbernays: but updated it to:
```
			initLock.Lock()
			defer initLock.Unlock()
			for _, details := range svcsDetail {
				client.ServicesManager.InitServices(details)
			}
```

disq: Noticed the weird diff but still managed to parse it :) thanks for the update

plugins/source/aws/client/spec.go
disq: There is no gain when using it with `1` which is the same as using the old code.

disq (approved):

website/pages/docs/plugins/sources/aws/configuration.md
```diff
@@ -101,6 +101,11 @@ This is the (nested) spec used by the AWS source plugin.
 
   In AWS organization mode, CloudQuery will source all accounts underneath automatically
 
+- `initialization_concurrency` (int) (default: 4)
```
yevgenypats, line 104: why such a low default if we have 50000 as the default for the syncing goroutines?

bbernays: 50,000 is a number that is designed to be spread out over every single table being resolved... This is only for setting up the SDK (assuming a role and getting the enabled/disabled regions)... If a user needs a higher value they can always increase it

plugins/source/aws/client/client.go
```diff
@@ -177,6 +184,7 @@ func (c *Client) withLanguageCode(code string) *Client {
 	return &newC
 }
 
+// This is the entrypoint into configuring the AWS plugin. It is called by the plugin initialization in resources/plugin/aws.go
```
hermanschaaf, line 187:
```suggestion
// Configure is the entrypoint into configuring the AWS plugin. It is called by the plugin initialization in resources/plugin/aws.go
```

website/pages/docs/plugins/sources/aws/configuration.md
erezrokah: Non blocking for this PR, but we could test by modifying the code to run on more accounts than we have in our playground.
In Azure I tested by creating 1800 duplicated subscriptions, see https://github.com/cloudquery/cloudquery/compare/main...erezrokah:test/azure_memory?expand=1#diff-c5caf1d71566c18f3cf9dfd107693cc1fca4ea980176d33db01d94b013685afeR226

The only catch is the need to keep client IDs unique

bbernays: So tested it with 7000 faked accounts... Only ran into rate limit issues when an account was duplicated... which should ever happen

erezrokah: Nice! Does this mean we can use a higher default concurrency? If we're not sure we can keep it lower and. change the default later based on users reports

PR #10182, opened 2023-04-18
Fixed missing character

still42:
#### Summary

Added a missing character so that the SQL statement returns the expected values.


still42:
Summary
Added a missing character so that the SQL statement returns the expected values.

PR #10183, opened 2023-04-18
fix(azure): Added missing character

still42:
Added missing character so the SQL statement returns the expected values.

#### Summary

Added missing character so that the SQL statement returns expected values.


disq (approved):

bbernays:
Looked into this... This wasn't caught in our tests because in our queries we do not explicitly state what columns we are inserting into, we use the implicit functionality, If we were to hard code the columns we are inserting into then this would fail.

PR #10184, opened 2023-04-18
fix(azure-policies): Fix Syntax for policies

bbernays:
#### Summary

Extension of fix opened in #10183 

This wasn't caught in our tests because in our queries we do not explicitly state what columns we are inserting into, we use the implicit functionality, If we were to hard code the columns we are inserting into then this would fail.


plugins/source/azure/policies/queries/web/function_app_should_only_be_accessible_over_https.sql
```diff
@@ -1,10 +1,10 @@
-insert into azure_policy_results
+INSERT INTO azure_policy_results (execution_time, framework, check_id, title, subscription_id, resource_id, status)
 SELECT
   :'execution_time',
   :'framework',
   :'check_id',
   'Function App should only be accessible over HTTPS',
-  subscription_id
+  subscription_id,
```
bbernays, line 7: This is handled in #10183

erezrokah (approved):

PR #10185, opened 2023-04-18
feat(gcp): Update GCP Instance Shielded VM Policy

aisha-als:
#### Summary

Updating GCP Shielded VM instance policy to include Secure Boot
Link: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#enable_secure_boot


erezrokah (approved):
This looks great thanks @aisha-als 🚀 I confirmed we use a new enough version of the GCP SDK to get this information

PR #10188, opened 2023-04-19
feat(gremlin)!: Update to SDK V2 (Arrow Support)

disq:
BEGIN_COMMIT_OVERRIDE

END_COMMIT_OVERRIDE


plugins/destination/gremlin/client/read.go
```diff
@@ -28,27 +30,36 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
+			if err != nil {
+				return err
 			}
-			res <- madeRow
+			//records = append(records, rec)
+			res <- rec
 		}
 	}
 
+	//syncTimeIndex := table.FieldIndices(schema.CqSyncTimeColumn.Name)[0]
```
disq, line 47: This would also work rather than sorting things in the SDK... but why should every dest plugin have to deal with the same issue, and also hardcoding internal cq-column types is... bad

plugins/destination/gremlin/client/transformer.go
```diff
@@ -1,127 +1,169 @@
...
+		for l := 0; l < col.Len(); l++ {
+			results[l][sc.Field(i).Name] = transformed[l]
+		}
+	}
+	return results
 }
 
 func stripNulls(s string) string {
-	return strings.ReplaceAll(s, "\x00", "")
+	return s
```
disq, line 74: This is currently not working due to the null thing not being merged yet

plugins/destination/gremlin/go.mod
```diff
@@ -3,15 +3,22 @@ module github.com/cloudquery/cloudquery/plugins/destination/gremlin
...
 	github.com/aws/aws-sdk-go-v2/config v1.18.21
 	github.com/cenkalti/backoff/v4 v4.2.0
-	github.com/cloudquery/plugin-sdk v1.45.0
+	github.com/cloudquery/plugin-sdk/v2 v2.3.4
 	github.com/rs/zerolog v1.29.0
 )
 
+replace github.com/apache/arrow/go/v12 => github.com/cloudquery/arrow/go/v12 v12.0.0-20230416141001-e8315b3fff1c
+
+replace github.com/cloudquery/plugin-sdk/v2 => github.com/disq/plugin-sdk/v2 v2.0.0-20230419141635-939b7cf8c8d7
```
disq, line 17: Using a pirated sdk version for the migrate row order fix.

yevgenypats (approved):
Nice! I think with https://github.com/cloudquery/plugin-sdk/pull/810 we can also remove the `Release` stuff

plugins/destination/gremlin/go.mod
```diff
@@ -3,15 +3,22 @@ module github.com/cloudquery/cloudquery/plugins/destination/gremlin
...
 	github.com/aws/aws-sdk-go-v2/config v1.18.21
 	github.com/cenkalti/backoff/v4 v4.2.0
-	github.com/cloudquery/plugin-sdk v1.45.0
+	github.com/cloudquery/plugin-sdk/v2 v2.3.7
 	github.com/rs/zerolog v1.29.0
 )
 
+replace github.com/apache/arrow/go/v12 => github.com/cloudquery/arrow/go/v12 v12.0.0-20230416141001-e8315b3fff1c
+
+replace github.com/cloudquery/plugin-sdk/v2 => github.com/cloudquery/plugin-sdk/v2 v2.3.8-0.20230420092138-d9d7d0ae7688
```
disq, line 17: Using unreleased SDK until 2.3.8 is released

candiduslynx (approved):

plugins/destination/gremlin/go.mod
candiduslynx, line 17: waiting till release

candiduslynx (changes requested):

plugins/destination/gremlin/go.mod
candiduslynx: released: https://github.com/cloudquery/plugin-sdk/pull/815

candiduslynx (approved):

Issue #9946, opened 2023-04-13
feat: Change default of `tables` to empty


yevgenypats:
Right now the default is `*` which means every time we release new tables or even the first user experience in big plugins will be super slow because we fetch thousands of API calls while the user usually just want to check out some simple behaviour as a first experience. The default should be fetch nothing and then the user will have to specify which tables they want to fetch.

bbernays:
A few questions about how is this change going to be implemented:
1. Is it going to be change to the default value? 
2. Are we going to make the `tables` field a required field and just not have an input?
3. Are there any changes to our documentation that need to be made so users can better identify the tables that are pertinent to them?
4. Does this change also impact the migrate command as well?
5. Should we document a subset of tables for each plugin in the example configurations to get users started?

erezrokah:
Good point @bbernays, I think most of the work here will be in the docs. Instead of having a long even growing list of tables to skip, we could document a short list of example tables (or even a single one).

hermanschaaf:
It would be nice if we could add some metadata to certain tables to have them appear on that default list, as well as on the integration pages with links to their docs. I've been trying to think of a way to highlight the main tables on the integration pages so this might intersect nicely :)

erezrokah:
> It would be nice if we could add some metadata to certain tables to have them appear on that default list,

💯 I think this is a scoped version of https://github.com/cloudquery/cloudquery/issues/2103 (like having `isDefault: true` on a table). We could probably start with a static list in the Website, and then extract it to an SDK feature if needed

disq:
It could be 'presets' of some kind. There can be multiple presets, for a variety of purposes for each provider (e.g. security teams don't need ec2_instance_types)

erezrokah:
https://github.com/cloudquery/cloudquery/pull/10189 is deployed and https://github.com/cloudquery/cloudquery/releases/tag/cli-v3.0.0 is released so I think we can close this.

If someone has a follow up suggestion for a default list/presets feel free to open a separate issue, as I believe it's out of scope of this one.

PR #10189, opened 2023-04-19
feat(docs): Update docs to reflect tables as a required configuration

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery/issues/9946.
Should be merged after https://github.com/cloudquery/cloudquery/pull/9864 is released as a major version bump

> To find the places to update I searched for `kind: source` (with/without single/double quotes) in our docs and chose the most suitable default for the use case. I might be missing some context so probably got some wrong


hermanschaaf (approved):
Nice 💯

PR #10194, opened 2023-04-19
feat(mongodb)!: Upgrade to SDK V2 (Arrow support)

yevgenypats:
This is ready apart from a small issue with the microseconds comparison.

Closes https://github.com/cloudquery/cloudquery/issues/10109 

BEGIN_COMMIT_OVERRIDE

END_COMMIT_OVERRIDE


plugins/destination/mongodb/client/migrate.go
```diff
@@ -21,56 +22,53 @@ func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
...
+			c.logger.Debug().Str("table", tableName).Err(err).Msg("skipped create index")
 		default:
-			return fmt.Errorf("create index on %s: %w", table.Name, err)
-		}
-	}
-
-	for _, subTable := range table.Relations {
-		if err := c.migrateTable(ctx, subTable); err != nil {
-			return err
+			return fmt.Errorf("create index on %s: %w", tableName, err)
```
disq, line 40: I would keep this as-is, wrapping errors causes migrate error messages to get wrapped also
```suggestion
			return err
```

plugins/destination/mongodb/client/read.go
```diff
@@ -4,111 +4,107 @@ import (
...
-			return fmt.Errorf("failed to read from table %s: %w", table.Name, err)
+			return fmt.Errorf("failed to read from table %s: %w", tableName, err)
+		}
+		rec, err := c.reverseTransformer(table, result)
+		if err != nil {
+			return fmt.Errorf("failed to read from table %s: %w", tableName, err)
 		}
-		values := c.createResultsArray(result, table)
-		res <- values
+		// values := c.createResultsArray(result, table)
```
disq, line 106:
```suggestion
```

plugins/destination/mongodb/client/write.go
```diff
@@ -2,63 +2,144 @@ package client
...
 )
 
-func (c *Client) appendTableBatch(ctx context.Context, table *schema.Table, resources [][]any) error {
-	documents := make([]any, len(resources))
-	for i, resource := range resources {
-		document := make(bson.M, len(table.Columns))
-		for l, col := range table.Columns {
-			document[col.Name] = resource[l]
+func transformArr(arr arrow.Array) []any {
+	pgArr := make([]any, arr.Len())
```
disq, line 17:
```suggestion
	dbArr := make([]any, arr.Len())
```

plugins/destination/mongodb/client/write.go
```diff
@@ -2,63 +2,144 @@ package client
...
+			var val any
+			if err := json.Unmarshal([]byte(a.ValueStr(i)), &val); err != nil {
+				panic(err)
+			}
+			dbArr[i] = val
+		case array.ListLike:
+			start, end := a.ValueOffsets(i)
+			nested := array.NewSlice(a.ListValues(), start, end)
+			dbArr[i] = transformArr(nested)
+			nested.Release()
```
hermanschaaf, line 56: Nit: can probably removed now if we're relying on the Go garbage collector

plugins/destination/mongodb/client/write.go
```diff
@@ -2,63 +2,144 @@ package client
...
+		default:
+			dbArr[i] = arr.ValueStr(i)
```
hermanschaaf, line 58: Will this handle structs in the way we want? Maybe we should use json marshaling here, rather than value string?

plugins/destination/mongodb/client/write.go
```diff
@@ -2,63 +2,144 @@ package client
...
 	return nil
 }
 
-func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resources [][]any) error {
+func (c *Client) WriteTableBatch(ctx context.Context, table *arrow.Schema, resources []arrow.Record) error {
+	defer func() {
+		for _, r := range resources {
+			r.Release()
+		}
+	}()
```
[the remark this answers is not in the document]
candiduslynx: +1

plugins/destination/mongodb/client/write.go
candiduslynx: +1

plugins/destination/mongodb/client/write.go
yevgenypats: structs are handled under `JSON` . Otherwise we will have the same problem for example with UUID having`"` if we wont use `ValueStr` or any other unknown type might have un-necessary `"`

hermanschaaf: Ah, what I mean is what happens if we receive an arrow struct type: even though source plugins don't do that today, that case is not handled explicitly here so we would have to come back and update it later it seems.

disq (approved):

PR #10202, opened 2023-04-19
fix: Add timestamp type

erezrokah:
#### Summary

Related to https://github.com/koltyakov/cq-source-sharepoint/issues/19#issuecomment-1514954567.

SQLite doesn't have an official `timestamp` type but the Go lib takes care of it, see https://github.com/mattn/go-sqlite3/issues/748.

We had this type in `v1`, see https://github.com/cloudquery/cloudquery/blob/plugins-destination-sqlite-v1.3.6/plugins/destination/sqlite/client/types.go#L58

### DIFF
--- a/plugins/destination/sqlite/client/types.go
+++ b/plugins/destination/sqlite/client/types.go
@@ -16,6 +16,8 @@ func (*Client) arrowTypeToSqliteStr(t arrow.DataType) string {
 		return "real"
 	case arrow.BOOL:
 		return "boolean"
+	case arrow.TIMESTAMP:
+		return "timestamp"
 	default:
 		return "text"
 	}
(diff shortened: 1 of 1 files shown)


plugins/destination/sqlite/client/types.go
```diff
@@ -50,7 +54,9 @@ func (*Client) sqliteTypeToArrowType(t string) arrow.DataType {
 		return arrow.BinaryTypes.LargeBinary
 	case "boolean":
 		return arrow.FixedWidthTypes.Boolean
+	case "timestamp":
+		return arrow.FixedWidthTypes.Timestamp_us
 	default:
-		panic("unknown type")
+		panic("unknown type: " + t)
```
erezrokah, line 60: Also added this bit so we'd know which type errored out

disq (approved):

PR #10209, opened 2023-04-20
fix: Update to SDK v2.3.7, remove release calls

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/810


plugins/destination/mysql/client/read.go
```diff
@@ -74,7 +74,6 @@ func (*Client) createResultsArray(table *arrow.Schema) []any {
 
 func reverseTransform(table *arrow.Schema, values []any) (arrow.Record, error) {
 	recordBuilder := array.NewRecordBuilder(memory.DefaultAllocator, table)
```
erezrokah, line 76: The default allocator is the Go one, so I don't think we need a release here

candiduslynx (approved):

plugins/destination/mysql/client/read.go
```diff
@@ -74,7 +74,6 @@ func (*Client) createResultsArray(table *arrow.Schema) []any {
...
-	defer recordBuilder.Release()
```
yevgenypats, line 77: I think here it's fine just because the builder is only used in this function scope. we can later try and test if it has any effect on memory vs not using it.

yevgenypats (changes requested):

plugins/destination/mysql/client/read.go
erezrokah: The default allocator is the Go one so I thought this is a no op. I can keep it if needed but removed to avoid confusion for someone who's later reading the code

hermanschaaf: Unfortunately `Release` is not a no-op in the Go allocator. But I think it's fine to remove it here and let the garbage collector clean it up for the sake of consistency.

erezrokah: Reverted the commit until we reach a conclusion

yevgenypats (approved):

PR #10220, opened 2023-04-20
fix!: Remove deprecated CSV plugin

hermanschaaf:
The CSV plugin was deprecated a couple of months ago, please use the `file` plugin instead.

Closes https://github.com/cloudquery/cloudquery/issues/10102


.release-please-manifest.json
```diff
@@ -27,8 +27,6 @@
   "plugins/destination/postgresql+FILLER": "0.0.0",
   "plugins/destination/test": "1.3.26",
   "plugins/destination/test+FILLER": "0.0.0",
-  "plugins/destination/csv": "1.2.2",
```
candiduslynx, line 30: should be left as is (used in the website to show the latest version)

candiduslynx (changes requested):

.release-please-manifest.json
hermanschaaf: It's being removed from the website

candiduslynx (approved):

.release-please-manifest.json
erezrokah: Also it's not used in the Website I think. We use https://github.com/cloudquery/cloudquery/blob/845a8a0f344d253534eceeee4b9320658f9652d4/website/versions/destination-csv.json

erezrokah (approved):
🧹

PR #10222, opened 2023-04-20
feat: Update to SDK v2.3.8

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery/issues/9946

BEGIN_COMMIT_OVERRIDE

feat: Update to SDK v2.3.8

BREAKING-CHANGE: With this CLI version the `tables:` source spec configuration is now required. We're making it required as the previous default of `["*"]` (all tables) is not suitable for production usage as it can be slow to sync all tables on popular plugins like AWS, GCP and Azure.

END_COMMIT_OVERRIDE


yevgenypats (approved):

erezrokah:
Another commit is coming to fix the tests

PR #10244, opened 2023-04-20
doc(aws-resources): Update AWS Policies Readme

sandorcloudopus:
In order to successfully run the AWS CIS V1.2.0 scripts, you require to extend the list of tables.


disq:
The list is auto-generated so we're taking a look at the code which generates this file and we'll come up with a fix shortly.

disq:
Seems like the queries used in views aren't documented. This affects all policies that depend on the `view_aws_log_metric_filter_and_alarm` view. All tables used in that view must be documented in the relevant policies. I've opened an issue to track this here: https://github.com/cloudquery/cloudquery/issues/10249

Feel free to reopen this PR or comment on the new issue if you think I've missed anything.

PR #10250, opened 2023-04-20
fix(aws-docs): Policy docs should include tables used in views

### LINKED ISSUE #10249: bug(policies): Dependent tables aren't correctly documented if used in views

### Current Behavior

If a policy is creating and using a view, the tables used in the view aren't documented in the policies section on the website.

Current docs: https://www.cloudquery.io/docs/plugins/sources/aws/policies

Some of the AWS policies seem to create and use a `view_aws_log_metric_filter_and_alarm` view, which depend on a quite a few tables, but some of them aren't really used on any other query so aren't included in the final list in the docs.`aws_cloudwatch_alarms` and `aws_sns_subscriptions` are missing (there might be more)

### Expected Behavior

Docs should contain all required tables for the specified policy to work.

```
```

### Steps To Reproduce

```
```

```
```

### CloudQuery version

latest main

### Additional Context

_No response_

### Pull request (optional)

disq:
Fixes #10249

Looks like https://cloudquery-ooopynfqx-cloudquery.vercel.app/docs/plugins/sources/aws/policies


plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -48,6 +52,7 @@ type PolicyInfo struct {
 	Name    string
 	Queries []Query
 	Tables  []string
+	Views   []string
```
disq, line 55: This is not used, included here for completeness.

plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -113,26 +128,39 @@ func extractQueries(prefix, sqlPath string) ([]Query, error) {
...
+	q := Query{}
+	if tableMatches := reTable.FindAllStringSubmatch(content, -1); len(tableMatches) > 0 {
 		for _, m := range tableMatches {
-			if !strings.HasPrefix(m[1], prefix) {
-				continue
+			switch {
+			case strings.HasPrefix(m[1], prefix):
+				q.Tables = append(q.Tables, m[1])
+				q.Path = sqlPath
+			case strings.HasPrefix(m[1], "view_"+prefix):
```
disq, line 138: This only works if names of views are clearly marked with `view_{prefix}`. We'd have to do a multipass if we don't want this requirement (which currently GCP doesn't fit)

hermanschaaf (approved):
Looks good, thanks for the fix! Just one question: would it now be possible to list the views a policy depends on as well? Listing those in a separate section could be really useful too.

hermanschaaf (approved):

plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -61,28 +67,58 @@ func newPolicyInfo(name string, queries []Query, allTables []Table) *PolicyInfo
...
+			if pi.CreatedViews[table] {
+				// This is a view, not a table
+				v[table] = true
+			}
+
+			t[table] = true
 			ancestors := allTables.FindAncestors(table)
 			for _, a := range ancestors {
-				t[a.Name] = struct{}{}
+				t[a.Name] = true
```
candiduslynx, line 98: better leave as it was (bool uses mem, empty struct doesn't)

plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -61,28 +67,58 @@ func newPolicyInfo(name string, queries []Query, allTables []Table) *PolicyInfo
...
+
+	unused := make(map[string]bool, len(pi.CreatedViews))
+	for k := range pi.CreatedViews {
+		if _, ok := v[k]; !ok {
+			unused[k] = true
+		}
 	}
-	sort.Strings(final)
-	pi.Tables = final
+	pi.UnusedViews = keysOf(unused)
```
candiduslynx, line 111: https://pkg.go.dev/golang.org/x/exp/maps#Keys

plugins/source/aws/scripts/policy_docs/main.go
```diff
@@ -195,6 +251,15 @@ func readTablesJSON(filepath string) ([]Table, error) {
 	return tables, err
 }
 
+func keysOf(m map[string]bool) []string {
```
candiduslynx, line 254: https://pkg.go.dev/golang.org/x/exp/maps#Keys

candiduslynx (changes requested):

plugins/source/aws/scripts/policy_docs/main.go
disq: Doesn't matter and this makes the template simpler (or, usable)

PR #10251, opened 2023-04-21
chore(k8s): Policy docs should include tables used in views

disq:
Just like https://github.com/cloudquery/cloudquery/pull/10250 but without any changes due to the fact that we don't have any views here.

[Looks the same](https://cloudquery-2to47f9br-cloudquery.vercel.app/docs/plugins/sources/k8s/policies)


plugins/source/k8s/scripts/policy_docs/main.go
```diff
@@ -61,28 +67,58 @@ func newPolicyInfo(name string, queries []Query, allTables []Table) *PolicyInfo
...
-				t[a.Name] = struct{}{}
+				t[a.Name] = true
 			}
 		}
 	}
-	var final []string
-	for k := range t {
-		final = append(final, k)
+
+	pi.Tables, pi.DependentViews = keysOf(t), keysOf(v)
```
candiduslynx, line 103: [maps.Keys](https://pkg.go.dev/golang.org/x/exp/maps#Keys)

plugins/source/k8s/scripts/policy_docs/main.go
```diff
@@ -61,28 +67,58 @@ func newPolicyInfo(name string, queries []Query, allTables []Table) *PolicyInfo
...
-			t[table] = struct{}{}
+			if pi.CreatedViews[table] {
+				// This is a view, not a table
+				v[table] = true
+			}
+
+			t[table] = true
 			ancestors := allTables.FindAncestors(table)
 			for _, a := range ancestors {
-				t[a.Name] = struct{}{}
```
candiduslynx, line 70: better to use struct (doesn't use memory)

plugins/source/k8s/scripts/policy_docs/main.go
```diff
@@ -195,6 +251,15 @@ func readTablesJSON(filepath string) ([]Table, error) {
 	return tables, err
 }
 
+func keysOf(m map[string]bool) []string {
```
candiduslynx, line 254: https://pkg.go.dev/golang.org/x/exp/maps#Keys

candiduslynx (dismissed):

candiduslynx (approved):

PR #10255, opened 2023-04-21
fix(aws): Remove Resource Specific Parallelization For S3 to use only the SDK parallelization 

bbernays:
#### Summary

Was trying to look into an issue with S3 ACL Pks and removed this refactor from my investigation code. Removed all custom/resource specific parallelization to rely solely on the SDK provided parallelization

### DIFF
--- a/plugins/source/aws/resources/services/s3/buckets.go
+++ b/plugins/source/aws/resources/services/s3/buckets.go
@@ -89,7 +89,7 @@ func resolveS3BucketsAttributes(ctx context.Context, meta schema.ClientMeta, r *
 		resource.Region = output
 	}
 
-	resolvers := []func(context.Context, schema.ClientMeta, *models.WrappedBucket, string) error{
+	resolvers := []func(context.Context, schema.ClientMeta, *models.WrappedBucket) error{
 		resolveBucketLogging,
 		resolveBucketPolicy,
 		resolveBucketVersioning,
@@ -99,7 +99,7 @@ func resolveS3BucketsAttributes(ctx context.Context, meta schema.ClientMeta, r *
 		resolveBucketOwnershipControls,
 	}
 	for _, resolver := range resolvers {
-		if err := resolver(ctx, meta, resource, resource.Region); err != nil {
+		if err := resolver(ctx, meta, resource); err != nil {
 			r.Item = resource
 			if isBucketNotFoundError(c, err) {
 				return nil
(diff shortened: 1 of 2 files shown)


plugins/source/aws/resources/services/s3/buckets.go
```diff
@@ -133,48 +88,27 @@ func resolveS3BucketsAttributes(ctx context.Context, meta schema.ClientMeta, res
...
+	}
+	for _, resolver := range resolvers {
+		if err := resolver(ctx, meta, resource, resource.Region); err != nil {
+			r.Item = resource
+			if isBucketNotFoundError(c, err) {
+				return nil
+			}
+			return err
+		}
 	}
```
bbernays, line 109: this function moved to `bucket_grants.go`

plugins/source/aws/resources/services/s3/buckets.go
```diff
@@ -52,41 +52,10 @@ func fetchS3Buckets(ctx context.Context, meta schema.ClientMeta, _ *schema.Resou
...
-		}
-	}()
-	done := make(chan struct{})
-	go func() {
-		for err = range errs {
-			cl.Logger().Err(err).Msg("failed to fetch s3 bucket")
-		}
-		close(done)
-	}()
-	wg.Wait()
```
bbernays, line 80: Using `Resolver` and `PreResourceResolver` instead of custom channels and go routines

plugins/source/aws/resources/services/s3/buckets.go
```diff
@@ -17,10 +16,11 @@ import (
 func Buckets() *schema.Table {
 	tableName := "aws_s3_buckets"
 	return &schema.Table{
-		Name:      "aws_s3_buckets",
-		Resolver:  fetchS3Buckets,
-		Transform: transformers.TransformWithStruct(&models.WrappedBucket{}),
-		Multiplex: client.AccountMultiplex(tableName),
+		Name:                "aws_s3_buckets",
+		Resolver:            listS3Buckets,
+		PreResourceResolver: resolveS3BucketsAttributes,
```
hermanschaaf, line 21: This wasn't using a `PreResourceResolver` before, so think this may slightly change the behavior. Should it still be considered a refactor? Maybe `fix` would be more appropriate.

bbernays: Fair.

hermanschaaf (approved):

PR #10264, opened 2023-04-23
test(issues): Fix race condition in tests, pre-initialize all issues

erezrokah:
#### Summary

This PR fixes a race condition in the issues test, see https://github.com/cloudquery/cloudquery/actions/runs/4753216187/jobs/8444531449.

We can't update the response in the request handler as the hander can be called from multiple go routines.
The solution is to pre-initialize all issues (based on the `total` count) and for each request return the relevant range for the requested page


plugins/source/snyk/resources/services/reporting/issues_mock_test.go
```diff
@@ -12,30 +13,48 @@ import (
...
 
-func createIssues(mux *httprouter.Router) error {
-	var resp snyk.ListReportingIssuesResponse
-	if err := faker.FakeObject(&resp); err != nil {
-		return err
-	}
-	resp.Total = 2001
-	i := 0
-	mux.POST("/v1/reporting/issues/", func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-		resp.Results[0].Issue.ID = fmt.Sprintf("test-%d", i)
```
erezrokah, line 23: This is the race condition as the handler can be called by multiple Go routines

hermanschaaf (approved):
Ah, nice find, thanks for fixing!

PR #10275, opened 2023-04-24
feat(duckdb): Update to SDK V2 (Arrow)

yevgenypats:
Closes https://github.com/cloudquery/cloudquery/issues/10103

BEGIN_COMMIT_OVERRIDE

END_COMMIT_OVERRIDE


plugins/destination/duckdb/json/writer.go
```diff
@@ -0,0 +1,71 @@
...
+		for j := 0; j < nc; j++ {
+			switch record.Column(j).DataType().ID() {
+			case arrow.TIMESTAMP:
+				tmp[w.schema.Field(j).Name] = record.Column(j).GetOneForMarshal(i)
+				if tmp[w.schema.Field(j).Name] != nil {
+					strTimestamp := tmp[w.schema.Field(j).Name].(string)
+					// I believe the real fix should be in how duckdb is handling timestamp as it seems like a bug
+					// so this is just a workaround.
+					// in anycase we do want a capability of custom marshaling where needed.
+					if !strings.Contains(strTimestamp, ".") {
```
disq, line 51: You can 'simplify' this by doing
```suggestion
					if ind := strings.Index(strTimestamp, "."); ind == -1 {
```
And then using `ind` directly below

plugins/destination/duckdb/client/types.go
```diff
@@ -3,100 +3,95 @@ package client
...
-		if isArray {
-			panic("unsupported type " + t + "[]")
-		}
-		return schema.TypeByteArray
+	case "tinyint":
+		return arrow.PrimitiveTypes.Uint8
+	case "smallint":
+		return arrow.PrimitiveTypes.Uint16
+	case "int", "int4", "integer":
+		return arrow.PrimitiveTypes.Uint32
```
hermanschaaf, line 65: This doesn't look right; it probably shouldn't be unsigned?

hermanschaaf: (also the ones above it)

plugins/destination/duckdb/client/write.go
```diff
@@ -3,156 +3,193 @@ package client
...
-			expanded := expandData(table, r.Data)
-			if _, err := c.db.Exec(sql, expanded...); err != nil {
-				return fmt.Errorf("failed to execute '%s': %w", sql, err)
-			}
-		} else {
-			err := c.upsert(table, r.Data)
-			if err != nil {
-				return fmt.Errorf("failed to execute upsert: %w", err)
-			}
+// func (c *Client) upsert(table *schema.Table, data []any) error {
```
hermanschaaf, line 16: This commented out section can probably be removed now

plugins/destination/duckdb/json/writer.go
```diff
@@ -0,0 +1,71 @@
...
+						tmp[w.schema.Field(j).Name] = strTimestamp + ".000000"
+					} else {
+						ind := strings.Index(strTimestamp, ".")
+						if len(strTimestamp[ind+1:]) < 6 {
+							tmp[w.schema.Field(j).Name] = strTimestamp + "000000"[len(strTimestamp[ind+1:]):]
```
hermanschaaf, line 56: Wow, I actually didn't know you can do `"000000"[i:i+1]` in Go without assigning a variable first 😄 

An alternative would be to use `strings.Repeat("0", 6 - len(strTimestamp[ind+1:])`

plugins/destination/duckdb/main.go
```diff
@@ -3,8 +3,8 @@ package main
 import (
 	"github.com/cloudquery/cloudquery/plugins/destination/duckdb/client"
 	"github.com/cloudquery/cloudquery/plugins/destination/duckdb/resources/plugin"
-	"github.com/cloudquery/plugin-sdk/plugins/destination"
-	"github.com/cloudquery/plugin-sdk/serve"
+	"github.com/cloudquery/plugin-sdk/v2/plugins/destination"
+	"github.com/cloudquery/plugin-sdk/v2/serve"
 )
 
 const (
```
hermanschaaf, line 10: Since the writer changed, I think we need to add `destination.WithManagedWriter()` to the `NewPlugin` call here 👇 Otherwise it panics if you run a sync

hermanschaaf (changes requested):
We need to add `destination.WithManagedWriter()`

plugins/destination/duckdb/main.go
hermanschaaf: I fixed it locally so I could test it out, and pushed the change up now

erezrokah (commented):
Added a few comments (non blocking) and will run locally too

plugins/destination/duckdb/client/types.go
```diff
@@ -3,100 +3,95 @@ package client
...
+	case "boolean", "bool", "logical":
+		return arrow.FixedWidthTypes.Boolean
+	case "double", "float8", "numeric", "decimal":
+		return arrow.PrimitiveTypes.Float64
+	case "float", "float4", "real":
+		return arrow.PrimitiveTypes.Float32
+	case "blob", "bytea", "binary", "varbinary":
+		return arrow.BinaryTypes.Binary
+	case "date":
+		return arrow.FixedWidthTypes.Date64
```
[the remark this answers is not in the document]
hermanschaaf: This is what I'm basing it on: https://arrow.apache.org/docs/python/generated/pyarrow.date64.html

And e.g. here, when converting to `time.Time`, it throws away all the millisecond precision being stored anyway and just returns the date (year/month/day) https://github.com/apache/arrow/blob/72191880d3a66e6ecf65d9c249dcb2856a697f47/go/arrow/datatype_fixedwidth.go#L99-L102

plugins/destination/duckdb/client/types.go
```diff
@@ -3,100 +3,95 @@ package client
...
-	case "blob":
-		if isArray {
-			panic("unsupported type " + t + "[]")
-		}
-		return schema.TypeByteArray
+	case "tinyint":
+		return arrow.PrimitiveTypes.Uint8
+	case "smallint":
+		return arrow.PrimitiveTypes.Uint16
+	case "int", "int4", "integer":
```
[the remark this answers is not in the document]
yevgenypats: Added the alias. Yeah it's better if handle the aliases as well (no reason not to handle them I think)

plugins/destination/duckdb/client/types.go
```diff
@@ -3,100 +3,95 @@ package client
...
-		return "text"
-	case schema.TypeMacAddrArray:
-		return "text[]"
-	case schema.TypeInet:
-		return "text"
-	case schema.TypeInetArray:
-		return "text[]"
-	case schema.TypeIntArray:
-		return "int[]"
+	case *arrow.Date64Type:
```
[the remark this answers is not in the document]
yevgenypats: I wasn't able to find that it stores it in milliseconds in the code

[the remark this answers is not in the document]
hermanschaaf: What do you mean? That's how the type is defined here https://arrow.apache.org/docs/python/generated/pyarrow.date64.html#pyarrow.date64 "milliseconds since UNIX epoch 1970-01-01"?

plugins/destination/duckdb/client/types.go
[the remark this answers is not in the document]
hermanschaaf: Yeah, I had a hard-to-debug error at some point where two dates like `2021-01-01` and `2021-01-01` were coming back as different, and it turned out to be a difference in milliseconds that isn't printed by the date64 string output.

plugins/destination/duckdb/client/types.go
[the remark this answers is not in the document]
yevgenypats: I think it's a python thing but anyway changed it as prob there is no reason for `date64` anyway as `date32` should be enough

hermanschaaf (approved):

PR #10283, opened 2023-04-24
feat(s3): Detect write access without writing

### LINKED ISSUE #9839: feat(s3): disable or cleanup s3 test data

s3 destination adds test data to the bucket, it would be nice to have an option to disable or clean up this.

i prefer the disable option
- cleaner
- no extra permissions required

### Pull request (optional)

disq:
Fixes #9839

I first tried creating an `io.Pipe` and starting an upload only to cancel it later, but that was hairier than expected.

Could also just remove this check as suggested in the issue.

### DIFF
--- a/plugins/destination/s3/client/client.go
+++ b/plugins/destination/s3/client/client.go
@@ -1,16 +1,20 @@
 package client
 
 import (
-	"bytes"
 	"context"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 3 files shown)


erezrokah:
Hi @disq and thank you for the PR. You'll need to re-open it from a non forked repo branch to get the CI passing

plugins/destination/s3/client/client.go
```diff
@@ -63,17 +67,75 @@ func New(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (de
...
 	}
+
 	return c, nil
 }
 
 func (*Client) Close(ctx context.Context) error {
 	return nil
 }
+
+func (c *Client) testWriteAccess(ctx context.Context, cfg aws.Config) error {
```
bbernays, line 81: So this only works if the bucket and user credentials are in the same account. I think we are better off just adding an option that allows users to skip the tests

PR #10284, opened 2023-04-24
feat(clickhouse)!: Upgrade to `github.com/cloudquery/plugin-sdk/v2`

### LINKED ISSUE #9847: feat: Migrate `plugin/destination/clickhouse` to `github.com/cloudquery/plugin-sdk/v2`

Part of #9846

Upgrade the code performing the major bump.
PR description should include the following part:

```
BEGIN_COMMIT_OVERRIDE

END_COMMIT_OVERRIDE
```

### Pull request (optional)

candiduslynx:
Closes #9847 

BEGIN_COMMIT_OVERRIDE
feat: Update to use [Apache Arrow](https://arrow.apache.org/) type system (#10284)

feat: Allow nullable columns in primary keys (#10284)
BREAKING-CHANGE: This change enables [`allow_nullable_key`](https://clickhouse.com/docs/en/operations/settings/settings#allow-nullable-key) for tables (#10284).

END_COMMIT_OVERRIDE


yevgenypats (changes requested):
Just marking so I've a chance to review before it is merged.

candiduslynx:
@yevgenypats @erezrokah 
As some code had to be copied/based on ClickHouse Go SDK, I think we might need to include this info per license requirements:
https://github.com/ClickHouse/clickhouse-go/blob/main/LICENSE

plugins/destination/clickhouse/typeconv/ch/values/list.go
```diff
@@ -0,0 +1,62 @@
...
+		if elem != nil {
+			val.Elem().Set(reflect.ValueOf(elem))
+		}
+		res.Index(i).Set(val)
+	}
+
+	return res.Interface(), nil
+}
+
+type listWrapper struct {
```
candiduslynx, line 52: I really think this should be implemented in the Arrow Go code (`*array.FixedSizeList` should conform to the `array.ListLike` interface)

candiduslynx: Proposed in https://github.com/apache/arrow/issues/35325, PR: https://github.com/apache/arrow/pull/35327

plugins/destination/clickhouse/typeconv/arrow/types/field.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2/lib/column"
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/cloudquery/cloudquery/plugins/destination/clickhouse/util"
+	"github.com/cloudquery/plugin-sdk/v2/types"
+)
+
+func fieldFromColumn(col column.Interface) (*arrow.Field, error) {
```
yevgenypats, line 13: why not just use the string representation?

plugins/destination/clickhouse/typeconv/arrow/types/field.go
```diff
@@ -0,0 +1,95 @@
...
+	name := util.UnquoteID(col.Name())
+	switch col := col.(type) {
+	case *column.Bool:
+		return &arrow.Field{Name: name, Type: new(arrow.BooleanType)}, nil
```
yevgenypats, line 17: there is no need to create those types. take a look here - https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/types_pg.go#L29

plugins/destination/clickhouse/typeconv/ch/values/time.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/apache/arrow/go/v12/arrow/array"
+)
+
+type date interface {
+	ToTime() time.Time
+}
+
+func dateValue[A date](arr primitive[A]) []*time.Time {
```
yevgenypats, line 14: what's the point of generics here? can't it be just an interface ?

plugins/destination/clickhouse/typeconv/ch/values/decimal.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/apache/arrow/go/v12/arrow"
+	"github.com/shopspring/decimal"
+)
+
+type decimalToBigInt interface {
+	BigInt() *big.Int
+}
+
+func decimalValue[A decimalToBigInt](arr primitive[A]) []*decimal.Decimal {
```
yevgenypats, line 14: Same. what the point of generics here? can it be just an interface?

plugins/destination/clickhouse/typeconv/ch/types/data_type.go
```diff
@@ -0,0 +1,81 @@
+package types
+
+import (
+	"strconv"
+
+	"github.com/apache/arrow/go/v12/arrow"
+)
+
+func dataType(_type arrow.DataType) (string, error) {
+	switch _type.ID() {
```
yevgenypats, line 10: seems there is a mix between usage of `ID` and type casting. can we use just one? similar to https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/types_pg.go#L29 ?

yevgenypats (changes requested):
lot's of code :) few initial comments - will take a break and then give it another fresh look tomorrow.

I did an initial research and seems this should be our best bet in clickhouse indeed in terms of transformation (transforming to json or to arrow will be too slow and we might loose some control that we need)

plugins/destination/clickhouse/typeconv/arrow/types/field.go
candiduslynx: This allows us not to parse a lot ourselves, and instead we just handle the SDK types.
Some (e.g., `Tuple`, precision & time zone for `DateTime64`, etc.) parts are still required to be parsed, but mostly we just can lean on the SDK to unwrap the nested types for us.

plugins/destination/clickhouse/typeconv/arrow/types/field.go
candiduslynx: Could you elaborate?

plugins/destination/clickhouse/typeconv/ch/values/time.go
candiduslynx: We're inferring the interface of the value returned by `Value(int)`. In this case, it's either `arrow.Date32` & `arrow.Date64`.
If we were to use interfaces there would've been 2 separate functions (we might as well just do 2 separate functions for each array type).

plugins/destination/clickhouse/typeconv/ch/values/decimal.go
candiduslynx: Same as Date32/64 but for Decimal128/256

plugins/destination/clickhouse/typeconv/ch/types/data_type.go
candiduslynx: Good point, I'll look into it

plugins/destination/clickhouse/typeconv/ch/values/list.go
candiduslynx: merged, updated the code in 6cda3a3

plugins/destination/clickhouse/typeconv/ch/types/data_type.go
candiduslynx: Done in fb53fc1

plugins/destination/clickhouse/typeconv/arrow/types/field.go
candiduslynx: This is the reverse type transformation (from ClickHouse to Apache Arrow) as it's required to perform comparison in the normalized schemas.

yevgenypats (changes requested):
I think let's try go easy on generics. maybe we should disable generics use across all our repos.

I don't see an immediate need for generics in this PR. For example all generics code under `values` can be replaces with one call to `.Values` available on all arrow arrays (or `.Uint16Values` for some reason it is called by different name sometimes but nevertheless it is available there so should save lots of code and test code in this PR which will make it hard to maintain)

candiduslynx:
> all generics code under `values` can be replaces with one call to `.Values` available on all arrow arrays

@yevgenypats are you referring to the `ch/values` package?
If so, the reason for the generics is that we do need concrete type for the values passed to the ClickHouse SDK, and the `Value` method for different array types returns different types altogether (not `any`).

If we take a look at the `arrow/values` and generics usage there, removing them would mean going for the separate functions per array (`Append` takes concrete types once again).

Could you give an example on what do you suggest doing and where?

yevgenypats (approved):
Going to stamp approve it.  Might go back to that if time allows.

PR #10290, opened 2023-04-25
fix(azure-consumption): Update Price Sheets next link on new response when paginating

erezrokah:
#### Summary

There was a bug in the original pagination code I wrote. See https://github.com/Azure/azure-sdk-for-go/issues/20671#issuecomment-1521307978


hermanschaaf (approved):

hermanschaaf:
Non-blocking, but any way we could have caught this in the tests?

erezrokah:
> Non-blocking, but any way we could have caught this in the tests?

Yes, we could simulate pagination in the tests I think (like you did with issues). But I'm kind of worried that it means implementing that logic. Also it will be only useful here as in all other cases we use a pager interface.

erezrokah:
> Yes, we could simulate pagination in the tests I think (like you did with Snyk issues). But I'm kind of worried that it means implementing that logic. Also it will be only useful here as in all other cases we use a pager interface.

I'll see if I can do a quick follow up to this with a test

PR #10292, opened 2023-04-25
chore(meilisearch): Remove redundant check (`IsValid(i) == !IsNull(i)`)

candiduslynx:


erezrokah (approved):

candiduslynx:
@erezrokah this doesn't require priority merge & hotfix release, as it is, really, a chore (basically, we were checking for the same value twice, that's all)

PR #10300, opened 2023-04-25
feat(azure-resources): Include signedIdentifiers in `azure_storage_tables`

### LINKED ISSUE #10142: feat: Add signedIdentifiers to azure_storage_tables source plugin
### Describe the resource

> List of stored access policies specified on the table.
> https://learn.microsoft.com/en-us/rest/api/storagerp/table/list?tabs=HTTP#tablesignedidentifier

Add access policies configured on the table.

### Use Case

Analyze the access policies attached an Azure Storage Table

### Link to API documentation

https://learn.microsoft.com/en-us/rest/api/storagerp/table/list?tabs=HTTP#tablesignedidentifier

### Additional Context

Very similar to #8339

### Pull request (optional)

disq:
Fixes #10142


plugins/source/azure/resources/services/storage/tables_fetch.go
```diff
@@ -29,3 +29,25 @@ func fetchTables(ctx context.Context, meta schema.ClientMeta, parent *schema.Res
...
+	}
+	item := res.Parent.Item.(*armstorage.Account)
+	group, err := client.ParseResourceGroup(*item.ID)
+	if err != nil {
+		return err
+	}
+
+	t := res.Item.(*armstorage.Table)
+
+	v, err := svc.Get(ctx, group, *item.Name, *t.Name, nil)
```
disq, line 47: `List` only seems to include `properties.tableName` and no `properties.signedIdentifiers`. So we're basically making a second API call just to get those. We could also move it to a different table (`azure_storage_tables_signed_identifiers`?) so that it can be excluded to avoid the extra API call?

erezrokah (approved):
This looks good and not be a separate table I think.
I would also open an issue here https://github.com/Azure/azure-sdk-for-go/issues to ask if the list call behavior is expected.

Looks like `signedIdentifiers` is missing from the example responses both for `list` and `get`:
https://learn.microsoft.com/en-us/rest/api/storagerp/table/list?tabs=HTTP#tableoperationlist
https://learn.microsoft.com/en-us/rest/api/storagerp/table/get?tabs=HTTP#tableoperationget

disq:
> This looks good and not be a separate table I think. I would also open an issue here [Azure/azure-sdk-for-go/issues](https://github.com/Azure/azure-sdk-for-go/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) to ask if the list call behavior is expected.
> 
> Looks like `signedIdentifiers` is missing from the example responses both for `list` and `get`: [learn.microsoft.com/en-us/rest/api/storagerp/table/list?tabs=HTTP#tableoperationlist](https://learn.microsoft.com/en-us/rest/api/storagerp/table/list?tabs=HTTP#tableoperationlist) [learn.microsoft.com/en-us/rest/api/storagerp/table/get?tabs=HTTP#tableoperationget](https://learn.microsoft.com/en-us/rest/api/storagerp/table/get?tabs=HTTP#tableoperationget)

signedIdentifiers might not be there if there's no policy attached though?

disq:
https://github.com/Azure/azure-sdk-for-go/issues/20691

PR #10354, opened 2023-04-25
fix(neo4j): Truncate Sync Timestamp to Microsecond Precision 

bbernays:
#### Summary

In Go the precision for `time.Now` depends on the operating system... https://github.com/golang/go/issues/41087

This is the minimum required fix to get this functionality working


yevgenypats (approved):
Good find! CI fails though.

candiduslynx (approved):

PR #10362, opened 2023-04-26
fix(azure): Improve query of CIS v1.3.0 Section 4.2.1

aruneko:
#### Summary
I noticed that the current query of CIS v1.3.0 Section 4.2.1 checks the ATP (Advanced Threat Protection) status by SQL databases. But the original document says that the we have to check the status by SQL Servers.

In addition, even though I have enabled the ATP settings on my Azure SQL Server, current query always returns `fails`, because the status is managed by not database level but server level.

To solve the problem, I have added a new table `azure_sql_server_advanced_threat_protection_settings` to store the status, and use the table to check the policy.


aruneko:
Hello, @yevgenypats -san, how is the status of reviewing this pull request?

hermanschaaf (approved):
Looks good to me, sorry for the delay in reviewing this @aruneko ! Thank you for the great contribution 👍

PR #10370, opened 2023-04-26
doc(deployment): Airflow deployment

hermanschaaf:
A basic guide on how to orchestrate CloudQuery syncs using Airflow and Kubernetes. There is a lot more that could be said here, but hopefully this can serve as a useful starting point to those wishing to use Airflow. In the future we can also follow up with more operators.


website/pages/docs/deployment/airflow.md
```diff
@@ -0,0 +1,163 @@
+---
+title: Orchestrating CloudQuery Syncs with Airflow
```
bbernays, line 2: I think it should be clear from the title that it is Airflow on K8s rather than just plain Airflow. The distinction is important because AWS and GCP both have managed airflow services

yevgenypats (approved):

website/pages/docs/deployment/airflow.md
hermanschaaf, line 2:
```suggestion
title: Orchestrating CloudQuery Syncs with Airflow on Kubernetes
```

website/pages/docs/deployment/airflow.md
```diff
@@ -0,0 +1,163 @@
...
+tag: tutorial
+date: 2023/04/26
+---
+
+# Orchestrating CloudQuery Syncs with Apache Airflow
```
hermanschaaf, line 7:
```suggestion
# Orchestrating CloudQuery Syncs with Apache Airflow and Kubernetes
```

website/pages/docs/deployment/airflow.md
hermanschaaf: Good point, updated. I hope in the future we can add a few more examples (maybe on a different page) about deploying to the managed AWS / GCP versions or deploying using the BashOperator.

PR #10376, opened 2023-04-26
feat(hubspot): Add properties support on hubspot source plugin

### LINKED ISSUE #10373: feat: Add properties spec to hubspot source plugin
### Describe the resource

Currently the hubspot plugin only enables us to collect objects from hubspot and its top-level properties that its API return by default:
```
"createdate": "2023-04-13T09:07:29.661Z",
"domain": "...",
"hs_lastmodifieddate": "2023-04-13T09:08:14.185Z",
"hs_object_id": "...",
"name": "..."
```

Although, most of the actual data and value that comes from HubSpot resides on the `properties` of these entities. For example, a `Company` entity might have a custom property named "company_type" that defines if the company is a `end-user` or `trader`.

Currently, getting these data points (also other hubspot default properties such as `hs_lead_status`) is impossible.

Ideally the current resources would have a Spec for each entity that would enable users to specify all the additional properties that they want to retrieve from hubspot.

### Use Case

As a user, I would like to be able to specify additional properties to collect from the hubspot source. An example:

```
kind: source
spec:
  name: "hubspot"
...
  spec:
    max_requests_per_second: 5
    company:
      properties: ["hs_lead_status", "hs_created_by_user_id", "company_type"]
    contact:
      properties: ["lifecyclestage"]
```

This would apply to every single resource within this plugin:
* companies
* contacts
* deals
* line_items
* owners
* pipelines
* products
* quotes
* tickets

### Link to API documentation

https://developers.hubspot.com/docs/api/crm/companies#properties

### Additional Context

As it seems, the SDK being used `go-hubspot` already supports passing properties into the client when calling the API, as it can be seen here:

https://github.com/clarkmcc/go-hubspot/blob/20c2f9cbf936/generated/v3/companies/api_basic.go#L482

### Pull request (optional)

fsschmitt:
#### Summary

Closes  #10373

The goal of this PR is to add support for properties when capturing information from HubSpot, this enables users to specify the properties for each entity that they want to retrieve.

This mimics the behaviour of [HubSpot API](https://developers.hubspot.com/docs/api/crm/companies#properties), where the user needs to pass as a query parameter the properties that it wants to be returned on the API call.

This will enable users to add the properties for each entity that they want to be extracted using  cloud query, as up until now, only the "core properties" were being collected:

### Before
A company would only return the core properties:
```
"properties": {
  "createdate": "2023-04-13T09:07:29.661Z",
  "domain": null,
  "hs_lastmodifieddate": "2023-04-13T09:08:14.185Z",
  "hs_object_id": "6316510341",
  "name": "John Doe Corp"
}
```

### After
The cloudquery user can specify the **additional** properties to be extracted, such as:
```
spec:
  companies:
    properties: ["hs_lead_status"]
```
response:
```
"properties": {
  "createdate": "2023-04-13T09:07:29.661Z",
  "domain": null,
  "hs_lastmodifieddate": "2023-04-13T09:08:14.185Z",
  "hs_object_id": "6316510341",
  "hs_lead_status": "cold",
  "name": "John Doe Corp"
}
```


yevgenypats:
@fsschmitt Thanks for this contribution! will review this shortly!

plugins/source/hubspot/client/spec.go
```diff
@@ -1,7 +1,42 @@
 package client
 
 type Spec struct {
-	MaxRequestsPerSecond *int `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
+	MaxRequestsPerSecond *int          `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
+	Companies            CompaniesSpec `yaml:"companies,omitempty" json:"companies,omitempty"`
```
candiduslynx, line 5: I feel like the following could be done instead:
```go
type TableOptions struct {
	Properties []string `yaml:"properties,omitempty" json:"properties,omitempty"`
}
```

And the spec would just have an extra field:
```go
TableOptions map[string]*TableOptions `yaml:"table_options,omitempty" json:"table_options,omitempty"`
```

This way in the spec you would just have:
```yaml
table_options:
  hubspot_crm_companies:
  - property1
  - property2
  hubspot_crm_deals:
  - deal_property
```

Additionally, it will make this code much more extendable and easy to maintain.

candiduslynx: @fsschmitt thank you for the PR!
Could you evaluate the `TableOptions` struct I propose in the comment above?

candiduslynx: > it would allow to extend each entity specs to include

That's one of the reasons why I proposed `TableOptions` struct instead of `map[string][]string` or something similar.

I do understand that HubSpot resources can have specific options for the API calls, but I just wanted to weigh in the possibility to use the easier approach until the specific table options are being implemented and do not fit into the common table options pattern.

I'd also love to get an input from @yevgenypats before going forward in any of those directions.

yevgenypats: Yeah I agree with @candiduslynx here. I would go with something of the above:

```
type TableOptions struct {
  properties []string
  propertiesWithHistory []string
}
```

as it seems all tables has the same options given all hubspot tables have those methods. If there will be tables that don't have those methods we can always add the validation to exclude those.

I think the `TableOptions` is hubspot specific as you said but it's just mentioned in the internal spec which is fine. We can also name it `api_options` or any other name would work but I think we should use one dynamic map config instead of static per table.

fsschmitt: Understood, makes total sense as all of them (except pipeline and owners) would follow the same pattern we can make them under the same struct.

Nice catch and thank you for the suggestion! I'll update it right now.

plugins/source/hubspot/client/spec.go
```diff
@@ -1,7 +1,12 @@
 package client
 
 type Spec struct {
-	MaxRequestsPerSecond *int `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
+	MaxRequestsPerSecond *int                     `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
+	TableOptions         map[string]*TableOptions `yaml:"table_options,omitempty" json:"table_options,omitempty"`
+}
+
+type TableOptions struct {
```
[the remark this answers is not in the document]
fsschmitt: Sounds like a good idea to reduce code duplication. Thanks! I've went ahead and changed the `Properties()` function to `GetProperties()` as we're not allowed to have a field (`Properties` within the struct) and a function with the same naming. 👍

plugins/source/hubspot/client/spec.go
```diff
@@ -1,10 +1,12 @@
 package client
 
 type Spec struct {
-	MaxRequestsPerSecond *int                     `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
-	TableOptions         map[string]*TableOptions `yaml:"table_options,omitempty" json:"table_options,omitempty"`
+	MaxRequestsPerSecond *int           `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty"`
+	TableOptions         TablesSettings `yaml:"table_options,omitempty" json:"table_options,omitempty"`
```
candiduslynx, line 5:
```suggestion
	TablesSettings         TablesSettings `yaml:"table_options,omitempty" json:"table_settings,omitempty"`
```
for consistency (or the other way around)

candiduslynx (changes requested):

website/pages/docs/plugins/sources/hubspot/overview.mdx
```diff
@@ -29,4 +29,28 @@ You can use the `max_requests_per_second` configuration option to change this va
...
+
+- `table_name` ([TableOptionsSpec](#table-options-spec) spec, optional. Default: `empty`)
+
+  Table name to define the options when collecting data from the entity on HubSpot.
+
+### Table Options Spec
+
+- `properties` (`[]string`, optional. Default: `[]`)
+
+  Additional properties to be retrieved from HubSpot when syncing the table entity
```
candiduslynx, line 56:
```suggestion
- `table_options` (map[string][TableOptions](#table-options) spec, optional. Default: `empty`)

  Table Options for HubSpot entities that will be synced.

### Table Options

- `properties` (`[]string`, optional. Default: empty)

  Additional properties to be retrieved from HubSpot when syncing the table entity
```

candiduslynx (changes requested):

website/pages/docs/plugins/sources/hubspot/overview.mdx
candiduslynx: no need for extra section about table options spec. map[string]TableOptions + section about TableOptions is fine

candiduslynx (approved):

yevgenypats (approved):

PR #10425, opened 2023-04-28
fix(postgresql): Consistent migration for append mode

candiduslynx:


plugins/destination/postgresql/client/migrate.go
```diff
@@ -151,7 +151,7 @@ func (c *Client) normalizeTable(table *arrow.Schema, pgTable *arrow.Schema) *arr
 	fields := make([]arrow.Field, len(table.Fields()))
 	for i, f := range table.Fields() {
 		metadata := make(map[string]string, 0)
-		if !schema.IsPk(f) {
+		if !schema.IsPk(f) || !c.enabledPks() {
 			metadata[schema.MetadataPrimaryKey] = schema.MetadataFalse
 		}
 		if c.enabledPks() && schema.IsPk(f) {
```
hermanschaaf, line 157: This is now the opposite of the case above, so probably for clarity we should just use `else` ?

hermanschaaf (approved):

hermanschaaf (approved):

PR #10453, opened 2023-04-28
fix(aws): Explicitly set region in all SDK calls

### LINKED ISSUE #10452: bug(AWS): excessive memory consumption with big number of accounts

### Current Behavior

Right now when having more than 1000 or 10,000 different AWS accounts the AWS client create all AWS services for every single account and every single **region**. 

image

As a quick benchmark - having 10,000 accounts will mean having at least 25GB RAM consistently.

### Expected Behavior

The amount of RAM shouldn't grow with the number of accounts.

What we need to do as a first step is to only create all AWS services per account and not per region. This in itself will will drop the memory consumption for 10,000 accounts from 25GB to under 1GB give or take (given 30 ~regions).

I believe we already had it in the past but for some reason I don't recall we unfortunately changed it. after fixing it we can introduce a regression test that will prevent such change in the future going in circles :) 

-

### Steps To Reproduce

-

-

### CloudQuery version

-

### Additional Context

_No response_

### Pull request (optional)

yevgenypats:
This updates all AWS resolvers to explicitly set the region in SDK calls. This also now gets validated by our mock tests. We are doing this so that in a follow-up release we can reduce memory usage for AWS clients by ~20x by not needing to keep a separate client for every region.

Part of #10452


hermanschaaf (approved):

hermanschaaf:
We touched too many files in this PR, so the `Wait for all required workflows to pass` step will fail. However, we know that all the tests that had to run pass, so I will manually merge once all the other steps have passed.

PR #10456, opened 2023-04-28
fix: Add assertions generated mocks to check use of Region

hermanschaaf:
(This PR is to the `fix/aws_mem` branch)

This adds a quick and dirty script to modify all our mocks to also assert that Region is set when they are called. This could absolutely be done in better ways, but this was quick to write and seems to work without issues.

This makes every AWS test fail, and shows that we'll need to update 950 places in the code for all the tests to pass again.


yevgenypats (approved):
nice!

PR #10457, opened 2023-04-28
feat!: Migrate BigQuery to Plugin SDK v3 for native Arrow support

### LINKED ISSUE #10713: feat: Migrate `plugin/destination/bigquery` to `github.com/cloudquery/plugin-sdk/v3`

Part of #9846

Upgrade the code.

### Pull request (optional)

hermanschaaf:
Closes #10713

BEGIN_COMMIT_OVERRIDE

END_COMMIT_OVERRIDE


plugins/destination/bigquery/go.mod
```diff
@@ -3,19 +3,24 @@ module github.com/cloudquery/cloudquery/plugins/destination/bigquery
...
+	github.com/apache/arrow/go/v13 v13.0.0-20230509040948-de6c3cd2b604
+	github.com/cloudquery/plugin-pb-go v1.0.8
+	github.com/cloudquery/plugin-sdk/v3 v3.5.1
+	github.com/goccy/go-json v0.9.11
 	github.com/rs/zerolog v1.29.0
 	golang.org/x/sync v0.1.0
 	google.golang.org/api v0.114.0
 )
 
+replace github.com/cloudquery/plugin-sdk/v3 v3.2.0 => ../../../../plugin-sdk
```
candiduslynx, line 16: is it still required?

candiduslynx (changes requested):

plugins/destination/bigquery/go.mod
hermanschaaf: Sorry this wasn't actually ready for review, I thought it was still in draft

plugins/destination/bigquery/client/types.go
```diff
@@ -2,46 +2,208 @@ package client
...
+		return bigquery.RecordFieldType
+	case typeOneOf(dataType,
+		arrow.FixedWidthTypes.MonthDayNanoInterval):
+		return bigquery.RecordFieldType
+	case typeOneOf(dataType,
+		types.ExtensionTypes.Inet,
+		types.ExtensionTypes.MAC,
+		types.ExtensionTypes.UUID):
+		return bigquery.StringFieldType
+	case typeOneOf(dataType,
```
yevgenypats, line 125: Might be nice also to be part of arrow

yevgenypats (approved):
Looks good. Didn't deep dive but as long as the tests pass I think it should be good to go for now, as easy to do any additional performance or code changes later in plugins.

plugins/destination/bigquery/client/read.go
```diff
@@ -2,89 +2,28 @@ package client
...
-
-func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- []any) error {
-	colNames := make([]string, 0, len(table.Columns))
-	for _, col := range table.Columns {
-		colNames = append(colNames, col.Name)
-	}
-	cols := "`" + strings.Join(colNames, "`, `") + "`"
-	stmt := fmt.Sprintf(readSQL, cols, c.pluginSpec.ProjectID, c.pluginSpec.DatasetID, table.Name)
+func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- arrow.Record) error {
+	colSQL := "`" + strings.Join(table.Columns.Names(), "`, `") + "`"
```
candiduslynx, line 25: nice!

plugins/destination/bigquery/client/read.go
```diff
@@ -97,20 +36,153 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
-			v[i] = values[i]
+			readItem(values, rb, i)
 		}
-		res <- v
+		res <- rb.NewRecord()
 	}
 	return nil
 }
+
+func readItem(values []bigquery.Value, rb *array.RecordBuilder, i int) error {
```
candiduslynx, line 58: maybe ditch this func and use `appendValue` directly?

plugins/destination/bigquery/client/read.go
```diff
@@ -97,20 +36,153 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
+		arr.Append(true)
+		valBuilder := arr.ValueBuilder()
+		for _, v := range lst {
+			if err := appendValue(valBuilder, v); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	switch rbv := bldr.(type) {
```
candiduslynx, line 95: maybe just continue the previous `switch`?

plugins/destination/bigquery/client/read.go
```diff
@@ -97,20 +36,153 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
+	value := values[i]
+	return appendValue(fieldBuilder, value)
+}
+
+func appendValue(bldr array.Builder, value any) error {
+	if value == nil {
+		bldr.AppendNull()
+		return nil
+	}
+	switch arr := bldr.(type) {
```
candiduslynx, line 69:
```suggestion
	switch bldr := bldr.(type) {
```
could improve readability, as `arr` suggests that we have `arrow.Array` rather then `array.Builder`

plugins/destination/bigquery/client/read.go
```diff
@@ -97,20 +36,153 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
+		rbv.Append(arrow.MonthInterval(t[0].(int64)))
+		return nil
+	case *array.TimestampBuilder:
+		unit := bldr.Type().(*arrow.TimestampType).Unit
+		switch unit {
+		case arrow.Nanosecond:
+			v := value.([]bigquery.Value)
+			t := v[0].(time.Time)
+			nano := v[1].(int64)
+			nanoTime := t.Add(time.Duration(nano) * time.Nanosecond)
```
candiduslynx, line 142:
```suggestion
			nanoTime := t.Add(time.Duration(nano))
```
as `time.Nanosecond` = `1`

plugins/destination/bigquery/client/read.go
```diff
@@ -97,20 +36,153 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
+			ts, err := arrow.TimestampFromString(nanoTime.Format(time.RFC3339Nano), arrow.Nanosecond)
+			if err != nil {
+				return fmt.Errorf("failed to call arrow.TimestampFromString: %w", err)
```
candiduslynx, line 145:
```suggestion
				return err
```
to match the next case

plugins/destination/bigquery/client/types.go
```diff
@@ -2,46 +2,208 @@ package client
...
+	sc := bigquery.FieldSchema{
+		Name:        col.Name,
+		Description: col.Description,
+		Type:        c.DataTypeToBigQueryType(col.Type),
+		Schema:      c.DataTypeToBigQuerySchema(col.Type),
+	}
+
+	// TODO: handle repeated; we currently don't handle the case where we get a list of lists,
+	//       but that's not a valid case right now and isn't being explicitly tested for.
+	if col.Type.ID() == arrow.LIST {
```
candiduslynx, line 34: You need to check for other list types as well (per [this](https://github.com/apache/arrow/blob/fbe5f641d327/go/arrow/datatype.go#L376-L382) code).
Probably, this would do better:
```suggestion
	if arrow.IsListLike(col.Type) && col.Type.ID() != arrow.MAP {
```

plugins/destination/bigquery/client/types.go
```diff
@@ -2,46 +2,208 @@ package client
...
+}
+
+func (c *Client) DataTypeToBigQueryType(dataType arrow.DataType) bigquery.FieldType {
+	// handle complex types
+	switch dataType.ID() {
+	case arrow.MAP:
+		return bigquery.JSONFieldType
+	case arrow.STRUCT:
+		return bigquery.RecordFieldType
+	case arrow.LIST:
```
candiduslynx, line 47: fixed size list?

plugins/destination/bigquery/client/types.go
```diff
@@ -2,46 +2,208 @@ package client
...
+		v := dataType.(*arrow.StructType)
+		fields := make([]*bigquery.FieldSchema, len(v.Fields()))
+		for i, field := range v.Fields() {
+			fields[i] = c.ColumnToBigQuerySchema(schema.Column{
+				Name: field.Name,
+				Type: field.Type,
+			})
+		}
+		return fields
+	case arrow.IsListLike(dataType.ID()):
```
candiduslynx, line 145: you should probably add a case for map handling above, as `arrow.IsListLike` matches maps, too

plugins/destination/bigquery/client/write.go
```diff
@@ -58,3 +64,62 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resou
...
+	case array.ListLike:
+		arr := col.(array.ListLike)
+		elems := make([]any, 0, arr.Len())
+		for j := 0; j < arr.Len(); j++ {
+			if arr.IsNull(j) {
+				continue
+			}
+			from, to := arr.ValueOffsets(j)
+			slc := array.NewSlice(arr.ListValues(), from, to)
+			for k := 0; k < int(to-from); k++ {
```
candiduslynx, line 90:
```suggestion
			for k := 0; k < slc.Len(); k++ {
```

plugins/destination/bigquery/client/write.go
```diff
@@ -58,3 +64,62 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, resou
...
+			format := "2006-01-02 15:04:05.999999"
+			return TimestampNanoseconds{
+				Timestamp:   t.Format(format),
+				Nanoseconds: t.Nanosecond() % 1000,
+			}
+		default:
+			return v.GetOneForMarshal(i)
+		}
+	case *types.JSONArray:
+		return v.ValueStr(i)
```
candiduslynx, line 122: Maybe using
```suggestion
		return v. GetOneForMarshal(i)
```
could be better?

candiduslynx (commented):
Somme comments

plugins/destination/bigquery/client/write.go
hermanschaaf: I had `GetOneForMarshal` before, but it doesn't work. Something about it being `json.RawMessage` seems to confuse BigQuery and it throws an error about it not being a record column

plugins/destination/bigquery/client/types.go
hermanschaaf: We don't have tests for maps yet, so I haven't handled them in this PR

plugins/destination/bigquery/client/types.go
hermanschaaf: Hmm, we should add a test for it if we want to support it.

plugins/destination/bigquery/client/read.go
hermanschaaf: I think it's okay to include the units for clarity / readability, even if the internal implementation makes it a no-op. If it was anything other than than `Nanosecond` we would have had to include it.

plugins/destination/bigquery/client/types.go
hermanschaaf: Thanks, updated

plugins/destination/bigquery/client/read.go
hermanschaaf: Thanks!

plugins/destination/bigquery/client/types.go
hermanschaaf: I've added some handling, but since we're not testing it I think officially we should say it's not supported yet

plugins/destination/bigquery/client/read.go
```diff
@@ -97,20 +36,145 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
...
+		m := value.([]bigquery.Value)
+		bldr.Append(true)
+		for f := 0; f < bldr.NumField(); f++ {
+			fieldBldr := bldr.FieldBuilder(f)
+			if err := appendValue(fieldBldr, m[f]); err != nil {
+				return err
+			}
+		}
+		return nil
+	case array.ListLikeBuilder:
```
candiduslynx, line 77: Just a note: it also matches map builder

candiduslynx (approved):

PR #10472, opened 2023-05-01
fix: AWS mem S-X

disq:


plugins/source/aws/resources/services/s3/buckets.go
```diff
@@ -73,12 +73,14 @@ func listBucketRegion(cl *client.Client) string {
...
 	bucket := r.Item.(types.Bucket)
 	resource := &models.WrappedBucket{Name: bucket.Name, CreationDate: bucket.CreationDate}
-	c := meta.(*client.Client)
-	mgr := c.Services().S3manager
+	cl := meta.(*client.Client)
+	mgr := cl.Services().S3manager
 
-	output, err := mgr.GetBucketRegion(ctx, *resource.Name)
+	output, err := mgr.GetBucketRegion(ctx, *resource.Name, func(o *s3.Options) {
+		o.Region = cl.Region
```
disq, line 80: Not sure if this should get the client's region or the output of `listBucketRegion(cl)` which decides on the "base s3 region" based on the client partition. @bbernays

bbernays: It should be the output of the `listBucketRegion(cl)` ... ListBuckets is a global/accountwide call... Bucket configurations are specific to a region

hermanschaaf (approved):

PR #10473, opened 2023-05-01
fix(postgresql): Ensure all table names are escaped

bbernays:
#### Summary

user reported unable to sync postgresql table named `user` (which is a pg reserved word). This PR adds sanitization to table names to ensure all identifiers are properly handled


yevgenypats (approved):
Nice!

hermanschaaf (approved):

tausif3:
Thanks for the patch :) When can I expect to see it in production?

hermanschaaf:
@tausif3 This change should be released as PostgreSQL source plugin `v1.1.1` within the next hour or so.

PR #10478, opened 2023-05-01
fix: AWS Memory set region when listing accounts

hermanschaaf:


disq (approved):

plugins/source/aws/client/organizations.go
```diff
@@ -28,7 +28,11 @@ func loadOrgAccounts(ctx context.Context, logger zerolog.Logger, awsPluginSpec *
 		return nil, nil, err
 	}
 	svc := organizations.NewFromConfig(awsCfg)
-	accounts, err := loadAccounts(ctx, awsPluginSpec, svc)
+	region := awsCfg.Region
+	if region == "" {
+		region = defaultRegion
+	}
```
bbernays, line 34: I think AWS Orgs is only supported in `us-east-1` so maybe we should just hard code it?

hermanschaaf: Sure, we can do that if you think that's correct; looking at `NewFromConfig`, it was using the region from the config before though, so maybe we should stick to that in case there are edge cases we're not aware of?

bbernays: Ok. In that case I am good with it

bbernays (approved):

yevgenypats (approved):

PR #10479, opened 2023-05-01
feat(aws): Make Cost And Forecast tables fixed time intervals

bbernays:
#### Summary

Original implementation had variable time windows for forecast and cost tables based on the current date. This makes the window static so for forecast it is always 30 days in the future (with `DAILY` granularity) and `cost` is always 30 days in the past (also with `DAILY` granularity)

This is not labeled as breaking because the tables have never been released only ever merged


website/tables/aws/aws_costexplorer_cost_30d.md
```diff
@@ -1,6 +1,6 @@
-# Table: aws_costexplorer_cost_current_month
+# Table: aws_costexplorer_cost_30d
 
-This table shows data for AWS Cost Explorer Cost Current Month.
+This table shows data for AWS Cost Explorer Cost 30d.
```
disq, line 3: nit: To improve this you could change the `titleTransformer` in `resources/plugin/plugin.go` and add an override (`awsExceptions` is just for the service name) so it says "AWS Cost Explorer costs for the last 30 days" etc...

disq (approved):

website/tables/aws/aws_costexplorer_cost_30d.md
bbernays: No need for exceptions in this case as we can just hard code the title... Once we get more similar exceptions (beyond just the service name)...

PR #10501, opened 2023-05-02
chore(postgresql): Update destination to SDK V3

yevgenypats:
Example for this sdk update: https://github.com/cloudquery/plugin-sdk/pull/832


yevgenypats:
New implementation here: https://github.com/cloudquery/cloudquery/pull/10783

Issue #7711, opened 2023-02-06
AWS Config Resources


jsonpr:
Retention Configuration: https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeRetentionConfigurations.html
Configuration Aggregators: https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeConfigurationAggregators.html
Remediation Configuration: https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeRemediationConfigurations.html

paul-e-allen:
I'd like to make a pitch to add resource-specific Config Rule evaluation results via the GetAggregateComplianceDetailsByConfigRule:
https://docs.aws.amazon.com/config/latest/APIReference/API_GetAggregateComplianceDetailsByConfigRule.html

These date can augment traditional CQ Policy query results to provide resource-specific compliance to Config Rules without having to create new Policy SQL (except to pull the results from the CQ tables once they are ingested.

hermanschaaf:
@paul-e-allen If I am not mistaken, the `aws_config_config_rule_compliances` table already provides this data, just not in aggregated form: https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_config_config_rule_compliances. Could you check whether this satisfies your use case?

hermanschaaf:
Thanks for the clarification @paul-e-allen, that makes sense. I've opened a PR for adding this table in https://github.com/cloudquery/cloudquery/pull/10544. I have no idea how many rows this will represent for a typical organization (and thus how much data it will pull), but I figure it can always be skipped if it's too much.

PR #10544, opened 2023-05-02
feat: Add AWS Config Compliance Details table

hermanschaaf:
This adds an `aws_config_config_rule_compliance_details` table, requested by a user in the thread on https://github.com/cloudquery/cloudquery/issues/7711#issuecomment-1531342412. 

This table allows you to get evaluation results for resources discovered through AWS config rule evaluations, which you can then query together with other CloudQuery resources.


plugins/source/aws/resources/services/config/retention_configurations.go
```diff
@@ -31,8 +31,7 @@ func fetchRetentionConfigurations(ctx context.Context, meta schema.ClientMeta, p
 	c := meta.(*client.Client)
 	svc := c.Services().Configservice
 
-	input := &configservice.DescribeRetentionConfigurationsInput{}
-	p := configservice.NewDescribeRetentionConfigurationsPaginator(svc, input)
+	p := configservice.NewDescribeRetentionConfigurationsPaginator(svc, nil)
```
hermanschaaf, line 34: Just a bit of clean-up requested in a previous PR; no functional changes

plugins/source/aws/resources/services/config/config_rule_compliance_details.go
```diff
@@ -0,0 +1,65 @@
...
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(true),
+			client.DefaultRegionColumn(true),
+			{
+				Name:     "config_rule_name",
+				Type:     schema.TypeString,
+				Resolver: schema.ParentColumnResolver("config_rule_name"),
+			},
+			{
+				Name:     "_evaluation_hash",
```
hermanschaaf, line 34: I'm not sure this is the best way to go; we could also not have a PK at all and have it default to `_cq_id`. Thoughts on this?

plugins/source/aws/resources/services/config/config_rule_compliance_details.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v2/schema"
+	"github.com/cloudquery/plugin-sdk/v2/transformers"
+)
+
+func configRuleComplianceDetails() *schema.Table {
+	tableName := "aws_config_config_rule_compliance_details"
+	return &schema.Table{
+		Name:        tableName,
+		Description: `https://docs.aws.amazon.com/config/latest/APIReference/API_GetComplianceDetailsByConfigRule.html`,
```
disq, line 17:
```suggestion
		Description: `https://docs.aws.amazon.com/config/latest/APIReference/API_EvaluationResult.html`,
```
Maybe? We generally link to the struct if there's a separate doc/type for it, and only link to the API call if the type is not separately documented.

hermanschaaf: Yeah good point, probably better to change it.

plugins/source/aws/resources/services/config/config_rule_compliance_details.go
disq: Defaulting to `_cq_id` could be better because if the user wants deterministic they can set `deterministic_cq_id` in the spec and have the _cq_id generated from all column values automatically. On the other hand that setting affects all tables.

hermanschaaf: Yeah exactly, `deterministic_cq_id` affects all tables which is probably not what users would want. We do use this `ResolveObjectHash` technique in two other places as well, but I think that predated `deterministic_cq_id`. What do you think - just use `_cq_id`?

disq: I'm unsure but yes the `ResolveObjectHash` method predates deterministic_cq_id so it's probably better to just keep it simple.

plugins/source/aws/resources/services/config/config_rule_compliance_details.go
disq: Other config resources may also have the same issue I don't think I paid attention to the description in those.

disq (approved):

PR #10550, opened 2023-05-03
fix(aws): Use GetBucketLocation for S3 bucket region

### LINKED ISSUE #10548: bug: unexpected anonymous HeadBucket API calls from CloudQuery

### Current Behavior

The call `s3manager.GetBucketRegion()` makes unsigned API requests generating unauthorized anonymous access attempts in AWS CloudTrail logs.

> The request will not be signed, and will not use your AWS credentials.
> https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#GetBucketRegion

### Expected Behavior

CloudQuery retrieve the S3 Bucket Region with an authenticated call. `GetBucketLocation` would be appropiate.

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Client.GetBucketLocation

Any CQ config with aws_s3_buckets as a table.

### Steps To Reproduce

Run a CQ config with `aws_s3_buckets` as a table in DEBUG mode.
You will see unsigned HEAD requests to buckets that respond with a 403.

n/a

### CloudQuery version

3.0.2

### Additional Context

_No response_

### Pull request (optional)

disq:
This is to prevent an s3manager unauthenticated call to get region (which then show up on CloudTrail events)

Fixes #10548


hermanschaaf (approved):

bbernays (approved):
This looks good... But my only question is why did it work? If it was making unsigned calls then how was it getting the information about private buckets?

spofdamon:
@bbernays If you look at the sample request in the docs, it's unsigned but still authorized:

https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html

I think the real answer here is probably "S3 is a very very old API".

spofdamon:
I have a feeling that it will answer `HEAD` requests that confirm region without any auth whatsoever. This is probably what the API itself does when you anonymously download something from S3 without putting the region in the host.

RyPeck:
@spofdamon that's correct. golang example below to run in a Docker container without any aws config for the curious. Works for public and private buckets.

```go
package main

import (
        "context"
        "fmt"
        "log"
        "errors"
        "os"

        "github.com/aws/aws-sdk-go-v2/config"
        "github.com/aws/aws-sdk-go-v2/feature/s3/manager"
        "github.com/aws/aws-sdk-go-v2/service/s3"
)

func main() {
        ctx := context.TODO()

        cfg, err := config.LoadDefaultConfig(ctx,
                config.WithRegion("us-west-1"),
        )

        if err != nil {
                log.Println("error:", err)
                return
        }

        bucket := "asdf"

        region, err := manager.GetBucketRegion(ctx, s3.NewFromConfig(cfg), bucket)

        if err != nil {
                var bnf manager.BucketNotFound
                if errors.As(err, &bnf) {
                        fmt.Fprintf(os.Stderr, "unable to find bucket %s's region\n", bucket)
                }
                return
        }

        fmt.Printf("Bucket %s is in %s region\n", bucket, region)
}
```

PR #10554, opened 2023-05-03
fix(azure): Storage Queue ACL should use storage keys and not OAuth

### LINKED ISSUE #10546: bug: azure_storage_queue_acl erroring and not populating

### Current Behavior

Syncing `azure_storage_queue_acl` and `azure_storage_queues` I see `azure_storage_queues` successful but not the acl table.

### Expected Behavior

The ACLs for each `azure_storage_queues` populate in `azure_storage_queue_acl`

```yaml
kind: source
spec:
  # Source spec section
  name: "azure"
  path: "cloudquery/azure"
  version: "v7.4.0"
  destinations: ["file"]
  skip_dependent_tables: true

  tables:
    - azure_storage_accounts
    - azure_storage_queues
    - azure_storage_queue_acl
    
  spec:
    subscriptions: ["<subscription_id>"]
```

### Steps To Reproduce

1. Connect an Azure Subscription with a Storage Account.
2. Create a Data Storage Queue.
3. Add an Access Policy to the Queue.
4. Sync.

```
2023-05-02T16:18:58Z ERR table resolver finished with error error="GET https://ryantest20230206.queue.core.windows.net/ryantest20230502\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 The specified resource does not exist.\nERROR CODE: ResourceNotFound\n--------------------------------------------------------------------------------\n\ufeff<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>ResourceNotFound</Code><Message>The specified resource does not exist.\nRequestId:9bc713eb-3003-0052-2b11-7d9c16000000\nTime:2023-05-02T16:18:58.1562817Z</Message></Error>\n--------------------------------------------------------------------------------\n" client=subscriptions/b9b2421c-adac-4ea8-aefa-b4504a0fb6db module=azure-src table=azure_storage_queue_acl
```

### CloudQuery version

3.0.2

### Additional Context

Error still occurs when no access policies are attached to the storage queue.

Recent changes to pull in ACLs to the signedIdentifiers property for `azure_storage_tables` are working.

### Pull request (optional)

disq:
Fixes #10546

Looks like the azure cli is somewhat cheating to get the data, but it works.


plugins/source/azure/client/client.go
```diff
@@ -50,6 +51,8 @@ type Client struct {
 	BillingProfile  *armbilling.Profile
 	BillingPeriods  map[string][]*armbilling.Period
 	BillingPeriod   *armbilling.Period
+
+	storageAccountKeys *sync.Map
```
disq, line 55: This is shared among clients/subscriptions and contains a single key for each storage account.

plugins/source/azure/client/client.go
```diff
@@ -363,3 +367,31 @@ func (c *Client) withBillingPeriod(billingPeriod *armbilling.Period) *Client {
 	newC.BillingPeriod = billingPeriod
 	return &newC
 }
+
+func (c *Client) GetStorageAccountKey(ctx context.Context, acc *armstorage.Account) (string, error) {
+	if val, ok := c.storageAccountKeys.Load(*acc.Name); ok {
```
hermanschaaf, line 372: If this function happens to get called for the same storage account in parallel by 100 goroutines, we could still end up making 100 requests, because there's some delay between this line and the `Store` call lower down.

If we want to avoid that potential, we should have some locking mechanism around this. Maybe a `sync.Map` with  that indicates whether an storage account is already in the process of being loaded?

There's probably a better way of doing this, but I'm thinking something along these lines:

```
	mut := c.storageAccountLoading.LoadOrStore(*acc.Name, &sync.Mutex{})
	mut.Lock()
	defer mut.Unlock()
	if val, ok := c.storageAccountKeys.Load(*acc.Name); ok {
		return val.(string), nil
	}
```

plugins/source/azure/client/helpers.go
```diff
@@ -22,3 +23,28 @@ func ParseResourceGroup(resourceID string) (string, error) {
...
+
+func loadOrStore(m *sync.Map, key string, f func() (any, error)) (any, error) {
+	temp, _ := m.LoadOrStore(key, &syncData{
+		data: nil,
+		once: &sync.Once{},
+	})
+	d := temp.(*syncData)
+	var err error
+	if d.data == nil {
+		d.once.Do(func() {
```
disq, line 40: > Do guarantees that when it returns, f has finished.

hermanschaaf (approved):

disq:
Testing with `-race` and multiple mock test items now.

plugins/source/azure/client/helpers.go
```diff
@@ -22,3 +23,28 @@ func ParseResourceGroup(resourceID string) (string, error) {
...
+			d.data, err = f()
+			if err != nil {
+				d.data = nil
+				d.once = &sync.Once{}
+			}
+			m.Store(key, d)
+		})
+	}
+	return d.data, err
```
hermanschaaf, line 49: Nice solution, but I think there might be a weird edge case here... if `f` returns an error. In this case the function to execute `d.once.Do` will have an error and return it, but all the other callers that were waiting on the `d.once.Do` to finish will return with nil data and no error, which is not quite right. Maybe there needs to be another check at the end of the function to return an error if `d.data` is still nil, or something along those lines.

plugins/source/azure/resources/services/storage/queue_access_policy_mock_test.go
```diff
@@ -26,5 +31,31 @@ func createQueueAccessPolicy(router *mux.Router) error {
...
+	keyResp.Keys[0].Permissions = &pp
+	keyResp.Keys[0].Value = to.Ptr(base64.StdEncoding.EncodeToString([]byte("testkey")))
+
+	router.HandleFunc("/subscriptions/{subscriptionId}/resourceGroups/debug/providers/Microsoft.Storage/storageAccounts/test string/listKeys", func(w http.ResponseWriter, r *http.Request) {
+		b, err := json.Marshal(&keyResp)
+		if err != nil {
+			http.Error(w, "unable to marshal request: "+err.Error(), http.StatusBadRequest)
+			return
+		}
+		fmt.Println(string(b))
```
hermanschaaf, line 53: This should be removed 😛

plugins/source/azure/client/helpers.go
disq: Should be fixed with [`733b79a` (#10554)](https://github.com/cloudquery/cloudquery/pull/10554/commits/733b79a7facafea5222cc34f563070419ce763b5)

plugins/source/azure/client/helpers.go
```diff
@@ -22,3 +23,36 @@ func ParseResourceGroup(resourceID string) (string, error) {
...
+				dt = nil
+				d.once = &sync.Once{}
+			}
+			m.Store(key, syncData{
+				data: dt,
+				err:  err,
+				once: d.once,
+			})
+		})
+		temp, _ = m.Load(key)
```
disq, line 53: Modifying `d` inside `Do()` triggers the race detector for some reason (if ptr `*syncData` is used inside the map) so we just store and load it again.

plugins/source/azure/client/client.go
candiduslynx: there's also sync.Once to be considered for init once (per sync)

hermanschaaf (approved):
👍

plugins/source/azure/client/client.go
```diff
@@ -363,3 +368,33 @@ func (c *Client) withBillingPeriod(billingPeriod *armbilling.Period) *Client {
...
+
+		keysResponse, err := svc.ListKeys(ctx, group, *acc.Name, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		for i := range keysResponse.Keys {
+			return *keysResponse.Keys[i].Value, nil
+		}
+		return nil, ErrNoStorageKeysFound
```
candiduslynx, line 394:
```suggestion
		if len(keysResponse.Keys) == 0 {
			return nil, ErrNoStorageKeysFound
		}
		return *keysResponse.Keys[0].Value, nil
```

plugins/source/azure/resources/services/storage/queue_access_policy.go
```diff
@@ -37,14 +38,34 @@ func fetchQueueACL(ctx context.Context, meta schema.ClientMeta, parent *schema.R
 	queue := parent.Item.(*armstorage.ListQueue)
 	acc := parent.Parent.Item.(*armstorage.Account)
 
+	storageKey, err := cl.GetStorageAccountKey(ctx, acc)
+	if err == client.ErrNoStorageKeysFound || storageKey == "" {
+		var lgr *zerolog.Event
+		if err != nil && err != client.ErrNoStorageKeysFound {
+			lgr = cl.Logger().Error().Err(err)
```
candiduslynx, line 45: shouldn't we also ret the err here?

candiduslynx (commented):
a couple of suggestions

plugins/source/azure/resources/services/storage/queue_access_policy.go
disq: We did below, but simplified in [`d4a690e` (#10554)](https://github.com/cloudquery/cloudquery/pull/10554/commits/d4a690e09f529d2b2c1e4db3d27fcf720fe17a06)

candiduslynx (approved):

yevgenypats (approved):

PR #10555, opened 2023-05-03
feat(clickhouse): Map support

candiduslynx:


plugins/destination/clickhouse/typeconv/ch/values/map.go
```diff
@@ -0,0 +1,75 @@
...
+		// we need to fill in for the in-depth recursive parsing by ClickHouse SDK
+		if arr.IsNull(i) {
+			res.Index(i).Set(val)
+			continue
+		}
+
+		rowKeys, rowItems := keysValue.Index(i).Elem(), itemsValue.Index(i).Elem() // ->[]*
+		for idx := 0; idx < rowKeys.Len(); idx++ {
+			// this is matched exactly by the items
+			reflect.Ma
```
disq, line 54: 🤔

candiduslynx: yeah, I was in a hurry, so this was a draft =)

PR #10564, opened 2023-05-04
feat(aws): Support Table level inputs

bbernays:
#### Summary

In this PR we add **experimental** support for 2 tables (`aws_inspector2_findings` and `aws_cloudtrail_events`). 

`TableOptions` is a nested structure that uses the CQ table name as the first level key. After that each table uses the aws API call to hold the actual API input. So in JSON form the object looks like this:
``` json
{
    "table_options":{
        "aws_inspector2_findings":{
            "ListFindings":{}
        }
        "aws_cloudtrail_events":{
            "LookupEvents":{}
        }
    }
}
```

``` yml
  spec:
    table_options:
      aws_inspector2_findings:
        list_findings:
          filter_criteria:
            finding_status:
              - comparison: EQUALS
                value: ACTIVE
      aws_accessanalyzer_analyzer_findings:
        list_findings:
          
      aws_cloudtrail_events:
        lookup_events:
          start_time: 2023-05-01T20:20:52Z
          end_time:   2023-05-03T20:20:52Z
          lookup_attributes:
            - attribute_key: EventName
              attribute_value: RunInstances
```

Update:
1. In order to maintain all inputs in snake case, I wrote a script that inputs the input struct from the aws sdk and makes a local copy of the struct that uses the SDK caser to ensure consistent names
2. All of the logic for accessing the inputs is in its own package under the client
3. All of the validation and copying from one struct type to the sdk type is validated with unit tests


plugins/source/aws/resources/services/inspector2/findings.go
```diff
@@ -48,8 +48,20 @@ The 'request_account_id' and 'request_region' columns are added to show from whe
...
-	input := inspector2.ListFindingsInput{MaxResults: aws.Int32(100)}
-	paginator := inspector2.NewListFindingsPaginator(svc, &input)
+	var err error
+	input := &inspector2.ListFindingsInput{}
+	if c.Spec.TableOptions.Inspector2Findings != nil {
+		input, err = c.Spec.TableOptions.Inspector2Findings.ListFindings()
+		if err != nil {
+			return err
+		}
+	}
```
disq, line 58: Would it be possible to 'hold onto' defaults here? Something like `if input.Maxresults == 0 { input.MaxResults = aws.Int32(100) }`
I understand it wouldn't work for values with a valid zero value

disq (approved):

hermanschaaf:
Nice, I like this, and it's also a really unexpectedly small code change 😄  

The only issue I see that I'd like to discuss a bit, is what if users want to apply multiple filters? For example, for `aws_cloudtrail_events`, `LookupAttributes` are unfortunately limited to a single key and value (`"Currently the list can contain only one item."`). If CQ users wanted to sync more... well, they'd need to do that in an entirely different sync.

One could argue that this is a limitation of the AWS API, take it up with them. But on the other hand, we do have ways to deal with this. For example, what if instead of using:

```
type ctAPIs struct {
	LookupEventsOpts cloudtrail.LookupEventsInput `json:"LookupEvents,omitempty"`
}
```

we did:

```
type ctAPIs struct {
	LookupEventsOpts []cloudtrail.LookupEventsInput `json:"LookupEvents,omitempty"`
}
```

(so `LookupEvents` becomes a slice). Then the config becomes this:

```
  spec:
    table_options:
      aws_cloudtrail_events:
        LookupEvents:
          - StartTime: 2023-05-01T20:20:52Z
            EndTime:   2023-05-03T20:20:52Z
            LookupAttributes:
              - AttributeKey: EventName
                AttributeValue: RunInstances
```

Which is almost exactly the same as before, except there's an extra dash. However it can then be used in this way:

```
  spec:
    table_options:
      aws_cloudtrail_events:
        LookupEvents:
          - StartTime: 2023-05-01T20:20:52Z
            EndTime:   2023-05-03T20:20:52Z
            LookupAttributes:
              - AttributeKey: EventName
                AttributeValue: RunInstances
          - StartTime: 2023-05-01T20:20:52Z
            EndTime:   2023-05-03T20:20:52Z
            LookupAttributes:
              - AttributeKey: EventName
                AttributeValue: StartInstances
          - StartTime: 2023-05-01T20:20:52Z
            EndTime:   2023-05-03T20:20:52Z
            LookupAttributes:
              - AttributeKey: EventName
                AttributeValue: StopInstances
```

You'd have to repeat the start time/end time if you need this, but at least you could then fetch all the events you need in a single sync. Inside the resolver, we could loop over each entry in the list and do them one-by-one (probably in separate goroutines, but not necessarily).

What do you think?

plugins/source/aws/resources/services/inspector2/findings.go
bbernays: If we want to do that I would suggest that we return an error when calling `ListFindings()`

bbernays:
@hermanschaaf - I like the idea of taking in an array of input arguments and then iterating over them! I think if we do it for a single table we should support it for all tables/APIs that support user provided inputs. One other thing that this will possibly generate will be duplicate data as the filters could theoretically overlap. Is that a concern?

disq:
Where do we stand on account-level filtering of these options? Each account gets its own list of TableOptions. But then you'd need catch-all etc. as well so it could get out of hand quickly. Maybe some kind of an account-id-match "metafield" can be included, which could take wildcards or multiple account ids. In a future iteration.

bbernays:
> Each account gets its own list of TableOptions

At this time I think we should start with just a single "global" `TableOptions` if users need an account specific filters for specific accounts/regions they can use multiple source blocks

bbernays:
I will not be including the documentation in this PR. I am working on a follow up PR that will solely be the documentation

plugins/source/aws/resources/services/accessanalyzer/analyzers.go
```diff
@@ -42,7 +42,6 @@ func fetchAccessanalyzerAnalyzers(ctx context.Context, meta schema.ClientMeta, p
 	svc := c.Services().Accessanalyzer
 	paginator := accessanalyzer.NewListAnalyzersPaginator(svc, &accessanalyzer.ListAnalyzersInput{})
 	for paginator.HasMorePages() {
-		// no need to override API call options anymore: https://github.com/aws/aws-sdk-go-v2/issues/1260
```
bbernays, line 45: we used to override the region due to a bug in the sdk, now we do it because our sdk clients aren't per account+region...

hermanschaaf:
I had a few comments on the codegen, but hoping we can get away with a shorter solution that doesn't require any codegen; here's a branch that demonstrates the idea (and some other changes): https://github.com/cloudquery/cloudquery/compare/aws-inputs-simplified 

The basic idea is to use a custom JSON Unmarshal function that first unmarshals to a map[string]any, then transforms all the key names from snake to camel, then marshalling the map back to JSON and calling the default config object unmarshal on _that_ instead. It seems to work in my testing and it's relatively simple I think

bbernays:
@hermanschaaf - I am fine with your approach... In the end both approaches achieve the same thing with the exact same public interface

plugins/source/aws/client/table_options_test.go
```diff
@@ -0,0 +1,56 @@
...
+	if len(tableOpts.CloudTrailEvents.LookupEventsOpts.LookupAttributes) != 1 {
+		t.Fatalf("CloudTrailEvents.LookupEventsOpts.LookupAttributes = %v, want 1 item", tableOpts.CloudTrailEvents.LookupEventsOpts.LookupAttributes)
+	}
+	if tableOpts.CloudTrailEvents.LookupEventsOpts.LookupAttributes[0].AttributeKey != "EventName" {
+		t.Fatalf("CloudTrailEvents.LookupEventsOpts.LookupAttributes[0].AttributeKey = %v, want EventName", tableOpts.CloudTrailEvents.LookupEventsOpts.LookupAttributes[0].AttributeKey)
+	}
+	if *tableOpts.CloudTrailEvents.LookupEventsOpts.LookupAttributes[0].AttributeValue != "ConsoleLogin" {
+		t.Fatalf("CloudTrailEvents.LookupEventsOpts.LookupAttributes[0].AttributeKey = %v, want EventName", *tableOpts.CloudTrailEvents.LookupEventsOpts.LookupAttributes[0].AttributeValue)
+	}
+}
```
hermanschaaf, line 56: Not blocking (and I know I wrote this in the first place 😄 ) but I think it might be better to have a test like this for each type individually in the `table_options` package. That way, when a new type is added, we can copy and modify the tests there and make sure that the underscore parsing functionality is preserved for new types as well.

Or maybe, even better, would be a single test here that marshals the entire `TableOptions` object, does the inverse operation of `changeCaseForObject`, and checks that when you load that JSON the object is the same? Then we don't have to remember to add new tests for new types and it can be validated for all the fields.

Again, not blocking, I'm happy to follow up and add a test like this after this is merged if you agree it's a good idea.

hermanschaaf (changes requested):
Looks great, thanks for the update! I think the only thing remaining now is to use a list instead of individual objects so that you can combine multiple filters.

plugins/source/aws/client/table_options_test.go
hermanschaaf: Added this test in https://github.com/cloudquery/cloudquery/pull/10564/commits/96739188c1e7b530c097687c27e2b4ce3fd5e513 and fixed an edge case caught by it in https://github.com/cloudquery/cloudquery/pull/10564/commits/a614f7759b4911694113f7728f6ab52906063814

hermanschaaf:
@bbernays I pushed a couple of commits to the branch:
 - A test to validate the snake case JSON parsing is working as expected
 - A fix for an edge case discovered by the above
 - Renaming `table_options` package `tableoptions` (not great for readability, but more consistent with our other packages)
 - A commit to use lists in the config like we discussed https://github.com/cloudquery/cloudquery/pull/10564/commits/600879cf9fe915f18b3667fa10925bbd69b60a7c This was a bit trickier than I thought for Cloudtrail events, because there we have to carefully combine it with the state backend. I kept it backwards-compatible, but please have a look there and let me know whether you agree this is the right way to implement it. None of these are using additional goroutines yet. I think that's okay for now

hermanschaaf (approved):
Approving so the PR is unblocked; since I pushed a couple of commits myself though it would be good to get a review of that :)

PR #10565, opened 2023-05-05
feat(cli): Move CLI to use plugin-pb instead of plugin-sdk

yevgenypats:
WIP to remove plugin-sdk dependency from CLI


cli/cmd/sync_v0_1.go
```diff
@@ -1,162 +0,0 @@
-package cmd
```
yevgenypats, line 1: I deleted this old version and just returned an error for users to upgrade source or use an older CLI.

hermanschaaf (approved):
LGTM 👍  Lots of changes here but seems like it's mostly code moved from the SDK. I pushed one small fix so the progress bar is finished before printing the final summary, but otherwise couldn't find any issues during testing. 

Overall I think it makes a lot of sense to keep the download and plugin management code in the CLI rather than the SDK 💯

cli/cmd/sync_v1.go
```diff
@@ -153,75 +102,40 @@ func syncConnectionV1(ctx context.Context, cqDir string, sourceSpec specs.Source
...
-				_ = bar.Add(1)
-				for i := range destSubscriptions {
-					select {
-					case <-gctx.Done():
-						return gctx.Err()
-					case destSubscriptions[i] <- resource:
-					}
-				}
-			case <-t.C:
-				_ = bar.Add(0)
```
hermanschaaf, line 200: This change here seems like a regression; previously the ticker ensured that the bar would update at least once a second, so it didn't seem like the CLI was frozen during a long initialization process or period where no resources are loaded (e.g. all the API endpoints return no data)

hermanschaaf: Fixed in https://github.com/cloudquery/cloudquery/pull/10565/commits/8eb5895d1c66d747803a905390d55d806149deed

PR #10567, opened 2023-05-05
feat(digitalocean): Update DO SDK, S3 SDK and add docs

disq:


plugins/source/digitalocean/resources/services/accounts/accounts.go
```diff
@@ -8,9 +8,10 @@ import (
 
 func Accounts() *schema.Table {
 	return &schema.Table{
-		Name:      "digitalocean_accounts",
-		Resolver:  fetchAccountsAccounts,
-		Transform: transformers.TransformWithStruct(&godo.Account{}),
+		Name:        "digitalocean_accounts",
+		Description: "https://pkg.go.dev/github.com/digitalocean/godo#Account",
```
hermanschaaf, line 12: I think we should rather link to the API docs (e.g. https://docs.digitalocean.com/reference/api/api-reference/#tag/Regions), where the columns have descriptions. The Go docs don't seem to offer much more context than we already have in our list of columns

hermanschaaf (approved):
👍

PR #10571, opened 2023-05-05
feat: Add Cloudformation Template Summaries table

nicl:
#### Summary

https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_GetTemplateSummary.html

Adds the template summary as a child table of the Cloudformation 'stacks' table. The initial motivation is to get template Metadata, but the other fields seem useful too.


plugins/source/aws/resources/services/cloudformation/stacks_mock_test.go
```diff
@@ -40,6 +41,21 @@ func buildStacks(t *testing.T, ctrl *gomock.Controller) client.Services {
 		nil,
 	)
 
+	var summary cloudformation.GetTemplateSummaryOutput
+	if err := faker.FakeObject(&summary); err != nil {
+		t.Fatal(err)
+	}
+	summary.Metadata = aws.String("{}") // TODO this feels hacky - what is a good solution here?
```
nicl, line 48: Note, this 'fixes' the failing test - an issue with the JSONinfication of the field when it is faked during this test -but I'm not sure I understand things well enough here. There must be a better solution?! Input welcome.

plugins/source/aws/resources/services/cloudformation/template_summaries.go
```diff
@@ -0,0 +1,56 @@
...
+func fetchTemplateSummary(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	c := meta.(*client.Client)
+	svc := c.Services().Cloudformation
+
+	stack := parent.Item.(types.Stack)
+
+	summary, err := svc.GetTemplateSummary(ctx, &cloudformation.GetTemplateSummaryInput{
+		StackName: stack.StackName,
+	})
+	if err != nil {
```
nicl, line 50: Note: not entirely sure if we should handle certain errors here or if the ones we want to ignore are already handled - e.g. stack is not in a state where getTemplateSummary succeed, because deleted etc.

hermanschaaf: There is always a chance that the in the time between the initial request to get the stack and getting a summary here something happens (like the stack getting deleted), so we can't necessarily rely on the higher level handling it. 

That said, I think just returning the error here is fine for now. We can refine this upon further feedback if necessary, but for the time being this will make users aware of issues so they can choose to act on it (or not)

plugins/source/aws/resources/services/cloudformation/template_summaries.go
```diff
@@ -0,0 +1,56 @@
...
+			{
+				Name:     "stack_id",
+				Type:     schema.TypeString,
+				Resolver: schema.ParentColumnResolver("id"),
+			},
+			{
+				Name:     "metadata",
+				Type:     schema.TypeJSON,
+				Resolver: schema.PathResolver("Metadata"),
+			},
```
hermanschaaf, line 36: I think bringing the CF stack ARN in as a column here would be useful, and if I'm not mistaken it could also serve as a primary key. We can do something like this:

```
{
   Name:     "stack_arn",
   Type:     schema.TypeString,
   Resolver: schema.ParentColumnResolver("arn"),
   CreationOptions: schema.ColumnCreationOptions{
	PrimaryKey: true,
  },
},
```

What do you think?

hermanschaaf (changes requested):
Looks great, thanks @nicl ! Just one comment about whether we want to bring in the parent ARN as a PK

hermanschaaf (approved):

hermanschaaf:
Thanks for the contribution @nicl ! We will release this in the new version of the AWS plugin shortly

PR #10624, opened 2023-05-09
chore(website): Add jobs link

yevgenypats:
Bring back jobs link


hermanschaaf:
Our top bar is getting a bit full, maybe we can use a "Careers" link in the footer? I'm happy to make the change

hermanschaaf:
I pushed an update but GitHub is acting up again

PR #10636, opened 2023-05-09
fix(cli): Fix panic in no-migrate mode

yevgenypats:


cli/cmd/sync_test.go
```diff
@@ -66,6 +66,16 @@ func TestSync(t *testing.T) {
 			logContent := string(b)
 			require.NoError(t, logFileError, "failed to read cloudquery.log")
 			require.NotEmpty(t, logContent, "cloudquery.log empty; expected some logs")
+
+			logFileName = path.Join(cqDir, "cloudquery1.log")
```
hermanschaaf, line 70: Maybe we should use the same log file in this command, otherwise it won't get closed by the `defer CloseLogFile()` call above

cli/cmd/sync_v0.go
```diff
@@ -39,20 +39,22 @@ func syncConnectionV0_2(ctx context.Context, sourceClient *managedsource.Client,
 	if err != nil {
 		return err
 	}
+	for i := range destinationsClients {
```
candiduslynx, line 42: todo: run in parallel? (nit)

cli/cmd/sync_v1.go
```diff
@@ -55,20 +55,22 @@ func syncConnectionV1(ctx context.Context, sourceClient *managedsource.Client, d
 	if err != nil {
 		return err
 	}
+	for i := range destinationsClients {
```
candiduslynx, line 58: todo: run in parallel? (nit)

candiduslynx (approved):

PR #10688, opened 2023-05-09
fix(cli): Close send after write

candiduslynx:


cli/cmd/sync_v0.go
```diff
@@ -84,6 +85,12 @@ func syncConnectionV0_2(ctx context.Context, sourceClient *managedsource.Client,
 		return err
 	}
 	writeClients := make([]destination.Destination_Write2Client, len(destinationsPbClients))
+	defer func() {
+		for i := range writeClients {
```
yevgenypats, line 89: nil check is missing

cli/cmd/sync_v0.go
```diff
@@ -84,6 +85,12 @@ func syncConnectionV0_2(ctx context.Context, sourceClient *managedsource.Client,
...
+			_ = writeClients[i].CloseSend()
```
yevgenypats, line 90: can you check err and print it to the log so we see that we are not hiding the same error that we were fixing?

yevgenypats (changes requested):
Looks good. few comments

cli/cmd/sync_v1.go
```diff
@@ -133,9 +140,10 @@ func syncConnectionV1(ctx context.Context, sourceClient *managedsource.Client, d
 	// Read from the sync stream and write to all destinations.
 	for {
 		r, err := syncClient.Recv()
-		if err == io.EOF {
-			break
-		} else if err != nil {
+		if err != nil {
```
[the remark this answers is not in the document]
candiduslynx: it will match the wrapped err as well (if we start using some client interceptors, for instance, that might wrap the EOF error via `%w`)

cli/cmd/sync_v0.go
candiduslynx: done in 9f1eed6ecd56ca45c169ec153c9059eee5a5aed0

cli/cmd/sync_v0.go
candiduslynx: done in 9f1eed6ecd56ca45c169ec153c9059eee5a5aed0

yevgenypats (approved):

PR #10692, opened 2023-05-10
fix(aws): Fix the case where resrouce_id is null in ECS.2 of foundational policy

aruneko:
#### Summary
When I run the query of ECS.2 in foundational policies, I have noticed that resource_id became null if there is no ECS service in a ECS cluster.

Thus, we don't have to use `left join` to detect the rule. Only just seek `aws_ecs_cluster_services` table.


jsonpr (approved):
Thanks for the contribution to CloudQuery and fix.

This looks good.  Good eye!

hermanschaaf (approved):

PR #10699, opened 2023-05-10
feat(gcp)!: Make Listing Enabled Services failures not block the sync

bbernays:
#### Summary
We currently require a successful `ListServices` for all projects in the org. If a user cannot make the call for a single project then the entire sync will fail. This PR changes the listing to be a best effort, which means that if CQ is not able to make the `ListServices` call it will just assume that all services are enabled


BenitezDev:
I think it is a useful solution, in my case it would help a lot

hermanschaaf (approved):

disq (approved):

PR #10700, opened 2023-05-10
feat(gcp)!: Initialization concurrency

bbernays:
#### Summary

We currently try and use the same concurrency value that we use for resolving all tables for parallelizing `ListServices` when the default is used then the concurrency can overload the credential provider as seen by [user on discord](https://discord.com/channels/872925471417962546/1105845358535385271).

This PR creates a separate concurrency option that enables users to have a large sync concurrency but a much more limited initialization concurrency


plugins/source/gcp/client/spec.go
```diff
@@ -9,6 +9,7 @@ type Spec struct {
 	ProjectFilter         string   `json:"project_filter"`
 	BackoffDelay          int      `json:"backoff_delay"`
 	BackoffRetries        int      `json:"backoff_retries"`
+	DiscoveryConcurrency  *int     `json:"discovery_concurrency"`
```
candiduslynx, line 12: I'd suggest to keep it a plain value and check for 0 instead of nil

candiduslynx (approved):

plugins/source/gcp/client/spec.go
disq: Since 0 would be meaningless (not a good idea to have infinite concurrency?) keeping it plain is an option, OTOH having this as a pointer is a better way of signalling to the user (or code reader/maintainer) that this is an optional value. Can go either way.

plugins/source/gcp/client/spec.go
```diff
@@ -19,4 +20,9 @@ func (spec *Spec) setDefaults() {
 	if spec.FolderRecursionDepth == nil {
 		spec.FolderRecursionDepth = &defaultRecursionDepth
 	}
+
+	var defaultDiscoveryConcurrency = 100
+	if spec.DiscoveryConcurrency == nil {
+		spec.DiscoveryConcurrency = &defaultDiscoveryConcurrency
```
disq, line 26: You could define `defaultDiscoveryConcurrency` inside the block.

hermanschaaf:
Why is this marked as a breaking change? Can it not be done in a backwards-compatible way?

bbernays:
I marked it as breaking because if a user had adjusted concurrency to a value other than default this would now change their configuration.

plugins/source/gcp/client/spec.go
bbernays: I was trying to match the style in lines 19-21...

plugins/source/gcp/client/spec.go
bbernays: I was trying to keep the same style as `FolderRecursionDepth` on line 8

hermanschaaf:
@BenitezDev Good point, I think we missed it on this one! Opened an issue so we don't forget: https://github.com/cloudquery/cloudquery/issues/11170

In short you should be able to set `discover_concurrency` at the same level as other options like `backoff_retries`. It defaults to 100

PR #10752, opened 2023-05-11
fix(aws): Change column type of `aws_cloudformation_stack_templates.template_body`

bbernays:
#### Summary

Change from type string to type JSON.

This is not a breaking change because the resource has never been released


plugins/source/aws/resources/services/cloudformation/stack_templates.go
```diff
@@ -29,6 +29,11 @@ func stackTemplates() *schema.Table {
 					PrimaryKey: true,
 				},
 			},
+			{
+				Name:     "template_body",
+				Type:     schema.TypeJSON,
```
candiduslynx, line 34: 1 question: did you verify that the template will always be JSON?
Is there a possibility to see YAML, for instance?
https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_GetTemplate.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-anatomy.html

Possible workaround would be to parse as YAML always (JSON is a valid YAML) and then marshal as JSON

candiduslynx (changes requested):

plugins/source/aws/resources/services/cloudformation/stack_templates.go
bbernays: It is always JSON. You can submit a template in YAML, but it is always returned to you in JSON

candiduslynx (approved):

PR #10783, opened 2023-05-14
feat(postgresql): Update PG to SDK V3 native arrow support

yevgenypats:
Closes https://github.com/cloudquery/cloudquery/issues/10727

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -138,9 +138,8 @@ func (c *Client) normalizeTable(table *schema.Table, pgTable *schema.Table) *sch
 	normalizedTable := schema.Table{
 		Name: table.Name,
 	}
-	for _, f := range table.Columns {
-		col := f
-		if c.enabledPks() && f.PrimaryKey {
+	for _, col := range table.Columns {
+		if c.enabledPks() && col.PrimaryKey {
 			col.NotNull = true
 		} else {
 			col.PrimaryKey = false
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -196,11 +196,11 @@ func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
 func (c *Client) normalizeTables(tables schema.Tables, pgTables schema.Tables) schema.Tables {
 	var result schema.Tables
 	for _, table := range tables {
-		pgTabe := pgTables.Get(table.Name)
-		if pgTabe == nil {
+		pgTable := pgTables.Get(table.Name)
+		if pgTable == nil {
 			result = append(result, table)
 		} else {
-			result = append(result, c.normalizeTable(table, pgTabe))
+			result = append(result, c.normalizeTable(table, pgTable))
 		}
 	}
 	return result
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -125,8 +125,8 @@ func (c *Client) listPgTables(ctx context.Context, pluginTables schema.Tables) (
 		}
 		schemaType := c.PgToSchemaType(columnType)
 		table.Columns = append(table.Columns, schema.Column{
-			Name: columnName,
-			Type: schemaType,
+			Name:       columnName,
+			Type:       schemaType,
 			PrimaryKey: isPrimaryKey,
 			NotNull:    notNull,
 		})
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -127,10 +127,8 @@ func (c *Client) listPgTables(ctx context.Context, pluginTables schema.Tables) (
 		table.Columns = append(table.Columns, schema.Column{
 			Name: columnName,
 			Type: schemaType,
-			CreationOptions: schema.ColumnCreationOptions{
-				PrimaryKey: isPrimaryKey,
-				NotNull:    notNull,
-			},
+			PrimaryKey: isPrimaryKey,
+			NotNull:    notNull,
 		})
 	}
 	return tables, nil
@@ -142,10 +140,10 @@ func (c *Client) normalizeTable(table *schema.Table, pgTable *schema.Table) *sch
 	}
 	for _, f := range table.Columns {
 		col := f
-		if c.enabledPks() && f.CreationOptions.PrimaryKey {
-			col.CreationOptions.NotNull = true
+		if c.enabledPks() && f.PrimaryKey {
+			col.NotNull = true
 		} else {
-			col.CreationOptions.PrimaryKey = false
+			col.PrimaryKey = false
 		}
 		col.Type = c.PgToSchemaType(c.SchemaTypeToPg(col.Type))
 		normalizedTable.Columns = append(normalizedTable.Columns, col)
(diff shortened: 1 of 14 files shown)


plugins/destination/postgresql/client/transformer.go
```diff
@@ -80,6 +85,26 @@ func transformArr(arr arrow.Array) []any {
...
+		case *array.Uint32:
+			pgArr[i] = pgtype.Int4{
+				Int32: int32(a.Value(i)),
+				Valid: a.IsValid(i),
+			}
+		case *array.Uint64:
+			pgArr[i] = pgtype.Int8{
+				Int64: int64(a.Value(i)),
+				Valid: a.IsValid(i),
+			}
```
hermanschaaf, line 107: I think here we should try and avoid loss of precision as far as possible: 

 - Uint8 -> cast to int16, store in int2
 - Uint16 -> cast to int32, store in int4
 - Uint32 -> cast to int64, store in int8
 - Uint64 -> store as NUMERIC maybe? 

This way we're doing the best we can within the restrictions imposed by Postgres, and can guarantee no data loss will occur

candiduslynx: I'm here to remind that we had the same issue in MSSQL migration to Apache Arrow and decided to use upscaling where possible (uin8 -> int16, etc)

hermanschaaf: Yeah, I think we should upscale here, like we did in MSSQL. This way queries will work as expected and there won't be gotchas for users where numbers they expect to always be positive (color values, enums, etc) can be negative if they don't cast it correctly. The cost is using a few extra bytes in the database, but I think 99.9% of users will be fine with that trade-off.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -115,67 +111,53 @@ func (c *Client) listPgTables(ctx context.Context, pluginTables schema.Schemas)
...
-		if c.enabledPks() && schema.IsPk(f) {
-			metadata[schema.MetadataPrimaryKey] = schema.MetadataTrue
-			f.Nullable = false
+func (c *Client) normalizeTable(table *schema.Table, pgTable *schema.Table) *schema.Table {
+	normalizedTable := schema.Table{
+		Name: table.Name,
+	}
+	for _, f := range table.Columns {
+		col := f
+		if c.enabledPks() && f.PrimaryKey {
```
hermanschaaf, line 143:
```suggestion
	for _, col := range table.Columns {
		if c.enabledPks() && col.PrimaryKey {
```

plugins/destination/postgresql/client/transformer.go
hermanschaaf: I seee we're using `numeric` in other places now, but the cast to `int64` should then be removed here

hermanschaaf (commented):
Looks good! Few final comments about uint handling

plugins/destination/postgresql/client/read_cockroach.go
```diff
@@ -0,0 +1,133 @@
...
+			b.Append(arrow.Timestamp(val.(time.Time).UnixNano()))
+		default:
+			return fmt.Errorf("unsupported timestamp unit %s", f.Type.(*arrow.TimestampType).Unit)
+		}
+	case *types.UUIDBuilder:
+		va, ok := val.([16]byte)
+		if !ok {
+			return fmt.Errorf("unsupported type %T with builder %T", val, bldr)
+		}
+		u, err := uuid.FromBytes(va[:])
```
candiduslynx, line 79: is it required? the `va` is already of the needed length, couldn't you just use `uuid.UUID(va)` instead?

plugins/destination/postgresql/client/read.go
```diff
@@ -113,24 +141,31 @@ func (c *Client) reverseTransform(f arrow.Field, bldr array.Builder, val any) er
...
 }
 
-func (c *Client) Read(ctx context.Context, table *arrow.Schema, sourceName string, res chan<- arrow.Record) error {
-	colNames := make([]string, 0, len(table.Fields()))
-	for _, col := range table.Fields() {
+func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- arrow.Record) error {
+	colNames := make([]string, 0, len(table.Columns))
+	for _, col := range table.Columns {
 		colNames = append(colNames, pgx.Identifier{col.Name}.Sanitize())
 	}
```
candiduslynx, line 166:
```suggestion
	colNames := make([]string, len(table.Columns))
	for i, col := range table.Columns {
		colNames[i] = pgx.Identifier{col.Name}.Sanitize()
	}
```

plugins/destination/postgresql/client/read.go
```diff
@@ -60,7 +68,18 @@ func (c *Client) reverseTransform(f arrow.Field, bldr array.Builder, val any) er
 	case *array.BinaryBuilder:
 		b.Append(val.([]byte))
 	case *array.TimestampBuilder:
-		b.Append(arrow.Timestamp(val.(time.Time).UnixMicro()))
+		switch b.Type().(*arrow.TimestampType).Unit {
```
candiduslynx, line 71: same as in snowflake PR (use single func instead of own impl)

plugins/destination/postgresql/client/migrate.go
```diff
@@ -212,34 +193,38 @@ func (*Client) canAutoMigrate(changes []schema.FieldChange) bool {
...
+	var result schema.Tables
 	for _, table := range tables {
-		pgTabe := pgTables.SchemaByName(schema.TableName(table))
-		result = append(result, c.normalizeTable(table, pgTabe))
+		pgTabe := pgTables.Get(table.Name)
+		if pgTabe == nil {
+			result = append(result, table)
+		} else {
+			result = append(result, c.normalizeTable(table, pgTabe))
+		}
```
candiduslynx, line 204:
```suggestion
		pgTable := pgTables.Get(table.Name)
		if pgTable == nil {
			result = append(result, table)
		} else {
			result = append(result, c.normalizeTable(table, pgTable))
		}
```

candiduslynx (changes requested):

plugins/destination/postgresql/client/types_cockroach.go
```diff
@@ -58,10 +56,14 @@ func (c *Client) CockroachToSchemaType(t string) arrow.DataType {
 	switch t {
 	case "boolean":
 		return arrow.FixedWidthTypes.Boolean
-	case "bigint", "int", "oid", "serial":
+	case "smallint":
+		return arrow.PrimitiveTypes.Int16
+	case "int4", "bigint", "int", "oid", "serial", "integer", "int8", "int64":
```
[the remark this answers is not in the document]
yevgenypats: no in cockroach for I used `int64` seems it had some issues with int32

plugins/destination/postgresql/client/read.go
yevgenypats: I think it's identitcal

candiduslynx: Unfortunately, append isn't identical to plain assignment by index, as it branches to verify the current length.
It can be somewhat summarized to the following piece of code:
```go
func append[A any](dst []A, elems...A) []A {
	l, c := len(dst), cap(dst)

	if l + len(elems) > c {
		// realloc dst, with capacity = 2*c
	}

	res = dst[0:l+len(elems)]
	for i := 0; i < len(elems); i++ {
		res[l+i] = elems[i]
	}
	return res
}
```

So, when you have an option to use the assignment by index, it's better to stick to it performance-wise.

candiduslynx: However, in this case it's a nit

candiduslynx (changes requested):
I'd like to address this [comment](https://github.com/cloudquery/cloudquery/pull/10783#discussion_r1196227641) before merging, the proposed solution is the same as [here](https://github.com/cloudquery/cloudquery/pull/10822#discussion_r1195733431)

PR #10794, opened 2023-05-15
feat(aws): Only instantiate a single SDK per Account

bbernays:
Closes https://github.com/cloudquery/cloudquery/issues/10452


plugins/source/aws/client/multiplexers.go
```diff
@@ -12,15 +10,17 @@ var AllNamespaces = []string{ // this is only used in applicationautoscaling
 }
 
 // Extract region from service list
-func getRegion(regionalMap map[string]*Services) string {
-	if len(regionalMap) == 0 {
+func getRegion(regionalList []string) string {
+	// We should try and find the closest region if possible. This will require checking the following locations:
```
hermanschaaf, line 14: I don't quite understand why the comment is so long but the function is so short; what am I missing?

bbernays: Sorry, that should have a `TODO` before of it. Didn't want to add all of that new logic in this PR

plugins/source/aws/client/client.go
```diff
@@ -64,53 +64,33 @@ var errUnknownRegion = func(region string) error {
...
 
-func (s *ServicesManager) InitServicesForPartitionAccountAndScope(partition, accountId string, svcs Services) {
-	if s.wafScopeServices == nil {
-		s.wafScopeServices = make(map[string]map[string]*Services)
-	}
-	if s.wafScopeServices[partition] == nil {
-		s.wafScopeServices[partition] = make(map[string]*Services)
-	}
-	s.wafScopeServices[partition][accountId] = &svcs
+	s.services[partition][accountId].Regions = funk.IntersectString(s.services[partition][accountId].Regions, svcs.Regions)
```
hermanschaaf, line 86: I might be missing something here, but I think this can be simplified to:

```
s.services[partition][accountId] = &svcs
```

(so no check for nil, and no `funk.IntersecString`). Otherwise, I don't understand how we can land in a situation where `InitServicesForPartitionAccount` is being called for a particular account, but has already been called for that same account before? And if we land in that situation somehow, is using the intersection of regions the right approach?

hermanschaaf: Yeah, not considering `wafScopeServices`, just looking at the old definition of `InitServicesForPartitionAccountAndRegion`. Before, it was:

```
	if s.services[partition][accountId] == nil {
		s.services[partition][accountId] = make(map[string]*Services)
	}
	s.services[partition][accountId][region] = &svcs
```

But I think that nil check was to do map initialization, so we can then safely assign to the map afterwards. This function also got called once per region, so that makes sense.

However, now we have `InitServicesForPartitionAccount` which I believe should only get called once, and we don't need to do a nil check before assigning. Or am I missing something here?

bbernays: `InitServicesForPartitionAccount` Gets called for every single account in the list of Accounts specified by the user in the spec (or discovered via Orgs)... Accounts are only deduped by this map structure because the accountID/partitions are resolved in parallel in concurrent go routines.

plugins/source/aws/client/multiplexers.go
```diff
@@ -12,15 +10,17 @@ var AllNamespaces = []string{ // this is only used in applicationautoscaling
...
+	// 3. Region defined by the AWS_DEFAULT_REGION environment variable
+	// 4. Region defined by the local config file
+	if len(regionalList) == 0 {
 		return ""
 	}
-	regions := make([]string, 0)
-	for i := range regionalMap {
-		regions = append(regions, i)
-	}
-	sort.Strings(regions)
```
hermanschaaf, line 23: What do you think about keeping the `sort.Strings(region)` in here to maintain consistency with the previous implementation? IIRC, we had it here to prevent some intermittent bugs caused by Go map ordering, but removing it now will change the order of regions again, so maybe some users will run into issues with the new ordering

hermanschaaf (approved):

plugins/source/aws/client/account.go
```diff
@@ -78,30 +77,14 @@ func (c *Client) setupAWSAccount(ctx context.Context, logger zerolog.Logger, aws
...
-	}
-
-	svcsDetails[len(account.Regions)] = svcsDetail{
+	svcsDetails := svcsDetail{
 		partition: iamArn.Partition,
 		accountId: *output.Account,
-		svcs:      initServices(cloudfrontRegion, awsCfg),
+		svcs:      initServices(awsCfg, account.Regions),
 	}
-	return svcsDetails, nil
```
bbernays, line 104: This is not needed anymore because we add the appropriate client in the multiplexer without having to have the region in the list

yevgenypats (approved):

PR #10797, opened 2023-05-15
feat(aws)!: Migrate to arrow native SDK

yevgenypats:
BEGIN_COMMIT_OVERRIDE

You will also need to update destinations depending on which one you use:
Azure Blob Storage >= v3.2.0
BigQuery >= v3.0.0
ClickHouse >= v3.1.1
DuckDB >= v1.1.6
Elasticsearch >= v2.0.0
File >= v3.2.0
Firehose >= v2.0.2
GCS >= v3.2.0
Gremlin >= v2.1.10
Kafka >= v3.0.1
Meilisearch >= v2.0.1
MSSQL >= v4.2.0
MongoDB >= v2.0.1
MySQL >= v2.0.2
Neo4j >= v3.0.0
PostgreSQL >= v4.2.0
S3 >= v4.4.0
Snowflake >= v2.1.1
SQLite >= v2.2.0

END_COMMIT_OVERRIDE


plugins/source/aws/go.mod
```diff
@@ -133,9 +134,10 @@ require (
 // TODO: remove once all updates are merged
 replace github.com/apache/arrow/go/v13 => github.com/cloudquery/arrow/go/v13 v13.0.0-20230509053643-898a79b1d3c8
 
+replace github.com/cloudquery/plugin-sdk/v3 => ../../../../plugin-sdk
```
bbernays, line 137: This needs to be updated

bbernays: I guess we can only do it once https://github.com/cloudquery/plugin-sdk/pull/864 is merged/released

hermanschaaf (approved):
🚀

PR #10821, opened 2023-05-16
feat(gremlin): Upgrade to SDK v3

disq:


plugins/destination/gremlin/client/transformer.go
```diff
@@ -39,11 +39,15 @@ func transformArr(arr arrow.Array) []any {
 		case *array.LargeString:
 			dbArr[i] = stripNulls(a.Value(i))
 		case *array.Timestamp:
-			dbArr[i] = a.Value(i).ToTime(a.DataType().(*arrow.TimestampType).Unit).UTC()
+			if isCQTime {
+				dbArr[i] = a.Value(i).ToTime(a.DataType().(*arrow.TimestampType).Unit).UTC()
+				continue
+			}
+			dbArr[i] = a.Value(i).ToTime(a.DataType().(*arrow.TimestampType).Unit).UTC().Format("2006-01-02 15:04:05.999999999")
```
disq, line 46: Gremlin time precision doesn't seem to allow `us` and `ns` tests so we stringify every timestamp except for `cq_time`. This is because if we do that to cq_time then `DeleteStale` as problems with the `gremlingo.P.Lt` operator when used on strings.

yevgenypats (approved):

PR #10831, opened 2023-05-17
feat(hackernews)!: Migrate to SDKv3

brucehsu:
#### Summary

This PR addresses #10749. 

- Migrate dependencies to plugin-sdk/v3
- Deprecate `ColumnCreationOptions` in favour of inline fields
- Migrate to Arrow types (`TypeInt`->`Int64`, `TypeTimestamp`->`Timestamp_us`)

BEGIN_COMMIT_OVERRIDE
feat: Update to use [Apache Arrow](https://arrow.apache.org/) type system (#10831)

- BigQuery >= v3.0.0
- ClickHouse >= v3.1.1
- DuckDB >= v1.1.6
- Elasticsearch >= v2.0.0
- File >= v3.2.0
- Firehose >= v2.0.2
- GCS >= v3.2.0
- Gremlin >= v2.1.10
- Kafka >= v3.0.1
- Meilisearch >= v2.0.1
- MongoDB >= v2.0.1
- MySQL >= v2.0.2
- Neo4j >= v3.0.0
- PostgreSQL >= v4.2.0
- S3 >= v4.4.0
- Snowflake >= v2.1.1
- SQLite >= v2.2.0

END_COMMIT_OVERRIDE


brucehsu:
I see that these breaking changes are due to SDKv3 using arrow.DataType names in doc generation directly [1][2].
Is this expected? Or should these be compatible with previous versions, namely v2?

[1] https://github.com/cloudquery/plugin-sdk/blob/e890385102e2668a16e35cff75fe2ffea32f2937/plugins/source/docs.go#L231
[2] https://github.com/apache/arrow/blob/07f142df646c9070fc9c627004dfdc96914fe487/go/arrow/datatype_binary.go#L28

candiduslynx:
@brucehsu the breaking change in sources is expected when performing the upgrade to plugin-sdk/v3.
I'll mark the PR as breaking accordingly.

hermanschaaf:
Nice one, thank you @brucehsu! 

We will have to wait for all destinations to be migrated to v3 (and released) before merging this, because sources using SDK v3 will require a destination that also supports SDK v3.

The other thing is that I think we should/will try and update the "Breaking Change" detection script that commented above to accept Arrow type names as aliases of the old names, so it looks a bit less scary and we can verify that there are no real breaking changes. Not saying you need to do this for this PR, I can take a look at this soon :)

brucehsu:
@hermanschaaf Gotcha! I was about to ask how do we cope with the change in types when say writing to an existing PostgreSQL database since that'd probably also change the types of columns.

plugins/source/hackernews/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.19
 
 require (
 	github.com/cloudquery/plugin-pb-go v1.0.8
-	github.com/cloudquery/plugin-sdk/v2 v2.7.0
+	github.com/cloudquery/plugin-sdk/v3 v3.5.1
```
candiduslynx, line 7:
```suggestion
	github.com/cloudquery/plugin-sdk/v3 v3.6.4
```

candiduslynx (changes requested):

candiduslynx (approved):

hermanschaaf (approved):

PR #10832, opened 2023-05-17
feat(website): Add AWS skip tables back in

jsonpr:
Adding information about tables to skip in AWS back to configuration.  This has come up recently with customer feedback.


website/pages/docs/plugins/sources/aws/configuration.md
```diff
@@ -234,3 +236,38 @@ This is used to specify one or more accounts to extract information from. Note t
...
+
+AWS has tables that may contain many resources, nested information, and AWS-provided data.  These tables may cause certain syncs to be slow due to the amount of AWS-provided data and may not be needed.  Below is a reference configuration of skip tables, where certain tables are skipped.  We recommend validating if the data contained in the tables below are necessary and to ensure syncing from required tables.
+
+```yaml
+kind: source
+spec:
+  # Source spec section
+  name: aws
+  path: cloudquery/aws
+  version: "v17.2.0"
```
hermanschaaf, line 252:
```suggestion
  version: "VERSION_SOURCE_AWS"
```

website/pages/docs/plugins/sources/aws/configuration.md
```diff
@@ -234,3 +236,38 @@ This is used to specify one or more accounts to extract information from. Note t
...
+  tables: ["*"]
```
hermanschaaf, line 253: I think listing the typically slow tables is fine, but we should also recommend against the practice of using `*`, because it can cause syncs to slow down over time as new tables are added.

yevgenypats: Yeah agree. I think the recommendation should be just specifying the needed tables (we can still show this example but recommend *strongly* against using `*`).

jsonpr: Yea makes sense.  I went back and forth on whether to recommend against using * but will add that in.

website/pages/docs/plugins/sources/aws/configuration.md
```diff
@@ -234,3 +236,38 @@ This is used to specify one or more accounts to extract information from. Note t
...
+kind: source
+spec:
+  # Source spec section
+  name: aws
+  path: cloudquery/aws
+  version: "VERSION_SOURCE_AWS"
+  tables: ["*"]
+  skip_tables:
+    - aws_ec2_vpc_endpoint_services 
+    - aws_cloudtrail_events
```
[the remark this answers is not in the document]
jsonpr: Added all 4 last_accessed_details tables for user clarity.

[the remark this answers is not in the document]
bbernays: It might be good to show that `skip_tables` supports wildcards?

bbernays (approved):

website/pages/docs/plugins/sources/aws/configuration.md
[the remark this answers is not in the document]
jsonpr: Good point, I think that should stay at the higher level configuration page and it will benefit users more here to have the skipped tables explicitly listed.

PR #10849, opened 2023-05-18
feat(neo4j)!: Update to SDK V3 with native arrow

yevgenypats:


plugins/destination/neo4j/client/write.go
```diff
@@ -5,32 +5,32 @@ import (
...
 	for _, record := range records {
 		rows = append(rows, transformValues(record)...)
 	}
 	var sb strings.Builder
 	sb.WriteString("UNWIND $rows AS row MERGE (t:")
-	sb.WriteString(tableName)
+	sb.WriteString(table.Name)
 	sb.WriteString(" {")
-	pks := schema.PrimaryKeyIndices(table)
+	pks := table.PrimaryKeysIndexes()
```
disq, line 24: Since it's all column names and nothing to do with indexes, can do it simpler: https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/gremlin/client/write.go#L35-L41

plugins/destination/neo4j/client/write.go
```diff
@@ -5,32 +5,32 @@ import (
...
 	if len(pks) == 0 {
 		// If no primary keys are defined, use _cq_id
-		pks = table.FieldIndices(schema.CqIDColumn.Name)
+		pks = []int{table.Columns.Index(schema.CqIDColumn.Name)}
```
candiduslynx, line 27: what happens if no such column is present (I think the error should be returned)

PR #10867, opened 2023-05-18
feat(mysql): Migrate to SDK V3 native arrow

bbernays:


yevgenypats:
Seems tests are failing.

bbernays:
> Seems tests are failing.

Thanks... I had been testing against MySQL 8, while these tests run against MySQL 5.7... I think we should probably start running the tests against MySQL 8 in addition to MySQL 5.7 because 5.7 will be EOL in October and I would imagine most people are now using 8 rather than 5.7

yevgenypats (approved):

PR #10871, opened 2023-05-19
feat(diff): Properly handle change from CQTypes to Apache Arrow ones

candiduslynx:
Inspired by the diff in #10831


hermanschaaf:
Nice, let's just hang back with this a little please, I think we're going to need to make some changes to the Arrow doc format first and I'm not sure how that's going to look yet.

candiduslynx:
@hermanschaaf I've updates the SDK to v3.6.3, that includes https://github.com/cloudquery/plugin-sdk/pull/896.
I presume now this code is OK to be merged and leaves us with a betted diff for src migration

hermanschaaf (approved):

PR #10874, opened 2023-05-21
feat(duckdb)!: Migrate to SDK V3

yevgenypats:
Few things in this PR:

1) Upgrade to SDK V3
2) Move from JSON to Parquet (extracting and importing from duckdb). Means that should be (in theory faster) as it all stays columnar. Also, was able to make zero-copy transformations and convert only types that needs to be transformed.

Ready for initial review but Im still working on adding more types and also solving a bug in duckdb driver which results inconsistency and fails overwrite mode.

I updated to duckdb v0.8.0

One note. There seem to be some sort of race condition between [delete](https://github.com/cloudquery/cloudquery/pull/10874/files#diff-4dd201c82c24eacf7914e3fd973841d115cf9e830d3675084b5b37ec15ecebabR155) and insert. Im not able to reproduce it in the new `v0.8.0` duckdb version but still it's not clear why at all it happens. The race is basically that when we execute insert data is still in the table and insert fails.

BEGIN_COMMIT_OVERRIDE
feat(duckdb): Migrate to SDK V3 (#10874)
BREAKING-CHANGE: Move DuckDB write to Parquet (#10874)
END_COMMIT_OVERRIDE


plugins/destination/duckdb/client/client_test.go
```diff
@@ -28,5 +33,15 @@ func TestPlugin(t *testing.T) {
 		destination.PluginTestSuiteTests{
 			MigrateStrategyOverwrite: migrateStrategy,
 			MigrateStrategyAppend:    migrateStrategy,
-		})
+			SkipOverwrite:            true,
+		},
+		// not supported in Parquet Writer
+		destination.WithTestSourceSkipIntervals(),
+		// not supported in Parquet Writer
```
disq, line 40:
```suggestion
```

yevgenypats: why remove? those two types are not supported in parquet writer for now?

disq: it's not really obvious why we have two of the same comment in succession. you should also separate the other types (times and so on) with a newline as well.

plugins/destination/duckdb/client/write.go
```diff
@@ -87,63 +89,77 @@ func (c *Client) copy_from_file(tableName string, fileName string, sc *arrow.Sch
...
+		parquet.WithMaxRowGroupLength(128*1024*1024), // 128M
+	)
+	arrprops := pqarrow.NewArrowWriterProperties(
+		pqarrow.WithStoreSchema(),
+	)
+	fw, err := pqarrow.NewFileWriter(sc, f, props, arrprops)
+	if err != nil {
+		return err
+	}
+	// defer fw.Close()
```
hermanschaaf, line 124: Should this still be commented out?

plugins/destination/duckdb/client/write.go
```diff
@@ -27,30 +31,29 @@ func nonPkIndices(sc *arrow.Schema) []int {
 // As a workaround, we delete the row and insert it again. This makes it non-atomic, unfortunately,
 // but this is unavoidable until support is added to duckdb itself.
 // See https://github.com/duckdb/duckdb/blob/c5d9afb97bbf0be12216f3b89ae3131afbbc3156/src/storage/table/list_column_data.cpp#L243-L251
-func containsList(sc *arrow.Schema) bool {
-	for _, f := range sc.Fields() {
+func containsList(sc *schema.Table) bool {
```
candiduslynx, line 34: we need to check for other list types as well (large list, fixed size list).
`arrow.IsListLike` (+ possibly check that the type ID isn't `arrow.MAP`, if we have special handling for those)

plugins/destination/duckdb/client/write.go
candiduslynx: +1

plugins/destination/duckdb/client/migrate.go
```diff
@@ -29,118 +28,95 @@ type tableInfo struct {
...
-				} else {
-					metadata[schema.MetadataPrimaryKey] = schema.MetadataFalse
-				}
-				if !schema.IsUnique(fields[i]) {
-					metadata[schema.MetadataUnique] = schema.MetadataFalse
-				}
+				normalizedColumn.PrimaryKey = false
+				normalizedColumn.Unique = false
+			} else if normalizedColumn.PrimaryKey {
+				normalizedColumn.NotNull = true
```
candiduslynx, line 73:
```suggestion
			} else {
				normalizedColumn.NotNull = normalizedColumn.NotNull || normalizedColumn.PrimaryKey
```

plugins/destination/duckdb/client/read.go
```diff
@@ -1,32 +1,47 @@
...
+func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- arrow.Record) error {
+	// sc := table.ToArrowSchema()
+	f, err := os.CreateTemp("", fmt.Sprintf("%s-*.parquet", table.Name))
 	if err != nil {
 		return err
 	}
 
-	defer os.Remove(f.Name())
-
+	// defer os.Remove(f.Name())
```
candiduslynx, line 27: should it be returned?

plugins/destination/duckdb/client/read.go
```diff
@@ -1,32 +1,47 @@
...
+	"github.com/cloudquery/plugin-sdk/v3/schema"
+	"github.com/cloudquery/plugin-sdk/v3/types"
+	"github.com/google/uuid"
 )
 
-func (c *Client) Read(ctx context.Context, sc *arrow.Schema, sourceName string, res chan<- arrow.Record) error {
-	tableName := schema.TableName(sc)
-	f, err := os.CreateTemp("", fmt.Sprintf("%s-*.json", tableName))
+func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- arrow.Record) error {
+	// sc := table.ToArrowSchema()
```
candiduslynx, line 21: it's the same as one below, rm unused or uncomment

plugins/destination/duckdb/client/write.go
```diff
@@ -154,3 +170,74 @@ func (c *Client) WriteTableBatch(ctx context.Context, sc *arrow.Schema, records
...
+}
+
+func transformArray(arr arrow.Array) arrow.Array {
+	dt := arr.DataType()
+	switch {
+	case arrow.TypeEqual(dt, types.ExtensionTypes.UUID) ||
+		arrow.TypeEqual(dt, types.ExtensionTypes.Inet) ||
+		arrow.TypeEqual(dt, types.ExtensionTypes.MAC) ||
+		arrow.TypeEqual(dt, types.ExtensionTypes.JSON) ||
+		dt.ID() == arrow.STRUCT:
```
candiduslynx, line 213:
```suggestion
	case arrow.TypeEqual(dt, types.ExtensionTypes.UUID),
		arrow.TypeEqual(dt, types.ExtensionTypes.Inet),
		arrow.TypeEqual(dt, types.ExtensionTypes.MAC),
		arrow.TypeEqual(dt, types.ExtensionTypes.JSON),
		dt.ID() == arrow.STRUCT:
```

plugins/destination/duckdb/client/write.go
```diff
@@ -154,3 +170,74 @@ func (c *Client) WriteTableBatch(ctx context.Context, sc *arrow.Schema, records
...
+}
+
+func transformToStringArray(arr arrow.Array) arrow.Array {
+	bldr := array.NewStringBuilder(memory.DefaultAllocator)
+	for i := 0; i < arr.Len(); i++ {
+		if arr.IsValid(i) {
+			bldr.Append(arr.ValueStr(i))
+		} else {
+			bldr.AppendNull()
+		}
```
candiduslynx, line 231:
```suggestion
		if arr.IsNull(i) {
			bldr.AppendNull()
			continue
		}
		bldr.Append(arr.ValueStr(i))
```

candiduslynx (changes requested):
1st round

plugins/destination/duckdb/client/write.go
yevgenypats: IsValid is better then IsNull afk as it checks for both.

plugins/destination/duckdb/client/write.go
yevgenypats: What's the different here? why it should be that one or the otherway? either way I think it's either in lint or not.

plugins/destination/duckdb/client/migrate.go
yevgenypats: What's the different? does it produce a different logic ?

yevgenypats:
> 1st round

Thanks for the review! I'll address the bugs - let's avoid any syntax comments (those goes into linter if they are important and if not then there will be just few styles of code in the codebase) as it makes it hard to focus on the important bits

plugins/destination/duckdb/client/migrate.go
candiduslynx: just eliminates branching here

plugins/destination/duckdb/client/write.go
yevgenypats: Right but in your suggestion I'll be adding something while it contains an invalid thing so in our conversion if an array has either a null or an invalid thing we will add a null (because there is no such thing as an invalid value in a database - at least in duckdb). this is why both `IsValid` -> maps to Null

candiduslynx: OK, 1 more time: `IsValid(i) ~ !IsNull(i)`:
* if `IsValid(i)` panics, `IsNull(i)` also panics and vice versa
* `IsValid(i) == true` -> `IsNull(i) == false`
* `IsValid(i) == false` -> `IsNull(i) == true`

So, basically, the code you have:
```go
if ok(i) {
	bldr.Append(arr.ValueStr(i))
} else {
	bldr.AppendNull()
}
```

And what I'm suggesting (early loop exit for less ident):
```go
if !ok(i) {
	bldr.AppendNull()
	continue
}
bldr.Append(arr.ValueStr(i))
```

candiduslynx: resolving, as it might be considered a nit

candiduslynx (approved):

candiduslynx (approved):

hermanschaaf (approved):
Nice! 💯

PR #10880, opened 2023-05-22
feat(okta): Add rate limiter config

### LINKED ISSUE #10570: bug: Okta source plugin rate limiting

### Current Behavior

The Okta source plugin gets rate limited, and loses some data.

The logs have messages like this:

```
{
    "level": "error",
    "module": "okta-src",
    "client": "okta",
    "error": "too many requests",
    "message": "table resolver finished with error",
    "table": "okta_application_group_assignments",
    "time": "2023-05-05T14:12:50Z"
}
```

### Expected Behavior

If the service (okta in this case) rate limits CQ< it should backoff and retry, and not complete until all the data is gathered (or retries are exhausted).

```
kind: source
spec:
  # Source spec section
  name: okta
  path: cloudquery/okta
  version: "v2.2.4"
  tables: ["*"]
  destinations: ["s3"]
  spec:
    # Required. Your Okta domain name
    domain: "https://${OKTA_DOMAIN}/"
    # Optional. Okta Token to access API, you can set this with OKTA_API_TOKEN environment variable
    # ⚠️ Warning - Your token should be kept secret and not committed to source control
    token: $${OKTA_TOKEN}
---
kind: destination
spec:
  name: "s3"
  path: "cloudquery/s3"
  version: "v3.1.2"
  write_mode: "append" # s3 only supports 'append' mode
  # batch_size: 10000 # optional
  # batch_size_bytes: 5242880 # optional
  spec:
    bucket: "${CQ_S3_BUCKET}"
    region: "${AWS_REGION}" # Example: us-east-1
    path: "cloudquery/{{TABLE}}/{{YEAR}}/{{MONTH}}/{{DAY}}/{{UUID}}.json"
    format: "json"
    athena: false # <- set this to true for Athena compatibility
```

### Steps To Reproduce

1. Run `cloudquery sync`
2. Observe errors in the log
3. Examine output table, find some missing rows (I noticed them in the join tables, eg `okta_group_users`, `okta_application_group_assignments`, etc)

```
{
    "level": "error",
    "module": "okta-src",
    "client": "okta",
    "error": "too many requests",
    "message": "table resolver finished with error",
    "table": "okta_application_group_assignments",
    "time": "2023-05-05T14:12:50Z"
}
```

### CloudQuery version

2.5.3

### Additional Context

Running in fargate, using `ghcr.io/cloudquery/cloudquery:2.5`

### Pull request (optional)

candiduslynx:
Closes #10570

### DIFF
--- a/plugins/source/okta/client/spec.go
+++ b/plugins/source/okta/client/spec.go
@@ -1,6 +1,60 @@
 package client
 
-type Spec struct {
-	Token  string `json:"token,omitempty"`
-	Domain string `json:"domain"`
+import (
+	"fmt"
+	"os"
+	"time"
+)
+
+type (
+	Spec struct {
+		Token     string     `json:"token,omitempty"`
+		Domain    string     `json:"domain,omitempty"`
+		RateLimit *RateLimit `json:"rate_limit,omitempty"`
+	}
+	RateLimit struct {
+		MaxBackoff time.Duration `json:"max_backoff,omitempty"`
+		MaxRetries int32         `json:"max_retries,omitempty"`
+	}
+)
+
+const (
+	OktaAPIToken = "OKTA_API_TOKEN"
+)
+
+func (s *Spec) setDefaults() {
+	const (
+		minRetries = int32(3)
+		minBackOff = 5 * time.Second
+	)
+
+	if s.RateLimit == nil {
+		s.RateLimit = new(RateLimit)
+	}
+
+	if s.RateLimit.MaxRetries < minRetries {
+		s.RateLimit.MaxRetries = minRetries
+	}
+
+	if s.RateLimit.MaxBackoff < minBackOff {
+		s.RateLimit.MaxBackoff = minBackOff
+	}
+
+	if len(s.Token) == 0 {
+		s.Token = os.Getenv(OktaAPIToken)
+	}
+}
+
+func (s *Spec) validate() error {
+	if len(s.Token) == 0 {
+		return fmt.Errorf("missing API token (should be set in the configuration or as %q environment variable)", OktaAPIToken)
+	}
+
+	const exampleDomain = "https://<CHANGE_THIS_TO_YOUR_OKTA_DOMAIN>.okta.com"
+	switch s.Domain {
+	case "", exampleDomain:
+		return fmt.Errorf("missing \"domain\" in plugin configuration")
+	}
+
+	return nil
 }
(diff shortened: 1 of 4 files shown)


plugins/source/okta/client/transformers.go
```diff
@@ -27,8 +30,25 @@ func typeTransformer(field reflect.StructField) (schema.ValueType, error) {
...
 func resolverTransformer(field reflect.StructField, path string) schema.ColumnResolver {
 	if field.Type == reflect.TypeOf(okta.NullableTime{}) {
-		return ResolveNullableTime(path)
+		return resolveNullableTime(path)
 	}
 
 	return transformers.DefaultResolverTransformer(field, path)
 }
+
+func resolveNullableTime(path string) schema.ColumnResolver {
```
candiduslynx, line 39: just moved from `client.go` and unexported

plugins/source/okta/client/spec.go
```diff
@@ -1,6 +1,60 @@
...
+		s.Token = os.Getenv(OktaAPIToken)
+	}
+}
+
+func (s *Spec) validate() error {
+	if len(s.Token) == 0 {
+		return fmt.Errorf("missing API token (should be set in the configuration or as %q environment variable)", OktaAPIToken)
+	}
+
+	const exampleDomain = "https://<CHANGE_THIS_TO_YOUR_OKTA_DOMAIN>.okta.com"
```
hermanschaaf, line 53: I don't think we really need to handle this specific case, we can just check for empty. But if we do handle this case, the example on the website right now is `YOUR_OKTA_DOMAIN`, not `CHANGE_THIS_TO_YOUR_OKTA_DOMAIN`

plugins/source/okta/client/spec.go
```diff
@@ -1,6 +1,60 @@
...
+)
+
+const (
+	OktaAPIToken = "OKTA_API_TOKEN"
+)
+
+func (s *Spec) setDefaults() {
+	const (
+		minRetries = int32(3)
+		minBackOff = 5 * time.Second
```
hermanschaaf, line 28: Looks like the default is 30s with 2 retries (based on a quick google finding this: https://github.com/okta/okta-sdk-golang/blob/master/okta/okta.go#L187); if users were already running into errors before we should probably keep the default backOff to 30s or so as well (the minimum can still be less)

disq (approved):

disq:
Foiled by automerge, sorry

PR #10887, opened 2023-05-22
feat(okta): Deprecation notice for `OKTA_API_TOKEN` env variable support

candiduslynx:


bbernays (changes requested):
Why would we want to break backwards compatibility for this?

candiduslynx:
> Why would we want to break backwards compatibility for this?

@bbernays 

It's a cleanup as we did in Git(Lab?Hub?) and Tailscale.

Additionally, I'd like to merge this along with the v3 migration, as that one would already be a breaking change.

bbernays:
> > Why would we want to break backwards compatibility for this?
> 
> @bbernays
> 
> It's a cleanup as we did in Git(Lab?Hub?) and Tailscale.
> 
> Additionally, I'd like to merge this along with the v3 migration, as that one would already be a breaking change.

What is the improved experience or functionality that users will get by us forcing them to adjust their configurations?


But the V3 migration shouldn't require users to adjust their configurations other than the destination version...

candiduslynx:
> What is the improved experience or functionality that users will get by us forcing them to adjust their configurations?

We streamline the code base here (no direct env lookups in plugins). It's more of the dev experience change.



> But the V3 migration shouldn't require users to adjust their configurations other than the destination version...

It will still be a breaking change per the types changes and the need to use updated destinations as well.
So, it feels like a good moment to add an additional breaking change.

hermanschaaf (dismissed):
I agree, let's not do a breaking change only for the sake of removing a few lines / consistency with other plugins.

Instead, maybe we can start by logging a warning if you use `OKTA_API_TOKEN` without setting `token` in the config, and say this is marked for deprecation and will be removed in a future version?

candiduslynx:
@hermanschaaf @bbernays 
I've changed the PR to only print out a warning about this variable + added the warning to the config example & overview.

plugins/source/okta/client/spec.go
```diff
@@ -42,6 +44,7 @@ func (s *Spec) setDefaults() {
 	}
 
 	if len(s.Token) == 0 {
+		logger.Warn().Msgf("usage of %q environment variable value is deprecated and will be dropped in the future releases", OktaAPIToken)
```
bbernays, line 47:
```suggestion
		logger.Warn().Msgf("usage of %q environment variable value is deprecated and will be dropped in a future release", OktaAPIToken)
```

bbernays (approved):

Issue #10782, opened 2023-05-14
bug: Subnets are excluded from aws_resources view


pchaganti:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The current aws_resources view SQL as posted on the CQ blog and found in GitHub - https://github.com/cloudquery/cloudquery/tree/main/plugins/source/aws/views
will not include any subnets. 

The SQL is using an INTERSECT to pull together resources with both an 'account_id' and an 'arn' column. The **aws_ec2_subnets** table does not have an account_id column. Here is the discord discussion on that - https://discord.com/channels/872925471417962546/1106918918423060500

### Expected Behavior

The resources view contains ec2 subnets.

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: aws
  path: cloudquery/aws
  version: "v17.2.0"
```

### CloudQuery (redacted) logs

```
```

### CloudQuery version

3.2.1

### Pull request (optional)

- [ ] I can submit a pull request

bbernays:
More resources than just `aws_ec2_subnets` are possibly also impacted by this issue and they include:
`aws_organizations_accounts`
`aws_inspector2_findings`
`aws_ec2_managed_prefix_lists`

Issue #10366, opened 2023-04-26
feat: Support service and type as columns in AWS plugin


hermanschaaf:
### Which problem is this feature request solving?

This is both a report of a discrepancy in a past blog and a feature request. Discord user `mingfang` [reported](https://discord.com/channels/872925471417962546/873606591335759872/1100653639179173898) that the [blog post on the AWS resources view](https://www.cloudquery.io/blog/aws-resources-view) mentions `service` and `type` columns, but this is no longer available on the current view (nor in the SQL query shown in the post). 

We should either add these columns back to the view, if possible, or if not possible then at least fix the blog post to be consistent. User mingfang also requested that these columns be added to all tables by the AWS plugin, as a performance improvement over parsing ARNs in the view, but this is something we can discuss if adding it to the view turns out not to be viable.

### Describe the solution you'd like

`service` and `type` columns available on the AWS resources view

### Pull request (optional)

- [ ] I can submit a pull request

pchaganti:
This would be really useful. Not needing to parse the ARN into corresponding tokens on the client side for large accounts can help performance.

PR #10896, opened 2023-05-23
chore(website): Fix AWS resources view ARN fields

disq:
Parse ARN according to the [spec](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html) and add `partition`, `service` and `type` fields.

Closes https://github.com/cloudquery/cloudquery/issues/10366
Closes https://github.com/cloudquery/cloudquery/issues/10782

Looks like https://cloudquery-lypoyt9im-cloudquery.vercel.app/blog/aws-resources-view


plugins/source/aws/views/resources.sql
```diff
@@ -14,14 +14,17 @@ FOR tbl IN
...
  	IF NOT (strSQL = ''::TEXT) THEN
 		strSQL = strSQL || ' UNION ALL ';
 	END IF;
 	-- create an SQL query to select from table and transform it into our resources view schema
-	strSQL = strSQL || FORMAT('
-        select _cq_id, _cq_source_name, _cq_sync_time, %L as _cq_table, account_id, %s as region, arn, %s as tags
+	strSQL = strSQL || FORMAT(E'
+        SELECT _cq_id, _cq_source_name, _cq_sync_time, %L AS _cq_table, account_id, %s AS region,
+        SPLIT_PART(arn, \':\', 2) AS PARTITION, SPLIT_PART(arn, \':\', 3) AS service,
+        CASE WHEN SPLIT_PART(SPLIT_PART(arn, \':\', 6), \'/\', 2) = \'\' AND SPLIT_PART(arn, \':\', 7) = \'\' THEN NULL ELSE SPLIT_PART(SPLIT_PART(arn, \':\', 6), \'/\', 1) END AS type,
```
disq, line 26:
```
arn:partition:service:region:account-id:resource-id
arn:partition:service:region:account-id:resource-type/resource-id
arn:partition:service:region:account-id:resource-type:resource-id
```

`resource-type` is only valid if there's a resource-id in the next field (separated by one of two separators) so this checks for both and sets type to `NULL` otherwise.

plugins/source/aws/views/resources.sql
```diff
@@ -14,14 +14,17 @@ FOR tbl IN
...
+        arn, %s AS tags
         FROM %s',
```
bbernays, line 28: Can we make this a little easier to read like this?
```suggestion
        SELECT _cq_id, 
        	_cq_source_name, 
        	_cq_sync_time, 
        	%L AS _cq_table, 
        	account_id, 
        	%s AS region,
        	SPLIT_PART(arn, \':\', 2) AS PARTITION, 
        	SPLIT_PART(arn, \':\', 3) AS service,
        	CASE
				WHEN SPLIT_PART(SPLIT_PART(ARN, \':\', 6), \'/\', 2) = \'\' AND SPLIT_PART(arn, \':\', 7) = \'\' THEN NULL
        		ELSE SPLIT_PART(SPLIT_PART(arn, \':\', 6), \'/\', 1)
        	END AS TYPE,
        	arn, %s AS tags
        FROM %s',
```

PR #10907, opened 2023-05-24
fix: Use NUMERIC for uint64

hermanschaaf:
`uint64` cannot be properly represented by INT64, so we should use NUMERIC.


plugins/destination/bigquery/client/types.go
```diff
@@ -14,13 +14,6 @@ type TimestampNanoseconds struct {
 	Nanoseconds int    `json:"nanoseconds" bigquery:"nanoseconds"`
 }
 
-// DurationNanoseconds is a struct to hold a duration with nanosecond precision,
-// because BigQuery does not support nanosecond precision durations.
-type DurationNanoseconds struct {
-	Duration    string `json:"duration" bigquery:"duration"`
-	Nanoseconds int    `json:"nanoseconds" bigquery:"nanoseconds"`
-}
```
hermanschaaf, line 22: This was unused

disq (approved):

PR #10917, opened 2023-05-24
feat: Add resource_record_sets to GCP plugin

paulopontesm:
#### Summary

This PR adds the `gcp_dns_resource_record_sets` table to the GCP plugin.

These are the columns for the new table: 

```csv
_cq_source_name,_cq_sync_time,_cq_id,_cq_parent_id,project_id,kind,name,routing_policy,rrdatas,signature_rrdatas,ttl,type
```

Tested this on my GCP project and seems to be working ok. Let me know if there's something I'm missing.


plugins/source/gcp/resources/plugin/tables.go
```diff
@@ -128,6 +128,7 @@ func PluginAutoGeneratedTables() []*schema.Table {
 		containeranalysis.Occurrences(),
 		dns.Policies(),
 		dns.ManagedZones(),
+		dns.ResourceRecordSets(),
```
paulopontesm, line 131: I'm not 100% sure if I should add this here also, or if the relationship with the managed zones enought. Can you confirm, please?

plugins/source/gcp/client/services.go
```diff
@@ -143,6 +143,7 @@ var GcpServices = map[string]bool{
 	"genomics.googleapis.com":                         true,
 	"gkebackup.googleapis.com":                        true,
 	"gkehub.googleapis.com":                           true,
+	"gkeonprem.googleapis.com":                        true,
```
paulopontesm, line 146: The codegen added this service. I'm pushing it here, but I didn't test anything around this. Let me know if this should be handled some other way.

plugins/source/gcp/resources/plugin/tables.go
hermanschaaf: The relationship with managed zones is enough, it shouldn't be added here as well (otherwise I would expect the sync to fail because there is no associated parent for the resolver to use)

plugins/source/gcp/resources/services/dns/resource_record_sets.go
```diff
@@ -0,0 +1,28 @@
+package dns
+
+import (
+	"github.com/cloudquery/plugin-sdk/v2/schema"
+	"github.com/cloudquery/plugin-sdk/v2/transformers"
+	"github.com/cloudquery/plugins/source/gcp/client"
+	"google.golang.org/api/dns/v1"
+)
+
+func ResourceRecordSets() *schema.Table {
```
hermanschaaf, line 10:
```suggestion
func resourceRecordSets() *schema.Table {
```

You can make this private to indicate that it's a relational table and shouldn't be used directly

plugins/source/gcp/client/services.go
hermanschaaf: It should be fine I believe, we can keep it here

plugins/source/gcp/resources/services/dns/resource_record_sets.go
```diff
@@ -0,0 +1,28 @@
...
+		Transform:   client.TransformWithStruct(&dns.ResourceRecordSet{}, transformers.WithPrimaryKeys("Name", "Type")),
+		Columns: []schema.Column{
+			{
+				Name:     "project_id",
+				Type:     schema.TypeString,
+				Resolver: client.ResolveProject,
+				CreationOptions: schema.ColumnCreationOptions{
+					PrimaryKey: true,
+				},
+			},
```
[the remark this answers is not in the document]
paulopontesm: That makes sense. It will make some queries a lot easier 🥳

paulopontesm:
@hermanschaaf I think I implemented all your suggestions 👍 

Also rebased and bumped the schema to v3.

hermanschaaf (approved):
Looks good, thanks @paulopontesm (and thanks for rebasing onto v3!) 🚀

Issue #10915, opened 2023-05-24
feat: support HubSpot associations extraction


aflopes:
### Which problem is this feature request solving?

Currently it seems that it is not possible to extract `associations` from HubSpot using the HubSpot source plugin, and this feature is to have this support.

### Describe the solution you'd like

According to the [HubSpot REST API documentation](https://developers.hubspot.com/docs/api/overview) it is possible to get associations for all the current supported objects that can be extracted into tables (companies, contacts, deals, line_items, products, quotes and tickets).

So it seems to me that it should be possible to have a plugin configuration like:
```
    table_options:
      hubspot_crm_companies:
        properties: []
        associations: []
      hubspot_crm_contacts:
        properties: []
        associations: []
      hubspot_crm_deals:
        properties: []
        associations: []
      hubspot_crm_line_items:
        properties: []
        associations: []
      hubspot_crm_products:
        properties: []
        associations: []
      hubspot_crm_quotes:
        properties: []
        associations: []
      hubspot_crm_tickets:
        properties: []
        associations: []
```

where the `associations` list would be transformed into the `associations` query parameter the REST API has available, pretty much just like what I guess that it is being done with the `properties` on plugin configuration.

### Pull request (optional)

- [ ] I can submit a pull request

yevgenypats:
> ### Which problem is this feature request solving?
> Currently it seems that it is not possible to extract `associations` from HubSpot using the HubSpot source plugin, and this feature is to have this support.
> 
> ### Describe the solution you'd like
> According to the [HubSpot REST API documentation](https://developers.hubspot.com/docs/api/overview) it is possible to get associations for all the current supported objects that can be extracted into tables (companies, contacts, deals, line_items, products, quotes and tickets).
> 
> So it seems to me that it should be possible to have a plugin configuration like:
> 
> ```
>     table_options:
>       hubspot_crm_companies:
>         properties: []
>         associations: []
>       hubspot_crm_contacts:
>         properties: []
>         associations: []
>       hubspot_crm_deals:
>         properties: []
>         associations: []
>       hubspot_crm_line_items:
>         properties: []
>         associations: []
>       hubspot_crm_products:
>         properties: []
>         associations: []
>       hubspot_crm_quotes:
>         properties: []
>         associations: []
>       hubspot_crm_tickets:
>         properties: []
>         associations: []
> ```
> 
[...]

PR #10919, opened 2023-05-24
feat(hubspot): Extract associations

aflopes:
#### Summary

Extracts the HubSpot objects associations and adds their JSON content as a column in the extracted table
Relates to https://github.com/cloudquery/cloudquery/issues/10915


hermanschaaf (approved):

aflopes:
That was fast! Thanks for accepting this so quickly!

PR #10934, opened 2023-05-25
feat(cli): Migrate support SDK V3 arrow native

yevgenypats:


candiduslynx:
@yevgenypats I've converted it to draft.
Kindly publish once the tests are passing & the code is ready

yevgenypats:
I don't think this is critical right now. we can do more fine tuning in the future like also turn the source and destination to the same plugin which will stream line the code. merging if no clear bugs spotted.

PR #10946, opened 2023-05-25
feat(snowflake): Update to SDK v3.6.3

yevgenypats:


candiduslynx:
I think we need to support `"null"` for string parsing (like, in [here](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/snowflake/client/read.go#L31))

candiduslynx (approved):

PR #10989, opened 2023-05-26
chore(ci): Allow 2 commits

candiduslynx:


hermanschaaf:
I see one here https://github.com/cloudquery/cloudquery/pull/10897

PR #11003, opened 2023-05-26
feat(plausible)!: Migrate to Arrow native SDK

### LINKED ISSUE #10762: feat: Migrate `plugin/source/plausible` to `github.com/cloudquery/plugin-sdk/v3`

Part of #10711

Upgrade the code.

### Pull request (optional)

disq:
Closes #10762

BEGIN_COMMIT_OVERRIDE
feat: Update to use [Apache Arrow](https://arrow.apache.org/) type system (#11003)

- BigQuery >= v3.0.0
- ClickHouse >= v3.1.1
- DuckDB >= v1.1.6
- Elasticsearch >= v2.0.0
- File >= v3.2.0
- Firehose >= v2.0.2
- GCS >= v3.2.0
- Gremlin >= v2.1.10
- Kafka >= v3.0.1
- Meilisearch >= v2.0.1
- MongoDB >= v2.0.1
- MySQL >= v2.0.2
- Neo4j >= v3.0.0
- PostgreSQL >= v4.2.0
- S3 >= v4.4.0
- Snowflake >= v2.1.1
- SQLite >= v2.2.0

END_COMMIT_OVERRIDE


plugins/source/plausible/resources/services/stats_timeseries.go
```diff
@@ -19,45 +20,41 @@ func StatsTimeseries() *schema.Table {
...
 			},
 			{
 				Name:        "date",
 				Description: "Date of the data point",
-				Type:        schema.TypeTimestamp,
-				CreationOptions: schema.ColumnCreationOptions{
-					PrimaryKey: true,
-				},
-				Resolver: ResolveDate,
+				Type:        arrow.FixedWidthTypes.Timestamp_s,
```
disq, line 30: This is different than our usual `Timestamp_us` approach. Initially wanted to use `Date64` but our scalar pkg doesn't support it yet.

hermanschaaf: Why can't we use `Timestamp_us`?

disq: we can, but it's a date in the end. Would've used Date64 if scalar was up to date

hermanschaaf: Even though it's a date, we don't yet have a compatibility mode, so we should try and stick to the Arrow types that we determined to be the equivalents of the old CQTypes for now (in this case `Timestamp_us`). This should guarantee that the schema doesn't break for the majority of destinations, if not all.

hermanschaaf: (More context: this is what the struct transformer is doing now https://github.com/cloudquery/plugin-sdk/blob/main/transformers/struct.go#L307C6-L350 so if we require manual column updates we should also map it in the same way)

disq: ok fixed :)

candiduslynx (approved):

hermanschaaf (approved):

PR #11022, opened 2023-05-26
chore(shopify): Update SDK to 3.6.5, remove unnecessary transformer

disq:


plugins/source/shopify/client/transformers.go
```diff
@@ -1,27 +1,10 @@
...
-	}
-}
-
-var options = []transformers.StructTransformerOption{
-	transformers.WithTypeTransformer(typeTransformer),
-}
-
 func TransformWithStruct(t any, opts ...transformers.StructTransformerOption) schema.Transform {
-	return transformers.TransformWithStruct(t, append(options, opts...)...)
+	return transformers.TransformWithStruct(t, opts...)
```
candiduslynx, line 9: if we don't have any common options, just drop `TransformWithStruct` func & replace with direct call to `transformers.TransformWithStruct`

candiduslynx (changes requested):

candiduslynx (approved):

PR #11031, opened 2023-05-26
fix(test-dst): Update to the latest SDK

candiduslynx:
Required for #11027


plugins/destination/test/go.mod
```diff
@@ -5,12 +5,12 @@ go 1.19
 require (
 	github.com/apache/arrow/go/v13 v13.0.0-20230509040948-de6c3cd2b604
 	github.com/cloudquery/plugin-pb-go v1.0.8
-	github.com/cloudquery/plugin-sdk/v3 v3.6.3
+	github.com/cloudquery/plugin-sdk/v3 v3.6.6
 	github.com/rs/zerolog v1.29.0
 )
 
 // TODO: remove once all updates are merged
-replace github.com/apache/arrow/go/v13 => github.com/cloudquery/arrow/go/v13 v13.0.0-20230509053643-898a79b1d3c8
```
hermanschaaf, line 13: Looks like we are still using the previously referenced version in the SDK https://github.com/cloudquery/plugin-sdk/blob/main/go.mod#L29 🤔

candiduslynx: It's just the latest cqmain, it'll be used instead of the one referenced in the SDK.
As we don't have diff from upstream, I think it's fine as long as the tests pass

hermanschaaf (approved):

hermanschaaf (approved):

PR #11035, opened 2023-05-26
feat(aws)!: Add Support For Organizational Unit Accounts

bbernays:
#### Summary

Users can now map Accounts back to the correct OU


plugins/source/aws/resources/services/organizations/organizational_unit_accounts.go
```diff
@@ -0,0 +1,48 @@
...
+
+func organizationalUnitAccounts() *schema.Table {
+	tableName := "aws_organizations_organizational_unit_accounts"
+	return &schema.Table{
+		Name:        tableName,
+		Description: `https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccountsForParent.html`,
+		Resolver:    fetchAccountsForParent,
+		Transform:   transformers.TransformWithStruct(&types.Account{}, transformers.WithPrimaryKeys("Arn")),
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(true),
```
bbernays, line 22: Ideally this should be `request_account_id` but that would require changing the parent table in a backwards incompatible way

erezrokah (approved):

jsonpr (commented):
I'd prefer to add a parent column to both aws_organizations_organizational_units and aws_organizations_accounts for a complete picture of the organizational structure.  With adding a parent column to both OUs and accounts, that would give a complete picture of the organizational structure all the way up to the root. 

I haven't tested this, but this implementation could work.  Does this give us an understanding of OUs within OUs?  An example would be an OU (with accounts) within an OU.

bbernays:
> I'd prefer to add a parent column to both aws_organizations_organizational_units and aws_organizations_accounts for a complete picture of the organizational structure. With adding a parent column to both OUs and accounts, that would give a complete picture of the organizational structure all the way up to the root.

I definitely agree that we should add a `parent_id` column to `aws_organizations_organizational_units` as otherwise there is no way to identify the structure of an org. We can't add the `parent_id` to `aws_organizations_accounts` because there is no API that can get that information (only way to get it is to traverse the OU structure...

With the addition of the `parent_id` on both `aws_organizations_organizational_units` and `aws_organizations_organizational_unit_accounts` the data is present to be able to generate a recursive query that can link the OUs, roots and accounts all together

jsonpr:
>We can't add the parent_id to aws_organizations_accounts because there is no API that can get that information (only way to get it is to traverse the OU structure...

Doesn't ListParents https://docs.aws.amazon.com/cli/latest/reference/organizations/list-parents.html return the immediate parent (OU or root) for the specified OU or account?  Wouldn't that give the information needed to add `parent_id` to `aws_organizations_accounts`?

bbernays:
> Doesn't ListParents https://docs.aws.amazon.com/cli/latest/reference/organizations/list-parents.html return the immediate parent (OU or root) for the specified OU or account? Wouldn't that API call give the information needed to add `parent_id` to `aws_organizations_accounts`?

`aws_organizations_accounts` uses the `ListAccounts` API to be able to get all accounts

jsoncow:
Does the implementation of `aws_organizations_organizational_unit_accounts` cover the use case of accounts with a parent of root?  

If not, how will customers determine which accounts have a parent of root?

bbernays:
> Does the implementation of `aws_organizations_organizational_unit_accounts` cover the use case of accounts with a parent of root?
> 
> If not, how will customers determine which accounts have a parent of root?

good call out. Will look into this

bbernays:
> Does the implementation of `aws_organizations_organizational_unit_accounts` cover the use case of accounts with a parent of root?
> 
> If not, how will customers determine which accounts have a parent of root?

I added a table `aws_organizations_root_accounts` as a relation to `aws_organizations_roots`. I don't love it the name because it makes it hard to understand the difference between a root account and accounts attached to the root.

- `aws_organizations_roots`
    - `aws_organizations_root_accounts`
- `aws_organizations_organizational_units`
    - `aws_organizations_organizational_unit_accounts`
- `aws_organizations_accounts`

Maybe what we want to do is just reorganize the tables so it goes:
- `aws_organizations_accounts`
- `aws_organizations_roots`
    - `aws_organizations_organizational_units`
        - `aws_organizations_organizational_unit_accounts`

jsonpr:
> > Does the implementation of `aws_organizations_organizational_unit_accounts` cover the use case of accounts with a parent of root?
> > If not, how will customers determine which accounts have a parent of root?
> 
> I added a table `aws_organizations_root_accounts` as a relation to `aws_organizations_roots`. I don't love it the name because it makes it hard to understand the difference between a root account and accounts attached to the root.
> 
> * `aws_organizations_roots`
>   
>   * `aws_organizations_root_accounts`
> * `aws_organizations_organizational_units`
>   
>   * `aws_organizations_organizational_unit_accounts`
> * `aws_organizations_accounts`
> 
> Maybe what we want to do is just reorganize the tables so it goes:
> 
> * `aws_organizations_accounts`
> * `aws_organizations_roots`
>   
>   * `aws_organizations_organizational_units`
>     
>     * `aws_organizations_organizational_unit_accounts`

Agreed that a table named `aws_organizations_root_accounts` would get confusing.  I'd also think customers would get confused with accounts existing in both `aws_organizations_root_accounts` and `aws_organizations_organizational_unit_accounts`.  

I think it also may not be ideal to have multiple levels of child tables.

Thinking back to a previous comment - 

>Doesn't ListParents https://docs.aws.amazon.com/cli/latest/reference/organizations/list-parents.html return the immediate parent (OU or root) for the specified OU or account? Wouldn't that API call give the information needed to add parent_id to aws_organizations_accounts?.

And your response: 
>aws_organizations_accounts uses the ListAccounts API to be able to get all accounts

PR #11034, opened 2023-05-26
chore(okta): Update SDK to 3.6.6 and regenerate docs

hermanschaaf:
Fixes some errors we are seeing on the website due to special characters in the table.

PR #11036, opened 2023-05-26
chore(diff): Ignore backticks for the migration

candiduslynx:
Inspired by #11034


candiduslynx:
> @candiduslynx Could you give me the summary here--why did it previously think the sort order changed when we introduced backticks?

@hermanschaaf 
Basically, the logic is: if there's no change in the column type/options, but we have diff -- it means the column was moved.

In our current behavior (we don't care about the columns order much) we might want just remove this message altogether, but I'd rather wait for @erezrokah to take such big step

hermanschaaf (approved):

Issue #10963, opened 2023-05-25
feat: Multi-Account Guide for AWS and default role


jsonpr:
### Which problem is this feature request solving?

Currently the multi-account guide by default uses `OrganizationAccountAccessRole` which comes with full administrator `*:*` permissions and customers may see this as overpermissive or dangerous.  For example, some enterprises don't use the the `OrganizationAccountAccessRole`

Would be good to keep permissions at a read/view-only access level.

### Describe the solution you'd like

For now, this should just be additional documentation and clarification.

Further conversation on what makes the most sense as a default would be needed as `OrganizationAccountAccessRole` is one of the most straightforward ways to set up CQ since it's by default created in all member accounts in an Organization but may not be the most secure.  

### Pull request (optional)

- [X] I can submit a pull request

PR #11045, opened 2023-05-27
doc(website): Update AWS Organizations Role documentation

jsonpr:
This is for https://github.com/cloudquery/cloudquery/issues/10963 to update documentation references to the OrganizationAccountAccessRole role.  

AWS creates the OrganizationAccountAccessRole role by default in created member accounts as part of an Organization.  However this role will come with a default trust policy that trusts all of the admin account (trusts the root principal) and has *:* (admin permissions).

Different organizations may not use this role, may not want a read-only tool to use this, and also may use this for alerting.  The OrganizationAccountAccessRole can also be used for privilege escalation to jump around within multiple AWS accounts.  Thus, it could raise concerns if someone installing CloudQuery wants to use the OrganizationAccountAccessRole.  This is especially true for enterprises with split responsibilities.  For example, a platform team may want to install CQ across the AWS Organization but has to work with a security team to install.  Asking for access to the ORganizationAccountAccessRole may raise alarms with the security team.

This also adds a reference to https://github.com/cloudquery/iam-for-aws-orgs as a reference for teams deploying CQ.


disq (approved):

website/pages/docs/plugins/sources/aws/multi-account.mdx
```diff
@@ -31,29 +31,31 @@ Prerequisites for using AWS Org functionality:
...
+* Reference IAM assets and the CloudFormation templates for deployment in an AWS Organization for CloudQuery can be found [here](https://github.com/cloudquery/iam-for-aws-orgs).
 
 Configuring AWS Organization:
 
 1. It is always necessary to specify a member role name:
 
    ```yaml copy
        org:
-         member_role_name: OrganizationAccountAccessRole
+         member_role_name: cloudquery-ro
```
bbernays, line 44: Maybe instead of `cloudquery-ro` we can indicate that it is a user specified value like 
```
member_role_name: <cloudquery-ro>
```

PR #11078, opened 2023-05-29
fix(deps): Update `github.com/cloudquery/plugin-sdk/v3` to `v3.8.1`

candiduslynx:
Extracted from #11043
Depends on:
* https://github.com/cloudquery/plugin-sdk/pull/913
* https://github.com/apache/arrow/pull/35823
* https://github.com/cloudquery/plugin-sdk/pull/921


plugins/destination/clickhouse/go.mod
erezrokah, line 1: Can we scope the changes in the file to only the SDK update? Is there a reason we need to update other dependencies too?

plugins/destination/clickhouse/go.mod
```diff
@@ -3,21 +3,18 @@ module github.com/cloudquery/cloudquery/plugins/destination/clickhouse
...
+	github.com/cloudquery/plugin-sdk/v3 v3.8.1
 	github.com/google/uuid v1.3.0
 	github.com/rs/zerolog v1.29.1
-	github.com/stretchr/testify v1.8.2
+	github.com/stretchr/testify v1.8.3
 	golang.org/x/sync v0.2.0
 )
 
-// TODO: remove once al updates are merged
-replace github.com/apache/arrow/go/v13 => github.com/cloudquery/arrow/go/v13 v13.0.0-20230509053643-898a79b1d3c8
```
erezrokah, line 17: Should we keep this and update the `github.com/cloudquery/arrow` reference? Other plugins still use the fork and it seem we'd always want to use our fork to avoid waiting for changes to be merged upstream.

Anyway this discussion seems out of scope for this PR

plugins/destination/clickhouse/client/client_test.go
```diff
@@ -52,5 +53,17 @@ func TestPlugin(t *testing.T) {
 
 			MigrateStrategyAppend: migrateStrategy,
 		},
+		destination.WithTestSourceAllowNull(func(dt arrow.DataType) bool {
```
erezrokah, line 56: Can you explain what's the reason we disallow null for these types but allow for others?

plugins/destination/clickhouse/typeconv/arrow/values/time.go
```diff
@@ -48,16 +70,5 @@ func timeToTimestamp(value time.Time, tsType *arrow.TimestampType) (arrow.Timest
...
-	case arrow.Millisecond:
-		return arrow.Timestamp(value.Unix()*1e3 + int64(value.Nanosecond())/1e6), nil
-	case arrow.Microsecond:
-		return arrow.Timestamp(value.Unix()*1e6 + int64(value.Nanosecond())/1e3), nil
-	case arrow.Nanosecond:
-		return arrow.Timestamp(value.UnixNano()), nil
-	default:
-		return arrow.Timestamp(0), fmt.Errorf("unsupported Apache Arrow time unit: %s", tsType.Unit.String())
-	}
+	return arrow.TimestampFromTime(value, tsType.Unit)
```
erezrokah, line 73: To confirm, this is a refactoring unrelated to the fixes needed for the upgrade?

plugins/destination/clickhouse/typeconv/ch/values/list.go
```diff
@@ -37,13 +41,33 @@ func listValue(arr array.ListLike) (any, error) {
...
-			val.Elem().Set(reflect.ValueOf(elem))
-		}
+		val.Elem().Set(reflect.ValueOf(elem))
 		res.Index(i).Set(val)
 	}
 
 	return res.Interface(), nil
 }
+
+// sanitizeNested will replace all null entries with empty ones as in CH nested types aren't nullable themselves
```
erezrokah, line 54: Thanks for adding this comment 🚀

plugins/destination/clickhouse/typeconv/ch/values/special.go
```diff
@@ -27,11 +27,11 @@ func float16Value(arr *array.Float16) []*float32 {
 	return res
 }
 
-func byteArrValue(arr primitive[[]byte]) [][]byte {
-	res := make([][]byte, arr.Len())
+func byteArrValue(arr primitive[[]byte]) []*string {
```
erezrokah, line 30: Can you explain the reason for this change? Is it related to the SDK update?

erezrokah (commented):
Thanks for the PR @candiduslynx, I did a first pass and mostly have questions on the some of the changes to ensure I understand everything correctly.

plugins/destination/clickhouse/go.mod
candiduslynx: This might go any other way, as I tend to get the review & merge upstream first & then sync forc, merge `main` into `cqmain` and bump the replace.

I think, for now it might be a place to start the migration to the upstream (and, if there's a long awaiting PR that we really need to propagate faster - we can always add the replace back (as we would still need to bump in anyways))

plugins/destination/clickhouse/client/client_test.go
candiduslynx: Yes, it's this: https://clickhouse.com/docs/en/sql-reference/data-types/nullable
Array, Map & Tuple can't be nullable (arow: list, map, struct)

plugins/destination/clickhouse/typeconv/arrow/values/string.go
```diff
@@ -10,5 +10,13 @@ func buildFromString(builder array.Builder, value any) error {
 		builder.AppendNull()
 		return nil
 	}
-	return builder.AppendValueFromString(v)
+
+	if len(v) > 0 {
+		return builder.AppendValueFromString(v)
+	}
+
+	// binary types are handled separately, so here we have a builder that most likely can't handle empty string.
```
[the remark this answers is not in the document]
candiduslynx: I didn't actually test if any of the builders that get here can handle the empty string & I can't be sure that they wont be added in the future.
Thus, this comment just explains why we opt for an empty value for the reading here

plugins/destination/clickhouse/typeconv/ch/values/list_test.go
```diff
@@ -32,20 +32,26 @@ func Test_listValue(t *testing.T) {
...
+		builder.AppendNull()
 	}
 
 	data, err := listValue(builder.NewListArray())
 	require.NoError(t, err)
 
 	uidSlices := data.([]*[]*uuid.UUID)
 
-	require.Equal(t, amount, len(uidSlices))
+	require.Equal(t, 2*amount, len(uidSlices))
```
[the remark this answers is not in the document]
candiduslynx: done in 69830d2c27cb961b4304ccb0bdff60d2ce35f785

erezrokah (approved):

PR #11084, opened 2023-05-29
fix(test): Support only implemented scalars

candiduslynx:
Required for #11027


yevgenypats:
But there can be sources that use `arrow` directly so not sure it's a good idea to ignore those in the `source` but rather on the destination until we just support all types on every destination.

candiduslynx:
> But there can be sources that use `arrow` directly so not sure it's a good idea to ignore those in the `source` but rather on the destination until we just support all types on every destination.

@yevgenypats 
This managed source shouldn't create values that will result in scalar package panic (now they do)

PR #11099, opened 2023-05-29
fix(test): Fix missing resolver

disq:
Works when used with https://github.com/cloudquery/plugin-sdk/pull/914


plugins/source/test/resources/services/testdata.go
```diff
@@ -13,7 +13,7 @@ func TestDataTable() *schema.Table {
 		if strings.HasPrefix(c.Name, "_cq_") {
 			table.Columns[i].Name = "test" + c.Name
 		}
-		table.Columns[i].PrimaryKey = false
+		table.Columns[i].PrimaryKey = c.Name == "int64"
```
disq, line 16: This was a wild moment to test something, should I remove it so that we have a table with no PKs?

candiduslynx: IMO it's good to have a table with no PKs in test source until the moment we ban them completely

plugins/source/test/resources/services/test.go
```diff
@@ -11,27 +11,29 @@ func TestSomeTable() *schema.Table {
 	return &schema.Table{
 		Name:        "test_some_table",
 		Description: "Test description",
-		Resolver:    fetchSomeTableData,
+		Resolver:    resolveSomeTable,
```
candiduslynx, line 14: Could we rename this back just for the sake of the less diff?

candiduslynx (changes requested):

plugins/source/test/resources/services/test.go
disq: done in [`714d1e9` (#11099)](https://github.com/cloudquery/cloudquery/pull/11099/commits/714d1e98df85fc04d99fb96f47c0032ce1cbc063)

candiduslynx (approved):

PR #11185, opened 2023-06-01
feat(aws): Add Support for fully customizable Cost Explorer Data syncing

### LINKED ISSUE #10805: feat: AWS Cost by service with dynamic date range

Enabling optimum data pulls with additional granularity

Hi Team, Thank you for implementing the AWS Cost Explorer API integration. 

Can we please extend it to having an option for last 7 days (or something like last 'x' which is configurable) of data pull, since predominantly the syncs happen on a weekly basis and would like to avoid pulling last 30 days worth of cost information for every run. 

Also can we please have the service captured, since it helps drill down on what the cost is tied to. Sample cli command is given for reference

aws ce get-cost-and-usage \
    --time-period Start=2023-03-01,End=2023-03-07 \
    --granularity DAILY \
    --metrics "BlendedCost" "UnblendedCost" "UsageQuantity" \
    --group-by Type=DIMENSION,Key=SERVICE Type=TAG,Key=Environment

Regards,
Hari

### Pull request (optional)

bbernays:
#### Summary

closes #10805


plugins/source/aws/resources/services/costexplorer/cost_custom.go
```diff
@@ -0,0 +1,115 @@
...
+				Type:        arrow.BinaryTypes.String,
+				Resolver:    schema.PathResolver("inputHash"),
+				PrimaryKey:  true,
+			},
+			{
+				Name:        "input_json",
+				Description: `The JSON of the input used to generate this result.`,
+				Type:        cqtypes.ExtensionTypes.JSON,
+				Resolver:    schema.PathResolver("inputJSON"),
+				PrimaryKey:  false,
```
disq, line 57:
```suggestion
```

plugins/source/aws/client/tableoptions/costexplorer.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/costexplorer"
+	"github.com/cloudquery/plugin-sdk/v3/caser"
+)
+
+type CostExplorerAPIs struct {
+	GetCostAndUsageOpts []CustomGetCostAndUsageInput `json:"get_cost_and_usage,omitempty"`
+}
+
+type CustomGetCostAndUsageInput struct {
+	costexplorer.GetCostAndUsageInput
```
disq, line 17: Maybe also add a `name` here people can fill in and use as a 'preset' while accessing data (if they have multiple inputs?)
It would be better than keeping track of what `input_hash` corresponds to what input, or selecting data directly from `input_json` (if it's more than start/end dates... which again since `GetCostAndUsageInput` doesn't contain a name, they'd have to match multiple fields)

bbernays: I see the issue you are raising, but I am concerned about deviating from the exact AWS defined input as it would make documentation and usage more complex. 

I think no matter what they would have to match on multiple fields because I don't think they can guarantee the `name` field would be unique enough as it would be 100% user defiend

plugins/source/aws/client/tableoptions/table_options.go
```diff
@@ -4,6 +4,7 @@ type TableOptions struct {
 	CloudTrailEvents       *CloudtrailAPIs         `json:"aws_cloudtrail_events,omitempty"`
 	AccessAnalyzerFindings *AccessanalyzerFindings `json:"aws_accessanalyzer_analyzer_findings,omitempty"`
 	Inspector2Findings     *Inspector2APIs         `json:"aws_inspector2_findings,omitempty"`
+	CustomCostExplorer     *CostExplorerAPIs       `json:"aws_costexplorer_cost_custom,omitempty"`
```
disq, line 7: you should do:

```go
	if t. CustomCostExplorer != nil {
		err := t. CustomCostExplorer.Validate()
		if err != nil {
			return err
		}
	}
```

In the `Validate()` method below.

disq (changes requested):

plugins/source/aws/client/tableoptions/costexplorer.go
bbernays: Why would they access by row number? I would expect users to parse the `input_json` column to ensure they are using the right input... I would make that column a PK, but not all destinations support JSON PKs...

Also we would have to name the input field something like `_cq_input_name` to ensure that we never have a collision with a field in an input

disq: ok let's revisit once/if it becomes a user-reported problem.

disq (approved):

plugins/source/aws/resources/services/costexplorer/cost_custom.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/costexplorer/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v3/schema"
+	"github.com/cloudquery/plugin-sdk/v3/transformers"
+	cqtypes "github.com/cloudquery/plugin-sdk/v3/types"
+	"github.com/mitchellh/hashstructure/v2"
+)
+
+func CustomCost() *schema.Table {
+	tableName := "aws_costexplorer_cost_custom"
```
disq, line 20:
```suggestion
	tableName := "aws_alpha_costexplorer_cost_custom"
```

plugins/source/aws/resources/services/costexplorer/cost_custom.go
```diff
@@ -0,0 +1,113 @@
...
+	return &schema.Table{
+		Name:     tableName,
+		Resolver: fetchCustom,
+		Title:    "AWS Cost Explorer costs based on custom inputs",
+		Description: `https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetCostAndUsage.html
+To sync this table you must set the 'use_paid_apis' option to 'true' in the AWS provider configuration as well as specify the request parameters in the 'table_options' attribute. `,
```
disq, line 26:
```suggestion
To sync this table you must set the 'use_paid_apis' option to 'true' in the AWS provider configuration as well as specify the request parameters in the 'table_options' attribute. 

Please note that this table is considered **alpha** (experimental) and may have breaking changes or be removed in the future.
`,
```

disq (changes requested):

plugins/source/aws/resources/services/costexplorer/cost_custom.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/mitchellh/hashstructure/v2"
+)
+
+var tableName = "aws_alpha_costexplorer_cost_custom"
+
+func CustomCost() *schema.Table {
+	return &schema.Table{
+		Name:     tableName,
+		Resolver: fetchCustom,
+		Title:    "AWS Cost Explorer costs based on custom inputs",
```
disq, line 25:
```suggestion
		Title:    "AWS Cost Explorer costs based on custom inputs (Alpha)",
```

This change would also affect the `This table shows data for...` entry in the markdown and might look a bit different, but I don't think it's a problem.

disq (approved):
LGTM with nits

PR #11116, opened 2023-05-31
fix(deps): Update `github.com/cloudquery/plugin-sdk/v3` to `v3.7.0`

candiduslynx:

PR #11186, opened 2023-06-01
chore(duckdb): Refactor

candiduslynx:
It's an effort to shorten #11116 diff

### DIFF
--- a/plugins/destination/duckdb/client/transform.go
+++ b/plugins/destination/duckdb/client/transform.go
@@ -16,7 +16,7 @@ func transformRecord(sc *arrow.Schema, rec arrow.Record) arrow.Record {
 }
 
 func transformArray(arr arrow.Array) arrow.Array {
-	if plainWrite(arr.DataType()) {
+	if arrow.TypeEqual(arr.DataType(), transformTypeForWriting(arr.DataType())) {
 		return arr
 	}
 
--- a/plugins/destination/duckdb/client/transform.go
+++ b/plugins/destination/duckdb/client/transform.go
@@ -16,35 +16,37 @@ func transformRecord(sc *arrow.Schema, rec arrow.Record) arrow.Record {
 }
 
 func transformArray(arr arrow.Array) arrow.Array {
-	dt := arr.DataType()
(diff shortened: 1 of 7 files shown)


plugins/destination/duckdb/client/types.go
```diff
@@ -7,10 +7,45 @@ import (
...
+	fields := sc.Fields()
+	for i := range fields {
+		fields[i].Type = transformTypeForWriting(fields[i].Type)
+	}
+	md := sc.Metadata()
+	return arrow.NewSchema(fields, &md)
+}
+
+func transformType(dt arrow.DataType) arrow.DataType {
+	return duckDBToArrow(arrowToDuckDB(dt))
```
bbernays, line 42: Why wrap this function call in a function? What do we get for this added abstraction layer?

plugins/destination/duckdb/client/transform.go
```diff
@@ -0,0 +1,92 @@
...
+	for i := 0; i < int(rec.NumCols()); i++ {
+		cols[i] = transformArray(rec.Column(i))
+	}
+	return array.NewRecord(sc, cols, rec.NumRows())
+}
+
+func transformArray(arr arrow.Array) arrow.Array {
+	if plainWrite(arr.DataType()) {
+		return arr
+	}
```
bbernays, line 21: Why wrap this single boolean expression in a function why not just directly call:
```suggestion
	if arrow.TypeEqual(arr.DataType(), transformTypeForWriting(arr.DataType())) {
		return arr
	}
```

plugins/destination/duckdb/client/types.go
candiduslynx: Done in dafa00461042b75497856cabe705d2bc3c9d535e

plugins/destination/duckdb/client/transform.go
candiduslynx: Merged this + removed in dafa00461042b75497856cabe705d2bc3c9d535e

disq (approved):

PR #11188, opened 2023-06-02
chore(refactor): Refactor DuckDB

candiduslynx:
Extracted from #11116 (pt2)

### DIFF
--- a/plugins/destination/duckdb/client/read.go
+++ b/plugins/destination/duckdb/client/read.go
@@ -14,10 +14,9 @@ import (
 	"github.com/apache/arrow/go/v13/parquet/pqarrow"
 	"github.com/cloudquery/plugin-sdk/v3/schema"
 	"github.com/cloudquery/plugin-sdk/v3/types"
-	"github.com/google/uuid"
 )
 
-func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- arrow.Record) error {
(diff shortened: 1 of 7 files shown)


plugins/destination/duckdb/client/read.go
```diff
@@ -82,170 +81,109 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName strin
 }
 
 func convertToSingleRowRecords(sc *arrow.Schema, rec arrow.Record) []arrow.Record {
-	records := make([]arrow.Record, rec.NumRows())
-	for i := int64(0); i < rec.NumRows(); i++ {
-		records[i] = reverseTransformRecord(sc, rec.NewSlice(i, i+1))
+	// transform arrays first
+	transformed := reverseTransformRecord(sc, rec)
```
erezrokah, line 85: Can you explain the purpose of this refactor?

erezrokah (commented):
This looks good, thanks for breaking the refactoring to a separate PR.
I have one question though

plugins/destination/duckdb/client/read.go
candiduslynx: If we get a record with several rows, we'll have much less function calls as a result.
In the previous implementation we would transform the data row by row, calling at least `rec.NumCols()` functions for each row.
Now we are calling the transformation first, resulting in a multi-row record being returned, and then we just slice it.

erezrokah (approved):

disq (approved):

PR #11201, opened 2023-06-02
feat(gcp): Add Support For Service Account Impersonation

### LINKED ISSUE #10903: feat: GCP service account impersonation

When running the GCP source plugin, it is sometimes necessary to impersonation another GCP service account.
E.g., an application default credential (ADC) is already available for service account A.
However, this service account does not have the permissions required by CloudQuery GCP plugin.
It is therefore necessary to impersonation service account B, using the already available credential for service account A.

CloudQuery should then perform all GCP API calls using the impersonated service account B.

The gcloud CLI has an `--impersonate-service-account` parameter that provides the same feature (see [GCP documentation](https://cloud.google.com/sdk/gcloud/reference#--impersonate-service-account)).

The GCP Go SDK also has [documentation on the impersonate function.](https://pkg.go.dev/google.golang.org/api/impersonate).

Ideally, there will be a new parameter in the GCP source plugin spec that allows to specify the ID of the service account to be impersonated. E.g. `impersonate-service-account: foo@project-id.iam.gserviceaccount.com`.

### Pull request (optional)

bbernays:
#### Summary

closes #10903

### DIFF
--- a/plugins/source/gcp/client/client.go
+++ b/plugins/source/gcp/client/client.go
@@ -166,6 +166,8 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source, opts source
 		c.ClientOptions = append(c.ClientOptions, option.WithCredentialsJSON(serviceAccountKeyJSON))
 	}
 	if gcpSpec.ServiceAccountImpersonation != nil {
+		c.logger.Info().Msgf("%v", gcpSpec.ServiceAccountImpersonation)
+
 		// Base credentials sourced from ADC or provided client options.
 		ts, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
 			TargetPrincipal: gcpSpec.ServiceAccountImpersonation.TargetPrincipal,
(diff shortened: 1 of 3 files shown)


plugins/source/gcp/client/client.go
```diff
@@ -164,7 +165,25 @@ func New(ctx context.Context, logger zerolog.Logger, s specs.Source, opts source
...
+		c.logger.Info().Msgf("%v", gcpSpec.ServiceAccountImpersonation)
+
+		// Base credentials sourced from ADC or provided client options.
+		ts, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+			TargetPrincipal: gcpSpec.ServiceAccountImpersonation.TargetPrincipal,
+			Scopes:          gcpSpec.ServiceAccountImpersonation.Scopes,
+			// Optionally supply delegates.
+			Delegates: gcpSpec.ServiceAccountImpersonation.Delegates,
+			// Specify user to impersonate
+			Subject: gcpSpec.ServiceAccountImpersonation.Subject,
```
bbernays, line 176: These options were not documented because I could not fully understand the use case where they can be used, but are required inputs. So if a user has a need for them we can add documentation for them

website/pages/docs/plugins/sources/gcp/configuration.md
```diff
@@ -75,6 +75,16 @@ If enabled CloudQuery will skip any resources that belong to a service that has
...
 
+- `service_account_impersonation` ([Rate limit](#service-account-impersonation-spec) spec, optional. Default: empty)
+
+  Service Account impersonation configuration.
+
+### Service Account Impersonation Spec
+
+- `target_principal` (`string`, optional. Default: empty)
+
+  	The email address of the service account to impersonate
```
disq, line 86:
```suggestion
    The email address of the service account to impersonate
```

PR #11212, opened 2023-06-05
feat(postgres)!: Update to SDK V3 Arrow native

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/10763

~~Keeping in draft until we sort out some issues related to internal CQ columns being handled differently in SDK v3~~

BEGIN_COMMIT_OVERRIDE
feat: Update to use [Apache Arrow](https://arrow.apache.org/) type system (#11212)

BREAKING-CHANGE: This release introduces an internal change to our type system to use [Apache Arrow](https://arrow.apache.org/). This should not have any visible breaking changes, however due to the size of the change we are introducing it under a major version bump to communicate that it might have some bugs that we weren't able to catch during our internal tests. If you encounter an issue during the upgrade, please submit a [bug report](https://github.com/cloudquery/cloudquery/issues/new/choose). You will also need to update destinations depending on which one you use:
- Azure Blob Storage >= v3.2.0
- BigQuery >= v3.0.0
- ClickHouse >= v3.1.1
- DuckDB >= v1.1.6
- Elasticsearch >= v2.0.0
- File >= v3.2.0
- Firehose >= v2.0.2
- GCS >= v3.2.0
- Gremlin >= v2.1.10
- Kafka >= v3.0.1
- Meilisearch >= v2.0.1
- Microsoft SQL Server >= v4.2.0
- MongoDB >= v2.0.1
- MySQL >= v2.0.2
- Neo4j >= v3.0.0
- PostgreSQL >= v4.2.0
- S3 >= v4.4.0
- Snowflake >= v2.1.1
- SQLite >= v2.2.0

END_COMMIT_OVERRIDE

PR #11214, opened 2023-06-05
feat(mysql)!: Update to SDK V3 Arrow native

erezrokah:
Closes https://github.com/cloudquery/cloudquery/issues/10757

BEGIN_COMMIT_OVERRIDE
feat: Update to use [Apache Arrow](https://arrow.apache.org/) type system (#11214)

- BigQuery >= v3.0.0
- ClickHouse >= v3.1.1
- DuckDB >= v1.1.6
- Elasticsearch >= v2.0.0
- File >= v3.2.0
- Firehose >= v2.0.2
- GCS >= v3.2.0
- Gremlin >= v2.1.10
- Kafka >= v3.0.1
- Meilisearch >= v2.0.1
- MongoDB >= v2.0.1
- MySQL >= v2.0.2
- Neo4j >= v3.0.0
- PostgreSQL >= v4.2.0
- S3 >= v4.4.0
- Snowflake >= v2.1.1
- SQLite >= v2.2.0

END_COMMIT_OVERRIDE


plugins/source/mysql/client/sync.go
```diff
@@ -97,10 +124,6 @@ func (c *Client) resourceFromValues(tableName string, values []any) (*schema.Res
 	table := c.Tables.Get(tableName)
 	resource := schema.NewResourceData(table, nil, values)
 	for i, col := range table.Columns {
-		// If the value points to an empty byte array array we set it to nil
-		if v, ok := values[i].(**[]byte); ok && *v == nil {
```
erezrokah, line 101: This was a workaround and no longer needed since the scalar UUID supports `**[]byte` due to https://github.com/cloudquery/plugin-sdk/pull/922

plugins/source/mysql/client/types.go
```diff
@@ -1,64 +1,143 @@
...
+		scale = defaultScale
+	}
+	return precision, scale
+}
+
+func SQLType(t arrow.DataType) string {
+	switch v := t.(type) {
+	case *arrow.BooleanType:
 		return "bool"
-	case schema.TypeInt:
```
erezrokah, line 11: To avoid losing data for unsigned bigint I added support to all various int types + unsigned ones.
This is a breaking change

plugins/source/mysql/client/types.go
```diff
@@ -1,64 +1,146 @@
...
+		return "bigint unsigned"
+	case *arrow.Float16Type, *arrow.Float32Type:
 		return "float"
-	case schema.TypeUUID:
+	case *arrow.Float64Type:
+		return "double"
+	case *arrow.Decimal128Type:
+		return fmt.Sprintf("decimal(%d,%d)", v.Precision, v.Scale)
+	case *arrow.Decimal256Type:
+		return fmt.Sprintf("decimal(%d,%d)", v.Precision, v.Scale)
```
candiduslynx, line 74:
```suggestion
	case arrow.DecimalType:
		return fmt.Sprintf("decimal(%d,%d)", v.GetPrecision(), v.GetScale())
```
nit

plugins/source/mysql/client/types.go
```diff
@@ -1,64 +1,146 @@
...
+	case *types.UUIDType:
 		return "binary(16)"
-	case schema.TypeByteArray:
+	case *arrow.BinaryType, *arrow.LargeBinaryType, *arrow.FixedSizeBinaryType:
 		return "blob"
```
candiduslynx, line 78: Do we want to make this account for the fixes binary width?
```suggestion
	case *arrow.BinaryType, *arrow.LargeBinaryType:
		return "blob"
	case *arrow.FixedSizeBinaryType:
		return "blob(" + strconv.Itoa(v.ByteWidth) + ")"
```

plugins/source/mysql/resources/plugin/plugin_test.go
```diff
@@ -65,7 +67,7 @@ func createTable(ctx context.Context, db *sql.DB, table *schema.Table) error {
 		builder.WriteString(" PRIMARY KEY (")
 		for i, pk := range pks {
 			builder.WriteString(client.Identifier(pk))
-			if table.Columns.Get(pk).Type == schema.TypeString {
+			if table.Columns.Get(pk).Type == arrow.BinaryTypes.String {
```
candiduslynx, line 70:
```suggestion
			if arrow.TypeEqual(table.Columns.Get(pk).Type, arrow.BinaryTypes.String) {
```
as otherwise you are comparing types and might get into trouble if someone uses `new(arrow.StringType)`

plugins/source/mysql/resources/plugin/plugin_test.go
```diff
@@ -93,13 +95,78 @@ func insertTable(ctx context.Context, db *sql.DB, table *schema.Table, data sche
...
-	dbData := schema.TransformWithTransformer(&client.Transformer{}, data)
-	if _, err := db.ExecContext(ctx, sb.String(), dbData...); err != nil {
-		return err
+	for _, record := range records {
+		transformedRecords, err := client.TransformRecord(record)
+		if err != nil {
+			return err
+		}
+		for _, transformedRecord := range transformedRecords {
+			if _, err := db.ExecContext(ctx, sb.String(), transformedRecord...); err != nil {
```
candiduslynx, line 104: save query once, no need to call `sb.String()` several times

plugins/source/mysql/resources/plugin/plugin_test.go
```diff
@@ -123,26 +190,27 @@ func TestPlugin(t *testing.T) {
...
-	otherTable := testdata.TestSourceTable("other_mysql_table")
+	otherTable := schema.TestTable("other_mysql_table", schema.TestSourceOptions{})
 	if _, err := db.ExecContext(ctx, "DROP TABLE IF EXISTS other_mysql_table"); err != nil {
 		t.Fatal(err)
 	}
 	if err := createTable(ctx, db, otherTable); err != nil {
 		t.Fatal(err)
 	}
-	otherData := testdata.GenTestDataV1(otherTable)
+	otherData := schema.GenTestData(otherTable, schema.GenTestDataOptions{MaxRows: 1, SyncTime: syncTime})
```
candiduslynx, line 213: does it allow for 2 rows (to test the null row)?

candiduslynx (commented):
some comments & nits

plugins/source/mysql/resources/plugin/plugin_test.go
erezrokah: We test for 2 rows a bit earlier in the code (see `data := schema.GenTestData`). `otherData` is inserted o a table that is not part of the spec tables and meant to check that we don't sync it.

plugins/source/mysql/client/types.go
candiduslynx: np, just wanted to check in with you on this feature.
Maybe `binary(n)` would work here better?

candiduslynx (approved):

candiduslynx (approved):

candiduslynx (approved):

erezrokah:
This is on hold until for the same reason as https://github.com/cloudquery/cloudquery/pull/11212 (an issue with CQ internal columns in SDK v3)

erezrokah:
Going to merge this (but not release the MySQL source) so I can use this as a baseline for the v4 upgrade

PR #11215, opened 2023-06-05
fix(race): Fix race in DuckDB delete-stale

candiduslynx:


plugins/destination/duckdb/client/write.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/apache/arrow/go/v13/arrow"
 	"github.com/apache/arrow/go/v13/parquet"
 	"github.com/apache/arrow/go/v13/parquet/pqarrow"
-	"github.com/cloudquery/plugin-pb-go/specs"
+	backoff "github.com/cenkalti/backoff/v4"
```
disq, line 12:
```suggestion
	"github.com/cenkalti/backoff/v4"
```

plugins/destination/duckdb/client/write.go
```diff
@@ -134,39 +125,46 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, recor
...
+	if err := c.deleteByPK(ctx, tmpTableName, table); err != nil {
+		return err
+	}
+
+	// per https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking we might need to retry
+	return backoff.Retry(
+		func() error {
+			return c.exec(ctx, "insert into "+table.Name+" from "+tmpTableName)
+		},
+		backoff.WithContext(backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 3), ctx),
```
hermanschaaf, line 176: How long is the backoff period in total after 3 retries? Should we make the number of retries configurable? Because I think the number you'll need will depend on how many rows are being inserted / deleted

candiduslynx: I moved to a constant backoff with 50ms.
We can address this if someone reports issues even after this fix.

hermanschaaf: It's hard to know whether this fix is going to generalize beyond the test, because we're only inserting a few rows there.  So I'd probably lean towards a higher number of retries, at least, if not making it fully configurable. But yeah we can see if we get any reports of this issue after the change.

plugins/destination/duckdb/client/write.go
```diff
@@ -134,39 +134,46 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, recor
...
+
+func (c *Client) deleteInsert(ctx context.Context, tmpTableName string, table *schema.Table) error {
+	if err := c.deleteByPK(ctx, tmpTableName, table); err != nil {
+		return err
+	}
+
+	// per https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking we might need to retry
+	return backoff.Retry(
+		func() error {
+			return c.exec(ctx, "insert into "+table.Name+" from "+tmpTableName)
```
hermanschaaf, line 175: is there any way we can detect the unique constraint violation error and only retry that, rather than retrying all errors?

candiduslynx: We could detect it only via comparing error strings, but I'd rather leave it to the further implementation if we ever need to touch this place once again.
I hope that the [issue](https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking) is fixed by DuckDB itself thus making this code obsolete and enabling us to remove this backoff sooner though.

hermanschaaf (approved):

PR #11228, opened 2023-06-06
fix(azure): Fix some CIS Azure policies

aruneko:
## Section 1.21

I have misunderstood the condition of CIS Azure Section 1.21.

It should be checked for only custom roles.

And I forgot to put `INSERT` query in this SQL file...

## Section 4.1.1

The join condition on from query has been fixed for actual relationship.

## Section 4.1.2

According to the CIS benchmark documentation, they say below.

> TDE cannot be used to encrypt the logical master database in SQL Database. The master database contains objects that are needed to perform the TDE operations on the user databases.

So, we have to exclude `master` database from results.


yevgenypats (approved):

erezrokah:
/gen sha=b48d18fa7177a2462618ae5cd840ed7cd7192ff5 plugin=azure

PR #11299, opened 2023-06-06
fix(ci): When checking number of commits in release summary, only include `cq-bot` commits

erezrokah:
#### Summary

To avoid a recursive workflow when adding the release summary to release PRs, we check that the number of commits is 1, and only then run the release summary workflow.
The issue is that sometimes Kodiak's is faster than our CI, and does a merge with main commit before we manage to add the summary commit, hence it never gets added.

This PR improves the CI by filtering out commits to only include `cq-bot` commits, so Kodiak commits are ignored.
See example PR where this race condition happened:
https://github.com/cloudquery/cloudquery/pull/11139
https://github.com/cloudquery/cloudquery/pull/11139/commits
https://github.com/cloudquery/cloudquery/actions/runs/5192224428/jobs/9361094081


.github/workflows/release_summary.yml
```diff
@@ -7,9 +7,31 @@ on:
       - ".github/workflows/release_summary.yml"
 
 jobs:
+  trigger-summary:
+    runs-on: ubuntu-latest
+    name: Commits Check
+    # only run on release-please branches
+    if:  startsWith(github.head_ref, 'chore/fix_release_summary_condition')
```
erezrokah, line 14: I'll revert this test in a bit

candiduslynx (approved):

Issue #11325, opened 2023-06-07
bug: char() bug for mysql source > mysql destination


pandaijoeyong:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when setting up a source of MariaDB to MySQL, (using the mysql connector), i get an char() error.

ec2 AMI: ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20230516

Error:
`Error: failed to sync v1 source mysql: rpc error: code = Unknown desc = got unknown MySQL type "char(32)" for column "USER" of table "users" while trying to convert it to CloudQuery internal schema type. Supported MySQL types are ["bigint" "bigint unsigned" "bigint(20)" "bigint(20) unsigned" "binary(16)" "blob" "datetime" "double" "float" "int" "int unsigned" "int(10) unsigned" "int(11)" "json" "smallint" "smallint unsigned" "smallint(5)" "smallint(5) unsigned" "smallint(6)" "text" "tinyint" "tinyint unsigned" "tinyint(1)" "tinyint(3) unsigned" "tinyint(4)"]`

### Expected Behavior

no errors.

### CloudQuery (redacted) config

<cloudquery-config/mysql_source.yaml>
```
kind: source
spec:
  name: "mysql"
  path: "cloudquery/mysql"
  version: "v1.2.0"
  tables: ["*"]
  destinations: ["mysql"]
  spec:
    connection_string: "user:password@tcp(dns:port)/dbname"
```

<cloudquery-config/mysql_destination.yaml>
```
kind: destination
spec:
  name: "mysql"
  registry: "github"
  path: "cloudquery/mysql"
  version: "v2.1.1"
  spec:
    connection_string: "user:password@tcp(dns:port)/dbname"
```

### Steps To Reproduce

1. In ubuntu-jammy-22.04, sudo nano the following cloudquery config files
2. run ./cloudquery sync cloudquery-config/mysql_source.yaml cloudquery-config/mysql_destination.yaml

### CloudQuery (redacted) logs

023-06-07T07:43:43Z INF Loading spec(s) args=["./cloudquery-config/mysql_source.yaml","cloudquery-config/mysql_destination.yaml"] module=cli
2023-06-07T07:43:44Z INF Source plugin server listening address=/tmp/cq-kOBmPLmBrRLPvAec.sock module=cli
[...]

PR #11360, opened 2023-06-07
feat(dest-mysql): Handle all MySQL types

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/11325
Same as https://github.com/cloudquery/cloudquery/pull/9143 for MySQL destination.
Type handling taken from https://github.com/cloudquery/cloudquery/pull/11214/files#diff-c4ee14fff2e1c7cbea7dfcf122879522b5d817aaa041f54564696293b1c6bc30R90.

Once we have the MySQL source migrated to v3 we can add support for more granular types like `char(x)` or other fixed size types (doing that will be a breaking change).


plugins/destination/mysql/client/types.go
```diff
@@ -1,55 +1,106 @@
...
+	if sqlType == "binary(16)" {
+		return types.ExtensionTypes.UUID
+	}
+	if sqlType == "tinyint(1)" {
+		return arrow.FixedWidthTypes.Boolean
+	}
+	if strings.HasPrefix(sqlType, "datetime") {
+		return arrow.FixedWidthTypes.Timestamp_us
+	}
+	if strings.HasPrefix(sqlType, "decimal") || strings.HasPrefix(sqlType, "numeric") {
```
hermanschaaf, line 60: It looks like this might fail in the future if we start supporting general array types via Arrow, but right now this is not yet supported on the source side so we're okay. It'll be a bigger change in any case

hermanschaaf (approved):

PR #11388, opened 2023-06-07
feat(azure)!: Force ID values to be all lowercase

disq:
Just a draft. I don't like the untestability of this as its current state (with injected PostResourceResolvers).

### DIFF
--- a/plugins/source/azure/client/middleware.go
+++ b/plugins/source/azure/client/middleware.go
@@ -1,0 +1,52 @@
+package client
+
+import (
+	"context"
+	"reflect"
+	"strings"
+
+	"github.com/cloudquery/plugin-sdk/v3/schema"
+)
+
+func LowercaseIDResolver(_ context.Context, _ schema.ClientMeta, resource *schema.Resource) error {
+	resource.SetItem(lowercaseID(resource.GetItem()))
+	return nil
+}
+
+func ChainRowResolvers(next ...schema.RowResolver) schema.RowResolver {
+	return func(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource) error {
+		for i := range next {
+			if next[i] == nil {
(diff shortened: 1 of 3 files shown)


erezrokah (commented):
Can we make this optional via configuration so it's not a breaking change?
If it works well we can make it the default.

Also maybe there's a way to hook into the Azure SDK and handle it here? The team is quite responsive so we can ask via an issue here https://github.com/Azure/azure-sdk-for-go/issues

disq:
> Can we make this optional via configuration so it's not a breaking change? If it works well we can make it the default.
done

> Also maybe there's a way to hook into the Azure SDK and handle it here? The team is quite responsive so we can ask via an issue here [Azure/azure-sdk-for-go/issues](https://github.com/Azure/azure-sdk-for-go/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc)
[Asked](https://github.com/Azure/azure-sdk-for-go/issues/20982), I don't think they'll recommend anything other than a custom `Transport` (or a `Policy`) entry in `ClientOptions.PerCallPolicies` which would be just as bad (to "detect" ids in the http.Response and process...) if not worse.

plugins/source/azure/client/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func LowercaseIDResolver(_ context.Context, c schema.ClientMeta, resource *schema.Resource) error {
+	if c.(*Client).pluginSpec.NormalizeIDs {
+		resource.SetItem(lowercaseID(resource.GetItem()))
+	}
+	return nil
+}
+
+func ChainRowResolvers(next ...schema.RowResolver) schema.RowResolver {
```
disq, line 18: This is currently unused and may be removed.

PR #11400, opened 2023-06-08
feat!: Rename tables

candiduslynx:


hermanschaaf (changes requested):
I don't think we need to do this, and I don't see it tied to a customer issue; maybe the names aren't perfect but we don't need to create unnecessary work for those relying on the Oracle plugin (who will need to go update their queries)

Issue #11236, opened 2023-06-06
feat: S3 Bucket storage information in cloudquery


rajeshwarakella1994:
### Which problem is this feature request solving?

Hello Team,

the aws_s3_bucket table in AWS source plugin today doesn't have anything related to bucket size. 
This information will be very useful to identity buckets which are large and such buckets can be looked into for optimisations to save opearational costs.

### Describe the solution you'd like

Eventhough the S3 API provided by AWS doesn't provide storage statistics, we can retrieve this information from cloudwatch metrics. Here is an example of aws cli which illustrates the same - 

aws cloudwatch get-metric-statistics --namespace AWS/S3 --start-time 2023-06-05T00:00:00 --end-time 2023-06-06T00:00:00 --period 86400 --statistics Average --metric-name BucketSizeBytes --unit Bytes --dimensions Name=BucketName,Value={Bucket-Name} Name=StorageType,Value=StandardStorage

### Pull request (optional)

- [ ] I can submit a pull request

bbernays:
CloudWatch metrics are not a free API. I suggest we use the new `table_options` parameter and `use_paid_apis` to enable users to define exactly what they want to sync.


This would follow the same pattern as #11185

disq:
@rajeshwarakella1994 This feature is now released AWS [18.4.0](https://github.com/cloudquery/cloudquery/compare/plugins-source-aws-v18.3.0...plugins-source-aws-v18.4.0), you can try it with a `table_options` entry in the Spec, like so:

```yaml
    table_options:
      aws_alpha_cloudwatch_metrics:
        - list_metrics:
            namespace: AWS/S3
            metric_name: BucketSizeBytes
          get_metric_statistics:
            - statistics:
                - Average
              start_time: "2023-06-06T00:00:00Z"
              end_time: "2023-06-11T00:00:00Z"
              period: 86400
```

This will list all `BucketSizeBytes` in the `AWS/S3` namespace (so every bucket will get its own metric entry) and get Average statistics with the given time period/options for each. Please let us know if you encounter anything unexpected. (To manage different or changing time periods, [environment variable substitution](https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution) feature can be used)

rajeshwarakella1994:
Hello @disq .. Thank you very much for the feature and sincere apologies for not being able to respond here.

[...]

PR #11402, opened 2023-06-08
feat(aws): Add `aws_alpha_cloudwatch_metric*` tables to fetch Cloudwatch Metrics and statistics

disq:
Fixes https://github.com/cloudquery/cloudquery/issues/11236

- **aws_alpha_cloudwatch_metrics**: List of metrics, changes with what's available (in terms of resources in the account, as in every S3 bucket will generate new metric entries) can be filtered with `table_options.aws_alpha_cloudwatch_metrics.list_metrics`. (If you want all metrics available in the account, you still have to create an empty entry)
  - **aws_alpha_cloudwatch_metric_statistics**: Statistics for the listed metrics. Needs to have `table_options.aws_alpha_cloudwatch_metrics[].get_metric_statistics` set with appropriate statistic and time periods. `Dimensions`, `MetricName` and `Namespace` is derived from the parent table (can't be set in the opts)
~- **aws_cloudwatch_metric_statistics_custom**: Custom statistics for any metric, set with `table_options.aws_cloudwatch_metric_stats_custom.get_metric_statistics`. No expansion is done, so e.g. for each S3 bucket the user needs to insert another entry.~

Example:
```yaml
    use_paid_apis: true
    table_options:
      aws_alpha_cloudwatch_metrics:
        - list_metrics:
            namespace: AWS/S3
          get_metric_statistics:
            - statistics:
                - Average
              start_time: "2023-06-06T00:00:00Z"
              end_time: "2023-06-11T00:00:00Z"
              period: 86400
        - list_metrics:
            namespace: AWS/Usage
            metric_name: CallCount
            dimensions:
              - name: Type
                value: API
              - name: Service
                value: EC2
          get_metric_statistics:
            - statistics:
                - Average
              start_time: "2023-06-01T00:00:00Z"
              end_time: "2023-06-06T00:00:00Z"
              period: 86400
```

Which will (a) query all Average statistics for every available metric in `AWS/S3` namespace using the given time period and (b) query CallCount statistics for the EC2 service in the `AWS/Usage` namespace for a longer period. It's also possible to query multiple time periods (for the same list_metric result) by adding another list-item under `get_metric_statistics`.


plugins/source/aws/resources/services/cloudwatch/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/cloudquery/plugin-sdk/v3/transformers"
+)
+
+func Metrics() *schema.Table {
+	tableName := "aws_cloudwatch_metrics_custom"
+	return &schema.Table{
+		Name:        tableName,
+		Description: `https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Metric.html`,
+		Resolver:    fetchCloudwatchMetrics,
+		Multiplex:   client.ServiceAccountRegionMultiplexer(tableName, "logs"),
```
disq, line 22: I'm not sure if `logs` is the correct service here but couldn't find anything suitable in our services json.

plugins/source/aws/resources/services/cloudwatch/metric_statistics.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func MetricStatistics() *schema.Table {
+	tableName := "aws_cloudwatch_metric_stats_custom"
+	return &schema.Table{
+		Name:        tableName,
+		Description: `https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html`,
+		Resolver:    fetchCloudwatchMetricStats,
+		Multiplex:   client.ServiceAccountRegionMultiplexer(tableName, "logs"),
+		Transform:   transformers.TransformWithStruct(&statOutput{}, transformers.WithSkipFields("ResultMetadata"), transformers.WithUnwrapAllEmbeddedStructs()),
```
disq, line 26: No PKs. Ideal PKs would have to include `input_json` which could complicate things for some dests.

plugins/source/aws/resources/services/cloudwatch/metrics.go
bbernays: You want to use `monitoring` ... https://docs.aws.amazon.com/general/latest/gr/cw_region.html

disq: Thanks, now fixed

plugins/source/aws/resources/services/cloudwatch/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func fetchCloudwatchMetrics(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+
+	if len(cl.Spec.TableOptions.CloudwatchMetrics) > 0 && !cl.Spec.UsePaidAPIs {
+		cl.Logger().Info().Msg("skipping `aws_cloudwatch_metrics` because `use_paid_apis` is set to false")
+		return nil
+	}
+
```
bbernays, line 47: If the input is `nil` we should output a message or a warning letting the user know why nothing was synced

plugins/source/aws/resources/services/cloudwatch/metric_statistics.go
bbernays: I think we should hash the `input_json` (or leave it as a string) so that all destinations can use it as a PK. 

There is a timestamp for each data point `statOutput.Datapoints[].Timestamp` I think that should also be a PK

plugins/source/aws/resources/services/cloudwatch/metrics.go
disq: done in https://github.com/cloudquery/cloudquery/pull/11402/commits/fe2d74315f1ee67feeef87865293f7151fda935d

website/tables/aws/aws_cloudwatch_metric_stats_custom.md
```diff
@@ -0,0 +1,21 @@
+# Table: aws_cloudwatch_metric_stats_custom
+
+This table shows data for Cloudwatch Metric Stats Custom.
+
+https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html
```
hermanschaaf, line 5: Can we update both the `Title` and the `Description` field of this table so that the documentation makes it clear what this table does and how it works?

plugins/source/aws/resources/services/cloudwatch/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+			metricStatistics(),
+		},
+	}
+}
+
+func fetchCloudwatchMetrics(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+
+	if len(cl.Spec.TableOptions.CloudwatchMetrics) > 0 && !cl.Spec.UsePaidAPIs {
+		cl.Logger().Info().Msg("skipping `aws_cloudwatch_metrics` because `use_paid_apis` is set to false")
```
hermanschaaf, line 44: I wonder if this shouldn't both return an error and log an error-level message, because there's a configuration issue if this is the case.

disq: Good catch

bbernays: If a user specifies `tables: "*"` there is no way for them to get a "clean" sync. They would have to add `aws_*_custom` to their skip tables

hermanschaaf: @bbernays I think that's fine. If they specify `*` then their intention seems to be to sync this API, so we should tell them when it's not working.

hermanschaaf: (Plus, there is another way to get a clean sync, and that is to enable `use_paid_apis`)

website/tables/aws/aws_cloudwatch_metric_stats.md
```diff
@@ -0,0 +1,25 @@
+# Table: aws_cloudwatch_metric_stats
```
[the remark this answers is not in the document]
disq: Fixed in https://github.com/cloudquery/cloudquery/pull/11402/commits/8b8abfa0791779f644145ac81afe08fbe2cb7fce

plugins/source/aws/resources/services/cloudwatch/metrics.go
bbernays: I agree, but if the `table_options` are still in experimental stage I am not sure that users should be required to specify anything for them if they are not actively using them...

website/tables/aws/aws_cloudwatch_metric_stats_custom.md
disq: Added a blurb, more explanation in configuration.md. Also since table_options is an experimental feature it doesn't have to be 100% bombproof documented, we can add a blog post for it later.

plugins/source/aws/resources/services/cloudwatch/metric_statistics.go
disq: I wouldn't know what to do with those timestamps since they're inside a list (unless you're suggesting each datapoint gets its own row in a sub-table) but I've added `input_hash` to all tables (and `parent_input_hash` to the metric_statistics table, to account for some potential PK clashes)

I couldn't bring myself to encode the json as string just to use it as a string-PK while at the same time making it painful (or inconsistent) to query later.

plugins/source/aws/resources/services/cloudwatch/metrics.go
disq: I think getting an error when using `*` is fine. Besides it probably won't be the single error they get ;-)

plugins/source/aws/resources/services/cloudwatch/metric_statistics.go
disq: > I do think that each data point should be its own row... That way if a user doesn't have to replace the entire object every single sync

@bbernays giving it a go in https://github.com/cloudquery/cloudquery/pull/11402/commits/ff50cd17ec8560a239b01af2bac425dbad293654

hermanschaaf (changes requested):
Shouldn't CloudWatch metrics be an incremental table? It seems like a perfect fit for this, because it will also do away with the requirement to update the time ranges in the config before every sync. But it gets a bit more complicated: we'd need to break up the time ranges into 1440-point increments to do a full refresh. (Check out `aws_cloudtrail_events` for a very close analogue to what we have here) WDYT?

hermanschaaf:
> @disq This would require us to assume the use cases.

I don't think so; what I mean is that we can / should use incremental updates for the start time only (and we can assume end time is now, unless specified otherwise). All the other options would still need to be specified, so we wouldn't need to assume anything.

disq:
`GetMetricStatistics` number of datapoints limitation:

> // Gets statistics for the specified metric. The maximum number of data points
// returned from a single call is 1,440. If you request more than 1,440 data
// points, CloudWatch returns an error. To reduce the number of data points, you
// can narrow the specified time range and make multiple requests across adjacent
// time ranges, or you can increase the specified period. Data points are not
// returned in chronological order.

We'd need to 'decide' on the next time period depending on the specified `Period` and how many datapoints there would be in that and so on. Assuming 'now' as end time makes sense. Another point is we shouldn't calculate the hash after setting our 'modified' inputs like modified start_time and end_time, so we should make sure that both `input_json` and the hash matches the original table_options input for that Metric entry.

website/tables/aws/aws_alpha_cloudwatch_metrics.md
```diff
@@ -0,0 +1,29 @@
+# Table: aws_alpha_cloudwatch_metrics
+
+This table shows data for Alpha Cloudwatch Metrics.
```
hermanschaaf, line 3: We may want to change the title of the table to `Cloudwatch Metrics (Alpha)`

website/tables/aws/aws_alpha_cloudwatch_metrics.md
```diff
@@ -0,0 +1,29 @@
...
+
+https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_ListMetrics.html
+To sync this table you must set the 'use_paid_apis' option to 'true' and set the relevant 'table_options' entry in the AWS provider configuration.
```
hermanschaaf, line 6: Also maybe in the description we should add:

```
Please note that this table is considered **alpha** (experimental) and may have breaking changes or be removed in the future.
```

hermanschaaf (approved):
LGTM, I think this is a great first step and will allow users to give us some feedback and set the future direction. I just had some nits about documentation for both tables, otherwise all good

Issue #11419, opened 2023-06-11
`WAFV2:GetWebACLForResource` Rate Limiting in `aws_elbv2_load_balancers`


bbernays:
Should the AWS plugin continue to make the `WAFV2:GetWebACLForResource` call as part of resolving `aws_elbv2_load_balancers`? It is an issue because users have no ability to not make that call currently because there is no way of skipping resolving a column.

This API has a low limit of 1 req/sec so if you have a lot of ALBs in a specific region you can easily be throttled.

A few options:
1. Drop the column, and have users use the `aws_wafv2_web_acls.resources_for_web_acl` column to connect an ALB to  to a webACL
2. Make the column a separate table so users can skip it if they want.
3. Leave it as is

PR #11421, opened 2023-06-11
fix(aws)!: Move `web_acl_arn` to its own table

bbernays:
#### Summary

Moves `web_acl_arn` column to its own table and removes the old column #11419 

If we want to do this in a backwards compatible way we could look into doing that as well, but I didn't go through that effort until we decided which path to go down


disq (approved):

hermanschaaf (approved):

website/tables/aws/aws_elbv2_load_balancer_web_acls.md
```diff
@@ -0,0 +1,39 @@
+# Table: aws_elbv2_load_balancer_web_acls
+
+This table shows data for Amazon Elastic Load Balancer (ELB) v2 Load Balancer Web ACLs.
+
+https://docs.aws.amazon.com/waf/latest/APIReference/API_GetWebACLForResource.html
+
+The primary key for this table is **_cq_id**.
```
hermanschaaf, line 7: Any primary key we could be using here?

bbernays: updated to use `load_balancer_arn,arn`

hermanschaaf (approved):

PR #11422, opened 2023-06-11
doc(website): Fix scaffold command

bbernays:


disq (approved):
According to the `.goreleaser.prerelease.yaml` line `binary: cq-scaffold` this is correct.

Issue #11410, opened 2023-06-09
bug: missing column _cq_source_name


BubbaJoe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

after upgrading to cloudquery 3.5.1 and updating firestore to 2.0.1, i am now getting this error below:

```
2023-06-09T18:55:33Z ERR exiting with error error="failed to sync v2 source firestore: rpc error: code = Unknown desc = table actorWallet is missing column _cq_source_name. please consider upgrading source plugin" module=cli
```

### Expected Behavior

No error

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: "firestore"
  registry: "github"
  path: cloudquery/firestore
  version: "v2.0.1"
  tables: ["taxis3"]
  destinations:
    - "json-s3-latest"
    - "json-s3-with-date"
    - "parquet-s3-latest"
    - "parquet-s3-with-date"
  spec:
    use_base64: true
    service_account_json: ${B64_SERVICE_ACCOUNT}
---
kind: destination
spec:
  name: "parquet-s3-latest"
  path: "cloudquery/s3"
  version: "v4.4.1"
  write_mode: "append"
  batch_size: 10_000_000 # optional
  batch_size_bytes: 1073741824 # optional
  spec:
    bucket: ${AWS_BUCKET}
    region: ap-northeast-1
    path: "tmp/{{TABLE}}-{{YEAR}}-{{MONTH}}-{{DAY}}.parquet"
    format: "parquet"
---
kind: destination
spec:
  name: "parquet-s3-with-date"
  path: "cloudquery/s3"
  version: "v4.4.1"
  write_mode: "append"
  batch_size: 5_000_000 # optional
  batch_size_bytes: 1073741824 # optional
  spec:
    bucket: ${AWS_BUCKET}
    region: ap-northeast-1
    path: "tmp/{{TABLE}}-latest.parquet"
    format: "parquet"
---
kind: destination
spec:
  name: "json-s3-latest"
  path: "cloudquery/s3"
  version: "v4.4.1"
  write_mode: "append"
  batch_size: 10_000_000 # optional
  batch_size_bytes: 1073741824 # optional
  spec:
    bucket: ${AWS_BUCKET}
    region: ap-northeast-1
    path: "tmp/{{TABLE}}-{{YEAR}}-{{MONTH}}-{{DAY}}.json"
    format: "json"
---
kind: destination
spec:
  name: "json-s3-with-date"
[...]

PR #11429, opened 2023-06-12
fix: Add internal columns

BubbaJoe:
#### Summary

This fixes the `missing column _cq_source_name` error for the firestore source integration.

Closes https://github.com/cloudquery/cloudquery/issues/11410


erezrokah:
Thanks for the fix @BubbaJoe, after doing some testing we had to revert this fix, see https://github.com/cloudquery/cloudquery/pull/11452 for reasoning.

There's an ongoing issue with SDK v3 and internal columns we're working on that should address the issue. We'll deprecate the firestore versions that are broken (versions >= `v2`)

BubbaJoe:
> Thanks for the fix @BubbaJoe, after doing some testing we had to revert this fix, see https://github.com/cloudquery/cloudquery/pull/11452 for reasoning.
> 
> 
> 
> There's an ongoing issue with SDK v3 and internal columns we're working on that should address the issue. We'll deprecate the firestore versions that are broken (versions >= `v2`)

Is reverting the correct decision? Without it it doesn't work at all and with it, it works but not internal fields which were not there before.

erezrokah:
> Is reverting the correct decision? Without it it doesn't work at all and with it, it works but not internal fields which were not there before.

Sorry for not being it clear. It doesn't work with the fix. Writing to the destinations fails (you can see errors the logs), so the data is not saved in the destination.

BubbaJoe:
@erezrokah Sorry I don't see the errors logs you are mentioning, when i tested it locally, it actually worked.

PR #11441, opened 2023-06-12
doc: Update default configurations to use non default concurrency value

bbernays:

### DIFF
--- a/website/pages/docs/plugins/sources/alicloud/_configuration.mdx
+++ b/website/pages/docs/plugins/sources/alicloud/_configuration.mdx
@@ -4,7 +4,7 @@ spec:
   name: "alicloud"
   path: "cloudquery/alicloud"
   version: "VERSION_SOURCE_ALICLOUD"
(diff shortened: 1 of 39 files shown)


website/pages/docs/plugins/sources/alicloud/_configuration.mdx
```diff
@@ -4,6 +4,8 @@ spec:
   name: "alicloud"
   path: "cloudquery/alicloud"
   version: "VERSION_SOURCE_ALICLOUD"
+  # adjust this value up or down to change the number of tables that are resolved in parallel
```
hermanschaaf, line 7:
```suggestion
  # adjust this value up or down to change the number of requests that are made in parallel
```

I think the current phrasing might leave the impression that concurrency is a 1:1 mapping to number of tables synced in parallel, so we may want to rephrase it slightly

hermanschaaf (approved):
LGTM, if it were up to me I'd probably drop the comment though (curious users can find it in the docs)

PR #11444, opened 2023-06-12
fix(cli): Propagate destinations errors, reports metrics

erezrokah:
#### Summary

`CloseAndRecv` returns the error reported by the destination (it also returns application level errors, not just protocol errors), so we need to report it. See https://grpc.io/docs/languages/go/basics/#client-side-streaming-rpc-1

This PR also reads the destination metrics and adds them to the summary, but I can split that into another PR.

If this looks good, I'll do it for sync v0 and v1


erezrokah:
Scoped the regression fix to https://github.com/cloudquery/cloudquery/pull/11445, so we should merge that first

yevgenypats (approved):

erezrokah:
Dropping the metrics fix, see https://github.com/cloudquery/plugin-sdk/pull/965

PR #11478, opened 2023-06-13
fix(sqlite): Escape PK constraint name

erezrokah:
#### Summary

Noticed this while working on the Firestore source. Looks like we were escaping everything but the constraint name.


plugins/destination/sqlite/client/migrate.go
```diff
@@ -216,15 +218,15 @@ func (c *Client) createTableIfNotExist(table *schema.Table) error {
...
+			primaryKeys = append(primaryKeys, identifier(col.Name))
 		}
 	}
 
 	if len(primaryKeys) > 0 {
 		// add composite PK constraint on primary key columns
 		sb.WriteString(", CONSTRAINT ")
-		sb.WriteString(table.Name)
-		sb.WriteString("_cqpk PRIMARY KEY (")
+		sb.WriteString(identifier(table.Name + "_cqpk"))
```
erezrokah, line 228: This is the fix

candiduslynx (approved):

PR #11516, opened 2023-06-13
fix: Don't update `UNIQUE` columns (DuckDB contraint)

candiduslynx:


plugins/destination/duckdb/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/cloudquery/plugin-pb-go v1.0.9
 	github.com/cloudquery/plugin-sdk/v3 v3.10.4
 	github.com/google/uuid v1.3.0
-	github.com/marcboeker/go-duckdb v1.3.0
+	github.com/marcboeker/go-duckdb v1.4.0
```
candiduslynx, line 11: rebuilt static libs at least

plugins/destination/duckdb/client/write.go
```diff
@@ -164,15 +152,61 @@ func (c *Client) WriteTableBatch(ctx context.Context, table *schema.Table, recor
...
+	if err != nil {
+		return "", err
+	}
+	defer f.Close() // we don't care here, as the happy-path will actually check the error
+	fileName = f.Name()
+
+	// prep file writer
+	fw, err := pqarrow.NewFileWriter(sc, f,
+		parquet.NewWriterProperties(
+			parquet.WithVersion(parquet.V2_4),
```
disq, line 169: why not `V2_LATEST` (which is `V2_6`?)

candiduslynx: 8e634aca3bce98b3c0c7b3c8135b38d9a5710669

erezrokah (commented):
Thanks for the PR @candiduslynx, do you mind updating the title to reflect what this change fixes so it shows up in the change log?

erezrokah (approved):

Issue #10964, opened 2023-05-25
Unable to use the OracleDB source plugin


niclasgrahm:
Hi, 

The following source config:
```
kind: source
spec:
  name: "oracledb"
  path: "cloudquery/oracledb"
  version: "v1.1.1"
  tables: ["*"]
  destinations: ["postgresql_destination"]
  spec:
    connection_string: "oracle://user:pass@host:1521/service"
```

Produces the following error: 
```
Error: failed to sync v1 source oracledb: rpc error: code = Internal desc = failed to init plugin: failed to create execution client for source plugin oracledb: failed to apply config to tables: tables include a pattern * with no matches.
```

I've verified that the user has access to multible tables in the db.

The database is OracleDb v19.

Relevant log output:
``` 
2023-05-25T13:42:02Z INF Start sync destinations=["file_destination","postgresql_destination"] module=cli source="oracledb (v1.1.1)" sync_time=2023-05-25T13:42:02Z
2023-05-25T13:42:02Z INF started call grpc.code=OK grpc.component=server grpc.method=Init grpc.method_type=unary grpc.service=cloudquery.source.v1.Source grpc.start_time=2023-05-25T15:42:02+02:00 grpc.time_ms=0 module=cli peer.address=@ protocol=grpc
2023-05-25T13:42:03Z ERR finished call grpc.code=Internal grpc.component=server grpc.error="rpc error: code = Internal desc = failed to init plugin: failed to create execution client for source plugin oracledb: failed to apply config to tables: tables include a pattern * with no matches" grpc.method=Init grpc.method_type=unary grpc.service=cloudquery.source.v1.Source grpc.start_time=2023-05-25T15:42:02+02:00 grpc.time_ms=234.306 module=cli peer.address=@ protocol=grpc
2023-05-25T13:42:03Z INF End sync destinations=["file_destination","postgresql_destination"] module=cli source="oracledb (v1.1.1)" sync_time=2023-05-25T13:42:02Z
2023-05-25T13:42:03Z INF Sending sync summary to analyticsv1.cloudquery.io:443 module=cli
2023-05-25T13:42:03Z INF waiting for destination plugin to terminate module=cli
2023-05-25T13:42:03Z INF destination plugin process exited with exit status 1 module=cli
2023-05-25T13:42:03Z INF waiting for destination plugin to terminate module=cli
2023-05-25T13:42:03Z INF destination plugin process exited with exit status 1 module=cli
[...]

PR #11163, opened 2023-06-01
fix: List tables from `ALL_TAB_COLS` instead of `USER_TAB_COLS`

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/10964

See explanation in https://github.com/cloudquery/cloudquery/issues/10964#issuecomment-1570648114

PR #11526, opened 2023-06-13
fix: Change list tables to exclude default namespaces except `USERS`

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/11163.
Hopefully fixes https://github.com/cloudquery/cloudquery/issues/10964#issuecomment-1580630670
See https://www.oracletutorial.com/oracle-administration/oracle-tablespace/:
image

If this causes issue the other option is to allow uses to configure it.


plugins/source/oracledb/client/list_tables.go
```diff
@@ -57,10 +57,14 @@ func (c *Client) updateTableConstraints(ctx context.Context, table *schema.Table
...
 	// Please note we don't use ORDER BY here because it's slower than sorting in memory via Go sort.SliceStable
+	// See more about the default namespaces we exclude in https://www.oracletutorial.com/oracle-administration/oracle-tablespace/
 	query := `
 	SELECT ALL_TAB_COLS.TABLE_NAME, COLUMN_ID, COLUMN_NAME, DATA_TYPE, DATA_LENGTH, NULLABLE FROM ALL_TABLES
 	LEFT JOIN ALL_TAB_COLS ON ALL_TABLES.TABLE_NAME = ALL_TAB_COLS.TABLE_NAME
-	WHERE ALL_TABLES.TABLESPACE_NAME = 'USERS'
+	WHERE ALL_TABLES.TABLESPACE_NAME != 'SYSAUX'
+	 AND ALL_TABLES.TABLESPACE_NAME != 'SYSTEM' 
+	 AND ALL_TABLES.TABLESPACE_NAME != 'TEMP' 
+	 AND ALL_TABLES.TABLESPACE_NAME != 'UNDOTBS1'
```
candiduslynx, line 67: I propose to use `NOT IN` instead
```suggestion
	WHERE ALL_TABLES.TABLESPACE_NAME NOT IN ('SYSAUX', 'SYSTEM', 'TEMP', 'UNDOTBS1')
```

candiduslynx (approved):

PR #11528, opened 2023-06-13
feat(shopify)!: Sync all orders not just open orders

bbernays:
#### Summary

Closes https://github.com/cloudquery/cloudquery/issues/11529

BEGIN_COMMIT_OVERRIDE
fix: List azure_costmanagement_views without scope (https://github.com/cloudquery/cloudquery/pull/11528)
BREAKING-CHANGE: Sync all orders rather than just open orders. This can cause sync times to drastically increase as there are more records to sync.
END_COMMIT_OVERRIDE

By default the orders api only returns `open` orders. This PR changes the behavior to return all orders

https://shopify.dev/docs/api/admin-rest/2023-04/resources/order#get-orders?status=any


erezrokah (approved):

candiduslynx:
Why is it marked as breaking?

disq:
Testing in a big shop with multiple years of orders, will report back.

disq:
60k orders in the last 40 minutes, still going (`min(created_at)` shows Jan 2022 so far). Too much? Maybe `table_options` later?

The Shopify resource suggests:

> Only the last 60 days' worth of orders from a store are accessible from the Order resource by default. If you want to access older orders, then you need to [request access to all orders](https://shopify.dev/api/usage/access-scopes#orders-permissions). If your app is granted access, then you can add the read_all_orders scope to your app along with read_orders or write_orders.

But looks like I'm able to fetch all orders.

bbernays:
@candiduslynx - It is marked as a major change because it can completely break users flow now that it is grabbing so much data

candiduslynx:
I think we might want to reconsider/split this data, so the users may skip it, or add an option at least

bbernays:
> I think we might want to reconsider/split this data, so the users may skip it, or add an option at least

There are a lot of different parameters that we could add. Adding them in a coherent manner is more than minor task. I would suggest we release it as is and if we recieve feedback that users need more control we can expose overrides

candiduslynx:
@bbernays @erezrokah I'm fine with releasing this as is, but could I ask to have the commit override to tell people what huge impact this change does have on the sync time?

bbernays:
> I'm fine with releasing this as is, but could I ask to have the commit override to tell people what huge impact this change does have on the sync time?

@candiduslynx Done

PR #11541, opened 2023-06-14
fix(ci): Remove `go get`

erezrokah:
#### Summary

I'm actually not sure why we have this step, but it can change `go.sum` and fail the `git diff` detection step.
If we want we can add a step that does `go mod tidy` and then check for changes, which will be more accurate for checking changes to `mod` and `sum` files.


erezrokah:
Will open not from a fork 🤦

PR #11542, opened 2023-06-14
fix(ci): Remove `go get` to avoid unwanted changes to `go.sum` files

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/11541 not from a fork.

>I'm actually not sure why we have this step, but it can change `go.sum` and fail the `git diff` detection step.
If we want we can add a step that does `go mod tidy` and then check for changes, which will be more accurate for checking changes to `mod` and `sum` files.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -61,7 +61,6 @@ jobs:
           working-directory: cli
           skip-pkg-cache: true
           skip-build-cache: true
-      - run: go mod tidy
       - name: Build
         run: go build .
       - name: Test
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -61,8 +61,7 @@ jobs:
           working-directory: cli
           skip-pkg-cache: true
           skip-build-cache: true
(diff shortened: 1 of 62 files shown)


hermanschaaf (approved):
Should we perhaps be replacing it with `go mod tidy` instead of removing it?

erezrokah:
> Should we perhaps be replacing it with `go mod tidy` instead of removing it?

Yeah I don't mind. Let me update the PR

erezrokah:
> Should we perhaps be replacing it with `go mod tidy` instead of removing it?

Done

erezrokah:
DuckDB if flaky regardless so I'll force merge once everything else passes

erezrokah:
OK so it seems `go mod tidy` in CI modifies `go.sum` for the AWS source plugin, but it doesn't reproduce locally (with different Go versions).

erezrokah:
I suggest we remove this step for now until we can figure out what's causing the changes, otherwise it's blocking PRs

disq (approved):

PR #11544, opened 2023-06-14
fix(ci): Replace `go get` with `go mod tidy`

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/11542 for the Scaffold CLI. This is a separate PR so it triggers a Scaffold release with a relevant change log


erezrokah:
Closing this for now until we figure this out, see https://github.com/cloudquery/cloudquery/pull/11542#issuecomment-1591511614

PR #11612, opened 2023-06-15
fix: Fix syncing multiple sources to multiple non-overlapping destinations

hermanschaaf:
This fixes the case where different sources are specified in one config, syncing to their own destinations. Previously the code incorrectly assumed that all sources should sync to all destinations.

Tested with the following config:

```yaml
kind: source
spec:
  name: xkcd1
  path: hermanschaaf/xkcd
  version: v2.0.0
  destinations: ["postgresql_schema1"]
  tables: ["*"]
  spec:
---
kind: source
spec:
  name: xkcd2
  path: hermanschaaf/xkcd
  version: v2.0.0
  destinations: ["postgresql_schema2"]
  tables: ["*"]
  spec:
---
kind: destination
spec:
  name: postgresql_schema1
  registry: "github"
  path: cloudquery/postgresql
  version: "v4.2.1"
  spec:
    connection_string: "postgresql://postgres:pass@localhost:5432/cloudquery?sslmode=disable&search_path=test1"
---
kind: destination
spec:
  name: postgresql_schema2
  registry: "github"
  path: cloudquery/postgresql
  version: "v4.2.1"
  spec:
    connection_string: "postgresql://postgres:pass@localhost:5432/cloudquery?sslmode=disable&search_path=test2"
```

Before this synced the XKCD comics to both schemas both times, but now it syncs the comics first to `test1`, then to `test2`. Currently it's difficult to have an automated test with assertions for this kind of setup, but we should work towards that being possible.

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -98,12 +98,18 @@ func sync(cmd *cobra.Command, args []string) error {
 
 	for _, cl := range sourcesClients {
 		maxVersion, err := cl.MaxVersion(ctx)
+		destinationClientsForSource := make([]*manageddestination.Client, 0)
+		for _, destination := range destinationsClients {
+			if slices.Contains(cl.Spec.Destinations, destination.Spec.Name) {
+				destinationClientsForSource = append(destinationClientsForSource, destination)
+			}
+		}
 		if err != nil {
 			return err
 		}
 		switch maxVersion {
 		case 2:
-			for _, destination := range destinationsClients {
+			for _, destination := range destinationClientsForSource {
 				versions, err := destination.Versions(ctx)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


cli/cmd/sync.go
```diff
@@ -98,12 +98,18 @@ func sync(cmd *cobra.Command, args []string) error {
 
 	for _, cl := range sourcesClients {
 		maxVersion, err := cl.MaxVersion(ctx)
+		destinationClientsForSource := make([]*manageddestination.Client, 0)
```
disq, line 101:
```suggestion
		var destinationClientsForSource []*manageddestination.Client
```
For the sake of simplicity?

disq (approved):
LGTM with nit

erezrokah (approved):

candiduslynx (approved):

PR #11621, opened 2023-06-15
fix!: List `azure_costmanagement_views` without scope

### LINKED ISSUE #9859: bug(azure): Cost management views only fetch non-private views

### Current Behavior

In `azure_costmanagement_views`, the fetched views don't contain the private views. This seems to because of the code using `svc.NewListByScopePager("subscriptions/"+cl.SubscriptionId, nil)`, whereas if it had used `NewListPager()` it will fetch the users views (but this time not any of the shared views - verified with api call)

### Expected Behavior

Fetch all views

### CloudQuery version

azure v7.1.0

### Additional Context

Reported on [Discord](https://discord.com/channels/872925471417962546/1085831052620201984)

### Pull request (optional)

candiduslynx:
Closes #9859

BEGIN_COMMIT_OVERRIDE
fix: List `azure_costmanagement_views` without scope (#11621)
BREAKING-CHANGE: Add `subscription_id` to `azure_costmanagement_views` primary key (#11621)
END_COMMIT_OVERRIDE

### DIFF
--- a/plugins/source/azure/resources/services/costmanagement/views.go
+++ b/plugins/source/azure/resources/services/costmanagement/views.go
@@ -20,20 +20,18 @@ func Views() *schema.Table {
 			transformers.WithNameTransformer(client.ETagNameTransformer),
 			transformers.WithPrimaryKeys("ID"),
 		),
-		Columns: schema.ColumnList{client.SubscriptionID},
-		Relations: []*schema.Table{
-			view_queries(),
-		},
+		Columns:   schema.ColumnList{client.SubscriptionIDPK},
+		Relations: schema.Tables{view_queries()},
 	}
 }
 
-func fetchViews(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+func fetchViews(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
 	cl := meta.(*client.Client)
 	svc, err := armcostmanagement.NewViewsClient(cl.Creds, cl.Options)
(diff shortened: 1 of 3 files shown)


plugins/source/azure/resources/services/costmanagement/views.go
```diff
@@ -20,20 +20,18 @@ func Views() *schema.Table {
...
 
-func fetchViews(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+func fetchViews(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
 	cl := meta.(*client.Client)
 	svc, err := armcostmanagement.NewViewsClient(cl.Creds, cl.Options)
 	if err != nil {
 		return err
 	}
-	pager := svc.NewListByScopePager("subscriptions/"+cl.SubscriptionId, nil)
+	pager := svc.NewListPager(nil)
```
disq, line 34: If no subscription then you need to update multiplexer to suit.

disq: Also doing this fails to get any of the 'shared views' as it was verified by api call in the original issue. I think shared views use case would be more important in large teams.

candiduslynx: Yeah, I just implemented this to be discussed during PR review

disq:
Maybe `azure_costmanagement_views` and `azure_costmanagement_private_views`? But then we need a sub-table for each for the view queries (that contain the actual data)

PR #11626, opened 2023-06-15
fix(aws): Upgrade all AWS SDK to latest version

bbernays:
#### Summary

Upgrades all aws dependencies... and fixes a breaking change that was introduced by [this pr](https://github.com/aws/aws-sdk-go-v2/commit/fe59fdbb04e9978e6237e4e3b44d320224dade24)


erezrokah:
Hi @bbernays 👋 We usually do an update per PR for the sake of the change log. I triggered some manually (out of the Monday schedule) via https://github.com/cloudquery/cloudquery/issues/760 to see what's up.

Also this should be a `fix` since we'd like it to trigger a release

PR #11649, opened 2023-06-16
Fix example query in analysing-gcp-cost-with-bigquery-and-cq.mdx

dfredell:
#### Summary

The SCHEMA for gcp_compute_disks.users is a REPEATED so you have to check for an empty array.

I found this solution via https://stackoverflow.com/a/67506844

When running the command in the doc, we get no results
image

Updating it to use the proposal in the MR gets data back

image

Investigating the data type of the `users` column
image


hermanschaaf (approved):
👍

PR #11650, opened 2023-06-16
feat(aws): Update to SDK V4

yevgenypats:
Still some more fine tuning to do but tests seem to pass. Im also going to test it shortly with real data.


plugins/source/aws/client/testing.go
```diff
@@ -24,77 +19,39 @@ type TestOptions struct {
...
-		},
-		newTestExecutionClient)
-	p.SetLogger(l)
-	source.TestPluginSync(t, p, specs.Source{
-		Name:         "dev",
-		Path:         "cloudquery/dev",
-		Version:      version,
-		Tables:       []string{table.Name},
-		Destinations: []string{"mock-destination"},
-	}, source.WithTestPluginAdditionalValidators(validateTagStructure))
```
hermanschaaf, line 66: I think we lost the additional validator for tag structure in the transition here, and I'd say it's quite important we keep it in one form or another

plugins/source/aws/client/client.go
```diff
@@ -88,9 +86,9 @@ func (s *ServicesManager) InitServicesForPartitionAccount(partition, accountId s
 	s.services[partition][accountId].Regions = funk.UniqString(append(s.services[partition][accountId].Regions, svcs.Regions...))
 }
 
-func NewAwsClient(logger zerolog.Logger, b backend.Backend, spec *Spec) Client {
+func NewAwsClient(logger zerolog.Logger, spec *Spec) Client {
 	return Client{
-		Backend: b,
+		// Backend: b,
```
hermanschaaf, line 91: If this is getting initialized somewhere else we can probably remove this comment now

plugins/source/aws/client/testing.go
```diff
@@ -24,77 +19,39 @@ type TestOptions struct {
...
-					}
-					value := resource.Get(column.Name)
-					var tags map[string]any
-					if err := json.Unmarshal(value.(*scalar.JSON).Value, &tags); err != nil {
-						t.Fatalf("failed to unmarshal tags column %s: %v", value.(*scalar.JSON).Value, err)
-					}
-				}
-			}
-		})
+	inserts := messages.InsertMessage()
```
hermanschaaf, line 50: Shouldn't we also be checking that the table migration messages are sent before the insert messages? Or is that being tested somewhere else? (Feels like we probably don't have to test that in every plugin, but we don't have to test it somewhere :) )

plugins/source/aws/resources/plugin/client.go
```diff
@@ -0,0 +1,55 @@
...
+	c.client, err = client.Configure(ctx, logger, spec)
+	if err != nil {
+		return nil, err
+	}
+
+	c.scheduler = scheduler.NewScheduler(
+		c.client,
+		scheduler.WithConcurrency(spec.Concurrency),
+		scheduler.WithLogger(logger),
+		scheduler.WithDeterministicCQId(true),
```
hermanschaaf, line 40: We probably don't want to _always_ set the CQ ID deterministically, or do we?

plugins/source/aws/resources/plugin/plugin.go
```diff
@@ -91,12 +91,11 @@ func titleTransformer(table *schema.Table) string {
...
-	return source.NewPlugin(
+func AWS() *plugin.Plugin {
+	return plugin.NewPlugin(
 		"aws",
 		Version,
-		tables(),
-		client.Configure,
-		source.WithTitleTransformer(titleTransformer),
+		New,
+		// plugin.WithTitleTransformer(titleTransformer),
```
hermanschaaf, line 99: I think once we run doc generation some titles will be incorrect without this transformer, unless it was moved somewhere and I missed it

plugins/source/aws/main.go
```diff
@@ -1,12 +1,14 @@
...
+
+	internalPlugin "github.com/cloudquery/cloudquery/plugins/source/aws/resources/plugin"
+	"github.com/cloudquery/plugin-sdk/v4/serve"
 )
 
 const sentryDSN = "https://6c6b72bc946844cb8471f49eba485cde@o1396617.ingest.sentry.io/6747636"
 
 func main() {
-	serve.Source(plugin.AWS(), serve.WithSourceSentryDSN(sentryDSN))
+	serve.Plugin(internalPlugin.AWS(), serve.WithPluginSentryDSN(sentryDSN)).Serve(context.Background())
```
disq, line 13: We seem to be ignoring `error` result from Serve, which gets caught by the linter.

plugins/source/aws/resources/plugin/client.go
bbernays: Is there a good reason to not make the default `true` and enable people to disable the functionality? If we don't have (or hear) a good reason to ever want a random CQID in the future we could remove the option

candiduslynx:
@yevgenypats I think this one needs an overhaul per updates to sdk

plugins/source/aws/client/testing.go
```diff
@@ -24,77 +20,48 @@ type TestOptions struct {
...
-		result = append(result, table)
-		result = append(result, extractTables(table.Relations)...)
+	sc := scheduler.NewScheduler(&c, scheduler.WithLogger(l))
+	messages, err := sc.SyncAll(context.Background(), tables)
+	if err != nil {
+		t.Fatal(err)
 	}
-	return result
+
+	records := filterInserts(messages).GetRecordsForTable(table)
```
disq, line 51:
```suggestion
	records := messages.GetInserts().GetRecordsForTable(table)
```

plugins/source/aws/resources/plugin/client.go
yevgenypats: I don't think we should change a default in this PR as already many changes.

plugins/source/aws/resources/plugin/plugin.go
yevgenypats: Fixed

PR #11655, opened 2023-06-18
refactor(cli): Update to managedplugin and decouple CLI config from proto

yevgenypats:
blocked by: https://github.com/cloudquery/plugin-pb-go/pull/16/files

Adds the following changes as pre-requirements that we can roll now and before migrating plugins to SDK V3:

- merge `managedplugin` code (from `managedsourceplugin` and `manageddestinationplugin` - now located at `plugin-pb-go`)
- decouple CLI config/spec from the one that is being sent over the wire to the SDK in protocol V1/V2. This is crucial requirement to be able to support multiple SDK versions. Also, this is a requirement for us to be able to introduce deprecation to config options and being able to print necessary warnings in the CLI.


cli/cmd/migrate.go
```diff
@@ -64,39 +79,51 @@ func migrate(cmd *cobra.Command, args []string) error {
...
-		maxVersion, err := cl.MaxVersion(ctx)
+	for _, source := range sources {
+		cl := managedSourceClients.ClientByName(source.Name)
+		versions, err := cl.Versions(ctx)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get source versions: %w", err)
+		}
+		maxVersion := findMaxVersion(versions)
+		if maxVersion >= 3 {
```
hermanschaaf, line 99: Shouldn't this be `if maxVersion <= 3`? Or am I missing something 🤔

hermanschaaf: @erezrokah Yeah I think this PR is not intended to handle SDK v3 yet; is that right @yevgenypats ? We may just want to update the title and description of this PR to make it clear this is mostly a refactor. Or are there any user-facing changes?

yevgenypats: Yeah I updated the title to `refactor` and removed `v4`. The only user facing change is that this CLI doesn't support an old version for sources which I dropped (`source_v0`). Once we move to single plugin protocol we can also have a schedule of how long we support a given protocol so we have a phase where we introduce a new protocol, add deprecation notice and then remove an old protocol support from the CLI. In any case users still can use an old version of the CLI.

cli/internal/specs/v0/spec_reader.go
```diff
@@ -0,0 +1,246 @@
...
+				return fmt.Errorf("source %s references unknown destination %s", source.Name, destination)
+			}
+			destinationToSourceKey := fmt.Sprintf("%s-%s", destination, source.Path)
+			if destinationSourceMap[destinationToSourceKey] == "" {
+				destinationSourceMap[destinationToSourceKey] = source.Path + "@" + source.Version
+			} else if destinationSourceMap[destinationToSourceKey] != source.Path+"@"+source.Version {
+				return fmt.Errorf("destination %s is used by multiple sources %s with different versions", destination, source.Path)
+			}
+		}
+	}
```
hermanschaaf, line 154: Is this validation step a new addition? I don't quite understand what it's doing / why we can't allow different source versions can't use the same destination

hermanschaaf: Looks like it's not new in any case; but I'm still curious why it exists 😄

PR #11657, opened 2023-06-19
feat(aws): Add PolicyStatus to WrappedBucket

tsutsun17:
#### Summary
Thank you for implementing this amazing project.

I added ~`is_public_by_policy`~ `policy_status` to `WrappedBucket`.
Though AWS shows the criteria ([criteria](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-block-public-access.html#access-control-block-public-access-policy-status)), It is difficult for me to conclude whether a s3 bucket which uses bucket policy is public or not.
so I suggest the way using `GetBucketPolicyStatus` API in this PR.

### DIFF
--- a/plugins/source/aws/resources/services/s3/models/s3.go
+++ b/plugins/source/aws/resources/services/s3/models/s3.go
@@ -29,6 +29,7 @@ type WrappedBucket struct {
 	LoggingTargetBucket   *string
 	LoggingTargetPrefix   *string
 	Policy                map[string]any
+	IsPublicByPolicy      bool
 	VersioningStatus      types.BucketVersioningStatus
 	VersioningMfaDelete   types.MFADeleteStatus
 	BlockPublicAcls       bool
--- a/plugins/source/aws/resources/services/s3/models/s3.go
+++ b/plugins/source/aws/resources/services/s3/models/s3.go
@@ -29,7 +29,7 @@ type WrappedBucket struct {
 	LoggingTargetBucket   *string
 	LoggingTargetPrefix   *string
 	Policy                map[string]any
-	IsPublicByPolicy      bool
+	PolicyStatus          *types.PolicyStatus
 	VersioningStatus      types.BucketVersioningStatus
 	VersioningMfaDelete   types.MFADeleteStatus
 	BlockPublicAcls       bool
--- a/website/tables/aws/aws_s3_buckets.md
+++ b/website/tables/aws/aws_s3_buckets.md
@@ -31,7 +31,7 @@ The following tables depend on aws_s3_buckets:
 |logging_target_bucket|`utf8`|
 |logging_target_prefix|`utf8`|
 |policy|`json`|
-|is_public_by_policy|`bool`|
+|policy_status|`json`|
 |versioning_status|`utf8`|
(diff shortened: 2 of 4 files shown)


plugins/source/aws/resources/services/s3/models/s3.go
```diff
@@ -29,6 +29,7 @@ type WrappedBucket struct {
 	LoggingTargetBucket   *string
 	LoggingTargetPrefix   *string
 	Policy                map[string]any
+	IsPublicByPolicy      bool
```
hermanschaaf, line 32: Could we please change this to contain the entire `PolicyStatus` object, so it gets stored as a JSON column called `policy_status`?

I'm aware this may seems like an odd request. This will mean queries will need to be written as `policy_status->is_public` instead of `is_public_by_policy`, but it conforms better to our current standards, which are:
 - stick as closely as possible to the original AWS SDK method / CLI command names
 - store nested objects as JSON columns rather than unnesting them

Both these are in place so that it's easier to map from the AWS API to CloudQuery table columns.

The S3 table happens to be one of our oldest tables, so for backwards-compatibility reasons we haven't updated all columns to these standards yet, and what you've done here is totally reasonable when looking at the other columns in the table. But I think for setting the precedent, new columns we should try and follow the rules above, which would mean storing this additional information as a JSON column. One other advantage of this approach is also that, if AWS were ever to add new fields to that API response, they will automatically be included when we upgrade the AWS SDK version without requiring a manual update here.

hermanschaaf (changes requested):
Hi @tsutsun17, thanks so much for this PR. What you did here looks good, but I have one request if you don't mind, outlined in the comment (and I tried to provide the reasoning).

plugins/source/aws/resources/services/s3/models/s3.go
tsutsun17: Thank you for review and I got it!

> if AWS were ever to add new fields to that API response, they will automatically be included when we upgrade the AWS SDK version without requiring a manual update here.

It means I should not use new struct, but use `map[string]any`, right?
so I'll change it as below. 
```go
  PolicyStatus map[string]any <- the result of `json.Unmarshal`
```

hermanschaaf: @tsutsun17 I think you should be able to do:

```
PolicyStatus          *types.PolicyStatus
```

and set with:

```
resource.PolicyStatus = policyStatusOutput.PolicyStatus
```

Worth testing this though of course :)

website/tables/aws/aws_s3_buckets.md
```diff
@@ -31,6 +31,7 @@ The following tables depend on aws_s3_buckets:
 |logging_target_bucket|`utf8`|
 |logging_target_prefix|`utf8`|
 |policy|`json`|
+|policy_status|`json`|
```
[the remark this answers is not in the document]
hermanschaaf: @tsutsun17 You don't have to, this is a limitation of our currently generated documentation that we are planning to address in an automated way :)

hermanschaaf (approved):

Issue #11550, opened 2023-06-14
bug(ci): DuckDB tests are flaky


erezrokah:
The CI fails occasionally for the DuckDB source.
We should fix it as it blocks dependencies updates PRs (like SDK) for all plugins as we group those.

erezrokah:
Haven't seen this in a while or either the move to v4 or the large runners solved it. Will re-open if I see it again

PR #11660, opened 2023-06-19
test(duckdb): Add debug logs

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/11550. Adds some debug logging to the DuckDB tests.
Hopefully we can trace back which query doesn't work. I initially thought the delete stale removes too much as the error is:
```
after overwrite expected 1 resource, got 0
```

but always got 1 row affected in the delete operation so 🤷  (the test is flaky on my machine too).
I would like to let the CI run a bit with the debug logs to see if it sheds more information on this issue.


plugins/destination/duckdb/client/client.go
```diff
@@ -73,6 +73,14 @@ func (c *Client) Metrics() destination.Metrics {
 }
 
 func (c *Client) exec(ctx context.Context, query string, args ...any) error {
-	_, err := c.db.ExecContext(ctx, query, args...)
+	r, err := c.db.ExecContext(ctx, query, args...)
+	if c.spec.Version == "development" {
```
erezrokah, line 77: Per a better way to communicate between the test framework and the plugin. Also only temporarily

erezrokah:
> Why only display this information during `development`? Why not always log this information?

That's a good point. Mostly since I don't have another use case for logging this information and it is quite verbose on the logs

bbernays:
> > Why only display this information during `development`? Why not always log this information?
> 
> That's a good point. Mostly since I don't have another use case for logging this information and it is quite verbose on the logs

Fair, to me debug logs are supposed to be verbose and if we were to unify the logging statement to:
```
c.logger.Debug().Str("query", query).Any("values", args).Int64("rowsAffected", rowsAffected).Err(rowsErr).Msg("exec query")
```

Then any future issues would be able to be debugged (or at least reproducible) outside of CQ by a user without having to dive into the code

erezrokah:
> Then any future issues would be able to be debugged (or at least reproducible) outside of CQ by a user without having to dive into the code

Agree, but I tend to add these once we need to, otherwise existing users of the DuckDB destination running in debug log might experience noise they don't need. But again I don't have a strong opinion here, my plan is to remove these logs once I'm done fixing our flaky CI, so maybe we'll decide to add them to all users. It does have a small impact on all users as it needs to call `RowsAffected` that can fail, but I don't think that's pretty minor

PR #11662, opened 2023-06-19
feat(postgresql)!: Migrate Postgres destination to SDK v4

hermanschaaf:
Upgrades the PostgreSQL destination to plugin SDK V4. This changes the plugin to use the new streaming interface, allowing migrations to happen more than once during a sync and for PostgreSQL to be used as a state backend.

While there shouldn't be anything breaking in this upgrade, I am marking this as a breaking change because it is a big internal refactor, and despite our best efforts there may be some things that we missed during testing.


plugins/destination/postgresql/main.go
```diff
@@ -1,17 +1,20 @@
...
 const (
 	sentryDSN = "https://19d1257d36854a51b17c06614e76dc2d@o1396617.ingest.sentry.io/4503896817336320"
 )
 
 func main() {
-	p := destination.NewPlugin("postgresql", plugin.Version, client.New, destination.WithDefaultBatchSize(1000))
-	serve.Destination(p, serve.WithDestinationSentryDSN(sentryDSN))
+	p := pluginSDK.NewPlugin("postgresql", plugin.Version, client.New)
+	server := serve.Plugin(p, serve.WithPluginSentryDSN(sentryDSN))
+	server.Serve(context.Background())
```
disq, line 19: We seem to be ignoring `error` result from Serve which gets caught in the linter.

hermanschaaf:
Seems like GitHub is lagging; I pushed a commit to fix the merge conflicts a few minutes ago already but it's not showing up yet.

yevgenypats (approved):

PR #11667, opened 2023-06-19
fix(aws)!: Remove ARN from `aws_cloudfront_cache_policies`

bbernays:
#### Summary

Cannot find any reference or documentation to an ARN for cache policies... Changed the PK from arn to resource id  + account id

### DIFF
--- a/plugins/source/aws/resources/services/cloudfront/cache_policies.go
+++ b/plugins/source/aws/resources/services/cloudfront/cache_policies.go
@@ -21,16 +21,11 @@ func CachePolicies() *schema.Table {
 		Multiplex:   client.ServiceAccountRegionMultiplexer(tableName, "cloudfront"),
 		Transform:   transformers.TransformWithStruct(&types.CachePolicySummary{}),
 		Columns: []schema.Column{
-			client.DefaultAccountIDColumn(false),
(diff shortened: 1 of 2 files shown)


disq:
As a side effect this would also break the [resources view](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/views/resources.sql) for this type. Is this a customer issue?

bbernays:
> As a side effect this would also break the [resources view](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/views/resources.sql) for this type. Is this a customer issue?

This is a field we made up so it is not like users would be able to do anything with this field...

disq:
> > As a side effect this would also break the [resources view](https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/views/resources.sql?rgh-link-date=2023-06-19T21%3A12%3A43Z) for this type. Is this a customer issue?
> 
> This is a field we made up so it is not like users would be able to do anything with this field...

Yeah, _but_ if there's no arn column a resource is not listed in the resources view. Granted this is the shortcoming of the resources view (lack of a better formed SQL) I was just wondering if this was an issue with any of our users or a general cleanup of sorts.

PR #11670, opened 2023-06-19
feat(aws): Add Support for Cloudtrail Channels

bbernays:

### DIFF
--- a/plugins/source/aws/resources/services/cloudtrail/channels.go
+++ b/plugins/source/aws/resources/services/cloudtrail/channels.go
@@ -1,0 +1,70 @@
+package cloudtrail
+
+import (
+	"context"
+
+	"github.com/apache/arrow/go/v13/arrow"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudtrail"
+	"github.com/aws/aws-sdk-go-v2/service/cloudtrail/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v3/schema"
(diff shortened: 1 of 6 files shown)


yevgenypats (approved):

plugins/source/aws/resources/services/cloudtrail/channels.go
```diff
@@ -0,0 +1,70 @@
...
+func Channels() *schema.Table {
+	tableName := "aws_cloudtrail_channels"
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_GetChannel.html`,
+		Resolver:            fetchCloudtrailChannels,
+		PreResourceResolver: getChannel,
+		Multiplex:           client.ServiceAccountRegionMultiplexer(tableName, "cloudtrail"),
+		Transform:           transformers.TransformWithStruct(&cloudtrail.GetChannelOutput{}, transformers.WithSkipFields("ResponseMetadata")),
+		IsIncremental:       true,
```
hermanschaaf, line 24:
```suggestion
		IsIncremental:       false,
```

I don't think this table is incremental?

hermanschaaf (changes requested):
I think it's incorrectly marked as incremental, otherwise LGTM

plugins/source/aws/resources/services/cloudtrail/channels.go
disq, line 24:
```suggestion
```
Good catch but how about just removing the line :P

hermanschaaf (approved):

PR #11680, opened 2023-06-20
feat: Provide better error message on mismatch between plugin and CLI protocol support

hermanschaaf:
This is a small change to the protocol version handling logic to make it more robust in the face of differing support between the CLI and source plugins.

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -20,7 +20,7 @@ cloudquery sync ./directory
 # Sync resources from directories and files
 cloudquery sync ./directory ./aws.yml ./pg.yml
 `
-	unknownFieldErrorPrefix = "code = InvalidArgument desc = failed to decode spec: json: unknown field "
+	maxSupportedProtocolVersion = 2
 )
 
 func NewCmdSync() *cobra.Command {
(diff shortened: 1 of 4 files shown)


cli/cmd/sync.go
```diff
@@ -36,17 +36,41 @@ func NewCmdSync() *cobra.Command {
...
-func findMaxVersion(versions []int) int {
-	max := -1
-	if len(versions) == 0 {
-		return max
+// findMaxCommonVersion finds the max common version between protocol versions supported by a plugin and those supported by the CLI.
+// If all plugin versions are lower than min CLI supported version, it returns -1.
+// If all plugin versions are higher than max CLI supported version, it returns -2.
+// In this way it is possible tell whether the source or the CLI needs to be updated:
+// if -1, the source needs to be updated or the CLI downgraded;
+// if -2, the CLI needs to be updated or the source downgraded.
```
hermanschaaf, line 44: I wanted to have a way to differentiate between source needing upgrade or CLI needing upgrade, and -1 and -2 were the simplest thing I could come up with. We could also introduce constants maybe, but this is a private function so I feel a comment explaining it is enough.

disq (approved):

erezrokah (approved):

hermanschaaf:
This was merged as part of https://github.com/cloudquery/cloudquery/pull/11683

PR #11685, opened 2023-06-20
chore: Homebrew gitignore update (second attempt)

hermanschaaf:
Seems like I got the syntax wrong in the first attempt 😮‍💨


disq (approved):

erezrokah (approved):

plugins/source/homebrew/.gitignore
```diff
@@ -1,3 +1,3 @@
 test
 dist
-./homebrew
\ No newline at end of file
+/homebrew
```
candiduslynx, line 3:
```suggestion
homebrew
```

hermanschaaf: No, there is an internal directory called `homebrew` that I don't want to ignore

PR #11686, opened 2023-06-20
fix: Use ServiceAccountRegion multiplexer for aws_availability_zones

hermanschaaf:
Without this fix, AZs are only listed for a single region.


disq (approved):

bbernays (approved):

getstek:
What I have noticed is that if you specify no region in the spec, it gives back 6 records for `ap-northeast-1` region. 

image

If you do specify regions in the spec though, it returns just the regions for the first region in the list.

`  spec:
    regions:
      - eu-west-1
      - us-east-1
    use_paid_apis: true
`
image

hermanschaaf:
Thanks @getstek, I think that makes sense considering what the bug was. Have you been able to try the fix (available in aws v19.0.0) to confirm if it solves the issue for you?

getstek:
I tried but no luck. There are no errors reported and it finished super quick.

```
[root@mc-host cloudquery-config]# cloudquery --log-level debug sync az.yml postgresql.yml
Loading spec(s) from az.yml, postgresql.yml
Starting migration for: aws (v19.0.0) -> [postgresql]
Starting sync for: aws (v19.0.0) -> [postgresql]
Sync completed successfully. Resources: 116, Errors: 0, Panics: 0, Time: 2s
[root@mc-host cloudquery-config]# vi az.yml
[root@mc-host cloudquery-config]# cloudquery --log-level debug sync az.yml postgresql.yml
Loading spec(s) from az.yml, postgresql.yml
Starting migration for: aws (v19.0.0) -> [postgresql]
Starting sync for: aws (v19.0.0) -> [postgresql]
Sync completed successfully. Resources: 116, Errors: 0, Panics: 0, Time: 3s
```
config file

```
kind: source
spec:
  ## Required. name of the plugin to use.
  ## This should be unique if you have number of aws plugins.
  name: "aws"
  path: "cloudquery/aws"
  version: "v19.0.0"
  tables: ["aws_availability_zones"]

  destinations: ["postgresql"]

  spec:
```
I also deleted the table completely before running the re-sync. Now it is just giving me records for 1 region but a different one this time :) 

image

getstek:
Hi Herman, 

It is better, I used the latest version and there are more records in the table. But I noticed not all the regions are reflected, so I can't query all AZ's. 

image

There are 31 regions in total, so for example I want to lookup AZ's for me-central-1. 

Config file specifies no regions

```
kind: source
spec:
  ## Required. name of the plugin to use.
  ## This should be unique if you have number of aws plugins.
  name: "aws"
  path: "cloudquery/aws"
  version: "v19.0.0"
  tables: ["aws_availability_zones"]

  destinations: ["postgresql"]

  spec:
```

bbernays:
@getstek - For this table CQ will only sync the regions that are specified in the spec... In this case you are specifying all `opted-in` regions, which by default is only the first 17 regions released... All regions introduced after March 20, 2019 require a manual opt-in...

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html

PR #11694, opened 2023-06-21
feat(aws): Support for `id` column in `aws_resources` view

### LINKED ISSUE #11676: bug(aws): Resources view doesn't include data from tables without the `arn` column

### Current Behavior

The resources view includes a list of all resources across all AWS tables, provided they have both `arn` and either `account_id` or `request_account_id` columns. Not all resources in AWS have ARNs so this becomes an issue.

The ARN is also used in extracting AWS partition/service name/resource type and making it available as separate columns in the view.

### Expected Behavior

Resources view should include _all_ resources regarding the AWS plugin. If there is no data available for some, just return nulls in place of those. `id` column is available (along with `account_id` and perhaps `region`) if there is no `arn`.

```yaml
```

### Steps To Reproduce

_No response_

```yaml
```

### CloudQuery version

latest

### Additional Context

_No response_

### Pull request (optional)

disq:
Fixes #11676

Now `arn` can be `NULL` in the view, in that case `id` should be there. Both `id` and `arn` can be non-null if both are in the table. `arn`-derived fields (`partition`, `service` and `type`) will be `NULL` if `arn` field is missing from the table.

### DIFF
--- a/plugins/source/aws/views/resources.sql
+++ b/plugins/source/aws/views/resources.sql
@@ -4,43 +4,56 @@ DO $$
 DECLARE
     tbl TEXT;
     strSQL TEXT = '';
+    arnField TEXT = '';
 BEGIN
--- iterate over every table in our information_schema that has an `arn` column available
+-- iterate over every table in our information_schema that has an `arn` or 'id' columns available
 FOR tbl IN
     SELECT DISTINCT table_name
     FROM information_schema.columns
     WHERE table_name LIKE 'aws_%s' and COLUMN_NAME IN ('account_id', 'request_account_id')
     INTERSECT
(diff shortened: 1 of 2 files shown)


erezrokah (approved):

bbernays:
This query will now include items from `aws_cloudformation_stack_set_operations`, `aws_apprunner_operations` and `aws_glue_job_runs` which are not really resources... Is this behavior alright? Maybe we generate a more general view that just normalizes all resources into a single view... Instead of using arn or id we could stringify all PK columns into a single value?

erezrokah:
Closing as stale, please re-open if not the case

PR #11696, opened 2023-06-21
feat(mysql)!: Update to SDK V4

erezrokah:
#### Summary

This is based on the MemDB plugin from https://github.com/cloudquery/plugin-sdk/pull/915

Fixes https://github.com/cloudquery/cloudquery/issues/11755

BEGIN_COMMIT_OVERRIDE
feat!: Upgrades the mysql source plugin to use plugin-sdk v4. This version does not contain any user-facing breaking changes, but because it is now using CloudQuery gRPC protocol v3, it does require use of a destination plugin that also supports protocol v3. All recent destination plugin versions support this.
END_COMMIT_OVERRIDE


plugins/source/mysql/resources/plugin/plugin_test.go
```diff
@@ -218,30 +166,55 @@ func TestPlugin(t *testing.T) {
...
+		t.Fatalf("expected 2 resource, got %d", len(actualRecords))
 	}
-	expectedStrings := getExpectedStrings(t, testTable, data)
-	actualStrings := getActualStrings(resources)
-	sortResults(testTable, actualStrings)
 
-	require.Equal(t, expectedStrings, actualStrings)
+	sortResults(testTable, actualRecords)
+
+	for recordIndex, expectedRecord := range expectedRecords {
```
erezrokah, line 200: Notice we switched from string comparison to value comparison as everything is arrow types now

plugins/source/mysql/client/transformer.go
```diff
@@ -1,11 +1,148 @@
...
+	"time"
+
 	"github.com/apache/arrow/go/v13/arrow"
 	"github.com/apache/arrow/go/v13/arrow/array"
-	"github.com/cloudquery/plugin-sdk/v3/types"
+	"github.com/cloudquery/plugin-sdk/v4/types"
+	"github.com/google/uuid"
 )
 
+func reverseTransform(bldr array.Builder, val any) error {
```
erezrokah, line 13: I'll probably do a follow and try to use the `scalar` package for this

yevgenypats (approved):

PR #11700, opened 2023-06-21
feat(s3): Upgrade to SDK v4

disq:
Due to flushing issues, plugintestsuite tests are skipped (still all there) and plugin specific test cases were added instead.


plugins/destination/s3/client/streamingbatch.go
```diff
@@ -0,0 +1,408 @@
...
+	migrateTableLock     *sync.Mutex
+	migrateTableMessages []*message.MigrateTable
+	deleteStaleLock      *sync.Mutex
+	deleteStaleMessages  []*message.DeleteStale
+
+	logger         zerolog.Logger
+	batchTimeout   time.Duration
+	batchSize      int
+	batchSizeBytes int
+}
```
candiduslynx, line 49: if you're going to use `StreamingBatchWriter` only by pointer there's no need to define sync primitives as pointers inside (define as value & don't alloc explicitly):
```suggestion
type StreamingBatchWriter struct {
	client           StreamingBatchWriterClient
	workers          map[string]*worker
	workersLock      sync.RWMutex
	workersWaitGroup sync.WaitGroup

	migrateTableLock     sync.Mutex
	migrateTableMessages []*message.MigrateTable
	deleteStaleLock      sync.Mutex
	deleteStaleMessages  []*message.DeleteStale

	logger         zerolog.Logger
	batchTimeout   time.Duration
	batchSize      int
	batchSizeBytes int
}
```

this allows for the init:
```go
c := &StreamingBatchWriter{
	client:           client,
	workers:          make(map[string]*worker),
	logger:           zerolog.Nop(),
	batchTimeout:     defaultBatchTimeoutSeconds * time.Second,
	batchSize:        defaultBatchSize,
	batchSizeBytes:   defaultBatchSizeBytes,
}
```

plugins/destination/s3/main.go
```diff
@@ -1,17 +1,28 @@
...
 )
 
 func main() {
-	p := destination.NewPlugin("s3", plugin.Version, client.New, destination.WithManagedWriter())
-	serve.Destination(p, serve.WithDestinationSentryDSN(sentryDSN))
+	p := plugin.NewPlugin("s3", internalPlugin.Version, client.New)
+	serveHelper(serve.Plugin(p, serve.WithPluginSentryDSN(sentryDSN), serve.WithDestinationV0V1Server()).Serve)
+}
+
+func serveHelper(f func(context.Context) error) {
```
candiduslynx, line 23: IMO this should just be in `main`

disq: it will probably removed, either moved to some sdk package or not accepted so I kept it easy to remove. :)

yevgenypats (approved):

PR #11701, opened 2023-06-21
feat(firestore)!: Update to SDK V4

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/11696. ~~Still need to test this~~

FIxes https://github.com/cloudquery/cloudquery/issues/11756

BEGIN_COMMIT_OVERRIDE
feat!: Upgrades the firestore source plugin to use plugin-sdk v4. This version does not contain any user-facing breaking changes, but because it is now using CloudQuery gRPC protocol v3, it does require use of a destination plugin that also supports protocol v3. All recent destination plugin versions support this.
END_COMMIT_OVERRIDE


plugins/source/firestore/client/sync.go
```diff
@@ -93,14 +93,12 @@ func (c *Client) syncTable(ctx context.Context, table *schema.Table, res chan<-
 	return err
 }
 
-func (c *Client) syncTables(ctx context.Context, res chan<- *schema.Resource) error {
-	eg, ctx := errgroup.WithContext(ctx)
-	eg.SetLimit(len(c.Tables))
```
erezrokah, line 98: No point of setting the limit to the number of tables as that's exactly the number of Go Routines that are created in the loop below

PR #11711, opened 2023-06-22
feat(oracledb)!: Update to SDK V4

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/11696

Fixes https://github.com/cloudquery/cloudquery/issues/11757

BEGIN_COMMIT_OVERRIDE
feat!: Upgrades the oracledb source plugin to use plugin-sdk v4. This version does not contain any user-facing breaking changes, but because it is now using CloudQuery gRPC protocol v3, it does require use of a destination plugin that also supports protocol v3. All recent destination plugin versions support this.
END_COMMIT_OVERRIDE


plugins/source/oracledb/client/types.go
```diff
@@ -49,14 +49,16 @@ func SQLType(t arrow.DataType) string {
 	case *arrow.Int8Type, *arrow.Int16Type, *arrow.Int32Type, *arrow.Int64Type,
 		*arrow.Uint8Type, *arrow.Uint16Type, *arrow.Uint32Type, *arrow.Uint64Type:
 		return "NUMBER(38)"
-	case *arrow.Float16Type, *arrow.Float32Type, *arrow.Float64Type:
+	case *arrow.Float16Type, *arrow.Float32Type:
```
erezrokah, line 52: This code is only used in the tests.
Previously the OracleDB types were not accurate enough and data was lost during insert. It was not discovered in the tests, due to doing string comparison

plugins/source/oracledb/client/transformer.go
```diff
@@ -50,7 +188,9 @@ func GetValue(arr arrow.Array, i int) (any, error) {
 		if err != nil {
 			return nil, err
 		}
-		return toTime(a.Value(i)), nil
+		t := toTime(a.Value(i))
```
erezrokah, line 191: This code was also used only in the tests during insert.
Apparently if you use Go `time.Time` type to insert data using the Oracle driver it allows only up to seconds precision. To fix this we need to use the built in `go_ora.TimeStamp`, see:

With `time.Time` we reach here:
https://github.com/sijms/go-ora/blob/41ca433ee475e64869448c27e6e20ad038613ae4/v2/parameter_encode.go#L93
https://github.com/sijms/go-ora/blob/41ca433ee475e64869448c27e6e20ad038613ae4/v2/converters/type_conversion.go#L1117

And with `go_ora.TimeStamp`:
https://github.com/sijms/go-ora/blob/41ca433ee475e64869448c27e6e20ad038613ae4/v2/parameter_encode.go#L116
https://github.com/sijms/go-ora/blob/41ca433ee475e64869448c27e6e20ad038613ae4/v2/parameter_encode.go#L116

plugins/source/oracledb/client/transformer.go
```diff
@@ -1,11 +1,149 @@
...
+
 	"github.com/apache/arrow/go/v13/arrow"
 	"github.com/apache/arrow/go/v13/arrow/array"
-	"github.com/cloudquery/plugin-sdk/v3/types"
+	"github.com/cloudquery/plugin-sdk/v4/types"
+	"github.com/google/uuid"
+	go_ora "github.com/sijms/go-ora/v2"
 )
 
+func reverseTransform(bldr array.Builder, val any) error {
```
erezrokah, line 14: I'll probably do a follow and try to use the `scalar` package for this

plugins/source/oracledb/client/client.go
```diff
@@ -3,23 +3,22 @@ package client
...
 	// Import the oracle driver
 	_ "github.com/sijms/go-ora/v2"
 )
 
 type Client struct {
-	logger      zerolog.Logger
-	metrics     *source.Metrics
-	Tables      schema.Tables
-	db          *sql.DB
-	Concurrency uint64
```
erezrokah, line 22: We use to read this from the top level spec, so this could be considered a breaking change as now we don't account for it

erezrokah:
Marked this as blocked since it's missing the concurrency option

plugins/source/oracledb/client/client.go
erezrokah: https://github.com/cloudquery/cloudquery/pull/11766 should handle this

erezrokah:
> Marked this as blocked since it's missing the concurrency option

https://github.com/cloudquery/cloudquery/pull/11766 should unblock this, but this update is still a breaking change

yevgenypats (approved):

PR #11719, opened 2023-06-23
feat(postgresql)!: Update to SDK V4

erezrokah:
Closes https://github.com/cloudquery/cloudquery/issues/11758

#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/11696.
~~I had to comment change some of the tests due to https://github.com/cloudquery/plugin-sdk/pull/982 and https://github.com/cloudquery/plugin-sdk/pull/915#issuecomment-1602751482, so not ready to merge~~

BEGIN_COMMIT_OVERRIDE
feat!: Upgrades the postgresql source plugin to use plugin-sdk v4. This version does not contain any user-facing breaking changes, but because it is now using CloudQuery gRPC protocol v3, it does require use of a destination plugin that also supports protocol v3. All recent destination plugin versions support this.

feat!: To enable CDC in this version you'll need to use the `cdc_id` configuration string property, instead of the `cdc` boolean one. Please refer to the [docs](https://www.cloudquery.io/docs/plugins/sources/postgresql/overview) for more information 
END_COMMIT_OVERRIDE


plugins/source/postgresql/client/client.go
```diff
@@ -16,15 +16,15 @@ import (
...
 	logger              zerolog.Logger
-	spec                specs.Source
-	metrics             *source.Metrics
 	pluginSpec          Spec
 	currentDatabaseName string
 	currentSchemaName   string
 	pgType              pgType
-	Tables              schema.Tables
+	tables              schema.Tables
+	cdcId               string
```
erezrokah, line 27: Since plugins don't have access to the top level spec we can't use the source name as the replication name like before. I added a plugin level configuration for now, but we should figure this out before merging this PR as this is a breaking change

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -77,51 +81,51 @@ func getTestCases(serialValue int64) []testCase {
...
-		panic(err)
-	}
-	timeMillisecond := scalar.Time{
-		Int: scalar.Int{
-			BitWidth: 32,
-		},
-		Unit: arrow.Millisecond,
-	}
-	err = timeMillisecond.Set("04:05:06.789")
+	err = inet.Set("192.168.0.0/24")
```
erezrokah, line 84: We can change this back to `192.168.0.1/24` after https://github.com/cloudquery/plugin-sdk/pull/982 is released

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -77,51 +81,51 @@ func getTestCases(serialValue int64) []testCase {
...
 	if err != nil {
 		panic(err)
 	}
+	// timeMicrosecond := scalar.Time{
```
erezrokah, line 88: All the commented scalars in this file are since the v4 SDK is missing those

erezrokah:
Marked this as blocked due to the breaking changes in the spec (adding a `CDCId`)

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -103,16 +107,16 @@ func getTestCases(serialValue int64) []testCase {
...
-		Type: &arrow.TimestampType{
-			Unit:     arrow.Millisecond,
-			TimeZone: "UTC",
-		},
-	}
-	err = timestampMillisecond.Set("1999-01-08 04:05:06.789")
-	if err != nil {
-		panic(err)
-	}
+	// timestampMillisecond := scalar.Timestamp{
```
erezrokah, line 110: Needs https://github.com/cloudquery/plugin-sdk/pull/986

erezrokah:
> Marked this as blocked due to the breaking changes in the spec (adding a `CDCId`)

Related to https://github.com/cloudquery/cloudquery/pull/11766. We'll need to chose the right naming for this plugin level configuration and document it

plugins/source/postgresql/resources/plugin/plugin_test.go
disq: Can be re-added now?

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -137,7 +141,7 @@ func getTestCases(serialValue int64) []testCase {
 		{"circle", "<(1,2),3>", &scalar.String{Value: "<(1,2),3>", Valid: true}},
 		{"date", "1999-01-08", &scalar.Date32{Value: 10599, Valid: true}},
 		{"double precision", 1.1, &scalar.Float{Value: 1.1, Valid: true, BitWidth: 64}},
-		{"inet", "192.168.0.1/24", &inet},
+		{"inet", "192.168.0.0/24", &inet},
```
disq, line 144: We can also revert this, should work normally

plugins/source/postgresql/resources/plugin/plugin_test.go
erezrokah: Good catch, done in https://github.com/cloudquery/cloudquery/pull/11719/commits/482c8f264089fcd8931fb8cdf0f3d8a50c645eb7

plugins/source/postgresql/resources/plugin/plugin_test.go
erezrokah: Good catch, done in https://github.com/cloudquery/cloudquery/pull/11719/commits/482c8f264089fcd8931fb8cdf0f3d8a50c645eb7

plugins/source/postgresql/client/spec.go
```diff
@@ -3,12 +3,5 @@ package client
...
 	PgxLogLevel      LogLevel `json:"pgx_log_level,omitempty"`
-	CDC              bool     `json:"cdc,omitempty"`
-}
-
-func (*Spec) SetDefaults() {
-}
-
-func (*Spec) Validate() error {
-	return nil
+	CDCId            string   `json:"cdc_id,omitempty"`
```
erezrokah, line 6: Updated the docs to reflect this

disq (approved):

PR #11727, opened 2023-06-23
fix: Github source pagination

nicl:
#### Summary

It appears that Github pagination is not working correctly. Specifically, we are missing out on the final page of results.

For the recommended way to perform pagination, see: https://github.com/google/go-github/#pagination.

It is fair to say that the Google Github library's behaviour here is a bit confusing and easy to get wrong!

*I am surprised that, if correct, this issue was not detected sooner - so please do take my findings with caution and confirm them for yourselves.*

**How diagnosed:**

*Note, I have only confirmed this impacts the `github_team_repositories` table but I assume from the code that this issue is actually much more widespread.*

The issue initially manifested as suspicious data. Running the following query on a Postgres instance we use as a destination:

`select team_id, count(team_id) from github_team_repositories group by team_id;`

yielded some surprisingly round numbers (all divisible by 100 other than results with < 100):

Screenshot 2023-06-23 at 15 31 15

Following these changes, the numbers look more 'natural'.


candiduslynx (approved):
Great find, thanks!

PR #11736, opened 2023-06-25
feat(sqlite): Migrate to SDK V4

yevgenypats:


plugins/destination/sqlite/client/client_test.go
```diff
@@ -1,32 +1,32 @@
...
-
 func TestPlugin(t *testing.T) {
-	destination.PluginTestSuiteRunner(t,
-		func() *destination.Plugin {
-			return destination.NewPlugin("sqlite", "development", New)
-		},
-		specs.Destination{
-			Spec: &Spec{
-				ConnectionString: ":memory:",
+	ctx := context.Background()
```
candiduslynx, line 12:
```suggestion
	ctx := context.TODO()
```
as `context.TODO()` exists for this exact reason

plugins/destination/sqlite/client/migrate.go
```diff
@@ -131,14 +130,14 @@ func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
 }
 
 // This is the responsibility of the CLI of the client to lock before running migration
-func (c *Client) Migrate(ctx context.Context, tables schema.Tables) error {
+func (c *Client) migrate(_ context.Context, force bool, tables schema.Tables) error {
```
candiduslynx, line 133: `ctx` should be utilized in the calls to database (so `c.createTableIfNotExist` should accept ctx, etc).
You should switch to [`sql.DB.ExecContext`](https://pkg.go.dev/database/sql#DB.ExecContext) call for proper `ctx` utilization

plugins/destination/sqlite/client/read.go
```diff
@@ -148,13 +148,14 @@ func reverseTransform(sc *arrow.Schema, values []any) (arrow.Record, error) {
 	return rec, nil
 }
 
-func (c *Client) Read(ctx context.Context, table *schema.Table, sourceName string, res chan<- arrow.Record) error {
+// Read(ctx context.Context, table *schema.Table, res chan<- arrow.Record) error
```
candiduslynx, line 151:
```suggestion
```

plugins/destination/sqlite/go.mod
```diff
@@ -5,14 +5,16 @@ go 1.19
...
-	github.com/cloudquery/plugin-sdk/v3 v3.10.6
+	github.com/cloudquery/plugin-sdk/v4 v4.0.0
 	github.com/mattn/go-sqlite3 v1.14.16
 	github.com/rs/zerolog v1.29.1
 )
 
 // TODO: remove once all updates are merged
 replace github.com/apache/arrow/go/v13 => github.com/cloudquery/arrow/go/v13 v13.0.0-20230624001442-f06019270fad
 
+replace github.com/cloudquery/plugin-sdk/v4 => ../../../../plugin-sdk
```
candiduslynx, line 16: should be removed

plugins/destination/sqlite/client/write.go
```diff
@@ -6,29 +6,52 @@ import (
...
+		case *message.MigrateTable:
+			if err := c.migrate(ctx, options.MigrateForce, schema.Tables{m.Table}); err != nil {
+				return fmt.Errorf("failed to process MigrateTable message: %w", err)
 			}
+		case *message.DeleteStale:
+			if err := c.deleteStale(ctx, m.Table, m.SourceName, m.SyncTime); err != nil {
+				return fmt.Errorf("failed to process DeleteStale message: %w", err)
+			}
+		case *message.Insert:
+			if err := c.insertMessage(m); err != nil {
```
candiduslynx, line 26: utilize ctx

plugins/destination/sqlite/client/write.go
```diff
@@ -6,29 +6,52 @@ import (
...
+	sc := m.Record.Schema()
+	var sql string
+	if len(table.PrimaryKeys()) == 0 {
+		sql = c.insert(sc)
+	} else {
+		sql = c.upsert(sc)
+	}
+	vals := transformRecord(m.Record)
+	for _, v := range vals {
+		if _, err := c.db.Exec(sql, v...); err != nil {
```
candiduslynx, line 48:
```suggestion
func (c *Client) insertMessage(ctx context.Context, m *message.Insert) error {
	table := m.GetTable()
	sc := m.Record.Schema()
	var sql string
	if len(table.PrimaryKeys()) == 0 {
		sql = c.insert(sc)
	} else {
		sql = c.upsert(sc)
	}
	vals := transformRecord(m.Record)
	for _, v := range vals {
		if _, err := c.db.ExecContext(ctx, sql, v...); err != nil {
```

candiduslynx (changes requested):
mostly fine, ctx needs to be used in db calls, though

plugins/destination/sqlite/client/client_test.go
yevgenypats: not sure I understand what is the reasoning here. moreover in a golang `Test` doens't provide a context so it's not a TODO it's background (which is the same thing just that TODO indicates we want to change in the future while we don't)

PR #11763, opened 2023-06-26
fix(ci): Paginate over all PR cache keys when listing caches to delete

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/11674.
I also opened a feature request to the `actions-cache` GitHub CLI extension here https://github.com/actions/gh-actions-cache/issues/75


hermanschaaf (approved):

erezrokah:
Also opened https://github.com/github/docs/pull/26252 and https://github.com/github/docs/issues/26251 as the example in misleading a bit

PR #11766, opened 2023-06-26
fix: CLI v4: issue config deprecation warnings, pass options to plugin-level spec

hermanschaaf:
This updates the V4 CLI to log deprecation warnings for config fields that are moving to the plugin spec level, and forward the settings to the plugin if they are still in use.


cli/cmd/sync_v3.go
```diff
@@ -50,10 +50,19 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
 		}
 		if destinationSpecs[i].WriteMode == specs.WriteModeAppend {
 			opts = append(opts, transformer.WithRemovePKs())
+		} else if destinationSpecs[i].PKMode == specs.PKModeCQID {
```
disq, line 53:
```suggestion
		}
		if destinationSpecs[i].PKMode == specs.PKModeCQID {
```

hermanschaaf: Hmm, I don't think this is equivalent. If write mode is append, we should remove all PKs and not add any back (not even `_cq_id`). If write mode is not append, we should check the pk mode and replace all PKs with CQID only if pk_mode == cq_id. This seems right to me, and it's what the logic was doing before

disq: OK makes sense now that you've explained it in detail :)

cli/internal/transformer/transformer.go
```diff
@@ -62,6 +69,9 @@ func (t *RecordTransformer) TransformSchema(sc *arrow.Schema) *arrow.Schema {
 		if _, ok := mdMap["cq:extension:primary_key"]; ok && t.removePks {
 			mdMap["cq:extension:primary_key"] = "false"
 		}
+		if field.Name == "_cq_id" && t.cqIDPrimaryKey {
```
disq, line 72:
```suggestion
		if field.Name == schema.CqIDColumn.Name && t.cqIDPrimaryKey {
```

cli/internal/transformer/transformer.go
```diff
@@ -62,6 +69,9 @@ func (t *RecordTransformer) TransformSchema(sc *arrow.Schema) *arrow.Schema {
 		if _, ok := mdMap["cq:extension:primary_key"]; ok && t.removePks {
 			mdMap["cq:extension:primary_key"] = "false"
```
disq, line 70: Not related to the functionality of this fix but why not delete the key from the map instead?
```suggestion
			delete(mdMap, "cq:extension:primary_key")
```

cli/internal/transformer/transformer.go
hermanschaaf: We are not supposed to import the plugin-sdk in the CLI anymore, it's only for plugins to import 🙈 

But we can create a constant in this package maybe

cli/cmd/specs.go
```diff
@@ -48,16 +48,16 @@ func CLISourceSpecToPbSpec(spec specs.Source) pbSpecs.Source {
 		Version:             spec.Version,
 		Path:                spec.Path,
 		Registry:            CLIRegistryToPbRegistry(spec.Registry),
-		Concurrency:         spec.Concurrency,
-		TableConcurrency:    spec.TableConcurrency,
-		ResourceConcurrency: spec.ResourceConcurrency,
+		Concurrency:         spec.Concurrency,         // nolint:staticcheck // allow use of deprecated options here for backwards-compatibility
```
disq, line 51: Maybe arrange the deprecated options down in the struct together... then the `allow use of...` comment can be at the top of the 'block'.

cli/internal/specs/v0/spec_reader.go
```diff
@@ -177,10 +190,12 @@ func (r *SpecReader) GetDestinationNamesForSource(name string) []string {
 
 func NewSpecReader(paths []string) (*SpecReader, error) {
 	reader := &SpecReader{
-		sourcesMap:      make(map[string]*Source),
-		destinationsMap: make(map[string]*Destination),
-		Sources:         make([]*Source, 0),
-		Destinations:    make([]*Destination, 0),
+		sourcesMap:             make(map[string]*Source),
+		destinationsMap:        make(map[string]*Destination),
+		Sources:                make([]*Source, 0),
```
disq, line 195: No need to `make` the slices.

disq (approved):
just a few nits for readability.

cli/internal/specs/v0/spec_reader.go
hermanschaaf: It was like this before, but we can clean it up in the next iteration

PR #11775, opened 2023-06-26
feat(mysql): Update to SDK V4

yevgenypats:
Still WIP


plugins/destination/mysql/client/client.go
```diff
@@ -52,13 +56,9 @@ func New(ctx context.Context, logger zerolog.Logger, spec specs.Destination) (de
 	db.SetMaxOpenConns(10)
 	db.SetMaxIdleConns(10)
 
-	return &Client{logger: logger.With().Str("module", "mysql").Logger(), db: db, spec: spec, mySQLSpec: mySQLSpec}, nil
+	return c, nil
 }
 
 func (c *Client) Close(ctx context.Context) error {
 	return c.db.Close()
```
disq, line 63:
```suggestion
	if err := c.writer.Close(ctx); err != nil {
		return err
	}
	return c.db.Close()
```

disq (changes requested):

plugins/destination/mysql/client/client.go
candiduslynx: we still need to add `_ = c.db.Close()` before `return err`, though

PR #11776, opened 2023-06-26
feat(aws): Add Support for Network firewall resources `Firewalls` and `TLSInspectionConfigurations`

bbernays:


plugins/source/aws/resources/services/networkfirewall/firewalls.go
```diff
@@ -0,0 +1,83 @@
...
+func Firewalls() *schema.Table {
+	tableName := "aws_networkfirewall_firewalls"
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/network-firewall/latest/APIReference/API_DescribeFirewall.html`,
+		Resolver:            fetchFirewalls,
+		PreResourceResolver: getFirewall,
+		Multiplex:           client.ServiceAccountRegionMultiplexer(tableName, "network-firewall"),
+		Transform: transformers.TransformWithStruct(
+			&models.FirewallWrapper{},
```
disq, line 25: Why not use `DescribeFirewallOutput` instead (could ignore `UpdateToken` if it's not wanted?)

plugins/source/aws/resources/services/networkfirewall/tls_inspection_configurations.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/cloudquery/plugin-sdk/v3/schema"
+	"github.com/cloudquery/plugin-sdk/v3/transformers"
+	sdkTypes "github.com/cloudquery/plugin-sdk/v3/types"
+)
+
+func TLSInspectionConfigurations() *schema.Table {
+	tableName := "aws_networkfirewall_tls_inspection_configurations"
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/network-firewall/latest/APIReference/API_RuleGroup.html`,
```
disq, line 20: Wrong link
```suggestion
		Description:         `https://docs.aws.amazon.com/network-firewall/latest/APIReference/API_DescribeTLSInspectionConfiguration.html`,
```

plugins/source/aws/resources/services/networkfirewall/tls_inspection_configurations.go
```diff
@@ -0,0 +1,83 @@
...
+		Resolver:            fetchTLSInspectionConfigurations,
+		PreResourceResolver: getTLSInspectionConfigurations,
+		Multiplex:           client.ServiceAccountRegionMultiplexer(tableName, "network-firewall"),
+		Transform: transformers.TransformWithStruct(
+			&models.TLSInspectionConfigurationWrapper{},
```
disq, line 25: again why not use `DescribeTLSInspectionConfigurationOutput` instead?

plugins/source/aws/resources/services/networkfirewall/firewalls.go
bbernays: I was trying to follow the same format as the 2 existing `networkfirewall` resources...

disq: If we're not merging data from 2 separate calls I think it's better to just use the existing structs. If the struct gets updated with new fields they will be explicitly shown when we update the dependency.

hermanschaaf (approved):
Agree we can probably get away without the extra model here, but since it's an internal refactor we can do later I don't see that as a blocker, LGTM

disq (approved):

plugins/source/aws/resources/services/networkfirewall/firewalls.go
bbernays: I agree that we lose the ability to automatically add new fields to this based on upgrading the SDK but:
1. I would rather maintain consistency within this service as we already have 2 resources that follow this `wrapper` model
2. We can't unwrap the `DescribeFirewallOutput` as the fields are not nested structs... So a new option for the transformer would be needed

PR #11783, opened 2023-06-27
feat(aws): Update all AWS dependencies

bbernays:


hermanschaaf (approved):

hermanschaaf:
That command works, though btw you should also be able to update it by ticking the appropriate boxes here https://github.com/cloudquery/cloudquery/issues/760 and triggering a dependency update run

bbernays:
That works as well, but as far as I am aware each package will be updated in its own PR... which means that to update 60+ dependencies would take days

hermanschaaf:
@bbernays I thought it would batch them together, but I could be wrong

bbernays:
> @bbernays I thought it would batch them together, but I could be wrong

@hermanschaaf - I could also be wrong, lets bring it up with erez when he is back (hopefully he doesn't see this while he is out)

PR #11794, opened 2023-06-28
feat(clickhouse): Migrate to SDK v4

### LINKED ISSUE #11747: Migrate `plugins/destination/clickhouse` to `github.com/cloudquery/plugin-sdk/v4`

candiduslynx:
Closes #11747


plugins/destination/clickhouse/client/spec.go
```diff
@@ -47,6 +53,18 @@ func (s *Spec) SetDefaults() {
 	if s.Engine == nil {
 		s.Engine = queries.DefaultEngine()
 	}
+
+	if s.BatchSize == 0 {
+		s.BatchSize = writers.DefaultBatchSize
```
disq, line 58: Use plugin's own defaults (we're removing the exported writer values here for this reason)

candiduslynx: 752b1df8ef92ab558baad9e4b57b425e4bfb4dbb

disq (approved):

plugins/destination/clickhouse/client/spec.go
```diff
@@ -47,6 +52,21 @@ func (s *Spec) SetDefaults() {
...
 		s.Engine = queries.DefaultEngine()
 	}
+
+	if s.BatchSize == 0 {
+		const defaultBatchSize = 10_000 // 10K
+		s.BatchSize = defaultBatchSize
+	}
+
+	if s.BatchSizeBytes == 0 {
+		const defaultBatchSizeBytes = 50 << 20 // 50 MiB
```
yevgenypats, line 62: Should it be 4MB ? otherwise memory can blow up as it's per table

yevgenypats (approved):

plugins/destination/clickhouse/client/client.go
```diff
@@ -20,30 +19,26 @@ type Client struct {
...
+var _ batchwriter.Client = (*Client)(nil)
 
-func (*Client) DeleteStale(context.Context, schema.Tables, string, time.Time) error {
-	return errors.New("DeleteStale is not implemented")
+func (*Client) DeleteStale(context.Context, []*message.WriteDeleteStale) error {
+	return plugin.ErrNotImplemented
 }
 
 func (c *Client) Close(context.Context) error {
 	return c.conn.Close()
```
disq, line 34:
```suggestion
	if err := c.writer.Close(ctx); err != nil {
		return err
	}
	return c.conn.Close()
```

plugins/destination/clickhouse/client/spec.go
candiduslynx: I can make it 5 MiB (or 4, if you insist)

plugins/destination/clickhouse/client/client.go
candiduslynx: but I think it should always issue conn close, too.
If we were on go 1.20 I'd use `errors.Join`, but I'd make a workaround for now

candiduslynx: 64e09937466a2cce483f05bc8e5da59badd933d0

plugins/destination/clickhouse/client/spec.go
candiduslynx: 5c0e1150d394f877575bb0920d46118afcbf13b6

PR #11795, opened 2023-06-28
fix(aws): Fix Timestream Regional issue

bbernays:
#### Summary

Workaround for https://github.com/aws/aws-sdk-go-v2/issues/2163

### DIFF
--- a/plugins/source/aws/resources/services/timestream/databases.go
+++ b/plugins/source/aws/resources/services/timestream/databases.go
@@ -49,8 +49,9 @@ func fetchTimestreamDatabases(ctx context.Context, meta schema.ClientMeta, _ *sc
 	svc := cl.Services().Timestreamwrite
 	// This should be removed once https://github.com/aws/aws-sdk-go-v2/issues/2163 is fixed
 	if cl.AWSConfig != nil && cl.AWSConfig.Region != cl.Region {
-		cl.AWSConfig.Region = cl.Region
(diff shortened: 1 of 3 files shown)


plugins/source/aws/resources/services/timestream/databases.go
```diff
@@ -46,8 +46,14 @@ func Databases() *schema.Table {
 
 func fetchTimestreamDatabases(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
 	cl := meta.(*client.Client)
+	svc := cl.Services().Timestreamwrite
+	// This should be removed once https://github.com/aws/aws-sdk-go-v2/issues/2163 is fixed
+	if cl.AWSConfig != nil && cl.AWSConfig.Region != cl.Region {
+		cl.AWSConfig.Region = cl.Region
```
hermanschaaf, line 52: Since `cl.AWSConfig` is a pointer (`*aws.Config`), couldn't we have a data race here when `fetchTimestreamDatabases` gets called by clients due to multiplexing? To be safe I would probably make a copy of `cl.AwsConfig` before changing the property and passing it in

bbernays: Good call. Updated accordingly

hermanschaaf (approved):
Seems like a pretty decent workaround given the constraints we have right now

PR #11816, opened 2023-06-29
fix(aws): Add test to ensure documentation is unique for each table

bbernays:
#### Summary

This PR adds a test to ensure that all tables have a unique description (or have been added to the list of table descriptions to ignore)

This will ensure that if/ when a developer copies a table from another resource developers will be required to update the documentation field or the tests won't pass


plugins/source/aws/resources/plugin/plugin_test.go
```diff
@@ -14,3 +14,31 @@ func TestAWS(t *testing.T) {
 		t.Errorf("Name() = %q, want %q", name, "aws")
 	}
 }
+
+// This test ensures that all tables have a unique description.
+func TestAWSTableDescriptions(t *testing.T) {
+	descriptions := make(map[string]string)
+	p := AWS()
+	tables := p.Tables()
+	for _, table := range tables {
```
bbernays, line 23: Need to update to handle nested tables
```suggestion
	for _, table := range tables.FlattenTables() {
```

hermanschaaf (approved):
Nice catch!

PR #11818, opened 2023-06-29
feat(aws): Add support for Route53resolver service

bbernays:
#### Summary

I put this new service under a new service called `route53resolver` rather than just `route53` to keep in line with AWS package name... This differs from how we handle `route53domain` package where we put it under the `route53` service and do not indicate it is in a separate package... for example if `aws_route53_domains` and `aws_route53_operations` were to follow the same convention as `route53resolver` it would be `aws_route53domains_domains` and `aws_route53domains_operations`


hermanschaaf:
I haven't reviewed yet, but see some errors in the build that we may want to address first:

```
Error: client/services.go:323:37: undefined: services.Route53resolverClient
```

hermanschaaf:
Also I think using `route53resolver` as the service name is the right call, since it's also what the AWS CLI uses.

website/tables/aws/aws_route53resolver_firewall_configs.md
```diff
@@ -0,0 +1,22 @@
+# Table: aws_route53resolver_firewall_configs
+
+This table shows data for Route53resolver Firewall Configs.
```
bbernays, line 3: Need an exception for this service name

hermanschaaf (approved):

PR #11822, opened 2023-06-30
fix(cli): Fix sync in protov3

yevgenypats:
This requires additional update in plugin-pb-go. For delete stale we actually don't want the schema but just the name of the table

### DIFF
--- a/cli/cmd/sync_v3.go
+++ b/cli/cmd/sync_v3.go
@@ -255,7 +255,7 @@ func deleteStale(client plugin.Plugin_WriteClient, tables map[string]bool, sourc
 				Delete: &plugin.Write_MessageDeleteStale{
 					SourceName: sourceName,
 					SyncTime:   timestamppb.New(syncTime),
-					TableName: tableName,
+					TableName:  tableName,
 				},
 			},
 		}); err != nil {
--- a/cli/cmd/sync_v3.go
+++ b/cli/cmd/sync_v3.go
@@ -250,14 +250,12 @@ func tableNameFromSchema(sc *arrow.Schema) string {
 
 func deleteStale(client plugin.Plugin_WriteClient, tables map[string]bool, sourceName string, syncTime time.Time) error {
 	for tableName := range tables {
-		// TODO remove this print
-		fmt.Println("Deleting stale records from table", tableName)
 		if err := client.Send(&plugin.Write_Request{
 			Message: &plugin.Write_Request_Delete{
 				Delete: &plugin.Write_MessageDeleteStale{
 					SourceName: sourceName,
 					SyncTime:   timestamppb.New(syncTime),
-					// TODO change table
+					TableName: tableName,
 				},
 			},
 		}); err != nil {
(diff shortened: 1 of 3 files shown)


cli/cmd/sync_v3.go
```diff
@@ -249,3 +242,28 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
 	fmt.Printf("%s. Resources: %d, Errors: %d, Warnings: %d, Time: %s\n", msg, totalResources, totals.Errors, totals.Warnings, syncTimeTook.Truncate(time.Second).String())
 	return nil
 }
+
+func tableNameFromSchema(sc *arrow.Schema) string {
+	tableName, _ := sc.Metadata().GetValue("cq:table_name")
```
disq, line 247: Is there a const we can use here? Maybe move `cq:*` consts to `plugin-pb-go` if we don't want to import the SDK?

yevgenypats: yeah we can move this later to `plugin-pb-go` just didn't want to do it now.

cli/cmd/sync_v3.go
```diff
@@ -249,3 +242,28 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
...
+	return tableName
+}
+
+func deleteStale(client plugin.Plugin_WriteClient, tables map[string]bool, sourceName string, syncTime time.Time) error {
+	for tableName := range tables {
+		// TODO remove this print
+		fmt.Println("Deleting stale records from table", tableName)
```
hermanschaaf, line 254: Should we remove the print now?

cli/cmd/sync_v3.go
```diff
@@ -249,3 +242,28 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
...
+		if err := client.Send(&plugin.Write_Request{
+			Message: &plugin.Write_Request_Delete{
+				Delete: &plugin.Write_MessageDeleteStale{
+					SourceName: sourceName,
+					SyncTime:   timestamppb.New(syncTime),
+					// TODO change table
```
hermanschaaf, line 260: We still need to  add the table name here, right?

cli/cmd/sync_v3.go
yevgenypats: yeah going to remove once proto-buf PR is merged

cli/cmd/sync_v3.go
```diff
@@ -249,3 +242,26 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
...
+}
+
+func deleteStale(client plugin.Plugin_WriteClient, tables map[string]bool, sourceName string, syncTime time.Time) error {
+	for tableName := range tables {
+		if err := client.Send(&plugin.Write_Request{
+			Message: &plugin.Write_Request_Delete{
+				Delete: &plugin.Write_MessageDeleteStale{
+					SourceName: sourceName,
+					SyncTime:   timestamppb.New(syncTime),
+					TableName: tableName,
```
hermanschaaf, line 258: It looks like this file is not gofmt-ed

hermanschaaf (approved):

PR #11839, opened 2023-06-30
fix(neo4j): Update to SDK v4.4.0-rc1

yevgenypats:
Update: https://github.com/cloudquery/plugin-sdk/releases/tag/v4.4.0-rc1

### DIFF
--- a/plugins/destination/neo4j/client/deletestale.go
+++ b/plugins/destination/neo4j/client/deletestale.go
@@ -11,11 +11,11 @@ import (
 
 const deleteCypher = "MATCH (n:%s) WHERE n._cq_source_name = $cq_source_name AND n._cq_sync_time < $cq_sync_time DETACH DELETE n"
 
-func (c *Client) DeleteStale(ctx context.Context, msgs []*message.WriteDeleteStale) error {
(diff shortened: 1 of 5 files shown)


candiduslynx (approved):

candiduslynx:
@yevgenypats the tests are failing, though

plugins/destination/neo4j/client/client_test.go
```diff
@@ -54,6 +54,9 @@ func TestPlugin(t *testing.T) {
 		plugin.WithTestIgnoreNullsInLists(),
 		plugin.WithTestDataOptions(schema.TestSourceOptions{
 			TimePrecision: time.Microsecond,
+			SkipMaps: 		true,
```
candiduslynx, line 57: why are we skipping those now? shouldn't we fix the issue instead?

candiduslynx (changes requested):
why won't we fix the writing instead?

yevgenypats:
> why won't we fix the writing instead?

MAP and Structs work now but had to skip lists due to list of list not working. 

It needs either more granular support on SDK site for test case generation or better support for neo4j. It's not really related to this upgrade so Im going to skip and opened a low-priority issue - https://github.com/cloudquery/cloudquery/issues/11849

PR #11855, opened 2023-06-30
feat!: Upgrade Hacker News plugin to SDK v4

hermanschaaf:
Upgrades Hacker News plugin to SDK v4, which allows any destination plugin to be used as a backend. 

This will only work with recent CLI versions and destinations that are also upgraded to V4. 

The backend should now be configured like this:

```
kind: source
spec:
  name: "hackernews"
  version: "vX.Y.Z"
  path: "cloudquery/hackernews"
  tables: 
    ["*"]
  destinations: 
    - "postgresql"
  backend_options:
    table_name: "cq_state_hackernews"
    connection: "@@plugins.postgresql.connection"
  spec:
    item_concurrency: 100
    start_time: "2023-07-01T00:00:00Z"
---
kind: destination
spec:
  name: "postgresql"
  version: "vX.Y.Z"
  registry: "github"
  path: "cloudquery/postgresql"
  spec:
    connection_string: "postgresql://postgres:pass@localhost:5432/cloudquery?sslmode=disable"
```


cli/cmd/sync_v3.go
```diff
@@ -81,6 +73,29 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
 			return err
 		}
 	}
+	// replace @destination.connection with the actual GRPC connection string from the client
+	// NOTE: if this becomes a stable feature, it can move out of sync_v3 and into sync.go
+	replacementValues := make([]specs.ReplacementValue, len(destinationSpecs))
```
yevgenypats, line 78: I think the search and replace should be on string and not on go structs.

plugins/source/hackernews/client/testing.go
```diff
@@ -7,52 +7,72 @@ import (
...
+		HackerNews: builder(t),
+		// Backend:    mockBackend, TODO(v4)
+		maxRetries: 0,
+		backoff:    1 * time.Millisecond,
+	}
+	tables := schema.Tables{table}
+	if err := transformers.TransformTables(tables); err != nil {
+		t.Fatal(err)
+	}
+	sc := scheduler.NewScheduler(schedulerClient, scheduler.WithLogger(l))
```
yevgenypats, line 40: unrelated to this PR but seems we will need to move the `schedulerClient` to `scheduler.Sync` because backend is a per `sync` option

plugins/source/hackernews/resources/plugin/client.go
```diff
@@ -0,0 +1,125 @@
...
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallRecvMsgSize(maxMsgSize),
+			grpc.MaxCallSendMsgSize(maxMsgSize),
+		),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial grpc source plugin at %s: %w", config.Backend.Connection, err)
+	}
+	stateClient, err := state.NewClient(ctx, conn, config.Backend.Table)
```
yevgenypats, line 108: we prob want to move this to `sync(` as `table` can be different depending on the sync

cli/cmd/sync_v3.go
yevgenypats: but maybe not crucial right now as we will are going to change the config anyway

hermanschaaf:
This is now ready and working, however we need to release all V4-compatible destinations first, before releasing this one.

yevgenypats (approved):

yevgenypats:
Looks great! Merging (will hold-off the release)

PR #11869, opened 2023-07-01
fix: Change placeholder logic

yevgenypats:
This updates a bit the logic of find/replace variables.

This still needs a few fixes to the HN plugins but I suggest we might want to do those separately ( I did test that HN is working with a few minor fixes and this SDK https://github.com/cloudquery/plugin-sdk/pull/1040).

Also, we might want to move the backend from `init` to `Sync`.


yevgenypats:
Closing in favor of https://github.com/cloudquery/cloudquery/pull/11870

PR #11886, opened 2023-07-03
fix: Close writers

disq:


disq:
Will reopen in main repo

PR #11889, opened 2023-07-03
feat(mssql): Migrate to SDK v4

### LINKED ISSUE #11842: Migrate `plugins/destination/mssql` to `github.com/cloudquery/plugin-sdk/v4`

candiduslynx:
Closes #11842


plugins/destination/mssql/client/spec.go
```diff
@@ -29,6 +34,21 @@ func (s *Spec) SetDefaults() {
 	if len(s.AuthMode) == 0 {
 		s.AuthMode = AuthModeMS
 	}
+
+	if s.BatchSize == 0 {
+		const defaultBatchSize = 1000 // 1K
```
disq, line 39: I think the inline `const` approach is unnecessary, I would just set `s.BatchSize = 1000`

candiduslynx: yeah, it's for documenting purposes only

yevgenypats (approved):

PR #11898, opened 2023-07-03
chore(cli): Use large runners

yevgenypats:


hermanschaaf:
@yevgenypats what's the summary / reasoning behind the change here?

yevgenypats:
> @yevgenypats what's the summary / reasoning behind the change here?

Im still playing with it but I think we should upgrade our CI to use large runners where possible. they are quite slow now for now reason and I think we can pay for the increased speed

.github/workflows/cli.yml
```diff
@@ -87,9 +67,9 @@ jobs:
           path: |
             ~/.cache/go-build
             ~/go/pkg/mod
-          key: ${{ runner.os }}-go-1.19.5-release-cache-${{ hashFiles('cli/go.sum') }}
+          key: ubuntu-go-1.19.5-release-cache-${{ hashFiles('cli/go.sum') }}
           restore-keys: |
-            ${{ runner.os }}-go-1.19.5-release-cache-cli
+            ubuntu-go-1.19.5-release-cache-cli
```
hermanschaaf, line 72: Shouldn't this be using `${{ matrix.os }}`?

yevgenypats: Im not sure this is needed anymore as this step doesn't run on matrix.

hermanschaaf: Ah I see, gotcha

hermanschaaf (approved):

PR #11909, opened 2023-07-03
chore(destinations): Update to SDK v4.6.1

yevgenypats:


disq:
Need to update testing as well.

PR #11941, opened 2023-07-04
chore: Remove mixpanel,gandi,launchdarkly plugins

yevgenypats:
Removing for now due to low usage.


hermanschaaf:
Also remember to remove their references from the website 🎗️

PR #11955, opened 2023-07-04
feat: Add table_options support for aws_securityhub_findings table

VictorCovalski:
#### Summary
This PR adds support for `table_options` for the table `aws_securityhub_findings`. This allows users to specify filters before fetching findings from Security Hub.


Dougoc:
👍

plugins/source/aws/client/tableoptions/securityhub.go
```diff
@@ -0,0 +1,45 @@
...
+	b, _ := json.Marshal(m)
+	return json.Unmarshal(b, &s.GetFindingsInput)
+}
+
+func (s *SecurityHubAPIs) validateGetFindingEvent() error {
+	for _, opt := range s.GetFindingsOpts {
+		if aws.ToString(opt.NextToken) != "" {
+			return errors.New("invalid input: cannot set NextToken in GetFindings")
+		}
+	}
```
bbernays, line 39: Maybe also validate that the MaxResults is in the appropriate range between 1 and 100 (inclusive)

plugins/source/aws/resources/services/securityhub/findings.go
```diff
@@ -44,19 +46,46 @@ This is useful when multi region and account aggregation is enabled.`,
...
+		for p.HasMorePages() {
+			response, err := p.NextPage(ctx, func(o *securityhub.Options) {
+				o.Region = cl.Region
+			})
+			if err != nil {
+				return err
+			}
+			res <- response.Findings
+		}
+		return nil
```
bbernays, line 69: Once all of the configs are in a single array you can just iterate through it without having to have the unique case of no custom inputs

```suggestion
	for _, w := range allConfigs {
		config = w.GetFindingsInput
		p := securityhub.NewGetFindingsPaginator(svc, &config)
		for p.HasMorePages() {
			response, err := p.NextPage(ctx, func(o *securityhub.Options) {
				o.Region = cl.Region
			})
			if err != nil {
				return err
			}
			res <- response.Findings
		}
	} 
```
```

**Review by bbernays — COMMENTED** [review]
Thank you for this PR, it looks really good! A few comments. And then if you could also add this to the docs [here](https://github.com/cloudquery/cloudquery/blob/aae59e21099c5c5552211e8d07c43cd530a90c92/website/pages/docs/plugins/sources/aws/configuration.md?plain=1#L108)

File: plugins/source/aws/client/tableoptions/securityhub.go (hunk 1 above)
Comment (VictorCovalski) [reply plugins/source/aws/client/tableoptions/securityhub.go]: Thanks for the input. I've just included this check in my last commit

File: plugins/source/aws/client/tableoptions/securityhub.go [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+	b, _ := json.Marshal(m)
+	return json.Unmarshal(b, &s.GetFindingsInput)
+}
+
+func (s *SecurityHubAPIs) validateGetFindingEvent() error {
+	for _, opt := range s.GetFindingsOpts {
+		if aws.ToString(opt.NextToken) != "" {
+			return errors.New("invalid input: cannot set NextToken in GetFindings")
+		}
+		if opt.MaxResults < 1 && opt.MaxResults > 100 {
```
hermanschaaf, line 39:
```suggestion
		if opt.MaxResults < 1 || opt.MaxResults > 100 {
```

It looks like this should be "or" not "and"

hermanschaaf (changes requested):
Looks great, thanks for the contribution! One small thing to fix, otherwise looks good to me 👍

VictorCovalski:
@hermanschaaf if `MaxResults` is not set in the configuration file then it will have the default value of 0 in the `GetFindingsInput` struct, which is not within valid range [1,100].

Would it be better to have this range go from [0,100]? Or perhaps we should let the AWS API validate this? 

As you can see, `MaxResults` is not validated in Cloudtrail `LookupEvents` table options.

https://github.com/cloudquery/cloudquery/blob/074aec2ae2d5b7b48e178e9df217d2b4ae55dca8/plugins/source/aws/client/tableoptions/cloudtrail.go#L34-L41

hermanschaaf (approved):
👍

hermanschaaf:
I think setting a default of 100 in the validation function makes sense for now, we generally want it to be the maximum 👍 Later on we can probably split it out and create a `SetDefaults` function that gets called before `Validate` for all table options, but we don't have to do that in this PR.

PR #11956, opened 2023-07-04
feat(shopify): Upgrade to SDK v4

disq:
Closes https://github.com/cloudquery/cloudquery/issues/11934


plugins/source/shopify/resources/plugin/client.go
```diff
@@ -0,0 +1,156 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return &Client{
+		config: config,
+		logger: logger,
+		scheduler: scheduler.NewScheduler(
+			scheduler.WithLogger(logger),
```
yevgenypats, line 128: concurrency is missing I think.

yevgenypats (approved):
Looks good. concurrency is missing imo

PR #11961, opened 2023-07-05
fix(gcp): Fix the query condition of the KMS Separation duties rule

aruneko:
#### Summary

The current query condition in the `kms_separation_of_duties.sql` is only checks that an user has the cloudkms.admin role. But the actually condition defined in the CIS Benchmarks document is an user should not have Cloud KMS Admin and any of the Cloud KMS CryptoKey Encrypter/Decrypter, Cloud KMS CryptoKey Encrypter or Cloud KMS CryptoKey Decrypter roles at the same time.

I fixed current condition to adjust for actual condition.


hermanschaaf (approved):
Thanks for the fix @aruneko !

PR #11973, opened 2023-07-05
fix(aws): Skip fetching tags for `aws_kafka_cluster_operations`

bbernays:
#### Summary

Resource doesn't support tagging. Rather than releasing a breaking change this will just skip the column resolver. In a future version we should delete the column


plugins/source/aws/resources/services/kafka/helpers.go
```diff
@@ -32,6 +32,10 @@ func describeClustersInput(resource *schema.Resource) kafka.DescribeClusterV2Inp
 
 func resolveKafkaTags(path string) schema.ColumnResolver {
 	return func(ctx context.Context, meta schema.ClientMeta, r *schema.Resource, c schema.Column) error {
+		if path == "ClusterOperations" {
```
hermanschaaf, line 35: We're passing in `OperationArn` in the table definition, shouldn't that be used here as well?

bbernays: Yes

hermanschaaf (approved):

hermanschaaf:
Was there an issue associated with this, or just something you discovered?

hermanschaaf (approved):

PR #11985, opened 2023-07-05
feat(facebookmarketing): Update to SDK V4

yevgenypats:
Closes https://github.com/cloudquery/cloudquery/issues/11928


plugins/source/facebookmarketing/resources/plugin/plugin.go
```diff
@@ -29,56 +37,122 @@ var customExceptions = map[string]string{
 	"customconversions": "Custom Conversions",
 }
 
-func titleTransformer(table *schema.Table) string {
+func titleTransformer(table *schema.Table) {
 	if table.Title != "" {
-		return table.Title
+		return
```
disq, line 42: Relations wouldn't be processed if title is set.

plugins/source/facebookmarketing/resources/plugin/plugin.go
```diff
@@ -29,56 +37,122 @@ var customExceptions = map[string]string{
...
 	}
 	for k, v := range customExceptions {
 		exceptions[k] = v
 	}
 	csr := caser.New(caser.WithCustomExceptions(exceptions))
 	t := csr.ToTitle(table.Name)
-	return strings.Trim(strings.ReplaceAll(t, "  ", " "), " ")
+	table.Title = strings.Trim(strings.ReplaceAll(t, "  ", " "), " ")
+	for _, rel := range table.Relations {
+		rel.Title = csr.ToTitle(rel.Name)
```
disq, line 55: Relation titles aren't subject to the same processing (if `Title` is set, don't set etc.) I think it's better to just make this fn as `func titleTransformer(table *schema.Table) error`, don't do the relations here, and use `transformers.Apply` which would take care of the relations.

PR #12003, opened 2023-07-06
feat(snyk): Update to SDK v4

### LINKED ISSUE #11996: Migrate `plugins/source/snyk` to `github.com/cloudquery/plugin-sdk/v4`

candiduslynx:
Closes #11996


plugins/source/snyk/client/transform.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+)
+
+func Transform(tables schema.Tables) schema.Tables {
+	if err := transformers.TransformTables(tables); err != nil {
+		panic(err)
+	}
+	for _, t := range tables {
```
hermanschaaf, line 12: We should be able to use `transformers.Apply` here so that it also gets applied to child tables https://github.com/cloudquery/plugin-sdk/blob/main/transformers/tables.go#L35

candiduslynx: Copied from digitalocean, tbh (cc: @disq )

hermanschaaf: This was added recently to the SDK by @disq himself :)

candiduslynx: But the `schema.AddCQIDs` already handles relations, so that shouldn't be done twice IMO

hermanschaaf: Hm okay, yeah then not... I guess 😅

hermanschaaf (approved):

PR #12052, opened 2023-07-07
feat(terraform)!: Migrate to SDK v4

### LINKED ISSUE #12050: Migrate `plugins/source/terraform` to `github.com/cloudquery/plugin-sdk/v4`

candiduslynx:
Closes #12050

BEGIN_COMMIT_OVERRIDE
feat!: Upgrades the Terraform source plugin to use plugin-sdk v4. 
This version contains no user-facing breaking changes, but because it is now using CloudQuery gRPC protocol v3, it does require use of a destination plugin that also supports protocol v3. All recent destination plugin versions support this.
END_COMMIT_OVERRIDE


plugins/source/terraform/client/spec.go
```diff
@@ -1,5 +1,28 @@
...
+		if err := bc.Validate(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Spec) SetDefaults() {
+	if s.Concurrency < 1 {
+		s.Concurrency = 10000
```
yevgenypats, line 26: const

disq: why? we have lots of plugins where concurrency isn't const (I migrated them)

yevgenypats (approved):

plugins/source/terraform/client/spec.go
yevgenypats: just a nit. I thought it's nice to have `const defaultConcurrency`

PR #12066, opened 2023-07-07
feat!: Upgrade Pagerduty to plugin-sdk v4

hermanschaaf:
BEGIN_COMMIT_OVERRIDE
feat!: Upgrades the pagerduty source plugin to use plugin-sdk v4. This version does not contain any user-facing breaking changes, but because it is now using CloudQuery gRPC protocol v3, it does require use of a destination plugin that also supports protocol v3. All recent destination plugin versions support this.
END_COMMIT_OVERRIDE

Closes https://github.com/cloudquery/cloudquery/issues/12064


yevgenypats (approved):

plugins/source/pagerduty/resources/plugin/all_tables.go
```diff
@@ -37,4 +38,12 @@ func AllTables() []*schema.Table {
 		users.Users(),
 		vendors.Vendors(),
 	}
+	if err := transformers.TransformTables(t); err != nil {
+		panic(err)
+	}
+	for _, table := range t {
+		schema.AddCqIDs(table)
+		titleTransformer(table)
```
disq, line 46: This doesn't apply the transformer to child tables of `table`, passing it to `transformers.Apply` would do it.

hermanschaaf: Good point, thanks, updated 👍

PR #12111, opened 2023-07-10
feat(aws): Support `networkmanager` resources

bbernays:

### DIFF
--- a/website/pages/docs/plugins/sources/aws/configuration.md
+++ b/website/pages/docs/plugins/sources/aws/configuration.md
@@ -105,9 +105,9 @@ This is the (nested) spec used by the AWS source plugin.
 
   When set to `true` plugin will sync data from APIs that incur a fee. Currently only `aws_costexplorer*` and `aws_alpha_cloudwatch_metric*` tables require this flag to be set to `true`.
 
-- **experimental** `table_options` (map) (default: not used)
+- **preview** `table_options` (map) (default: not used)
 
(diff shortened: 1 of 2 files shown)


plugins/source/aws/client/testing.go
```diff
@@ -16,9 +16,14 @@ import (
...
 type TestOptions struct {
 	TableOptions tableoptions.TableOptions
+	Region       string
 }
 
 func AwsMockTestHelper(t *testing.T, table *schema.Table, builder func(*testing.T, *gomock.Controller) Services, testOpts TestOptions) {
+	if testOpts.Region == "" {
+		testOpts.Region = "us-east-1"
+	}
+
```
bbernays, line 26: This is necessary because `networkmanager` is the first service that is not available in `us-east-1` it is only available in `us-west-2`

website/tables/aws/aws_networkmanager_global_networks.md
```diff
@@ -0,0 +1,31 @@
+# Table: aws_networkmanager_global_networks
+
+This table shows data for Networkmanager Global Networks.
+
+https://docs.aws.amazon.com/networkmanager/latest/APIReference/API_GlobalNetwork.html
+The  'request_region' column is added to show region of where the request was made from.
```
hermanschaaf, line 6: It doesn't seem like there is a `request_region` column on this table

bbernays: This has been resolved

PR #12121, opened 2023-07-11
fix(aws): Validate table relations not just top level table

bbernays:
#### Summary

Currently we are only validating that the top level table has no empty columns... This change ensures we validate all relations as well


hermanschaaf (approved):
Nice catch!

disq (approved):

PR #12129, opened 2023-07-12
chore(ci): Use large runners in all components

erezrokah:
#### Summary

This PR switches all monorepo components to use large runners for test workflows and releases


erezrokah:
I pushed an additional commit https://github.com/cloudquery/cloudquery/pull/12129/commits/1e4bdc029e040306e4b25174e938a4d150bf5d90 to cancel in progress runs if someone pushes a new commit to the branch. This should help reduce the usage.

We can also try without this commit and see how it goes

yevgenypats (approved):

PR #12133, opened 2023-07-12
fix(website): Update Simple Analytics Guide configuration

jsonpr:
Fixing the CQ example configuration by removing a `-` in path


hermanschaaf (approved):
👍

PR #12145, opened 2023-07-12
fix(deps): Upgrade destination plugins to SDK v4.0.0 release

disq:
More dests find/replace didn't catch.

After this there should be no `-rc1` references in `go.mod`, except for CH and ES which have their own PRs.


plugins/destination/mongodb/go.mod
```diff
@@ -13,6 +13,8 @@ require (
 // TODO: remove once all updates are merged
 replace github.com/apache/arrow/go/v13 => github.com/cloudquery/arrow/go/v13 v13.0.0-20230710001530-a2a76ebbb85f
 
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/disq/plugin-sdk/v4 v4.0.0-20230712231129-e87946c51a31
```
disq, line 16: Needs a tagged release

PR #12152, opened 2023-07-13
fix: Fix paths for premium plugins

hermanschaaf:
Fixes the static integration paths for premium plugins that don't have directories in the prebuilt location anymore


erezrokah (approved):
🚀

PR #12155, opened 2023-07-13
fix(clickhouse): Update to SDK v4.2.1

disq:
Skipped the failing tests


erezrokah (approved):

candiduslynx (approved):

candiduslynx:
Maybe its about empty json?
Like, could you check that the json type (arrow extension) works OK if we construct array with single append empty & then use marshal json on this?

disq:
> Maybe its about empty json? Like, could you check that the json type (arrow extension) works OK if we construct array with single append empty & then use marshal json on this?

My debugging showed the JSON had a bell character inside so it's really weird

erezrokah (approved):

plugins/destination/clickhouse/client/client_test.go
```diff
@@ -43,5 +44,9 @@ func TestPlugin(t *testing.T) {
 			},
 		},
 		plugin.WithTestSourceAllowNull(types.CanBeNullable),
+		plugin.WithTestDataOptions(schema.TestSourceOptions{
+			SkipStructs: true, // panic during marshal during diff
+			SkipMaps:    true, // ordering doesn't match when read back
```
candiduslynx, line 49: There is an option for approx equal that handles this.
We should use it here, too

PR #12158, opened 2023-07-13
feat(firehose): Add batch options

disq:


erezrokah (approved):

plugins/destination/firehose/client/spec.go
```diff
@@ -7,16 +7,35 @@ import (
...
+	defaultMaxBatchRecords    = 500
+	defaultMaxBatchSizeBytes  = 4194000
+)
+
 func (s *Spec) SetDefaults() {
 	if s.MaxRetries == nil {
 		s.MaxRetries = new(int)
 		*s.MaxRetries = 5
 	}
+	if s.MaxRecordSizeBytes == 0 {
```
disq, line 30: This cheeky `== 0` allows to use -1 max record size for no limit.

hermanschaaf: Why would you want no limit? I don't think it will work if you send anything larger than 1MB to Kinesis https://aws.amazon.com/kinesis/data-firehose/faqs/

> The maximum size of a record (before Base64-encoding) is 1024 KB if your data source is Direct PUT or Kinesis Data Streams.

disq: Ok, agreed. Btw it says before the base64 encoding, so I don't think the limiting code is counting the bytes correctly.

PR #12161, opened 2023-07-13
fix(website): Move Mongo Atlast and Render tables to correct location

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/12160.

We render tables dynamically otherwise our build time takes too long


erezrokah:
This also ensures these tables show up in the search

hermanschaaf (approved):

PR #12170, opened 2023-07-14
chore(deps): Disable `plugin-pb-go` dep updates in plugins

erezrokah:
#### Summary

This should reduce the noise of PRs such as https://github.com/cloudquery/cloudquery/pull/12166 by not updating `plugin-pb-go` in plugins. `plugin-pb-go` is an indirect dependency for plugins and should be updated via the SDK anyways.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -20,6 +20,11 @@
       matchPackageNames: ["github.com/jackc/pgx/v4"],
       allowedVersions: "<5",
     },
+    {
+      matchPackageNames: ["github.com/cloudquery/plugin-pb-go"],
+      matchFileNames: ["plugins/**"],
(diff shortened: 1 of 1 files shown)


yevgenypats (approved):

.github/renovate.json5
```diff
@@ -20,6 +20,11 @@
       matchPackageNames: ["github.com/jackc/pgx/v4"],
       allowedVersions: "<5",
     },
+    {
+      matchPackageNames: ["github.com/cloudquery/plugin-pb-go"],
```
erezrokah, line 24: Per renovate docs, all matchers need to match for the rule to apply

hermanschaaf (approved):

PR #12214, opened 2023-07-14
test: Validate all source plugin tables

disq:
~Some plugins _may_ have issues with PKs, will be fixed in future commits.~


erezrokah (approved):
Looks good 🚀 I changed the title to indicate a non user facing change

bbernays (approved):

PR #12238, opened 2023-07-17
fix: Fail if can't read current_schema or current_database

disq:
It also slightly modifies the logic in list_tables, could be unneeded but it looks like a safer approach to me.


plugins/source/postgresql/client/list_tables.go
```diff
@@ -65,20 +65,26 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
...
+		if table == nil {
+			table = &schema.Table{
 				Name:    tableName,
 				Columns: make([]schema.Column, 0),
-			})
+			}
+			tableMap[tableName] = table
+			tables = append(tables, table)
 		}
-		table := tables[len(tables)-1]
```
disq, line 81: This line caused problems for a user in discord, can't be sure why (resultset isn't empty but ordinalPosition didn't start at 1...)

erezrokah (approved):

yevgenypats (approved):

PR #12249, opened 2023-07-17
fix(broken-links-checker): Exclude `register_for_cloud`

erezrokah:
#### Summary

Our broken links checker considers redirects as broken links, so we need to exclude them.


erezrokah:
Going to merge this to fix the CI

Issue #12273, opened 2023-07-18
feat: Add AWS Backup Jobs to AWS source plugin


jsonpr:
### Describe the resource

https://docs.aws.amazon.com/aws-backup/latest/devguide/API_BackupJob.html

### Use Case

AWS Backup Job Information

### Link to API documentation

https://docs.aws.amazon.com/aws-backup/latest/devguide/API_BackupJob.html

### Pull request (optional)

- [ ] I can submit a pull request

PR #12389, opened 2023-07-18
feat(resources): Add Backup Jobs

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12273. Tested on real resources.

There's also https://docs.aws.amazon.com/aws-backup/latest/devguide/API_DescribeBackupJob.html but is seems it doesn't add too much new information


plugins/source/aws/resources/services/backup/jobs.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func Jobs() *schema.Table {
+	tableName := "aws_backup_jobs"
+	return &schema.Table{
+		Name:        tableName,
+		Description: `https://docs.aws.amazon.com/aws-backup/latest/devguide/API_BackupJob.html`,
+		Resolver:    fetchBackupJobs,
+		Multiplex:   client.ServiceAccountRegionMultiplexer(tableName, "backup"),
+		Transform:   transformers.TransformWithStruct(&types.BackupJob{}, transformers.WithPrimaryKeys("BackupJobId")),
```
erezrokah, line 21: `BackupJobId` could be enough as a PK based on the API in https://docs.aws.amazon.com/aws-backup/latest/devguide/API_DescribeBackupJob.html (it only accepts `BackupJobId`) but I couldn't find any other docs that confirm it

plugins/source/aws/resources/services/backup/jobs.go
```diff
@@ -0,0 +1,44 @@
...
+			client.DefaultAccountIDColumn(true),
+			client.DefaultRegionColumn(true),
+		},
+	}
+}
+
+func fetchBackupJobs(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services().Backup
+	params := backup.ListBackupJobsInput{ByAccountId: aws.String(cl.AccountID), MaxResults: aws.Int32(1000)} // maximum value from https://docs.aws.amazon.com/aws-backup/latest/devguide/API_ListBackupJobs.html
```
erezrokah, line 32: `ByAccountId` can be `*` to list all jobs for the account, but that doesn't sit well with our multiplexing.
It seems better to multiplex and save the account and region data

yevgenypats (approved):

jsonpr (approved):

PR #12402, opened 2023-07-18
feat(aws)!: Define composite primary key for lambda_function_versions

smokentar:
#### Summary

This PR is aimed to address #12391 by creating a composite primary key on `aws_lambda_function_versions` table comprising `function_arn` and `version`.

Since `function_arn` will always be unique, `account_id` or `region` are not required to be part of the primary key.

#### Summary

This is my first PR against the codebase, extra pair of eyes are highly recommended!


erezrokah (approved):
Looks great @smokentar 🚀 I added another commit to fix the styling and make the CI happy

disq (approved):

erezrokah:
Thanks for the PR @smokentar, see the expected release version in https://github.com/cloudquery/cloudquery/pull/12406.
We usually release those on Tuesdays to avoid too many releases with breaking changes.
If you'd like to submit more breaking fixes, we can group them into the same release

PR #12408, opened 2023-07-19
feat(aws)!: Define primary key for s3_bucket_encryption_rules

smokentar:
#### Summary

This PR is aimed to address #12392 by creating a primary key (`bucket_arn`) on `aws_s3_bucket_encryption_rules` table.

### DIFF
--- a/plugins/source/aws/resources/services/s3/bucket_encryption_rules.go
+++ b/plugins/source/aws/resources/services/s3/bucket_encryption_rules.go
@@ -22,9 +22,10 @@ func bucketEncryptionRules() *schema.Table {
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
+				Name:       "bucket_arn",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.ParentColumnResolver("arn"),
+				PrimaryKey: true,
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


plugins/source/aws/resources/services/s3/bucket_encryption_rules.go
```diff
@@ -22,9 +22,10 @@ func bucketEncryptionRules() *schema.Table {
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
+				Name:       "bucket_arn",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.ParentColumnResolver("arn"),
+				PrimaryKey: true,
```
erezrokah, line 28: Thanks for the PR @smokentar, I'm not this is sufficient as wit the parent bucket as the only primary key it will allow saving a single rule per bucket.

Maybe the solution should be to remove this table and have the rules as an additional JSON column on the bucket table

erezrokah (changes requested):
Thanks for the PR @smokentar, added a comment. Please let me know what you think

plugins/source/aws/resources/services/s3/bucket_encryption_rules.go
erezrokah: Thanks for the additional context @smokentar! I only looked at the AWS Go SDK struct of `ServerSideEncryptionConfiguration.Rules` and assumed that if it's an array, multiple items can be returned.

If there can only be a single rule (for now), this looks great 🚀

erezrokah (approved):

PR #12415, opened 2023-07-19
feat(aws)!: Define composite primary key for regions

smokentar:
#### Summary

This PR is aimed to address #12414 by creating a composite primary key on `aws_regions` table comprising `account_id` and `region`.

With this key there will be one set of regions (27) per discovered AWS Account.
The reason why we're not limiting them to only one set of regions for all accounts is that different AWS Accounts can have different `opt_in_status` / `enabled` values for different regions.


erezrokah (approved):
🚀

Issue #12416, opened 2023-07-19
bug: Empty `aws_acm_certificates` table from v19.1.0


akash1810:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

SInce v19.1.0 of the AWS source plugin, specifically https://github.com/cloudquery/cloudquery/pull/11761, ACM certificates are not collected.

Using the AWS CLI, I have 74 certificates:

```shell
➜ aws acm list-certificates --region eu-west-1 | jq -r '.CertificateSummaryList[].CertificateArn' | wc -l
      74
```

They are all RSA_2048:

```shell
➜ aws acm list-certificates --region eu-west-1 | jq -r '.CertificateSummaryList[].KeyAlgorithm' | uniq
RSA-2048
```

They are also all DNS validated, with a mixture of in use statuses.

However when running CloudQuery v19.1.0 and later, 0 items are collected.

Using any previous version yields the desired result.

### Expected Behavior

All ACM certificates should be collected, regardless of their key algorithm.

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: aws
  path: cloudquery/aws
  version: "v19.1.0"
  tables: ["aws_acm_certificates"]
  destinations: ["file"]
  spec:
    regions:
      - eu-west-1
---
kind: destination
spec:
  name: "file"
  path: "cloudquery/file"
  version: "v3.3.0"
  spec:
    path: "data/{{TABLE}}/{{UUID}}.{{FORMAT}}"
    format: "csv"
    format_spec:
      delimiter: ","
```

### Steps To Reproduce

1. Create an ACM certificate
2. With the above config, run `sync --log-console --log-level debug`
3. Observe an empty `aws_acm_certificates` table

### CloudQuery (redacted) logs

```
2023-07-19T18:47:21Z INF Loading spec(s) args=["aws.yaml"] module=cli
2023-07-19T18:47:21Z INF Source plugin server listening address=/var/folders/2k/kh_57mss3254t3gx2qyt3_48___18m/T/cq-SKxGBNEMEVDUhwwv.sock module=cli
2023-07-19T18:47:22Z INF Plugin server listening address=/var/folders/2k/kh_57mss3254t3gx2qyt3_48___18m/T/cq-rNWAHlSfVzHJCHWO.sock module=cli
[...]

PR #12436, opened 2023-07-21
fix(resources-acm-certificates): Filter by all key usages

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12416


disq (approved):

akash1810:
Thanks for this!

Issue #12423, opened 2023-07-20
bug: table aws_appconfig_applications - column "deployment_duration_in_minutes" of relation "aws_appconfig_applications" does not exist


ozbillwang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```
Starting sync for: aws (v21.1.0) -> [postgresql (v5.0.1)]
\ Syncing resources... (659/-, 278 resources/s) [1s] Error: failed to sync v3 source aws: write client returned error (insert): plugin returned error: failed to execute batch with pgerror: severity: ERROR, code: 42703, message: **column "deployment_duration_in_minutes" of relation "aws_appconfig_applications" does not exist**, detail :, hint: , position: 130, internal_position: 0, internal_query: , where: , schema_name: , table_name: , column_name: , data_type_name: , constraint_name: , file: parse_target.c, line: 1061, routine: checkInsertTargets: ERROR: column "deployment_duration_in_minutes" of relation "aws_appconfig_applications" does not exist (SQLSTATE 42703)
```

### Expected Behavior

successfully sync the aws account

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: "aws"
  path: "cloudquery/aws"
  version: "v21.1.0"
  tables: ["*"]
  skip_tables:
    - aws_ec2_vpc_endpoint_services
    - aws_cloudtrail_events
    - aws_docdb_cluster_parameter_groups
    - aws_docdb_engine_versions
    - aws_ec2_instance_types
    - aws_elasticache_engine_versions
    - aws_elasticache_parameter_groups
    - aws_elasticache_reserved_cache_nodes_offerings
    - aws_elasticache_service_updates
    - aws_iam_group_last_accessed_details
    - aws_iam_policy_last_accessed_details
    - aws_iam_role_last_accessed_details
    - aws_iam_user_last_accessed_details
    - aws_neptune_cluster_parameter_groups
    - aws_neptune_db_parameter_groups
    - aws_rds_cluster_parameter_groups
    - aws_rds_db_parameter_groups
    - aws_rds_engine_versions
    - aws_servicequotas_services
  destinations: ["postgresql"]
  spec:
```

### Steps To Reproduce

I have excluded the table 
```
  skip_tables:
    - ...
    - aws_appconfig_applications
```

then sync command works.

### CloudQuery (redacted) logs

```
Starting sync for: aws (v21.1.0) -> [postgresql (v5.0.1)]
[...]

PR #12437, opened 2023-07-21
doc: Update destinations postgresql with usage instructions improve

ozbillwang:
#### Summary
Update destinations postgresql with usage instructions improve (#12423)


erezrokah (commented):
Hi @ozbillwang, thanks for the PR.
Forced migration is not a PosgreSQL only feature.
We have it documented in https://www.cloudquery.io/docs/reference/destination-spec#migrate_mode and https://www.cloudquery.io/docs/advanced-topics/migrations.

Not sure what's the best way to expose those from individual plugins. Maybe a note that the plugin supports forced migration?

erezrokah:
Thanks for the PR @ozbillwang. I'll close it as stale for now. Happy to reopen per a response for my comment

PR #12465, opened 2023-07-25
fix(bigquery): Create less clients per dataset/table

disq:
Seems like we've been creating one extra client per dataset and then one extra client per table.

This PR changes it so that we're creating a single client for all datasets ~~, and another client for all tables per dataset~~ per multiplexer. Should help with the [429 errors](https://discord.com/channels/872925471417962546/873606591335759872/1132870724667834439) as each client authenticates again.

The PreResourceResolver pattern is only used in bigquery resources.

### DIFF
--- a/plugins/source/gcp/resources/services/bigquery/datasets_mock_test.go
+++ b/plugins/source/gcp/resources/services/bigquery/datasets_mock_test.go
@@ -2,7 +2,6 @@ package bigquery
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


plugins/source/gcp/resources/services/bigquery/datasets_mock_test.go
```diff
@@ -111,8 +110,6 @@ func createBigqueryDatasets(mux *httprouter.Router) error {
 	}
 
 	mux.GET("/projects/testProject/datasets/testDataset/tables/:table", func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-		fmt.Println("what")
```
erezrokah, line 114: 🙈

erezrokah (approved):

PR #12473, opened 2023-07-25
feat(azure): Add azure private endpoints

kumaya:
#### Summary

Adds azure Private Endpoints


erezrokah (approved):
Thanks @kumaya, looks great 🎸

Issue #12440, opened 2023-07-21
bug: Error copying table structure from Postgres -> DuckDB


nicosuave:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

On first run from this particular app's database, I encounter

```
terminate called after throwing an instance of 'duckdb::CatalogException'                                                                                                                                               │
 what():  Catalog Error: Column with name created_at already exists!  
 ```

Have tried updating to the most recent versions of each plugin (Postgres source, DuckDB destination) without any luck.

### Expected Behavior

For the structure to be replicated in the destination

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: abc
  registry: "github"
  path: "cloudquery/postgresql"
  version: "v2.0.2"
  destinations: ["xyz"]
  tables: ["*"]
  spec:
    connection_string: ...
---
kind: destination
spec:
  name: xyz
  path: cloudquery/duckdb
  version: "v4.1.2"
  write_mode: overwrite
  spec:
    connection_string: ...
```

### CloudQuery (redacted) logs

```
... Loading spec(s) from ....yml
... Downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-postgresql-v2.0.2/postgresql_linux_amd64.zip
... 
Downloading   0% |                                      | ( 0 B/7.5 MB) [0s:0s]
                                                                               

Downloading  15% |████                          | (1.2/7.5 MB, 18 MB/s) [0s:0s]
                                                                               

Downloading 100% |██████████████████████████████| (7.5/7.5 MB, 75 MB/s)
... Downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-destination-duckdb-v4.1.2/duckdb_linux_amd64.zip
... 
Downloading   0% |                                       | ( 0 B/17 MB) [0s:0s]
                                                                               

Downloading  73% |██████████████████████         | (13/17 MB, 193 MB/s) [0s:0s]
                                                                               

Downloading 100% |███████████████████████████████| (17/17 MB, 200 MB/s)
... Starting sync for: ... (postgresql@v2.0.2) -> [motherduck (duckdb@v4.1.2)]
... terminate called after throwing an instance of 'duckdb::CatalogException'
...   what():  Catalog Error: Column with name id already exists!
... SIGABRT: abort
... PC=0x7fd51196accc m=7 sigcode=18446744073709551610
... signal arrived during cgo execution
... 
... goroutine 25 [syscall]:
[...]

PR #12518, opened 2023-07-25
fix: Don't create duplicate columns for unique constraints

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12440

We can't use `like` otherwise an ordinal of `1` matches both `indkey = 11` and `indkey = 1`


hermanschaaf:
Tricky, nice find! Would it be feasible / easy to add a test case for this you think?

erezrokah:
> Would it be feasible / easy to add a test case for this you think?

Good point, added a test in https://github.com/cloudquery/cloudquery/pull/12518/commits/8b4e25c40d09c8f547432d430f11c1e3e51328cb (fixed the lint warnings in a separate commit)

plugins/source/postgresql/resources/plugin/plugin_test.go
```diff
@@ -294,9 +323,9 @@ func assertRecord(t *testing.T, actualRecord arrow.Record, expected []testCase)
 	if len(expected) != int(actualRecord.NumCols()) {
 		t.Fatalf("expected record to have %d columns, got %d", len(expected), actualRecord.NumCols())
 	}
-	schema := actualRecord.Schema()
+	sc := actualRecord.Schema()
```
erezrokah, line 326: Need to prevent shadowing the new import `"github.com/cloudquery/plugin-sdk/v4/schema"`

hermanschaaf (approved):

Issue #12528, opened 2023-07-26
AWS EMR Additional Resources: Studio and Studio Session Mapping


mnorbury:
Relates: https://github.com/cloudquery/cloudquery/issues/7708

Add the following new resources to the `emr` service in the `aws` plugin:

- [x] Studio Session Mappings: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/emr/list-studio-session-mappings.html
- [x] Studios: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/emr/list-studios.html

PR #12529, opened 2023-07-26
feat(resources): Add additional emr resources for studio and studio session mapping

mnorbury:
#### Summary
Fixes: https://github.com/cloudquery/cloudquery/issues/12528

### DIFF
--- a/plugins/source/aws/resources/services/emr/studio_session_mapping.go
+++ b/plugins/source/aws/resources/services/emr/studio_session_mapping.go
@@ -1,0 +1,57 @@
+package emr
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/service/emr"
+	"github.com/aws/aws-sdk-go-v2/service/emr/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+)
+
+func studioSessionMapping() *schema.Table {
+	tableName := "aws_emr_studio_session_mapping"
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/emr/latest/APIReference/API_SessionMappingSummary.html`,
+		Resolver:            fetchEmrStudioSessionMapping,
+		PreResourceResolver: getSessionMapping,
+		Transform:           transformers.TransformWithStruct(&types.SessionMappingDetail{}),
+	}
+}
+
+func fetchEmrStudioSessionMapping(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	p := parent.Item.(*types.Studio)
+	svc := cl.Services().Emr
+	paginator := emr.NewListStudioSessionMappingsPaginator(svc, &emr.ListStudioSessionMappingsInput{
+		StudioId: p.StudioId,
+	})
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(ctx, func(options *emr.Options) {
+			options.Region = cl.Region
+		})
+		if err != nil {
+			return err
+		}
+		res <- page.SessionMappings
+	}
+	return nil
+}
+
+func getSessionMapping(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services().Emr
+	response, err := svc.GetStudioSessionMapping(ctx, &emr.GetStudioSessionMappingInput{
+		StudioId:     resource.Item.(types.SessionMappingSummary).StudioId,
+		IdentityType: resource.Item.(types.SessionMappingSummary).IdentityType,
+		IdentityId:   resource.Item.(types.SessionMappingSummary).IdentityId,
+	}, func(options *emr.Options) {
+		options.Region = cl.Region
+	})
+	if err != nil {
+		return err
+	}
+	resource.Item = response.SessionMapping
+	return nil
+}
--- a/plugins/source/aws/resources/services/emr/studio_session_mapping.go
+++ b/plugins/source/aws/resources/services/emr/studio_session_mapping.go
@@ -11,7 +11,7 @@ import (
 )
 
 func studioSessionMapping() *schema.Table {
-	tableName := "aws_emr_studio_session_mapping"
+	tableName := "aws_emr_studio_session_mappings"
 	return &schema.Table{
 		Name:                tableName,
 		Description:         `https://docs.aws.amazon.com/emr/latest/APIReference/API_GetStudioSessionMapping.html`,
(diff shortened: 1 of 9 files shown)


plugins/source/aws/resources/services/emr/studio_session_mapping.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+)
+
+func studioSessionMapping() *schema.Table {
+	tableName := "aws_emr_studio_session_mapping"
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/emr/latest/APIReference/API_SessionMappingSummary.html`,
```
disq, line 16:
```suggestion
		Description:         `https://docs.aws.amazon.com/emr/latest/APIReference/API_SessionMappingDetail.html`,
```
Should match the final type the user will access (if not then the call - `API_GetStudioSessionMapping.html`)

plugins/source/aws/resources/services/emr/studios.go
```diff
@@ -0,0 +1,57 @@
...
+	tableName := "aws_emr_studios"
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/emr/latest/APIReference/API_Studio.html`,
+		Resolver:            fetchEmrStudios,
+		PreResourceResolver: getStudio,
+		Multiplex:           client.ServiceAccountRegionMultiplexer(tableName, "elasticmapreduce"),
+		Transform:           transformers.TransformWithStruct(&types.Studio{}),
+		Relations:           []*schema.Table{studioSessionMapping()},
+		Columns: []schema.Column{
```
disq, line 22: I would swap Relations and Columns, also indent relations as multiline so it's easier to spot :)

disq (commented):
Good work! Few comments + I think both tables also need primary keys of some sort, if possible

plugins/source/aws/resources/services/emr/studios.go
```diff
@@ -0,0 +1,65 @@
...
+		Transform:           transformers.TransformWithStruct(&types.Studio{}),
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(false),
+			client.DefaultRegionColumn(false),
+			{
+				Name:       "studio_id",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.PathResolver("StudioId"),
+				PrimaryKey: true,
+			}},
```
bbernays, line 30: In `aws_emr_studio_session_mapping` we use both `account_id` and `region` as PKs, but here we only use `studio_id`... If `studio_id` is globally unique (across all regions and accounts) we should remove those 2 PKs from the relation table, otherwise maybe we should add Region+Account to this PK?

bbernays (commented):
This looks really good! The only thing that I think is missing is running the `make gen-docs` to generate the table documentation for the new resources

plugins/source/aws/resources/services/emr/studios.go
mnorbury: Good point. I _think_ `studio_id` would be unique, but the [documentation](https://docs.aws.amazon.com/cli/latest/reference/emr/create-studio.html#output) doesn't make it completely clear. In this scenario do you think it would be better to add the 2PKs to this table also?

bbernays: We only ever assume an ID is going to be unique within a single region + account unless AWS clearly documents something different... For example s3 bucket name is guaranteed to be globally unique... So it is better to add the region + account ID as part of the PK

bbernays, line 30: So just looked it up in the docs and EMR Studio does actually have an ARN...

`arn:aws:elasticmapreduce:<region>:<account-id>:studio/<studio-id>`

https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-studio-manage-users.html

So I would make the PK a single field `arn` like this example: https://github.com/cloudquery/cloudquery/blob/91b2c2b8e43b733d35356669ddd24e579e9a5fb3/plugins/source/aws/resources/services/s3/multi_region_access_points.go#L27-L33

mnorbury: Okay, I've modified to use the `arn` instead. Since the `arn` is already present in the data, I think I can continue to use the `schema.PathResolver` to extract the value.

plugins/source/aws/resources/services/emr/studio_session_mapping.go
```diff
@@ -0,0 +1,61 @@
...
+	return &schema.Table{
+		Name:                tableName,
+		Description:         `https://docs.aws.amazon.com/emr/latest/APIReference/API_GetStudioSessionMapping.html`,
+		Resolver:            fetchEmrStudioSessionMapping,
+		PreResourceResolver: getSessionMapping,
+		Transform:           transformers.TransformWithStruct(&types.SessionMappingDetail{}, transformers.WithPrimaryKeys("StudioId", "IdentityType", "IdentityId")),
+		Columns: []schema.Column{
+			client.DefaultAccountIDColumn(true),
+			client.DefaultRegionColumn(true),
+		},
```
bbernays, line 23: Now that the parent table has just `arn` as the PK, you can change this PK to be just `studio_arn:IdentityType:IdentityId` and example of this can be found in the `aws_cloudwatchlogs_log_group_data_protection_policies` table

bbernays (dismissed):

plugins/source/aws/resources/services/emr/studio_session_mapping.go
mnorbury: Okay, that should now be the case in the latest commit https://github.com/cloudquery/cloudquery/pull/12529/commits/bbc2805a0afe607553ff5c333b98ba43a2d63f2f

plugins/source/aws/resources/services/emr/studio_session_mapping.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func getSessionMapping(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services().Emr
+	response, err := svc.GetStudioSessionMapping(ctx, &emr.GetStudioSessionMappingInput{
+		StudioId:     resource.Item.(types.SessionMappingSummary).StudioId,
+		IdentityType: resource.Item.(types.SessionMappingSummary).IdentityType,
+		IdentityId:   resource.Item.(types.SessionMappingSummary).IdentityId,
+	}, func(options *emr.Options) {
```
bbernays, line 61: Rather than casting the `resource.Item` 3 times I would probably just cast it once and save it to a  variable
```suggestion
	sms := resource.Item.(types.SessionMappingSummary)
	response, err := svc.GetStudioSessionMapping(ctx, &emr.GetStudioSessionMappingInput{
		StudioId:     sms.StudioId,
		IdentityType: sms.IdentityType,
		IdentityId:   sms.IdentityId,
	}, func(options *emr.Options) {
```

bbernays (approved):
LGTM 🚢 🚀

Issue #12526, opened 2023-07-26
bug: can't switch "write_mode" between "overwrite-delete-stale", "overwrite", "append"


ozbillwang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have run the sync successfully, but after change the `write_mode` from "overwrite-delete-stale" to "append" , or "append" to "overwrite-delete-stale", the sync command will throw errors. 

1. If not set migrate_mode, it recommends to enable `migrate_mode `to `forced`

2. after set `migrate_mode `to `forced`, it starts to give errors as below

```
\ Syncing resources... (56/-, 186 resources/s) [0s] Error: failed to sync v3 source azure: write client returned error (insert): plugin returned error: failed to drop tabl
e azure_windowsiot_services: ERROR: cannot drop table azure_windowsiot_services because other objects depend on it (SQLSTATE 2BP01)
```
follow the instruction, I drop that table
```
drop table azure_windowsiot_services CASCADE;
```
it will ask to drop another one, no end. 
```
| Syncing resources... (59/-, 17 resources/s) [4s] Error: failed to sync v3 source azure: write client returned error (insert): plugin returned error: failed to drop table
 azure_policy_assignments: ERROR: cannot drop table azure_policy_assignments because other objects depend on it (SQLSTATE 2BP01)
\ Syncing resources... (59/-, 15 resources/s) [4s]
```
Finally I have to drop all tables with `CASCADE`, but that's not correct way, right?

Got similar issue for both AWS and Azure

### Expected Behavior

should be no issue

### CloudQuery (redacted) config

```
$ cat postgresql.yml 

kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v5.0.1"
  #write_mode: "append"
  write_mode: "overwrite-delete-stale"
  migrate_mode: forced # recommended, when upgrade with new version of plugins
  spec:
    connection_string: ""
```

### Steps To Reproduce

1. run sync command, but switch "write_mode" between "overwrite-delete-stale", "overwrite", "append"

### CloudQuery (redacted) logs

```
\ Syncing resources... (56/-, 186 resources/s) [0s] Error: failed to sync v3 source azure: write client returned error (insert): plugin returned error: failed to drop tabl
e azure_windowsiot_services: ERROR: cannot drop table azure_windowsiot_services because other objects depend on it (SQLSTATE 2BP01)
```

### CloudQuery version

cloudquery version 3.10.1

[...]

PR #12531, opened 2023-07-26
chore(docs-spec): Add `write_mode`, `migrate_mode` clarifications

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/12526#issuecomment-1651384771

### DIFF
--- a/website/pages/docs/reference/destination-spec.md
+++ b/website/pages/docs/reference/destination-spec.md
@@ -66,13 +66,16 @@ Specifies the update method to use when inserting rows. The exact semantics depe
 - `overwrite`: Same as `overwrite-delete-stale`, but doesn't delete stale rows from previous `sync`s.
 - `append`: Rows are never overwritten or deleted, only appended.
 
+> Switching from `overwrite-delete-stale` or `overwrite` to `append`, or from `append` to `overwrite-delete-stale` or `overwrite` is not supported without dropping all the tables.
+To drop all tables automatically, use the `migrate_mode: forced` option.
+
 <!-- vale off -->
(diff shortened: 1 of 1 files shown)


website/pages/docs/reference/destination-spec.md
```diff
@@ -66,13 +66,16 @@ Specifies the update method to use when inserting rows. The exact semantics depe
 - `overwrite`: Same as `overwrite-delete-stale`, but doesn't delete stale rows from previous `sync`s.
 - `append`: Rows are never overwritten or deleted, only appended.
 
+> Switching from `overwrite-delete-stale` or `overwrite` to `append`, or from `append` to `overwrite-delete-stale` or `overwrite` is not supported without dropping all the tables.
```
disq, line 69: 'all' only if the user has `tables: *`, no? One can theoretically sync different tables with separate/incompatible modes as long as they are not synced in the same run

hermanschaaf (approved):

website/pages/docs/reference/destination-spec.md
hermanschaaf: Oh sorry got bit by automerge, but I think "all tables matched by your table config" is the most obvious interpretation of the word `all` here

hermanschaaf: Maybe we can change it to what I said just to be clear we won't drop other tables though

erezrokah: Good point, follow up in https://github.com/cloudquery/cloudquery/pull/12532

Issue #12524, opened 2023-07-25
bug: type mismatch: got=utf8, want=date64


nicosuave:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Seems the `date` datatype in Postgres is unhandled when copying data. Timestamp type seems to move over fine FWIW. 

I now see (what I assume to be staging tables based on the uuid suffix) sibling tables in the destination as well, as I saw a panic prior to any rows being copied if helpful

### Expected Behavior

For `date` datatype to be handled

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: abc
  registry: "github"
  path: "cloudquery/postgresql"
  version: "v2.0.3"
  destinations: ["xyz"]
  tables: ["*"]
  spec:
    connection_string: ...
---
kind: destination
spec:
  name: xyz
  path: cloudquery/duckdb
  version: "v4.1.2"
  write_mode: overwrite
  spec:
    connection_string: ...
```

### Steps To Reproduce

Create a table with a `date` datatype and try to sync that

### CloudQuery (redacted) logs

```
/ Syncing resources... (34/-, 337 resources/s) [0s] 
                                                    

- Syncing resources... (34/-, 169 resources/s) [0s] 
                                                    

\ Syncing resources... (34/-, 113 resources/s) [0s] 
                                                    

| Syncing resources... (34/-, 85 resources/s) [0s] 
                                                    

/ Syncing resources... (34/-, 68 resources/s) [0s] 
                                                    

- Syncing resources... (34/-, 68 resources/s) [0s] 
                                                    

\ Syncing resources... (34/-, 68 resources/s) [0s] 
                                                    

| Syncing resources... (34/-, 68 resources/s) [0s] 
                                                    

/ Syncing resources... (34/-, 68 resources/s) [0s] 
                                                    

- Syncing resources... (34/-, 68 resources/s) [1s] 
                                                    

\ Syncing resources... (34/-, 34 resources/s) [1s] 
                                                    

| Syncing resources... (34/-, 34 resources/s) [1s] 
                                                    

/ Syncing resources... (34/-, 34 resources/s) [1s] 
                                                    

- Syncing resources... (34/-, 34 resources/s) [1s] 
                                                    

\ Syncing resources... (34/-, 34 resources/s) [1s] 
                                                    

| Syncing resources... (34/-, 34 resources/s) [1s] 
                                                    

/ Syncing resources... (34/-, 23 resources/s) [1s] 
[...]

PR #12533, opened 2023-07-26
fix: Support syncing tables with a single column

erezrokah:
#### Summary

Discovered this while working on https://github.com/cloudquery/cloudquery/issues/12524.
If the table has a single column the PostgreSQL source fails to sync it

### DIFF
--- a/plugins/source/postgresql/client/sync.go
+++ b/plugins/source/postgresql/client/sync.go
@@ -108,7 +108,7 @@ func (c *Client) syncTables(ctx context.Context, snapshotName string, filteredTa
 }
 
 func syncTable(ctx context.Context, tx pgx.Tx, table *schema.Table, res chan<- message.SyncMessage) error {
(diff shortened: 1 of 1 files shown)


plugins/source/postgresql/client/sync.go
```diff
@@ -108,7 +108,7 @@ func (c *Client) syncTables(ctx context.Context, snapshotName string, filteredTa
 }
 
 func syncTable(ctx context.Context, tx pgx.Tx, table *schema.Table, res chan<- message.SyncMessage) error {
-	colNames := make([]string, 0, len(table.Columns)-2)
```
disq, line 111: Where do you think this came from? internal columns?

disq (approved):

plugins/source/postgresql/client/sync.go
erezrokah: Not sure, maybe 🤷 Maybe CoPilot?

PR #12536, opened 2023-07-26
feat(source-spec): Warn if ineffective options are detected

disq:
Didn't want to add these to `Source.Validate()` to error out just yet.

### DIFF
--- a/cli/internal/specs/v0/source.go
+++ b/cli/internal/specs/v0/source.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/thoas/go-funk"
+	"golang.org/x/exp/slices"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cli/internal/specs/v0/source.go
```diff
@@ -89,6 +90,13 @@ func (s *Source) GetWarnings() Warnings {
 	if s.ResourceConcurrency != 0 {
 		warnings["resource_concurrency"] = "the `resource_concurrency` option is deprecated. Please use the plugin-level concurrency option instead"
 	}
+	if s.SkipDependentTables && slices.Contains(s.Tables, "*") {
+		warnings["skip_dependent_tables"] = "the `skip_dependent_tables` option is ineffective when used with '*' `tables`"
+	}
+	if slices.Contains(s.Tables, "*") && len(s.Tables) > 1 {
```
erezrokah, line 96: Can't we have `tables: ["*", "!aws_ec*"] to exclude some tables?

disq: You'd need to add `aws_ec*` to `skip_tables`.

erezrokah (approved):

Issue #12404, opened 2023-07-19
bug (GCP): gcp_resourcemanager_folders is only collecting top level folders


chr-b:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The table `gcp_resourcemanager_folders` will only list those folders that are directly below the organization root.
E.g., assuming the following folder structure
```
org-root
|
| - A
   | - A-B
| - B
```
only folders `A` and `B` will be listed.
Folder `A-B` is missing.

In other words: not all existing folders are listed in this table.

### Expected Behavior

List all folders that exist in a GCP organization.

The problem is probably that the [GCP API folders.list](https://cloud.google.com/resource-manager/reference/rest/v2/folders/list) has an input parameter that can be either a) the organization ID or b) a folder ID.

When you only provide the organization ID as input, you will only retrieve the list of top-level folders.
The solution will require recursively iterating over all folders and use those folder IDs as input to obtain the children folder IDs.

### CloudQuery (redacted) config

n/a

### Steps To Reproduce

Run a CloudQuery sync for the table `gcp_resourcemanager_folders` in a GCP organization that has nested folders.

### CloudQuery (redacted) logs

n/a

### CloudQuery version

3.8.0

### Additional Context

GCP plugin v9.2.1

### Pull request (optional)

- [ ] I can submit a pull request

PR #12539, opened 2023-07-26
feat(resources): Fetch lower level folders

disq:
Fixes https://github.com/cloudquery/cloudquery/issues/12404

Adds `gcp_resourcemanager_subfolders` table to allow easy skipping of lower level folders.


erezrokah (commented):
~~I'm not sure if the requirement is to list one level down, or recursively all levels~~

Ah, noticed the recursion now

PR #12540, opened 2023-07-26
ci: Validate Scope in PR Title

bbernays:

### DIFF
--- a/.github/workflows/validate_pr_title.yml
+++ b/.github/workflows/validate_pr_title.yml
@@ -30,6 +30,10 @@ jobs:
             doc
           # Configure that a scope must always be provided.
           requireScope: false
+          scopes: |
+            deps
+            resources
+            main
           # Configure additional validation for the subject based on a regex.
           # This example ensures the subject starts with an uppercase character.
           subjectPattern: ^[A-Z].+$


erezrokah (commented):
Thanks for the PR @bbernays! Since the trigger for this PR is `pull_request_target` is will only start working after we merge it to `main`.

We should:
1. Verify that an empty scope is allowed
2. Go over are auto generated PRs and ensure their scope is valid. Searching for `schedule` in `.github/workflows` should work to find the relevant ones

bbernays:
> Thanks for the PR @bbernays! Since the trigger for this PR is `pull_request_target` is will only start working after we merge it to `main`.

Thanks! I have been trying to figure out why the current title `chore(non2main)` is being allowed

bbernays:
> Thanks for the PR @bbernays! Since the trigger for this PR is `pull_request_target` is will only start working after we merge it to `main`.
> 
> We should:
> 
> 1. Verify that an empty scope is allowed
After looking at the code and docs, because we do not require scope: `requireScope: false` an empty scope should still work
> 2. Go over are auto generated PRs and ensure their scope is valid. Searching for `schedule` in `.github/workflows` should work to find the relevant ones
I have searched through all of the automated PRs in the repo using the following regex `\(.*\): ` which found GCP and AWS automated PRs... I added another option to the scopes list `service` to handle the 2 automated prs

erezrokah (approved):

Issue #5195, opened 2022-11-30
File source plugin


cqgaurav:
This allows users to filter using arbitrary values that they can provide.

PR #12239, opened 2023-07-17
chore: Fix vale linting for buy pages

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/12219.
The buy pages are autogenerated during build, so they should be `git` ignored. I think https://github.com/cloudquery/cloudquery/pull/12219 removed the ignore by accident.

This PR also changes it so we use the ellipsis icon instead of three dots `...`

PR #12546, opened 2023-07-27
feat: Add premium file plugin to Website

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/5195

### TODO

- [x] Create stipe buy links, maybe @hermanschaaf can help me with that

### DIFF
--- a/plugins/source/aws/docs/tables/README.md
+++ b/plugins/source/aws/docs/tables/README.md
@@ -307,6 +307,8 @@
   - [aws_emr_cluster_instance_groups](../../../../../website/tables/aws/aws_emr_cluster_instance_groups.md)
   - [aws_emr_cluster_instances](../../../../../website/tables/aws/aws_emr_cluster_instances.md)
 - [aws_emr_security_configurations](../../../../../website/tables/aws/aws_emr_security_configurations.md)
+- [aws_emr_studios](../../../../../website/tables/aws/aws_emr_studios.md)
+  - [aws_emr_studio_session_mappings](../../../../../website/tables/aws/aws_emr_studio_session_mappings.md)
 - [aws_eventbridge_api_destinations](../../../../../website/tables/aws/aws_eventbridge_api_destinations.md)
 - [aws_eventbridge_archives](../../../../../website/tables/aws/aws_eventbridge_archives.md)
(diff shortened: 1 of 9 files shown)


website/components/pluginData.tsx
```diff
@@ -142,6 +142,18 @@ export const ALL_PLUGINS: Plugin[] = [
...
   },
+  {
+    name: "File",
+    logo: "/images/logos/plugins/file.svg",
+    id: "file",
+    kind: "source",
+    availability: "premium",
+    category: "cloud-finops",
+    buyLinks: {
+      'standard': 'https://buy.stripe.com/6oE28c1rHfRZaCA03r',
```
erezrokah, line 153: Need to update those with new links

website/pages/buy/file-extended.mdx
```diff
@@ -0,0 +1,15 @@
+---
```
hermanschaaf, line 1: Not sure if I made a mistake or if this is something that changed in the last week or so, but I thought these files don't need to be committed because they're generated as part of the prebuild step?

website/pages/docs/plugins/sources/file/overview.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+### Example
+
+This example configures a File source with a directory to sync files from. The (top level) spec section is described in the [Source Spec Reference](/docs/reference/source-spec).
+
+<Configuration/>
+
+## File spec
+
+This is the (nested) spec used by the MySQL destination plugin.
```
hermanschaaf, line 28:
```suggestion
This is the (nested) spec used by the File source plugin.
```

website/pages/buy/file-extended.mdx
erezrokah: Discussed in https://github.com/cloudquery/cloudquery/pull/12239. The `buy` files have a manually generated file too now, so until we separate it from the auto generated ones we need to commit them

hermanschaaf:
Looks good to me but I think the Stripe links still need to be updated to the new ones in the committed pages?

erezrokah:
> Looks good to me but I think the Stripe links still need to be updated to the new ones in the committed pages?

I committed the new files, good catch 🚀

hermanschaaf (approved):

PR #12547, opened 2023-07-27
fix: Align premium plugins docs and versions

erezrokah:
#### Summary

While working on https://github.com/cloudquery/cloudquery/pull/12546, I noticed that the way we display premium plugins on the Website is not consistent:

1. In the [sources overview page](https://www.cloudquery.io/docs/plugins/sources/overview) some show with the latest version of the free plugin, while some show as "Unpublished":
image

2. In the [sources overview page](https://www.cloudquery.io/docs/plugins/sources/overview) the GA premium plugins stage is `GA (Premium)`, while the preview premium plugin' (Jira) stage is `Preview (Paid)`

3. In some premium plugins configuration we use `version: "PREMIUM"`, and in others we point to the latest version of the free plugin.

4. Most premium plugins configurations are missing the `registry: local` property

**This PR has the following fixes:**

1. In both the sources overview page and plugins configurations, the version for premium plugins is `PREMIUM`.

2. Changes the only `Preview (Paid)` plugin (Jira) to `Preview (PREMIUM)`

3. Removes the version JSON files for premium plugins as those are not needed anymore

4. Adds `registry: local` when it was missing

5. Adds a note in the Tailscale blog post that it is now a premium plugin


website/pages/blog/cloudquery_tailscale.mdx
```diff
@@ -38,8 +45,9 @@ kind: source
 # Common source-plugin configuration
 spec:
   name: tailscale
-  path: cloudquery/tailscale
-  version: "VERSION_SOURCE_TAILSCALE"
+  path: /path/to/downloaded/plugin # Buy from here: https://cloudquery.io/integrations/tailscale
+  registry: local
+  version: "VERSION_PREMIUM"
```
disq, line 50: Not sure about these... just resolves to `version: "PREMIUM"` currently

erezrokah: Yeah, premium plugins don't really have SemVer versions. This is similar to the `Render` and `Jira` plugins.
I don't mind what we put there (I initially put `v1.0.0`) as long as we're consistent

erezrokah: But this should also be `"PREMIUM"` and not `"VERSION_PREMIUM"`

yevgenypats (approved):

PR #4341, opened 2022-11-10
fix(aws): Ensure all AWS table names are plural

bbernays:
BREAKING CHANGE: Renamed `aws_apprunner_auto_scaling_configuration` to `aws_apprunner_auto_scaling_configurations`
BREAKING CHANGE: Renamed `aws_apprunner_vpc_connector` to `aws_apprunner_vpc_connectors`
BREAKING CHANGE: Renamed `aws_apprunner_vpc_ingress_connection` to `aws_apprunner_vpc_ingress_connections`
BREAKING CHANGE: Renamed `aws_ec2_regional_config` to `aws_ec2_regional_configs`
BREAKING CHANGE: Renamed `aws_lightsail_disk_snapshot` to `aws_lightsail_disk_snapshots`
BREAKING CHANGE: Renamed `aws_xray_encryption_config` to `aws_xray_encryption_configs`

PR #12549, opened 2023-07-27
test: Add Test to validate Table names are plural

bbernays:
#### Summary

This brings back a test we had implemented [in our codegen](https://github.com/cloudquery/cloudquery/pull/4341), but got removed once we got rid of the codegen stuff


plugins/source/aws/resources/plugin/plugin_test.go
```diff
@@ -15,12 +16,15 @@ func TestAWS(t *testing.T) {
...
 
-// This test ensures that all tables have a unique description.
-func TestAWSTableDescriptions(t *testing.T) {
+// This test ensures that all tables have proper name and description.
+func TestAWSTables(t *testing.T) {
 	descriptions := make(map[string]string)
 	tables := getTables().FlattenTables()
 	for _, table := range tables {
-		if ignoreTable(table.Name) {
+		if !ignorePluralName(table.Name) && !strings.HasSuffix(table.Name, "s") {
```
erezrokah, line 24: We might have discussed this before, but should we use https://github.com/gertd/go-pluralize/tree/master for this? It covers more use cases

erezrokah:
Looks good, I updated the PR title to `test:*` since this is not a user facing change

Issue #7708, opened 2023-02-06
AWS EMR Additional Resources: Notebook Executions, Release Labels, Steps, Studio Session mappings, Studios


jsonpr:
- [ ] Notebook Executions: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/emr/list-notebook-executions.html
- [ ] Release Labels: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/emr/list-release-labels.html
- [ ] Steps: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/emr/list-steps.html
- [x] Studio and Studio Mapping: https://github.com/cloudquery/cloudquery/issues/12528
- [ ] Commands
- [ ] Instance Type Specifications

PR #12562, opened 2023-07-28
feat(resources): Adding additional EMR cluster resources

mnorbury:
#### Summary

Adding further EMR resources to support notebook executions, cluster steps, release labels and supported instance types.

Fixes: https://github.com/cloudquery/cloudquery/issues/7708


plugins/source/aws/resources/services/emr/release_labels.go
```diff
@@ -0,0 +1,60 @@
...
+func getReleaseLabel(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services().Emr
+	releaseLabel := resource.Item.(string)
+	response, err := svc.DescribeReleaseLabel(ctx, &emr.DescribeReleaseLabelInput{ReleaseLabel: &releaseLabel}, func(options *emr.Options) {
+		options.Region = cl.Region
+	})
+	if err != nil {
+		return err
+	}
```
bbernays, line 57: I think this is a paginated API, so in this implementation you will just get the first page

bbernays (commented):
PR looks good. one comment about the code

yevgenypats (approved):

website/tables/aws/aws_emr_notebook_executions.md
```diff
@@ -0,0 +1,38 @@
+# Table: aws_emr_notebook_executions
+
+This table shows data for Amazon EMR Notebook Executions.
+
+https://docs.aws.amazon.com/emr/latest/APIReference/API_NotebookExecution.html
+
+The composite primary key for this table is (**cluster_arn**, **arn**).
```
hermanschaaf, line 7: Isn't the `arn` alone enough in this case? Why include `cluster_arn` in the PK as well?

plugins/source/aws/resources/services/emr/clusters.go
```diff
@@ -37,7 +37,7 @@ func Clusters() *schema.Table {
 				Resolver: client.ResolveTags,
 			},
 		},
-		Relations: []*schema.Table{clusterInstanceFleets(), clusterInstanceGroups(), clusterInstances()},
+		Relations: []*schema.Table{clusterInstanceFleets(), clusterInstanceGroups(), clusterInstances(), notebookExecutions(), steps()},
```
disq, line 40: I would update this to be one table() per line so that further modifications are easily visible and mergable

website/tables/aws/aws_emr_notebook_executions.md
hermanschaaf: Yeah, I think keeping `cluster_arn` but with PK set to false should do the trick 👍

PR #12565, opened 2023-07-28
feat(resources): Add square plugin in python

yevgenypats:


python-plugins/square/plugin/tables/payments.py
```diff
@@ -0,0 +1,33 @@
...
+            Column("id", pa.string(), primary_key=True),
+            Column("status", pa.string()),
+            Column("order_id", pa.string()),
+        ])
+
+class PaymentsResolver(TableResolver):
+    def __init__(self) -> None:
+        super().__init__(Payments())
+    
+    def resolve(self, client: Client, parent_resource) -> Generator[Any, None, None]:
```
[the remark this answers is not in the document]
yevgenypats: because it can be either a dict or a dataclass as the scheduler tries to resolve it automatically.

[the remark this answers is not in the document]
hermanschaaf: Oh I see, sorry I was confused by the signature having to `None`s, but looks like it's a Generator-specific thing https://docs.python.org/3/library/typing.html#typing.Generator

python-plugins/square/plugin/tables/merchants.py
```diff
@@ -0,0 +1,47 @@
...
+                Column("language_code", pa.string()),
+                Column("currency", pa.string()),
+                Column("status", pa.string()),
+                Column("main_location_id", pa.string()),
+                Column("created_at", pa.timestamp("ms")),
+            ],
+        )
+
+    @property
+    def resolverClass(self):
```
disq, line 30: There's already a `resolver` defined in Table
```python
    def resolver(self, client: Client, parent=None) -> Generator[Any]:
        raise NotImplementedError()
```
It fits the signature of `resolve` below, so I can do something like:
```python
    def resolver(self, client: Client, parent=None) -> Generator[Any]:
        return MerchantsResolver(self).resolve(client, parent)
```

But then the `self._scheduler.sync(self._client, resolvers)` is not possible because sync expects `List[TableResolver]` which we now can't have.

Maybe remove resolver abstractclass from Table?

yevgenypats: where do we have resolver on table? if we have we might want to resolve it as from what I recall it causes bunch of conflicts and circular dependencies.

disq: @yevgenypats https://github.com/cloudquery/plugin-sdk-python/blob/main/cloudquery/sdk/schema/table.py#L39

yevgenypats: Oh seems like old stuff. should remove prob

disq: ok fixed here, will update SDK later

PR #12566, opened 2023-07-28
chore: Fix test release workflow

erezrokah:
#### Summary

Put the `Add Docker builder` in the correct place


erezrokah:
Also reverts https://github.com/cloudquery/cloudquery/pull/12564 so we can re-release

PR #12570, opened 2023-07-28
chore: Disable docker Windows image for source-test

erezrokah:


erezrokah:
Also reverts the latest release so we can retag it

PR #12599, opened 2023-07-31
chore: Move policies to main navbar

yevgenypats:

PR #12600, opened 2023-07-31
chore: Add some more scopes to PR validator

erezrokah:
#### Summary

We usually use these for `chore` PRs for clarity


erezrokah:
Example in https://github.com/cloudquery/cloudquery/pull/12599

hermanschaaf (approved):

Issue #12577, opened 2023-07-28
Sync azure (v9.1.0) -> [mssql (v4.3.2)] fails - write client returned error (insert)


rafaelrodrigues3092:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After upgrading the source (azure) and target plugins (mssql), the sync fails with error:

Error: failed to sync v3 source azure: write client returned error (insert): error reading from server: read unix @->C:\Users\rdadmin\AppData\Local\Temp\cq-srPiAsIdsKzFQWEm.sock: wsarecv: An existing connection was forcibly closed by the remote host.

Sync was working before the upgrade with versions azure (v7.1.1) -> [mssql (v3.0.4)]. Database tables and objects are created without issue.

Things already verified:
- SQL Server permissions (login has db_owner on cloudquery database)
- Tables, user-defined table types, and stored procedures are created OK
- There are no user-created dependent objects
- Connection string format (tried different options - see below in config)
- There are other tools accessing the same database server (different database)  from the same location with no issues. Previous versions of plugins were OK.

### Expected Behavior

Expecting a successful sync

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: azure
  path: cloudquery/azure
  version: "v9.1.0"
  destinations: [
    "mssql"
  ]
  concurrency: 100000
  skip_dependent_tables: true
  tables: [
    "azure_advisor_recommendation_metadata",
    "azure_advisor_recommendations",
    "azure_advisor_suppressions",
    "azure_appservice_certificate_orders",
    "azure_appservice_certificates",
    "azure_appservice_deleted_web_apps",
    "azure_appservice_domains",
    "azure_appservice_environments",
    "azure_appservice_plans",
    "azure_appservice_recommendations",
    "azure_appservice_resource_health_metadata",
    "azure_appservice_static_sites",
    "azure_appservice_top_level_domains",
    "azure_appservice_web_apps",
    "azure_appservice_web_app_configurations",
    "azure_appservice_web_app_vnet_connections",
    "azure_compute_availability_sets",
    "azure_compute_capacity_reservation_groups",
    "azure_compute_capacity_reservations",
    "azure_compute_cloud_services",
    "azure_compute_disk_accesses",
    "azure_compute_disk_encryption_sets",
    "azure_compute_disks",
    "azure_compute_galleries",
    "azure_compute_gallery_images",
    "azure_compute_gallery_image_versions",
    "azure_compute_images",
[...]

PR #12645, opened 2023-07-31
fix: Don't use row index for accessing chunk data

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12577 (can also be reproduces by syncing only `azure_compute_skus`)

It was a bit hard to for me to understand the code, but it seems we use `NewTableFromRecords` to map an array of records with length of `batch_size` to a single Arrow table. This means the resulting table will have `batch_size` number of rows, with each column having `number_of_schema_columns` chunks, where each chunk of arrays is of `batch_size` length. Each array that is a part of the chunk of arrays has the underlying data from the source (think of a chunk of a way to manage a collection of arrays).

See
https://github.com/cloudquery/arrow/blob/b7ee35bf784acdf7af50f7776ce8b9dd8117a4c2/go/arrow/array/table.go#L177
https://github.com/cloudquery/arrow/blob/b7ee35bf784acdf7af50f7776ce8b9dd8117a4c2/go/arrow/array/table.go#L193

The current code increments the row index for each array in the chunk:
https://github.com/cloudquery/cloudquery/blob/5fa1af103a6e0e0b62d05f4d5d6f902c23a9e03e/plugins/destination/mssql/queries/values.go#L17

and for each value in the chunk data array:
https://github.com/cloudquery/cloudquery/blob/5fa1af103a6e0e0b62d05f4d5d6f902c23a9e03e/plugins/destination/mssql/queries/values.go#L18

then uses that value as an index to access the data in each chunk.

As mentioned before each column's chunk references `batch_size` number of arrays so we cannot use that length as a indicator on how to access each array's data.

**The fix is to use the `row` index as the index to the preallocated row array, and `0` as the index to access the underlying data of each array in the chunk. This works since sources don't save data in arrays longer than 1, so we know there's always a single value.**

> Technically we should be "unwrapping" each data array, but that can be done in a follow up (and maybe remove the usage of `NewTableFromRecords` as we could create the query rows directly from the records array).

### DIFF
--- a/plugins/destination/mssql/queries/values.go
+++ b/plugins/destination/mssql/queries/values.go
@@ -14,14 +14,13 @@ func GetRows(table arrow.Table) ([][]any, error) {
 	for n := 0; n < int(table.NumCols()); n++ {
 		col := table.Column(n)
 		row := 0
-		for _, chunk := range col.Data().Chunks() {
-			for i := 0; i < chunk.Len(); i++ {
-				rows[row][n], err = getColValue(chunk, row)
-				if err != nil {
-					return nil, err
-				}
-				row++
+		for _, chunkArray := range col.Data().Chunks() {
+			// We only support records with arrays of length == 1
+			rows[row][n], err = getColValue(chunkArray, 0)
+			if err != nil {
+				return nil, err
 			}
+			row++
 		}
 	}
 	return rows, nil
--- a/plugins/destination/mssql/queries/values.go
+++ b/plugins/destination/mssql/queries/values.go
@@ -15,7 +15,7 @@ func GetRows(table arrow.Table) ([][]any, error) {
 		col := table.Column(n)
 		row := 0
 		for _, chunkArray := range col.Data().Chunks() {
-			// We only support records with arrays of length == 1
+			// We only support records with arrays of length == 1, so we always extract the first value
 			rows[row][n], err = getColValue(chunkArray, 0)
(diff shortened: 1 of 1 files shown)


plugins/destination/mssql/queries/values.go
```diff
@@ -14,14 +14,13 @@ func GetRows(table arrow.Table) ([][]any, error) {
 	for n := 0; n < int(table.NumCols()); n++ {
 		col := table.Column(n)
 		row := 0
-		for _, chunk := range col.Data().Chunks() {
-			for i := 0; i < chunk.Len(); i++ {
-				rows[row][n], err = getColValue(chunk, row)
```
disq, line 19: This comment might not make sense, but why not `getColValue(chunk, i)` instead?

erezrokah: Yeah, overwriting the row might not make sense either (even if it only runs a single iteration every time)

erezrokah: Maybe `We only support records with arrays of length == 1, so we always extract the first value`

disq (approved):

plugins/destination/mssql/queries/values.go
candiduslynx, line 19: this was a one-liner of 
```go
rows[row][n], err = getColValue(chunk, i)
```

Issue #12643, opened 2023-07-31
Cyclic dependency in scaffold source plugin


mnorbury:
When creating a new source plugin using the Cloudquery scaffold CLI:

```bash
cq-scaffold source cloudquery <name>
```

The API client is used in the [resources](https://github.com/cloudquery/cloudquery/blob/main/scaffold/cmd/templates/source/resources/table.go.tpl#L24-L26) package by casting to the `client.Client` type from the `client` package. However, the `client` package now has a dependency on the [resources](https://github.com/cloudquery/cloudquery/blob/main/scaffold/cmd/templates/source/client/client.go.tpl#L77-L88) package since this is where the `schema.Tables` collection is assembled. This results in a cyclic dependency between `client` and `resources` package.

PR #12649, opened 2023-07-31
fix: Create separate plugin client package

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12643

Had to refactor the templating engine code a bit as now we have different template files with the same name (`client.go.tpl` under `client` and under `resources/plugin`.

Go templating engine doesn't load templates recursively so I used `fs.WalkDir` to avoid having to manually map template files to output files.

### DIFF
--- a/scaffold/cmd/source.go
+++ b/scaffold/cmd/source.go
@@ -4,8 +4,10 @@ import (
 	"embed"
 	"fmt"
 	"go/format"
+	"io/fs"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"text/template"
 
@@ -33,21 +35,6 @@ func newCmdScaffoldSource() *cobra.Command {
 	return cmd
 }
 
-var scaffoldTemplates = map[string]string{
-	"release.yaml.tpl":     ".github/workflows/release.yaml",
-	"test.yaml.tpl":        ".github/workflows/test.yaml",
-	".goreleaser.yaml.tpl": ".goreleaser.yaml",
-	"go.mod.tpl":           "go.mod",
-	"main.go.tpl":          "main.go",
-	"Makefile.tpl":         "Makefile",
-	"README.md.tpl":        "README.md",
-	"client.go.tpl":        "client/client.go",
-	"spec.go.tpl":          "client/spec.go",
-	"plugin.go.tpl":        "plugin/plugin.go",
-	"table.go.tpl":         "resources/table.go",
-	".gitignore.tpl":       ".gitignore",
-}
-
 //go:embed templates/source/*
 var sourceFS embed.FS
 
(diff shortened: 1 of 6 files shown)


scaffold/cmd/source.go
```diff
@@ -57,44 +44,47 @@ type scaffoldData struct {
...
-	)
-	if err != nil {
-		return fmt.Errorf("failed to parse templates: %w", err)
-	}
-
 	data := scaffoldData{
 		Org:  org,
 		Name: name,
 	}
-	for templatePath, filePath := range scaffoldTemplates {
```
erezrokah, line 75: Before we had an hardcoded map of template names filenames to output files. This didn't allow different templates with the same base name of a file (e.g. `client/client.go.tpl` and `plugin/client.go.tpl`)

scaffold/cmd/source.go
```diff
@@ -57,44 +44,47 @@ type scaffoldData struct {
...
-		baseDir := path.Dir(fullPath)
-		if err := os.MkdirAll(baseDir, 0755); err != nil {
-			return fmt.Errorf("failed to create directory %s: %w", baseDir, err)
+		if d.IsDir() {
+			return nil
 		}
-		if strings.HasSuffix(filePath, ".go") {
-			formattedContent, err := format.Source(content)
+		if strings.HasSuffix(templatePath, ".tpl") {
+			tpl, err := template.New(filepath.Base(templatePath)).ParseFS(sourceFS, templatePath)
```
erezrokah, line 60: Mostly the same code as before, and now getting the output path based on the template filename

scaffold/cmd/templates/source/client/client.go.tpl
```diff
@@ -2,25 +2,13 @@ package client
 
 import (
 	"context"
-	"encoding/json"
```
erezrokah, line 5: Most of the code here was moved to `plugin/client.go.tpl`

scaffold/cmd/templates/source/resources/plugin/client.go.tpl
```diff
@@ -0,0 +1,88 @@
...
+	logger    zerolog.Logger
+	config    client.Spec
+	tables    schema.Tables
+	syncClient *client.Client
+	scheduler *scheduler.Scheduler
+
+	plugin.UnimplementedDestination
+}
+
+func New(ctx context.Context, logger zerolog.Logger, spec []byte, opts plugin.NewClientOptions) (plugin.Client, error) {
```
disq, line 28: Maybe call this `Configure` so it doesn't get confused with `client.New`? Currently we have 13 `New*` vs. 21 `Configure` in the monorepo source plugins.

disq (approved):

PR #12670, opened 2023-08-01
fix: Detecting conditions for CIS AWS v1.5.0 Section 1

aruneko:
#### Summary
I have fixed some detecting conditions for CIS AWS v1.5.0 Section 1.

- Section 1.9
    - Fixed `case` condition.
    - The valid value of `password_reuse_prevention` is 1 to 24 (ref [official doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_passwords_account-policy.html#password-policy-details)).
    - The CIS Benchmarks says `Ensure "Number of passwords to remember" is set to 24`.
- Section 1.15
    - Consider the case of inline policy against to an IAM user.


erezrokah (approved):
Thanks @aruneko 🚀

Issue #12652, opened 2023-07-31
bug: DeleteStale from V0/v1 Sources Don't Work with V3 Destinations 


bbernays:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

V0/V1 Sources are able to correctly write data to V3 destinations. But the `DeleteStale` function is never called.

In [earlier versions](https://github.com/cloudquery/plugin-sdk/blob/8a6a2cc87e303d66c810e629a497f9acd3c4982c/internal/servers/destination/v1/destinations.go#L103) of the destination, the destination plugin was responsible for calling `DeleteStale` at the end of the `Write()` call. This functionality was replaced with a DeleteStale Message that is now generated by the CLI. In the V0/V1 backwards [compatible servers](https://github.com/cloudquery/plugin-sdk/blob/17049ae7594e0f6733e9aa216c1b2926d3c00682/internal/servers/destination/v1/destinations.go#L103-L190) we never actually generate that new DeleteStale Message

### Expected Behavior

DeleteStale should be called at the end of the sync for each table

### CloudQuery (redacted) config

A v2 source with a v3 destination

### Steps To Reproduce

Run a sync with an old source and the newest destination

### CloudQuery (redacted) logs

N/A

### CloudQuery version

cloudquery version 3.11.0

### Pull request (optional)

- [ ] I can submit a pull request

PR #12684, opened 2023-08-01
fix: Add missing `DeleteStale` messages for old destinations

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12652


cli/cmd/sync_v1.go
```diff
@@ -164,6 +164,20 @@ func syncConnectionV1(ctx context.Context, sourceClient *managedplugin.Client, d
...
+				Tables:    tablesRes.Tables,
+				Source:    sourceSpec.Name,
+				Timestamp: timestamppb.New(syncTime),
+			})
+			if err != nil {
+				return err
+			}
+		}
+	}
+
```
candiduslynx, line 180:
```suggestion

	deleteStaleMsg := &destination.DeleteStale_Request{
		Tables:    tablesRes.Tables,
		Source:    sourceSpec.Name,
		Timestamp: timestamppb.New(syncTime),
	}

	for i := range destinationsClients {
		if destinationSpecs[i].WriteMode == specs.WriteModeOverwriteDeleteStale {
			if _, err := destinationsPbClients[i].DeleteStale(ctx, deleteStaleMsg); err != nil {
				return err
			}
		}
	}

```

cli/cmd/sync_v2.go
```diff
@@ -153,6 +153,17 @@ func syncConnectionV2(ctx context.Context, sourceClient *managedplugin.Client, d
 		}
 	}
 	for i := range destinationsClients {
+		if destinationSpecs[i].WriteMode == specs.WriteModeOverwriteDeleteStale {
```
candiduslynx, line 156: same here, alloc `deleteStaleMsg` outside of loop, reuse)

erezrokah: Personally I find it easier to follow when the allocation is close to where it's used, and I don't think we expect `destinationsClients` to be very large

disq (approved):

Issue #12674, opened 2023-08-01
docs: Update docs for `backend`, `backend_spec `, `scheduler` and `concurrency` spec options


erezrokah:
With the v4 update those were moved to the source plugin spec from the top level to the source plugin spec.
We should update the docs as using one of the above options logs a warning (except `backend_spec` which is a separate bug to fix in the CLI).

PR #12686, opened 2023-08-01
fix(website): Move deprecated top level spec options to plugin level

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/12674 + destinations options + some indentation fixes.
I chose to keep the options on the same page and only move them to the bottom.
This way links to the options don't break.

We should also add the specific options to each plugin, but that's more time consuming


website/pages/docs/plugins/sources/firestore/_configuration.mdx
```diff
@@ -5,9 +5,8 @@ spec:
   name: firestore
   path: cloudquery/firestore
   version: "VERSION_SOURCE_FIRESTORE"
-  concurrency: 10000
```
erezrokah, line 8: Not supported in Firestore (like most other DB sources)

website/pages/docs/plugins/sources/mysql/_configuration.mdx
```diff
@@ -4,9 +4,8 @@ spec:
   name: "mysql"
   path: "cloudquery/mysql"
   version: "VERSION_SOURCE_MYSQL"
-  concurrency: 10000
```
erezrokah, line 7: Not supported in MySQL (like most other DB sources)

website/pages/docs/reference/source-spec.md
```diff
@@ -142,3 +126,28 @@ If set to `true`, the exporter will not verify the server will connect via `http
...
+
+(`int`, optional, default: `<unique-per-plugin>`, introduced in CLI `v3.6.0`)
+
+A best effort maximum number of Go routines to use. Lower this number to reduce memory usage.
+
+#### scheduler
+
+(`string`, optional, default: `dfs`, introduced in CLI `v3.6.0`, **EXPERIMENTAL**)
+
+The scheduler to use when determining the priority of resources to sync. Currently, the only supported values are `dfs` (depth-first search) and `round-robin`. This is an experimental feature, and may be removed in the future. For more information about this, see [performance tuning](/docs/advanced-topics/performance-tuning).
```
hermanschaaf, line 141: Should concurrency and scheduler still be documented here, if they're part of the plugin spec?

website/pages/docs/reference/source-spec.md
```diff
@@ -142,3 +126,28 @@ If set to `true`, the exporter will not verify the server will connect via `http
 (`object`, optional)
 
 Plugin-specific configurations. Visit [source plugins](/docs/plugins/sources/overview) documentation for more information.
+The following options are available for most source plugins:
```
hermanschaaf, line 129: Oh I missed this line. I think we should remove it from this page and move it to each plugin page where applicable (but understand if you don't want to do it in this PR necessarily)

hermanschaaf: Maybe this line should be under a new heading at least

website/pages/docs/reference/source-spec.md
```diff
@@ -142,3 +126,28 @@ If set to `true`, the exporter will not verify the server will connect via `http
...
+
+The scheduler to use when determining the priority of resources to sync. Currently, the only supported values are `dfs` (depth-first search) and `round-robin`. This is an experimental feature, and may be removed in the future. For more information about this, see [performance tuning](/docs/advanced-topics/performance-tuning).
+
+#### backend
+
+(`string`, optional, default: `local`, introduced in CLI `v3.6.0`)
+
+The backend to use for storing the state of incremental tables. Currently, the only supported value is `local` (store the state in the local filesystem). For more information, see [managing incremental tables](/docs/advanced-topics/managing-incremental-tables).
+
+#### backend_spec
```
hermanschaaf, line 149: `backend_spec` is deprecated since v4, it's now called `backend_options`, and contains `table_name` and `connection`. It supports referencing values from other plugins, as documented in the blog post: https://www.cloudquery.io/blog/proto-v3 

Would be great to update the documentation for this here

website/pages/docs/reference/source-spec.md
erezrokah: I think I'll do it in a new heading. The reasons I kept them here are:
1. It's not a small effort going through each plugin and understand which options it uses, how and if the defaults have changed
2. Avoid having broken backlinks to this page. We can't fix those using redirects as the specific options are referenced  using `#`, so we'll need to do some client side routing to fix those.

plugins/source/alicloud/client/spec.go
```diff
@@ -16,6 +16,9 @@ type AccountSpec struct {
 }
 
 func (s *Spec) SetDefaults() {
+	if s.Concurrency == 0 {
```
erezrokah, line 19: The default was missing from this plugin

plugins/source/cloudflare/client/spec.go
```diff
@@ -1,7 +1,7 @@
 package client
 
 const (
-	defaultConccurency = 10000
+	defaultConcurrency = 10000
```
erezrokah, line 4: Typo

plugins/source/datadog/client/spec.go
```diff
@@ -11,3 +11,9 @@ type Account struct {
 	AppKey string `json:"app_key"`
 	APIUrl string `json:"api_url,omitempty"`
 }
+
+func (s *Spec) SetDefaults() {
+	if s.Concurrency == 0 {
+		s.Concurrency = 10000
```
erezrokah, line 17: Added missing default

erezrokah:
OK so I went through each plugin's implementation and updated the docs for each plugin with the relevant `concurrency`, `scheduler` (only GCP) and `backend_options` options and defaults.

I also remove the `concurrency: 10000` we had in many configurations as now each plugin has a sensible default that should work without needing to override it.

hermanschaaf (approved):

PR #12689, opened 2023-08-01
chore(website): Update Shopify link

erezrokah:
#### Summary

Looks like this link is a redirect now and fails our broken links check


disq (approved):

erezrokah:
Ah, still broken, I'll exclude it

PR #12690, opened 2023-08-01
chore(website): Exclude Shopify link

erezrokah:
#### Summary

Our broken links checker is not happy about this link for some reason. Follow up to https://github.com/cloudquery/cloudquery/pull/12689


disq (approved):
Or just don't link it :)

PR #12691, opened 2023-08-01
fix: Bring chunk traversal back

candiduslynx:
Follow-up for #12645.
See https://github.com/cloudquery/cloudquery/pull/12645/files#r1280845493 for more details

### DIFF
--- a/plugins/destination/mssql/queries/values.go
+++ b/plugins/destination/mssql/queries/values.go
@@ -16,10 +16,10 @@ func GetRows(table arrow.Table) ([][]any, error) {
 		for _, chunk := range col.Data().Chunks() {
 			for i := 0; i < chunk.Len(); i++ {
 				rows[row][c], err = getColValue(chunk, i)
+				row++
 				if err != nil {
 					return nil, err
 				}
-				row++
 			}
 		}
 	}
--- a/plugins/destination/mssql/queries/values.go
+++ b/plugins/destination/mssql/queries/values.go
@@ -11,16 +11,16 @@ func GetRows(table arrow.Table) ([][]any, error) {
 	rows := prealloc(table.NumRows(), table.NumCols())
 	var err error
 
-	for n := 0; n < int(table.NumCols()); n++ {
-		col := table.Column(n)
-		row := 0
-		for _, chunkArray := range col.Data().Chunks() {
-			// We only support records with arrays of length == 1, so we always extract the first value
-			rows[row][n], err = getColValue(chunkArray, 0)
-			if err != nil {
-				return nil, err
+	for c := 0; c < int(table.NumCols()); c++ {
+		row, col := 0, table.Column(c)
+		for _, chunk := range col.Data().Chunks() {
+			for i := 0; i < chunk.Len(); i++ {
+				rows[row][c], err = getColValue(chunk, i)
+				if err != nil {
+					return nil, err
+				}
+				row++
 			}
-			row++
 		}
 	}
 	return rows, nil
--- a/plugins/destination/mssql/queries/values_test.go
+++ b/plugins/destination/mssql/queries/values_test.go
@@ -1,0 +1,45 @@
+package queries
+
+import (
+	"github.com/apache/arrow/go/v13/arrow"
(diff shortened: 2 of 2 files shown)


plugins/destination/mssql/queries/values.go
```diff
@@ -11,16 +11,16 @@ func GetRows(table arrow.Table) ([][]any, error) {
...
-		for _, chunkArray := range col.Data().Chunks() {
-			// We only support records with arrays of length == 1, so we always extract the first value
-			rows[row][n], err = getColValue(chunkArray, 0)
-			if err != nil {
-				return nil, err
+	for c := 0; c < int(table.NumCols()); c++ {
+		row, col := 0, table.Column(c)
+		for _, chunk := range col.Data().Chunks() {
+			for i := 0; i < chunk.Len(); i++ {
+				rows[row][c], err = getColValue(chunk, i)
```
erezrokah, line 18: Doesn't this mean we keep overriding `rows[row][c]` per each chunk value?

candiduslynx: `row` is increased after each `getColValue` call

candiduslynx: Basically, we just fill in every row at index `c`

erezrokah: But what happens if `chunk.Len() > 1`. Won't we override the same value over again?
Can you add a test when `chunk.Len() > 1`?

candiduslynx: We do increment `c` in the outer loop, but that's for the index of `col` in the `table` only.
Basically, we do (`if we assume `table.Columns() to exist & return the slice of the corresponding chunked arrays):
```go
for colIndex, col := range table.Columns() {
	row := 0
	for _, chunk := range col.Data().Chunks() {
		for i := 0; i < chunk.Len(); i++ {
			data[row][colIndex] = getVal(chunk, i)
			row++
		}
	}
}
```

I've tested with `azure_compute_skus` and it worked fine.

candiduslynx: so, when `chunk.Len() > 1` we increment `row` var value right after extracting the `i`th value from the `chunk`

candiduslynx: I've moved the `row` increment closer to the value extraction, so that it's easier to understand: 67c2fef

candiduslynx: There's only 1 single place where we increment `row` variable (right after `getColValue(chunk, i)` call).
Could you point me the other one?

As for this comment:
> the number of allocated rows is always equal to the number of records in the batch

We do allocate the `table.NumRows()` times `table.NumCols()` which corresponds to the `total rows in all of the records combined` times `all the columns in the schema` per the Arrow implementation.

So, if we have 3 records 100 rows each (& the schema with 7 columns), we'll `prealloc` `300` rows times `7` columns.

candiduslynx: test added in a5b1e0f

erezrokah: I meant `row` will be incremented `len(col.Data().Chunks()) * chunk.Len()` times

candiduslynx: > I meant `row` will be incremented `len(col.Data().Chunks()) * chunk.Len()` times for each column

yes, that's intended, as we'll allocate that amount or fows anyways (assuming all chunks are of the same len).
If the chunks are of different len then we'll alloc the total rows for all of the records & we'll increment `row` var the same amount of times

erezrokah: Thank you for adding the test. It makes sense now. `NewTableFromRecords` flattens the data in each chunk. That's the piece I was missing 👍

candiduslynx: I've added a non-trivial test, too in [1e90aa4](https://github.com/cloudquery/cloudquery/pull/12691/commits/1e90aa4c24f72ee3706c137a378dd4e32deda789)

Issue #12694, opened 2023-08-01
bug(scheduler): Scheduler option has no impact


erezrokah:
We need to add it to all source plugins spec's and use it when creating a new scheduler

PR #12696, opened 2023-08-01
fix: Add and use scheduler option

erezrokah:
#### Summary

Discovered while working on https://github.com/cloudquery/cloudquery/issues/12674

Fixes https://github.com/cloudquery/cloudquery/issues/12694.

~~Added the option to all plugins except the DB based ones as those don't have a scheduler. Also didn't add it to Hacker News as it has a single table.~~

~~Still needs to test it on a plugin (e.g. GCP)~~

Only did this for GCP and test it for now

### DIFF
--- a/plugins/source/awspricing/client/spec.go
+++ b/plugins/source/awspricing/client/spec.go
@@ -1,15 +1,10 @@
 package client
 
-import (
-	"github.com/cloudquery/plugin-sdk/v4/scheduler"
-)
-
 type Spec struct {
 	RegionCodes []string `json:"region_codes,omitempty"`
 	OfferCodes  []string `json:"offer_codes,omitempty"`
 
-	Concurrency int                `json:"concurrency,omitempty"`
-	Scheduler   scheduler.Strategy `json:"scheduler,omitempty"`
+	Concurrency int `json:"concurrency,omitempty"`
 }
 
 func (s *Spec) SetDefaults() {
--- a/plugins/source/test/client/spec.go
+++ b/plugins/source/test/client/spec.go
@@ -1,21 +1,12 @@
 package client
 
-import (
-	"github.com/cloudquery/plugin-sdk/v4/scheduler"
-)
-
 type Spec struct {
-	NumClients  int                `json:"num_clients"`
-	Concurrency int                `json:"concurrency,omitempty"`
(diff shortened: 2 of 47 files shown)


plugins/source/test/client/spec.go
```diff
@@ -1,13 +1,22 @@
 package client
 
+import (
+	"github.com/cloudquery/plugin-sdk/v4/scheduler"
+)
+
 type Spec struct {
-	NumClients int `json:"num_clients"`
+	NumClients  int                `json:"num_clients"`
+	Concurrency int                `json:"concurrency,omitempty"`
```
erezrokah, line 9: Added these for reference

plugins/source/awspricing/client/spec.go
```diff
@@ -1,10 +1,15 @@
...
+import (
+	"github.com/cloudquery/plugin-sdk/v4/scheduler"
+)
+
 type Spec struct {
 	RegionCodes []string `json:"region_codes,omitempty"`
 	OfferCodes  []string `json:"offer_codes,omitempty"`
 
-	Concurrency int `json:"concurrency,omitempty"`
+	Concurrency int                `json:"concurrency,omitempty"`
```
candiduslynx, line 11: IMO this should be a separate struct in the plugin SDK called `SchedulerConfig` & we might need to have some setting like
```yaml
scheduler:
  concurrency: 1000
  strategy: round-robin
```
instead of putting the sched cfg to the spec top level

disq (approved):

plugins/source/awspricing/client/spec.go
erezrokah: This PR aims to first fix the bug. We can discuss changing the spec (which we'll need to do in a backwards compatible way or as a breaking change) in a separate PR (or even better, an issue). WDYT?

candiduslynx: 100% this should go into a separate discussion, maybe even a separate issue in the plugin-sdk

yevgenypats:
Looks good though I thought this is only in use by gcp hence it wasn't use by other plugins.

erezrokah:
> Looks good though I thought this is only in use by gcp hence it wasn't use by other plugins.

It was implemented for GCP in mind, but all plugins supported it

PR #12702, opened 2023-08-02
fix: CIS AWS Benchmarks section 2

aruneko:
#### Summary
I have fixed some detecting condition in CIS AWS Section 2.

- Section 2.1.2
    - in fact, current implementation always returns `fail`.
- Section 2.1.5
    - consider the statuses of `aws_s3_accounts`.
- Section 2.3.3
    - consider if a rds instance enabled `publicly_accessible` has internet connectivity or not.


jsonpr (approved):
Thanks for submitting these updates!

yevgenypats:
Im going to close this as we are not maintaining those anymore at the moment. and we moved to dbt as well as to premium policies - https://hub.cloudquery.io/addons/transformation to be able to maintain those in sustainable way. Thanks again for your contributions and old policies still available in the git history.

PR #12705, opened 2023-08-02
chore(ci): Add square CI

erezrokah:
#### Summary

Adds CI for the Square plugin


.github/workflows/source_square.yml
```diff
@@ -0,0 +1,54 @@
+name: Source Plugin Squre Workflow
```
hermanschaaf, line 1:
```suggestion
name: Source Plugin Square Workflow
```

PR #12707, opened 2023-08-02
chore(ci): Add docker release

erezrokah:
#### Summary

Based on https://github.com/cloudquery/python-plugins-release-playground.

I used the same `release_plugin.yml` workflow with an additional `prepare` step to detect the release type.

### DIFF
--- a/.github/workflows/release_plugin.yml
+++ b/.github/workflows/release_plugin.yml
@@ -7,15 +7,18 @@ on:
       - "!plugins-destination-sqlite-v*.*.*"
       - "!plugins-destination-snowflake-v*.*.*"
       - "!plugins-source-test-v*.*.*"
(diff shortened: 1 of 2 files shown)


.github/workflows/release_plugin.yml
```diff
@@ -7,15 +7,18 @@ on:
       - "!plugins-destination-sqlite-v*.*.*"
       - "!plugins-destination-snowflake-v*.*.*"
       - "!plugins-source-test-v*.*.*"
-env:
-  CGO_ENABLED: 0
 
 jobs:
-  release-plugin-binary:
```
erezrokah, line 14: Renamed to `release-plugin-go`

disq (approved):

PR #12732, opened 2023-08-02
feat: Typeform plugin

hermanschaaf:
This adds a Typeform plugin with two tables (`forms` and `form_responses`), serving as an example of a source plugin written in Python.


website/pages/docs/plugins/sources/typeform/overview.mdx
```diff
@@ -0,0 +1,48 @@
...
+# Typeform Source Plugin
+
+import { getLatestVersion } from "../../../../../utils/versions";
+import { Badge } from "../../../../../components/Badge";
+import Configuration from "./_configuration.mdx";
+import Authentication from "./_authentication.mdx";
+
+<Badge text={"Latest: " + getLatestVersion("source", `typeform`)}/>
+
+The CloudQuery Typeform plugin pulls data from [Typeform](https://www.typeform.com/) and loads it into any supported CloudQuery destination (e.g. PostgreSQL, BigQuery, Snowflake, and [more](/docs/plugins/destinations/overview)).
```
bbernays, line 16: Do the new python plugins work with all versions of the CLI (that support the latest protocol) or is there a minimum supported version?

hermanschaaf: There is a minimum supported CLI version of v3.12.0, because it requires the `docker` registry. I've added this to the text, thanks for pointing it out 👍

plugins/source/typeform/plugin/plugin.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def get_tables(self, options: plugin.TableOptions) -> List[plugin.Table]:
+        all_tables: List[plugin.Table] = [
+            tables.Forms(),
+        ]
+
+        # set parent table relationships
+        for table in all_tables:
+            for relation in table.relations:
+                relation.parent = table
```
disq, line 39: Was this needed or just 'nice to have'? I don't remember doing this in the Square plugin, which seemed to work (at least query) with child tables.

hermanschaaf: It's necessary for the docs to indicate the parent-child relationship on the child table

release-please-config.json
```diff
@@ -161,6 +161,17 @@
     "plugins/source/test": {
       "component": "plugins-source-test"
     },
+    "plugins/source/typeform": {
+      "component": "plugins-source-typeform",
+      "release-type": "simple",
+      "extra-files": [
+        "plugins/source/typeform/version.py"
```
erezrokah, line 168: I think this file needs to exist with a comment like in https://github.com/googleapis/python-aiplatform/blob/68e84e9bf97cabceb7f16f0e7c6d488c81852c24/google/cloud/aiplatform/gapic_version.py#L16

erezrokah (approved):
The CI stuff looks good. One comment

release-please-config.json
erezrokah: Yes it should

hermanschaaf:
Going to merge to check that the release-please config and other release steps work. We can iterate further in follow-up PRs, so feel free to still comment here if you notice something that should be fixed

PR #12736, opened 2023-08-02
feat: Adding support for OIDC Token in Azure plugin

joytandon17:
This PR is created to support OIDC Token in Azure plugin. This PR would enable Azure plugin to be used without AZURE_CLIENT_SECRET. This follows principle of Azure AD workload identity federation with AWS.


candiduslynx:
Hi @joytandon17!

Shouldn't this already work with the current code given that the default creds will attempt to use OIDC based on the env variables set?
See https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#-option-2-use-workload-identity

joytandon17:
> Hi @joytandon17!
> 
> Shouldn't this already work with the current code given that the default creds will attempt to use OIDC based on the env variables set? See https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#-option-2-use-workload-identity




> 

Hi @candiduslynx I think it only supports MSI(Managed Service Identity) Token from Azure itself, This PR would enable to use OIDC token which is not generated from Azure but for example AWS Cognito, Please see this document
https://blog.identitydigest.com/azuread-federate-aws/

hermanschaaf:
Hi @joytandon17 :wave Thanks for this PR! I think it makes sense from the perspective of supporting a new feature.

I'd like to request one or two changes if you don't mind:
- could you please remove the reliance on the `OIDC_TOKEN` environment variable and instead add an option to the spec (in `client/spec.go`) with the name `oidc_token`? The other environment variables like `AZURE_TENANT_ID` are special because they are recognized by the `az` CLI tool, but in general we prefer to add spec options rather than environment variables. This way users can use any environment variable name by using [environment variable substitution](https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution) in the config
- could you then also add some documentation for this new option to these two pages: https://github.com/cloudquery/cloudquery/blob/main/website/pages/docs/plugins/sources/azure/_configuration.mdx and https://github.com/cloudquery/cloudquery/blob/main/website/pages/docs/plugins/sources/azure/configuration.mdx. Just a short sentence or two would be fine. 

Let me know if that makes sense and if we can help in any way!

joytandon17:
@hermanschaaf can you please review this PR https://github.com/cloudquery/cloudquery/pull/12871? In this PR I pushed without rebasing(I though I had rebased but that was not the case)

joytandon17:
@hermanschaaf can you please review the PR? 
Also please ignore the tags added and removed by cqbot, I committed(by mistake) after merging main to my PR with some issues hence all the commits of main were showing here so I had to force push to previous commit to fix that.

plugins/source/azure/client/client.go
```diff
@@ -265,14 +266,31 @@ func New(ctx context.Context, logger zerolog.Logger, s *Spec) (schema.ClientMeta
...
 	if c.Options != nil {
 		credsOptions = &azidentity.DefaultAzureCredentialOptions{ClientOptions: c.Options.ClientOptions}
 	}
 	var err error
-	c.Creds, err = azidentity.NewDefaultAzureCredential(credsOptions)
-	if err != nil {
-		return nil, err
+	if oidcToken != "" {
+		// Accessing Azure using OIDC token
+		tenantID := os.Getenv("AZURE_TENANT_ID")
```
candiduslynx, line 277: `AZURE_TENANT_ID` & `AZURE_CLIENT_ID` shouldn't be looked up directly by plugin.
Instead, they should also be a part of the spec.

Additionally, the logic verifying that both of these params should be added.

joytandon17: @candiduslynx will add the required changes in a new PR.

joytandon17: @candiduslynx Thanks

PR #12737, opened 2023-08-02
feat: Replace DelegationSetId with full DelegationSet in aws_route53_hosted_zones

kfritchman46:
#### Summary

The default (not reusable) DelegationSet of AWS public HostedZones are only visible when describing the HostedZone. The DelegationSet API doesn't include them. Sync should include the DelegationSet as a json field of the `aws_route53_hosted_zones` table so changes in Nameservers can be detected.


erezrokah (changes requested):
Thanks for the PR @kfritchman46, can we still keep `DelegationSetId`? That would make this change non breaking for existing users

kfritchman46:
@erezrokah Absolutely, I wasn't sure how strict this project is on clutter vs reducing breaking changes. I'll update.

erezrokah (approved):

erezrokah:
> @erezrokah Absolutely, I wasn't sure how strict this project is on clutter vs reducing breaking changes. I'll update.

Thanks for following up, looks great 🚀 . We try to avoid breaking changes as it requires consumers of the data to drop existing tables in most cases

PR #12750, opened 2023-08-03
fix: Workaround 10k fetch incident limit in PagerDuty

dysosmus:
#### Summary

PagerDuty incidents listing API will fail when trying to retrieve more than 10 000 incidents.
```
HTTP response failed with status code 400, message: Arguments Caused Error (code: 2002): Offset+limit exceeds maximum allowed value of 10000. Please try to refine your search to reduce the returned set of records below this number, like adding a date range.
```
This PR update `fetchIncidents` to fetch incidents in increments of 90 days and workaround this limitation.

### DIFF
--- a/plugins/source/pagerduty/resources/services/incidents/incidents_fetch.go
+++ b/plugins/source/pagerduty/resources/services/incidents/incidents_fetch.go
@@ -2,6 +2,7 @@ package incidents
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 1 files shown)


hermanschaaf (approved):
LGTM, thanks for the PR @dysosmus ! Maybe in the future we can make the 90 days configurable, but this is fine for now.

Issue #12754, opened 2023-08-03
bug: postgres plugin - null handling for json


haris81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Team, Can you please review the handling of nulls in json for postgres plugin. Believe after migration to arrow, null handling logic is not being done anymore.. I am getting below error (which does not happen in sqlite plugin)

```
Starting sync for: azure (v9.1.0) -> [postgresql (v5.0.2)]

\ Syncing resources... (21/-, 19 resources/s) [1s] Error: failed to sync v3 source azure: rpc error: code = Internal desc = write failed: failed to execute batch with pgerror: severity: ERROR, code: 22P05, message: unsupported Unicode escape sequence, detail :\u0000 cannot be converted to text., hint: , position: 0, internal_position: 0, internal_query: , where: JSON data, line 1: ...playStatus":"Not Ready","level":"Info","message":...
unnamed portal parameter $6 = '...', schema_name: , table_name: , column_name: , data_type_name: , constraint_name: , file: jsonfuncs.c, line: 625, routine: json_ereport_error: ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)
\ Syncing resources... (21/-, 19 resources/s) [1s] %                                                                  
```

### Expected Behavior

Nulls should be handled for postgres destination plugin

### CloudQuery (redacted) config

N/A.. checked a few latest versions and all have errors with postgres destination

### CloudQuery (redacted) logs

```
2023-08-03T15:39:05Z INF Migrating table module=pg-dest table=azure_compute_virtual_machine_patch_assessments
2023-08-03T15:39:05Z INF Table exists, auto-migrating module=pg-dest table=azure_compute_virtual_machine_patch_assessments
2023-08-03T15:39:06Z INF End sync destinations=["postgresql (v5.0.2)"] module=cli source="azure (v9.1.0)" sync_time=2023-08-03T15:39:00Z
2023-08-03T15:39:06Z ERR BatchClose err="ERROR: unsupported Unicode escape sequence (SQLSTATE 22P05)" module=pgx pid=9214
[...]

PR #12755, opened 2023-08-03
fix: Strip NUL bytes in JSON

disq:
Fixes https://github.com/cloudquery/cloudquery/issues/12754

### DIFF
--- a/plugins/destination/postgresql/client/transformer.go
+++ b/plugins/destination/postgresql/client/transformer.go
@@ -146,7 +146,7 @@ func transformArr(arr arrow.Array) []any {
 			nested := array.NewSlice(a.ListValues(), start, end)
 			pgArr[i] = transformArr(nested)
 		case *types.JSONArray:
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

plugins/destination/postgresql/client/transformer.go
```diff
@@ -146,7 +146,7 @@ func transformArr(arr arrow.Array) []any {
 			nested := array.NewSlice(a.ListValues(), start, end)
 			pgArr[i] = transformArr(nested)
 		case *types.JSONArray:
-			pgArr[i] = a.Storage().(*array.Binary).Value(i)
+			pgArr[i] = stripNullsFromMarshalledJson(a.Storage().(*array.Binary).Value(i))
```
candiduslynx, line 149: IMO we could use the `GetOneForMarshal` method here

disq: You're right (same thing though)

PR #12763, opened 2023-08-04
chore(website): Doc config changes (destination plugins `F` through `K`)

candiduslynx:


plugins/destination/gremlin/client/spec.go
```diff
@@ -59,20 +59,18 @@ func (s *Spec) SetDefaults() {
 		s.AuthMode = authMode(strings.ToLower(string(s.AuthMode)))
 	}
 
-	if s.MaxRetries < 0 {
-		s.MaxRetries = 0
```
disq, line 63: Good catch. There's no handling code that allows for 0 maxretries (may it be infinite retries or whatever it may be...)

disq (approved):

PR #12765, opened 2023-08-04
feat: Make PostgreSQL destination use streaming batch writer (writes per table)

candiduslynx:
~Depends on https://github.com/cloudquery/plugin-sdk/pull/1131~


plugins/destination/postgresql/client/client.go
```diff
@@ -106,7 +106,7 @@ func (c *Client) Write(ctx context.Context, res <-chan message.WriteMessage) err
 	return c.writer.Write(ctx, res)
 }
 
-func (c *Client) Close(ctx context.Context) error {
+func (c *Client) Close(context.Context) error {
 	var err error
```
disq, line 110: should do `c.writer.Close(ctx)` here

plugins/destination/postgresql/client/insert.go
```diff
@@ -13,61 +13,36 @@ import (
...
+				sql = c.upsert(table)
+			} else {
+				sql = c.insert(table)
+			}
+		}
+
+		for _, row := range transformValues(msg.Record) {
+			batch.Queue(sql, row...)
 		}
 	}
```
disq, line 47: batch should be sent on channel close

candiduslynx: well, batch.Queue only adds the rows, the sending is done after iterating over mesages

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -53,9 +53,8 @@ ORDER BY
 	table_name ASC, ordinal_position ASC;
 `
 
-func (c *Client) listTables(ctx context.Context, include, exclude []string) (schema.Tables, error) {
-	var tables schema.Tables
-	whereClause := c.whereClause(include, exclude)
+func (c *Client) getDBTable(ctx context.Context, name string) (*schema.Table, error) {
+	whereClause := "AND pg_class.relname = '" + inClause(name) + "'"
```
disq:
```suggestion
	whereClause := "AND pg_class.relname IN " + inClause(name)
```
I don't think this complex code is necessary as the migrate table is invoked separately for each table

plugins/destination/postgresql/pgarrow/to_arrow.go
```diff
@@ -45,6 +45,10 @@ func Pg10ToArrow(t string) arrow.DataType {
 		return arrow.PrimitiveTypes.Int32
 	case "bigint", "int8":
 		return arrow.PrimitiveTypes.Int64
+	case "numeric(20,0)":
```
disq, line 48: Why not check for `strings.HasPrefix(t, "numeric(") && strings.HasSuffix(t, ",0)")`? (numerics with float components should still remain strings)

plugins/destination/postgresql/client/client.go
```diff
@@ -107,15 +107,14 @@ func (c *Client) Write(ctx context.Context, res <-chan message.WriteMessage) err
 }
 
 func (c *Client) Close(ctx context.Context) error {
-	var err error
 	if c.conn == nil {
 		return fmt.Errorf("client already closed or not initialized")
 	}
 	if c.conn != nil {
```
disq, line 113:
```suggestion
	err := c.writer.Close(ctx)
	if c.conn != nil {
```

disq (approved):

PR #12766, opened 2023-08-04
fix: Return a datadog client instance instead of nil

merlindorin:
#### Summary

The last versions of Datadog plugin do not work and generate the following message:

>  plugin not initialized. call Init() first

The `newClient` function return `nil,nil` if the configuration is good. It should return the client initialized.

I wrote also some extra tests to prevent regression.


disq:
Oops. Thanks for the fix!

disq (approved):

candiduslynx (approved):

Issue #12759, opened 2023-08-03
bug: `gcp_containeranalysis_occurrences` table is missing columns


AlecR:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The `gcp_containeranalysis_occurrences` table has the following columns:

```
_cq_id | uuid
-- | --
_cq_parent_id | uuid
project_id | utf8
name (PK) | utf8
resource | json
note_name | utf8
kind | utf8
remediation | utf8
create_time | timestamp[us, tz=UTC]
update_time | timestamp[us, tz=UTC]
```

### Expected Behavior

According to the docs, fields such as `vulnerability` and `discovered` should also be returned from the API for this table. We'd like to be syncing these columns when syncing this table.

### CloudQuery (redacted) config

N/A

### CloudQuery (redacted) logs

N/A

### CloudQuery version

N/A

### Pull request (optional)

- [ ] I can submit a pull request

bbernays:
The desired fields *should* be in the `details` field, but because the type is of type `interface` CQ does not properly grab the data

PR #12768, opened 2023-08-04
fix(resources): Handle previously missing interface fields

disq:
Closes https://github.com/cloudquery/cloudquery/issues/12759


candiduslynx:
I wonder if the same would apply well to AWS...

Issue #12769, opened 2023-08-04
bug: aws_organizations duplication


smokentar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`aws_organizations` -> an organization node is created for every account that belongs to the organization.
The loading cypher query uses `account_id` and `arn` as node properties. 
Example:
1 organization with 8 accounts results in 8 `aws_organizations` nodes (because `account_id` is different for each account)

Suggestion to avoid this duplication is using the `arn` field as a single node property. 
This way new organization nodes will only be added if the `arn` (a unique identifier for an organization) doesn't already exist.

Current load cypher query:
`TRC RUN "UNWIND $rows AS row MERGE (t:aws_organizations {account_id: row.account_id, arn: row.arn}) SET t = row"{...}`

### Expected Behavior

`aws_organizations` nodes are unique.
A CloudQuery sync will not create a new node for each account that is part of an organization. Instead it will only create a node if one doesn't already exist with the specified properties.

Desired load cypher query:
`TRC RUN "UNWIND $rows AS row MERGE (t:aws_organizations {arn: row.arn}) SET t = row"{...}`

### CloudQuery (redacted) config

```
kind: source
spec:
  name: "aws"
  registry: "github"
  path: "cloudquery/aws"
  version: "v22.1.0"
  tables: ["*"]
  skip_tables:
    - aws_ec2_vpc_endpoint_services
    - aws_cloudtrail_events
    - aws_docdb_cluster_parameter_groups
    - aws_docdb_engine_versions
    - aws_ec2_instance_types
    - aws_elasticache_engine_versions
    - aws_elasticache_parameter_groups
    - aws_elasticache_reserved_cache_nodes_offerings
    - aws_elasticache_service_updates
    - aws_elasticsearch_versions
    - aws_neptune_cluster_parameter_groups
    - aws_neptune_db_parameter_groups
    - aws_rds_cluster_parameters
    - aws_rds_cluster_parameter_groups
    - aws_rds_cluster_parameter_group_parameters
    - aws_rds_db_parameter_groups
    - aws_rds_engine_versions
    - aws_servicequotas_services
    - aws_servicequotas_quotas
    - aws_iam_role_last_accessed_details
    - aws_iam_user_last_accessed_details
    - aws_iam_policy_last_accessed_details
    - aws_directconnect_locations
    - aws_ram_resource_types
    - aws_lambda_runtimes
    - aws_docdb_event_categories
[...]

Issue #12648, opened 2023-07-31
AWS Plugin Quality of Life Improvements


bbernays:
These changes are breaking, so they should be bulked into a single major release.

### Tasks to be considered:
- [ ] Refactor SNS resolver to better follow the APIs... https://github.com/cloudquery/cloudquery/issues/12068#issuecomment-1628412100
- [x] Investigate switching from `GetAccountAuthorizationDetails` to `ListPolicies` for `aws_iam_policies` as `GetAuthorizationDetails` returns a lot of information that we do not need/use

### Tasks that are possibly out of scope:
- [ ] Refactor to better handle regions #10893 (we are not going to handle it in v23)
- [ ] https://github.com/cloudquery/cloudquery/issues/15391
- [ ] https://github.com/cloudquery/cloudquery/issues/12968

```[tasklist]
### Tasks
- [ ] https://github.com/cloudquery/cloudquery/pull/15377
- [ ] https://github.com/cloudquery/cloudquery/pull/15378
- [ ] https://github.com/cloudquery/cloudquery/pull/15380
- [ ] https://github.com/cloudquery/cloudquery/issues/15381
- [ ] https://github.com/cloudquery/cloudquery/issues/3466
- [ ] https://github.com/cloudquery/cloudquery/issues/15372
- [ ] https://github.com/cloudquery/cloudquery/issues/12974
- [ ] https://github.com/cloudquery/cloudquery/pull/15370
- [ ] https://github.com/cloudquery/cloudquery/pull/15398
- [ ] https://github.com/cloudquery/cloudquery/issues/15404
- [ ] https://github.com/cloudquery/cloudquery/issues/15411
- [ ] https://github.com/cloudquery/cloudquery/issues/15408
- [ ] https://github.com/cloudquery/cloudquery/pull/15423
- [ ] https://github.com/cloudquery/cloudquery/issues/12769
- [ ] https://github.com/cloudquery/cloudquery/issues/11029
- [ ] https://github.com/cloudquery/cloudquery/issues/15387
- [ ] https://github.com/cloudquery/cloudquery/issues/12646
```

jsonpr:
https://github.com/cloudquery/cloudquery/issues/12968 as well.

PR #12770, opened 2023-08-04
feat(aws): Define primary key for aws_organizations and remove accoun…

smokentar:
#### Summary

This PR is aimed to address #12769 by limiting the primary key on `aws_organizations` table to `arn`.

Additionally it removes `account_id` from the table as an organization can have many accounts.
Indicating the `account_id` from which the organization was discovered isn't necessary.

### DIFF
--- a/plugins/source/aws/resources/services/organizations/organizations.go
+++ b/plugins/source/aws/resources/services/organizations/organizations.go
@@ -20,11 +20,12 @@ func Organizations() *schema.Table {
 			&types.Organization{},
 			transformers.WithSkipFields(
 				"AvailablePolicyTypes", // deprecated and misleading field according to docs
+				"AccountId" // an organization doesn't belong to an account, other way around
 			),
 			transformers.WithPrimaryKeys("Arn"),
 		),
 		Multiplex: client.ServiceAccountRegionMultiplexer(tableName, "organizations"),
-		Columns:   []schema.Column{client.DefaultAccountIDColumn(true)},
+		Columns:   []schema.Column{client.DefaultAccountIDColumn(false)},
 	}
 }
 
--- a/website/tables/aws/aws_organizations.md
+++ b/website/tables/aws/aws_organizations.md
@@ -4,7 +4,7 @@ This table shows data for Organizations.
 
 https://docs.aws.amazon.com/organizations/latest/APIReference/API_Organization.html
 
-The composite primary key for this table is (**account_id**, **arn**).
+The primary key for this table is **arn**.
 
 ## Columns
 
@@ -12,7 +12,6 @@ The composite primary key for this table is (**account_id**, **arn**).
 | ------------- | ------------- |
 |_cq_id|`uuid`|
 |_cq_parent_id|`uuid`|
-|account_id (PK)|`utf8`|
 |arn (PK)|`utf8`|
 |feature_set|`utf8`|
 |id|`utf8`|


erezrokah (commented):
Thanks for the PR @smokentar, I'm not sure I understand how this fixes duplication as before the PK was a composite key of **both** `account_id` **and** `arn`, and with this PR the PK is **only** the `arn`.

smokentar:
Hey @erezrokah thanks for the review!

The situation is that during different syncs of different accounts, CQ generates multiple `aws_organizations` nodes for each account:

Example:
When 8 accounts from the same organization are synced separately, this results in 8 `aws_organizations` (because `account_id` is different for each account sync).
As @bbernays mentioned in https://github.com/cloudquery/cloudquery/issues/12769, account_id is there to identify what is visible from each account during a sync.

The duplication of those nodes happens because of the composite PK which is a combination of the unique `account_id` + the same `arn` (ARN for the Org is always unique as it includes the `master_account_id` in it instead of the `account_id` from which the Org has been discovered)

By removing `account_id` from the composite PK we ensure only one organization will be created even during multiple syncs across different accounts (that are part of the same organization).

The reason why the `account_id` is removed is because if we leave only the `arn` as PK then we encounter a situation where the single `aws_organizations` discovered will exist with a specific `account_id` - most likely the first account that was ever sync'ed from this organization.

Maybe that's not really a problem, looking for your opinion @erezrokah @bbernays

erezrokah:
Thanks for explaining @smokentar. Won't this be a problem for each resource that has `account_id` as a primary key?

smokentar:
Good question, not for every resource - the problem would only exist with **centralised** resources that are visible from **multiple** accounts.
In such scenarios PK should only be the `arn` of the resource.

Example: 
`aws_ec2_instances` can only be visible from the account that it is deployed to (eg. `1234567890`) and will have that as `account_id` value.

`aws_organizations` can be visible from all accounts that are part of it (eg. `1234567890`, `0987654321`). 
Since it's visible from all these accounts, it gets re-created multiple times with a different `account_id`.
However, only one organization exists in the master account.

That's why we'd want to use only `arn` as the PK - it's always the same for an organization, regardless from which account it's discovered.

Example values for duplicate `aws_organizations`:
```
account_id: 1111111111
master_account_id: 0987654321
arn: arn:aws:organizations::0987654321:organization/o-uuuuuuu

account_id: 222222222
master_account_id: 0987654321
arn: arn:aws:organizations::0987654321:organization/o-uuuuuuu
```

And for the `account_id` field:
- One option for such resources is to derive `account_id` from the `arn` of the resource, however that would break the logic of `account_id` displaying which account a resource is visible from.
- A second option is to remove `account_id` from such resources, but then it's challenging to figure out which all of these could be.
- A third option is to keep `account_id` with the value of the first account the centralised resource was discovered from

bbernays:
@smokentar - Please correct me if I am wrong/mistaken/missing something, but the removing the `aws_account` from the PK is designed to enable simpler querying of the data because one doesn't need a `group by` or `distinct` clause (depending on the database being used). Are there any other use cases that this change unblocks?

smokentar:
Hey @bbernays thanks for getting involved.
My goal with removing `account_id` from the PK was not to enable simpler querying but to prevent the duplication of `aws_organizations` in the destination.
From a data lake perspective, if you have 100 accounts in an organization and you sync them, you will get 100 duplicate `aws_organizations`.

If the zen of CQ sync is "_load everything that is visible from the account regardless of existence in the backend_" then I'm happy to roll with that and leave this unchanged.
I'll follow your guidance.

bbernays:
Thanks for that clarification! Yeah in general CloudQuery tries to do as few transformations of the data as possible and also tries persist as much of the data that we get from the APIs as possible in order to unlock the most number of use cases. 

That being said, I do think that the `account_id` field should be renamed to `request_account_id` to better indicate that the `organization` is not owned by that account, but that the request was made against a specific account. I have added this change as a task in a [general cleanup/improvement ticket](https://github.com/cloudquery/cloudquery/issues/12648) (that will require breaking changes)

bbernays:
Thank you for this, it looks good! A few notes:
1. Can you update the title and description to reflect the changes
2. Can you add a description to the table about that new field see [example here](https://github.com/cloudquery/cloudquery/blob/ea292948c822ef6bb15eabc5a0f3474bd5e4bcb4/plugins/source/aws/resources/services/organizations/account_parents.go#L18)

Just a heads up, we probably won't merge this PR for a few weeks because we are trying to batch our breaking changes to reduce migrations for users and currently we are focused on adding support for plugins in other languages (JavaScript and Java) by creating new sdks like we have in Go and Python.

bbernays:
@smokentar - If you could resolve the conflicts on this PR we can merge it.

PR #12789, opened 2023-08-07
chore(website): Doc config changes (destination plugins `M` through `S`)

candiduslynx:


plugins/destination/postgresql/client/spec.go
```diff
@@ -16,16 +15,12 @@ const (
...
 	ConnectionString string              `json:"connection_string,omitempty"`
 	PgxLogLevel      LogLevel            `json:"pgx_log_level,omitempty"`
-	MigrateMode      string              `json:"migrate_mode,omitempty"`
 	BatchSize        int                 `json:"batch_size,omitempty"`
 	BatchSizeBytes   int                 `json:"batch_size_bytes,omitempty"`
 	BatchTimeout     configtype.Duration `json:"batch_timeout,omitempty"`
 }
 
 func (s *Spec) SetDefaults() {
-	if s.MigrateMode == "" {
```
erezrokah, line 26: This is technically a breaking change, so we could log a warning instead.
On the other hand this is probably fine as if someone is using this for `forced` migrate mode, and it will stop working they'll get an error message that tells them to use a forced migration when needed

candiduslynx: IMO this will just error out, as the option wasn't being used anymore (so I'd say not a breaking change).

erezrokah (approved):

PR #12844, opened 2023-08-08
chore: Exclude external links in broken links checker

hermanschaaf:
Opening this one to see what others think. I like the broken links checker, but external links seem to just lead to a never-ending stream of toil to fine-tune what we exclude, and is prone to failure when external websites are having downtime. I'd rather test what we can control, and ignore the rest.


bbernays:
What if we just did a single daily/weekly check for external broken links? That way we would probably be able to fix issues before users ran into issues, but it wouldn't block normal work?

hermanschaaf:
> What if we just did a single daily/weekly check for external broken links? That way we would probably be able to fix issues before users ran into issues, but it wouldn't block normal work?

I like that idea; it will just require some additional work to notify us of failures, since it won't block any PRs. I can open an issue to tackle that as a follow-up if we want to go that route

erezrokah:
The broken links check doesn't block PRs as it is, and we do run it on a schedule already https://github.com/cloudquery/cloudquery/blob/333745b9085c67ace7274f35213b9ad04ba18c21/.github/workflows/broken_links.yml#L6

It's not a required check so PRs can get merged without it passing

hermanschaaf:
@erezrokah It's not required but I _think_ it blocks the automerge functionality, plus we also can't just leave it failing; someone has to go look into it every time.

erezrokah:
> but external links seem to just lead to a never-ending stream of toil to fine-tune what we exclude, and is prone to failure when external websites are having downtime

It does catch broken links in our docs, usually when we link to external doc sites and the content on those site changes.

erezrokah:
> It's not required but I _think_ it blocks the automerge functionality, plus we also can't just leave it failing; someone has to go look into it every time.

It doesn't block auto merge, as it only waits for required checks.

Maybe we can force the status check to pass via https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error but send a slack message instead?

erezrokah:
Please note that using Slack notification will be tricky to do in forked PRs as it requires authentication to Slack, GitHub secrets, etc.

.github/workflows/broken_links_scheduled.yml
```diff
@@ -0,0 +1,102 @@
...
+              echo "check=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Setup node
+        if: steps.broken-links.outputs.check != 'false'
+        uses: actions/setup-node@v3
+        with:
+          node-version: "lts/*"
+
+      - name: Waiting for Vercel Preview
```
hermanschaaf, line 38: Can probably remove all this from this file

.github/workflows/broken_links_scheduled.yml
```diff
@@ -0,0 +1,68 @@
+name: Check for broken links
```
erezrokah, line 1:
```suggestion
name: Check for internal and external broken links
```

And maybe rename the other one too

erezrokah (approved):

PR #12975, opened 2023-08-10
fix: Normalization logic for new & existing tables differences

candiduslynx:
Based on the findings from #12765 & #12870:
1. Handle `numeric(20,0)` as `uint64` from the database schema
2. Fix `numeric(20,0)[]` reported as `numeric[]` from CockroachDB
3. Always perform normalization even for the new tables

### DIFF
--- a/plugins/destination/postgresql/client/list_tables.go
+++ b/plugins/destination/postgresql/client/list_tables.go
@@ -23,6 +23,9 @@ SELECT
 	pg_class.relname AS table_name,
 	pg_attribute.attname AS column_name,
 	CASE
+	    -- This is required per the differences in pg_catalog.format_type implementations
+	    -- between PostgreSQL & CockroachDB.
+	    -- namely, numeric(20,0)[] is returned as numeric[] unless we use the typelem format + []
 	    WHEN pg_type.typcategory = 'A' AND pg_type.typelem != 0
 		THEN pg_catalog.format_type(pg_type.typelem, pg_attribute.atttypmod) || '[]'
 		ELSE pg_catalog.format_type(pg_attribute.atttypid, pg_attribute.atttypmod)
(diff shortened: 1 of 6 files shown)


plugins/destination/postgresql/client/list_tables.go
```diff
@@ -22,7 +22,11 @@ SELECT
 	columns.ordinal_position AS ordinal_position,
 	pg_class.relname AS table_name,
 	pg_attribute.attname AS column_name,
-	pg_catalog.format_type(pg_attribute.atttypid, pg_attribute.atttypmod) AS data_type,
+	CASE
+	    WHEN pg_type.typcategory = 'A' AND pg_type.typelem != 0
+		THEN pg_catalog.format_type(pg_type.typelem, pg_attribute.atttypmod) || '[]'
+		ELSE pg_catalog.format_type(pg_attribute.atttypid, pg_attribute.atttypmod)
+	END AS data_type,
```
hermanschaaf, line 29: What's the reasoning behind the change here?

candiduslynx: It's the point 2 in the description.

There's a difference in `pg_catalog.format_type` implementations between CocroachDB & PostgreSQL.

The `list<uint64>` Apache Arrow type is stored as `numeric(20,0)[]` in both databases, but then `pg_catalog.format_type(pg_attribute.atttypid, pg_attribute.atttypmod)` would return:
* `numeric(20,0)[]` in PostgreSQL
* `numeric[]` in CockroachDB

This query change allows us to have `numeric(20,0)[]` returned from both databases

hermanschaaf: I see, thanks for the explanation! Could we maybe add a comment to the query here to explain that this is for CockroachDB compatibility? I'm thinking that we may want to split CockroachDB into its own plugin down the line because of all the subtle differences, so knowing what is for Cockroach will be useful then

hermanschaaf (approved):

plugins/destination/postgresql/client/list_tables.go
candiduslynx: Added a comment in 13cd014

PR #13005, opened 2023-08-11
feat: Add ability to generate arbitrary number of rows, add sub table

disq:
Uses untagged SDK due to the timestamp issue (as `test_testdata_table` uses `NewTestDataGenerator` to generate all of our testdata as a table)


erezrokah (approved):

plugins/source/test/client/spec.go
```diff
@@ -1,13 +1,23 @@
 package client
 
 type Spec struct {
-	NumClients int `json:"num_clients"`
+	NumClients int  `json:"num_clients"`  // Number of clients to create. Default: 1
+	NumRows    *int `json:"num_rows"`     // Number of rows to generate in test_some_table. Defaul: 1
```
candiduslynx, line 5: make this simple `int` values & update `<=0` to `1` in `SetDefaults`

candiduslynx (changes requested):

plugins/source/test/resources/services/test_some_table.go
```diff
@@ -36,13 +37,19 @@ func TestSomeTable() *schema.Table {
...
 
 func fetchSomeTableData(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
-	res <- map[string]any{
-		"column1": "test_project_id",
-		"column2": 123,
+	cl := meta.(*client.Client)
+	for i := 0; i < *cl.Spec.NumRows; i++ {
+		res <- map[string]any{
+			"resource_id": i + 1,
+			"column2":     fmt.Sprintf("test_column_data_%d", i%3),
```
candiduslynx, line 51: why `%3`?

plugins/source/test/resources/services/test_sub_table.go
```diff
@@ -0,0 +1,51 @@
...
+		},
+	}
+}
+
+func fetchSubTableData(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	for i := 0; i < *cl.Spec.NumSubRows; i++ {
+		res <- map[string]any{
+			"sub_resource_id": i,
+			"data_column":     fmt.Sprintf("sub_data_%d", i%3),
```
candiduslynx, line 47: why `%3`?

candiduslynx (changes requested):

plugins/source/test/client/spec.go
disq: No we want 0 to be available for no rows.

plugins/source/test/resources/services/test_some_table.go
disq: Simple counter, but also duplicates data to some extent.

candiduslynx (approved):

PR #13007, opened 2023-08-11
feat: "go" registry support in CLI

icio:
Follows https://github.com/cloudquery/plugin-pb-go/pull/85


yevgenypats (changes requested):
Commented on the slack thread. Let's avoid introducing another specific registry and use either `local` or `grpc`

PR #13059, opened 2023-08-14
feat: Instantiate services at sync time rather than during init phase of sync

bbernays:
#### Summary

This will reduce memory usage in most cases, unless a user is running `tables:["*"]`. This will instantiate the service client the first time a client needs it rather than instantiating every single supported client during initialization phase of the sync.

If a user doesn't use a service, it will never be instantiated


plugins/source/aws/client/services.go
```diff
@@ -375,3 +256,249 @@ type Services struct {
...
+		s.Wafv2 = wafv2.NewFromConfig(c)
+	case "wafregional":
+		s.Wafregional = wafregional.NewFromConfig(c)
+	case "wellarchitected":
+		s.Wellarchitected = wellarchitected.NewFromConfig(c)
+	case "workspaces":
+		s.Workspaces = workspaces.NewFromConfig(c)
+	case "xray":
+		s.Xray = xray.NewFromConfig(c)
+	}
```
hermanschaaf, line 503: Let's add a default case here that panics if the function is called with an unrecognized service

plugins/source/aws/client/client.go
```diff
@@ -33,8 +33,8 @@ type Client struct {
 	Backend              state.Client
 	specificRegions      bool
 	Spec                 *Spec
-	// Do not rely on this field, it will be removed once https://github.com/aws/aws-sdk-go-v2/issues/2163 is resolved
```
hermanschaaf, line 36: Is this commend not longer true? Will we keep AWSConfig?

plugins/source/aws/client/services.go
```diff
@@ -375,3 +256,249 @@ type Services struct {
 	Workspaces                   services.WorkspacesClient
 	Xray                         services.XrayClient
 }
+
+func (s *Services) InitService(awsConfig *aws.Config, service string) {
+	c := awsConfig.Copy()
+	switch strings.ToLower(service) {
+	case "accessanalyzer":
```
hermanschaaf, line 263: Not necessarily blocking, but I think it would be great if we could expose constants for all the supported services, which could then be used by the resolvers rather than hardcoded strings. This would also make discovery of where the names come from easier for developers using IDEs

hermanschaaf (approved):
Couple of suggestions & questions but overall LGTM and I like the idea!

plugins/source/aws/client/client.go
bbernays: We will need to because the `initService` function uses it to initialize the service:
```
func (s *Services) InitService(awsConfig *aws.Config, service string) {
	c := awsConfig.Copy()
```

plugins/source/aws/client/client.go
```diff
@@ -115,7 +116,15 @@ func (c *Client) ID() string {
...
 }
 
-func (c *Client) Services() *Services {
+func (c *Client) Services(service_names ...string) *Services {
+	for _, service := range service_names {
+		svcConfig := funk.Get(c.ServicesManager.ServicesByPartitionAccount(c.Partition, c.AccountID), strings.ToUpper(service[:1])+service[1:])
+		if svcConfig == nil {
+			c.accountMutex[c.AccountID].Lock()
+			c.ServicesManager.ServicesByPartitionAccount(c.Partition, c.AccountID).InitService(c.AWSConfig, service)
+			c.accountMutex[c.AccountID].Unlock()
```
bbernays, line 125: @hermanschaaf - The linter was telling me not to have a `defer c.accountMutex[c.AccountID].Unlock()` in a for loop. Now that `InitService` can panic we need a way to be able to unlock the mutex in that case. Any thoughts on how to structure the code other than ignoring the linting rule?

plugins/source/aws/client/services.go
bbernays: Done in https://github.com/cloudquery/cloudquery/pull/13059/commits/a08568ce98f8c0f120c6fd44ed7d0ebdae6fc8aa

plugins/source/aws/client/client.go
hermanschaaf: Do we really need to unlock the mutex if the code panics? I would have thought all the goroutines get stopped anyway

hermanschaaf: But alternatively you can move just this bit of code:

```
			c.accountMutex[c.AccountID].Lock()
			c.ServicesManager.ServicesByPartitionAccount(c.Partition, c.AccountID).InitService(c.AWSConfig, service)
			c.accountMutex[c.AccountID].Unlock()
```

into its own function like:

```
func (c *Client) update(...) {
			c.accountMutex[c.AccountID].Lock()
			defer c.accountMutex[c.AccountID].Unlock()
			c.ServicesManager.ServicesByPartitionAccount(c.Partition, c.AccountID).InitService(c.AWSConfig, service)
}
```

bbernays: This gets executed as part of a table resolver, so the go routine will get caught and I imagine the mutex would never be released

bbernays: Maybe I am wrong though?

hermanschaaf: Oh I see yeah, we catch the panics from table resolvers. Ok well my second suggestion should work in that case

hermanschaaf:
Because this PR touches more than 500 files we will have to manually merge it

PR #13063, opened 2023-08-15
fix: CIS AWS v1.5.0 Sec5

aruneko:
#### Summary
I've fixed some queries of CIS AWS v1.5.0 Section 5.

- view `view_aws_nacl_ingress_rules`
  - store deny rules to detect `pass` resources in section 5.2
- 5.2
  - make results grouped by security group id
- 5.3
  - fix type in the detecting condition
  - make results grouped by security group id


yevgenypats:
Im going to close this as we are not maintaining those anymore at the moment. and we moved to dbt as well as to premium policies - https://hub.cloudquery.io/addons/transformation to be able to maintain those in sustainable way. Thanks again for your contributions and old policies still available in the git history.

PR #13065, opened 2023-08-15
fix: AWS foundational security controls for S3.8 query reference

### LINKED ISSUE #13064: bug: duplication in AWS foundational security controls for S3

### Current Behavior

Currently in `cloudquery/plugins/source/aws/policies/foundational_security/s3.sql` checks S3.1 and S3.8 refer to the same include file:
```
\set check_id 'S3.1'
\echo "Executing check S3.1"
\ir ../queries/s3/account_level_public_access_blocks.sql       <-- correct

\set check_id 'S3.2'
\echo "Executing check S3.2"
\ir ../queries/s3/publicly_readable_buckets.sql

[..skip..]

\set check_id 'S3.8'
\echo "Executing check S3.8"
\ir ../queries/s3/account_level_public_access_blocks.sql     <-- duplicate
```

### Expected Behavior

According to https://docs.aws.amazon.com/securityhub/latest/userguide/s3-controls.html :
```
[S3.8] S3 Block Public Access setting should be enabled at the bucket-level
```
..so I believe the correct file to include should be `s3/bucket_level_public_access_blocks.sql`
ie. bucket instead of the (duplicated) account 

N/A

### Steps To Reproduce

_No response_

N/A

### CloudQuery version

(I actually don't know)

### Additional Context

_No response_

### Pull request (optional)

candiduslynx:
Closes #13064


jsonpr (changes requested):
Good update.  Let's keep the control consistent with Foundational Security Best Practices Naming.  In this case, `S3 Block Public Access setting should be enabled at the bucket-level`

That will require a change to https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/policies/queries/s3/bucket_level_public_access_blocks.sql

jsonpr (approved):

PR #13112, opened 2023-08-15
fix: Handle differences in `format_type` implementation between PostgreSQL & CockroachDB

candiduslynx:
Copied the logic from #12975 for data types


plugins/source/postgresql/client/list_tables.go
```diff
@@ -31,13 +38,15 @@ SELECT
 		ELSE false
 	END AS not_null,
     CASE
-		WHEN conkey IS NOT NULL AND (contype = 'p' OR contype = 'u') AND array_position(conkey, pg_attribute.attnum) > 0 THEN true
+		WHEN conkey IS NOT NULL AND contype = 'u' AND array_position(conkey, pg_attribute.attnum) > 0 THEN true
```
erezrokah, line 41: I think we want to return `is_unique` when the constraint a primary key, as that implicitly says the column is unique

hermanschaaf (approved):

hermanschaaf (changes requested):
Agree this is a real bug, but decided we'll discuss this change internally a bit more before deciding on the fix

plugins/source/postgresql/client/list_tables.go
erezrokah: Maybe the fix should be to update the query to mark the column as unique only if there's a single primary key

hermanschaaf (approved):

erezrokah (approved):

PR #13129, opened 2023-08-15
feat: Simplify value transformation

candiduslynx:
1. Instead of performing type switch for every row for every column, just remember it once for the value transformation
2. reuse the record builder (per `NewRecord` resetting the record builder for reuse)

### DIFF
--- a/plugins/source/postgresql/client/sync.go
+++ b/plugins/source/postgresql/client/sync.go
@@ -108,6 +108,10 @@ func (c *Client) syncTables(ctx context.Context, snapshotName string, filteredTa
 }
 
 func syncTable(ctx context.Context, tx pgx.Tx, table *schema.Table, res chan<- message.SyncMessage) error {
+	arrowSchema := table.ToArrowSchema()
+	builder := array.NewRecordBuilder(memory.DefaultAllocator, arrowSchema)
+	transformers := transformersForSchema(arrowSchema)
+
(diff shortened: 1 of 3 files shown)


hermanschaaf:
I think what's mainly happening here is the logic is getting pulled out of the loop so that it's more efficient; is this correct? Could you add a PR description to explain the change? 🙇

plugins/source/postgresql/client/sync.go
```diff
@@ -118,27 +122,29 @@ func syncTable(ctx context.Context, tx pgx.Tx, table *schema.Table, res chan<- m
...
 	for rows.Next() {
 		values, err := rows.Values()
 		if err != nil {
 			return err
 		}
 
-		arrowSchema := table.ToArrowSchema()
-		rb := array.NewRecordBuilder(memory.DefaultAllocator, arrowSchema)
-		for i := range values {
-			val, err := prepareValueForResourceSet(arrowSchema.Field(i).Type, values[i])
```
hermanschaaf, line 130: Should we do the same in `cdc.go` and then remove `prepareValueForResourceSet`, rather than having similar logic in two places?

candiduslynx: 9c693b3086fde8d5787fb03f8ad6e480ca06cc65

hermanschaaf (approved):

PR #13133, opened 2023-08-15
feat: Improve error message on table with PKs

erezrokah:
#### Summary

Related to https://github.com/justmiles/cq-source-crowdstrike/issues/5.
A better way would be for the CLI to get what modes are supported by the destination.
I removed the backend error message as I found it confusing and it seems the `write_mode` error is more common (and it is always correct to set it).


erezrokah:
Will open not from a fork

PR #13134, opened 2023-08-15
fix!: Don't mark columns unique if they're part of the PK or composite unique constraint

### LINKED ISSUE #13131: bug: (Postgres Source) Composite unique columns not translated to destination correctly
When there are composite PKs or unique columns, the Postgres Source plugin incorrectly marks the individual columns as unique. 

Since we don't currently support composite unique columns at the SDK level, it seems like we shouldn't enforce any uniqueness constraints in the case where there are composite PKs or unique columns at this point.

candiduslynx:
Closes #13131

### DIFF
--- a/plugins/source/postgresql/client/list_tables.go
+++ b/plugins/source/postgresql/client/list_tables.go
@@ -38,7 +38,12 @@ SELECT
 		ELSE false
 	END AS not_null,
     CASE
-		WHEN conkey IS NOT NULL AND (contype = 'p' OR contype = 'u') AND array_position(conkey, pg_attribute.attnum) > 0 THEN true
+		WHEN
+		    conkey IS NOT NULL
+		    AND contype = 'u'
+		    AND array_position(conkey, pg_attribute.attnum) > 0
+		    AND array_length(conkey, 1) = 1 -- we don't handle composite unique keys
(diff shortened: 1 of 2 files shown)


plugins/source/postgresql/client/list_tables.go
```diff
@@ -38,7 +38,12 @@ SELECT
 		ELSE false
 	END AS not_null,
     CASE
-		WHEN conkey IS NOT NULL AND (contype = 'p' OR contype = 'u') AND array_position(conkey, pg_attribute.attnum) > 0 THEN true
+		WHEN
+		    conkey IS NOT NULL
+		    AND contype = 'u'
```
erezrokah, line 43: Don't we want to set it as true if it's a single PK?

candiduslynx: do we?

erezrokah: Yes, since that implies the column is unique in Postgres

candiduslynx: 9e4063ef7663bf23c021d8d67a5f3952e61106e2

erezrokah (approved):

PR #13137, opened 2023-08-15
feat!: Allow multiple rows in sync tables per sent record

### LINKED ISSUE #13110: bug: CQ gets super slow that it reaches 2 resources/s sometimes

### Current Behavior

I'm trying to extract data from 22 tables with 1gb data and 6 million records in a Postgres server to another Postgres server. Both instances are on the same machine (32 CPU and 270GB RAM) where CQ is installed by Docker. The rate I get is somewhere between 2-5000 resources/s and most of the time 500s. 
I then tested the same config on my local machine with another 2 Postgres databases installed locally and got almost the same result. 

### Expected Behavior

I was expecting getting at least 10K resources/s where the performance of CQ would make it possible to transfer all the data in a time shorter than Airbyte (in Airbyte it took 15 min to extract and load this dataset). However, CQ took 3.5 hours to finish. It looks like there is a bug or something wrong in my config. 

kind: source
spec:
  name: "pg-source"
  registry: "github"
  path: "cloudquery/postgresql"
  version: "v2.0.6"
  tables: ["*"]
  destinations: ["pg-destination"]
  spec:
    connection_string: "postgresql://***:***@localhost:5445/postgres?sslmode=disable"
    pgx_log_level: debug
    
---
kind: destination
spec:
  name: "pg-destination"
  registry: "github"
  path: "cloudquery/postgresql"
  version: "v5.0.4"
  spec:
    connection_string: "postgresql://***:***@localhost:5442/postgres?sslmode=disable"
    pgx_log_level: debug

### Steps To Reproduce

1. On docker container or locally, install CQ
2. Install 2 Postgres instances and fill it with random tables
3. Use the attached config.yml
4. Run the sync
5. Check the speed on the logs
[config.txt](https://github.com/cloudquery/cloudquery/files/12343724/config.txt)

[cloudquery_logs.txt](https://github.com/cloudquery/cloudquery/files/12343716/cloudquery_logs.txt)

### CloudQuery version

3.13.1

### Additional Context

_No response_

### Pull request (optional)

candiduslynx:
Closes #13110


plugins/source/postgresql/client/sync.go
```diff
@@ -141,7 +142,17 @@ func syncTable(ctx context.Context, tx pgx.Tx, table *schema.Table, res chan<- m
 
 			scalar.AppendToBuilder(builder.Field(i), s)
 		}
-		res <- &message.SyncInsert{Record: builder.NewRecord()} // NewRecord resets the builder for reuse
+
+		processed++
+		if processed%c.pluginSpec.RowsPerRecord == 0 {
```
hermanschaaf, line 147: nit: I think we should either not reset `processed` to 0 or use `processed >= c.pluginSpec.RowsPerRecord` here, but not both, it sends mixed messages of what should be happening

plugins/source/postgresql/client/sync.go
```diff
@@ -141,7 +142,17 @@ func syncTable(ctx context.Context, tx pgx.Tx, table *schema.Table, res chan<- m
...
+			res <- &message.SyncInsert{Record: builder.NewRecord()} // NewRecord resets the builder for reuse
+			processed = 0
+		}
+	}
+
+	record := builder.NewRecord()
+	if record.NumRows() > 0 { // only send if there are some unsent rows
```
hermanschaaf, line 154: I think it's safe, but it's been a while since I've worked with this so just checking: is it safe to call `builder.NewRecord` on a builder with 0 rows?

hermanschaaf: Yeah no problem there

website/pages/docs/plugins/sources/postgresql/overview.mdx
```diff
@@ -55,16 +55,20 @@ This is the (nested) spec used by the PostgreSQL source Plugin.
...
+- `cdc_id` (`string`) (optional)
 
-  If set to a non empty string the source plugin will start syncing CDC via PostgreSQL logical replication in real-time.
+  If set to a non-empty string the source plugin will start syncing CDC via PostgreSQL logical replication in real-time.
   The value should be unique across all sources.
 
+- `rows_per_record` (`integer`) (optional) (default: `1`)
+
+  Amount of rows to be packed into a single Apache Arrow record to be sent over the wire during sync (or initial sync in the CDC mode).
+  We suggest using amounts (e.g., `5000`) to sync from large databases/tables.
```
hermanschaaf, line 71:
```suggestion
  We suggest using significantly more than the default (e.g.`5000`) to sync from large databases/tables.
```

hermanschaaf:
Is this still meant to be marked as a breaking change?

erezrokah (approved):

PR #13160, opened 2023-08-16
fix(resources): Link to Go docs in `gcp_storage_buckets`

erezrokah:
#### Summary

Context in https://discord.com/channels/872925471417962546/1141351009545879612/1141361426284425246.
The GO SDK for buckets doesn't mimic the JSON API, they have a different object struct for some reason, see https://github.com/googleapis/google-cloud-go/blob/fbe78a28e01043b059e86c6e960d092b299ac9a4/storage/bucket.go#L753

Not sure why they wrap is and normalize some properties instead of retuning the raw bucket (which they have).
This PR links to the Go docs instead of the JSON API docs so it's easier to understand the data CloudQuery sync

### DIFF
--- a/plugins/source/gcp/resources/services/storage/buckets.go
+++ b/plugins/source/gcp/resources/services/storage/buckets.go
@@ -11,7 +11,7 @@ import (
 func Buckets() *schema.Table {
 	return &schema.Table{
 		Name:        "gcp_storage_buckets",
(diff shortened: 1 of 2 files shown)


candiduslynx (approved):

candiduslynx:
Maybe it should be `chore(docs)`?

erezrokah:
> Maybe it should be `chore(docs)`?

This is a user facing change since if someone tries to generate the tables for the plugin they see a different description, hence the `fix`

PR #13174, opened 2023-08-17
fix: Airtable `dateTime` type handling

erezrokah:
#### Summary

Fixes detecting and parsing of Airtable `dateTime` type. Also adds a log message when getting the tables

### DIFF
--- a/plugins/source/airtable/src/tables.ts
+++ b/plugins/source/airtable/src/tables.ts
@@ -1,15 +1,5 @@
 import type { DataType } from '@cloudquery/plugin-sdk-javascript/arrow';
-import {
-  Utf8,
-  Timestamp,
-  Date_,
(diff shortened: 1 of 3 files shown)


plugins/source/airtable/src/tables.ts
```diff
@@ -78,17 +68,7 @@ const airtableFieldToArrowField = (field: APIField): DataType => {
     }
     case APIFieldType.createdTime:
     case APIFieldType.lastModifiedTime: {
-      switch (field.options.result.type) {
-        case 'date': {
-          return new Date_(DateUnit.DAY);
```
erezrokah, line 83: We don't want to use Arrow Date as most (all?) of our destinations map it to string

candiduslynx (approved):

PR #13179, opened 2023-08-18
feat: Improve logging of discovery phase

erezrokah:
#### Summary

Better logging for the discovery phase so it doesn't seem the CLI hangs during that phase.

### DIFF
--- a/plugins/source/airtable/package.json
+++ b/plugins/source/airtable/package.json
@@ -66,7 +66,8 @@
     "eslint-plugin-you-dont-need-lodash-underscore": "^6.12.0",
     "prettier": "^3.0.1",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.5",
+    "winston": "^3.10.0"
   },
   "engines": {
     "node": ">=16.17.0"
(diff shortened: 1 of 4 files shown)


plugins/source/airtable/src/tables.ts
```diff
@@ -37,8 +38,9 @@ const options = (apiKey: string) => ({
   timeout,
 });
 
-const getBases = (apiKey: string, endpointURL: string) => {
-  return got(`${endpointURL}/v0/meta/bases`, options(apiKey)).json();
+const getBases = async (apiKey: string, endpointURL: string) => {
```
erezrokah, line 41: Made this function consistent with `getBaseTables`

plugins/source/airtable/package.json
```diff
@@ -66,7 +66,8 @@
     "eslint-plugin-you-dont-need-lodash-underscore": "^6.12.0",
     "prettier": "^3.0.1",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.5",
+    "winston": "^3.10.0"
```
erezrokah, line 70: Added `winston` as a dev dependency as we only import the type

mnorbury (approved):
👍

Issue #13163, opened 2023-08-17
bug(github): Inconsistent results based on org order 


mvbattista:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The actual total of repositories of organizations A-D (in the same Github enterprise) are:

- Organization A: 15
- Organization B: 103
- Organization C: 122
- Organization D: 128

Count of repos when orgs are in the following order in `spec.spec.orgs` for multiple clean runs

- A: 15
- C: 3
- B: 122
- D: 28

===

- B: 103
- C: 22
- D: 28
- A: 0

===

- C: 122
- A: 0
- B: 3
- D: 28

===

- D: 128
- A: 0
- C: 22
- B: 3

When each organization is isolated or in the first position will have the complete list.

### Expected Behavior

No matter what the order in `spec.spec.orgs`, all repositories should be pulled into the destination. We plan to configure this for many more organizations, but this bug is a major blocker.

### CloudQuery (redacted) config

```
kind: source
spec:
  destinations:
  - postgresql
  name: github
  path: cloudquery/github
  skip_tables:
  - github_releases
  - github_release_assets
  - github_repository_dependabot_alerts
  - github_repository_dependabot_secrets
  - github_organization_dependabot_alerts
  - github_organization_dependabot_secrets
  spec:
    app_auth:
    - org: <organization_A>
      private_key_path: cloudquery-github-app-key
      app_id: '<app_id>'
      installation_id: '<org_A_installation_id>'
    - org: <organization_B>
      private_key_path: cloudquery-github-app-key
      app_id: '<app_id>'
      installation_id: '<org_B_installation_id>'
    - org: <organization_C>
      private_key_path: cloudquery-github-app-key
      app_id: '<app_id>'
      installation_id: '<org_C_installation_id>'
    - org: <organization_D>
      private_key_path: cloudquery-github-app-key
      app_id: '<app_id>'
      installation_id: '<org_D_installation_id>'
    orgs:
    - <organization_A>
    - <organization_B>
    - <organization_C>
    - <organization_D>
  tables:
  - github_repositories
  - github_organization_members
  - github_organizations
  version: v7.1.3
---
kind: destination
spec:
  name: postgresql
  path: cloudquery/postgresql
  registry: github
  spec:
    connection_string: <connection_string>
  version: v5.0.0
```

### Steps To Reproduce

1. Install github app into multiple github organizations with all read-only permissions
2. Adjust config to account for app installations
[...]

PR #13182, opened 2023-08-18
fix: Re-init `RepositoryListByOrgOptions` for each organization when listing repositories

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/13163. We were re-using the list options (hence the pagination page) between organizations


erezrokah:
I'll open a separate PR to add make the discovery process concurrent with a `discovery_concurrency` option

yevgenypats (approved):

candiduslynx (approved):

PR #13188, opened 2023-08-18
feat: Add `discovery_concurrency` sync option

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/13163. Adds `discovery_concurrency` like other plugins to improve discovery speed (currently defaults to `1` as can be considered breaking).

Using the following config I saw a `25%` improvement (Google has ~2.5k repositories and Microsoft has ~5.8K`):

```yaml
kind: source
spec:
   ...
  tables:
    - github_repositories
  skip_dependent_tables: true
  spec:
    discovery_concurrency: -1
    orgs:
      - microsoft
      - google
  destinations:
    - postgresql
``` 

image

### DIFF
--- a/plugins/source/github/client/client.go
+++ b/plugins/source/github/client/client.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/beatlabs/github-auth/app/inst"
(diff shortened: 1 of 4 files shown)


mnorbury (approved):
Looks good. Only thought I had was what a negative discoveryConcurrency would do, but you have covered that clearly in the documentation. 👍

PR #13201, opened 2023-08-20
feat: Add optional login to our future registry

yevgenypats:
This adds an optional login to our CLI. This is not connected to our hub yet.

This works similarly to `gcloud auth login`. 

1) Goes to https://accounts.cloudquery.io 
2) Users logs in
3) browser send token to CLI via local http server
4) CLI saves token in `~/.cloudquery/token`

The command currently is Hidden because our hub is not up yet where it will be needed.


cli/cmd/root.go
```diff
@@ -30,6 +32,28 @@ Find more information at:
 	logFile         *os.File
 )
 
+func getUserConfigDir() string {
```
erezrokah, line 35: We should probably use https://github.com/adrg/xdg#xdg-base-directory to cover all use cases.
I think it might be better to use `XDG_CONFIG_HOME`. Then we won't need a CLI flag as we follow the `XDG` spec (i.e. if someone sets `XDG_CONFIG_HOME`)

erezrokah (changes requested):
Looks good overall, requesting changes due to the config comment. Please let me know what you think

cli/cmd/login.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	if serverErr != nil {
+		return serverErr
+	}
+
+	if refreshToken == "" {
+		return fmt.Errorf("failed to get refresh token")
+	}
+	tokenPath := path.Join(userConfig, "token")
```
erezrokah, line 174: What is the format of the token file?
Maybe it use be a JSON with the following format:

`~/.config/cloudquery/auth.json`

```js
{
  "version": 1,
  "token": ...
}
```

(We could even have a list of tokens for multiple accounts, but probably best to do it when we need it)

cli/cmd/root.go
yevgenypats: Nice! Will check this out.

candiduslynx:
I'll be overtaking the code + move to https://pkg.go.dev/github.com/pkg/browser instead of own impl

candiduslynx:
/gen sha=4384386092d0b78e0ba7fc31d323d4f94fec9887 component=cli

erezrokah:
> /gen sha=4384386092d0b78e0ba7fc31d323d4f94fec9887 component=cli

@candiduslynx This command is only supported for source plugins at the moment, see https://github.com/cloudquery/cloudquery/blob/9b6f9d60159b0ef26a003827be53b71256efb9ef/.github/workflows/manual_commands_triggers.yml#L40

cli/cmd/login.go
```diff
@@ -0,0 +1,167 @@
...
+		_, _ = io.Copy(io.Discard, resp.Body)
+		if resp.StatusCode == http.StatusOK {
+			return nil
+		}
+		return fmt.Errorf("failed to connect to local server. error code: %d", resp.StatusCode)
+	}, backoff.WithContext(backoff.NewConstantBackOff(100*time.Millisecond), ctx))
+}
+
+func runLogin(ctx context.Context) (err error) {
+	tokenFilePath, err := xdg.CacheFile("cloudquery/token")
```
erezrokah, line 83: I don't think `CacheFile` is the correct method to use, see https://wiki.archlinux.org/title/XDG_Base_Directory

image

erezrokah: `.cq` would make more sense for cache directory

candiduslynx: yeah, it's a typo (I wanted `xdg.ConfigFile`)

erezrokah: Actually we might want to use `DataHome` https://github.com/mislav/hub/issues/1912

candiduslynx: Done in 6a1a2ab895f36ccb57452b7c582a7835cfae16a7

candiduslynx:
@erezrokah tested the code manually (seems to work with current deployment)

candiduslynx (dismissed):

cli/cmd/login.go
```diff
@@ -0,0 +1,168 @@
...
+const (
+	loginShort = "Login to CloudQuery Hub."
+	loginLong  = `Login to CloudQuery Hub.
+
+This is required to download plugins from CloudQuery Hub.
+
+Local plugins and different registries don't need login.
+`
+
+	accountsURL = "https://accounts.cloudquery.io"
```
erezrokah, line 31: [non blocking] Can we add a hidden option to override this?

erezrokah (approved):

PR #13275, opened 2023-08-22
chore(deps): Update `github.com/cloudquery/plugin-sdk/v4` to v4.5.3

candiduslynx:
Extracted from https://github.com/cloudquery/cloudquery/pull/13262/


erezrokah (approved):
Looks good, we usually do SDK updates as a `fix`. This one has an arrow update too (not only the test changes) so could be a user visible change

erezrokah:
Never mind the auto merge got me before I could change it

PR #13289, opened 2023-08-23
fix(deps): Update `github.com/cloudquery/plugin-sdk/v4` to v4.5.5

candiduslynx:
Extracted from https://github.com/cloudquery/cloudquery/pull/13262
Part of https://github.com/cloudquery/cloudquery/issues/13153

### DIFF
--- a/plugins/destination/gremlin/client/write.go
+++ b/plugins/destination/gremlin/client/write.go
@@ -57,6 +57,7 @@ func (c *Client) WriteTableBatch(ctx context.Context, tableName string, msgs mes
 
 	g := gremlingo.Traversal_().WithRemote(session).V().HasLabel(tableName)
 	for i := range rows {
+		g = g.V().HasLabel(tableName)
 		for _, colName := range pks {
 			g = g.Has(colName, rows[i][colName])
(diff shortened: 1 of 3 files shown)


plugins/destination/gremlin/client/write.go
```diff
@@ -44,19 +43,21 @@ func (c *Client) WriteTableBatch(ctx context.Context, tableName string, msgs mes
...
+		for _, col := range table.Columns {
+			if !col.PrimaryKey {
+				valueColumns = append(valueColumns, col.Name)
+			}
 		}
 	}
 
 	g := gremlingo.Traversal_().WithRemote(session).V().HasLabel(tableName)
 	for i := range rows {
+		g = g.V().HasLabel(tableName)
```
candiduslynx, line 60: that's the bull of the fix, got it by looking at https://docs.aws.amazon.com/neptune/latest/userguide/gremlin-efficient-upserts.html#gremlin-upserts-vertices

bbernays (approved):
I am approving this because the tests pass. I have never used gremlin/neo4j though

PR #13309, opened 2023-08-24
feat(resources): Add `gcp_compute_osconfig_inventories` table

### LINKED ISSUE #13301: feat: Add `gcp_compute_osconfig_inventories ` to `gcp` source plugin
### Describe the resource

Info about host OS, version, etc.

### Use Case

Intersect with instances & provide more details on the vulnerability exposure

### Link to API documentation

https://pkg.go.dev/google.golang.org/api/osconfig/v1#ProjectsLocationsInstancesInventoriesListCall.Pages

### Additional Context

_No response_

### Pull request (optional)

candiduslynx:
Fixes #13301

Multiplex changes were made as listing inventories requires location param


erezrokah (commented):
Thanks for the PR @candiduslynx.
We usually don't discover a specific service locations during client initialization, since it doesn't allow users to filter those API calls. That is we will get the locations even if a user doesn't sync that table.

I would first try `/locations/-` or `/locations/global` to see if that works to return all instances, and if not, make instances a relation of locations, see example in https://github.com/cloudquery/cloudquery/blob/c51bd0bebb377329c99da7eb917c39f2ac49f294/plugins/source/gcp/resources/services/run/locations.go or https://github.com/cloudquery/cloudquery/blob/c51bd0bebb377329c99da7eb917c39f2ac49f294/plugins/source/gcp/resources/services/aiplatform/pipeline_locations.go

erezrokah:
I think the only place we use `ProjectLocationMultiplexEnabledServices` until now is for an API that doesn't support listing those so we hard code them (see https://github.com/cloudquery/cloudquery/blob/c51bd0bebb377329c99da7eb917c39f2ac49f294/plugins/source/gcp/resources/services/vision/products.go#L17)

candiduslynx:
@erezrokah I've moved the resource to be a child table of `gcp_compute_zones`

plugins/source/gcp/resources/services/compute/zones.go
```diff
@@ -46,13 +50,12 @@ func fetchZones(ctx context.Context, meta schema.ClientMeta, parent *schema.Reso
 	it := gcpClient.List(ctx, req, c.CallOptions...)
 	for {
 		resp, err := it.Next()
-		if err == iterator.Done {
-			break
-		}
 		if err != nil {
+			if errors.Is(err, iterator.Done) {
```
erezrokah, line 54: Can we not do this refactoring in this PR? I checked a few resources and they use the current pattern of not nesting the check for `iterator.Done`.

plugins/source/gcp/resources/services/compute/zones_test.go
```diff
@@ -18,6 +18,7 @@ func createZones(mux *httprouter.Router) error {
 	}
 	emptyStr := ""
 	zones.NextPageToken = &emptyStr
+	zones.Items = zones.Items[:1] // leave only 1
```
erezrokah, line 21: What's the reason this is needed? I thought  `faker.FakeObject` only adds a single item per array

candiduslynx: just to ensure that we indeed only have 1 call in the children resolvers

plugins/source/gcp/client/resolvers.go
```diff
@@ -14,9 +14,14 @@ import (
...
 )
 
-func ResolveProject(_ context.Context, meta schema.ClientMeta, r *schema.Resource, _ schema.Column) error {
+func ResolveProject(_ context.Context, meta schema.ClientMeta, r *schema.Resource, c schema.Column) error {
 	client := meta.(*Client)
-	return r.Set("project_id", client.ProjectId)
+	return r.Set(c.Name, client.ProjectId)
+}
+
+func ResolveLocation(_ context.Context, meta schema.ClientMeta, r *schema.Resource, c schema.Column) error {
```
erezrokah, line 22: It this used anywhere?

erezrokah (approved):

Issue #13281, opened 2023-08-22
bug: User report of delete-stale not working with CLI v13.1.1 and Postgres v5.0.1 when `--no-migrate` is used


hermanschaaf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Discord link: https://discord.com/channels/872925471417962546/873606591335759872/1143482036317868093

> Hi Team, Noticed that in cloudquery cli version v3.13.1, while we are syncing aws resources to postgres with   write_mode: "overwrite-delete-stale", migrate_mode: forced & pk_mode: cq-id-only , stale entries are not getting deleted. are we tracking this issue for future release?

> aws: "v22.5.1" pgx: "v5.0.1"

### Expected Behavior

Stale entries should be deleted.

### CloudQuery (redacted) config

```
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v5.0.1"
  write_mode: "overwrite-delete-stale"
  migrate_mode: forced
  pk_mode: cq-id-only
  spec:
    connection_string: <dsn>
---
kind: source
spec:
  name: aws
  path: "cloudquery/aws"
  version: "v22.5.1"
  tables: <tables>
  destinations: <dest>
  deterministic_cq_id: true
```

### Steps To Reproduce

Create a resource, then run a sync. Delete that resource and run a sync with `--no-migrate`. You should see that the resource is not deleted, and shows up in the DB with an old sync time.

### CloudQuery (redacted) logs

N/A

### CloudQuery version

v13.1.1

### Pull request (optional)

- [ ] I can submit a pull request

PR #13310, opened 2023-08-24
fix: Send delete stale when `--no-migrate` is used

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/13281

### DIFF
--- a/cli/cmd/sync_v3.go
+++ b/cli/cmd/sync_v3.go
@@ -197,15 +197,15 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
 				}
 			}
 		case *plugin.Sync_Response_MigrateTable:
-			if noMigrate {
-				continue
-			}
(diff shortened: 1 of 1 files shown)


cli/cmd/sync_v3.go
```diff
@@ -197,15 +197,15 @@ func syncConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, d
...
 		case *plugin.Sync_Response_MigrateTable:
-			if noMigrate {
-				continue
-			}
 			sc, err := plugin.NewSchemaFromBytes(m.MigrateTable.Table)
 			if err != nil {
 				return err
 			}
 			tableName := tableNameFromSchema(sc)
 			tables[tableName] = true
```
erezrokah, line 205: We use the `tables` map later in the code to send delete stale messages

candiduslynx (approved):

PR #13315, opened 2023-08-24
chore: Bring back plugin checksums, remove signing

erezrokah:
#### Summary

Plugin checksums have been broken since https://github.com/cloudquery/cloudquery/pull/4286 and https://github.com/cloudquery/cloudquery/pull/3471.
This PR brings them back.

It also removes checksums signing as that didn't work as we were not passing the private key to GoReleaser (it used a default key from the machine it's running on).
See https://github.com/cloudquery/cloudquery/blob/2a11e99c553aa310ef062a4d2797cedf8b958ca1/plugins/.goreleaser.template.yaml#L23 vs https://github.com/goreleaser/goreleaser-action/tree/a10d0e32dce991f86e53faa7b4f269dedb82632f#signing.


.github/workflows/release_plugin.yml
```diff
@@ -155,16 +155,10 @@ jobs:
...
-      - name: Import GPG key
-        id: import_gpg
-        uses: crazy-max/ghaction-import-gpg@v5
-        with:
-          gpg_private_key: ${{ secrets.PRIVATE_KEY }}
       - name: Run GoReleaser
         run: goreleaser release --timeout 50m --clean -f ./${{needs.prepare.outputs.plugin_dir}}/.goreleaser.yaml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
```
erezrokah, line 167: I accidentally commented this out in https://github.com/cloudquery/cloudquery/pull/12707. But it doesn't work anyway

plugins/.goreleaser.template.yaml
```diff
@@ -10,21 +10,11 @@ archives:
       - policies/**
 checksum:
   name_template: "checksums.txt"
-  ids:
```
erezrokah, line 13: `ids` need to match the ones in `archives` or remove them to checksum everything

plugins/.goreleaser.template.yaml
```diff
@@ -10,21 +10,11 @@ archives:
...
-    - default
 changelog:
   sort: asc
   filters:
     exclude:
       - "^docs:"
       - "^test:"
-signs:
-  - artifacts: checksum
-    args:
```
erezrokah, line 23: This should be `args: ["--batch", "-u", "{{ .Env.GPG_FINGERPRINT }}", "--output", "${signature}", "--detach-sign", "${artifact}"]` for the key from GitHub actions secrets to be used

candiduslynx (approved):

PR #13318, opened 2023-08-24
fix: Improve Postgres performance

erezrokah:
#### Summary

Still need to figure out how to replace the logic I deleted, but listing information schema on each batch insert is creating a bottleneck (saw it on `pprof`). Probably due to read/write locks or our queries to list tables and columns are slow.
Changes in this PR result in about x10 improvement. 

Spec used (I used an old source since I tested on an old Postgres version to get the expected numbers):

```yaml
kind: source
spec:
  name: aws
  registry: "github"
  path: "cloudquery/aws"
  version: "v19.0.0"
  tables:
    - "*"
  skip_tables:
    - "aws_cloudtrail_*"
    - "aws_iam_*"
    - "aws_servicequotas_*"
  destinations:
    - postgresql
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v5.0.5"
  migrate_mode: forced
  spec:
```

Before:
image

After (with Postgres running from localhost with this fix):
image

### DIFF
--- a/plugins/destination/postgresql/client/spec.go
+++ b/plugins/destination/postgresql/client/spec.go
@@ -8,8 +8,8 @@ import (
 
 const (
 	defaultBatchSize      = 10000
-	defaultBatchSizeBytes = 1000000
-	defaultBatchTimeout   = 10 * time.Second
+	defaultBatchSizeBytes = 100000000
+	defaultBatchTimeout   = 60 * time.Second
 )
 
 type Spec struct {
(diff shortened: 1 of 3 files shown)


plugins/destination/postgresql/client/spec.go
```diff
@@ -8,8 +8,8 @@ import (
 
 const (
 	defaultBatchSize      = 10000
-	defaultBatchSizeBytes = 1000000
-	defaultBatchTimeout   = 10 * time.Second
+	defaultBatchSizeBytes = 100000000
```
candiduslynx, line 11: also should be in the doc + should be a major bump

erezrokah:
Going to split this PR so we can release the removal of list tables in a non breaking change

erezrokah:
Closing in favor of https://github.com/cloudquery/cloudquery/pull/13324 and https://github.com/cloudquery/cloudquery/pull/13323

PR #13323, opened 2023-08-25
fix: Don't list Postgres tables during insert

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/13318. Witnessed the bottleneck using `pprof`.
With default batch settings I'm getting `2m29s` sync time instead of ~`11m`,
With `batch_size_bytes: 100000000` and `batch_timeout: 60s` I'm getting `1m40s`. (Fixed in https://github.com/cloudquery/cloudquery/pull/13324)

Used spec:

```yaml
kind: source
spec:
  name: aws
  path: "cloudquery/aws"
  version: "v19.0.0"
  tables:
    - "*"
  skip_tables:
    - "aws_cloudtrail_*"
    - "aws_iam_*"
    - "aws_servicequotas_*"
  destinations:
    - postgresql
---
kind: destination
spec:
  name: "postgresql"
  registry: "grpc"
  path: localhost:8888
  # path: cloudquery/postgresql
  # version "v5.0.5"
  spec:
    # batch_size_bytes: 100000000
    # batch_timeout: "60s"
```

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -25,7 +25,7 @@ func (c *Client) MigrateTableBatch(ctx context.Context, messages message.WriteMi
 	if err != nil {
 		return fmt.Errorf("failed listing postgres tables: %w", err)
 	}
-	tables = c.normalizeTables(tables, pgTables)
+	tables = c.normalizeTables(tables)
 
 	safeTables := map[string]bool{}
 	for _, msg := range messages {
@@ -80,7 +80,7 @@ func (c *Client) MigrateTableBatch(ctx context.Context, messages message.WriteMi
 	return nil
 }
 
-func (c *Client) normalizeTable(table *schema.Table, pgTable *schema.Table) *schema.Table {
+func (c *Client) normalizeTable(table *schema.Table) *schema.Table {
 	normalizedTable := schema.Table{
 		Name: table.Name,
 	}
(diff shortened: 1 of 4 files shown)


plugins/destination/postgresql/client/migrate.go
```diff
@@ -189,10 +187,10 @@ func (c *Client) addColumn(ctx context.Context, tableName string, column schema.
 
 func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table) error {
 	var sb strings.Builder
-	tName := table.Name
-	tableName := pgx.Identifier{tName}.Sanitize()
+	tableName := table.Name
```
erezrokah, line 190: Renamed these since I initially used `tableName` for the dictionary key which is wrong since it had the sanitized value

yevgenypats (approved):

plugins/destination/postgresql/client/migrate.go
```diff
@@ -216,19 +214,22 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		}
 	}
 
+	pkConstraintName := tableName + "_cqpk"
```
candiduslynx, line 217: shouldn't it actually use the `c.pgTablesToPKConstraints` value if that's available?

erezrokah: This is inside `createTableIfNotExist` so there should not be any value

candiduslynx: right, but what [about](https://pkg.go.dev/github.com/cloudquery/plugin-sdk/v4@v4.5.5/schema#Table) `table.PkConstraintName`?

PR #13324, opened 2023-08-25
feat: Increase default batch size bytes and timeout

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/13318.

BEGIN_COMMIT_OVERRIDE
feat: Increase default batch size bytes to `100000000` (100 MB) and default batch timeout to `60` seconds.
BREAKING-CHANGE: Increase default batch size bytes to `100000000` (100 MB) and default batch timeout to `60` seconds. We discovered a default higher batch size bytes and timeout settings provide better out of the box performance for the PostgreSQL destination. We're marking it as a breaking change as it might increase memory consumption in some environments.

END_COMMIT_OVERRIDE


plugins/destination/postgresql/client/spec.go
```diff
@@ -8,8 +8,8 @@ import (
 
 const (
 	defaultBatchSize      = 10000
-	defaultBatchSizeBytes = 1000000
-	defaultBatchTimeout   = 10 * time.Second
+	defaultBatchSizeBytes = 100000000
```
candiduslynx, line 11: add doc change as well

candiduslynx (changes requested):

plugins/destination/postgresql/client/spec.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/13324/commits/8e4f613b79aee3ff25c4d20d4870f7fe7b7c700b

erezrokah:
I'll hold off merging this until next week so we can release https://github.com/cloudquery/cloudquery/pull/13323 and https://github.com/cloudquery/cloudquery/pull/13327 in a non breaking change

candiduslynx (approved):

PR #13331, opened 2023-08-25
refactor: Clean up list tables method

erezrokah:
#### Summary

`exclude` was always passed as an empty list. Also we can use `tables.TableNames()`

### DIFF
--- a/plugins/destination/postgresql/client/list_tables.go
+++ b/plugins/destination/postgresql/client/list_tables.go
@@ -62,10 +62,10 @@ ORDER BY
 	table_name ASC, ordinal_position ASC;
 `
 
-func (c *Client) listTables(ctx context.Context, include, exclude []string) (schema.Tables, error) {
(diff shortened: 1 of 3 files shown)


plugins/destination/postgresql/client/list_tables.go
```diff
@@ -102,17 +102,14 @@ func (c *Client) listTables(ctx context.Context, include, exclude []string) (sch
...
+func (c *Client) whereClause(include []string) string {
+	if len(include) == 0 {
 		return ""
 	}
 	var where string
 	if len(include) > 0 {
 		where = fmt.Sprintf("AND pg_class.relname IN (%s)", c.inClause(include))
 	}
-	if len(exclude) > 0 {
-		where = fmt.Sprintf("AND pg_class.relname NOT IN (%s)", c.inClause(exclude))
```
erezrokah, line 114: I think there was a bug here. There should be a space between the include and exclude `AND` clauses

candiduslynx (approved):

PR #13332, opened 2023-08-25
fix: Drop include clause from list tables

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/13331.
On AWS this means we search an array of ~600 items. Since at the moment we only have a single `MigrateTableBatch` call for all the tables, no point of having the `WHERE` clause (and even with multiple migrate batches might be faster just to query everything).

See also https://www.postgresql.org/docs/current/arrays.html:
image
We can't really follow that tip though


hermanschaaf (approved):

hermanschaaf:
Looks good, though I don't know if this should be marked as a `refactor`, as it does change behavior?

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -62,10 +61,10 @@ ORDER BY
...
 `
 
-func (c *Client) listTables(ctx context.Context, include []string) (schema.Tables, error) {
+func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
 	c.pgTablesToPKConstraints = map[string]string{}
 	var tables schema.Tables
-	whereClause := c.whereClause(include)
+	var whereClause string
 	if c.pgType == pgTypeCockroachDB {
 		whereClause += " AND information_schema.columns.is_hidden != 'YES'"
```
candiduslynx, line 69:
```suggestion
	var whereClause string
	if c.pgType == pgTypeCockroachDB {
		whereClause = " AND information_schema.columns.is_hidden != 'YES'"
```

candiduslynx (approved):

erezrokah:
> Looks good, though I don't know if this should be marked as a `refactor`, as it does change behavior?

Yeah, I'm not sure. It shouldn't have any user facing impact, but maybe I'll do a `fix` to play it safe

PR #13357, opened 2023-08-28
doc: Event-Based Sync for AWS Plugin

bbernays:


website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,91 @@
+import { Callout } from 'nextra-theme-docs';
+
+<Callout type="info">
+This feature is currently in a closed preview. Sign up for access using this [form](https://admin.typeform.com/form/Dvdn6P2u/create?block=a48d83b4-e41c-4e1c-a980-89f769933f7e)
+</Callout>
+
+# Event Based Sync
```
hermanschaaf, line 7:
```suggestion
# Event-Based Sync
```

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,91 @@
...
+cloudquery sync config.yml
+```
+
+This will start a long lived process that will only stop when there is an error or you stop the process
+
+
+### Limitations:
+- Kinesis Stream can only have a single shard. (This is a limitation that we expect to remove in the future)
+- Stale records will only be deleted if the plugin stops consuming the Kinesis Stream, which only can occur if there is an error
+- Incremental table's state is not updated by event based syncs
```
hermanschaaf, line 91:
```suggestion
- Incremental tables' state is not updated by event based syncs
```

Are any incremental tables supported by event-based syncs right now? Maybe we can remove this limitation?

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,91 @@
...
+    - aws_iam_user_last_accessed_details
+  spec:
+    event_based_sync:
+      - account:
+          local_profile: "<ROLE-NAME>"
+        kinesis_stream_arn: <OUTPUT-FROM-CLOUDFORMATION-STACK>
+```
+
+4. Sync the data! 
+```
```
hermanschaaf, line 81:
```suggestion
```bash
```

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,91 @@
...
+## Configuration:
+
+1. Configure a AWS CloudTrail Trail to send management events to a Kinesis Data Stream via CloudWatch Logs. The most straight forward way to do this is to use the CloudQuery provided CloudFormation template.
+
+```
+aws cloudformation deploy --template-file ./streaming-deployment.yml --stack-name <STACK-NAME> --capabilities CAPABILITY_IAM --disable-rollback --region <DESIRED-REGION>
+```
+
+2. Copy the ARN of the Kinesis stream. If you used the CloudFormation template you can run the following command:
+```
```
hermanschaaf, line 45:
```suggestion
```bash
```

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,91 @@
...
+iam.amazonaws.com | CreateUser
+rds.amazonaws.com | CreateDBCluster
+rds.amazonaws.com | CreateDBInstance
+
+
+## Configuration:
+
+1. Configure a AWS CloudTrail Trail to send management events to a Kinesis Data Stream via CloudWatch Logs. The most straight forward way to do this is to use the CloudQuery provided CloudFormation template.
+
+```
```
hermanschaaf, line 40:
```suggestion
```bash
```

website/pages/docs/plugins/sources/aws/configuration.mdx
```diff
@@ -263,6 +268,28 @@ This is used to specify one or more accounts to extract information from. Note t
...
+
+  ARN for the Kinesis stream that will hold all of the CloudTrail records
+
+- `start_time` (timestamp) 
+
+  Defines the place in the stream where record processing should begin. By default, the time at which the sync began will be used.
+
+- `event_based_sync_only` (`bool`) (default: `false`)
+
+  By default CQ will do a full sync on the specified tables before starting to consume the events in the stream. This parameter enables users to skip the pull based sync and go straight to the streaming sync.
```
hermanschaaf, line 286:
```suggestion
  By default CQ will do a full sync on the specified tables before starting to consume the events in the stream. This parameter enables users to skip the pull-based sync and go straight to the streaming sync.
```

website/pages/docs/plugins/sources/aws/configuration.mdx
```diff
@@ -159,26 +155,35 @@ The `table_options` parameter is deprecated and will be removed soon. Please rea
...
   ```
+    - [ListFindings](https://docs.aws.amazon.com/access-analyzer/latest/APIReference/API_ListFindings.html)
+    - [ListMetrics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_ListMetrics.html)
+    - [GetMetricStatistics](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html)
+    - [LookupEvents](https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_LookupEvents.html)
+    - [ListFindings](https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListFindings.html)
+    - [GetFindings](https://docs.aws.amazon.com/securityhub/1.0/APIReference/API_GetFindings.html)
+    - [ListTasks](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListTasks.html)
+
+- **enterprise version only** `streaming_sync` (`array`) (default: empty)
```
hermanschaaf, line 186:
```suggestion
- **enterprise version only** `event_based_sync` (`array`) (default: empty)
```

website/pages/docs/plugins/sources/aws/configuration.mdx
```diff
@@ -263,6 +268,28 @@ This is used to specify one or more accounts to extract information from. Note t
 
   List of OU member accounts to skip. This is useful in conjunction with `organization_units` if there are accounts under the selected OUs that should be ignored.
 
+### event_based_sync
```
hermanschaaf, line 271:
```suggestion
### event_based_sync

Event-based syncing is only supported by the enterprise-version AWS plugin. [Read more about event-based syncs]([/docs/plugins/sources/aws/event-based-sync](https://github.com/cloudquery/cloudquery/pull/13357/files#diff-d6b3e143b73c78cee7e26fed26091b5d29df661282616ba5ba58f010dadee2be)).
```

hermanschaaf: (I think we should just re-iterate that this is enterprise-only, in case users didn't see it in the section higher up)

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+4. Sync the data! 
+```
+cloudquery sync config.yml
+```
+
+This will start a long lived process that will only stop when there is an error or you stop the process
+
+
+### Limitations:
```
hermanschaaf, line 88:
```suggestion
### Limitations
```

hermanschaaf (commented):
Looks good, just left a couple of suggestions

hermanschaaf (approved):

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,92 @@
...
+ec2.amazonaws.com | RevokeSecurityGroupIngress
+ec2.amazonaws.com | RunInstances
+iam.amazonaws.com | CreateGroup
+iam.amazonaws.com | CreateRole
+iam.amazonaws.com | CreateUser
+rds.amazonaws.com | CreateDBCluster
+rds.amazonaws.com | CreateDBInstance
+
+
+## Configuration
```
erezrokah, line 38: Can we add a diagram of how this setup looks like?

For people not using the service yet it would be useful to know what's required on their end. Especially since the `streaming-deployment.yml` is private (maybe it shouldn't be).

If we could add a step by step guide how on to do it in the console that would be great (but non blocking for the PR). Maybe link to existing AWS docs/other guides regarding this topic

erezrokah: Maybe we could create something similar to https://github.com/nexthink-oss/terraform-aws-cloudtrail-s3-to-kinesis/

plugins/source/aws/client/spec.go
```diff
@@ -36,6 +38,12 @@ type AwsOrg struct {
 	ChildAccountRegions         []string `json:"member_regions,omitempty"`
 }
 
+type EventBasedSync struct {
+	Account            Account    `json:"account"`
+	KinesisStreamARN   string     `json:"kinesis_stream_arn"`
+	StartTime          *time.Time `json:"start_time,omitempty"`
+	EventBasedSyncOnly bool       `json:"event_based_sync_only"`
```
erezrokah, line 45: I would remove both `StartTime` and `EventBasedSyncOnly` if possible for simplicity.

If you have `event_based_sync` configured it should only do event based. If users wants a full sync they can have a separate spec without `event_based_sync`.

Same for `start_time`. Can we make so it syncs whatever that's in the stream? Otherwise users will have to track when they started and stopped the CLI. The CLI can even stop due to an unexpected error which makes it harder to track sync time and ensure the same event is not handled twice. `start_time` or any other checkpoints might be useful if there are other consumers to the stream, and we need to track what we've read.

If users want to clean a stream they can always `aws kinesis get-records --shard-iterator ...` to drop messages.

erezrokah: Maybe keep `start_time` but make it so the default syncs from whatever is unprocessed in the stream

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,92 @@
...
+    - aws_ec2_security_groups
+    - aws_ec2_subnets
+    - aws_ec2_vpcs
+    - aws_ecs_cluster_tasks
+    - aws_iam_groups
+    - aws_iam_roles
+    - aws_iam_users
+    - aws_rds_instances
+  destinations: ["postgresql"]
+  skip_tables:
```
erezrokah, line 71: Is skipping these tables required? What happens if we remove this part of the configuration?

bbernays: Since we only support `create` events right now and not modify, those records would always be nil/empty but take extra API calls, which adds to latency.... It is a tradeoff between latency and actionable data

plugins/source/aws/client/spec.go
bbernays: You cannot delete messages/records out of a kinesis stream. There is no way of knowing what is processed/unprocessed

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
erezrokah: Can you add a bit more context? Why those records will always by `nil` or empty? And why would we need to do extra API calls for them? If we know these are not supported in advance, could we skip them in the code?

bbernays: Currently we are only supporting triggering syncing on top level tables... These are nested tables... We could just remove the skip tables and fix the performance issues if they arise... this might have been a pre mature optimization

erezrokah: Is there a reason the plugin can't ignore what it doesn't support?

bbernays: I don't think we want to skip all dependent tables.

What do you mean ignore?

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
bbernays: Yeah I think more information about this setup could definitely be added. Especially before this feature gets released publicly. Right now I am not that concerned because it is a closed preview

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
erezrokah: > I don't think we want to skip all dependent tables.

We can if this is only an event based sync

> What do you mean ignore?

Maybe I'm missing something, but the plugin knows which tables it supports, so I'm not sure why the user needs to configure those in `skip_tables`.

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
erezrokah: > Right now I am not that concerned because it is a closed preview

The idea is to explain how the features works in a public to everyone, not only to the private beta users. If people don't know how this works, they might not register in the first place

bbernays: Fair. We can use the diagram that Michel has in the blog post

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
bbernays: It is a balance between performance and syncing data... We do not know how the user will be deploying this. If a user is syncing from a stream and catching up to real time, a user might not want to skip these tables as time has elapsed and there might be valuable usage data returned... So I don't think we can change any of the code to optimize for these different syncing patterns

erezrokah: So I'm probably missing something because I thought we don't support these tables as a part of the event based sync.

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,96 @@
...
+    - aws_ecs_cluster_tasks
+    - aws_iam_groups
+    - aws_iam_roles
+    - aws_iam_users
+    - aws_rds_instances
+  destinations: ["postgresql"]
+  spec:
+    event_based_sync:
+      - account:
+          local_profile: "<ROLE-NAME>"
```
erezrokah, line 82:
```suggestion
     ## - account:
     ##     local_profile: "<ROLE-NAME>"
```

Maybe comment these as the defaults might work for people

website/pages/docs/plugins/sources/aws/configuration.mdx
```diff
@@ -265,6 +274,36 @@ This is used to specify one or more accounts to extract information from. Note t
...
+
+- `kinesis_stream_arn` (string) (**required**)
+
+  ARN for the Kinesis stream that will hold all of the CloudTrail records
+
+- `start_time` (timestamp) 
+
+  Defines the place in the stream where record processing should begin. By default, the time at which the sync began will be used.
+
+- `event_based_sync_only` (`bool`) (default: `false`)
```
erezrokah, line 298:
```suggestion
- `full_sync` (`bool`) (default: `true`)
```

website/pages/docs/plugins/sources/aws/configuration.mdx
```diff
@@ -265,6 +274,36 @@ This is used to specify one or more accounts to extract information from. Note t
...
+
+  By default CQ will do a full sync on the specified tables before starting to consume the events in the stream. This parameter enables users to skip the pull-based sync and go straight to the streaming sync.
```
erezrokah, line 300:
```suggestion
  By default CQ will do a full sync on the specified tables before starting to consume the events in the stream. This parameter enables users to skip the full sync and go straight to the event based sync.
```

website/pages/docs/plugins/sources/aws/event-based-sync.mdx
```diff
@@ -0,0 +1,96 @@
...
+    - aws_ec2_internet_gateways
+    - aws_ec2_security_groups
+    - aws_ec2_subnets
+    - aws_ec2_vpcs
+    - aws_ecs_cluster_tasks
+    - aws_iam_groups
+    - aws_iam_roles
+    - aws_iam_users
+    - aws_rds_instances
+  destinations: ["postgresql"]
```
erezrokah, line 78:
```suggestion
  destinations: ["DESTINATION_NAME"]
```

We use this placeholder to keep the naming consistent across the Website

website/public/images/docs/aws/event-based-sync-architecture.png
erezrokah, line 1: We should use the up to date diagram from https://github.com/cloudquery/cloudquery/pull/13618/files#diff-fbaa609d4c9b0848f3d357ae90fd3e7f30ba71cc0f93679e4d9f24964cfe0f3e

bbernays: I actually didn't update the diagram on purpose, this diagram shows what the cloudformation is meant to highlight what the cloudformation is creating. The up to date diagram is highlighting the entire data flow. But I will defer to @pilvikala on this one

pilvikala: Yeah, I don't think they need to be the same, I agree with @bbernays.

erezrokah: Cool, makes sense

erezrokah (approved):

PR #13369, opened 2023-08-29
doc: Update configuration for some aws, github, gitlab, oracle and typeform

hermanschaaf:
Some doc updates, mostly around not using `tables: ["*"]`, or in the case of AWS, suggesting the `aws_ec2_instances` table rather than `aws_s3_buckets` (which has child tables).


website/pages/docs/plugins/sources/aws/configuration.mdx
```diff
@@ -27,6 +27,8 @@ spec:
     aws_debug: false
 ```
 
+See [tables](/docs/plugins/sources/aws/tables) for a list of all available tables.
```
hermanschaaf, line 30: I'm adding this in a couple of places because right now our website's sidebar is so long that the page you're on is not always above the fold. We should improve this in the new docs website

erezrokah (approved):

PR #13411, opened 2023-08-29
fix: Call `types.RegisterAllExtensions` only once in `tables` command

erezrokah:
#### Summary

Noticed while working on https://github.com/cloudquery/cloudquery/pull/13410. `RegisterAllExtensions` fails if you call it multiple times. To reproduce run on a spec with multiple sources


erezrokah:
Looking at the build failures

erezrokah:
Ah, need to use a newer Go version

cli/cmd/tables_test.go
```diff
@@ -32,40 +32,35 @@ func TestTables(t *testing.T) {
...
 	}
 
 	for _, tc := range configs {
 		t.Run(tc.name, func(t *testing.T) {
 			defer CloseLogFile()
-			cmd, _ := getTablesCommand(t, tc.config, tc.format)
+			cmd, cqDir := getTablesCommand(t, tc.config, tc.format)
 			commandError := cmd.Execute()
-			require.Error(t, commandError, tc.err)
+			require.NoError(t, commandError)
```
erezrokah, line 57: These tests were passing since we were still throwing an error. When I wrote these I should have validated the error message too

cli/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cloudquery/cloudquery/cli
 
-go 1.20
+go 1.21.0
```
erezrokah, line 3: Need for `OnceValue`

cli/cmd/tables_v3.go
```diff
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	gosync "sync"
```
candiduslynx, line 6:
```suggestion
	"sync"
```

cli/cmd/tables_v3.go
```diff
@@ -11,8 +12,10 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/types"
 )
 
+var registerOnce = gosync.OnceValue(types.RegisterAllExtensions)
```
candiduslynx, line 15:
```suggestion
var registerOnce = sync.OnceValue(types.RegisterAllExtensions)
```

cli/cmd/tables_v3.go
erezrokah: We already have a package named `sync` that's the reason for the different import name

erezrokah: Here https://github.com/cloudquery/cloudquery/blob/cc30fb09d1a039948754cbda09a1adf1b58d0c54/cli/cmd/sync.go#L76

candiduslynx (approved):

Issue #13412, opened 2023-08-29
bug: aws_cloudtrail_events table resyncs all data even when backend options are specified


hermanschaaf:
This seems to be because we are not flushing the backend.

hermanschaaf:
This will be fixed in AWS source `v22.8.2` and Postgres destination `v6.0.2`.

PR #13429, opened 2023-08-29
fix: Drop include clause from list tables

hermanschaaf:
Somehow the change in https://github.com/cloudquery/cloudquery/pull/13332 got reverted, I think unintentionally.

This is however necessary as part of a fix for https://github.com/cloudquery/cloudquery/issues/13412

PR #13430, opened 2023-08-29
fix: Flush state backend

hermanschaaf:
Closes https://github.com/cloudquery/cloudquery/issues/13412 (together with https://github.com/cloudquery/cloudquery/pull/13429)


erezrokah (approved):
Should we do this for other stateful plugins?

PR #13445, opened 2023-08-30
fix: Remove stale code

candiduslynx:
Caused by https://github.com/cloudquery/cloudquery/pull/13341

### DIFF
--- a/plugins/destination/postgresql/client/spec.go
+++ b/plugins/destination/postgresql/client/spec.go
@@ -8,8 +8,8 @@ import (
 
 const (
 	defaultBatchSize      = 10000
-	defaultBatchSizeBytes = 1000000
-	defaultBatchTimeout   = 10 * time.Second
+	defaultBatchSizeBytes = 100000000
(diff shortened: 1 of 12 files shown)


yevgenypats (approved):

erezrokah (approved):

plugins/destination/postgresql/client/spec.go
```diff
@@ -8,8 +8,8 @@ import (
 
 const (
 	defaultBatchSize      = 10000
-	defaultBatchSizeBytes = 1000000
-	defaultBatchTimeout   = 10 * time.Second
+	defaultBatchSizeBytes = 100000000
```
erezrokah, line 11: We should release a new Postgres version with this

erezrokah: I made https://github.com/cloudquery/cloudquery/releases/tag/plugins-destination-postgresql-v6.0.1 a pre-release since it has the old batch settings and also doesn't have https://github.com/cloudquery/cloudquery/pull/13332

PR #13475, opened 2023-08-30
feat: Allow use of backend that is not in list of destinations

hermanschaaf:
Discovered while writing the documentation for backend options: we should be able to support using a backend that is different from the destination being synced to, but currently this doesn't work. 

Without the change in this PR, destinations like BigQuery and Snowflake wouldn't be able to use state with incremental tables. With this change, it can be achieved by using a different destination as the backend (e.g. PostgreSQL, SQLite, or any other that supports `overwrite` mode)


erezrokah (approved):

hermanschaaf:
Had to make some adjustments to the variable replacement code, but the tests are now passing on Windows as well

Issue #12791, opened 2023-08-07
bug(docs): Managing Incremental Tables doc is out of date


erezrokah:
We should update https://www.cloudquery.io/docs/advanced-topics/managing-incremental-tables per the new v4 implementation of backends via PostgreSQL. See:

https://github.com/cloudquery/cloudquery/issues/6665#issuecomment-1625575560
https://github.com/cloudquery/cloudquery/issues/6665#issuecomment-1665737717
https://www.cloudquery.io/blog/proto-v3#unified-protocol

PR #13480, opened 2023-08-30
doc: Update docs about backend state management

hermanschaaf:
Includes some updates and fixes related to backend state management.


hermanschaaf:
> Fixes https://github.com/cloudquery/cloudquery/issues/12791 probably

@erezrokah Ah nice yeah I think so, thanks

PR #13482, opened 2023-08-30
fix: Skip incremental tables when performing delete-stale

hermanschaaf:
This fixes a bug introduced with the migration to SDK v4 / Protocol v3; when `overwrite-delete-stale` write mode is used, stale entries should not be removed from incremental tables when a state backend is active.


hermanschaaf:
> Looks good, we might need it in sync_v2 too

@erezrokah You're probably right; v2 was still a batch protocol so it's quite different, it's not immediately clear to me how we can fix it there. I think I'm going to defer fixing this, and if we get a report of a community v2 source that also uses incremental tables we can take another look (or upgrade the source to v3, which might be easier)

PR #13494, opened 2023-08-31
chore: Add factors.ai to Website

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/588 (internal issue).

Opening for feedback


website/components/_app.tsx
```diff
@@ -73,6 +82,10 @@ export default function Nextra({ Component, pageProps }) {
           referrerPolicy="no-referrer-when-downgrade"
         />
       </noscript>
+      <CQCookieConsent
+        onAccept={() => setConsent(true)}
+        onDecline={() => setConsent(false)}
```
hermanschaaf, line 87: When consent is withdrawn, should this also clear the existing cookies?

hermanschaaf: Ok, as long as the scripts check the value of that cookie it should be fine. But the cookies are technically on our domain so I would expect we can delete/unset them

erezrokah: It's done in https://github.com/cloudquery/cloudquery/pull/13494/files#diff-a4b784516284a9bb578fb84489f0480b3982436caeaace9b094c81297a65a48fR76

hermanschaaf (approved):

Issue #13555, opened 2023-08-31
feat(github): Support reading private key from string


patriknordlen:
### Which problem is this feature request solving?

When authenticating using Github App credentials, the `github` source plugin currently supports only reading the key material from a file via the `private_key_path` configuration directive.

In scenarios like when cloudquery is run in a CI pipeline job, any key material is typically injected via environment variables so currently one needs to have a separate build step where the contents of the environment variable holding secrets are echoed into a file, so that cloudquery can then use it. This works but is cumbersome.

### Describe the solution you'd like

I'd like to have a configuration directive `private_key` that lets me supply the private key as a string, for instance from an environment variable. This would be mutually exclusive with `private_key_path` to avoid confusion about what key material is being used.

### Pull request (optional)

- [X] I can submit a pull request

erezrokah:
Hi @patriknordlen, thanks for the issue. I marked it as `good first issue` if you'd like to contribute. Please note we support env variables substitution https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution

patriknordlen:
> Hi @patriknordlen, thanks for the issue. I marked it as `good first issue` if you'd like to contribute. Please note we support env variables substitution https://www.cloudquery.io/docs/advanced-topics/environment-variable-substitution

Great, I have an incoming PR for it. Arriving shortly. :)

erezrokah:
Resolved by https://github.com/cloudquery/cloudquery/pull/13556 and released in https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-github-v7.3.0

Thanks @patriknordlen!

PR #13556, opened 2023-08-31
feat: Support reading private key from string

patriknordlen:
#### Summary

This PR addresses #13555 by adding a `private_key` configuration key that can be used to read the private key from a string instead of a file.


plugins/source/github/client/client.go
```diff
@@ -81,7 +82,12 @@ func New(ctx context.Context, logger zerolog.Logger, spec Spec) (schema.ClientMe
 
 	ghServices := map[string]GithubServices{}
 	for _, auth := range spec.AppAuth {
-		k, err := key.FromFile(auth.PrivateKeyPath)
+		var k *rsa.PrivateKey
```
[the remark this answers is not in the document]
patriknordlen: Thanks for pushing a fix! The PR is ready for review.

plugins/source/github/client/spec.go
```diff
@@ -47,8 +48,11 @@ func (s *Spec) Validate() error {
 		if appAuth.Org == "" {
 			return fmt.Errorf("missing org in app auth configuration")
 		}
-		if appAuth.AppID != "" && appAuth.PrivateKeyPath == "" {
-			return fmt.Errorf("missing private key path in configuration")
+		if appAuth.AppID != "" && (appAuth.PrivateKeyPath == "" && appAuth.PrivateKey == "") {
+			return fmt.Errorf("missing private key specification in configuration")
```
erezrokah, line 52:
```suggestion
			return fmt.Errorf("missing private key specification in configuration. Please specify using either `private_key` or `private_key_path`")
```

plugins/source/github/client/spec.go
```diff
@@ -47,8 +48,11 @@ func (s *Spec) Validate() error {
...
+		}
+		if appAuth.AppID != "" && (appAuth.PrivateKeyPath != "" && appAuth.PrivateKey != "") {
+			return fmt.Errorf("both private key and private key path specified in configuration")
```
erezrokah, line 55:
```suggestion
			return fmt.Errorf("both private key and private key path specified in configuration. Please remove the configuration for either `private_key_path` or `private_key`")
```

erezrokah (changes requested):
Thanks for following up @patriknordlen, added a couple of comments regarding the error messages

patriknordlen:
> Thanks for following up @patriknordlen, added a couple of comments regarding the error messages

Thanks for the suggestions, accepted and committed!

erezrokah (approved):

Issue #13486, opened 2023-08-30
bug: Failed to Write in MySQL Error 1071 (42000)


bbernays:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Run a sync and get the following error:
```
ERR exiting with error error="failed to sync v3 source aws: write client returned error (insert): plugin returned error: failed to write: Error 1071 (42000): Specified key was too long; max key length is 3072 bytes" module=cli
```

Reported on discord here: https://discord.com/channels/872925471417962546/1146540225997320272

### Expected Behavior

No error

### CloudQuery (redacted) config

mysql.yml:
```
kind: destination
spec:
  name:     "mysql"
  registry: "github"
  path:     "cloudquery/mysql"
  version:  "v3.0.4"
  spec:
    connection_string: "user:password@host/cloudquery"
```

aws.yml:
```
kind: source
spec:
  name: "aws"
  path: "cloudquery/aws"
  version: "v22.8.2"
  tables: ["*"]
  destinations: ["mysql"]
`

cloudquery version:
$ cloudquery  --version
cloudquery version 3.16.1


### Steps To Reproduce

```
AWS_PROFILE=some-profile cloudquery sync ./cloudquery-config/
```


### CloudQuery (redacted) logs

none

### CloudQuery version

3.16.1

### Additional Context

_No response_

### Pull request (optional)

- [ ] I can submit a pull request

PR #13557, opened 2023-08-31
fix: Normalize PKs to not null

erezrokah:
#### Summary

Saw this while working on https://github.com/cloudquery/cloudquery/issues/13486. If you run migration twice on the same spec it fails. The reason is that MySQL Implicitly sets PKs to non null in the DB, so we have to account for that.

We had this in the code and it seems it got lost in the migration to v3 or v4.
~~I also cleaned up the tests as those are passing without that loop on test options.~~ Tests cleanup moved to https://github.com/cloudquery/cloudquery/pull/13565

Not sure why the double migration tests didn't catch this


plugins/destination/mysql/client/client_test.go
```diff
@@ -32,27 +30,16 @@ func TestPlugin(t *testing.T) {
...
-				SafeMigrations: plugin.SafeMigrations{
-					AddColumn:    true,
-					RemoveColumn: true,
-				},
+	plugin.TestWriterSuiteRunner(t,
+		p,
+		plugin.WriterTestSuiteTests{
+			SafeMigrations: plugin.SafeMigrations{
+				AddColumn:           true,
+				AddColumnNotNull:    false,
```
candiduslynx, line 38: I think it just needs the true options, the false is the default.

erezrokah:
Tests fixed in https://github.com/cloudquery/cloudquery/pull/13557/commits/a429640d33d67750b90d0ff8f1c630ee4c8fb16f

.github/workflows/dest_mysql.yml
```diff
@@ -24,19 +24,6 @@ jobs:
     defaults:
       run:
         working-directory: ./plugins/destination/mysql
-    services:
```
candiduslynx, line 27: Any specific reason for this?

erezrokah: Maps and Lists don't work without it, see https://github.com/cloudquery/cloudquery/pull/13557/commits/a429640d33d67750b90d0ff8f1c630ee4c8fb16f

candiduslynx: Ah, OK.
Could you add a comment to the workflow describing this? So that we won't try to move back to `services` in this workflow.

plugins/destination/mysql/client/client_test.go
```diff
@@ -32,27 +30,16 @@ func TestPlugin(t *testing.T) {
...
+				RemoveColumn:        true,
+				RemoveColumnNotNull: false,
+				ChangeColumn:        false,
```
candiduslynx, line 41:
```suggestion
				AddColumn:    true,
				RemoveColumn: true,
```

plugins/destination/mysql/client/migrate.go
```diff
@@ -22,6 +22,8 @@ func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
 func (c *Client) normalizeTable(table *schema.Table) *schema.Table {
 	columns := make([]schema.Column, len(table.Columns))
 	for i, col := range table.Columns {
+		// In MySQL primary keys are implicitly not null
+		col.NotNull = col.NotNull || col.PrimaryKey
```
candiduslynx, line 26: I'm a bit worried that we modify column here instead of working on the copy.
Could we move it to
```go
normalizedCol := schema.NewColumnFromArrowField(*normalized)
normalizedCol.NotNull = normalizedCol.NotNull || normalizedCol.PrimaryKey
columns[i] = normalizedCol
```
or something similar?

.github/workflows/dest_mysql.yml
erezrokah: There's a comment there `# We don't use services as we need to pass additional arguments to the container`

plugins/destination/mysql/client/migrate.go
erezrokah: Cleaned up the PR. Also thought about removing the conversion to arrow and back to column, but probably safer to keep it so we remove any non relevant column fields.

Also moved the tests change to https://github.com/cloudquery/cloudquery/pull/13565

plugins/destination/mysql/client/migrate.go
```diff
@@ -22,20 +22,22 @@ func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
...
+	field := col.ToArrowField()
 	normalizedType := mySQLTypeToArrowType(arrowTypeToMySqlStr(field.Type))
-	return &arrow.Field{
+	// In MySQL primary keys are implicitly not null
+	notNull := col.NotNull || col.PrimaryKey
+	return schema.NewColumnFromArrowField(arrow.Field{
 		Name:     field.Name,
 		Type:     normalizedType,
-		Nullable: field.Nullable,
+		Nullable: !notNull,
```
candiduslynx, line 38:
```suggestion
		Nullable: !(col.NotNull || col.PrimaryKey), // In MySQL primary keys are implicitly not null
```

plugins/destination/mysql/client/migrate.go
```diff
@@ -22,20 +22,22 @@ func (c *Client) normalizeTables(tables schema.Tables) schema.Tables {
...
 	return &schema.Table{Name: table.Name, Columns: columns}
 }
 
-func (*Client) normalizeField(field arrow.Field) *arrow.Field {
+func (*Client) normalizeColumn(col schema.Column) schema.Column {
+	field := col.ToArrowField()
 	normalizedType := mySQLTypeToArrowType(arrowTypeToMySqlStr(field.Type))
-	return &arrow.Field{
+	// In MySQL primary keys are implicitly not null
+	notNull := col.NotNull || col.PrimaryKey
```
candiduslynx, line 34:
```suggestion
```

candiduslynx (approved):

PR #13558, opened 2023-08-31
doc: Add multiline text substitution option in the documentation

ukbea:
Multiline JSON substitution using pipe operator.

#### Summary

The environment variable substitution documentation is suggesting that new line characters should be escaped for multiline content. This is an unnecessary effort which complicates the scripts to setup CloudQuery environment. 

Users can rather substitute the environment variable using pipe '|' operator and a tabbed new line before the substitution. This is a standard way of importing a multiline text into YAML files.


hermanschaaf:
Hi @ukbea, thanks for this PR! This is related to the discussion in https://github.com/cloudquery/plugin-sdk/issues/480 and I'll have to revisit the code there to wrap my head around the current state again. 

Before we document this, I'd want to make sure we have a test case for it in the SDK and that it also works when using multi-line values that are not JSON. This is on me, but please expect a bit of delay on this review for that reason 😄

hermanschaaf:
Hi @ukbea, I reviewed this and found that the pipe technique indeed specifically only works for JSON - other types of multiline text fail to parse right now, due to the issues highlighted in https://github.com/cloudquery/plugin-sdk/issues/480. 

In any case, I think this is useful context for users to have, I just updated the PR to make it clear it's specific to multi-line JSON right now.

Thanks again for the contribution, I'm sure it's going to help others down the line!

erezrokah (approved):

hermanschaaf (approved):

PR #13560, opened 2023-08-31
fix: Use Pagination for `aws_ec2_images`

bbernays:
#### Summary

Rather than returning all images in a single API call, this PR changes it to use pagination... It will return 1000 images at a time. This will improve performance for large accounts where it is faster to batch the results rather than return them all at once


hermanschaaf (approved):

hermanschaaf:
Was there an issue attached to this, or just something you noticed?

PR #13565, opened 2023-09-01
test: Clean up MySQL tests

erezrokah:
#### Summary

No need for the loop in the tests or the the `sleep`. Also we can support lists and maps if we increase `innodb_log_file_size`.


mnorbury (approved):
LGTM

plugins/destination/mysql/client/client_test.go
```diff
@@ -32,27 +30,16 @@ func TestPlugin(t *testing.T) {
...
-				},
+	plugin.TestWriterSuiteRunner(t,
+		p,
+		plugin.WriterTestSuiteTests{
+			SafeMigrations: plugin.SafeMigrations{
+				AddColumn:           true,
+				AddColumnNotNull:    false,
+				RemoveColumn:        true,
+				RemoveColumnNotNull: false,
+				ChangeColumn:        false,
```
candiduslynx, line 41:
```suggestion
				AddColumn:    true,
				RemoveColumn: true,
```

PR #13566, opened 2023-09-01
doc(website): Adding bitbucket documentation

mnorbury:


website/pages/docs/plugins/sources/bitbucket/overview.mdx
```diff
@@ -0,0 +1,58 @@
...
+    username: "${BITBUCKET_USERNAME}" # required
+    password: "${BITBUCKET_PASSWORD}" # required
+  ...
+```
+
+## Configuration Reference
+
+This is the (nested) spec used by the Bitbucket source plugin:
+
+- `username` (string, required):
```
candiduslynx, line 52:
```suggestion
- `username` (`string`) (required):
```

website/pages/docs/plugins/sources/bitbucket/overview.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+  The Bitbucket username associated with the [app password](https://support.atlassian.com/bitbucket-cloud/docs/app-passwords/).
+
+- `password` (string, required):
```
candiduslynx, line 56:
```suggestion
- `password` (`string`) (required):
```

website/pages/docs/plugins/sources/bitbucket/_configuration.mdx
```diff
@@ -0,0 +1,20 @@
...
+  tables: ["*"]
+  destinations: ["DESTINATION_NAME"]
+  # bitbucket-specific configuration
+  spec:
+    username: "${BITBUCKET_USERNAME}" # required
+    password: "${BITBUCKET_PASSWORD}" # required
+```
+
+<Callout type="info">
+This example uses [environment variable expansion](/docs/advanced-topics/environment-variable-substitution) to read the token from an `BITBUCKET_USERNAME` and `BITBUCKET_PASSWORD` environment variables. You can also hardcode the value in the configuration file, but this is not advised for production settings.
```
candiduslynx, line 19:
```suggestion
This example uses [environment variable expansion](/docs/advanced-topics/environment-variable-substitution) to read the `username` and `password` options from `BITBUCKET_USERNAME` and `BITBUCKET_PASSWORD` environment variables respectively. You can also hardcode the value in the configuration file, but this is not advised for production settings.
```

plugins/source/bitbucket/testdata/config.yml
```diff
@@ -0,0 +1,15 @@
+kind: source
+spec:
+  name: bitbucket
+  version: v1.0.0
+  destinations: [test]
+  path: "bitbucket:latest"
```
candiduslynx, line 6: is it a correct path?

candiduslynx (changes requested):

plugins/source/bitbucket/testdata/config.yml
mnorbury: It is in this case. This config is used to generated the markdown tables using a locally built docker image called `bitbucket:latest`

candiduslynx (approved):

website/tables/bitbucket/bitbucket_workspaces.md
```diff
@@ -0,0 +1,20 @@
...
+| ------------- | ------------- |
+|_cq_id|`uuid`|
+|_cq_parent_id|`uuid`|
+|uuid (PK)|`uuid`|
+|name|`utf8`|
+|type|`utf8`|
+|slug|`utf8`|
+|is_private|`bool`|
+|created_on|`timestamp[ms, tz=Z]`|
+|updated_on|`timestamp[ms, tz=Z]`|
```
hermanschaaf, line 19: I wonder why this is `timestamp[ms, tz=Z]` instead of `timestamp[ms, tz=UTC]` that we normally get in the docs 🤔

hermanschaaf (approved):

hermanschaaf:
Looks great!

website/tables/bitbucket/bitbucket_workspaces.md
candiduslynx: https://github.com/apache/arrow/blob/4b7d699ab9ee/go/arrow/datatype_fixedwidth.go#L819

hermanschaaf: @candiduslynx That definition seems to have `UTC` in it, why is it `Z` in the generated docs here?

candiduslynx: @hermanschaaf because almost all of our plugins use Go's `Timestamp_us` while in Java SDK we might just skip the timezone?

candiduslynx: Although we use UTC there, too:
https://github.com/cloudquery/plugin-sdk-java/blob/main/lib/src/main/java/io/cloudquery/scalar/Timestamp.java#L16-L17
https://github.com/cloudquery/plugin-sdk-java/blob/main/lib/src/main/java/io/cloudquery/transformers/TypeTransformer.java#L43-L44

erezrokah (approved):
🚀  Amazing work

website/tables/bitbucket/bitbucket_workspaces.md
candiduslynx: `Z` comes from the `ZoneOffset.UTC`, though.
I wonder if it's something of an issue

mnorbury: Running the CLI table conversion through a debugger, I can see the `Z` string from Java makes it all the way though to the data type creation [here](https://github.com/apache/arrow/blob/main/go/arrow/ipc/metadata.go#L947) in the CLI process. The timezone is just a string in Go with no additional validation (although `Z` is a valid timezone).

hermanschaaf: Ah cool, that makes sense now. No worries in that case, I think we can live with the minor discrepancy in the docs. Thanks for looking into it!

PR #11982, opened 2023-07-05
fix(mysql)!: Limit truncation of `text` and `blob` columns used as primary Keys by setting the prefix to the largest value

bbernays:
#### Summary

Closes #11962 
BEGIN_COMMIT_OVERRIDE
fix!: This changes the prefix size for `text` and `blob` primary keys. Previous to this PR only the first 64 characters were included in the primary key, but now it will use maximum supported number of characters (191) in the primary key. Warning logs will be printed if there is any possible truncation. If you are using a source plugin that has a primary key on a `text` or `blob` column you will have to use `migrate_mode: forced` to upgrade.
END_COMMIT_OVERRIDE

PR #13569, opened 2023-09-01
fix!: Split max index length between PKs

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/13486. The max length of an index is either `3072` bytes or `767` bytes.
Since we use `utf8mb4` which can use up to 4 bytes per char, we max out at either `768` or `191` chars (the number in bytes divided by `4`).
The current code allocates `191` chars per PK which is OK if a table has 4 PKs since `191*4= 764`, but fails if a table has more than that. See list of tables [here](https://github.com/cloudquery/cloudquery/issues/13486#issuecomment-1701366092)

The solution is to get the max number of bytes based on the row format, and evenly distribute the max length between PKs.
This is still not the best solution as we assign the same number of bytes to a UUID and a `text` column (for example CQId is `UUID`), but I don't think it's very common to have one UUID PK and another `text` PK and enough other PKs so the prefix is too short.


plugins/destination/mysql/client/schema.go
```diff
@@ -13,20 +13,13 @@ func identifier(name string) string {
...
 }
 
-const maxPrefixLength = 191
-
 const columnQuery = `SELECT 
 cols.COLUMN_NAME,
 COLUMN_TYPE,
 IS_NULLABLE,
-constraint_type,
-sub_part
```
erezrokah, line 23: I couldn't understand why we need `sub_part`. Please comment if you know why

plugins/destination/mysql/client/schema.go
```diff
@@ -72,10 +64,6 @@ func (c *Client) getTableColumns(ctx context.Context, tableName string) ([]schem
 		if constraintType != nil {
 			primaryKey = strings.Contains(*constraintType, "PRIMARY KEY")
 		}
-		// subpart only non nil for pks on blob/text columns
-		if subpart != nil && *subpart != maxPrefixLength {
```
erezrokah, line 76: Not sure why we disable the PK for columns that has a `subpart` but it doesn't match our own number

plugins/destination/mysql/client/schema.go
candiduslynx: https://github.com/cloudquery/cloudquery/pull/11982#discussion_r1255691867

erezrokah: Saw that comment and still don't understand the reason. It is because we want the migration to fail?

candiduslynx: I presume so.
I think @bbernays is the one who can answer this question precisely.

plugins/destination/mysql/client/schema.go
erezrokah: I think this could create issues in a source DB -> dest MySQL sync, for example MySQL -> MySQL

plugins/destination/mysql/client/schema.go
bbernays: Yes, without it migrations wouldn't catch (and then update) the subpart for existing DBs. Not sure if this was the best tradeoff (as I agree that it inhibits the DB->DB) use case... 

But by shortening the subpart, we are risking losing data in cases when the beginning of PKs are the same...

erezrokah: > But by shortening the subpart, we are risking losing data in cases when the beginning of PKs are the same

Yeah, but we won't be able to create the table otherwise as we would exceed the allowed limit to the sync won't work at all. At least with the shorter prefix we'll have the log warning about truncation

erezrokah: Also, please note that with this fix, a single PK will have a longer prefix than `191`

erezrokah: Those won't really help source DB -> MySQL users

bbernays: It will help all but mysql-mysql right? Which in that case we would not want to alter the length of PKs at all

plugins/destination/mysql/client/schema.go
```diff
@@ -172,15 +160,15 @@ func (c *Client) createTable(ctx context.Context, table *schema.Table) error {
 	if len(primaryKeysIndices) > 0 {
 		builder.WriteString(",\n  ")
 		builder.WriteString(" PRIMARY KEY (")
+		lengthPerPk := c.maxIndexLength / len(primaryKeysIndices)
```
bbernays, line 163: A few things:
1. This makes the assumption that all keys in the composite PK have the same length... So in cases where the PK is something like `req_region`, `req_account_id` and `arn` the first 2 columns will be vastly over provisioned as we know an account_id is always 12 characters.
2. Don't we need to factor in the type of each column in the PK? If one column is an int, doesn't that take up a lot less space than an text/blob that requires the subpart?

erezrokah: 1. Yes, but we don't have any way of knowing that in advance unless the source adds a column length in bytes to each column. Please note that currently we are still provisioning the same length for each key which is `191`, so this PR doesn't make it worse, only better.

2. Yes it's not 100% yet, I commented on it in the PR description, see https://github.com/cloudquery/cloudquery/pull/13569#issue-1877070038. Please note this probably only impacts tables with many PKs, so this optimization might not be needed yet.

plugins/destination/mysql/client/schema.go
erezrokah: We could force users to use `pk_mode: cq-id-only` and `deterministic_cq_id`, but I'm not sure that's the best default.
For example in Azure we hardly ever have more than 1 PK (it's always the ID), same with GCP and `Name` and AWS and `arn`, and so on.

Most tables do have a single field that acts as a PK, and in that case we would use the maximum possible length for the key with this PR. Without this fix we always use `191` which is not the current maximum value.

And it not just about mysql-mysql, it's also Postgres->MySQL and OracleDB->MySql, and maybe Firestore->MySQL.

erezrokah: Also, with this fix we won't be shortening the prefix for tables, only making it longer. If the tables exists we would skip it (we don't detect the prefix change) and if it doesn't exist we create it.
It can't exist with a longer prefix, since that's exactly the thing that was preventing it from getting created.

erezrokah: Technically we could detect the prefix change, but I believe that would be quite error prone

bbernays (approved):

erezrokah:
Going to merge and release this next week

PR #13588, opened 2023-09-01
feat: Bulk all messages in a batch into a single bulk write

candiduslynx:
Extracted from https://github.com/cloudquery/cloudquery/pull/13553


hermanschaaf (approved):
👍

PR #13591, opened 2023-09-01
fix!: Upgrade SDK to v4.6.2, introduce `index_name_format` option

hermanschaaf:


candiduslynx:
Could we close it in favor of https://github.com/cloudquery/cloudquery/pull/13600?

hermanschaaf:
Yes, this requires a bit more work. There's also a problem around the use of `time.Now` instead of the actual sync time which makes delete-stale tricky to implement.

Issue #13605, opened 2023-09-04
Add S3 Event Notifications to AWS source plugin


candiduslynx:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/EventNotifications.html

PR #13607, opened 2023-09-04
fix!: Primary keys for `aws_s3_buckets` relations

candiduslynx:
Found during work on https://github.com/cloudquery/cloudquery/issues/13605

### DIFF
--- a/plugins/source/aws/resources/services/s3/bucket_cors_rules.go
+++ b/plugins/source/aws/resources/services/s3/bucket_cors_rules.go
@@ -22,9 +22,10 @@ func bucketCorsRules() *schema.Table {
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
+				Name:       "bucket_arn",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.ParentColumnResolver("arn"),
+				PrimaryKey: true,
 			},
 		},
 	}
--- a/plugins/source/aws/resources/services/s3/bucket_cors_rules.go
+++ b/plugins/source/aws/resources/services/s3/bucket_cors_rules.go
@@ -3,7 +3,6 @@ package s3
 import (
 	"context"
 
-	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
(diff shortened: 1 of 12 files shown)


plugins/source/aws/resources/services/s3/bucket_cors_rules.go
```diff
@@ -22,9 +22,10 @@ func bucketCorsRules() *schema.Table {
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
+				Name:       "bucket_arn",
```
erezrokah, line 25: Doesn't this mean we always save a single rule per bucket in the DB? 

Can we have multiple rules per https://docs.aws.amazon.com/AmazonS3/latest/userguide/ManageCorsUsing.html#cors-example-1?

plugins/source/aws/resources/services/s3/bucket_lifecycles.go
```diff
@@ -22,9 +22,10 @@ func bucketLifecycles() *schema.Table {
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
+				Name:       "bucket_arn",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.ParentColumnResolver("arn"),
+				PrimaryKey: true,
```
erezrokah, line 28: We should verify the API for Bucket lifecycle always returns a single item

plugins/source/aws/resources/services/s3/bucket_websites.go
```diff
@@ -21,9 +21,10 @@ func bucketWebsites() *schema.Table {
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
+				Name:       "bucket_arn",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.ParentColumnResolver("arn"),
+				PrimaryKey: true,
```
erezrokah, line 27: We should verify the API for Bucket Website always returns a single item

plugins/source/aws/resources/services/s3/bucket_lifecycles.go
candiduslynx: Added `ID` to PK in c2825823e5a4d53ed753f7cc26e2867e8e0ae0d1

plugins/source/aws/resources/services/s3/bucket_cors_rules.go
candiduslynx: Added `ID` to PK in 6f45f34aff6ef778f06dd815fcbd81247237c429

erezrokah (approved):

plugins/source/aws/resources/services/s3/bucket_cors_rules.go
```diff
@@ -18,15 +17,8 @@ func bucketCorsRules() *schema.Table {
...
-		Transform:   transformers.TransformWithStruct(&types.CORSRule{}),
-		Columns: []schema.Column{
-			client.DefaultAccountIDColumn(false),
-			{
-				Name:     "bucket_arn",
-				Type:     arrow.BinaryTypes.String,
-				Resolver: schema.ParentColumnResolver("arn"),
-			},
-		},
+		Transform:   transformers.TransformWithStruct(&types.CORSRule{}, transformers.WithPrimaryKeys("ID")),
```
bbernays, line 20: We need to verify that all of these fields are required fields.

I just checked this one and it is not required. In [the Examples](https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-cors.html) for the AWS CLI they don't even populate:

```
{
    "CORSRules": [
        {
            "AllowedHeaders": [
                "*"
            ],
            "ExposeHeaders": [
                "x-amz-server-side-encryption"
            ],
            "AllowedMethods": [
                "PUT",
                "POST",
                "DELETE"
            ],
            "MaxAgeSeconds": 3000,
            "AllowedOrigins": [
                "http://www.example.com"
            ]
        },
        {
            "AllowedHeaders": [
                "Authorization"
            ],
            "MaxAgeSeconds": 3000,
            "AllowedMethods": [
                "GET"
            ],
            "AllowedOrigins": [
                "*"
            ]
        }
    ]
}
```

bbernays (changes requested):

plugins/source/aws/resources/services/s3/bucket_cors_rules.go
candiduslynx: nice catch!
I'll close this PR for now as noone actually asked to fix those

PR #13616, opened 2023-09-04
fix: Fix query for finding unused target groups for AWS

Ramesh7:
#### Summary

The current query doesn't help to find the unused target group as the where clause always compares with integer but the `array_length(load_balancer_arns, 1)` returns NULL if no value exists.
```
cloudquery=# select arn, array_length(load_balancer_arns, 1) as target_group_count from aws_elbv2_target_groups where array_length(load_balancer_arns, 1) is NULL limit 1;
-[ RECORD 1 ]------+------------------------------------------------------------------------------------------------------------------
arn                | arn:aws:elasticloadbalancing:us-west-2:XXXXXXXXXX:targetgroup/a2bfea1561b04450d8afa8af519fffed/b0d1f4eda194f1ab
target_group_count |
```

Fix : 
CASCADE NULL value to 0 so that we always get integer output : 
```
cloudquery=# select arn, COALESCE(array_length(load_balancer_arns, 1), 0) as target_group_count from aws_elbv2_target_groups where COALESCE(array_length(load_balancer_arns, 1), 0) =0 limit 1;
-[ RECORD 1 ]------+------------------------------------------------------------------------------------------------------------------
arn                | arn:aws:elasticloadbalancing:us-west-2:XXXXXXXXX:targetgroup/a2bfea1561b04450d8afa8af519fffed/b0d1f4eda194f1ab
target_group_count | 0

```


hermanschaaf:
/gen commit=b01ac41789b3b5985de54e569796864caea9ec5e plugin=aws

PR #13621, opened 2023-09-05
chore(website): Embed form for register for cloud

hermanschaaf:
Embeds the Typeform to register for CQ Cloud so that users don't need to be redirected to another domain.

Closes https://github.com/cloudquery/cloudquery-issues/issues/589 (internal issue)


erezrokah (approved):

website/components/Typeform.tsx
```diff
@@ -0,0 +1,19 @@
+import { Widget } from '@typeform/embed-react';
+
+export const Typeform = ({
+                             id,
+                         }: {
+    id: string;
+}) => {
```
candiduslynx, line 7: IDK if this weird formatting a norm in JS
```suggestion
export const Typeform = ({id}: {id: string}) => {
```

candiduslynx (approved):

Issue #13433, opened 2023-08-29
bug: aws policy elbv2_redirect_http_to_https.sql should match on lowercase "redirect" 


obormot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

plugins/source/aws/policies/queries/elb/elbv2_redirect_http_to_https.sql:
```
   protocol = 'HTTP' and (
        da->>'Type' != 'REDIRECT' or da->'RedirectConfig'->>'Protocol' != 'HTTPS')
    then 'fail'
```
this code tries to match on uppercase "REDIRECT", while in my case I found the keyword as returned by AWS is lowercase. 
e.g. running `select default_actions from aws_elbv2_listeners limit 1;` gives this:
```
[
  {
    "Type": "redirect",
    "Order": 1,
    "ForwardConfig": null,
    "RedirectConfig": {
      "Host": "#{host}",
      "Path": "/#{path}",
      "Port": "443",
      "Query": "#{query}",
      "Protocol": "HTTPS",
      "StatusCode": "HTTP_301"
    },
    "TargetGroupArn": null,
    "FixedResponseConfig": null,
    "AuthenticateOidcConfig": null,
    "AuthenticateCognitoConfig": null
  }
]
```

### Expected Behavior

the correct code in elbv2_redirect_http_to_https.sql should be
```
   protocol = 'HTTP' and (
        da->>'Type' != 'redirect' or da->'RedirectConfig'->>'Protocol' != 'HTTPS')
    then 'fail'
```
OR
```
   protocol = 'HTTP' and (
        UPPER(da->>'Type') != 'REDIRECT' or da->'RedirectConfig'->>'Protocol' != 'HTTPS')
    then 'fail'
```

### CloudQuery (redacted) config

N/A

### Steps To Reproduce

N/A

### CloudQuery (redacted) logs

N/A

### CloudQuery version

I don't know

### Pull request (optional)

- [X] I can submit a pull request

sabe6:
+1

PR #13751, opened 2023-09-07
fix: Issue #13433 - match on lowercase "redirect" in aws policy 

obormot:
#### Summary

Fix a bug in `plugins/source/aws/policies/queries/elb/elbv2_redirect_http_to_https.sql` that tries to match on uppercase "REDIRECT" while the data in `aws_elbv2_listeners` shows the lowercase "redirect" (see issue https://github.com/cloudquery/cloudquery/issues/13433 for detail).  This bug causes false positive matches in the aws policy.

Fixes https://github.com/cloudquery/cloudquery/issues/13433

### DIFF
--- a/plugins/source/aws/policies/queries/elb/elbv2_redirect_http_to_https.sql
+++ b/plugins/source/aws/policies/queries/elb/elbv2_redirect_http_to_https.sql
@@ -8,7 +8,7 @@ select
   arn as resource_id,
   case when
    protocol = 'HTTP' and (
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
Reference https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listener-action.html#cfn-elasticloadbalancingv2-listener-action-type.

Thanks @obormot 👍

PR #13757, opened 2023-09-07
feat: Add scheduler option to AWS

hermanschaaf:
This wasn't moved into the AWS plugin before because there wasn't a strong need for it, but apparently some APIs (like account contacts) can benefit from this scheduler.

### DIFF
--- a/plugins/source/aws/client/spec.go
+++ b/plugins/source/aws/client/spec.go
@@ -130,7 +130,4 @@ func (s *Spec) SetDefaults() {
 		fullSync := true
 		s.EventBasedSync.FullSync = &fullSync
 	}
-	if string(s.Scheduler) == "" {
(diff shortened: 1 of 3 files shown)


bbernays (approved):

plugins/source/aws/client/spec.go
```diff
@@ -60,6 +61,7 @@ type Spec struct {
 	TableOptions              *tableoptions.TableOptions `json:"table_options,omitempty"`
 	Concurrency               int                        `json:"concurrency"`
 	EventBasedSync            *EventBasedSync            `json:"event_based_sync,omitempty"`
+	Scheduler                 scheduler.Strategy         `json:"scheduler,omitempty"`
```
hermanschaaf, line 64: The default is DFS (it's an enum with zero value) so we don't need to set a default explicitly

PR #13855, opened 2023-09-13
feat: Updated query of security_groups_with_access_to_unauthorized_ports

ronsh12:
A change was made because duplicate rows were inserted into the aws_policy_results table. Additionally, the IP condition should be combined with port conditions using parentheses for proper logical grouping.


plugins/source/aws/policies/queries/ec2/security_groups_with_access_to_unauthorized_ports.sql
```diff
@@ -9,12 +9,14 @@ SELECT
...
     OR to_port IS DISTINCT FROM 80
     OR from_port IS DISTINCT FROM 443
-    OR to_port IS DISTINCT FROM 443
+    OR to_port IS DISTINCT FROM 443)
     then 'fail'
     else 'pass'
-  end
+  end as status
 FROM view_aws_security_group_ingress_rules
+GROUP BY account_id, resource_id, status
```
jsonpr, line 21: What happens here when a Security Group has multiple rules - some that pass and some that fail?  Will the SG show up twice - where there will be multiple rows for the same SG (but each SG will have 2 rows - one for a pass and one for a fail).

jsonpr (changes requested):

plugins/source/aws/policies/queries/ec2/security_groups_with_access_to_unauthorized_ports.sql
```diff
@@ -1,20 +1,34 @@
...
-  end
-FROM view_aws_security_group_ingress_rules
+	:'execution_time' as execution_time,
+	:'framework' as framework,
+	:'check_id' as check_id,
+	'Aggregates rules of security groups with ports and IPs including ipv6' as title,	
+    account_id,
+    resource_id,
+    CASE
+      WHEN SUM(CASE WHEN status = 'fail' THEN 1 ELSE 0 END) > 0 THEN 'fail'
```
jsonpr, line 29: Approving this for now - let's choose a consistent way across our logic to do aggregate checks like this.

jsonpr (approved):

PR #13859, opened 2023-09-13
feat: Updated query unused_acls.sql

ronsh12:
Change query - 'Unused network acl' would mean it does not have any subnet associations. and as result should be fail.


jsonpr (approved):

erezrokah:
Going to force merge as the CI failures are unrelated

PR #13860, opened 2023-09-13
feat: Updated query api_gw_cache_encrypted.sql

ronsh12:
Change query - because opposite condition.


plugins/source/aws/policies/queries/apigateway/api_gw_cache_encrypted.sql
```diff
@@ -8,11 +8,12 @@ select
   arn as resource_id,
   case
     when stage_caching_enabled is true
-        or (
+        and (
```
jsonpr, line 11: Does this check for all methods in a stage?

jsonpr (changes requested):

plugins/source/aws/policies/queries/apigateway/api_gw_cache_encrypted.sql
ronsh12: You're correct; perhaps we should consider reviewing each stage and then grouping by ARN. 
However, for the time being, I think to leave it as is, as potential modifications to the table are being considered (table view).

jsonpr (approved):
Approving for now and we may need to revisit.

Issue #13768, opened 2023-09-08
bug: elasticbeanstalk/advanced_health_reporting_enabled.sql always has not match health_status value


sabe6:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

https://docs.aws.amazon.com/ja_jp/securityhub/latest/userguide/elasticbeanstalk-controls.html#elasticbeanstalk-1

[ElasticBeanstalk.1](https://docs.aws.amazon.com/securityhub/latest/userguide/elasticbeanstalk-controls.html#elasticbeanstalk-1) control.

When I confured enhanced health reporting enabled, health_status became '' empty string, not null.

### Expected Behavior

Patch it works:

```
diff --git a/plugins/source/aws/policies/queries/elasticbeanstalk/advanced_health_reporting_enabled.sql b/plugins/source/aws/policies/queries/elasticbeanstalk/advanced_health_reporting_enabled.sql
index 527020f92..0ffcc98b5 100644
--- a/plugins/source/aws/policies/queries/elasticbeanstalk/advanced_health_reporting_enabled.sql
+++ b/plugins/source/aws/policies/queries/elasticbeanstalk/advanced_health_reporting_enabled.sql
@@ -7,7 +7,7 @@ select
     account_id,
     arn as resource_id,
     case when
-        health_status is null
+        (health_status is null or health_status = '')
         or health is null
         then 'fail'
         else 'pass'

```

### CloudQuery (redacted) config

N/A

### CloudQuery (redacted) logs

N/A

### CloudQuery version

main

### Pull request (optional)

- [ ] I can submit a pull request

PR #13861, opened 2023-09-13
fix: Added check for empty string health_status

john-s58:
following issue #13768  
https://github.com/cloudquery/cloudquery/issues/13768


erezrokah:
@john-s58 You can copy the `sha` from the commits tab of the PR:
image

jsonpr (approved):

Issue #13773, opened 2023-09-08
bug: secretsmanager/remove_unused_secrets_manager_secrets.sql where condition


sabe6:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/policies/queries/secretsmanager/remove_unused_secrets_manager_secrets.sql

[SecretsManager.3](https://docs.aws.amazon.com/securityhub/latest/userguide/secretsmanager-controls.html#secretsmanager-3) control.

The check 'fail' condition would became 'last_accessed_date before 90 days' since last_accessed_date was not updated within 90 days. Current sql would be false positive. 

### Expected Behavior

Patch it works:

```
diff --git a/plugins/source/aws/policies/queries/secretsmanager/remove_unused_secrets_manager_secrets.sql b/plugins/source/aws/policies/queries/secretsmanager/remove_unused_secrets_manager_secrets.sql
index f9ba93ae4..97eb9e4f8 100644
--- a/plugins/source/aws/policies/queries/secretsmanager/remove_unused_secrets_manager_secrets.sql
+++ b/plugins/source/aws/policies/queries/secretsmanager/remove_unused_secrets_manager_secrets.sql
@@ -7,7 +7,7 @@ select
     account_id,
     arn as resource_id,
     case when
-        (last_accessed_date is null and created_date > now() - INTERVAL '90 days')
-        or (last_accessed_date is not null and last_accessed_date > now() - INTERVAL '90 days')
+        (last_accessed_date is null and created_date < now() - INTERVAL '90 days')
+        or (last_accessed_date is not null and last_accessed_date < now() - INTERVAL '90 days')
     then 'fail' else 'pass' end as status
 from aws_secretsmanager_secrets

```

### CloudQuery (redacted) config

N/A

### CloudQuery (redacted) logs

N/A

### CloudQuery version

main

### Pull request (optional)

- [ ] I can submit a pull request

PR #13862, opened 2023-09-13
fix: Flipped condition on secrets manager policy

john-s58:
for issue #13773


erezrokah:
@john-s58 You can copy the `sha` from the commits tab of the PR:
image

jsonpr (approved):

PR #13876, opened 2023-09-16
feat(main): Add JSON schema for spec

yevgenypats:
Blocked by: https://github.com/cloudquery/plugin-sdk/pull/1214


disq (approved):

plugins/source/gcp/client/spec/spec.go
```diff
@@ -1,31 +1,62 @@
...
-	OrganizationIDs             []string           `json:"organization_ids"`
-	OrganizationFilter          string             `json:"organization_filter"`
-	ServiceAccountImpersonation *CredentialsConfig `json:"service_account_impersonation"`
 	Concurrency                 int                `json:"concurrency"`
 	Scheduler                   scheduler.Strategy `json:"scheduler,omitempty"`
+	ServiceAccountImpersonation *CredentialsConfig `json:"service_account_impersonation"`
+}
+
+func (spec *Spec) SetDefaults() {
+	if spec.BackoffRetries <= 0 {
```
bbernays, line 29: If a user wanted `0` retries how would they do that?

candiduslynx: Addressed in d426712e9bbb647b13eadb92186f57f17f0e6039

hermanschaaf:
What do we expect to happen if we merge this PR? Will specs then be validated using JSONSchema, returning an error if validation fails?

candiduslynx:
> What do we expect to happen if we merge this PR? Will specs then be validated using JSONSchema, returning an error if validation fails?

Yes, exactly

disq (approved):

Issue #13937, opened 2023-09-18
feat: Support downloading required plugins separately


disq:
### Which problem is this feature request solving?

CLI should be able to download plugin binaries from a given config separately, without requiring authentication and proceeding further. This would help to prepare self-contained Docker images where the `cli` and required plugins are ready to go.

### Describe the solution you'd like

Just like the `migrate` command, we would have a `download` command where a given config is processed, `cq-dir` is evaluated and missing binaries are downloaded.

### Pull request (optional)

- [ ] I can submit a pull request

pilvikala:
@disq could we make this `cloudquery install` command instead of `download`? We discussed with @hermanschaaf today that this would be less confusing.

PR #13941, opened 2023-09-19
feat: Add `install` command to pre-install required plugins

disq:
Fixes https://github.com/cloudquery/cloudquery/issues/13937
Requires https://github.com/cloudquery/plugin-pb-go/pull/103 (merged/tagged)


cli/cmd/root.go
```diff
@@ -127,6 +127,7 @@ func NewCmdRoot() *cobra.Command {
 		NewCmdSync(),
 		NewCmdMigrate(),
 		newCmdDoc(),
+		newCmdDownload(),
```
disq, line 130: Export or not export?

cli/go.mod
```diff
@@ -68,3 +68,5 @@ require (
 )
 
 replace github.com/apache/arrow/go/v14 => github.com/cloudquery/arrow/go/v14 v14.0.0-20230916001126-ffb7089245ac
+
+replace github.com/cloudquery/plugin-pb-go => github.com/cloudquery/plugin-pb-go v1.9.5-0.20230919081227-761ada95bcf0
```
disq, line 72: Using untagged version for now

hermanschaaf (changes requested):
This looks good 🚀  Just missing a test for the new command :)

disq:
> This looks good 🚀 Just missing a test for the new command :)

Added in [`980ac31` (#13941)](https://github.com/cloudquery/cloudquery/pull/13941/commits/980ac31eb2c6de72e8030553d87fcdf0b2a8be23)

cli/cmd/install_test.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		{
+			name:   "multiple_destinations",
+			config: "multiple-destinations.yml",
+			wantFiles: []string{
+				"cloudquery.log",
+				"plugins/destination/cloudquery/test/v2.1.0/plugin",
+				"plugins/destination/cloudquery/test/v2.1.0/plugin.zip",
+				"plugins/destination/cloudquery/test/v2.2.7/plugin",
+				"plugins/destination/cloudquery/test/v2.2.7/plugin.zip",
```
hermanschaaf, line 51: I believe these versions get updated automatically by bots (https://github.com/cloudquery/cloudquery/blob/main/cli/cmd/testdata/multiple-destinations.yml#L19) so we will need to read them from the config or keep them in sync some other way for those automated PRs to succeed

hermanschaaf (changes requested):
Looks good, just one comment about testing

cli/cmd/install_test.go
erezrokah: This is the regex we use https://github.com/cloudquery/.github/blob/c43e6ce1c7e117aa0da14aff24ec15df002fe0a6/.github/renovate-default.json5#L87 so if you create a new spec without the comment it won't automatically update the versions

erezrokah: For example https://github.com/cloudquery/cloudquery/blob/54cecdc91bb1af0501ac2ff4f65b4a4a448305bc/cli/cmd/testdata/multiple-destinations.yml#L6 pins the source version

erezrokah (approved):

PR #13943, opened 2023-09-19
fix: Updated query dynamodb/autoscale_or_ondemand.sql

ronsh12:
Change the query to be more organized and only include the relevant rows.

### DIFF
--- a/plugins/source/aws/policies/queries/dynamodb/autoscale_or_ondemand.sql
+++ b/plugins/source/aws/policies/queries/dynamodb/autoscale_or_ondemand.sql
@@ -1,28 +1,78 @@
-insert into aws_policy_results
-select
-    :'execution_time' as execution_time,
-    :'framework' as framework,
-    :'check_id' as check_id,
-    'DynamoDB tables should automatically scale capacity with demand' as title,
+WITH dynamodb_tables AS (
+    SELECT
+        account_id,
+        arn,
+        table_name,
+        billing_mode_summary->>'BillingMode' IS DISTINCT FROM 'PAY_PER_REQUEST' AS is_not_on_demand
+    FROM aws_dynamodb_tables
+),
+replica_auto_scalings AS (
+    SELECT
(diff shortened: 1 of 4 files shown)


jsonpr (changes requested):
From reading this control https://docs.aws.amazon.com/securityhub/latest/userguide/dynamodb-controls.html#dynamodb-1, our logic shouldn't check for replicas and their status.  This check should cover `whether an Amazon DynamoDB table can scale its read and write capacity as needed.` and specifically `passes if the table uses either on-demand capacity mode or provisioned mode with auto scaling configured`

@ronsh12  - could you check on this, I believe we'll need additional information not contained in here: https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_dynamodb_tables.  Specifically information about the table's billing mode (Provisioned or On-Demand) and then linking that to autoscaling (https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_applicationautoscaling_policies).

ronsh12:
> From reading this control https://docs.aws.amazon.com/securityhub/latest/userguide/dynamodb-controls.html#dynamodb-1, our logic shouldn't check for replicas and their status. This check should cover `whether an Amazon DynamoDB table can scale its read and write capacity as needed.` and specifically `passes if the table uses either on-demand capacity mode or provisioned mode with auto scaling configured`
> 
> @ronsh12 - could you check on this, I believe we'll need additional information not contained in here: https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_dynamodb_tables. Specifically information about the table's billing mode (Provisioned or On-Demand) and then linking that to autoscaling (https://www.cloudquery.io/docs/plugins/sources/aws/tables/aws_applicationautoscaling_policies).

Changed this.

ronsh12:
in DBT changed too - 
https://github.com/cloudquery/policies-premium/pull/178

erezrokah:
Closing as stale, please re-open if not the case

PR #13953, opened 2023-09-20
chore: Add publish to Hub CI workflow

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloud/issues/303 (internal issue).

### DIFF
--- a/.github/workflows/publish_plugin_to_hub.yml
+++ b/.github/workflows/publish_plugin_to_hub.yml
@@ -1,0 +1,77 @@
+name: Publish plugin to hub
+on:
+  push:
+    tags:
+      - "plugins-*-*-v*.*.*"
+jobs:
+  publish-plugin-to-hub:
+    timeout-minutes: 60
+    runs-on: large-ubuntu-monorepo
+    steps:
+      - name: Split tag
+        id: split
+        run: |
+          tag=${{ github.ref_name }}
+          plugin_type=$(echo $tag | cut -d- -f2)
+          plugin_name=$(echo $tag | cut -d- -f3)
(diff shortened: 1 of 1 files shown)


erezrokah:
Waiting for https://github.com/cloudquery/cloudquery/pull/13880 to be released

.github/workflows/publish_plugin_to_hub.yml
```diff
@@ -0,0 +1,76 @@
...
+              owner: "cloudquery",
+              repo: context.repo.repo,
+              tag: context.ref.replace('refs/tags/', ''),
+            });
+            return data.body;
+
+      - name: Run publish package command
+        working-directory: ${{ steps.split.outputs.plugin_dir }}
+        run: |
+          go run main.go package -m "${{ steps.release-notes.outputs.result }}" . ${{ steps.split.outputs.plugin_version }}
```
hermanschaaf, line 65: Right now this will only work for Go plugins, are other plugins being excluded somehow?

erezrokah: Ah right, I'll update it to ignore other plugins

erezrokah: Fixed and based on the logic from https://github.com/cloudquery/cloudquery/blob/7d3aaa8950383cbaf5de5a016c9cd594bbbbc054/.github/workflows/release_plugin.yml#L48

I intentionally duplicated the file so not to interfere with existing publishing flow. We can merge them later on if needed

hermanschaaf (approved):
Nice 🚀

PR #13962, opened 2023-09-20
fix: Stub docs when publishing

erezrokah:
#### Summary

Stub plugin docs during publish to the Hub. Some plugins don't have docs and others that do have MDX docs mostly, so this stubs the docs for starts until we figure out how to handle MDX docs.


plugins/source/test/README.md
```diff
@@ -1,3 +1,3 @@
 # CloudQuery source test plugin
 
-This source plugin is used in various testing scenarios mainly in our CLI code, such as downloading, running and so on.
\ No newline at end of file
+This source plugin is used in various testing scenarios mainly in our CLI code, such as downloading, running and so on.
```
erezrokah, line 3: This change is here so we trigger a new test plugin release

hermanschaaf (approved):

PR #13975, opened 2023-09-20
feat: Move careers to ATS

yevgenypats:
We are moving to ATS - https://apply.workable.com/cloudquery/ as we are integrating our jobs with other systems as well and upgrading/automating some of our recruting process


website/vercel.json
```diff
@@ -344,7 +344,15 @@
     },
     {
       "source": "/apply_for_job",
-      "destination": "https://cloudquery.typeform.com/to/flNY3OaO"
+      "destination": "https://apply.workable.com/cloudquery/"
+    },
+    {
+      "source": "/careers",
```
hermanschaaf, line 350: In order to not have our internal links point directly to a redirect (which is bad for SEO), maybe we can put up a page at `/careers` similar to the buy pages https://github.com/cloudquery/cloudquery/blob/main/website/pages/buy/slack-standard.mdx. The alternative would be to update all our links from `/careers` to `https://apply.workable.com/cloudquery/`

yevgenypats: Im not sure how critical this for SEO (at least right now - one month ago we had it in notion anyway and I changed it because I was mostly playing with typeform but now it's part for the workable solution) - we can also add our custom domain later for workable as that might help as well. we can go even further and do our custom careers page that will use the API at workable to get the description but I think we are not there yet.

hermanschaaf: I could be wrong, but I think it hurts our overall domain score if we link to many internal redirects (that's the idea I got from SEO tools at least). But in any case, we should do something to at least not have the broken links checker fail, even if that is to change every link to `/careers` to the new location manually

yevgenypats: I believe I did it - it is just failing because it's preview. Isn't that https://github.com/cloudquery/cloudquery/pull/13975/files#diff-21378a71578c160a16aa86a2ccc6120c09d1d10a3deaf16af651a228bd0cfde4R350 redirect doing that?

hermanschaaf: Probably easiest to just ignore it for now by adding a line here: https://github.com/cloudquery/cloudquery/blob/main/.github/workflows/broken_links.yml#L76

erezrokah (approved):

PR #13993, opened 2023-09-22
chore(ci): Pass plugin kind to package command

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/1238


hermanschaaf (approved):
👍

PR #13997, opened 2023-09-22
feat: Support plugin kind in CLI publish command

erezrokah:
#### Summary

Replaces https://github.com/cloudquery/cloudquery/pull/13996. Supports plugin kind in the CLI when publishing plugins


cli/cmd/publish.go
```diff
@@ -18,7 +18,6 @@ import (
 
 	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
 	"github.com/cloudquery/cloudquery/cli/internal/auth"
-	"github.com/cloudquery/plugin-sdk/v4/plugin"
```
erezrokah, line 21: The CLI should not reference the SDK, only https://github.com/cloudquery/plugin-pb-go

cli/cmd/publish.go
```diff
@@ -69,13 +68,16 @@ type PackageJSONVersion struct {
...
-	Name             string             `json:"name"`
-	Message          string             `json:"message"`
-	Version          string             `json:"version"`
-	Protocols        []int              `json:"protocols"`
-	SupportedTargets []TargetBuild      `json:"supported_targets"`
-	PackageType      plugin.PackageType `json:"package_type"`
+	Name             string                    `json:"name"`
+	Message          string                    `json:"message"`
+	Version          string                    `json:"version"`
+	Kind             cloudquery_api.PluginKind `json:"kind"`
```
erezrokah, line 77: I could also make it a string or any other wrapper type and convert, but this works as we keep the enums consistent

cli/cmd/publish.go
erezrokah: Actually I see it's used for the docs command too so we might want to refactor it at some point. The API surface the CLI should interact with is https://github.com/cloudquery/plugin-pb-go

hermanschaaf: Yes, we should refactor this at some point. It will mean copying quite a bit of code from the SDK into plugin-pb-go I think, e.g. all the table struct definitions and how to convert from Arrow to that.

hermanschaaf (approved):

Issue #14030, opened 2023-09-25
feat: Add support for AWS VPC Endpoint Connections


hermanschaaf:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcEndpointConnections.html

PR #14044, opened 2023-09-26
feat: Add support for AWS EC2 VPC Endpoint Connections

bbernays:
#### Summary

closes: #14030


mnorbury (commented):
Looks good, just one question.

Is a VPC endpoint connection a child resource of a VPC endpoint (or an endpoint service)?

bbernays:
> Looks good, just one question.
> 
> Is a VPC endpoint connection a child resource of a VPC endpoint (or an endpoint service)?

I don't think we should consider it to be, because the API doesn't require any inputs. It is always better to have a top level table for resources because they can be better parallelized

mnorbury (approved):

PR #14054, opened 2023-09-27
fix: Slugify docs names before publishing

erezrokah:
#### Summary

See https://github.com/cloudquery/cloudquery/actions/runs/6323457020/job/17171062956#step:8:10

We started enforcing this recently


cli/cmd/publish_test.go
```diff
@@ -265,6 +266,12 @@ func checkCreateDocsRequest(t *testing.T, r *http.Request) {
 				"title":            "Configuration",
 				"ordinal_position": float64(2),
 			},
+			map[string]any{
+				"content":          customDocContent,
+				"name":             "custom-doc",
```
hermanschaaf, line 271: Is this slug expected to come from the name of the doc, or from the slug in the frontmatter? Maybe we should make the `slug` in the frontmatter differ to make this clear?

erezrokah: Updated the tests a bit, good point

hermanschaaf: Great, thanks :) I think the `slug` field in the frontmatter is not being used, should we remove it?

hermanschaaf (approved):

PR #14060, opened 2023-09-27
chore: Add cross compile Hub publishing workflows

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloud/issues/394 (internal issue).

I didn't add `GO111MODULE=on` as that's the default in newer Go versions

### DIFF
--- a/.github/workflows/publish_plugin_to_hub_duckdb.yml
+++ b/.github/workflows/publish_plugin_to_hub_duckdb.yml
@@ -57,7 +57,6 @@ jobs:
       image: ghcr.io/cloudquery/golang-cross:v10.0.0
       env:
         CGO_ENABLED: 1
-        GO111MODULE: on
         CC: /usr/bin/gencc.sh
         CXX: /usr/bin/gencpp.sh
     needs: prepare
--- a/.github/workflows/publish_plugin_to_hub_duckdb.yml
+++ b/.github/workflows/publish_plugin_to_hub_duckdb.yml
@@ -1,11 +1,8 @@
-name: Publish plugin to hub
+name: Publish plugin to DuckDB to hub
 on:
   push:
     tags:
-      - "plugins-*-*-v*.*.*"
-      - "!plugins-destination-duckdb-v*.*.*"
-      - "!plugins-destination-sqlite-v*.*.*"
-      - "!plugins-destination-snowflake-v*.*.*"
+      - "plugins-destination-duckdb-v*.*.*"
 jobs:
   prepare:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/publish_plugin_to_hub_duckdb.yml
```diff
@@ -0,0 +1,114 @@
...
+              return 'go';
+            }
+          result-encoding: string
+  publish-plugin-to-hub:
+    timeout-minutes: 60
+    runs-on: large-ubuntu-monorepo
+    container:
+      image: ghcr.io/cloudquery/golang-cross:v10.0.0
+      env:
+        CGO_ENABLED: 1
```
erezrokah, line 59: Based on https://github.com/cloudquery/cloudquery/blob/54fc73e3b0993ee132d0750a1b24084d2c775638/plugins/destination/duckdb/.goreleaser.yaml#L20

.github/workflows/publish_plugin_to_hub_sqlite.yml
```diff
@@ -0,0 +1,114 @@
...
+              return 'go';
+            }
+          result-encoding: string
+  publish-plugin-to-hub:
+    timeout-minutes: 60
+    runs-on: large-ubuntu-monorepo
+    container:
+      image: ghcr.io/cloudquery/golang-cross:v10.0.0
+      env:
+        CGO_ENABLED: 1
```
erezrokah, line 59: Based on https://github.com/cloudquery/cloudquery/blob/54fc73e3b0993ee132d0750a1b24084d2c775638/plugins/destination/sqlite/.goreleaser.yaml#L20

.github/workflows/publish_plugin_to_hub_snowflake.yml
```diff
@@ -0,0 +1,115 @@
...
+            if (pluginFiles.includes('.goreleaser.yaml')) {
+              return 'go';
+            }
+          result-encoding: string
+  publish-plugin-to-hub:
+    timeout-minutes: 60
+    runs-on: large-ubuntu-monorepo
+    container:
+      image: ghcr.io/cloudquery/golang-cross:v10.0.0
+      env:
```
erezrokah, line 58: A mix of https://github.com/cloudquery/cloudquery/blob/54fc73e3b0993ee132d0750a1b24084d2c775638/plugins/destination/snowflake/.goreleaser.yaml#L20 and https://github.com/cloudquery/cloudquery/blob/54fc73e3b0993ee132d0750a1b24084d2c775638/plugins/destination/snowflake/.goreleaser.yaml#L43

Not sure if this works. If it doesn't we'll need to support different configurations per target

hermanschaaf (approved):

Issue #14105, opened 2023-09-27
bug: Snowflake SQL errors on Windows


mnorbury:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A [user](https://discord.com/channels/872925471417962546/873606591335759872/1156605427946111046) on discord reported the following SQL error message when syncing from OnFleet -> Snowflake.

```
In attempting to sync to Snowflake, using snowflake(v3.2.3), I receive an error, but a sync to SQLite, using sqlite (v2.4.9) is successful.
The Snowflake sync error in the log is:

ERR failed to write batch error="failed to put file into stage with last resource put file://C:\Users\ED29FE~1.WIN\AppData\Local\Temp\onfleet_admins.json.1159616259 @cq_plugin_stage auto_compress=true: 001003 (42000): SQL compilation error:
syntax error line 1 at position 26 unexpected '~'.
parse error line 1 at position 33 near '65'.
parse error line 1 at position 41 near '76'.
parse error line 1 at position 47 near '84'.
parse error line 1 at position 52 near '111'.
syntax error line 1 at position 66 unexpected '.'.
syntax error line 1 at position 71 unexpected '.1159616259'.
syntax error line 1 at position 83 unexpected '@cq_plugin_stage'." duration=68.2582 len=2 module=snowflake-src table=onfleet_admins
```

The Snowflake [put](https://docs.snowflake.com/en/sql-reference/sql/put) documentation suggests that file URIs need to be enclosed in single quotes and the separator be a forward slash (/).

> Windows
You must include the drive and backslash in the path (e.g. file://C:\temp\load).
> If the directory path and/or filename includes special characters, the entire file URI must be enclosed in single quotes. Note that the separator character is a forward slash (/) in enclosed URIs (e.g. 'file://C:/temp/load data' for a path in Windows that includes a directory named load data).

The SQL statement [here](https://github.com/cloudquery/cloudquery/blob/678380691ca6afe7b40175e559d7160f7561d925/plugins/destination/snowflake/client/write.go#L18) needs to be modified to satisfy the above caveat on Windows.

### Expected Behavior

No SQL error is encountered when syncing to Snowflake

### CloudQuery (redacted) config

N/A

### CloudQuery (redacted) logs

N/A

### CloudQuery version

n/a

### Pull request (optional)

- [ ] I can submit a pull request

PR #14107, opened 2023-09-27
fix: Properly handle file paths for Windows

bbernays:
#### Summary

closes: #14105 

Tested that this works properly on both MacOS and Windows


plugins/destination/snowflake/client/write.go
```diff
@@ -54,11 +55,13 @@ func (c *Client) WriteTableBatch(ctx context.Context, name string, msgs message.
 	if err := f.Close(); err != nil {
 		return fmt.Errorf("failed to close temp file with last resource %s: %w", f.Name(), err)
 	}
-	sql := fmt.Sprintf(putFileIntoStage, f.Name())
+
+	sql := fmt.Sprintf(putFileIntoStage, strings.ReplaceAll(f.Name(), "\\", "\\\\"))
```
disq, line 59: Could use `strconv.Quote` or refactor into its own little fn...

disq (approved):

plugins/destination/snowflake/client/write.go
```diff
@@ -54,13 +55,20 @@ func (c *Client) WriteTableBatch(ctx context.Context, name string, msgs message.
...
+	sql = fmt.Sprintf(copyIntoTable, tableName, escapePath(path.Base(f.Name())))
 	if _, err := c.db.ExecContext(ctx, sql); err != nil {
 		return fmt.Errorf("failed to copy file into table with last resource %s: %w", sql, err)
 	}
 	return err
 }
+
+// escapePath properly escapes the `\` character in window's file paths.
+func escapePath(p string) string {
+	return strings.ReplaceAll(p, "\\", "\\\\")
```
mnorbury, line 73: Is escaping the backslash the same as converting to a forward slash - which the docs recommended?

bbernays: The solution actually comes from the Snowflake driver docs:
https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Support_For_PUT_and_GET

plugins/destination/snowflake/client/write.go
```diff
@@ -54,13 +55,20 @@ func (c *Client) WriteTableBatch(ctx context.Context, name string, msgs message.
...
-	sql = fmt.Sprintf(copyIntoTable, tableName, path.Base(f.Name()))
+	sql = fmt.Sprintf(copyIntoTable, tableName, escapePath(path.Base(f.Name())))
 	if _, err := c.db.ExecContext(ctx, sql); err != nil {
 		return fmt.Errorf("failed to copy file into table with last resource %s: %w", sql, err)
 	}
 	return err
 }
+
+// escapePath properly escapes the `\` character in window's file paths.
+func escapePath(p string) string {
```
candiduslynx, line 72: why not use https://pkg.go.dev/path/filepath package & funcs from it?

candiduslynx: Specifically, they do handle this: https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/path/filepath/path_windows.go

bbernays: What function do you want me to use?

candiduslynx: Oh, I missed that it is escape, not unescape

PR #14135, opened 2023-09-28
chore: Fix module path for GCP plugin

candiduslynx:


erezrokah (commented):
Nice catch, I think we will need to update it in the plugin imports too

Issue #14133, opened 2023-09-28
ci(stripe): The CI workflow for the Stripe plugin is failing


erezrokah:
See https://github.com/cloudquery/cloudquery/actions/runs/6336725121/job/17210420493?pr=13930
On https://github.com/cloudquery/cloudquery/pull/13930:
![image](https://github.com/cloudquery/cloudquery/assets/26760571/261269bc-679c-4eee-88e1-c1e7198a3e6e)

We should look into it

PR #14163, opened 2023-09-28
test: Use matching version of stripe API spec for tests

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/14133

The issue is that we use the spec definitions from `master` and generate the fixtures based on the current SDK version in `go.mod` so those drift apart as time passes.

The solution is to use the SDK tag from `go.mod` to read the spec version from https://github.com/stripe/stripe-go/blob/master/OPENAPI_VERSION.

Then once we have the spec version we can download the relevant one from https://github.com/stripe/openapi/tags


plugins/source/stripe/Makefile
```diff
@@ -5,8 +5,10 @@ build:
 # Test unit
 .PHONY: test
 test: gen-fixtures
-	wget -nc https://raw.githubusercontent.com/stripe/openapi/master/openapi/spec3.json -P ./resources/testdata
-	go test -race -timeout 3m ./...
+	stripeSDKTag=$$(go list -m github.com/stripe/stripe-go/v74 | awk '{print $$2}') ;\
+	stripeOpenAPITag=$$(curl --silent https://raw.githubusercontent.com/stripe/stripe-go/$$stripeSDKTag/OPENAPI_VERSION) ;\
+	curl --silent https://raw.githubusercontent.com/stripe/openapi/$$stripeOpenAPITag/openapi/spec3.json > ./resources/testdata/spec3.json ;\
```
disq, line 10: `wget` being replaced by `curl` here...

disq (approved):

plugins/source/stripe/Makefile
disq: `wget -nc` is "no-clobber" so it won't overwrite if the file already exists... which makes the tests go faster

erezrokah: Used `curl` for consistency. We can use `wget` in both. I'd rather overwrite to avoid having a stale file, for example after updating the SDK. Also in CI it doesn't matter

disq: Stale file shouldn't happen as it's in `.gitignore`, but one could upgrade the module and then this would create issues... safe to use `curl` unless we put the sdk tag in the filename I guess.

erezrokah: > but one could upgrade the module and then this would create issues

Yeah, that's what I meant

Issue #14187, opened 2023-09-29
bug: Missing users from Datadog source sync


mnorbury:
A user reported that when using the Datadog source plugin they were only seeing 10 users in their organisation, even though they have many more.

The `datadog-api-client-go` project have recently added pagination support for users in this [PR](https://github.com/DataDog/datadog-api-client-go/pull/2166) (v2.17.0)

Prior to this it is likely that pagination wasn't fully supported in this endpoint.

PR #14188, opened 2023-09-29
fix!: Add pagination support for list users

mnorbury:
fixes: https://github.com/cloudquery/cloudquery/issues/14187


erezrokah (approved):

disq (approved):

erezrokah:
I'm assuming the column type change is due to the SDK update. We should update the title of this PR to `fix!: Add pagination support for list users` so it creates a new major version

erezrokah (approved):

PR #14192, opened 2023-09-29
feat: Enable Users to Disable TLS Verification when using Custom Endpoint  

bbernays:


erezrokah (approved):

plugins/destination/s3/client/spec.go
```diff
@@ -40,6 +40,10 @@ func (s *Spec) SetDefaults() {
 		b := true
 		s.TestWrite = &b
 	}
+	if s.EndpointSkipTLSVerify == nil {
+		b := true
+		s.EndpointSkipTLSVerify = &b
```
disq, line 45: Setting skip to true means it would be skipped by default? Is that what we really want?

bbernays: Definitely not

plugins/destination/s3/client/spec.go
```diff
@@ -19,12 +19,12 @@ type Spec struct {
...
 
-	Endpoint     string `json:"endpoint,omitempty"`
-	UsePathStyle bool   `json:"use_path_style,omitempty"`
-
-	BatchSize      *int64               `json:"batch_size"`
-	BatchSizeBytes *int64               `json:"batch_size_bytes"`
-	BatchTimeout   *configtype.Duration `json:"batch_timeout"`
+	Endpoint              string               `json:"endpoint,omitempty"`
+	UsePathStyle          bool                 `json:"use_path_style,omitempty"`
+	EndpointSkipTLSVerify *bool                `json:"endpoint_skip_tls_verify,omitempty"`
```
disq, line 24: Since the default is `false` there's no reason this to be a ptr. `bool` zero value conforms to our default.

disq (approved):
I think we also want documentation update?

disq (approved):

Issue #14265, opened 2023-10-02
feat: Add support for configurable site


mnorbury:
Datadog uses different regional servers or [sites](https://docs.datadoghq.com/getting_started/site/). To talk to a non-default site (e.g. `eu`) you need to configure Go request context - see [here](https://docs.datadoghq.com/getting_started/site/)

PR #14266, opened 2023-10-02
feat: Add configurable site

mnorbury:
Add configurable site to allow other DDog regional sites to be used, if no site is specified the existing behaviour is preserved.

fixes: https://github.com/cloudquery/cloudquery/issues/14265

### DIFF
--- a/website/pages/docs/plugins/sources/datadog/configuration.mdx
+++ b/website/pages/docs/plugins/sources/datadog/configuration.mdx
@@ -16,6 +16,10 @@ This is the (nested) spec used by the Datadog source plugin.
 
   Specify which accounts to sync data from.
 
+- `site` (`string`, optional, default: "")
+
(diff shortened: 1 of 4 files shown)


website/pages/docs/plugins/sources/datadog/configuration.mdx
```diff
@@ -16,6 +16,10 @@ This is the (nested) spec used by the Datadog source plugin.
 
   Specify which accounts to sync data from.
 
+- `site` (`string`, optional, default: "")
```
erezrokah, line 19: Is setting an empty string (using the default) the same as setting it to `datadoghq.com`? Maybe we should clarify that if that's the case

erezrokah (approved):

erezrokah:
Looks good, one non blocking comment

website/pages/docs/plugins/sources/datadog/configuration.mdx
mnorbury: Yes, it falls back on [this](https://github.com/DataDog/datadog-api-client-go/blob/master/api/datadog/configuration.go#L126) configuration if there is not `site` override in the context.

Issue #13865, opened 2023-09-14
bug: Shopify Fails to Retry on `http2: server sent GOAWAY and closed the connection` error


bbernays:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Table resolver will fail with the following error:
```
ERR table resolver finished with error error="http2: server sent GOAWAY and closed the connection; LastStreamID=19999, ErrCode=NO_ERROR, debug=\"\"" client=shopify
```

### Expected Behavior

Resolver should retry this request

### CloudQuery (redacted) config

n/a

### CloudQuery (redacted) logs

```
2023-09-14T01:01:49Z ERR table resolver finished with error error="http2: server sent GOAWAY and closed the connection; LastStreamID=19999, ErrCode=NO_ERROR, debug=\"\"" client=shopify module=shopify-src table=shopify_orders
```

### CloudQuery version

n/a

### Additional Context

This seems to be a known issue with AWS load balancers:
https://github.com/golang/go/issues/18639

### Pull request (optional)

- [ ] I can submit a pull request

PR #14273, opened 2023-10-02
fix: Retry on http2 GOAWAY error

disq:
Fixes https://github.com/cloudquery/cloudquery/issues/13865

### DIFF
--- a/plugins/source/shopify/internal/shopify/shopify.go
+++ b/plugins/source/shopify/internal/shopify/shopify.go
@@ -97,7 +97,7 @@ func (s *Client) request(ctx context.Context, edge string, params url.Values) (r
 		}
 
 		temporary := false
-		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) || isErrH2GoAway(err) {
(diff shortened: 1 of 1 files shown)


erezrokah (approved):

plugins/source/shopify/internal/shopify/shopify.go
```diff
@@ -361,3 +361,7 @@ func getNextPage(hdr http.Header) string {
 
 	return ""
 }
+
+func isErrH2GoAway(err error) bool {
```
candiduslynx, line 365: why not check for `errors.Is(http2.GoAwayError)`?
https://pkg.go.dev/golang.org/x/net/http2#GoAwayError

disq: @candiduslynx `http2.GoAwayError` is a type and not a variable, that `errors.Is` won't catch it. [`dfe06d9` (#14273)](https://github.com/cloudquery/cloudquery/pull/14273/commits/dfe06d95b8fe6c1aee925ce64445886d8f8b02cf) uses `errors.As` now, but it isn't as pretty.

bbernays (approved):

romerocw:
Based on your fix I restarted the sync on one of my smaller stores. I'll keep you posted on the progress.

PR #14295, opened 2023-10-04
doc: Replace stripe redirects with links to typeform

pilvikala:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/596

Removed stripe redirects from premium policies in early access and replaced them with links to a signup form


erezrokah:
Tagged @hermanschaaf for the review as I think he has the most context on how these were generated

hermanschaaf:
There's a prebuild script that updates these that we should update instead: https://github.com/cloudquery/cloudquery/blob/main/website/scripts/prebuild.tsx#L17-L35

PR #14381, opened 2023-10-05
fix: Redact Code Location and Fall back to other APIs to fully resolve `aws_lambda_functions`

bbernays:
#### Summary

Redact Code Location and Fall back to other APIs to fully resolve `aws_lambda_functions`


plugins/source/aws/resources/services/lambda/functions.go
```diff
@@ -120,7 +121,9 @@ func getFunction(ctx context.Context, meta schema.ClientMeta, resource *schema.R
 
 		return err
 	}
-
+	if funcResponse.Code != nil {
+		funcResponse.Code.Location = aws.String("REDACTED_FOR_SECURITY_PURPOSES")
```
candiduslynx, line 125: This should be:
1. A constant defined somewhere outside of `lambda` dir, maybe in `client` pkg
2. Be documented for AWS source plugin (something like `sensitive information like source code download URLs are masked`)
3. Possibly, we want to leave the host + protocol path & update/remove only the key/presigned part

bbernays: It points to hosts that are owned and controlled by AWS so the values don't matter

bbernays: How does having the string value be a constant impact anything?

candiduslynx: https://go.dev/play/p/VpRRmG7Bbnl

```go
func strip(presigned string) (string, error) {
	u, err := url.Parse(presigned)
	if err != nil {
		return "", err
	}

	u.User = nil
	u.ForceQuery = false
	u.RawQuery = ""
	return u.String(), nil
}
```

getglad:
@bbernays - I would have a high value for finding a way to source this information without ever making an API call that could pull the URL. In order to perform the action, the principal would still have to have a role with that permission. It feels like an unnecessary risk for what should otherwise be a configuration-only operation/permission set.

bbernays:
> @bbernays - I would have a high value for finding a way to source this information without ever making an API call that could pull the URL. In order to perform the action, the principal would still have to have a role with that permission. It feels like an unnecessary risk for what should otherwise be a configuration-only operation/permission set.

@getglad - I have updated the logic so that if the principal doesn't have access for `GetFunction` it will fall back and call `GetConcurrency` and `ListTags`... This balances performance and security as those orgs that don't even want to make the API call that could return sensitive information don't have to provide access, but it will require 2 more API calls to fully resolve. Does this work for you?

disq (approved):

plugins/source/aws/resources/services/lambda/functions.go
```diff
@@ -62,6 +63,24 @@ func Functions() *schema.Table {
 				Type:     arrow.BinaryTypes.String,
 				Resolver: resolveRuntimeManagementConfig,
 			},
+			{
+				Name: "code",
```
erezrokah, line 67: What's the reason we need to declare `code` and `configuration`?

erezrokah (approved):

plugins/source/aws/resources/services/lambda/functions.go
bbernays: This PR has all of the functional changes but fails because of `gen` changes
https://github.com/cloudquery/cloudquery/actions/runs/6433569162/job/17470896579

PR #14440, opened 2023-10-06
chore: Use  for wait for all workflow

erezrokah:
#### Summary

We only get [1000 requests per hour per repository](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions) with the built-in token, so this should help avoid reaching rate limits (personal access tokens get 5000 requests per hour).

I also contacted GitHub support to see if they can increase the limit for us


erezrokah:
Going to merge this to fix the CI

erezrokah:
Actually we can't use this token for forks

PR #14458, opened 2023-10-09
feat: Move gcp to dbt (only 1 section)

yevgenypats:
As we are trying to increase adoption to our policies, create documentation, testing, support for other database we need better tooling. DBT provides most of that (at least for now). I moved gcp section 1 to dbt. Full policy with dbt will be paid.

Things to do:

- documentation and README 
- testing
- making sure the project structure is built in such way that we can import it from our pro/full policy


erezrokah:
https://github.com/cloudquery/cloudquery/pull/14460 should fix the markdown linting failures

disq (approved):

Issue #14464, opened 2023-10-09
chore: Hitting rate limits in our CI with the built-in GitHub token


erezrokah:
See https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions

This is high priority since it's blocking PRs

```[tasklist]
### Related work
- [x] Disable https://github.com/cloudquery/cloudquery/blob/04a500d3620f23a4fb89a80adbe8b290540e121d/.github/workflows/greetings.yml
- [x] Use personal access token where possible https://github.com/cloudquery/cloudquery/pull/14463, https://github.com/cloudquery/cloudquery/pull/14473
- [x] Use `git` based diffing instead of GitHub API, less polling in wait for all https://github.com/cloudquery/cloudquery/pull/14467
```

erezrokah:
This is done I think and also splitting the open source monorepo to the private one helped too

PR #14473, opened 2023-10-09
chore: Use `cq-bot` token in cache cleanup

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/14464


.github/workflows/clean_actions_cache.yml
```diff
@@ -1,16 +1,13 @@
 name: Cleanup caches by a branch
 on:
-  pull_request:
+  pull_request_target:
```
erezrokah, line 3: So we can use `secrets.GH_CQ_BOT`

PR #14474, opened 2023-10-09
chore: Move `hashicorp-vault` pages under `vault`

erezrokah:
#### Summary

This makes the `vault` plugin consistent with how we structure the repository, see https://github.com/cloudquery/cloudquery/tree/32b5a9019833d44177caf6ec716215d0b4c8348b/plugins/source

It also fixes the docs broken link here https://cloud-hub.cloudquery.io/plugins/source/cloudquery/vault


hermanschaaf:
I think the makefile will also need to be updated: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/vault/Makefile#L28-L33

erezrokah:
> I think the makefile will also need to be updated: https://github.com/cloudquery/cloudquery/blob/main/plugins/source/vault/Makefile#L28-L33

Good catch! Updated in https://github.com/cloudquery/cloudquery/pull/14474/commits/635e8d6912df01865518df6fb1b5897a349e4a38

erezrokah:
/gen sha=635e8d6912df01865518df6fb1b5897a349e4a38 plugin=vault

PR #14476, opened 2023-10-09
fix: Fallback to `eu-west-1` if a given buckets LocationConstraint is `EU`

AshCorr:
#### Summary

When using the AWS Plugin we noticed that some of our buckets weren't being indexed by Cloudquery. After some investigation we noticed that those buckets seem to have a LocationConstraint of `EU` instead of an expected region such as `eu-west-1`

```bash
➜  aws git:(aa/EU-Region-Buckets) ✗ aws s3api get-bucket-location --bucket (redacted working bucket) --no-paginate

{
    "LocationConstraint": "eu-west-1"
}
```

```bash
➜  aws git:(aa/EU-Region-Buckets) ✗ aws s3api get-bucket-location --bucket (redacted broken bucket) --no-paginate

{
    "LocationConstraint": "EU"
}
```

This PR changes how we decide which region to use fetching s3 bucket data, falling back to `eu-west-1` when a buckets LocationConstraint is set to `EU`.

#### Why?!

It looks like this fallback to `eu-west-1` was a built in behaviour of the V1 AWS SDK https://github.com/aws/aws-sdk-go/blob/592707ce45b91664d30c270ad0057de033ce4dfe/service/s3/bucket_location.go#L23C6-L32 but was not ported to the V2 API which is used for Cloudquery https://github.com/aws/aws-sdk-go-v2/issues/1473#issuecomment-961732783

In the long run we should switch to the [HeadBucket](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html) instead, which has this behaviour built in to the AWS API response.


AshCorr:
Actually doesn't look too bad to switch to HeadBucket instead. DNM.

jsonpr:
Hey @AshCorr - interesting find and thanks for the PR!  Could you switch this PR to `Draft` for now or closing this until you're ready with the `HeadBucket` change or whichever change you think is appropriate at the time?

bbernays:
@AshCorr - I think that switching to `HeadBucket` while is the recommended solution, it would require us to release a breaking version of the AWS plugin because users would be forced to change their permissions.

While I am not opposed to making that change, I think that we should immediately make the change you have proposed to correctly handle the `EU` value.

AshCorr:
> I think that switching to HeadBucket while is the recommended solution, it would require us to release a breaking version of the AWS plugin because users would be forced to change their permissions.

Ahh good point. I've re-opened the PR.

bbernays (approved):

PR #14484, opened 2023-10-10
feat: Move Alicloud plugin docs to Markdown

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloud/issues/457 (internal issue)

~~Still work in progress, but this should allow us to switch to Markdown instead of MDX and inject our own components for `badge, configuration, authentication` gradually.~~

~~It uses `remark-directive` to extend markdown capabilities. I'll verify this works for our Hub as well~~

Instead of MDX we can use MD with custom directives and pass the related components. This should allow us to move existing docs gradually.


erezrokah:
Actually I think I should move `_configuration.mdx` and `_authentication.mdx` to markdown too

website/utils/components.tsx
```diff
@@ -0,0 +1,7 @@
+import AlicloudConfiguration from "../pages/docs/plugins/sources/alicloud/_configuration.md";
+import AlicloudAuthentication from "../pages/docs/plugins/sources/alicloud/_authentication.md";
+
+export const components = {
```
erezrokah, line 4: We need to update this manually each time we add a plugin, but this will be temporary until we move everything to the Hub.

I tried using dynamic imports via `const component = await import(componentPath)` but it seems nextra doesn't support it directly in the config

.github/workflows/publish_plugin_to_hub.yml
```diff
@@ -104,9 +107,9 @@ jobs:
       - name: Run package command
         working-directory: ${{ needs.prepare.outputs.plugin_dir }}
         run: |
-          mkdir -p stubdocs
-          touch stubdocs/README.md
-          go run main.go package --docs-dir stubdocs -m ${{ steps.release-notes.outputs.result }} ${{ needs.prepare.outputs.plugin_kind }} ${{ needs.prepare.outputs.plugin_version }} .
+          rsync -v --exclude='tables.md' ../../../website/pages/docs/plugins/${{ needs.prepare.outputs.plugin_kind_plural }}/${{ needs.prepare.outputs.plugin_name }}/*.md docsdir/
```
erezrokah, line 110: Needs https://github.com/cloudquery/cloudquery/pull/14498 to work. We exclude the tables since those are handled differently

mnorbury (approved):

PR #14493, opened 2023-10-10
feat: Add team context selection via `switch`

hermanschaaf:
This change is focused around adding the ability to set a team context via a new `cloudquery switch` command that takes a team name, but also introduces a few other features that are adjacent to that:
 - Adds `cloudquery switch` (prints current team) and `cloudquery switch <team>` (switches team)
 - Adds a `cloudquery logout` command to log out
 - Adds a `--team` option to `cloudquery login` as shorthand for `cloudquery login` followed by `cloudquery switch <team>`
 - Introduces a `config` library for managing config files. We currently only store a `team` value, but this may be extended in the future.
 - Removes the `--url` parameter from `cloudquery publish` and uses an environment variable for this instead. Usually we don't use environment variables, but in this case it is intentional, because you would only ever need to set the URL in tests.
 - All tests now use environment variables for overriding the API and accounts URLs
 - Renaming `CQ_API_KEY` to `CLOUDQUERY_API_KEY` for clarity and consistency (we will have to update the publish CI job to handle this CC @erezrokah)


hermanschaaf:
Ah I will need to fix the `switch` test, it was working locally because I had an auth token...

cli/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+// GetValue reads the value of a config key from the config file
+func GetValue(key string) (string, error) {
+	if !slices.Contains(configKeys, key) {
+		return "", fmt.Errorf("invalid config key %v (options are: %v)", key, strings.Join(configKeys, ", "))
+	}
+	configFilePath, err := xdg.ConfigFile(configPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to get config file path: %w", err)
+	}
+	b, err := os.ReadFile(configFilePath)
```
erezrokah, line 28: Can we verify that reading a value from the config doesn't fail if the file doesn't exist? For example if I run `cloudquery switch` without arguments and before setting any team seems like we fail with an error now.

I would expect it to print the help instead if there's no team

erezrokah (approved):
Looks good, one question on `cloudquery switch`

erezrokah:
> * Renaming `CQ_API_KEY` to `CLOUDQUERY_API_KEY` for clarity and consistency (we will have to update the publish CI job to handle this CC @erezrokah)

This is fine since we don't use API keys at the moment https://github.com/cloudquery/cloudquery/blob/914f5badf75b87814d54814658886e9335da6ddd/.github/workflows/publish_plugin_to_hub.yml#L119

cli/internal/config/config.go
hermanschaaf: Good point, will add a test for this and fix if needed

hermanschaaf: Added a test for this and fixed in [75ec8e6](https://github.com/cloudquery/cloudquery/pull/14493/commits/75ec8e68422b85ffadcffa724ef1d20a24effa92)

PR #14528, opened 2023-10-13
fix: Late setup for format and stage

disq:
Shouldn't attempt to `CREATE STAGE` and `CREATE FORMAT` before migrate starts so we can have `cloudquery migrate` with fewer permissions required.

### DIFF
--- a/plugins/destination/snowflake/client/client.go
+++ b/plugins/destination/snowflake/client/client.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"sync"
 
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 	"github.com/cloudquery/plugin-sdk/v4/writers/batchwriter"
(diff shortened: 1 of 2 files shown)


plugins/destination/snowflake/client/client.go
```diff
@@ -19,12 +20,15 @@ type Client struct {
 	logger zerolog.Logger
 	spec   Spec
 	writer *batchwriter.BatchWriter
+
+	setupWriteOnce *sync.Once
```
erezrokah, line 24: We could use the fancy new https://pkg.go.dev/sync#OnceValue now that we're on Go 1.21

erezrokah (approved):

plugins/destination/snowflake/client/client.go
disq: We could, but init wasn't a good place for write-related code anyway. Plus if we did that we'd need to use the `ctx` from `New` and not `Write`?

erezrokah: Yeah, non blocking comment this is good to go

PR #14535, opened 2023-10-15
doc: Status badges to readme

R3DRUN3:
#### Summary
*Safe to merge PR*  
added `license`, `go report`, `ci` and `latest release` status badges to readme to improve docs quality.  
Take a look at the edited readme in [my fork](https://github.com/R3DRUN3/cloudquery/blob/main/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 <h1 align="center"><img alt="cloudquery logo" width=75% src="https://github.com/cloudquery/cloudquery/raw/main/cli/docs/images/logo.png"/></h1>
 <!-- markdownlint-enable MD033 -->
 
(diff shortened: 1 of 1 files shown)


candiduslynx:
Hi @R3DRUN3!
Thanks for the PR, it looks great!

Just one thing I'd like to change: we have a monorepo here, so the latest release will point sometimes to some plugins, sometimes to CLI. Is it possible to get the latest CLI release only? We tag them in a specific pattern, so it'll be easy to query.

candiduslynx (changes requested):
Need to figure out a way to point only to the CLI release (or drop the release badge altogether)

R3DRUN3:
> Need to figure out a way to point only to the CLI release (or drop the release badge altogether)

@candiduslynx unfortunately, it appears that there is no straightforward and efficient method for selectively referencing only the CLI releases.  
Consequently, I have removed the badge for the latest release.

erezrokah (approved):

PR #14543, opened 2023-10-16
chore: Disable setup-go cache in workflows we use cache action directly

erezrokah:
#### Summary

`setup-go` has caching enabled by default. In the validate-release/release workflows we use the cache action directly to control the cache key, so we should disable it for `setup-go`. Please note `setup-go` cache didn't work for these workflows anyway since it couldn't find the `go.sum` file. This PR doesn't change any cache related behavior it only fixes the warnings we're getting like:
https://github.com/cloudquery/cloudquery/actions/runs/6531867504
image

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -79,6 +79,7 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version-file: cli/go.mod
+          cache: false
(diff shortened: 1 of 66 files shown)


hermanschaaf (approved):

erezrokah:
Ah, forgot to open this not from a fork, so workflows that require elevated context will fail

PR #14547, opened 2023-10-16
Chore: bump plugin-pb-go to v1.12.3 and plugin-sdk to v4.13.0

maaarcelino:
Manual bump of plugin-pb-go and plugin-sdk


erezrokah (approved):
This only does sources, do we want destinations too?

erezrokah:
I suggest we close this PR in favor of https://github.com/cloudquery/cloudquery/pull/14551, WDYT?

PR #14556, opened 2023-10-16
feat: Add JSON schema for plugin spec

### LINKED ISSUE #14450: Add JSON schema to `plugins/source/k8s`

candiduslynx:
Closes #14450


erezrokah (approved):

plugins/source/k8s/client/spec/schema.json
```diff
@@ -0,0 +1,32 @@
...
+            }
+          ]
+        },
+        "concurrency": {
+          "type": "integer",
+          "minimum": 1,
+          "default": 50000
+        }
+      },
+      "additionalProperties": false,
```
erezrokah, line 28: Can we not enforce `additionalProperties: false`? Seems a bit too strict

erezrokah: I think this means that if we add a new config property, update our docs to include, then someone copy pastes the docs and still uses an older version they'll get a validation error

candiduslynx: Currently it's not failing, only producing an err log because of https://github.com/cloudquery/plugin-sdk/pull/1280.

We can make a breaking change once we decide to actually fail on the wrong plugin cfg.

PR #14564, opened 2023-10-16
feat: Addon CLI

disq:
Implements https://github.com/cloudquery/cloudquery-issues/issues/600

`cloudquery install` is now `cloudquery plugin install` (`cloudquery install` still works...)
`cloudquery publish` => `cloudquery plugin publish`

new command: `cloudquery addon publish`


cli/cmd/addon_publish.go
```diff
@@ -0,0 +1,245 @@
...
+
+	return cmd
+}
+
+type ManifestJSONV1 struct {
+	Type      string `json:"type"` // always "addon"
+	TeamName  string `json:"team_name"`
+	AddonName string `json:"addon_name"`
+	// AddonType   string   `json:"addon_type"`   // unused
+	// AddonFormat string   `json:"addon_format"` // unused
```
disq, line 65: These fields belong to the `addon` entry and not `addon_version`, so there's no use for them here. We could try to `GET` the addon and compare/verify the values if we really want to make sure.

hermanschaaf: Seems fine, though I would probably still read the properties, but just not use them

cli/cmd/addon_publish.go
```diff
@@ -0,0 +1,245 @@
...
+
+	zipPath := filepath.Join(manifestDir, manifest.PathToZip)
+	if _, err := os.Stat(zipPath); err != nil {
+		return fmt.Errorf("could not read file %s: %w", zipPath, err)
+	}
+
+	name := fmt.Sprintf("%s/%s@%s", manifest.TeamName, manifest.AddonName, version)
+	fmt.Printf("Publishing %s to CloudQuery Hub...\n", name)
+
+	c, err := cloudquery_api.NewClientWithResponses(getEnvOrDefault("CLOUDQUERY_API_URL", defaultAPIURL),
```
hermanschaaf, line 101: Let's create a constant in `constants.go` for `CLOUDQUERY_API_URL`, since it's used in multiple places now

hermanschaaf (approved):
Nice, looks good 👍

cli/cmd/addon_publish.go
disq: Done in [`fcda030` (#14564)](https://github.com/cloudquery/cloudquery/pull/14564/commits/fcda03061b03e7560b91ffcf3eb53f60dcaa6eff)

cli/cmd/addon_publish.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return nil
+}
+
+func createNewAddonDraftVersion(ctx context.Context, c *cloudquery_api.ClientWithResponses, manifest ManifestJSONV1, version, manifestDir, zipPath string) error {
+	if manifest.PluginDeps == nil {
+		manifest.PluginDeps = []string{}
+	}
+	if manifest.AddonDeps == nil {
+		manifest.AddonDeps = []string{}
```
disq, line 156: otherwise we get:
> Error: failed to create new draft version: 400: request body has an error: doesn't match schema: Error at "/addon_deps": Value is not nullable

.github/workflows/publish_plugin_to_hub.yml
```diff
@@ -121,4 +121,4 @@ jobs:
         run: |
           mkdir -p ~/.local/share/cloudquery
           echo ${{ secrets.CQ_CI_CLOUDQUERY_HUB_TOKEN }} > ~/.local/share/cloudquery/token
-          cloudquery publish cloudquery/${{ needs.prepare.outputs.plugin_name }} --finalize
+          cloudquery plugin publish cloudquery/${{ needs.prepare.outputs.plugin_name }} --finalize
```
erezrokah, line 124: For this to work we actually need to release a new CLI version and update it in the workflow here https://github.com/cloudquery/cloudquery/blob/8946c36362c4344115e015f0716b8aafb7cabd83/.github/workflows/publish_plugin_to_hub_duckdb.yml#L96

erezrokah: See https://github.com/cloudquery/cloudquery/actions/runs/6562442141/job/17824393847#step:11:7.

I'll release https://github.com/cloudquery/cloudquery/pull/14659 and renovate should handle the update for us

PR #14572, opened 2023-10-16
fix: Github source plugin hanging on missing permissions

AshCorr:
#### Summary

Bump `github.com/gofri/go-github-ratelimit` to v1.0.4 in the Github source plugin.

v1.0.3 treats `403 Forbidden` errors from the Github API due to missing permissions as breaching the rate limit, this causes the Cloudquery to wait until the rate limit reset (potentially up to an hour!) before trying again, at which point it will keep repeating this until the token is granted the correct permissions or the job is terminated.

v1.0.4 improves the rate limit detection behaviour so that it doesn't detect missing permission errors as being rate limited https://github.com/gofri/go-github-ratelimit/pull/13


AshCorr:
I think we need to wait until v1.0.5 of this dep is released. Now its not detecting things as breaking the rate limit when it should.

Issue #14548, opened 2023-10-16
bug: DuckDB tests sometimes fail


hermanschaaf:
There seems to be a race between flushing and reading rows in the latest version of the DuckDB plugin, leading to this occasional test failure:

```
--- FAIL: TestPlugin (5.50s)
    testing_write_delete.go:142: 
        	Error Trace:	/home/runner/go/pkg/mod/github.com/cloudquery/plugin-sdk/v4@v4.12.5/plugin/testing_write_delete.go:142
        	            				/home/runner/go/pkg/mod/github.com/cloudquery/plugin-sdk/v4@v4.12.5/plugin/testing_write.go:147
        	Error:      	Not equal: 
        	            	expected: int(10)
        	            	actual  : int64(20)
        	Test:       	TestPlugin
        	Messages:   	unexpected amount of items after second delete stale
    --- FAIL: TestPlugin/TestDeleteStale (1.[9](https://github.com/cloudquery/cloudquery/actions/runs/6532488275/job/17735758206#step:6:10)4s)
        --- FAIL: TestPlugin/TestDeleteStale/All (1.92s)
            testing.go:[14](https://github.com/cloudquery/cloudquery/actions/runs/6532488275/job/17735758206#step:6:15)90: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
        testing.go:1490: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

(Note: expected is 10, actual is 20, we can ignore the int64/int difference)

hermanschaaf:
I spent some time on this but haven't yet found a satisfactory solution. I found that if you add a 1 second sleep on this line: https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/duckdb/client/delete_stale.go#L29 the tests mostly pass when running with `go test -count 10 -run TestPlugin/TestDeleteStale -race -timeout 3m ./...`, whereas without it they don't. I also tried some other ideas, like running a vacuum optimize operation or querying the tables to wait for the rows to be deleted before exiting the deletestale function, but these did not solve the issue. 

I opened https://github.com/cloudquery/cloudquery/pull/14573 to fix another testing issue discovered while testing this, but I'm going to put this aside for now, hoping a future a DuckDB library version will either fix this or give us the ability to ensure a command has been flushed before exiting the delete stale function.

erezrokah:
Closing as stale. Have not seen this pop up and we can re-open if we do

PR #14573, opened 2023-10-16
fix: Remove unnecessary type registration

hermanschaaf:
Without this, running `go test -count 2` fails with:

```
--- FAIL: TestPlugin (0.00s)
    client_test.go:16: arrow: type extension with name uuid already defined
```

No other plugins call this in the tests anymore, and I think it's a leftover from a previous migration. The registration now happens automatically in the plugin serve method.


disq (approved):

hermanschaaf:
Tests failing due to https://github.com/cloudquery/cloudquery/issues/14548, the issue I was actually trying to fix when finding this issue 🙃

Issue #14465, opened 2023-10-09
bug: SQLite plugin cannot be use with the docker image


moadibfr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello there,
I was trying to use the docker image to run a sync with a SQLite destination and after sucessfully donwloading the plugin I got the error that the plugin was not found:
```
docker run -v $PWD:$PWD --workdir $PWD ghcr.io/cloudquery/cloudquery:latest sync aws.yml sqlite.yml
Loading spec(s) from aws.yml, sqlite.yml
Downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-source-aws-v22.13.2/aws_linux_amd64.zip
Downloading 100% |████████████████████████████████| (59/59 MB, 33 MB/s)        
Downloading https://github.com/cloudquery/cloudquery/releases/download/plugins-destination-sqlite-v2.4.10/sqlite_linux_amd64.zip
Downloading 100% |██████████████████████████████| (8.0/8.0 MB, 24 MB/s)        
Error: failed to start plugin .cq/plugins/destination/cloudquery/sqlite/v2.4.10/plugin: fork/exec .cq/plugins/destination/cloudquery/sqlite/v2.4.10/plugin: no such file or directory
```

### Expected Behavior

I expect the sync to work as intended.

### CloudQuery (redacted) config

```
❯ cat aws.yml
spec:
  path: cloudquery/aws
  tables:
  - aws_cloudtrail_events
  destinations:
  - sqlite
  name: aws
  version: v22.13.2
  spec:
    custom_endpoint_url: http://127.0.0.1:32835
    custom_endpoint_hostname_immutable: "true"
    max_retries: 0
    custom_endpoint_partition_id: aws
    custom_endpoint_signing_region: us-east-1
kind: source
❯ cat sqlite.yml
spec:
  version: v2.3.0
  name: sqlite
  spec:
    connection_string: ./db.sqlite
  path: cloudquery/sqlite
kind: destination

```

### Steps To Reproduce

`docker run -v $PWD:$PWD --workdir $PWD ghcr.io/cloudquery/cloudquery:latest sync aws.yml sqlite.yml`
with the provided config files

### CloudQuery (redacted) logs

```
2023-10-09T08:00:41Z INF Loading spec(s) args=["aws.yml","sqlite.yml"] module=cli
[...]

PR #14612, opened 2023-10-17
fix: Compile DuckDB, SQLite and Snowflake plugins for running under Alpine Linux 

maaarcelino:
We currently have three CGO_ENABLED plugins: DuckDB, SQLite and Snowflake. These don't correctly run under Alpine Linux due to missing libraries in the compiled executable. This adds statically linked C libraries to the executable for those plugins so they can run on those distros.

See https://github.com/cloudquery/cloudquery/issues/14465 for more info.


erezrokah (approved):
🚀

erezrokah:
Test failures seems to be related to https://github.com/cloudquery/cloudquery/issues/14548

PR #14631, opened 2023-10-17
refactor: Move auth package to `cloudquery-api-go`

mnorbury:
Updates to the latest `cloudquery-api-go` version

fixes: https://github.com/cloudquery/cloudquery-issues/issues/668


cli/cmd/login.go
```diff
@@ -160,20 +161,21 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
...
 
 	if cmd.Flags().Changed("team") {
-		team := cmd.Flag("team").Value.String()
-		token, err := auth.GetToken()
+		selectedTeam := cmd.Flag("team").Value.String()
+		tc := auth.NewTokenClient()
+		token, err := tc.GetToken()
 		if err != nil {
-			return fmt.Errorf("failed to get auth token: %w", err)
+			return fmt.Errorf("failed to get team token: %w", err)
```
hermanschaaf, line 168: this doesn't seem right; it's still a per-user auth token

mnorbury: Whoops, over enthusiastic renaming by Intellij

cli/cmd/logout.go
```diff
@@ -41,7 +41,7 @@ func newCmdLogout() *cobra.Command {
 }
 
 func runLogout(_ context.Context, cmd *cobra.Command) error {
-	err := auth.Logout()
+	err := team.Logout()
```
hermanschaaf, line 44: I think `Logout` should be under auth, not under `team`

mnorbury: `auth` in the current repository or the `cloudquery-api-go`? I resisted putting it in the `cloudquery-api-go` on the basis that it felt more like a [user](https://github.com/cloudquery/cloudquery/blob/main/cli/internal/auth/logout.go) of the shared code e.g. it defines what keys to store.

hermanschaaf (approved):

hermanschaaf:
I think this should be marked as `refactor` or `chore`, right? We don't want to have this generate a new release

disq (approved):

PR #14638, opened 2023-10-17
feat(resources): Add support for GCP organization and project tags

DougEverly:
#### Sync GCP Organization and Project Tags

* Sync tagKeys and tagValues for organizations
* Sync tagKeys, tagValues, and tagBindings for projects.


plugins/source/gcp/resources/services/resourcemanager/organization_tag_values.go
```diff
@@ -0,0 +1,57 @@
...
+	pb "cloud.google.com/go/resourcemanager/apiv3/resourcemanagerpb"
+	"context"
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/cloudquery/cloudquery/plugins/source/gcp/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+	"google.golang.org/api/iterator"
+)
+
+func OrganizationTagValues() *schema.Table {
```
erezrokah, line 14: Can we rename this function to `organizationTagValues` so it becomes a private one? We usually do it for non top level tables so they don't get mistakenly used outside their package

plugins/source/gcp/resources/services/resourcemanager/tag_keys.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/cloudquery/cloudquery/plugins/source/gcp/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+	"google.golang.org/api/iterator"
+)
+
+func TagKeys() *schema.Table {
+	return &schema.Table{
+		Name:        "gcp_resourcemanager_tag_keys",
```
erezrokah, line 16: Should we call this table `gcp_resourcemanager_project_tag_keys`? That would make it consistent with `gcp_resourcemanager_organization_tag_values`. If so we should rename `TagKeys` to `ProjectTagKeys`

plugins/source/gcp/resources/services/resourcemanager/tag_values.go
```diff
@@ -0,0 +1,57 @@
...
+	pb "cloud.google.com/go/resourcemanager/apiv3/resourcemanagerpb"
+	"context"
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/cloudquery/cloudquery/plugins/source/gcp/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+	"google.golang.org/api/iterator"
+)
+
+func TagValues() *schema.Table {
```
erezrokah, line 14: Similar comment about making this function private

erezrokah (commented):
Nice work  @DougEverly, added a few comments regarding conventions we have.
Also do you mind add mock tests? See example in https://github.com/cloudquery/cloudquery/blob/e5a367e54fe67fc1fca0448a8d4cd9fac670c95a/plugins/source/gcp/resources/services/resourcemanager/projects_mock_test.go

plugins/source/gcp/resources/services/resourcemanager/tag_keys.go
DougEverly: Updated to use `organization` and `project` prefixes for consistency.

plugins/source/gcp/resources/services/resourcemanager/tag_values.go
DougEverly: Made dependent table methods private.

plugins/source/gcp/resources/services/resourcemanager/organization_tag_values.go
DougEverly: Made dependent table methods private.

erezrokah (approved):
Looks great @DougEverly, thanks for adding the tests and following up on my comments. I pushed a small commit to removing a `Printf` and fix a typo

PR #14647, opened 2023-10-18
fix: Add descriptions to JSON schema

candiduslynx:
See https://github.com/invopop/jsonschema/pull/112 (https://github.com/cloudquery/jsonschema/pull/9)

Blocked by https://github.com/cloudquery/codegen/pull/60


.github/workflows/source_aws.yml
```diff
@@ -50,6 +50,8 @@ jobs:
         uses: cloudquery/setup-cloudquery@v3
         with:
           version: v3.23.1
+      - name: vendor
```
erezrokah, line 53: What's the reason we need this?

candiduslynx: `gen` step (`make gen` -> `make gen-spec-schema`) requires code to be present to load comments from AWS SDK

candiduslynx: I can move `go mod vendor` to `gen-spec-schema` step, it will have the same effect

erezrokah (approved):

.github/workflows/source_aws.yml
candiduslynx: Moved in 11e4350

PR #14649, opened 2023-10-18
chore: Removes the xdg (direct) dependency

mnorbury:
The xdg logic for configuraiton management is now available through the `cloudquery-api-go` dependency and is no longer needed as a direct dependency in the CLI.


erezrokah (approved):

erezrokah:
Go we need a `go mod tidy` to remove it from `go.mod`? Or is it used somewhere else?

Issue #14632, opened 2023-10-17
feat: Add EFS FileSystemPolicy to AWS source plugin


mMeijden:
### Describe the resource

EFS supports File system policies to be configured. These policies are currently not collected by CloudQuery.
https://docs.aws.amazon.com/efs/latest/ug/API_DescribeFileSystemPolicy.html 

Blog post:
https://aws.amazon.com/blogs/aws/new-for-amazon-efs-iam-authorization-and-access-points/ 

Can this be added to the EFS table or as a separate table perhaps?

### Use Case

Our use case is to build a full cloud resource inventory and perform security and compliance analyses on top of that. 
One of the reason we require the file systems is to ensure that they are not overprivileged or public.

### Link to API documentation

https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#EFS.DescribeFileSystemPolicy

### Pull request (optional)

- [ ] I can submit a pull request

cychiang:
I'm working on this at the moment, and it looks nature to me by extending the `filesystems` table with the `filesystem_policy` column.

PR #14672, opened 2023-10-18
feat: Add `policy` column to `aws_efs_filesystems` table

cychiang:
#### Summary

resolves cloudquery/cloudquery#14632

I'm looking for a way to reduce the duplicate implementations, for example, the following part is a way to get a service to resolve data for the table. It works well when a table relays on a single resolver. But if a table needs multiple resolver then might be a good way to do it. 

```go
	p := resource.Item.(types.FileSystemDescription)
	config := efs.DescribeFileSystemPolicyInput{
		FileSystemId: p.FileSystemId,
	}
	cl := meta.(*client.Client)
	svc := cl.Services(client.AWSServiceEfs).Efs
```


plugins/source/aws/resources/services/efs/filesystems.go
```diff
@@ -82,3 +87,26 @@ func ResolveEfsFilesystemBackupPolicyStatus(ctx context.Context, meta schema.Cli
...
+		FileSystemId: p.FileSystemId,
+	}
+	cl := meta.(*client.Client)
+	svc := cl.Services(client.AWSServiceEfs).Efs
+	response, err := svc.DescribeFileSystemPolicy(ctx, &config, func(options *efs.Options) {
+		options.Region = cl.Region
+	})
+	if err != nil {
+		if cl.IsNotFoundError(err) {
+			return resource.Set(c.Name, types.StatusDisabled)
```
bbernays, line 103: Instead of setting this to a value, can you just leave it unset to follow the convention of all of the other resources? Also a 404 doesn't automatically mean that it is disabled, the resource could have also been deleted in the time between the table resolver being called and the column resolver being called.

bbernays (changes requested):

plugins/source/aws/resources/services/efs/filesystems.go
```diff
@@ -82,3 +87,26 @@ func ResolveEfsFilesystemBackupPolicyStatus(ctx context.Context, meta schema.Cli
 
 	return resource.Set(c.Name, response.BackupPolicy.Status)
 }
+
+func ResolveEfsFilesystemPolicy(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
bbernays, line 91:
```suggestion
func resolveEfsFilesystemPolicy(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```

This should be a private function and not available outside of the package

cychiang: Done, I also change another function to the lower case as well because I see that function also a private funciton.

plugins/source/aws/resources/services/efs/filesystems.go
```diff
@@ -82,3 +87,23 @@ func ResolveEfsFilesystemBackupPolicyStatus(ctx context.Context, meta schema.Cli
...
+		FileSystemId: p.FileSystemId,
+	}
+	cl := meta.(*client.Client)
+	svc := cl.Services(client.AWSServiceEfs).Efs
+	response, err := svc.DescribeFileSystemPolicy(ctx, &config, func(options *efs.Options) {
+		options.Region = cl.Region
+	})
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
cychiang: I see, then I will change it as following for `resolveEfsFilesystemBackupPolicyStatus` and `resolveEfsFilesystemPolicy`  for the error handling

```go
 if err != nil { 
 	if cl.IsNotFoundError(err) { 
 		return nil 
 	} 
 	return err 
 } 
```

plugins/source/aws/resources/services/efs/filesystems.go
```diff
@@ -72,7 +77,7 @@ func ResolveEfsFilesystemBackupPolicyStatus(ctx context.Context, meta schema.Cli
 	})
 	if err != nil {
 		if cl.IsNotFoundError(err) {
-			return resource.Set(c.Name, types.StatusDisabled)
+			return nil
```
bbernays, line 80: Lets not change this in this pr because then we will have to mark the next release as a breaking change. I have added that [change to our list of](https://github.com/cloudquery/cloudquery/issues/12648) changes to include in the next breaking change (which is tbd)

plugins/source/aws/resources/services/efs/filesystems.go
```diff
@@ -32,7 +32,12 @@ func Filesystems() *schema.Table {
 			{
 				Name:     "backup_policy_status",
 				Type:     arrow.BinaryTypes.String,
-				Resolver: ResolveEfsFilesystemBackupPolicyStatus,
+				Resolver: resolveEfsFilesystemBackupPolicyStatus,
+			},
+			{
+				Name:     "policy",
```
bbernays, line 38:
```suggestion
				Name:     "file_system_policy",
```
so it is clear what is in the column, I think the name should be `file_system_policy`

bbernays (changes requested):
This PR is almost there! one last issues/change about the column name

bbernays (approved):

PR #14696, opened 2023-10-19
doc: Add a note on sensitive information when debug logging is enabled

pilvikala:
#### Summary

Based on customer feedback, I have added a warning about CLI potentially printing sensitive information in logs when debug logging is enabled.


website/pages/docs/troubleshooting.mdx
```diff
@@ -24,6 +26,10 @@ There are a couple of ways to get help for any CloudQuery-related issues or ques
 
 Usually the first step that will be needed to debug/resolve an issue is to run `cloudquery` with `--log-level debug` to enable verbose logging.
 
+<Callout type="warning">
+The CLI may print out environment variables including sensitive information when the verbose logging is enabled.
```
hermanschaaf, line 30:
```suggestion
The CLI may print out environment variables, including sensitive information, when verbose logging is enabled.
```

erezrokah (commented):
Can you try the following MD syntax to add the Callout https://github.com/cloudquery/cloudquery/blob/6f766014836e1ebf481e5bac028536e814c06bf0/website/pages/docs/plugins/sources/aws/event-based-sync.md?plain=1#L1?

It should help us migrate this doc to the Hub if we need to

pilvikala:
> :::callout{type="info"}

This turned out to be a bit complicated because there were compilation issues and I had to fix a function used by the `customPlugin`. Let me know if this is not a good fix.

erezrokah (approved):

erezrokah:
> Let me know if this is not a good fix.

I think it's fine I might update the code in a follow up PR to only get `kind` and `name` when we need them

PR #14698, opened 2023-10-19
fix: Set plugin metadata

hermanschaaf:
This sets plugin metadata on all the CloudQuery-maintained plugins, following an update to plugin-sdk v4.16.0, which requires this metadata to be present for the `package` command to succeed.

### DIFF
--- a/plugins/source/firestore/resources/plugin/plugin.go
+++ b/plugins/source/firestore/resources/plugin/plugin.go
@@ -5,12 +5,19 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 )
 
-var version = "development"
+var (
+	Name    = "firestore"
+	Kind    = "source"
+	Team    = "cloudquery"
+	Version = "development"
+)
 
 func Plugin() *plugin.Plugin {
 	return plugin.NewPlugin(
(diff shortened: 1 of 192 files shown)


plugins/source/firestore/resources/plugin/plugin.go
```diff
@@ -5,12 +5,19 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 )
 
-var version = "development"
```
hermanschaaf, line 8: The fact that `version` was unexported here was actually a bug, as even official versions of the plugin right now report their version to be `development`:

```
./plugin --version
firestore version development
```

disq:
Couple of questions:

- Why `var` when they are constants?
- Why not use strings inside plugin init and define them as vars (or consts) when it's only in one place anyway?

disq (approved):

hermanschaaf:
@disq Keeping them var because we may want to replace these values with `ldflag` in the future (like we already do for version), and keeping them grouped because I think it's nice to have this metadata visible/editable in one place, rather than in a function call.

hermanschaaf:
Closing for now, I did a bad merge

PR #14715, opened 2023-10-19
fix: Set plugin metadata

hermanschaaf:
This sets plugin metadata on all the CloudQuery-maintained plugins, following an update to plugin-sdk v4.16.0, which requires this metadata to be present for the `package` command to succeed.

Second attempt of [#14698](https://github.com/cloudquery/cloudquery/pull/14698)

### DIFF
--- a/plugins/destination/gcs/main.go
+++ b/plugins/destination/gcs/main.go
@@ -15,7 +15,7 @@ const (
 )
 
 func main() {
-	p := plugin.NewPlugin("file", internalPlugin.Version, client.New)
+	p := plugin.NewPlugin(internalPlugin.Name, internalPlugin.Version, client.New, plugin.WithKind(internalPlugin.Kind), plugin.WithTeam(internalPlugin.Team))
 	if err := serve.Plugin(p, serve.WithPluginSentryDSN(sentryDSN), serve.WithDestinationV0V1Server()).Serve(context.Background()); err != nil {
 		log.Fatalf("failed to serve plugin: %v", err)
 	}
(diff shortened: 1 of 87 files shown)


disq (approved):

plugins/source/k8s/resources/plugin/plugin.go
```diff
@@ -33,7 +33,12 @@ import (
 	"golang.org/x/exp/maps"
 )
 
-var Version = "Development"
+var var (
+	Name = "gcp"
```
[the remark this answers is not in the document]
hermanschaaf: Nice catch, thanks!

mnorbury (approved):
👍

plugins/destination/gcs/main.go
```diff
@@ -15,7 +15,7 @@ const (
 )
 
 func main() {
-	p := plugin.NewPlugin("file", internalPlugin.Version, client.New)
```
hermanschaaf, line 18: This was also a bug before

plugins/source/k8s/resources/plugin/plugin.go
```diff
@@ -33,7 +33,12 @@ import (
 	"golang.org/x/exp/maps"
 )
 
-var Version = "Development"
+var var (
```
hermanschaaf, line 36:
```suggestion
var (
```

hermanschaaf:
Ok GitHub is struggling to process all the jobs in here, in retrospect I probably should've split this up into several PRs. I'm going to force it through and follow up with updates to save on CI time.

Issue #14652, opened 2023-10-18
What should Cloudquery plugins consider as "errors"


AshCorr:
### False positive for Errors

We've noticed a few tables from the AWS source plugin throwing seemingly false positives for errors in perfectly normal situations:

1. `aws_s3_bucket_object_lock_configurations`

    An error will be thrown by every S3 bucket that doesn't have Object Lock enabled as the AWS API will return a 404 response code when requesting bucket lock information.
    
2. `aws_lambda_functions`

    Any Lambda that uses a Image runtime will throw 2 errors when it tries to call the `GetRuntimeManagementConfig` and `GetFunctionCodeSigningConfig` AWS APIs, as neither API support image runtime Lambdas.

3. `aws_ecr_repository_lifecycle_policies`

    An error will be thrown by every ECR Repository that doesn't have a lifecycle policy as the AWS API will return a 404 response code when requesting lifecycle policy information.

_this is not an exclusive list_
    
In my opinion, it feels like these should be considered either "warnings" or "info", not "errors", as they're not caused by any underlying api outage/cloudquery misconfiguration/rate limit/etc but instead caused by perfectly normal AWS Resources.

### Impact

Ideally we'd be able to scan our logs for errors and alert us if anything comes up that needs our attention, for example if an API outage caused gaps in our data.

Theres a "table sync finished" log message that gets posted after every table has finished syncing which has a "errors" field indicating the total number of errors encountered. But with how many false positives there are it becomes un-usable

<img width="843" alt="image" src="https://github.com/cloudquery/cloudquery/assets/21217225/11a7e704-ae2e-405a-9f56-829886f4d20a">

We're happy to look at fixing some of these errors ourselves, but wanted to understand what the Cloudquery team considers an "error" to be before we do any work! Maybe some suggestions to plugin developers on what should be considered an error?

erezrokah:
Hi @AshCorr, thanks for raising this issue. Our current approach is to propagate errors by default as it's hard to predict how the AWS (or other APIs) respond. Some will throw 404, some will not error out, etc.

Sometimes we do handle the errors on a case to case basic when it makes sense, for example https://github.com/cloudquery/cloudquery/blob/4e669f9c13f7ed4b7bcdc1e8df5ceacf60917731/plugins/source/aws/resources/services/account/alternate_contacts.go#L45

We're happy to discuss more if you submit a PR per each case

bbernays:
[...]

PR #14729, opened 2023-10-19
fix: Don't call GetFunctionCodeSigningConfig or GetRuntimeManagementConfig on Lambdas packaged as Images (`aws_lambda_functions`)

AshCorr:
#### Summary

When syncing `aws_lambda_functions` Cloudquery has 2 resolvers for GetFunctionCodeSigningConfig and GetRuntimeManagementConfig. The AWS API does not support calling either of these APIs with Lambdas that use containerised images and this results in errors in Cloudquery.

Thankfully Cloudquery is great and doesn't panic when it encounters these errors and theres no loss in data, but it does log it as "something bad happening" when in reality this is expected behaviour from the AWS API.

It looks like there was an attempt to fix this before for GetFunctionCodeSigningConfig but it doesn't seem to be preventing the errors.

```
2:39PM ERR column resolver finished with error error="operation error Lambda: GetFunctionCodeSigningConfig, https response error StatusCode: 400, RequestID: (redacted), InvalidParameterValueException: Code signing is not supported for functions created with container images." client=(redacted) module=aws-src table=aws_lambda_functions
2:39PM ERR column resolver finished with error error="operation error Lambda: GetRuntimeManagementConfig, https response error StatusCode: 400, RequestID: (redacted), InvalidParameterValueException: Lambda couldn't get a runtime management configuration because (redacted) is a container image function." client=(redacted) module=aws-src table=aws_lambda_functions
```

Part of #14652


plugins/source/aws/resources/services/lambda/functions.go
```diff
@@ -153,9 +152,7 @@ func resolveCodeSigningConfig(ctx context.Context, meta schema.ClientMeta, resou
 	svc := cl.Services(client.AWSServiceLambda).Lambda
 
 	// skip getting CodeSigningConfig since containerized lambda functions does not support this feature
-	// value can be nil if the caller doesn't have GetFunctionConfiguration permission and only has List*
```
AshCorr, line 156: This is still the case but theres a nil check for `r.Configuration` on https://github.com/cloudquery/cloudquery/pull/14729/files#diff-93c24c91e5c0e22f86d6415f1b98b47062f7d8974b9337c0b13c01322b8df981L149

plugins/source/aws/resources/services/lambda/functions.go
```diff
@@ -153,9 +152,7 @@ func resolveCodeSigningConfig(ctx context.Context, meta schema.ClientMeta, resou
...
-	lambdaType := resource.Get("code_repository_type").(*scalar.String)
```
AshCorr, line 157: Not entirely sure why we previously relied on untyped `resource.Get` instead of directly accessing values from `r.Configuration`. Either way, this approach seemed to not be working for us as we were still getting errors from `GetRuntimeManagementConfig`.

bbernays (changes requested):
Great find!

plugins/source/aws/resources/services/lambda/functions.go
```diff
@@ -153,9 +152,7 @@ func resolveCodeSigningConfig(ctx context.Context, meta schema.ClientMeta, resou
...
-	if lambdaType != nil && lambdaType.Value == "ECR" {
+	if r.Configuration.PackageType == "Image" {
```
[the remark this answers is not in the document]
AshCorr: Ah nice find! Will make this change tomorrow

bbernays (approved):

Issue #11225, opened 2023-06-06
feat: Add a backoff 


kvish:
### Which problem is this feature request solving?

Add similar parameters as the aws source plugin to control retry and rate limiting errors.

```
max_retries (int) (default: 10)

Defines the maximum number of times an API request will be retried

max_backoff (int) (default: 30)

Defines the duration between retry attempts
```

### Describe the solution you'd like

Add configuration options in to the config.yml file for Azure to pass client settings for backoff

### Pull request (optional)

- [ ] I can submit a pull request

erezrokah:
Hi @kvish, can you please share under which scenario you're seeing rate limiting errors with Azure? It would help us verify that adding retry settings will solve those issues.
Also it would be interesting to see if using a different scheduling algorithm can help too. See https://www.cloudquery.io/docs/advanced-topics/performance-tuning#tune-concurrency and https://www.cloudquery.io/docs/advanced-topics/performance-tuning#use-a-different-scheduler

Sharing the logs when the errors happen will be very useful.

shai-gundersen:
Hi @erezrokah 
I am getting a lot of rate limit errors when syncing tables of type `azure_storage_*`. 
Sharing some of the logs below.
I also notice that CQ is waiting, and after 5 minute as stated in `Read_ObservationWindow_00:05:00`, we actually do get the data and the sync ends successfully.

`2023-08-09T15:55:45Z ERR table resolver finished with error error="GET https://management.azure.com/subscriptions/xxxxx/resourceGroups/OVHWEPRD-QUO-RG/providers/Microsoft.Storage/storageAccounts/ovhweprdquostorage/tableServices/default/tables\n--------------------------------------------------------------------------------\nRESPONSE 429: 429 Too Many Requests\nERROR CODE: TooManyRequests\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"TooManyRequests\",\n    \"message\": \"The request is being throttled as the limit has been reached for operation type - Read_ObservationWindow_00:05:00. For more information, see - https://aka.ms/srpthrottlinglimits\"\n  }\n}\n--------------------------------------------------------------------------------\n" client=subscriptions/43e6231a-e00b-46fa-9d7b-aeb81c00b10d module=azure-src table=azure_storage_tables
[...]

Issue #9860, opened 2023-04-12
bug(azure): Cost management queries hitting rate limits - 429 Too Many Requests


disq:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Cost management queries seem to hit the rate limits and error out if there are many views to be queried.

### Expected Behavior

Fetch successfully

### CloudQuery (redacted) logs

```
3:05PM ERR table resolver finished with error error="POST https://management.azure.com/subscriptions/<uuid>/providers/Microsoft.CostManagement/query\n--------------------------------------------------------------------------------\nRESPONSE 429: 429 Too Many Requests\nERROR CODE: 429\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"429\",\n    \"message\": \"Too many requests. Please retry.\"\n  }\n}\n--------------------------------------------------------------------------------\n" client=subscriptions/<uuid> module=azure-src table=azure_costmanagement_view_queries
```

### CloudQuery version

azure v7.1.0

### Pull request (optional)

- [ ] I can submit a pull request

erezrokah:
Sharing https://github.com/Azure/azure-sdk-for-go/blob/ef4cd3ef8f9fc2deb4240daa0f822c3c9486e665/sdk/azcore/policy/policy.go#L83 for context, as we might want to look into modifying the default retry policy for this resource

erezrokah:
We released https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-azure-v10.1.0 that allows configuring the Azure SDK retry options to help with this issue

PR #14745, opened 2023-10-19
feat: Expose Azure Go SDK retry configuration

erezrokah:
#### Summary

Resolves https://github.com/cloudquery/cloudquery/issues/9860
Resolves https://github.com/cloudquery/cloudquery/issues/11225

For some Azure APIs the default retry options don't cut it. This PR exposes the retry configuration globally.
We could also expose per table but that seems a bit too much.

### TODO
 
- [x] Generate docs


website/pages/docs/plugins/sources/azure/_configuration.md
```diff
@@ -16,4 +16,9 @@ spec:
     # skip_subscriptions: []
     # normalize_ids: false
     # oidc_token: ""
+    # retry_options:
+    #   max_retries: 3
+    #   try_timeout_seconds: 0
+    #   retry_delay_seconds: 4
+    #   max_retry_delay_seconds: 60
```
erezrokah, line 23: Did not add `status_codes` as I think that's quite an advanced option and also I'd have to hard code the default codes from the Azure SDK

plugins/source/azure/client/spec/spec.go
```diff
@@ -42,6 +42,17 @@ type Spec struct {
...
+	// Retry options to pass to the Azure Go SDK, see more details here https://github.com/Azure/azure-sdk-for-go/blob/f951bf52fb68cbb978b7b95d41147693c1863366/sdk/azcore/policy/policy.go#L86
+	RetryOptions *RetryOptions `json:"retry_options"`
+}
+
+type RetryOptions struct {
+	MaxRetries           *int32 `json:"max_retries"`
+	TryTimeoutSeconds    *int   `json:"try_timeout_seconds" jsonschema:"minimum=0"`
+	RetryDelaySeconds    *int   `json:"retry_delay_seconds" jsonschema:"minimum=0"`
+	MaxRetryDelaySeconds *int   `json:"max_retry_delay_seconds" jsonschema:"minimum=0"`
+	StatusCodes          *[]int `json:"status_codes" jsonschema:"minItems=1,uniqueItems=true"`
```
hermanschaaf, line 55: Have you validated / do we have tests to check that this json schema spec allows `nil` to be specified?

hermanschaaf (approved):

plugins/source/azure/client/spec/spec.go
erezrokah: Yes, it allows it

erezrokah: Let me check about adding tests for it

erezrokah: Ah, OK so there are existing tests that don't set it in https://github.com/cloudquery/cloudquery/blob/bb141c8baaa9674df82cb4d70476c0e7258f9f36/plugins/source/azure/client/spec/spec_test.go#L185 I'll add some more

hermanschaaf: Cool, yeah I'm specifically interested in the `StatusCodes` case, because it's not clear from the schema that it allows nil to be specified, but maybe it's implied.

plugins/source/azure/client/spec/spec_test.go
```diff
@@ -198,5 +198,54 @@ func TestSpec_JSONSchemaExtend(t *testing.T) {
 			Name: "proper discovery_concurrency",
 			Spec: `{"discovery_concurrency":123}`,
 		},
+		{
+			Name: "proper retry_options.max_retries",
```
candiduslynx, line 202: should probably be tested separately

Issue #14725, opened 2023-10-19
bug: `backend_options` documented under nested plugin specs


candiduslynx:
See https://github.com/cloudquery/cloudquery/pull/12686
We need to move `backend_options` only under the top-level spec & just mention them where we know the plugin supports incremental.

candiduslynx:
cc: @erezrokah @hermanschaaf

candiduslynx:
Proposed solution:
1. Remove `backend_options` documentation from underneath the source plugins nested spec
2. Fix example in [`proto/v3` blog](https://www.cloudquery.io/blog/proto-v3)
3. Add a section regarding incremental syncs like in [AWS source plugin](https://www.cloudquery.io/docs/plugins/sources/aws/configuration#incremental-tables) to the plugins that support this
4. Link the [Managing Incremental Tables](https://www.cloudquery.io/docs/advanced-topics/managing-incremental-tables) doc to blog & plugin docs.

PR #14767, opened 2023-10-20
doc(website): Proper `backend_options` documentation

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/14725


website/pages/docs/plugins/sources/homebrew/_configuration.md
```diff
@@ -8,8 +8,8 @@ spec:
   tables: ["*"]
   destinations:
     - "DESTINATION_NAME"
+  backend_options:
+    table_name: "test_state_table"
```
hermanschaaf, line 12: Bit of a nit, but can we maybe call it something other than `test_state_table` (here and in the other configs). Maybe `homebrew_state_table`?

erezrokah (approved):

website/pages/docs/plugins/sources/homebrew/_configuration.md
candiduslynx: we have only a single place where we did this, so I left others as is

candiduslynx: Renamed to `cq_state_<plugin>` in 036cfc5

hermanschaaf: Ah yeah I see, thanks for the fix!

Issue #14699, opened 2023-10-19
bug: No url configuration found for aws lambda functions without aliases


alanjose10:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, the `url_config` column is in the `aws_lambda_function_aliases` table. But this is causing an issue since there can be lambda functions with URLs without there being an alias for it. Hence, I are not able to view the url configuration of some lambda functions since they do not have aliases.

### Expected Behavior

Maybe this column can be moved to the `aws_lambda_functions` table? Or a new table named `aws_lambda_function_url_configurations` can be created?

### CloudQuery (redacted) config

na

### CloudQuery (redacted) logs

na

### CloudQuery version

3.13.1

### Pull request (optional)

- [ ] I can submit a pull request

candiduslynx:
@alanjose10 `aws_lambda_function_url_configs` support is released in [`v22.18.0`](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-aws-v22.18.0)

PR #14913, opened 2023-10-24
feat: Add `aws_lambda_function_url_configs` table

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/14699

I wonder if we want to drop the `url_config` column in `aws_lambda_function_aliases` table (pre-resource resolver, slower), too. This will make this change a breaking one, though. Can be done in a separate PR.


hermanschaaf (approved):
@candiduslynx Maybe let's add it to the list of quality improvements that would be breaking here: https://github.com/cloudquery/cloudquery/issues/12648

PR #14920, opened 2023-10-24
chore: Fix docs publishing for DuckDB, SQLite, Snowflake

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/14484. This fixes docs publishing for plugins with `CGP_ENABLED=1`


.github/workflows/publish_plugin_to_hub_duckdb.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     outputs:
       plugin_name: ${{ steps.split.outputs.plugin_name }}
       plugin_kind: ${{ steps.split.outputs.plugin_kind }}
+      plugin_kind_plural: ${{ steps.split.outputs.plugin_kind_plural }}
```
hermanschaaf, line 12: What do we need the plural form for?

erezrokah: The Website directories are in the format `docs/plugins/sources` and `docs/plugins/destinations`, and we copy the docs from them.
I could hard code the values too since this workflow is scoped to a specific plugin, but that actually makes it harder to sync with https://github.com/cloudquery/cloudquery/blob/49ab5669d0bef1ab3ce907238feb2b9664a4b802/.github/workflows/publish_plugin_to_hub.yml

hermanschaaf (approved):

PR #14922, opened 2023-10-24
feat!: Support creating & recreating indexes for tables with primary keys

### LINKED ISSUE #13754: bug: Slow load times to Neo4j destination due to lack of indexing

### Current Behavior

Example: `aws_cloudtrail_events`
Big table with lots of data.

Loading to Neo4j starts off really quick (couple of seconds to load thousands of events) then speed exponentially decreases.
The more events have been loaded, the more time it takes to load.

This happens because CQ uses `MERGE` which creates a node if it doesn't already exist in the DB based on primary keys defined.
In [write.go](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/neo4j/client/write.go) there isn't a functionality for creating an index on the primary keys defined.
Without indexes, Neo4j does a full scan of all nodes to determine whether the entity in question exists.
This causes the exponential decrease in performance.
This would impact any table from any source that contains a lot of data.

### Expected Behavior

Before starting the load, CQ would create indexes on the primary keys of the node.
This could be done in write.go or a separate file.
Example: 
```
CREATE INDEX IF NOT EXISTS FOR (n:aws_cloudtrail_events) ON (n.event_id);
```

```
kind: source
spec:
  name: "aws"
  registry: "github"
  path: "cloudquery/aws"
  version: "v22.1.0"
  tables: ["*"]
  skip_tables:
    - aws_ec2_vpc_endpoint_services
    - aws_docdb_cluster_parameter_groups
    - aws_docdb_engine_versions
    - aws_ec2_instance_types
    - aws_elasticache_engine_versions
    - aws_elasticache_parameter_groups
    - aws_elasticache_reserved_cache_nodes_offerings
    - aws_elasticache_service_updates
    - aws_elasticsearch_versions
    - aws_neptune_cluster_parameter_groups
    - aws_neptune_db_parameter_groups
    - aws_rds_cluster_parameters
    - aws_rds_cluster_parameter_groups
    - aws_rds_cluster_parameter_group_parameters
    - aws_rds_db_parameter_groups
    - aws_rds_engine_versions
    - aws_servicequotas_services
    - aws_servicequotas_quotas
    - aws_iam_role_last_accessed_details
    - aws_iam_user_last_accessed_details
    - aws_iam_policy_last_accessed_details
    - aws_directconnect_locations
    - aws_ram_resource_types
    - aws_lambda_runtimes
    - aws_docdb_event_categories
  destinations: ["neo4j"]
  spec:
    accounts:
      - id: "${ACCOUNT_ID}"
        role_arn: "${ROLE_ARN}"
        role_session_name: "Discovery"
    aws_debug: false

kind: destination
spec:
  name: "neo4j"
  registry: "github"
  path: "cloudquery/neo4j"
  version: "v4.0.1"
  # batch_size: 10000 # optional
  # batch_size_bytes: 5242880 # optional
  spec:
    connection_string: "${DB_ENDPOINT}"
    username: "${DB_USERNAME}"
    password: "${DB_PASSWORD}"
```

### Steps To Reproduce

_No response_

```
N/A
```

### CloudQuery version

v3.8.0

### Additional Context

[Neo4j Indexing](https://neo4j.com/docs/cypher-manual/current/query-tuning/indexes/)

### Pull request (optional)

candiduslynx:
Closes #13754

BEGIN_COMMIT_OVERRIDE
feat: Support creating & recreating indexes for tables with primary keys (#14922), closes (#13754)
BREAKING-CHANGE: Support creating & recreating indexes for tables with primary keys (#14922). CloudQuery Neo4j plugin will create or recreate indexes for tables with primary keys. This feature is marked breaking, as `CREATE INDEX` & `DROP INDEX` privileges are now required.
END_COMMIT_OVERRIDE


erezrokah (approved):
Let's do a breaking change as I'm not sure what the impact of this will be

plugins/destination/neo4j/client/write.go
```diff
@@ -20,7 +20,7 @@ func (c *Client) WriteTableBatch(ctx context.Context, tableName string, msgs mes
 		return err
 	}
 
-	session := c.LoggedSession(ctx, neo4j.SessionConfig{})
+	session := c.Session(ctx, neo4j.SessionConfig{AccessMode: neo4j.AccessModeWrite})
```
bbernays, line 23: What is the default access mode? does this require users to potentially change credentials or permissions?

candiduslynx: Default is `0` = `write`:
https://pkg.go.dev/github.com/neo4j/neo4j-go-driver/v5@v5.13.0/neo4j#AccessMode

candiduslynx: The change, however, requires users to have `CREATE INDEX` & `DROP INDEX` privileges

candiduslynx:
> Let's do a breaking change as I'm not sure what the impact of this will be

I'll wait for the acknowledgement (& maybe test) from the issue author, but I also think the major bump is required here.

erezrokah:
> I'll wait for the acknowledgement (& maybe test) from the issue author, but I also think the major bump is required here.

Maybe we can tag a release candidate and post the version on the issue?

bbernays:
I think we can test this out ourselves by syncing the `cloudtrail_events` table from our org

candiduslynx:
> I think we can test this out ourselves by syncing the `cloudtrail_events` table from our org

I've tested with `aws_rds_cluster_parameters` instead.

The resulting speed is very good:
* pre: 30K rows for 15+ minutes, stopped
* post: 97K resources get written in a couple of minutes (but there's some lagging pattern, maybe caused by flushing everything onto disk, that gets me hanging for several minutes more)

PR #15000, opened 2023-10-26
feat: Pass auth token to download client if needed

erezrokah:
#### Summary

Goes with https://github.com/cloudquery/cloudquery-issues/issues/737.
Opening in draft since I need to extract some refactoring to make the diff smaller

PR #15001, opened 2023-10-26
fix: Add missing options to install command

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/15000

Added some missing options to the `install` command.
We could probably also do `managedplugin.WithDirectory(cqDir)` directly without the condition as `cqDir` has a default so the only way it can be an empty string is when someone does `--cq-dir ""` (I tested that and it sets `cqDir` to an empty string).
Not doing the `cqDir` fix as that could be considered a breaking change fix (doing `--cq-dir ""` with the current code will use the default `.cq`)

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -49,9 +49,15 @@ func installPlugin(cmd *cobra.Command, args []string) error {
 	}
 	sources := specReader.Sources
 	destinations := specReader.Destinations
-	opts := []managedplugin.Option{managedplugin.WithNoExec()}
+	opts := []managedplugin.Option{
+		managedplugin.WithNoExec(),
+		managedplugin.WithLogger(log.Logger),
+	}
 	if cqDir != "" {
 		opts = append(opts, managedplugin.WithDirectory(cqDir))
+	}
+	if disableSentry {
+		opts = append(opts, managedplugin.WithNoSentry())
 	}
 
 	sourcePluginConfigs := make([]managedplugin.Config, 0, len(sources))


cli/cmd/install.go
```diff
@@ -49,9 +49,15 @@ func installPlugin(cmd *cobra.Command, args []string) error {
...
 	sources := specReader.Sources
 	destinations := specReader.Destinations
-	opts := []managedplugin.Option{managedplugin.WithNoExec()}
+	opts := []managedplugin.Option{
+		managedplugin.WithNoExec(),
+		managedplugin.WithLogger(log.Logger),
+	}
 	if cqDir != "" {
-		opts = append(opts, managedplugin.WithDirectory(cqDir))
+		managedplugin.WithDirectory(cqDir)
```
disq, line 57: This is a no-op?

disq: Isn't it this?
```go
func WithDirectory(directory string) Option {
	return func(c *Client) {
		c.directory = directory
	}
}
```
How does this work if we're not using `Option`?

erezrokah: There's a default https://github.com/cloudquery/plugin-pb-go/blob/f92d39c67536f41473abef7d6d9ebc30015c393e/managedplugin/plugin.go#L133

erezrokah:
Looking at the tests failures

cli/cmd/install.go
erezrokah: That's for the Slack discussion I misunderstood you. I updated the PR

disq (approved):

PR #15012, opened 2023-10-26
feat: Updated query s3_cross_region_replication.sql

ronsh12:


plugins/source/aws/policies/queries/s3/s3_cross_region_replication.sql
```diff
@@ -1,18 +1,40 @@
...
+	from
+	aws_s3_buckets
+),
+sec as (
+	select bt.arn,
+		bt.region,
+		bt.dest,
+		asb.region as other,
+		bt.stat
+		from basic_table as bt
```
bbernays, line 16: Can we use more descriptive name here as this will be in the public documentation

plugins/source/aws/policies/queries/s3/s3_cross_region_replication.sql
```diff
@@ -1,18 +1,40 @@
+WITH basic_table as (
+select 
+	arn,
+	region,
+	jsonb_array_elements(replication_rules) -> 'Destination' ->> 'Bucket' as dest,
+	jsonb_array_elements(replication_rules) ->> 'Status' as stat
+	from
+	aws_s3_buckets
+),
+sec as (
```
bbernays, line 10: What is `sec`?

bbernays (changes requested):
The logic looks good, but I think the subquery names could be a bit more descriptive so it is easier to understand.

bbernays (approved):

PR #15070, opened 2023-10-31
chore: Use large runners only for AWS, GCP, Azure CI workflows

erezrokah:
#### Summary

Same as https://github.com/cloudquery/cloudquery/pull/15069 not from a fork so workflows that require secrets can pass


hermanschaaf (approved):

erezrokah:
Wait for all fails since this PR's commit is the same as the closed one https://github.com/cloudquery/cloudquery/pull/15069

PR #15086, opened 2023-11-01
chore: Fix docs publishing for DuckDB, Snowflakse, SQLite

erezrokah:
#### Summary

We were still using stub docs for these plugins. This PR fixes it


erezrokah:
Going to merge so I can fix these plugins in the Hub

PR #15088, opened 2023-11-01
chore: Use relative link in product updates, exclude www.g2.com

erezrokah:
#### Summary

Fixes the broken links check


.github/workflows/broken_links_scheduled.yml
```diff
@@ -54,6 +54,7 @@ jobs:
             --exclude regex101.com \
             --exclude help.shopify.com/en/manual/apps/app-types/custom-apps \
             --exclude hub.cloudquery.io \
+            --exclude www.g2.com \
```
erezrokah, line 57: Seems like they have a bot protection that fails our check

website/pages/blog/product-updates-9.mdx
```diff
@@ -52,9 +52,9 @@ Join us for a live chat on November 29, at 10 AM EST (3 PM UTC). We will talk ab
 
 ## Blog
 
-- We have launched the **Hub**! Read about what it is in our [announcement](https://cloudquery.io/blog/announcing-cloudquery-new-hub).
```
erezrokah, line 55: This link creates a redundant redirect and also opens in a new tab

hermanschaaf (approved):

PR #15090, opened 2023-11-01
chore: Fix docs publishing cross-compile image

erezrokah:
#### Summary

Seems like `rsync` [doesn't work as expected](https://github.com/cloudquery/cloudquery/actions/runs/6719832688/job/18262200668#step:9:11) on the cross compile image so trying a different approach


erezrokah:
can't really test this locally since https://github.com/nektos/act fails to download the cross compile image

PR #15106, opened 2023-11-01
chore: Add latest version to public paid plugins

erezrokah:
#### Summary

Part of cloudquery/cloudquery-issues#768 (internal issue). Adds the latest versions for the paid public plugins, see https://hub.cloudquery.io/plugins/source?tiers=paid.

I also removed the buy links from the configuration comments as I think the CLI should guide users to login when using these plugins.

Once this PR is merged we can publish new versions of the plugins with the new docs.

**Not in this PR** - latest versions in https://www.cloudquery.io/docs/plugins/sources/overview as it requires a bigger change to get the versions in the React component here https://github.com/cloudquery/cloudquery/blob/0971c8a49c6415eb9330f4d078f5ab7a37bd78ae/website/components/PluginsTable.tsx#L32

### DIFF
--- a/website/pages/docs/plugins/sources/file/overview.md
+++ b/website/pages/docs/plugins/sources/file/overview.md
@@ -6,7 +6,7 @@ description: CloudQuery File source plugin documentation
 ---
 # File Source Plugin
 
-:badge{text="Premium"}
+:badge
 
 This is a premium plugin that you can buy [here](/integrations/file).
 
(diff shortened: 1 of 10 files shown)


website/pages/docs/plugins/sources/file/overview.md
```diff
@@ -6,7 +6,7 @@ description: CloudQuery File source plugin documentation
 ---
 # File Source Plugin
 
-:badge{text="Premium"}
```
erezrokah, line 9: No attributes defaults to using the latest version

yevgenypats (approved):
Looks good. I don't think we should do any changes in the old docs as we are going to remove them shortly.

erezrokah:
> Looks good. I don't think we should do any changes in the old docs as we are going to remove them shortly.

These docs end up in https://github.com/cloudquery/cloudquery-private/tree/main and used in the Hub

PR #15136, opened 2023-11-03
doc: Containers with read-only file system

pilvikala:
Adding a documentation on how to build a container with a readonly file system


website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
...
+  name: xkcd
+  path: hermanschaaf/xkcd
+  version: v2.0.0
+  destinations: ["postgresql"]
+  tables: ["*"]
+---
+kind: destination
+spec:
+  name: "postgresql"
+  registry: "github"
```
hermanschaaf, line 37:
```suggestion
  registry: "cloudquery"
```

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
...
+  path: "cloudquery/postgresql"
+  version: "v6.1.3"
```
hermanschaaf, line 39:
```suggestion
  version: "VERSION_DESTINATION_POSTGRESQL"
```

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
...
+- a start script used as an entry point for the container that will start the plugins and run the sync
+
+In the example below, we will use the [XKCD](https://github.com/hermanschaaf/cq-source-xkcd) source plugin and [PostgreSQL](https://hub.cloudquery.io/plugins/destination/cloudquery/postgresql/v6.1.3/docs) destination plugin.
+
+### Create a config file for the installation
+
+This config file should reference the plugins that you want to install with the actual source where they should be downloaded from:
+
+```yaml
+# install.yaml
```
hermanschaaf, line 25:
```suggestion
```yaml copy filename="install.yaml"
```

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
...
+  spec:
+    connection_string: "postgresql://cq:password123@host.docker.internal:5432/cq-test?sslmode=disable"
+```
+
+### Create a config file for the sync
+
+This config file should reference the plugins running as gRPC servers (fill in the `connection_string`). We will run XKCD on port 7778 and PostgreSQL on port 7779:
+
+```yaml
+# sync.yaml
```
hermanschaaf, line 49:
```suggestion
```yaml copy filename="sync.yaml"
```

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
...
+  tables: ["*"]
+---
+kind: destination
+spec:
+  name: "postgresql"
+  registry: "github"
+  path: "cloudquery/postgresql"
+  version: "v6.1.3"
+  spec:
+    connection_string: "postgresql://cq:password123@host.docker.internal:5432/cq-test?sslmode=disable"
```
hermanschaaf, line 41: Is the connection string actually needed here? I think you should be able to leave it empty when using with the `cloudquery plugin install` command. If not, we should file a bug for it

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+## Run the Container
+
+First, build the container as you would normally do:
+
+```bash
+docker build ./ -t my-cq-container:latest
+```
+
+Run the container with `--read-only` option and a command to pass to our start script. You will also need to add the `--no-log-file` option to skip logging to a file. Here is an example:
```
hermanschaaf, line 112: I would suggest rather using `--log-console` so that you can still access the logs at least

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,122 @@
+---
+title: Docker - Read-only Containers
+description: Learn how to run containers with read-only file system.
```
hermanschaaf, line 3:
```suggestion
description: Learn how to run containers with a read-only file system.
```

hermanschaaf (approved):
Very nice tutorial! I left a few non-blocking comments

website/pages/docs/deployment/docker-read-only.mdx
pilvikala: no, I'll fix it

website/pages/docs/deployment/docker-read-only.mdx
pilvikala: actually, `--no-log-file` is still required. I also didn't see a difference in the log when used the `--log-console`

hermanschaaf: Ah okay yeah seems like `--log-console` results in logs being written to both file and console. So in that case I'd suggest using both: `--log-console --no-log-file` so that the output would only go to stdout/stderr

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,120 @@
+---
+title: Docker - Read-only Containers
+description: Learn how to run containers with a read-only file system.
+---
+
+# Docker - Read-only Containers
+
+By default, CloudQuery requires a read-write file system to create a socket file used by the plugins to communicate with the server. It is however not possible to create a read-write socket on a read-only file system. For this to work, a plugin needs to be started as a gRPC server locally.
```
bbernays, line 8:
```suggestion
By default, CloudQuery requires a read-write file system to create a socket file used by the plugins to communicate with the cli. It is however not possible to create a read-write socket on a read-only file system. For this to work, a plugin needs to be started as a gRPC server locally.
```

website/pages/docs/deployment/docker-read-only.mdx
```diff
@@ -0,0 +1,120 @@
+---
+title: Docker - Read-only Containers
+description: Learn how to run containers with a read-only file system.
+---
+
+# Docker - Read-only Containers
```
bbernays, line 6: Can we call this Something Like `Containers- Read Only Root Filesystem`

Because that is the exact wording the users used and it is how AWS and other security vendors refer to this 
https://docs.aws.amazon.com/config/latest/developerguide/ecs-containers-readonly-access.html

bbernays (commented):
Looks great a few comments

website/pages/docs/deployment/docker-read-only.mdx
pilvikala: renamed

PR #15124, opened 2023-11-02
chore(website): Redirect old plugin docs to hub

yevgenypats:
Part of migration to the cloudquery hub documentation: https://github.com/cloudquery/cloudquery-issues/issues/773

PR #15137, opened 2023-11-03
chore: Migrating mysql source plugin to premium paid

mnorbury:


mnorbury:
> Looks great 🚀 I think we can remove
> 
> https://github.com/cloudquery/cloudquery/blob/4d2da14f655eafdff799c84fb049a3e5fa3cb89d/.github/pr_labeler.yml#L32
> 
> too.
> Don't change anything under `website/`. We should handle it in a similar way to other premium closed source plugins via #15124

👍 The source labeler not the destination though.

erezrokah (approved):

erezrokah:
Once this is merged we should close https://github.com/cloudquery/cloudquery/pull/14981

Issue #14796, opened 2023-10-20
bug: `failed to sync records: failed to sync unmanaged client: parsing time "111111-11-11 11:11:00" as "2006-01-02 15:04:05.999999999": cannot parse "11-11-11 11:11:00" as "-"`


nicosuave:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Postgres -> DuckDB sync is failing. I don't believe Postgres can store such a value as `111111-11-11 11:11:00` in a timestamp field so am a little unclear of what's happening.

### Expected Behavior

No failure

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: ...
  registry: "github"
  path: "cloudquery/postgresql"
  version: "v3.0.6"
  destinations: ["motherduck"]
  tables: ['*']
  spec:
    rows_per_record: 10000
    pgx_log_level: debug
    connection_string: ...
  backend_options:
    table_name: "cq_..._state"
    connection: "@@plugins.motherduck.connection"
---
kind: destination
spec:
  name: motherduck
  path: cloudquery/duckdb
  version: "v5.0.3"
  write_mode: overwrite
  migrate_mode: forced
  spec:
    connection_string: ...
    batch_size: 10000
    debug: true
```

### CloudQuery (redacted) logs

In title. `debug` option also not seeming to do anything as far as surfacing more detailed logs around the failure

### CloudQuery version

cloudquery version 3.9.0

### Pull request (optional)

- [ ] I can submit a pull request

hermanschaaf:
@nicosuave Can you share some more information about the tables being synced from postgres? Especially their column types; if there are in time-related columns there I suspect that could get us closer to the root cause of the issue, but I will admit the error is quite strange :)

nicosuave:
@hermanschaaf sure— I struggled with the `debug` option on the duckdb destination; am I using that correctly? I didn't see any additional output.

Yeah, my thoughts were any time-like columns in Postgres would in effect reject that invalid data upstream, so am also confused. Will dig through the tables that have been partially copied over and go from there.

candiduslynx:
@nicosuave Could you try with `postgresql` [`v4.0.0`](https://hub.cloudquery.io/plugins/source/cloudquery/postgresql/v4.0.0)? This has some fix for `time` columns, which might help.

Note that it has been moved to CloudQuery Hub & you would also need to use `cloudquery login` or create an API key.

[...]

PR #15139, opened 2023-11-03
fix: Use Arrow marshaling code for `arrow.Time32` & `arrow.Time64` types

candiduslynx:
Should help with https://github.com/cloudquery/cloudquery/issues/14796


erezrokah (approved):
Looks good. Is there a way to test this?

candiduslynx:
@erezrokah we have the tests for time32/time64 in our test cases for this plugin, however, no additional test came to my mind.
This doesn't necessarily fix the issue as the timestamps produced there might actually be caused by `arrow.TimestampType` stuff, but this is a first attempt of understanding what might've went wrong.

PR #15142, opened 2023-11-03
chore: Change MySQL to premium plugin

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/15137


mnorbury (approved):
Whoops 🙇

PR #15144, opened 2023-11-03
chore: Migrate postgres source to premium

mnorbury:
Migrating postgres to premium plugins.

Same as: https://github.com/cloudquery/cloudquery/pull/15137


erezrokah (approved):

dylanschoenmakers:
Wow this is suprising. We were using this source, why would you migrate this completely and not let a community source exists?

pilvikala:
Hey @dylanschoenmakers, the maintenance of this plugin requires a substantial effort. We have decided to make it a premium option and while the current version remains open and free, further development will be closed-source and available for installation via the [Hub](https://hub.cloudquery.io/).

The old versions of the plugin will remain on GitHub and it will be possible to configure CloudQuery to download the plugins from there by setting registry: github in the config. Similarly, the release tags remain so you can pull the source code.

This PR is removing the source code so it does not become stale.

yevgenypats:
Hi @dylanschoenmakers sorry for the surprise effect here. To add to that it's only those plugins which were in preview (full announcement here - https://www.cloudquery.io/blog/announcing-premium-plugins). Source is still available in the git and if someone wants to maintain and fork it - we would be more than happy and we will also publish the free version on our hub (It's actually already possible to so now self-serve via cloud.cloudquery.io and publish to hub.cloudquery.io ). We were just the sole maintainers of this plugin and given we already maintain almost 20 free plugins we couldn't fund additional free plugins ourselves.

dylanschoenmakers:
Hi @pilvikala and @yevgenypats, we were just starting out using cloudquery and since we have a couple of postgres destinations and sources we would like to test it out further. A colleague of mine actually just recently opened an issue about the date parsing so this came at an inconvenient time where the fix was published. We will have no problem with purchasing premium once we take this into production. 

In any case, thanks for the effort!

yevgenypats:
Hi @dylanschoenmakers, That sounds good - you can try out the postgresql premium plugin https://hub.cloudquery.io/plugins/source/cloudquery/postgresql/v4.0.0/docs for free as it's free in the next month (unlimited quota while in preview) and then free up to a quota so testing is still possible after that for small scale. Please let me know if you run into any issues.

PR #15154, opened 2023-11-05
feat: Updated view api_gateway_method_settings

ronsh12:
When a new instance is created in the API Gateway, the default value for the method_settings becomes an empty JSON object ({}). When the JSONB_EACH_TEXT function is used, it doesn't expand the JSON data at all, and it remains empty.
As it's a JSON object, attempts to expand it using jsonb_array_elements do not work.
best approach is to handle this case specifically with checking the method_settings column and creating default values if needed.


bbernays:
Is this a `feat` or a `fix`?

erezrokah:
Closing as stale. Please re-open if not the case

PR #15166, opened 2023-11-06
feat: Add `addon download` command

disq:
```
cloudquery addon download addon-team/addon-type/addon-name@v1.0.0
cloudquery addon download addon-team/addon-type/addon-name@v1.0.0 -t /path/to/dest/dir
cloudquery addon download addon-team/addon-type/addon-name@v1.0.0 -t - | jar t
```

(Apparently the regular infozip `unzip` can't read from stdin? but `jar` can...)

### DIFF
--- a/cli/cmd/addon_download.go
+++ b/cli/cmd/addon_download.go
@@ -99,17 +99,15 @@ func runAddonDownload(ctx context.Context, cmd *cobra.Command, args []string) er
 func getAddonMetadata(ctx context.Context, c *cloudquery_api.ClientWithResponses, teamName, addonType, addonName, version string) (*cloudquery_api.Addon, *cloudquery_api.AddonVersion, error) {
 	addonResp, err := c.GetAddonWithResponse(ctx, teamName, cloudquery_api.AddonType(addonType), addonName)
 	if err != nil {
-		if addonResp == nil {
-			return nil, nil, fmt.Errorf("failed to get addon: %w", err)
-		}
+		return nil, nil, fmt.Errorf("failed to get addon: %w", err)
+	} else if addonResp.StatusCode() != http.StatusOK {
 		return nil, nil, fmt.Errorf("failed to get addon: %w", errorFromHTTPResponse(addonResp.HTTPResponse, addonResp))
 	}
 
 	addonVersionResp, err := c.GetAddonVersionWithResponse(ctx, teamName, cloudquery_api.AddonType(addonType), addonName, version)
 	if err != nil {
-		if addonVersionResp == nil {
-			return nil, nil, fmt.Errorf("failed to get addon version: %w", err)
-		}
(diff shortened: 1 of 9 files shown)


cli/cmd/addon_download.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create hub client: %w", err)
+	}
+
+	target, err := cmd.Flags().GetString("target")
+	if err != nil {
+		return err
+	}
+
+	addon, addonVersion, err := getAddonMetadata(ctx, c, addonParts[0], addonParts[1], addonVer[0], addonVer[1])
```
disq, line 91: We need to make two separate calls before attempting a download, because we don't know about the addon format (file extension) or this version's checksum beforehand.

cli/cmd/addon_download.go
```diff
@@ -0,0 +1,180 @@
...
+			return fmt.Errorf("failed to create file: %w", err)
+		}
+		fileWriter = f
+
+		defer func() {
+			if retErr != nil {
+				_ = os.Remove(zipPath)
+				return
+			}
+			fmt.Fprintf(os.Stderr, "Wrote %d bytes to %s\n", size, zipPath)
```
disq, line 157: Closing `os.Stdout` takes care of not writing anything else, but we still want to give some feedback to the user.

cli/cmd/addon_download.go
disq: In the future we should look into amending the http headers (I'm not sure if it works alongside 302/Location or we might look into setting them in the bucket object) to carry the filename and expected checksum so we don't need to query these at all.

website/pages/docs/reference/cli/cloudquery_addon_download.md
```diff
@@ -0,0 +1,49 @@
...
+---
+## cloudquery addon download
+
+Download addon from CloudQuery Hub.
+
+### Synopsis
+
+Download addon from CloudQuery Hub.
+
+This downloads an addon from CloudQuery Hub to local disk.
```
pilvikala, line 12: You could add a link (or links) so people would know where to find the addons: https://hub.cloudquery.io/addons/transformation, https://hub.cloudquery.io/addons/visualization

disq: @pilvikala 
There are no hub links so far in any of the cli commands, if we're going to do that maybe we should amend all our "cli frontend" with links? The only link in the cli is in the main help output, and goes like:

```
CloudQuery CLI

Open source data integration at scale.

Find more information at:
	https://www.cloudquery.io
```

hermanschaaf (approved):
LGTM!

(I just pushed a commit to fix a panic in the case if you're not logged in or are not authorized to view the addon)

yevgenypats (approved):

PR #15173, opened 2023-11-07
fix: Better addon downloads

disq:
Get addon downloads to work without making 2 extra api requests but still verifying checksum.
~Needs https://github.com/cloudquery/cloud/pull/660 to be solved/merged first (internal issue)~

Also updates plugin-pb-go to verify the plugin downloads from the hub ([PR](https://github.com/cloudquery/plugin-pb-go/pull/158))

~We might keep this out and use the AddonWithTeam endpoint which is going to be implemented in the backend soon.~


cli/cmd/addon_download.go
```diff
@@ -74,7 +75,16 @@ func runAddonDownload(ctx context.Context, cmd *cobra.Command, args []string) er
 		return fmt.Errorf("invalid addon ref %q: version must start with 'v'", args[0])
 	}
 
+	checksum := ""
+	hc := *http.DefaultClient
+	hc.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		if len(via) == 1 && req.Response != nil {
+			checksum = req.Response.Header.Get("X-Checksum-Sha256") // get checksum from first request
```
disq, line 82: This is a bit weird: `via[0].Response` is supposed to be `nil` for some reason, _that_ response is available in `req.Response`

cli/cmd/addon_download.go
```diff
@@ -130,16 +110,21 @@ func downloadAddon(ctx context.Context, c *cloudquery_api.ClientWithResponses, t
...
 		size       int64
+		err        error
 	)
 
 	switch targetDir {
 	case "-":
 		fileWriter = os.Stdout
 	default:
-		zipPath := filepath.Join(targetDir, addonFilename(teamName, addonType, addonName, version, format))
+		zipPath := filepath.Join(targetDir, path.Base(res.Request.URL.Path))
```
disq, line 127: Here we infer the filename from the final URL.

cli/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/apache/arrow/go/v14 v14.0.0-20231031200323-c49e24273160
 	github.com/bradleyjkemp/cupaloy/v2 v2.8.0
 	github.com/cenkalti/backoff/v4 v4.2.1
-	github.com/cloudquery/cloudquery-api-go v1.4.3
+	github.com/cloudquery/cloudquery-api-go v1.4.4-0.20231110172139-6b00213353ac
```
disq, line 9: Use 1.4.4 when we do the same thing with plugin downloads

cli/go.mod
```diff
@@ -6,8 +6,8 @@ require (
 	github.com/apache/arrow/go/v14 v14.0.0-20231031200323-c49e24273160
 	github.com/bradleyjkemp/cupaloy/v2 v2.8.0
 	github.com/cenkalti/backoff/v4 v4.2.1
-	github.com/cloudquery/cloudquery-api-go v1.4.3
-	github.com/cloudquery/plugin-pb-go v1.13.5
+	github.com/cloudquery/cloudquery-api-go v1.4.4
+	github.com/cloudquery/plugin-pb-go v1.13.6-0.20231113121727-245cf96ea3ea
```
disq, line 10: using untagged for now

cli/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/apache/arrow/go/v14 v14.0.0-20231031200323-c49e24273160
 	github.com/bradleyjkemp/cupaloy/v2 v2.8.0
 	github.com/cenkalti/backoff/v4 v4.2.1
-	github.com/cloudquery/cloudquery-api-go v1.4.6
+	github.com/cloudquery/cloudquery-api-go v1.5.1
```
disq, line 9: We now use `1.5.x` to infer if we need to use the /teams/* endpoint or not.

erezrokah (approved):

PR #15177, opened 2023-11-07
feat: Addon publishing command to take directory input

disq:
Better `cloudquery addon publish`

> cloudquery addon publish /path/to/manifest.json /path/to/addon-dir v1.0.0

This way every addon archive can have the `cloudquery_visualization_test_v1.2.3/` type wrapper directory.
We also add the version as a zip comment: `cloudquery/visualization/test@v1.2.3`


cli/cmd/addon_publish.go
```diff
@@ -246,3 +249,48 @@ func readManifestJSON(manifestPath string) (ManifestJSONV1, error) {
 	}
 	return manifest, nil
 }
+
+func zipAddon(addonPath, wrapDirectory, comment string) (string, string, error) {
+	s := sha256.New()
+	zipFile, err := os.CreateTemp("", "cq-addon*.zip")
```
disq, line 255: In theory we could just zip it into a io.Writer and upload it directly (while at the same time calculating checksums) but then the `checksum` field would need to be updated with a separate API call (or left empty if the user doesn't call `--finalize`).

yevgenypats (changes requested):
Might missed the previous one but it seemed to me that the previous command worked well. what didn't work?

disq:
> Might missed the previous discussion but it seemed to me that the previous command worked well. what didn't work?

@yevgenypats explained in the linked private issue

PR #15178, opened 2023-11-07
feat: Add `aws_budgets` & `aws_budget_actions` tables

candiduslynx:
Closes https://github.com/cloudquery/cloudquery-issues/issues/813


plugins/source/aws/resources/services/budgets/budgets.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func fetchBudgets(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services(client.AWSServiceBudgets).Budgets
+	input := &budgets.DescribeBudgetsInput{AccountId: aws.String(cl.AccountID)}
+
+	p := budgets.NewDescribeBudgetsPaginator(svc, input, func(options *budgets.DescribeBudgetsPaginatorOptions) {
+		options.StopOnDuplicateToken = true
```
bbernays, line 32: Is this an option that is unique for this service?

plugins/source/aws/resources/services/budgets/budgets.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/budgets"
+	"github.com/aws/aws-sdk-go-v2/service/budgets/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+)
+
+func Budgets() *schema.Table {
+	tableName := "aws_budgets"
```
bbernays, line 15: Why is this not `aws_budget_budgets`? In GCP we have `gcp_workflows_workflows`. I think it is probably better to be consistent in our naming rather than try and make it the most readable/understandable

bbernays (changes requested):

plugins/source/aws/resources/services/budgets/budgets.go
candiduslynx: No, not a unique stuff, but no all paginators have this.
This is useful when the service decides to return the next token twice.

plugins/source/aws/resources/services/budgets/budgets.go
candiduslynx: We already have `aws_organizations` & `aws_regions` tables.
Besides, I feel like having `aws_%svc%_%resource%` is an artifact of codegen, so we might just go with better names for the new resources.

bbernays: I would be in favor in our next breaking change to add the service name to those tables that only have a resource name...If we are going to move away from the `aws_%svc%_%resource%` naming convention we should do it explicitly with the entire teams buy in.

yevgenypats (changes requested):
Putting on hold till meeting/slack discussion.

Issue #13585, opened 2023-09-01
bug: ERROR: zero-length delimited identifier at or near \"\"\"\" (SQLSTATE 42601)"


hermanschaaf:
This error can happen in the postgres destination if the table being inserted into doesn't exist (or wasn't found by the `listTables` query). We should handle this case better and/or give a more informative error message.

PR #15181, opened 2023-11-08
fix: Fail insert if the table to insert to is unknown to the client

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/13585


plugins/destination/postgresql/client/insert.go
```diff
@@ -43,10 +43,10 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
 		if !ok {
 			return fmt.Errorf("table name not found in metadata")
 		}
-		table := tables.Get(tableName)
-		if table == nil {
+		if _, ok = c.pgTablesToPKConstraints[tableName]; !ok {
```
disq, line 46: if we store the table with `c.pgTablesToPKConstraints[tableName] = ""` we need to get the value and check it for empty too (or just that since it's a string)
```suggestion
		if c.pgTablesToPKConstraints[tableName] == "" {
```

candiduslynx: We should never store such table, actually (pg always names constraints automatically, if no name is provided).
So we either query & get the proper name or we store during table creation & still have non-empty data.

candiduslynx: We should never store such table, actually (pg always names constraints automatically, if no name is provided).
So we either query & get the proper name or we store during table creation & still have non-empty data.

candiduslynx: When we "store" `""` as a value we basically say that we actually saw the table.
However, I see that it might be confusing, so, knowing that the constraint will always have a name in PG, I'll just remove the confusing `switch` from `list_tables.go`

candiduslynx: Simplified code in 254ce2abd18ba9dd0fabebff89d30c90f95f9a91

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -88,7 +88,10 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
 			})
 		}
 		table := tables[len(tables)-1]
-		if pkName != "" {
+		// Note: constraints always have a name in PostgreSQL.
+		// It can either be an assigned name or an automatically generated one,
+		// if the constraint name is skipped in the query adding this constraint.
+		if len(pkName) > 0 { // ~ pkName == "" but without string comparison
```
disq, line 94:
```suggestion
		if len(pkName) > 0 { // ~ pkName != "" but without string comparison
```

disq (approved):

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -88,7 +88,10 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
 			})
 		}
 		table := tables[len(tables)-1]
-		if pkName != "" {
+		// Note: constraints always have a name in PostgreSQL.
+		// It can either be an assigned name or an automatically generated one,
+		// if the constraint name is skipped in the query adding this constraint.
+		if len(pkName) > 0 { // ~ pkName != "" but without string comparison
```
hermanschaaf, line 94: Let's just do `pkName != ""` if that's what we mean, or drop the comment? There's no reason to optimize this and make the code more obscure unless there's a benchmark showing it matters in a significant way. In this case I doubt it

hermanschaaf (approved):

plugins/destination/postgresql/client/insert.go
candiduslynx: Had to move back in 08cc7bc476131bcd64f68c24c799c87ba0a21af0 (added comment why we do that)

plugins/destination/postgresql/client/list_tables.go
candiduslynx: done in 08cc7bc476131bcd64f68c24c799c87ba0a21af0

PR #14925, opened 2023-10-24
feat: Add  gcp `networkconnectivity` tables

shani1998:
Added network connectivity tables
- `gcp_networkconnectivity_locations`
  - `gcp_networkconnectivity_internal_ranges`

PR #15186, opened 2023-11-08
doc: Remove `isolation` column from `gcp_bigtableadmin_app_profiles`

candiduslynx:
Follow-up to https://github.com/cloudquery/cloudquery/pull/14925
Noticed in https://github.com/cloudquery/cloudquery/pull/15183


candiduslynx:
I'm not really sure if we want this to be a breaking change or not (dropping a value (non-PK) column)

disq (approved):

hermanschaaf:
This is only changing documentation, why is it breaking? If the change was already accidentally included in a previous version, what's the point of releasing a breaking change for it now? Are we going to retract the previous version?

candiduslynx:
> This is only changing documentation, why is it breaking? If the change was already accidentally included in a previous version, what's the point of releasing a breaking change for it now? Are we going to retract the previous version?

I don't know if we're going to retract previously released version, I can make this non-brealing

hermanschaaf:
@candiduslynx If it's only changing documentation it's still not a feature either, maybe `chore` or `doc`?

candiduslynx:
> @candiduslynx If it's only changing documentation it's still not a feature either, maybe `chore` or `doc`?

made it `doc:`

hermanschaaf (approved):

PR #15209, opened 2023-11-09
chore(website): Update integration links for premium plugins to point to Hub

hermanschaaf:
Points premium plugins to their new home in the hub, and removes the outdated payment links.

Also removes the "Integrations" link from the top header (while keeping the pages for users coming from Google), in favor of the Hub.


mnorbury (approved):
Looks good

PR #15223, opened 2023-11-10
fix: Optimize batch writing

candiduslynx:
BEGIN_COMMIT_OVERRIDE
feat: Cache queries for writing (#15223)
fix: Don't iterate over batch messages twice (#15223)
END_COMMIT_OVERRIDE

I've tested the sync (`--no-migrate`) with the following test source config (destinations used local pg in Docker + `write_mode: overwrite`):

```yaml
kind: source
spec:
  name: test
  path: cloudquery/test
  registry: github
  version: "v3.1.14"
  tables:
  - "*"
  destinations:
  - postgresql
  spec:
    num_rows: 1000000
    num_sub_rows: 0
```

There's no significant difference in speed or RAM/CPU usage:

| Plugin        | Time           | CPU % | RAM  |
| --- | --- | --- | --- |
| `v7.0.1` |  2m19 | 150% dropping to 50% | 177 MiB |
| `PR` |  2m16 | 130% dropping to 40% | 177 MiB |


plugins/destination/postgresql/client/insert.go
```diff
@@ -46,12 +50,19 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
...
+		sql, ok := queries[tableName]
+		if !ok {
+			// cache the query
+			table := tables.Get(tableName) // will always be present, panic should be produced if not
+			if len(table.PrimaryKeysIndexes()) > 0 {
+				sql = c.upsert(table)
+			} else {
+				sql = c.insert(table)
+			}
+			queries[tableName] = sql
```
bbernays, line 63: Do we have any data or even idea of how much time this will save? Based purely on my intuition I am skeptical that it will save any meaningful time in comparison to the latency involved communicating with the DB over the network.

If in fact this does represent a meaningful performance benefit, why not store the queries in the client that way they can be used across multiple batches?

plugins/destination/postgresql/client/insert.go
```diff
@@ -34,8 +34,12 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
 		return err
 	}
 
-	var sql string
 	batch := &pgx.Batch{}
+
+	// Queries cache.
+	// We may consider LRU cache in the future, but even for 10K records it may be OK to just save.
+	queries := make(map[string]string, 100)
```
bbernays, line 41: Can we guarantee that every insert record has the same query? This would stop users from using different versions of the same plugin in the same configuration.

if a user is using `deterministic_cq_id: true` and `pk_mode: cq_id_only` then they can safely ignore PK changes and therefore  could mix plugin versions

bbernays: If this is an optimization we really want to use we could use all of the cq_meta data fields as cache keys rather than just the table name

plugins/destination/postgresql/client/insert.go
candiduslynx: Consider large batch sizes (100K+).
If there is only a handful of different tables in the batch, we can win a lot of time by simply reusing query instead of constructing it over & over.

plugins/destination/postgresql/client/insert.go
candiduslynx: We can guarantee, as the migrate will first flush the inserts.

bbernays: Not if they run with the flag `--no-migrate`

yevgenypats (changes requested):
As far as Im aware pgx is already doing caching and prepared statement. why we need to add more caching logic. given it's the most used plugin I'd be cautions with any such changes without clear benchmarks  or test cases or references to performance issues we encountered.

plugins/destination/postgresql/client/insert.go
candiduslynx: In this case we still refresh the info with list tables call at the beginning of the flush, so we're also OK there

bbernays: Why would we be doing a flush if we are not doing a migrate?

candiduslynx:
@yevgenypats 
> As far as Im aware pgx is already doing caching and prepared statement. why we need to add more caching logic. given it's the most used plugin I'd be cautions with any such changes without clear benchmarks or test cases or references to performance issues we encountered.

For **each** row we pass to the write batch func we will perform the statement creation from schema. Instead of doing this (if we assume that many records would likely share the same schema), we could just save the insert/upsert result & reuse it once we encounter the record for the same table. That's what this PR is all about.

This doesn't affect the query cache from pgx side, but rather just allows us not to recreate insert/upsert query over & over.

plugins/destination/postgresql/client/insert.go
bbernays: I ran a benchmark to see how much performance increase we could expect for a batch with 100k records. Very surprisingly I found that the caching/map version was significantly longer... 
```
=== RUN   TestGenInsert
    insert_test.go:42: With caching enabled took 29.649439459s
    insert_test.go:52: With caching disabled took 12.558788125s
--- PASS: TestGenInsert (183.31s)
```

You can see the code [here](https://github.com/cloudquery/cloudquery/blob/82b4946937b52ad74960502977d108e117ad9fb1/plugins/destination/postgresql/client/insert_test.go)... My intuition was just like yours that the caching would reduce some of the latency, and had no expectation that it would add latency. I would be super interested to learn why it is in fact slower or what I did wrong in the test code!

plugins/destination/postgresql/client/insert.go
candiduslynx: PG plugin uses ([mixed](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/client.go#L95)) batch writer, so we always perform writing in a batch (to different tables in a single batch, but still).

When performing flush (writing a batch) we first check if we need to [refresh info about tables](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/insert.go#L24-L30). This data is either filled in during migration phase to normalize tables (& updated when we create new table), or updated right before insert.

So, in both simple & `--no-migrate` modes we will end up with already properly filled in info for normalized tables.
Additionally, when [looking up the table to insert to](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/postgresql/client/insert.go#L49) we use `schema.Tables.Get` that will just perform a lookup in a list & return the first encountered entry, so the schema returned is always the same.

hermanschaaf (changes requested):
Before moving forward, can we see a benchmark to show that this change has a (significant) positive impact on performance and that the memory/computation trade-off is worth it?

plugins/destination/postgresql/client/insert.go
candiduslynx: On my machine:
10K
```
=== RUN   TestGenInsert
    insert_test.go:42: With caching enabled took 3.340375ms
    insert_test.go:52: With caching disabled took 103.888792ms
--- PASS: TestGenInsert (6.39s)
PASS
```

One of the issues in the setup you had is that you also measure the alloc time for all of the 100K entries for sql queries.
I've updated the code to use the benchmarks instead in 70c7e80f32091883ef48c7b88fd9315517dca582, here are the results:

#### No caching
Performed by the following command in `plugins/destination/postgresql/client` dir:
```sh
go test \
-test.run=BenchmarkNoCache \
-test.bench=BenchmarkNoCache \
-test.count 10 -test.benchmem -test.benchtime 10000x
```

<details><summary>`BenchmarkNoCache` results</summary>

```
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/cloudquery/plugins/destination/postgresql/client
BenchmarkNoCache-10        10000              9932 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000             10271 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000             10171 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000             16025 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000              9939 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000             10006 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000              9953 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000             10075 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000              9953 ns/op            6016 B/op        241 allocs/op
BenchmarkNoCache-10        10000             10027 ns/op            6016 B/op        241 allocs/op
PASS
ok      github.com/cloudquery/cloudquery/plugins/destination/postgresql/client  62.879s
```

</details>

#### With caching
Performed by the following command in `plugins/destination/postgresql/client` dir:
```sh
go test \
-test.run=BenchmarkCache \
-test.bench=BenchmarkCache \
-test.count 10 -test.benchmem -test.benchtime 10000x
```

<details><summary>`BenchmarkCache` results</summary>

```
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/cloudquery/plugins/destination/postgresql/client
BenchmarkCache-10          10000               306.9 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               298.9 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               518.6 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               308.7 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               304.1 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               331.6 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               324.6 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               317.9 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               327.0 ns/op             0 B/op          0 allocs/op
BenchmarkCache-10          10000               320.7 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/cloudquery/cloudquery/plugins/destination/postgresql/client  61.732s
```

</details>

candiduslynx: I added build tags to emulate the diff better in ba2a986d091bdf7346835264c64ae02fcaae1999, the results:

<details><summary>No cache</summary>

```
go test \
-test.run=BenchmarkSQL \
-test.bench=BenchmarkSQL \
-test.count 10 -test.benchmem -test.benchtime 10000x \
-test.timeout 15m
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/cloudquery/plugins/destination/postgresql/client
BenchmarkSQL-10            10000             14682 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10610 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10628 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10675 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10716 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10539 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10784 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             11285 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10721 ns/op            6016 B/op        241 allocs/op
BenchmarkSQL-10            10000             10842 ns/op            6016 B/op        241 allocs/op
PASS
ok      github.com/cloudquery/cloudquery/plugins/destination/postgresql/client  67.698s
```
</details>

<details><summary>Cache</summary>

```
go test \
-test.run=BenchmarkSQL \
-test.bench=BenchmarkSQL \
-test.count 10 -test.benchmem -test.benchtime 10000x \
-test.timeout 15m \
-tags cache
goos: darwin
goarch: arm64
pkg: github.com/cloudquery/cloudquery/plugins/destination/postgresql/client
BenchmarkSQL-10            10000               339.3 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               349.1 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               357.8 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               351.4 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               339.7 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               351.0 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               345.8 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               335.2 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               351.1 ns/op             0 B/op          0 allocs/op
BenchmarkSQL-10            10000               341.2 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/cloudquery/cloudquery/plugins/destination/postgresql/client  66.661s

```
</details>

candiduslynx:
@hermanschaaf 
> Before moving forward, can we see a benchmark to show that this change has a (significant) positive impact on performance and that the memory/computation trade-off is worth it?

See https://github.com/cloudquery/cloudquery/pull/15223#discussion_r1391128012

plugins/destination/postgresql/client/insert.go
bbernays: This config I don't think we work as a single batch can have records from different tables.
```
kind: source
spec:
  # Source spec section
  name: aws-1
  version: version: v21.0.1
  deterministic_cq_id: true
  tables:
    - aws_ec2_images
  destinations: ["postgresql"]
  spec:
    org:
      member_role_name: OrganizationAccountAccessRole
---
kind: source
spec:
  name: aws-2
  version: v22.0.1
  deterministic_cq_id: true
  tables:
    # This table added new columns in v22.0.1
    - aws_ec2_images
  destinations: ["postgresql"]
  spec:
    accounts:
      - id: test
        local_profile: test-profile
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  registry: "github"
  version: "v6.1.3"
  pk_mode: cq-id-only
  migrate_mode: forced
  spec:
```

Notice how in the 2 source configs they are syncing the same table name, but because they are from different versions so there could be columns included in `aws-2` that are not in `aws-1`... This is definitely an edge case but it is a valid case that we have to consider...

candiduslynx: @bbernays we don't sync them in parallel. (also see sync_v3.go & flushes between those parts)

hermanschaaf (approved):
I ran this locally and measured similar improvements; not major but maybe about 1%. 

I also like the introduction of `flushBatch`, makes the code a bit cleaner :+1:

PR #15231, opened 2023-11-13
chore: Update links to destinations overview

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/actions/runs/6840548429/job/18599942290.
No need to do multiple redirects when we can update our internal links to point to the target


website/vercel.json
erezrokah, line 1: Changes to this file prevent multiple redirect hops and take users directly to the Hub

hermanschaaf (approved):

Issue #15234, opened 2023-11-13
feat: Add a azure adaptive_application_controls source plugin


mayudong200333:
### Describe the plugin

Gets a list of application control machine groups for the subscription.

### Link to API documentation

https://learn.microsoft.com/ja-jp/rest/api/defenderforcloud/adaptive-application-controls/list?view=rest-defenderforcloud-2020-01-01&tabs=HTTP

### Pull request (optional)

- [X] I can submit a pull request

bbernays:
closed by https://github.com/cloudquery/cloudquery/pull/15236

PR #15236, opened 2023-11-13
feat: Added plugins for azure adaptive application controls

mayudong200333:
#### Summary

- Added azure security plugin `adaptive_application_controls`
- Added corresponding mock test
- Ref: https://github.com/cloudquery/cloudquery/issues/15234


erezrokah (changes requested):
Hi @mayudong200333, thanks for the PR! I think you'll need to reference this new table here https://github.com/cloudquery/cloudquery/blob/20b75f3b8162c07090a67b98d11e81688d209303/plugins/source/azure/resources/plugin/tables.go#L318

mayudong200333:
@erezrokah 
Thank you for your review, I had reflected changes on [4deae64](https://github.com/cloudquery/cloudquery/pull/15236/commits/4deae64b1e829cf47d667338aef55f08cfe236c7)

erezrokah (approved):
Tested it, looks great (got 0 results but that's because I don't have these resources)

PR #15264, opened 2023-11-14
feat: Add `plugin docs` commands for doc updates

disq:
```
cloudquery plugin docs download test-team/test-plugin@v1.0.0 [-D docs]
cloudquery plugin docs upload test-team/test-plugin@v1.0.0 [--sync] [-D docs]
```


erezrokah (commented):
Code looks good but I would keep this command hidden for now so we can iterate on this feature. I would say it would be easier to allow doc editing via https://cloud.cloudquery.io.

Looks like we already allow editing release notes in the UI

erezrokah (approved):

PR #15266, opened 2023-11-14
chore: Fix tables diff script

erezrokah:
#### Summary

Since we removed the tables generation for most plugins to `website/tables` in favor generating them under each plugin, this PR updates the table diff script to take the tables files from the correct location.
It also adds the missing table docs for `bitbucket`, `square` and `typeform` (`airtable` has dynamic tables so not needed).

Docs summary runs on `pull_request_target` so this will only take effect once merged to `main`


disq (approved):

.github/workflows/source_bitbucket.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: "temurin"
-          java-version: "18"
+          java-version: "20"
```
erezrokah, line 34: Hopefully fixes https://github.com/cloudquery/cloudquery/actions/runs/6866733430/job/18673589655#step:5:30

PR #15268, opened 2023-11-15
chore: Better names for sub-tests

candiduslynx:


hermanschaaf (changes requested):
table.Name is used in the for loop, it should be part of the name as well so it's possible to differentiate which one failed

plugins/source/aws/client/testing.go
```diff
@@ -90,7 +90,7 @@ func AwsCreateMockClient(t *testing.T, ctrl *gomock.Controller, builder func(*te
 
 func validateTagStructure(t *testing.T, tables schema.Tables) {
 	for _, table := range tables.FlattenTables() {
-		t.Run(table.Name, func(t *testing.T) {
+		t.Run("validate tags", func(t *testing.T) {
```
disq, line 93:
```suggestion
		t.Run("validate tags " + table.Name, func(t *testing.T) {
```
maybe?

disq: I was going to use `:` and so on as separator but in the end they all get converted to `_` anyway

candiduslynx: Addressed in 7aaa291ecb0ba282223bee6b8a80066c5279a7b4

plugins/source/aws/client/testing.go
```diff
@@ -119,7 +119,7 @@ func validateMultiplexers(t *testing.T, parentTable *schema.Table) {
 
 func validateSkippedColumns(t *testing.T, tables schema.Tables) {
 	for _, table := range tables.FlattenTables() {
-		t.Run(table.Name, func(t *testing.T) {
+		t.Run(table.Name+"/validate skipped columns", func(t *testing.T) {
```
hermanschaaf, line 122: the `/` will get converted to an underscore anyway I believe

hermanschaaf (approved):

plugins/source/aws/client/testing.go
hermanschaaf: oh ok, good to know

Issue #15267, opened 2023-11-14
scaffold generates broken CI


kd7lxl:
I used the scaffold to generate a source plugin and pushed it to github.
https://github.com/cloudquery/cloudquery/blob/1d051d391a580cfbf94753ac478fffb74dffa9dc/scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl#L18-L25
The test workflow failed in `golanglint-ci`. The first failure was a timeout. After fixing the timeout, `golangci-lint` has trouble restoring its cache. This appears to be some conflict between the `setup-go` action's default cache and the `golangci-lint` action's default cache. I worked around it by disabling caching, but this isn't a good long term solution.

Example errors:
```
Run golangci/golangci-lint-action@v3
prepare environment
  Checking for go.mod: go.mod
  Received 88080384 of 525990884 (16.7%), 84.0 MBs/sec
  Received 272629760 of 525990884 (51.8%), 129.9 MBs/sec
  Received 452984832 of 525990884 (86.1%), 144.0 MBs/sec
  Cache Size: ~502 MB (525990884 B)
  /usr/bin/tar -xf /home/runner/work/_temp/5dc58b45-a00c-4ad3-be70-055e4a0a2007/cache.tzst -P -C /home/runner/work/cq-source-disco/cq-source-disco --use-compress-program unzstd
  Received 525990884 of 525990884 ([10](https://github.com/.../actions/runs/6868901301/job/18680529528#step:4:11)0.0%), [12](https://github.com/.../actions/runs/6868901301/job/18680529528#step:4:13)5.3 MBs/sec
  /usr/bin/tar: ../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.4.linux-amd64/VERSION: Cannot open: File exists
  /usr/bin/tar: ../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.4.linux-amd64/SECURITY.md: Cannot open: File exists
  /usr/bin/tar: ../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.21.4.linux-amd64/bin/gofmt: Cannot open: File exists
  /usr/bin/tar: ../../../go/pkg/mod/golang.org/toolchain@v0.0.1-go1.[21](https://github.com/.../actions/runs/6868901301/job/18680529528#step:4:22).4.linux-amd64/bin/go: Cannot open: File exists
[...]

PR #15273, opened 2023-11-15
fix: Increase test and lint timeouts in tests, fix caching

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/15267


scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -22,7 +22,10 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.51.2
+          version: v1.55.2
+          skip-pkg-cache: true
+          skip-build-cache: true
```
erezrokah, line 27: We use this approach in our tests to only keep the lint results cache. `setup-go` handles build and package cache already

scaffold/cmd/templates/source/.github/workflows/test.yaml.tpl
```diff
@@ -22,7 +22,10 @@ jobs:
...
+          args: --timeout=10m
```
kd7lxl, line 28: This is very similar to the change I made locally to workaround https://github.com/cloudquery/cloudquery/issues/15267.

kd7lxl (approved):

mnorbury (approved):

PR #15274, opened 2023-11-15
chore: Update transformations

yevgenypats:
Remove old policies section and redirect to hub


website/vercel.json
```diff
@@ -865,6 +865,15 @@
     {
       "source": "/docs/plugins/overview",
       "destination": "https://hub.cloudquery.io"
+    },
+    {
+      "source": "/docs/core-concepts/policies",
```
erezrokah, line 870: I think only the first redirect that matches take, so not sure we should have 2 redirects with the exact same source.

website/vercel.json
erezrokah, line 1: In this file we also have:

```json
    {
      "source": "/docs/policies",
      "destination": "/docs/core-concepts/policies"
    },
```

I suggest we update it to avoid multiple hops

erezrokah (commented):
Added a couple of comments. We might want to also do a find replace on our docs to replace `/docs/core-concepts/policies` with `https://hub.cloudquery.io/addons/transformation` to avoid extra hops when navigating our Website

yevgenypats:
> Added a couple of comments. We might want to also do a find replace on our docs to replace `/docs/core-concepts/policies` with `https://hub.cloudquery.io/addons/transformation` to avoid extra hops when navigating our Website

@erezrokah mind overtaking this please?

erezrokah (approved):

PR #15278, opened 2023-11-15
feat: Add teamName to managed plugin configuration

mnorbury:
This adds the team name to the managed plugins configuration, which will use the `/team/*` endpoints to download plugins - allowing plugin downloads to be attributed to the correct team

fixes: https://github.com/cloudquery/cloudquery-issues/issues/835

### DIFF
--- a/cli/cmd/tables.go
+++ b/cli/cmd/tables.go
@@ -63,9 +63,10 @@ func tables(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to get auth token: %w", err)
 	}
+	//TODO: add team name
 	opts := []managedplugin.Option{
 		managedplugin.WithLogger(log.Logger),
-		managedplugin.WithAuthToken(authToken),
+		managedplugin.WithAuthToken(authToken.Value),
 	}
 	if cqDir != "" {
 		opts = append(opts, managedplugin.WithDirectory(cqDir))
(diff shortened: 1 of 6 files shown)


cli/cmd/tables.go
```diff
@@ -63,9 +63,10 @@ func tables(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to get auth token: %w", err)
 	}
+	//TODO: add team name
```
hermanschaaf, line 66: Any blockers to adding the team name here as well?

hermanschaaf (approved):

cli/cmd/tables.go
mnorbury: Nope, added now - missed it on first pass.

cli/internal/auth/team.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+	"github.com/cloudquery/cloudquery-api-go/auth"
+	"github.com/cloudquery/cloudquery-api-go/config"
+)
+
+// GetTeamForToken returns the team for the given token
+// If the token is a bearer token, we need to get the team from the configuration.
+// For api keys the team name is not required as the key is bound to a team name already.
+func GetTeamForToken(token auth.Token) (string, error) {
```
disq, line 12: Looks like this can be utilized in the addon flow as well

PR #15279, opened 2023-11-15
chore: Embed latest version in docs during publish

erezrokah:
#### Summary

Instead of doing version replacement in the UI this does it during publish.

### DIFF
--- a/.github/workflows/publish_plugin_to_hub.yml
+++ b/.github/workflows/publish_plugin_to_hub.yml
@@ -105,6 +105,13 @@ jobs:
             });
             return shellescape([data.body]);
 
+      - name: Find and Replace
+        uses: jacobtomlinson/gha-find-replace@a51bbcd94d000df9ca0fcb54ec8be69aad8374b0
+        with:
+          find: "(?i)version_${{ needs.prepare.outputs.plugin_kind }}_${{ needs.prepare.outputs.plugin_name }}"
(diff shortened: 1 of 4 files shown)


bbernays:
This lowers the likelihood of a user seeing an example configuration in the hub that is incorrect but it also means that a user can find an outdated configuration on the hub

erezrokah:
> means that a user can find an outdated configuration on the hub

I don't think we can call it outdated. The version in the configuration will match the version you're browsing

bbernays:
> > means that a user can find an outdated configuration on the hub
> 
> I don't think we can call it outdated. The version in the configuration will match the version you're browsing. We have a callout at if you're not browsing the latest version


If we have a config that shows how to sync `source_plugin_a` to `destination_plugin_b`. As soon as we release a new version of `destination_plugin_b` the demo config will be out of date. The callout will only be present if we have released a new `source_plugin_a`, but not if we have released a new `destination_plugin_b`.

Maybe I am missing an assumption or something else that goes into the docs generation/display.

erezrokah:
> As soon as we release a new version of `destination_plugin_b` the demo config will be out of date

I'm not saying this replaces https://github.com/cloudquery/frontend/pull/432. We would need both for configs that reference plugins other than the one we publish.
I still think it makes sense to embed the version in the docs (like we embed it in the binary)

jackcloudquery (approved):

PR #15285, opened 2023-11-15
fix: Simplify `addon download` team reading

disq:
follow up to https://github.com/cloudquery/cloudquery/pull/15278

### DIFF
--- a/cli/cmd/addon_download.go
+++ b/cli/cmd/addon_download.go
@@ -91,9 +91,9 @@ func runAddonDownload(ctx context.Context, cmd *cobra.Command, args []string) er
 		return err
 	}
 
-	currentTeam, err := auth.GetTeamForToken(token)
+	currentTeam, err := auth.RequireTeamForToken(token)
 	if err != nil {
-		return fmt.Errorf("failed to get team name: %w", err)
+		return fmt.Errorf("failed to get team name for token: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


cli/cmd/addon_download.go
```diff
@@ -112,17 +116,7 @@ func runAddonDownload(ctx context.Context, cmd *cobra.Command, args []string) er
...
 }
 
-func getAddonMetadata(ctx context.Context, c *cloudquery_api.ClientWithResponses, tokenType auth.TokenType, addonTeam, addonType, addonName, addonVersion string) (location, checksum string, retErr error) {
-	var currentTeam string
-
-	if tokenType != auth.APIKey {
-		var err error
-
-		currentTeam, err = config.GetValue("team")
-		if err != nil && !errors.Is(err, os.ErrNotExist) {
```
disq, line 122: We don't have this check inside `auth.GetTeamForToken()`. Not sure if it's fine (can already logged in users still have no team - so no file?)
`tc.GetToken()` would fail first if there was no login.

disq: fixed with https://github.com/cloudquery/cloudquery/pull/15285/commits/2b8525244474d27671d5689a139c06fd5f223630

Other commands might need this as well.

mnorbury (approved):

Issue #13415, opened 2023-08-29
feat: AWS Inspector V2 coverage 


JohnCalhoun:
### Which problem is this feature request solving?

I would like to ingest aws inspector V2 coverage data

docs: https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListCoverage.html

### Describe the solution you'd like

I would like another AWS Table for inspector v2 coverage data

### Pull request (optional)

- [ ] I can submit a pull request

PR #15350, opened 2023-11-16
feat: Add aws inspector2 coverages table

rubenandre:
#### Summary
This PR aims to create a new table aws_inspector2_coverages that maps https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListCoverage.html

Added Tables
- aws_inspector2_coverages

Resolve issue #13415


plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func Coverages() *schema.Table {
+	tableName := "aws_inspector2_coverages"
+	return &schema.Table{
+		Name: tableName,
+		Description: `https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListCoverage.html
+The 'request_account_id' and 'request_region' columns are added to show from where the request was made.`,
+		Resolver:  fetchInspector2Coverages,
+		Transform: transformers.TransformWithStruct(&types.CoveredResource{}, transformers.WithSkipFields("ResourceId")),
```
bbernays, line 20: We always aim to do as little transformation as possible, so it is fine to have columns with duplicate data... So no need skip the `ResourceId`
```suggestion
		Transform: transformers.TransformWithStruct(&types.CoveredResource{}),
```

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,48 @@
...
+		Multiplex: client.ServiceAccountRegionMultiplexer(tableName, "inspector2"),
+		Columns: []schema.Column{
+			{
+				Name:       "arn",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   schema.PathResolver("ResourceId"),
```
bbernays, line 26: According to my testing `ResourceId` doesn't have to be an ARN... It can be an `instanceID` or a `accountID`... So I think we will need to use a compound PK including things like Region, accountID

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,48 @@
...
+func Coverages() *schema.Table {
+	tableName := "aws_inspector2_coverages"
+	return &schema.Table{
+		Name: tableName,
+		Description: `https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListCoverage.html
+The 'request_account_id' and 'request_region' columns are added to show from where the request was made.`,
+		Resolver:  fetchInspector2Coverages,
+		Transform: transformers.TransformWithStruct(&types.CoveredResource{}, transformers.WithSkipFields("ResourceId")),
+		Multiplex: client.ServiceAccountRegionMultiplexer(tableName, "inspector2"),
+		Columns: []schema.Column{
```
bbernays, line 22: We are missing the default columns (req_account_id and req_account_id) that we include in all tables that support cross account and cross region aggregation

https://github.com/cloudquery/cloudquery/blob/9fe1ce1c5008250d8282f1fa31b6071f68f4fd50/plugins/source/aws/resources/services/inspector2/findings.go#L25-L36

bbernays (changes requested):

plugins/source/aws/resources/services/inspector2/coverages.go
rubenandre: Thanks for the explanation. 
Removed.

plugins/source/aws/resources/services/inspector2/coverages.go
rubenandre: Done.

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func Coverages() *schema.Table {
+	tableName := "aws_inspector2_coverages"
+	return &schema.Table{
+		Name: tableName,
+		Description: `https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListCoverage.html
+The 'request_account_id' and 'request_region' columns are added to show from where the request was made.`,
+		Resolver:  fetchInspector2Coverages,
+		Transform: transformers.TransformWithStruct(&types.CoveredResource{}),
```
bbernays, line 20: I think that we are still missing a Primary Key (or keys). While it is important to include the `request_` fields in the primary key, we need to specify some other field(s) so that each item in the response from the API can be uniquely identified.  Maybe it should be `ResourceID`?

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/aws/aws-sdk-go-v2/service/inspector2"
+	"github.com/aws/aws-sdk-go-v2/service/inspector2/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+)
+
+func Coverages() *schema.Table {
+	tableName := "aws_inspector2_coverages"
```
candiduslynx, line 14: I suggest renaming it to `aws_inspector2_covered_resources` for consistency
```suggestion
	tableName := "aws_inspector2_covered_resources"
```

Also, rename files to `covered_resources*.go`

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
...
+	return &schema.Table{
+		Name: tableName,
+		Description: `https://docs.aws.amazon.com/inspector/v2/APIReference/API_ListCoverage.html
```
candiduslynx, line 17: Link to the structure doc instead
```suggestion
		Description: `https://docs.aws.amazon.com/inspector/v2/APIReference/API_CoveredResource.html
```

plugins/source/aws/resources/services/inspector2/coverages.go
candiduslynx, line 20: Add primary keys:
```suggestion
		Transform: transformers.TransformWithStruct(&types.CoveredResource{}, transformers.WithPrimaryKeys("AccountId", "ResourceId")),
```

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
...
+The 'request_account_id' and 'request_region' columns are added to show from where the request was made.`,
+		Resolver:  fetchInspector2Coverages,
+		Transform: transformers.TransformWithStruct(&types.CoveredResource{}),
+		Multiplex: client.ServiceAccountRegionMultiplexer(tableName, "inspector2"),
```
candiduslynx, line 21: you can use simplified code here:
```suggestion
		Multiplex: client.ServiceAccountRegionMultiplexer(tableName, client.AWSServiceInspector2.String()),
```

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
...
+				Resolver:   client.ResolveAWSAccount,
+				PrimaryKey: true,
+			},
+			{
+				Name:       "request_region",
+				Type:       arrow.BinaryTypes.String,
+				Resolver:   client.ResolveAWSRegion,
+				PrimaryKey: true,
+			},
+		},
```
candiduslynx, line 35: there are helper columns in `client` package:
```suggestion
		Columns: []schema.Column{
			client.RequestAccountIDColumn(true),
			client.RequestRegionColumn(true),
		},
```

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func fetchInspector2Coverages(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	cl := meta.(*client.Client)
+	svc := cl.Services(client.AWSServiceInspector2).Inspector2
+	input := inspector2.ListCoverageInput{}
```
candiduslynx, line 42: add `MaxResults` param
```suggestion
	input := inspector2.ListCoverageInput{MaxResults: aws.Int32(200)}
```

plugins/source/aws/resources/services/inspector2/coverages.go
```diff
@@ -0,0 +1,54 @@
+package inspector2
+
+import (
+	"context"
+	"github.com/apache/arrow/go/v14/arrow"
```
candiduslynx, line 5: If you apply helper columns from `client` package & the `MaxResults` fix, the imports will need to be changed:
```suggestion
	"context"

	"github.com/aws/aws-sdk-go-v2/aws"
```

candiduslynx (changes requested):
Hello @rubenandre!

Thank you for the contribution!
I also have a PR in the workings for this table, so I think we could get your PR up to speed & merge it instead.

Here are the things I'd like to sort out before merging your code.

candiduslynx:
@rubenandre I've extracted the changes to https://github.com/cloudquery/cloudquery/pull/15706 (applied my suggestions, too).

Could you please tell me if it's OK to give you a shoutout in the release notes as well?

PR #15351, opened 2023-11-17
fix: Addon publish command should take `message` as-is unless prefixed with `@`

disq:
For consistency.

### DIFF
--- a/cli/cmd/addon_publish.go
+++ b/cli/cmd/addon_publish.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strings"
 	"syscall"
 
 	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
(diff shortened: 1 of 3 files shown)


erezrokah (approved):

cli/cmd/addon_publish.go
```diff
@@ -167,12 +168,17 @@ func createNewAddonDraftVersion(ctx context.Context, c *cloudquery_api.ClientWit
 		body.Doc = string(b)
 	}
 
-	if manifest.PathToMessage != "" {
-		b, err := os.ReadFile(filepath.Join(manifestDir, manifest.PathToMessage))
-		if err != nil {
-			return fmt.Errorf("failed to read message file: %w", err)
+	if manifest.Message != "" {
```
erezrokah, line 171: Isn't `message` required? We can add a validation in a separate PR if we don't have it already

disq: @erezrokah it's required, but the remote endpoint checks for it anyway. I don't think we need to duplicate all of that in client code.

PR #15371, opened 2023-11-19
fix!: Remove all `result_metadata` columns

bbernays:


candiduslynx:
I propose to move this PR to the [`feat/aws/v23-quality-of-life`](https://github.com/cloudquery/cloudquery/tree/feat/aws/v23-quality-of-life) branch (see https://github.com/cloudquery/cloudquery/issues/12648 description)

disq:
Should consider client.TransformWithStruct approach (with a global SkipFields) later so result_metadata doesn't come back with newer resources. Currently it's not possible without writing a full wrapper around `transformers`, but https://github.com/cloudquery/plugin-sdk/pull/1379 will help.

bbernays:
> Should consider "**client**.TransformWithStruct" approach (with a global SkipFields) later so result_metadata doesn't come back with newer resources. Currently it's not possible without writing a full wrapper around `transformers`, but [cloudquery/plugin-sdk#1379](https://github.com/cloudquery/plugin-sdk/pull/1379) will help.

Being able to remove the explicit `ignore` would be great, but we have a test that automatically runs for every table to make sure it doesn't come back

candiduslynx (approved):

PR #15395, opened 2023-11-20
fix!: Refactor `aws_s3_buckets` to use nested tables rather than that column resolvers

bbernays:


disq:
Except for `PolicyStatus` which is a single bool field (behind an api call...)

erezrokah (approved):

Issue #12974, opened 2023-08-10
bug: `aws_cloudtrail_trail_event_selectors` is missing advanced selectors


hermanschaaf:
The trail event selectors call (https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/resources/services/cloudtrail/trail_event_selectors.go) returns a response containing `EventSelectors` as well as `AdvancedEventSelectors`, but currently the `aws_cloudtrail_trail_event_selectors` table only includes the `EventSelectors`. We should add the missing advanced event selectors. Since advanced event selectors have a different schema, this may need to be done in a different table if we want to keep the change backwards-compatible.

bbernays:
We have a breaking change that will come in the near(ish) future (#12648) lets just include this change in that

PR #15397, opened 2023-11-21
feat: Add `aws_cloudtrail_trail_advanced_event_selectors` table

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/12974

Alternative to https://github.com/cloudquery/cloudquery/pull/15400

### DIFF
--- a/plugins/source/aws/resources/services/cloudtrail/trail_advanced_event_selectors.go
+++ b/plugins/source/aws/resources/services/cloudtrail/trail_advanced_event_selectors.go
@@ -1,0 +1,47 @@
+package cloudtrail
+
+import (
+	"context"
+
+	"github.com/apache/arrow/go/v14/arrow"
+	"github.com/aws/aws-sdk-go-v2/service/cloudtrail"
+	"github.com/aws/aws-sdk-go-v2/service/cloudtrail/types"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
+	"github.com/cloudquery/cloudquery/plugins/source/aws/resources/services/cloudtrail/models"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+)
+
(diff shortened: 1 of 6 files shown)


plugins/source/aws/resources/services/cloudtrail/trail_advanced_event_selectors.go
```diff
@@ -0,0 +1,47 @@
...
+			},
+		},
+	}
+}
+
+func fetchCloudtrailTrailAdvancedEventSelectors(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
+	r := parent.Item.(*models.CloudTrailWrapper)
+	cl := meta.(*client.Client)
+	svc := cl.Services(client.AWSServiceCloudtrail).Cloudtrail
+	response, err := svc.GetEventSelectors(ctx, &cloudtrail.GetEventSelectorsInput{TrailName: r.TrailARN}, func(options *cloudtrail.Options) {
```
disq, line 39: Doing the `GetEventSelectors` call twice to get two separate fields from it? Is there a way of caching this call (or does the AWS client cache it?) in a field in `cl` so that it doesn't hit the endpoint twce?

disq: it would also need a sync primitive in case the resource is resolved in parallel

candiduslynx: We could move to the https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_GetEventSelectors.html result instead for the table (& as a bonus, get the PK from trail ARN).

candiduslynx: alt PR: https://github.com/cloudquery/cloudquery/pull/15400

candiduslynx:
Closed in favor of https://github.com/cloudquery/cloudquery/pull/15400

PR #15398, opened 2023-11-21
feat!: Remove `aws_alpha_*` tables

candiduslynx:
Part of https://github.com/cloudquery/cloudquery/issues/12648


hermanschaaf (approved):
For others reading this: normally we don't consider changes to (or removal of) alpha tables breaking, but since we're bundling many breaking changes together in the next release we're marking this as breaking as well.

PR #15400, opened 2023-11-21
fix!: Include advanced event selectors in `aws_cloudtrail_trail_event_selectors`

### LINKED ISSUE #12974: bug: `aws_cloudtrail_trail_event_selectors` is missing advanced selectors

candiduslynx:
Closes #12974

Alternative to https://github.com/cloudquery/cloudquery/pull/15397

### DIFF
--- a/plugins/source/aws/docs/tables/aws_cloudtrail_trail_event_selectors.md
+++ b/plugins/source/aws/docs/tables/aws_cloudtrail_trail_event_selectors.md
@@ -2,9 +2,9 @@
 
 This table shows data for AWS CloudTrail Trail Event Selectors.
 
-https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_EventSelector.html
+https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_GetEventSelectors.html
 
-The primary key for this table is **_cq_id**.
+The primary key for this table is **trail_arn**.
 
 ## Relations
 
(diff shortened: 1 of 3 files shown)


disq (approved):

plugins/source/aws/policies_v1/views/log_metric_filter_and_alarm.sql
```diff
@@ -2,19 +2,27 @@ create or replace view view_aws_log_metric_filter_and_alarm as
...
+tes as (
+  select trail_arn from aws_cloudtrail_trail_event_selectors
+  where exists(
+    select * from jsonb_array_elements(event_selectors) as es
+    where es ->>'ReadWriteType' = 'All' and (es->>'IncludeManagementEvents')::boolean = TRUE
+  ) or exists(
+    select * from jsonb_array_elements(advanced_event_selectors) as aes
+    where not exists(select * from jsonb_array_elements(aes ->'FieldSelectors') as aes_fs where aes_fs ->>'Field' = 'readOnly')
+  )
 )
```
bbernays, line 15: These queries should be removed. `v23` will not have policies in the repo.

candiduslynx: So I can just revert the diff from this PR, right?

candiduslynx: 8bae49f2cd5e58ec51b8717668c4e52d265a142c

candiduslynx:
@bbernays could you please take a look before merging?

bbernays (approved):

PR #15405, opened 2023-11-21
fix!: Remove deprecated `table_options`

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/15404


plugins/source/aws/resources/services/ecs/cluster_tasks.go
```diff
@@ -48,34 +48,33 @@ func fetchEcsClusterTasks(ctx context.Context, meta schema.ClientMeta, parent *s
 
 	cl := meta.(*client.Client)
 	svc := cl.Services(client.AWSServiceEcs).Ecs
+	input := &ecs.ListTasksInput{Cluster: cluster.ClusterArn}
```
bbernays, line 51: Can we set the `MaxResults = aws.Int32(100)` like we were doing in the `table_options`

candiduslynx: 3fb8f53b37fe9e4ba012642be20964a2531ad046

bbernays (approved):

PR #15409, opened 2023-11-21
feat!: Make `registry: cloudquery` the default registry instead of `registry: github`

disq:
Uses ~https://github.com/cloudquery/plugin-pb-go/pull/170~ https://github.com/cloudquery/plugin-pb-go/pull/172 (now tagged as 1.14.2)

Part of https://github.com/cloudquery/cloudquery-issues/issues/924 (Internal issue)


cli/internal/specs/v0/spec.go
```diff
@@ -83,7 +83,22 @@ func (s *Spec) UnmarshalJSON(data []byte) error {
...
 	dec.DisallowUnknownFields()
-	return dec.Decode(s.Spec)
+	if err = dec.Decode(s.Spec); err != nil {
+		return err
+	}
+
+	var r struct {
+		Registry string `json:"registry"`
+	}
+	if err := json.Unmarshal(b, &r); err == nil && r.Registry == "" {
```
disq, line 93: The gist of it: During unmarshal, if `registry` (as a string) was empty, we mark it as such.

cli/internal/specs/v0/destination.go
```diff
@@ -39,8 +42,8 @@ func (d *Destination) SetDefaults(defaultBatchSize, defaultBatchSizeBytes int) {
 	if d.Spec == nil {
 		d.Spec = make(map[string]any)
 	}
-	if d.Registry.String() == "" {
-		d.Registry = RegistryGithub
+	if d.registryInferred && d.Registry == 0 {
+		d.Registry = RegistryCloudQuery
```
disq, line 46: The gist part 2, for destinations: if registry was an empty string during unmarshal and defaulted to the zero-value of `int`, we override it to be our default.

cli/internal/specs/v0/source.go
```diff
@@ -107,8 +110,8 @@ func (s *Source) SetDefaults() {
 	if s.Spec == nil {
 		s.Spec = make(map[string]any)
 	}
-	if s.Registry.String() == "" {
-		s.Registry = RegistryGithub
+	if s.registryInferred && s.Registry == 0 {
+		s.Registry = RegistryCloudQuery
```
disq, line 114: Same as the destination.go change.

cli/internal/specs/v0/spec.go
erezrokah: Can't we make the JSON registry property a pointer then check for nullability instead?

disq: We can but it's everywhere, lots of places will have `to.Ptr` or `RegistryPtr` and dereferencing, which is not nice. I would rather make it a string if we're not doing this.

disq: @erezrokah https://github.com/cloudquery/cloudquery/pull/15409/commits/f99f20e56101e8c33eb91df4090fa06f9bcd0a58 uses `string` for `Registry`, no funny business.

cli/internal/specs/v0/registry.go
```diff
@@ -1,56 +1,34 @@
...
 	var registry string
 	if err := json.Unmarshal(data, &registry); err != nil {
 		return err
 	}
-	if *r, err = RegistryFromString(registry); err != nil {
-		return err
-	}
-	return nil
+	*r = Registry(registry)
+	return r.IsValid()
```
disq, line 24: This make sure if `registry` is specified it needs to be a valid value. It still stays zero value (empty string) if it's not specified.

erezrokah:
What if we change https://github.com/cloudquery/cloudquery/blob/720fc24f3017d474caceae599814b9593cd66cbd/cli/internal/specs/v0/registry.go#L12 to start with `None` then if it's missing it will be unmarshalled to `None`, which we can set the default for to `RegistryCloudQuery`.

Not sure if we rely on the int values not changing somewhere. Based on the `MarshalJSON` code we use the string value

disq:
> What if we change
> 
> https://github.com/cloudquery/cloudquery/blob/720fc24f3017d474caceae599814b9593cd66cbd/cli/internal/specs/v0/registry.go#L12
> 
> to start with `None` then if it's missing it will be unmarshalled to `None`, which we can set the default for to `RegistryCloudQuery`.
> Not sure if we rely on the int values not changing somewhere. Based on the `MarshalJSON` code we use the string value

@erezrokah I don't think it's used anywhere else other than specs/v0, let's do that. But I don't really believe it's necessary to use an integer enum here at all. I'll keep it out of SetDefaults so we can remove `registryInferred` logic too.

cli/cmd/install.go
```diff
@@ -72,31 +72,55 @@ func installPlugin(cmd *cobra.Command, args []string) error {
...
+		return enrichClientError(clist, destinationRegInferred, err)
+	}
+
+	return nil
+}
 
-	if _, err := managedplugin.NewClients(ctx, managedplugin.PluginSource, sourcePluginConfigs, opts...); err != nil {
+// enrichClientError gets the index of the failed client (which is one more than the last one on the list) and checks if the registry was inferred.
+// If so, adds a hint to the error message.
+func enrichClientError(clist managedplugin.Clients, infs []bool, err error) error {
```
erezrokah, line 109:
```suggestion
func enrichClientError(clientsList managedplugin.Clients, inferredList []bool, err error) error {
```

Maybe longer var naming will help, not sure

cli/cmd/install.go
```diff
@@ -72,31 +72,55 @@ func installPlugin(cmd *cobra.Command, args []string) error {
...
+	il := len(infs)
+	if l > il {
+		return err // shouldn't happen
+	}
+	if !infs[l] {
 		return err
 	}
 
-	return nil
+	return fmt.Errorf("%w. Hint: make sure plugin version exists in hub.cloudquery.io or if it's an older version, use `registry: github`", err)
```
erezrokah, line 125:
```suggestion
	return fmt.Errorf("%w. Hint: make sure to use the latest plugin version from hub.cloudquery.io or to keep using an outdated version add `registry: github` to your configuration", err)
```

erezrokah (approved):
Added a few comments, non blocking. We should test this with older plugins to make sure the enum change works

disq:
Changed the enums in https://github.com/cloudquery/cloudquery/pull/15409/commits/562edd501da1531f50a245da133321696b28cbb9

inferred bool had to stay, otherwise (if you make it a behavior change in the parsing) it's not unit testable.

cli/cmd/install.go
disq: I just copy pasted the message from the original issue. Here we don't know if it's an outdated version or the plugin is not in Hub for other reasons (3rd party, not published, unsupported docker etc.)

disq:
@erezrokah Tested with AWS v13.0.0 and v14.0.0 successfully
AWS v10/v11/v12 got me `Error: please upgrade source old or use an older CLI version, between v3.0.1 and v3.5.3`

PR #15412, opened 2023-11-22
fix!: Remove premium-only `event_based_sync` spec

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/15411


erezrokah:
Can we not change the docs until the premium version is available in the Hub? Once it's available we can update the docs to link to it. WDYT?

candiduslynx:
> Can we not change the docs until the premium version is available in the Hub? Once it's available we can update the docs to link to it. WDYT?

IMO that's not required, as we intentionally documented the event-based-sync as premium only feature.
So, just removing this would be fine (no real usability for it in the free version)

erezrokah:
> So, just removing this would be fine (no real usability for it in the free version)

It's about making the premium features visible when someone navigates the free version, so users know they can pay and get more features

candiduslynx:
> It's about making the premium features visible when someone navigates the free version, so users know they can pay and get more features

IMO this should be a separate task to document premium-only features in premium & link to them from the free version (& out of scope of this change).

erezrokah:
> IMO this should be a separate task to document premium-only features in premium & link to them from the free version (& out of scope of this change).

But if we remove the docs in this PR, we'll be removing the only publicly visible information we have about the premium plugin. My suggestion is not to change the docs in this PR at all, and document the premium-only features in premium & link to them like you suggested in a separate task once we have the premium plugin available in the Hub.

candiduslynx:
> My suggestion is not to change the docs in this PR at all, and document the premium-only features in premium & link to them like you suggested in a separate task once we have the premium plugin available in the Hub.

@erezrokah I've returned the docs about this feature & mentioned the premium-only part in the separate doc for the feature as well

plugins/source/aws/docs/event-based-sync.md
```diff
@@ -1,5 +1,5 @@
 :::callout{type="info"}
-This feature is currently in a closed preview. Sign up for access using this [form](https://cloudquery.typeform.com/to/Dvdn6P2u)
```
erezrokah, line 2: Can we keep the sign up link? I don't see a reason to remove it yet. If someone signs up we can tell them we'll release the premium version soon

candiduslynx: d8876ca0b96fa0570b02712cc188a5f743681094

erezrokah (approved):

PR #15421, opened 2023-11-22
chore: Explain ETL vs ELT

rngadam:
small typo fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,14 @@ See the **[Quickstart guide](https://www.cloudquery.io/docs/quickstart)** for in
 - **Cloud Security Posture Management**: Use as an open source CSPM solution to monitor and enforce security policies across your cloud infrastructure for AWS, GCP, Azure and many more.
 - **Cloud Asset Inventory**: First-class support for major cloud infrastructure providers such as AWS, GCP and Azure allow you to collect and unify configuration data.
 - **Cloud FinOps**: Collect and unify billing data from cloud providers to drive financial accountability.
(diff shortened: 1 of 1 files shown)


erezrokah (commented):
Hi @rngadam, can you explain the reason this is a typo? Our main Website also states we're an `ELT` framework https://www.cloudquery.io/

rngadam:
> Hi @rngadam, can you explain the reason this is a typo? Our main Website also states we're an `ELT` framework https://www.cloudquery.io/


https://www.techtarget.com/searchdatamanagement/definition/Extract-Load-Transform-ELT

>ELT vs. ETL: The differences between ELT and a traditional ETL process are more significant than just switching the L and the T. The biggest determinant is how, when and where the data transformations are performed. With ETL, the raw data is not available in the data warehouse because it is transformed before it is loaded. With ELT, the raw data is loaded into the data warehouse (or [data lake](https://www.techtarget.com/searchaws/definition/data-lake)) and transformations occur on the stored data. Staging areas are used for both ELT and ETL, but with ETL the staging areas are built into the ETL tool being used. With ELT, the staging area is in a database used for the data warehouse.

wow, ok, I assumed it was a typo because I've only ever seen ETL but apparently ELT is also a thing.

confusing. I thought it was maybe something else but no, it's reordering an acronym that is already well known to mean something subtly different in the process.

erezrokah (commented):
Hi @rngadam, thanks for the PR. As you linked in https://www.techtarget.com/searchdatamanagement/definition/Extract-Load-Transform-ELT there are plenty of resource that explain what ELT is and the differences from ETL.

We prefer not to duplicate that content. I'll go ahead and close the PR.
Please let me know if you have any concerns

PR #15430, opened 2023-11-22
feat: Allow reading spec(s) from stdin with `-`

disq:
Loading from stdin in addition to filesystem also works:
```
cat ./src.yml | ./cli sync - ./dest.yml
```
> Loading spec(s) from -, ./dest.yml

### DIFF
--- a/cli/cmd/migrate_test.go
+++ b/cli/cmd/migrate_test.go
@@ -35,7 +35,7 @@ func TestMigrate(t *testing.T) {
 		{
 			name:   "should fail with missing path error when path is missing",
 			config: "sync-missing-path-error.yml",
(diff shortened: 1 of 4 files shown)


erezrokah (commented):
What this feature requested by someone? Not sure if we want this added complexity especially when users can `cat` to a file instead

disq:
> What this feature requested by someone? Not sure if we want this added complexity especially when users can `cat` to a file instead

No, it's a nice to have feature. I just played around to see if it would be too bad code-wise.

erezrokah:
I would say let's skip it since it's easy to output STDIN to file. If we want to have it we'd need to document it, maintain it, etc.

PR #15415, opened 2023-11-22
fix!: Provide primary keys for tables (D)

candiduslynx:
Part of https://github.com/cloudquery/cloudquery/issues/15381

BEGIN_COMMIT_OVERRIDE
fix: Provide primary keys for tables (C) (https://github.com/cloudquery/cloudquery/pull/15415)
BREAKING-CHANGE: Provide primary keys for tables (C) (https://github.com/cloudquery/cloudquery/pull/15415). The following tables are affected:
* `aws_directconnect_gateway_associations`
* `aws_directconnect_gateway_attachments`
* `aws_docdb_cluster_parameters`
* `aws_docdb_event_categories`
* `aws_docdb_event_subscriptions`
* `aws_docdb_events`
* `aws_docdb_orderable_db_instance_options`
* `aws_docdb_pending_maintenance_actions`
* `aws_dynamodb_table_replica_auto_scalings`

END_COMMIT_OVERRIDE

Issue #15381, opened 2023-11-20
Verify all `_cq_id` primary keys in `plugins/source/aws`


candiduslynx:
```[tasklist]
### Tasks
- [ ] https://github.com/cloudquery/cloudquery/pull/15389
- [ ] https://github.com/cloudquery/cloudquery/pull/15400
- [ ] https://github.com/cloudquery/cloudquery/pull/15413
- [ ] https://github.com/cloudquery/cloudquery/pull/15415
- [ ] https://github.com/cloudquery/cloudquery/pull/15432
- [ ] https://github.com/cloudquery/cloudquery/pull/15434
- [ ] https://github.com/cloudquery/cloudquery/pull/15447
- [ ] https://github.com/cloudquery/cloudquery/pull/15468
- [ ] https://github.com/cloudquery/cloudquery/pull/15469
- [ ] https://github.com/cloudquery/cloudquery/pull/15464
- [ ] https://github.com/cloudquery/cloudquery/issues/15476
```

PR #15432, opened 2023-11-23
fix: Provide primary keys for tables (EC2)

candiduslynx:
Part of https://github.com/cloudquery/cloudquery/issues/15381

BEGIN_COMMIT_OVERRIDE
fix: Provide primary keys for tables (EC2) (https://github.com/cloudquery/cloudquery/pull/15415)
BREAKING-CHANGE: Provide primary keys for tables (EC2) (https://github.com/cloudquery/cloudquery/pull/15415). The following tables are affected:
* `aws_ec2_transit_gateway_*`

END_COMMIT_OVERRIDE

### DIFF
--- a/plugins/source/aws/resources/services/ec2/transit_gateway_vpc_attachments.go
+++ b/plugins/source/aws/resources/services/ec2/transit_gateway_vpc_attachments.go
@@ -10,7 +10,6 @@ import (
 	"github.com/cloudquery/cloudquery/plugins/source/aws/client"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/cloudquery/plugin-sdk/v4/transformers"
-	sdkTypes "github.com/cloudquery/plugin-sdk/v4/types"
 )
 
 func transitGatewayVpcAttachments() *schema.Table {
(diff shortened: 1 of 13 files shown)


plugins/source/aws/resources/services/ec2/transit_gateway_vpc_attachments.go
```diff
@@ -19,19 +18,21 @@ func transitGatewayVpcAttachments() *schema.Table {
 		Name:        tableName,
 		Description: `https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_TransitGatewayVpcAttachment.html`,
 		Resolver:    fetchEc2TransitGatewayVpcAttachments,
-		Transform:   transformers.TransformWithStruct(&types.TransitGatewayVpcAttachment{}),
+		Transform:   transformers.TransformWithStruct(&types.TransitGatewayVpcAttachment{}, transformers.WithResolverTransformer(client.TagsResolverTransformer)),
 		Columns: []schema.Column{
 			client.DefaultAccountIDColumn(false),
 			client.DefaultRegionColumn(false),
```
bbernays, line 24: The parent table uses arn + accountID + region... I think this is because TGW can be used cross account. So We should also add these fields in all nested tables

bbernays (approved):

PR #15433, opened 2023-11-23
fix: Add missing tables for Docker plugins, remove on-demand rendering of tables

erezrokah:
#### Summary

Direct links to Docker plugins tables are broken. Example https://docs.cloudquery.io/docs/plugins/sources/typeform/tables/typeform_forms

The plugins are hidden on the Website at the moment, but we kept the old content in an attempt to not break direct links.
It seems Nextra fails to identify the table files should be uploaded, so Vercel + Next.js fails to on demand render the, see https://vercel.com/cloudquery/docs/logs?page=1&timeline=past30Minutes&startDate=1700733152253&endDate=1700734952253&selectedLogId=1700734861599973971413875058&selectedLogTimestamp=1700734861599&forceEndDate=1700734896000

(It works locally if you do `next dev`).

This PR:
1.  Removes the on demand rendering, since it was used to avoid having a long build time. We don't need it anymore since Docker plugins have a small amount of tables. Changes to the makefile are intended to put the tables under each plugin instead of a root level `tables` directory (on demand rendering took the content from that directory)
2. Deletes some left over table files for plugins that we migrated to the Hub
3. Cleans up some sitemap generation code related to tables


hermanschaaf (approved):

plugins/source/bitbucket/testdata/config.yml
```diff
@@ -1,6 +1,6 @@
 kind: source
 spec:
-  name: bitbucket
+  name: tables
```
hermanschaaf, line 3: this doesn't look right :)

erezrokah: That's actually on purpose. The docs command uses the name attribute for the name of the directory the docs will be generated to

.github/workflows/source_typeform.yml
```diff
@@ -36,20 +36,6 @@ jobs:
           pip install -r requirements.txt
       - name: Check formatting
         run: make fmt-check
-      - name: Setup CloudQuery
```
erezrokah, line 39: Couldn't figure out why this is failing (wasn't able to reproduce by running the workflow locally).

We'll be deprecating this way of generating docs soon, so I think it's better to remove it for now

PR #15434, opened 2023-11-23
fix!: Streamline primary keys for `aws_ecr_repository_*`

candiduslynx:
Part of https://github.com/cloudquery/cloudquery/issues/15381

BEGIN_COMMIT_OVERRIDE
fix: Streamline primary keys for `aws_ecr_repository_*` (https://github.com/cloudquery/cloudquery/pull/15434)
BREAKING-CHANGE: Streamline primary keys for `aws_ecr_repository_*` (https://github.com/cloudquery/cloudquery/pull/15434). The following tables are affected:
* `aws_ecr_repository_image_scan_findings`
* `aws_ecr_repository_images`
* `aws_ecr_repository_lifecycle_policies`
* `aws_ecr_repositories`
* `aws_ecr_repository_policies`

END_COMMIT_OVERRIDE


plugins/source/aws/resources/services/ecr/repository_images.go
```diff
@@ -54,16 +51,3 @@ func fetchEcrRepositoryImages(ctx context.Context, meta schema.ClientMeta, paren
 	}
 	return nil
 }
-func resolveImageArn(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
candiduslynx, line 57: there's no such thing as image ARN according to the [docs](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonelasticcontainerregistry.html#amazonelasticcontainerregistry-resources-for-iam-policies)

plugins/source/aws/resources/services/ecr/repository_image_scan_findings.go
```diff
@@ -17,49 +17,88 @@ func repositoryImageScanFindings() *schema.Table {
 		Name:        "aws_ecr_repository_image_scan_findings",
 		Description: `https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_ImageScanFindings.html`,
 		Resolver:    fetchEcrRepositoryImageScanFindings,
-		Transform:   transformers.TransformWithStruct(&models.ImageScanWrapper{}),
+		Transform: transformers.TransformWithStruct(&ecr.DescribeImageScanFindingsOutput{},
+			transformers.WithSkipFields("ImageId", "NextToken", "ResultMetadata"),
```
bbernays, line 21: Why skip `ImageId`

plugins/source/aws/resources/services/ecr/repository_images.go
bbernays: Nice catch! But just so you know ARNs can exist beyond what is supported in IAM policies

plugins/source/aws/resources/services/ecr/repository_image_scan_findings.go
candiduslynx: [`imageId`](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_DescribeImageScanFindings.html#ECR-DescribeImageScanFindings-response-imageId) field is an object of [`ImageIdentifier`](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_ImageIdentifier.html) type. It will be a JSOn column (that we already have info about). We can leave this in the response, though, WDYT?

bbernays: I see no reason to drop data that is returned from the API. Lets try and stay as close as possible to the API, in the past when we have deviated, that is when issues have arisen especially when one user wants it one way and another wants it in a different way

candiduslynx: c4b934974e37b8028ed8810966f05feac440ff8c

plugins/source/aws/resources/services/ecr/lifecycle_policy.go
```diff
@@ -16,10 +17,16 @@ func lifeCyclePolicy() *schema.Table {
 		Name:        "aws_ecr_repository_lifecycle_policies",
 		Description: `https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetLifecyclePolicy.html`,
 		Resolver:    fetchRepositoryLifecyclePolicy,
-		Transform:   transformers.TransformWithStruct(&ecr.GetLifecyclePolicyOutput{}, transformers.WithPrimaryKeys("RepositoryName", "RegistryId"), transformers.WithSkipFields("ResultMetadata")),
```
disq, line 19: From what I understand from the godoc below, `RegistryId` can differentiate between duplicate ARNs, so it's best to have it in the PK? (registry id isn't included in the arn)

```
	// The Amazon Web Services account ID associated with the registry that contains
	// the repository.
	RegistryId *string

	// The Amazon Resource Name (ARN) that identifies the repository. The ARN contains
	// the arn:aws:ecr namespace, followed by the region of the repository, Amazon Web
	// Services account ID of the repository owner, repository namespace, and
	// repository name. For example,
	// arn:aws:ecr:region:012345678910:repository-namespace/repository-name .
	RepositoryArn *string
```

candiduslynx: I've addressed this in a5931bc293dfcc3067f14a9d0a36d765ffdaf7ca

candiduslynx: & 7f93bf84f3d451b546ca5033d9b6d020c842eb28

PR #15499, opened 2023-11-29
feat: Add `skip_answers` option to spec to have more privacy options

pilvikala:
We need to load typeform responses in an environment where we don't want to store any personal data. `skip_answers` will remove the answers column, thus allowing collecting general statistics, but not the actual answers.


erezrokah (commented):
I don't mind approving but usually CloudQuery doesn't transform the data and saves it as raw as possible in the DB. Then you can write a query that skips the `answers` column for example

pilvikala:
I understand. The point of this change is to prevent the personal data from arriving in the destination. Is there another way to achieve that with CloudQuery? Should we actually merge this, if it goes against the current principles?

erezrokah:
> Should we actually merge this, if it goes against the current principles?

Closing the PR. We can run a query post sync to remove PII if needed. Please comment if we need to re-open

PR #15555, opened 2023-12-05
fix: Adjust resolvers to avoid errors in AWS tables

bbernays:


plugins/source/aws/resources/services/wellarchitected/lenses.go
```diff
@@ -80,6 +80,10 @@ func getLens(ctx context.Context, meta schema.ClientMeta, resource *schema.Resou
 	if summary.LensType == types.LensTypeAwsOfficial {
 		input.LensVersion = nil // official lenses don't support versions
 	}
+	if l.LensAlias == nil {
+		// LensAlias is required so if it's nil we can't do anything
+		return nil
```
disq, line 85: At least `resource.SetItem(l)` so that type on the ending resource ends up matching what we specified?

disq (approved):

PR #15556, opened 2023-12-06
feat: Add Source Pagerduty Oncalls resource

spangenberg:
#### Summary

Currently the Pagerduty source plugin doesn't support the actual Oncall resource which is kinda the most important one :)


erezrokah (commented):
Thanks for the PR @spangenberg looks good 🚀 

Looking at the [API docs](https://developer.pagerduty.com/api-reference/3a6b910f11050-list-all-of-the-on-calls) I have a few questions

1. Should we set the `include` array to get `escalation_policies, users, schedules`? Or is that the default already
2. Should we set `time_zone` to UTC? That would make it easier to query the data I think, otherwise you'd need to know the account timezone
3. Both `since` and `until` default to current time. That seems OK as the API will return any on calls that overlap with those. Not sure if we want a bit of buffer to be safe
4. Is there anything we can use as a primary key for the table? Based on the API docs I couldn't find a valid option (even a combination of `user.id, schedule.id, escalation_policy.id, start, end` won't work I think)

spangenberg:
> 1. Should we set the `include` array to get `escalation_policies, users, schedules`? Or is that the default already?

Since this API Operation seems to have special rate limiting I tried to stay as lean as possible: `This API operation has operation specific rate limits.`
If include is not set if will just contain the self references which can be used to join the corresponding tables. Also if I include all options it takes way longer to fetch.

e.g.
```
JOIN pagerduty_escalation_policies ON pagerduty_escalation_policies.NAME = ?
JOIN pagerduty_oncalls ON pagerduty_oncalls.escalation_policy ->> 'id' = pagerduty_escalation_policies.ID
JOIN pagerduty_users ON pagerduty_users.ID = pagerduty_oncalls.USER ->> 'id'
```

> 2. Should we set `time_zone` to UTC? That would make it easier to query the data I think, otherwise you'd need to know the account timezone

Absolutely

> 3. Both `since` and `until` default to current time. That seems OK as the API will return any on calls that overlap with those. Not sure if we want a bit of buffer to be safe

I don't have a strong opinion here but I suspect that not setting anything here resembles the data similar to as one would use the UI.

> 4. Is there anything we can use as a primary key for the table? Based on the API docs I couldn't find a valid option (even a combination of `user.id, schedule.id, escalation_policy.id, start, end` won't work I think)

I don't think that's possible, for that reason @shimonp21 also skipped the resource while adding the plugin: https://github.com/cloudquery/cloudquery/pull/5547
> OnCall resources require, due to the fact that the IDs of all nested structs should be PKs. https://github.com/PagerDuty/go-pagerduty/blob/b6aa0d76f875e1de499b2a246c82882f022a0f8a/on_call.go#L26

erezrokah (approved):
Thanks for the follow up @spangenberg and the additional context. Looks great 🚀

erezrokah:
Thanks @spangenberg, released in [`v3.1.0`](https://hub.cloudquery.io/plugins/source/cloudquery/pagerduty/v3.1.0/docs)

PR #15561, opened 2023-12-06
fix: Login with no browser

disq:
Should fix https://github.com/cloudquery/cloudquery-issues/issues/1019 (internal issue)

Another issue is, the token shown in the frontend doesn't seem to work for me when I paste it, but if I capture the "callback" token and paste that it works.


cli/cmd/login.go
```diff
@@ -129,21 +130,37 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 
 	url := accountsURL + "?returnTo=" + localServerURL + "/callback"
 	if err := browser.OpenURL(url); err != nil {
-		fmt.Printf("Failed to open browser at %s. Please open the URL manually.\n", accountsURL)
+		fmt.Printf("Failed to open browser. Please open %s manually and paste the token below:\n", accountsURL)
+		oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
```
disq, line 134: Without these, any type of reading (bufio scanner/reader, scanf, io.ReadAll with CTRL+D chaser) hangs after 1024 characters for me.

cli/cmd/login.go
```diff
@@ -129,21 +130,37 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
...
+			fmt.Println("Context cancelled. Shutting down server.")
+		case <-gotToken:
+		}
 	}
 
 	if refreshToken == "" {
 		return fmt.Errorf("failed to get refresh token")
 	}
+
+	fmt.Println("Authenticating...")
```
disq, line 162: Just thought some kind of feedback after the input/click is complete, before we hit the server would be nice...

erezrokah (approved):

PR #15600, opened 2023-12-11
chore: Remove old links from docs

yevgenypats:
we had old links. I think now those are not needed and we can just link to home as all the guides are under the blog. I can also add a solutions section just like in our home but it will mean we will need to maintain both which might be a bit annoying.


website/pages/solutions/open-source-cspm.mdx
erezrokah, line 1: @yevgenypats I added a direct for pricing to `https://www.cloudquery.io/pricing` via https://github.com/cloudquery/cloudquery/pull/15600/commits/2bd74c8e6ea6ef2f9c0fc9d998d7abfaba6a9dab.

Where should we redirect `open-source-cspm`? Could not find it in https://www.cloudquery.io/ under solutions

yevgenypats: We should redirect it here - https://www.cloudquery.io/solutions/for-cloud-teams . Im also opening another ticket to move this https://docs.cloudquery.io/solutions/open-source-cspm to blog.

erezrokah (approved):

website/pages/solutions/open-source-cspm.mdx
erezrokah: Fixed in https://github.com/cloudquery/cloudquery/pull/15600/commits/fc293a449fd358a7eb7ba333f4ee3b80f4113fb8, I also removed the link to pricing from the footer https://github.com/cloudquery/cloudquery/pull/15600/commits/1ec4e2a0669eb615a034ed0b7524c0057590702d (we can do a follow up if we need to remove more links)

PR #15698, opened 2023-12-14
feat: Support publishing to a private docker registry

erezrokah:
#### Summary

Example (images generated with https://github.com/cloudquery/plugin-sdk-javascript/pull/106)
image

The images were generated by https://github.com/cloudquery/plugin-sdk-javascript/pull/106 (replacing `registry.cloudquery.io` with `ghcr.io`)

### TODO

- [ ] Merge the different platforms into a single image with `docker manifest`. The is done by `docker buildx --push` automatically at the moment in https://github.com/cloudquery/cloudquery/blob/67c10c38a04dcdd1512bf6dc739f89bc11baa888/.github/workflows/release_plugin.yml#L94


cli/internal/publish/plugins.go
```diff
@@ -184,3 +218,89 @@ func UploadPluginBinary(ctx context.Context, c *cloudquery_api.ClientWithRespons
...
+		return fmt.Errorf("failed to parse docker load response: %v", err)
+	}
+	if loadResponse.Stream != "" {
+		fmt.Print(loadResponse.Stream)
+	}
+
+	return nil
+}
+
+func PublishToDockerRegistry(ctx context.Context, token, distDir string, pkgJSON PackageJSONV1) error {
```
erezrokah, line 267: This is the bulk of the change

cli/internal/publish/plugins.go
```diff
@@ -26,10 +34,95 @@ type PackageJSONV1 struct {
...
+		progressbar.OptionOnCompletion(func() {
+			fmt.Fprint(os.Stdout, "\n")
+		}),
+		progressbar.OptionSpinnerType(14),
+		progressbar.OptionFullWidth(),
+		progressbar.OptionSetRenderBlankState(true),
+	)
+}
+
+func (pr *dockerProgressReader) Read(_ []byte) (n int, err error) {
```
erezrokah, line 90: Based on https://github.com/cloudquery/plugin-pb-go/blob/7eb82cb9204b9d1ce0d8652b2312d61ab4243ba9/managedplugin/docker.go#L32

cli/cmd/plugin_publish.go
```diff
@@ -83,6 +82,7 @@ func runPluginPublish(ctx context.Context, cmd *cobra.Command, args []string) er
 			return errors.New("invalid plugin name. Must be in format <team_name>/<plugin_name>")
 		}
 		teamName, pluginName = parts[0], parts[1]
+		pkgJSON.Team, pkgJSON.Name = teamName, pluginName
```
erezrokah, line 85: This is backwards compatibility code from when we had `plugin publish <team>/<name>` instead of inside `package.json`.

I cleaned up some of the arguments to `UploadPluginBinary` to pass only the package JSON so we need to set these here

erezrokah:
https://github.com/cloudquery/cloudquery/pull/15698/commits/7a5784c732e94867e4e0e96d52afd0434199a831 makes it so we merge all platforms into a single image:
image

cli/internal/publish/plugins.go
```diff
@@ -26,10 +43,115 @@ type PackageJSONV1 struct {
...
+
+type transportWithRegistryAuth struct {
+	http.RoundTripper
+	baseTransport *http.Transport
+	registryAuth  string
+}
+
+func newTransportWithRegistryAuth(insecureSkipVerify bool, registryAuth string) *transportWithRegistryAuth {
+	baseTransport := http.DefaultTransport.(*http.Transport).Clone()
+	baseTransport.TLSClientConfig.InsecureSkipVerify = insecureSkipVerify
```
maaarcelino, line 85: I think we want to remove this for prod, right?

cli/internal/publish/plugins.go
```diff
@@ -26,10 +43,115 @@ type PackageJSONV1 struct {
...
+	return &transportWithRegistryAuth{
+		baseTransport: baseTransport,
+		registryAuth:  registryAuth,
+	}
+}
+
+func (t *transportWithRegistryAuth) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+	req.Header.Set("Authorization", "Bearer "+t.registryAuth)
```
maaarcelino, line 93: Which endpoints is this for? The Docker auth expects `Basic` not `Bearer`.

cli/internal/publish/plugins.go
erezrokah: Only done for `localhost` images https://github.com/erezrokah/cloudquery/blob/6086d56a7a4f9d07bfc2e8ec2684776cad1537cb/cli/internal/publish/plugins.go#L598

cli/internal/publish/plugins.go
erezrokah: The code does basic auth with the auth endpoint to get an access token which then it uses as a `Bearer` token with the registry. This is described in https://distribution.github.io/distribution/spec/auth/token/#how-to-authenticate.

The TLDR: you start with the request you'd like to do, then if auth is required you get a challenge back to do the basic auth. With the basic auth you get a bearer token you can perform actions wiht.

1. Challenge request https://github.com/erezrokah/cloudquery/blob/6086d56a7a4f9d07bfc2e8ec2684776cad1537cb/cli/internal/publish/plugins.go#L389
2. Do basic auth https://github.com/erezrokah/cloudquery/blob/6086d56a7a4f9d07bfc2e8ec2684776cad1537cb/cli/internal/publish/plugins.go#L417
3. Get token from auth response https://github.com/erezrokah/cloudquery/blob/6086d56a7a4f9d07bfc2e8ec2684776cad1537cb/cli/internal/publish/plugins.go#L435
4. Use token to contact registry https://github.com/erezrokah/cloudquery/blob/6086d56a7a4f9d07bfc2e8ec2684776cad1537cb/cli/internal/publish/plugins.go#L468

maaarcelino (approved):

PR #15703, opened 2023-12-14
fix: OU Filtering in AWS Org mode

bbernays:
#### Summary

Currently OU Filtering is only used when a user specifies an `OU`. This could be either an `ou-` or the `root`. With this change the plugin will resolve all of the Accounts in the OU (and below) and then exclude them from the sync


hermanschaaf (approved):
LGTM 👍  Just a tiny thing about the test case name (it's currently the same as the test above it)

PR #15708, opened 2023-12-15
chore(resources): Validate empty JSON Columns

disq:
Uses https://github.com/cloudquery/plugin-sdk/pull/1418 and fixes some of the current tests.


plugins/source/aws/client/testing.go
```diff
@@ -56,7 +58,15 @@ func AwsMockTestHelper(t *testing.T, parentTable *schema.Table, builder func(*te
 		t.Fatal(err)
 	}
 
-	plugin.ValidateNoEmptyColumns(t, tables, messages)
+	for _, table := range tables.FlattenTables() {
+		cols, _ := testOpts.SkipEmptyCheckColumns[table.Name]
+		if len(cols) > 0 {
+			continue // TODO ignore these columns but check the rest
```
disq, line 64: This needs a later fix... Maybe something like
`plugin.ValidateNoEmptyColumns(t, schema.Tables{table}, messages, cols...)` (`except` columns)

plugins/source/aws/resources/services/s3/buckets_mock_test.go
```diff
@@ -84,5 +84,7 @@ func buildS3Buckets(t *testing.T, ctrl *gomock.Controller) client.Services {
 }
 
 func TestS3Buckets(t *testing.T) {
-	client.AwsMockTestHelper(t, Buckets(), buildS3Buckets, client.TestOptions{})
+	client.AwsMockTestHelper(t, Buckets(), buildS3Buckets, client.TestOptions{
+		SkipEmptyCheckColumns: map[string][]string{"aws_s3_bucket_notification_configurations": {"event_bridge_configuration"}},
```
disq, line 88: This is an empty struct so the JSON comes up as empty.

plugins/source/aws/resources/services/kafka/clusters_mock_test.go
```diff
@@ -38,5 +38,7 @@ func buildKafkaClustersMock(t *testing.T, ctrl *gomock.Controller) client.Servic
 	}
 }
 func TestKafkaClusters(t *testing.T) {
-	client.AwsMockTestHelper(t, Clusters(), buildKafkaClustersMock, client.TestOptions{})
+	client.AwsMockTestHelper(t, Clusters(), buildKafkaClustersMock, client.TestOptions{
+		SkipEmptyCheckColumns: map[string][]string{"aws_kafka_cluster_operations": {"tags"}},
```
disq, line 42: Here `tags` field is redundant and is always empty. The resolver ignores this but the field ends up empty.

erezrokah (approved):
Let's make this a `chore:` if it impacts only tests

PR #15710, opened 2023-12-15
chore(resources): Validate empty JSON Columns

disq:
Uses https://github.com/cloudquery/plugin-sdk/pull/1418 and fixes a single test with missing field.


erezrokah (approved):
Let's make this a `chore:` if it impacts only tests

Issue #15727, opened 2023-12-18
bug: AWS Lambda Function tags are not being set


jutley:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Row in the table `aws_lambda_functions` have their `tags` column set to null, despite having tags defined.

### Expected Behavior

Tags actually show up.

### CloudQuery (redacted) config

```yaml
---
kind: source
spec:
  # Source spec section
  name: aws
  path: cloudquery/aws
  version: "v23.2.0"
  skip_dependent_tables: true
  concurrency: 10000
  tables:
    - aws_acm_certificates
    - aws_athena_work_groups
    - aws_autoscaling_groups
    - aws_cloudtrail_trails
    - aws_cloudwatchlogs_log_groups
    - aws_dynamodb_tables
    - aws_dynamodb_global_tables
    - aws_ec2_instances
    - aws_elasticache_clusters
    - aws_elasticsearch_domains
    - aws_elbv2_load_balancers
    - aws_lambda_functions
    - aws_glue_databases
    - aws_glue_jobs
    - aws_guardduty_detectors
    - aws_organizations_accounts
    - aws_rds_clusters
    - aws_rds_instances
    - aws_redshift_clusters
    - aws_route53_domains
    - aws_route53_hosted_zone_resource_record_sets
    - aws_s3_buckets
    - aws_securityhub_hubs
    - aws_sqs_queues
    - aws_support_cases
    - aws_support_case_communications
    - aws_transfer_servers
  destinations:
    - postgresql
  spec:
    aws_debug: false
    org: <redacted>
```

### Steps To Reproduce

1. Have a Lambda with tags
2. Run Cloudquery with the aws_lambda_functions table enabled
3. Check the tags. They will be null.

### CloudQuery (redacted) logs

```
2023-12-18T21:12:53Z INF top level table resolver started client=<account_id>:eu-west-1 module=aws-src table=aws_lambda_functions
2023-12-18T21:12:54Z WRN location of lambda function redacted for security purposes account_id=<account_id> module=aws-src region=eu-west-1
2023-12-18T21:12:54Z WRN location of lambda function redacted for security purposes account_id=<account_id> module=aws-src region=eu-west-1
2023-12-18T21:12:54Z INF table sync finished client=<account_id>:eu-west-1 errors=0 module=aws-src resources=2 table=aws_lambda_functions
```

### CloudQuery version

4.3.4

### Additional Context

[...]

PR #15728, opened 2023-12-18
fix: Update lambda functions resolveTags() to use Tags field instead of Concurrency

jutley:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/15727. Currently Lambda tags don't make it into the table due to a simple bug.

~I haven't tested this, and I'm not really sure how at the moment. There aren't already any unit tests I can leverage and I don't know how to install a custom plugin. Since this is such a simple change, I figured it was worth a shot submitting it anyway.~

I found the docs to test locally and used CloudQuery to write to a local file. The difference between my local plugin and the upstream plugin showed that this change worked as expected!

### DIFF
--- a/plugins/source/aws/resources/services/lambda/functions.go
+++ b/plugins/source/aws/resources/services/lambda/functions.go
@@ -306,7 +306,7 @@ func resolveTags(ctx context.Context, meta schema.ClientMeta, resource *schema.R
 
 	// setting tags value from GetFunction call
 	if r.Code != nil {
-		return resource.Set(col.Name, r.Concurrency)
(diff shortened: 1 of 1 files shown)


bbernays:
@jutley - Thank you for this PR. We can merge it as it is, the test coverage doesn't check for the value in the column only that every column is set.

You can see in this PR how we actually mock the API calls to be run in tests: #15729 

Instructions on how you can build a plugin and run it locally  can be found here: https://docs.cloudquery.io/docs/developers/running-locally

bbernays (approved):

PR #15729, opened 2023-12-18
chore: Tags for Lambda Functions properly set

bbernays:
#### Summary

The happy path of `GetFunction` succeeding means that `ListTags` should not be called. This PR adds tests to ensure all paths are properly implemented


erezrokah (approved):
Looks good. Changed this to a `chore` as it impacts only the tests

PR #15783, opened 2023-12-22
chore: Fix GCP Testing

bbernays:
#### Summary

- Prior to this PR we had 2 methods for testing `MockTestRestHelper` and `MockTestGrpcHelper` this PR combines all of the logic into 1 function where each test can pass in the servers that it needs.
- `MockTestGrpcHelper` did not run `TransformTables` so some of the tests were passing erroneously


plugins/source/gcp/resources/plugin/tables.go
```diff
@@ -54,28 +54,28 @@ import (
 
 func PluginAutoGeneratedTables() schema.Tables {
 	tables := []*schema.Table{
-		aiplatform.JobLocations(),
```
erezrokah, line 57: Can we do this change in a separate PR? Also this is a user facing change as it impacts the order of how tables are synced so should probably be a `fix`

plugins/source/gcp/resources/services/aiplatform/operations.go
```diff
@@ -30,6 +30,11 @@ func Operations() *schema.Table {
 				Resolver:   client.ResolveProject,
 				PrimaryKey: true,
 			},
+			// {
```
erezrokah, line 33: Should we remove this comment?

plugins/source/gcp/resources/services/aiplatform/pipeline_jobs.go
```diff
@@ -46,7 +44,7 @@ func fetchPipelineJobs(ctx context.Context, meta schema.ClientMeta, parent *sche
 	}
 
 	clientOptions := c.ClientOptions
```
erezrokah, line 46: What's the reason this code is commented out?

plugins/source/gcp/resources/services/aiplatform/pipeline_locations.go
```diff
@@ -45,7 +43,7 @@ func fetchPipelineLocations(ctx context.Context, meta schema.ClientMeta, parent
 	}
 
 	clientOptions := c.ClientOptions
-	clientOptions = append([]option.ClientOption{option.WithEndpoint("us-central1-aiplatform.googleapis.com:443")}, clientOptions...)
```
erezrokah, line 48: What's the reason this code is commented out?

erezrokah (changes requested):
Great change. I have some comments on code changes that are not related to the tests.
Also please note this will mostly like break the sync to the private version of the plugin (we'll have to update tests for paid tables)

plugins/source/gcp/resources/services/aiplatform/pipeline_jobs.go
bbernays: it should not be. fixed in https://github.com/cloudquery/cloudquery/pull/15783/commits/5a7f6645af55ed6553eddde543466a3d3d58c15f

plugins/source/gcp/resources/services/aiplatform/operations.go
bbernays: Removed in https://github.com/cloudquery/cloudquery/pull/15783/commits/de0a35bc5bd4fe949ffcd19f4bc9daeabd70a79c

plugins/source/gcp/resources/plugin/tables.go
bbernays: reverted in https://github.com/cloudquery/cloudquery/pull/15783/commits/0e8d9b0cf065b18f5905f908bcfc3659bad886b1

erezrokah (approved):

PR #15794, opened 2023-12-25
fix: RDS Snapshot Attribute Throttling

bbernays:
#### Summary

Currently the AWS RDS Snapshot attributes are resolved in a column resolver, this means that it uses the Resource Semaphore to limit concurrency, right now that value is set at a minimum of 100 concurrent resolvers. In a future version we may want to switch this to be a nested relation so that we can take advantage of concurrency at the multiplexer level, but in the meantime this fix is a backwards compatible work around to ensure that throttling does not occur. This may cause some performance slow down as API requests will be handled sequentially, but it also might speed things up because little to no throttling should occur as there won't be up to 100 concurrent API requests occuring each backing off up to 30 seconds 10 times.


plugins/source/aws/resources/services/rds/db_snapshots.go
```diff
@@ -55,40 +56,61 @@ func fetchRdsDbSnapshots(ctx context.Context, meta schema.ClientMeta, parent *sc
 		if err != nil {
```
hermanschaaf, line 56: I might be missing something, but shouldn't the column resolver on line 41 be removed?

bbernays: If I do that, then the order of the columns changes, which I was trying to avoid

hermanschaaf: Ah I see okay. We don't consider column ordering changes breaking but I guess it's better if we can not change it at all

hermanschaaf (approved):

PR #15800, opened 2023-12-27
fix: Call `GetPolicyVersion` to get PolicyDocument

bbernays:
#### Summary

Tested this changes coming in `v4.23.0` of plugin-sdk and no throttling issues while in current version of scheduler there are throttling issues


plugins/source/aws/resources/services/iam/policy_versions.go
```diff
@@ -57,6 +58,30 @@ func fetchPolicyVersion(ctx context.Context, meta schema.ClientMeta, parent *sch
...
+		&iam.GetPolicyVersionInput{
+			PolicyArn: policy.Arn,
+			VersionId: pv.VersionId},
+		func(options *iam.Options) {
+			options.Region = cl.Region
+		},
+	)
+	if err != nil {
+		cl.Logger().Warn().Err(err).Msg("Failed to get policy version")
+		resource.SetItem(pv)
```
disq, line 76: `resource.Item` is already `pv`, why do we need this?

yevgenypats (approved):

disq (approved):

PR #15854, opened 2023-12-29
fix: Streamline `aws_guardduty_` tables primary keys

candiduslynx:
Follow-up for https://github.com/cloudquery/policies-premium/pull/264

ARNs are constructed based on https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonguardduty.html#amazonguardduty-resources-for-iam-policies

Based on the [`CreateDetector` API documentation](https://docs.aws.amazon.com/guardduty/latest/APIReference/API_CreateDetector.html) we don't need to use `request_account_id` & `request_region` columns, but rather their plain variants (`account_id` & `region` respectively).
The only issue is `aws_guardduty_detector_members` table, where `account_id` is a part of the resource itself. It's solved by renaming the resource `account_id` column to `member_account_id`.


bbernays (changes requested):
I think some of the PKs need to change. But I wonder if we can even make this backwards compatible by leaving the old columns in place and just adding the new ones?

candiduslynx (commented):
@bbernays I've brought this PR up to speed after recent transition to the CQ ID as PK.

Kindly review.

bbernays (approved):

PR #15952, opened 2024-01-02
Add pendo source plugin

mkorolyov:
Adds support of pendo.io as a source plugin.


yevgenypats:
Hey @mkorolyov , thanks for the PR! This looks great. Per our previous chat it is best if you put this code in a private repo and publish this to hub.cloudquery.io from there. As this would be a paid plugin and you will be maintaining it. I'd be happy to chat here or on upwork for additional help

PR #16001, opened 2024-01-03
refactor: Extract common plugin spec fields into metadata struct

candiduslynx:
Extracted from https://github.com/cloudquery/cloudquery/pull/15998

### DIFF
--- a/cli/internal/specs/v0/metadata.go
+++ b/cli/internal/specs/v0/metadata.go
@@ -36,7 +36,21 @@ func (m *Metadata) Validate() error {
 	if m.Path == "" {
 		msg := "path is required"
 		// give a small hint to help users transition from the old config format that didn't require path
-		officialPlugins := []string{"aws", "azure", "gcp", "k8s", "postgresql", "clickhouse"}
+		officialPlugins := []string{
+			"aws",
+			"azure",
+			"gcp",
+			"digitalocean",
+			"github",
+			"heroku",
+			"k8s",
+			"okta",
+			"terraform",
+			"cloudflare",
+			"postgresql",
+			"csv",
+			"clickhouse",
+		}
 		if slices.Contains(officialPlugins, m.Name) {
 			msg += fmt.Sprintf(". Hint: try setting path to cloudquery/%s in your config", m.Name)
 		}
--- a/cli/internal/specs/v0/metadata.go
+++ b/cli/internal/specs/v0/metadata.go
@@ -14,7 +14,7 @@ type Metadata struct {
 	// Version of the plugin to be used
 	Version string `json:"version"`
 
-	// Path is the canonical path to the source plugin in a given registry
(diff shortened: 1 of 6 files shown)


cli/internal/specs/v0/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Spec part to define exact plugin: name, version & location.
+type Metadata struct {
+	// Name of the plugin to use
+	Name string `json:"name" jsonschema:"required,minLength=1"`
+
+	// Version of the plugin to be used
+	Version string `json:"version"`
+
+	// Path is the canonical path to the source plugin in a given registry
```
hermanschaaf, line 17:
```suggestion
	// Path is the canonical path to the plugin in a given registry
```

cli/internal/specs/v0/source.go
```diff
@@ -95,36 +77,15 @@ func (s *Source) UnmarshalSpec(out any) error {
...
-		if !strings.HasPrefix(s.Version, "v") {
-			return fmt.Errorf("version must start with v")
-		}
-	}
 	if len(s.Destinations) == 0 {
 		return fmt.Errorf("at least one destination is required")
 	}
 
-	return nil
+	return s.Metadata.Validate()
```
hermanschaaf, line 88: This slightly changes the ordering of the validation steps, let's rather keep it as similar as possible to how it was before and validate `s.Metadata` first, returning an error if it fails

cli/internal/specs/v0/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (m *Metadata) Validate() error {
+	if m.Name == "" {
+		return fmt.Errorf("name is required")
+	}
+
+	if m.Path == "" {
+		msg := "path is required"
+		// give a small hint to help users transition from the old config format that didn't require path
+		officialPlugins := []string{"aws", "azure", "gcp", "k8s", "postgresql", "clickhouse"}
```
hermanschaaf, line 39: It seems like we lost a number of plugins from the list in the refactor here. The full list should be:

```
[]string{"aws", "azure", "gcp", "digitalocean", "github", "heroku", "k8s", "okta", "terraform", "cloudflare, "postgresql", "csv"}
```

candiduslynx: 66a5f405caf90330917a2601d8ed16309035f153

candiduslynx: 66a5f405caf90330917a2601d8ed16309035f153

cli/internal/specs/v0/source.go
candiduslynx: 76e952ca9debfff063b2632abf4c45e112292bda

hermanschaaf (approved):

disq (approved):

PR #16033, opened 2024-01-04
feat: Upload images from markdown files to CDN when publishing

disq:
Limited testing results:

```
# Test Addon README
![Troll face](https://storage.googleapis.com/cq-cloud-images/kemal/6027b3c3f2afcb2cf2c9405637a40690adb9a675_troll.png "Trollface")
## Installation
![](https://storage.googleapis.com/cq-cloud-images/kemal/4b0b26dd0e5113344c3d655dce252c8778ffadd3_tldr.png)
![](https://storage.googleapis.com/cq-cloud-images/kemal/bf66c0a36876f9854f80883c609694a9ec653dac_1000.png)
## One more heading
![Troll face 2](https://storage.googleapis.com/cq-cloud-images/kemal/6027b3c3f2afcb2cf2c9405637a40690adb9a675_troll.png)
```

Newly supported:
- ~HTML `` tags in markdown (will try to fix this, but need to make sure it doesn't replace inside code blocks)~
- ~Reference-style image syntax (with ids, documented [here](https://daringfireball.net/projects/markdown/syntax#img))~


cli/go.mod
```diff
@@ -142,3 +143,5 @@ require (
 
 // github.com/cloudquery/jsonschema @ cqmain
 replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0-20231018073309-6c617a23d42f
+
+replace github.com/cloudquery/cloudquery-api-go => github.com/cloudquery/cloudquery-api-go v1.6.4-0.20240104155404-515b000bd747
```
disq, line 147: Using untagged for now

hermanschaaf: what's the reason for this? let's release the next version if that's what you need

cli/internal/publish/images.go
```diff
@@ -0,0 +1,367 @@
...
+			}
+			// test = append(append(test[:ir.startPos], bytes.Repeat([]byte("!"), ir.endPos-ir.startPos)...), test[ir.endPos:]...)
+			for i := ir.startPos; i < ir.endPos; i++ {
+				test[i] = '!'
+			}
+
+			ir := ir
+			reps = append(reps, ir)
+		}
+	}
```
hermanschaaf, line 146: I think this could be simplified / improved by using sweep-line approach, at least in terms of memory & efficiency:

 1. create one list of all important positions (start / end) `struct {pos int, start bool}`, sort it by position. while doing this, check that start < end in all cases (error out if not)
 2. keep track of number of open ranges (`open := 0`)
 3. iterate over list; if start node then `open++`, if end node `open--`. If open is > 1, we have overlapping ranges and should error out.

cli/internal/publish/images.go
```diff
@@ -0,0 +1,367 @@
...
+			startPos, endPos int
+		)
+		switch el := n.(type) {
+		case *ast.Image:
+			imgRef = string(el.Destination)
+			p := el.BaseNode.Parent()
+			for p != nil {
+				if p.Kind() == ast.KindParagraph {
+					for i := 0; i < p.Lines().Len(); i++ {
+						// we can have multiple lines in a paragraph, so we need to check each one to match destination/title, hoping there are no dupes
```
hermanschaaf, line 272: you probably saw this question coming, but what happens if there are dups? 😄

cli/internal/publish/images_test.go
```diff
@@ -0,0 +1,220 @@
...
+	if filepath.Separator == '\\' {
+		specialAbsEscaped = "/" + specialAbsEscaped // absolute paths in file:// mode start with "/", not C:\ (`file:///C:/`, not `file://C:/`)
+	}
+
+	const mockURL = "https://example.com/file.ext"
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			tc.contents = strings.ReplaceAll(tc.contents, "${ABS_IMAGE}", specialAbsEscaped)
+			out, err := findMarkdownImages(tc.contents, tempdir)
```
hermanschaaf, line 178: It feels like it would simplify the validation of test cases if we tested `findMarkdownImages` and `replaceMarkdownImages` in separate tests?

Because if I'm reading this test right, you need to now adjust the startPos and endPos to what it will be *after* replacement, rather than before. I'm also a bit confused by how endPos can be negative, maybe I misunderstand what it's doing 🤔

cli/internal/publish/images.go
```diff
@@ -0,0 +1,367 @@
...
+
+type imageReference struct {
+	ref     string // image filename (to replace with URL)
+	absFile string // absolute path to image file, to upload
+	url     string // result of upload
+
+	startPos, endPos int
+}
+
+var htmlImageRe = regexp.MustCompile(`<img\s+(?:[^>]*?\s+)?src="([^"]*)"`)
```
hermanschaaf, line 35: Am I right in thinking there are some cases this can't handle yet, like `" src="sda">` and maybe `<div></div>`?

hermanschaaf: We could use https://pkg.go.dev/golang.org/x/net/html maybe

hermanschaaf (commented):
Did a first pass review; I think overall it looks good enough for a first version. Most of my comments are just suggestions for possible improvement, perhaps worth some comments in the code at least if we don't want to improve them now.

My two main comments are:
 - the one about the `replace` directive on `github.com/cloudquery/cloudquery-api-go` before merging
 - the one about the test cases seeming hard to write / validate, I think it would make sense to split the tests

cli/internal/publish/images.go
disq: We'll replace the first one we find and leave the other(s).

cli/internal/publish/images_test.go
disq: No the positions are before replacement, except for the ABS_IMAGE case, which we don't and can't know the exact path before running the test due to the nature of temp dirs.

cli/internal/publish/images.go
disq: I don't think we should make the effort to support that first one, it's supposed to be a markdown document after all.

" src="https://storage.googleapis.com/cq-cloud-images/kemal/6027b3c3f2afcb2cf2c9405637a40690adb9a675_troll.png">

Does github support it?
And the second one might just work, but it's still a lot of html in a markdown. Using `` in markdown is allowed only because there are some attributes you might want to set, like `width`.

cli/go.mod
disq: the reason was being unsure if anything else (like content type) was needed in the request :) will release shortly.

cli/internal/publish/images.go
disq: Done in [`3e8252a` (#16033)](https://github.com/cloudquery/cloudquery/pull/16033/commits/3e8252a5c51e0fd87305a3cb4a2c36ddcfaeda4b)

hermanschaaf (approved):

cli/internal/publish/images_test.go
hermanschaaf: Why are positions negative sometimes?

cli/internal/publish/images.go
disq: Fixed this case in https://github.com/cloudquery/cloudquery/pull/16033/commits/143410ac2835bc377df7719ae5857ff65827f1bf

cli/internal/publish/images.go
disq: First case fixed in https://github.com/cloudquery/cloudquery/pull/16033/commits/56421e7a110bc2e6587e132aa8efe35d0230d970
Second case was previously fixed

cli/internal/publish/images_test.go
disq: They aren't really, clarified in https://github.com/cloudquery/cloudquery/pull/16033/commits/aeb4f25e986159dc469a9b205a45110d798a819d

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+	return contents, nil
+}
+
+func findMarkdownImages(contents, docDir string) (map[string][]imageReference, error) {
+	imf := &imageFinder{
+		docDir: docDir,
+	}
+	p := goldmark.New(
+		goldmark.WithParserOptions(
+			parser.WithASTTransformers(util.Prioritized(imf, 999999)),
```
disq, line 114: I'm not sure if this priority matters, or if it should be lowest

candiduslynx (commented):
just some suggestions

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+
+	fmt.Println("Preparing to upload images...")
+
+	reqs := make([]cloudquery_api.TeamImageCreate, 0, len(ims))
+	for k := range ims {
+		refParts := strings.SplitN(k, ":", 2)
+		checksum, name := refParts[0], refParts[1]
+
+		reqs = append(reqs, cloudquery_api.TeamImageCreate{
+			Name:     name,
```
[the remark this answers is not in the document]
disq: You're correct it's solved elsewhere (see the relevant PR in the backend repo) the backend will do a HEAD and if there is an object on the same 'calculated path' it won't return an upload url.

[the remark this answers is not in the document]
disq: File size limits are also not a discussion for clients, we also have the ability to lock out each team from uploading.

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+	return imf.images, nil
+}
+
+func convertMarkdownReferences(ims map[string][]imageReference) ([]imageReference, error) {
+	type pos struct {
+		at  int
+		end bool
+	}
+	reps := make([]imageReference, 0, len(ims))
+	posList := make([]pos, 0, 2*len(ims))
```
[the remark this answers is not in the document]
disq: 2* makes sense here as each reference has two positions.

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+		if p.end {
+			open--
+		} else {
+			open++
+		}
+		if open > 1 {
+			return nil, fmt.Errorf("found overlapping range: %d-%d", lastPos, p.at)
+		}
+		lastPos = p.at
+	}
```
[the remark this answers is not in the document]
disq: how about positions starting at the same point? this start/end stuff leads to more complicated code for this use case.

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+	if strings.HasPrefix(filename, "file://") {
+		u, err := url.Parse(filename)
+		if err != nil {
+			return "", err
+		}
+		if u.Host != "" && u.Host != "localhost" {
+			return "", fmt.Errorf("invalid file URL %s", filename)
+		}
+		p := u.Path
+		if strings.HasPrefix(p, "/") && os.PathSeparator == '\\' {
```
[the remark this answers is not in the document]
disq: > Sure, here's an example of a file:// URL for both Windows and Linux systems:
> Windows: file:///C:/Users/username/Documents/file.txt
> Linux: file:///home/username/Documents/file.txt

So the windows file:// URL has an extra `/` before the `C:/`

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+		return "", fmt.Errorf("failed to prepare doc images: %w", err)
+	}
+	if resp.HTTPResponse.StatusCode > 299 {
+		return "", fmt.Errorf("failed preparing: %w", hub.ErrorFromHTTPResponse(resp.HTTPResponse, resp))
+	}
+
+	for _, item := range resp.JSON201.Items {
+		item := item
+
+		key := item.Checksum + ":" + item.Name
```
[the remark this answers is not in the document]
disq: Fixed with the struct in https://github.com/cloudquery/cloudquery/pull/16033/commits/73bf1a32c437e33ecc1e3764e6d9a2dc2bb0b135

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	// sort by start position, descending
+	sort.Slice(reps, func(i, j int) bool {
+		return reps[i].startPos > reps[j].startPos
+	})
+
+	// replace in reverse order
+	for _, ir := range reps {
+		literalTag := contents[ir.startPos:ir.endPos]
```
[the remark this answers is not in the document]
disq: Added in the struct definition in https://github.com/cloudquery/cloudquery/pull/16033/commits/73bf1a32c437e33ecc1e3764e6d9a2dc2bb0b135

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+	_, err = io.Copy(s, f)
+	if err != nil {
+		return "", err
+	}
+	result := s.Sum(nil)
+	return fmt.Sprintf("%x", result), nil
+}
+
+type imageFinder struct {
+	images map[string][]imageReference
```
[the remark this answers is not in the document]
disq: Done in https://github.com/cloudquery/cloudquery/pull/16033/commits/73bf1a32c437e33ecc1e3764e6d9a2dc2bb0b135

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+
+type imageReference struct {
+	ref     string // image filename inc. all paths (to replace with URL)
+	absFile string // absolute path to image file, to upload
+	url     string // result of upload
+
+	startPos, endPos int // start,end of complete markdown tag. if html: start,end of actual ref
+}
+
+var htmlImageRe = regexp.MustCompile(`<img\s+(?:[^>"']|"[^"]*"|'[^']*')*\s*src="([^"]*)"`)
```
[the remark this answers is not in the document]
disq: Fixed in https://github.com/cloudquery/cloudquery/pull/16033/commits/d26e88d9d2eb5bb39481d337ee80510f7df90a3e but since Go's regex doesn't have backreferences I can't make sure quote types match each other (I can do it in code but then it would be too late)

cli/internal/publish/images.go
```diff
@@ -0,0 +1,458 @@
...
+			Name:     name,
+			Checksum: checksum,
+		})
+	}
+
+	resp, err := c.CreateTeamImagesWithResponse(ctx, teamName, cloudquery_api.CreateTeamImagesJSONRequestBody{Images: reqs})
+	if err != nil {
+		return "", fmt.Errorf("failed to prepare doc images: %w", err)
+	}
+	if resp.HTTPResponse.StatusCode > 299 {
```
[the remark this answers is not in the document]
disq: It's defined as 201 in the API and we also rely on `resp.JSON201` two lines below this.

cli/internal/publish/images.go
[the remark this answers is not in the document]
disq: Removed in https://github.com/cloudquery/cloudquery/pull/16033/commits/015d93250a0e8aa498f7e712e32592b6e997725b in favour of x/net/html

cli/internal/publish/images.go
disq: @hermanschaaf Using x/net/html as of https://github.com/cloudquery/cloudquery/pull/16033/commits/015d93250a0e8aa498f7e712e32592b6e997725b

cli/internal/publish/images.go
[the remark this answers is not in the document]
disq: Also see @hermanschaaf 's comment in https://github.com/cloudquery/cloudquery/pull/16033#discussion_r1443123190

Issue #15391, opened 2023-11-20
AWS: use `sync.Map` in the `client.ServiceManager`


candiduslynx:

bbernays:
Do we really want to do this in this large release? Technically I think we can do the refactor in a minor version because nothing user facing would be impacted...

candiduslynx:
We thought about this as a major change as it impacts the access for the core structure in AWS plugin.
I'm OK with moving it out of scope, but a safer bet would be to mark this change breaking, too.

bbernays:
Lets leave this one till after all of the others that are 100% breaking... I am not completely sold on it requiring a major release

PR #16066, opened 2024-01-05
refactor: Use `sync.Map` in service manager

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/15391

Tested locally (scenario without specifying any regions)


bbernays (commented):
Do we expect there to be any performance benefits to this change?

For me this change makes the overall code harder to read...

PR #16068, opened 2024-01-05
refactor: Ease `client.Services` & `client.ServiceManager` logic

candiduslynx:
Overtakes https://github.com/cloudquery/cloudquery/pull/16066

### DIFF
--- a/plugins/source/aws/client/account.go
+++ b/plugins/source/aws/client/account.go
@@ -19,7 +19,7 @@ import (
 type svcsDetail struct {
 	partition string
 	accountId string
-	svcs      Services
+	svcs      *Services
 }
 
 func (c *Client) setupAWSAccount(ctx context.Context, logger zerolog.Logger, awsPluginSpec *spec.Spec, adminAccountSts AssumeRoleAPIClient, account spec.Account) (*svcsDetail, error) {
--- a/plugins/source/aws/client/client.go
+++ b/plugins/source/aws/client/client.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	wafv2types "github.com/aws/aws-sdk-go-v2/service/wafv2/types"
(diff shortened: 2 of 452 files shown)


hermanschaaf:
Thanks for the PR @candiduslynx! Could you explain the reasoning behind the change? Is there future work that depends on this? Or why should we do this now?

candiduslynx:
Hi @hermanschaaf!
This PR restructures the locking mechanism for the services in the aws plugin client: instead of decoupled services map & mutexes map members we mutexes incorporated into the service struct, providing easier init code & less error-prone mechanism (keep muted close to where the locked data is instead of having it layers above).

hermanschaaf:
> This PR restructures the locking mechanism for the services in the aws plugin client: instead of decoupled services map & mutexes map members we mutexes incorporated into the service struct, providing easier init code & less error-prone mechanism (keep muted close to where the locked data is instead of having it layers above).

@candiduslynx Got it, thanks.

While this looks like an improvement, if we don't have an active need for this refactor then I think it's better to leave the code as it is right now, as we know it was working. Unfortunately this section of the code is quite risky and needs to be carefully reviewed and tested to make sure we haven't inadvertently introduced deadlocks or other subtle issues. Unless there's another change waiting to be built on this I think this can wait?

candiduslynx:
> @hermanschaaf 
> 
> While this looks like an improvement, if we don't have an active need for this refactor then I think it's better to leave the code as it is right now, as we know it was working. Unfortunately this section of the code is quite risky and needs to be carefully reviewed and tested to make sure we haven't inadvertently introduced deadlocks or other subtle issues. Unless there's another change waiting to be built on this I think this can wait?

Although it can totally wait, the original issue (https://github.com/cloudquery/cloudquery/issues/15391) was designed to be fixed in the v23 breaking batch. I see that the code may be hard to review because of the seeming amount of changes, but the whole change is constrained in the `plugins/aws/client` package (the other code is just changing from a value to a pointer in the testing framework).

I did a thorough testing locally with different init strategies, so I'm pretty sure about the change, but once again, it can 💯 wait.

yevgenypats (changes requested):
Im also not a big fan of refactors for sake of refactors (as it creates work for reviewers and then testing major versions with performance and other stuff) at least at this stage. maybe better to take a stab at the snyk plugin.

erezrokah:
Should we close this or stale?

candiduslynx:
> Should we close this or stale?

I still think it has value & it's not stale.

PR #16122, opened 2024-01-09
refactor: Apply perfsprint linter

catenacyber:
#### Summary

Optimization: applies perfsprint linter for better performance

happy new year @yevgenypats ;-)


erezrokah (commented):
Thanks for the PR @catenacyber, is this something we can add to our linter configuration to avoid having to run the linter manually?

The CLI config is here:
https://github.com/cloudquery/cloudquery/blob/e61860801b2745a82ed43ddfe02c30cc66d40c8c/cli/.golangci.yml

catenacyber:
> Thanks for the PR @catenacyber, is this something we can add to our linter configuration to avoid having to run the linter manually?

It should work indeed, see last commit

erezrokah (approved):
Thanks for the PR @catenacyber, sorry for the delay in merging

PR #16159, opened 2024-01-16
fix: Handle `docker.cloudquery.io` docker registry when checking if auth is needed

erezrokah:
#### Summary

Updates the CLI logic a bit to try and get auth token if someone references our own docker registry. We use this logic to recommend users to `cloudquery login` if they reference our registry


cli/internal/auth/token.go
```diff
@@ -8,9 +8,13 @@ import (
 	"github.com/rs/zerolog"
 )
 
+func tokenNeeded(registry specs.Registry, path string) bool {
+	return registry == specs.RegistryCloudQuery || registry == specs.RegistryDocker && strings.HasPrefix(path, "docker.cloudquery.io")
```
disq, line 12:
```suggestion
	return registry == specs.RegistryCloudQuery || (registry == specs.RegistryDocker && strings.HasPrefix(path, "docker.cloudquery.io"))
```

erezrokah: 👍

disq (approved):

PR #16165, opened 2024-01-16
feat: Persist KMS Key data even if AccessDenied on `kms:DescribeKey`

bbernays:
#### Summary

### DIFF
--- a/plugins/source/aws/resources/services/kms/keys.go
+++ b/plugins/source/aws/resources/services/kms/keys.go
@@ -81,6 +81,15 @@ func getKey(ctx context.Context, meta schema.ClientMeta, resource *schema.Resour
 		options.Region = cl.Region
 	})
 	if err != nil {
+		if client.IsAWSError(err, "AccessDenied") || client.IsAWSError(err, "AccessDeniedException") {
+			resource.Item = &types.KeyMetadata{
+				Arn:   item.KeyArn,
+				KeyId: item.KeyId,
+			}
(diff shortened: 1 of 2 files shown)


erezrokah (approved):

hermanschaaf:
Let's please update the description before merging 😅

plugins/source/aws/resources/services/kms/keys.go
```diff
@@ -81,6 +81,15 @@ func getKey(ctx context.Context, meta schema.ClientMeta, resource *schema.Resour
 		options.Region = cl.Region
 	})
 	if err != nil {
+		if client.IsAWSError(err, "AccessDenied") || client.IsAWSError(err, "AccessDeniedException") {
+			resource.Item = &types.KeyMetadata{
+				Arn:   item.KeyArn,
+				KeyId: item.KeyId,
+			}
+			cl.Logger().Warn().Err(err).Msg("metadata retrieved from ListKeys will still be persisted")
```
hermanschaaf, line 89: I wonder if we really need a warning for every instance; I think one warning, or even just a mention in the documentation of this table, should suffice

bbernays: Users could have access to some keys but not others... and this follows the pattern we use in Lambda

hermanschaaf: I'm just worried it's going to be a bit noisy if users have many keys. But it's okay, we can start with this and see if it's an issue for anyone

hermanschaaf (approved):

Issue #16199, opened 2024-01-19
feat: Add all mfa_devices to AWS source plugin


jeromewir:
### Describe the resource

From [AWS](https://aws.amazon.com/what-is/mfa/):
Multi-factor authentication (MFA) is a multi-step account login process that requires users to enter more information than just a password. For example, along with the password, users might be asked to enter a code sent to their email, answer a secret question, or scan a fingerprint.

### Use Case

Being able to list both physical and virtual MFA devices for AWS IAM users.

### Link to API documentation

https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListMFADevices.html

### Additional Context

At the moment, cloudquery allows to list virtual MFA devices but not physical ones.
All virtual MFA devices can be retrieved at once with the [ListVirtualMFADevices](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListVirtualMFADevices.html) but the physical ones are retrieved per user: [ListMFADevice](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListMFADevices.html)

### Pull request (optional)

- [X] I can submit a pull request

PR #16200, opened 2024-01-19
feat: Add mfa_devices for AWS IAM users

jeromewir:
#### Summary

Resolves https://github.com/cloudquery/cloudquery/issues/16199

Add all `mfa_devices` listing for AWS IAM users.
Those ressources need to be fetch per user as the physical MFA devices cannot be retrieved via a list endpoint.


bbernays:
Hi @jeromewir - Thank you for this PR, but please note that we have moved the AWS, Azure and GCP plugins to an Open-Core model. This means that we are limiting the PRs we accept for those plugins. In this case we will accept and merge this PR, but  in the future I would suggest waiting to hear back from a member of the CloudQuery team about whether we would accept a PR for an open issue. 

Blog post with more information about this move to `Open-Core` can be found on our blog: https://www.cloudquery.io/blog/announcing-open-core-plugins

x

PR #16203, opened 2024-01-19
feat: Ensure all insert queries can be cached

bbernays:
#### Summary

Currently we only preallocate for the first 100 sql queries. This PR changes it so that the cache can hold the SQL statement for all tables by using the number of tables present in the DB as the upper limit


hermanschaaf:
> This PR changes it so that the cache can hold the SQL statement for all tables by using the number of tables present in the DB as the upper limit

It seems like the cache was already holding all the SQL statements, it was just potentially pre-allocating more/less memory.

hermanschaaf (approved):

PR #16220, opened 2024-01-22
fix: Add `N/A` as `stack_id` when the value is unavailable

candiduslynx:
Closes https://github.com/cloudquery/cloudquery-issues/issues/1174 (Internal issue)


plugins/source/aws/resources/services/cloudformation/stack_instances.go
```diff
@@ -62,9 +63,13 @@ func fetchStackInstanceSummary(ctx context.Context, meta schema.ClientMeta, pare
 		if err != nil {
 			return err
 		}
-		for i := range page.Summaries {
+		for _, summary := range page.Summaries {
+			if summary.StackId == nil {
```
hermanschaaf, line 67: It seems like we could have some duplicate PKs if StackId is null? Can it happen that a single stack set has multiple stacks in an error state? I don't see a better way of handling this but maybe it's something to document

candiduslynx: IDK if there can be multiple stack instances in a bad state (inoperable?) inside a single stack set.
I can add a more comprehensive comment to this, but IDK if it's a good idea to add this to a doc somewhere.

hermanschaaf: Why is it not a good idea to mention this in the table description? If it can happen, it's a limitation users should be aware of without needing to read the code

candiduslynx: cf41bfebb1156798d9d923c7654c0145bcbf0aeb

hermanschaaf (approved):

bbernays:
I think this has issues, A stackset can have multiple stacks, and if multiple stacks have deployment issues this will result in loss of data

candiduslynx:
> I think this has issues, A stackset can have multiple stacks, and if multiple stacks have deployment issues this will result in loss of data

Yeah, but it's not worse than the current state of things (it implies all of them having `nil` value in `stack_id`)

bbernays:
Are we just pushing the issue down the road? If a `stackId` is not available are there any other fields that are not available, because the related table would then fail...

PR #16226, opened 2024-01-22
feat: Add Auto Migrate Functionality for migrating to using a `_cq_id` as the only Primary Key

bbernays:
#### Summary

This functionality will allow users to auto update their primary key from a composite key to `_cq_id` without dropping any data.

This PR is blocked by the following outstanding issues:
- [x] ~Need to add a test case in the plugin sdk to enable this type of migration:~ completed in https://github.com/cloudquery/plugin-sdk/pull/1480
- [x] Need `v4.26.0` of the plugin-sdk [PR here](https://github.com/cloudquery/plugin-sdk/pull/1468) to be released


plugins/destination/postgresql/client/migrate.go
```diff
@@ -167,6 +170,50 @@ func (c *Client) dropTable(ctx context.Context, tableName string) error {
...
+func (c *Client) migrateToCQID(ctx context.Context, tableName string, _ schema.Column) error {
+	// Steps:
+	// acquire connection
+	conn, err := c.conn.Acquire(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to acquire connection: %w", err)
+	}
+	defer conn.Release()
+	// start transaction
+	tx, err := conn.BeginTx(ctx, pgx.TxOptions{})
```
candiduslynx, line 182: maybe use serializable explicitly rather than relying on the default level being enough?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -167,6 +170,50 @@ func (c *Client) dropTable(ctx context.Context, tableName string) error {
...
+	_, err = tx.Exec(ctx, "ALTER TABLE "+sanitizedTableName+" ADD CONSTRAINT "+sanitizedPKName+" Primary Key ("+pgx.Identifier{schema.CqIDColumn.Name}.Sanitize()+")")
+	if err != nil {
+		c.logger.Error().Err(err).Str("table", tableName).Msg("Failed to create new primary key on _cq_id")
+		if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
+			return fmt.Errorf("failed to rollback transaction: %w", rollbackErr)
+		}
+		return err
+	}
+	// Commit transaction if no error
+	err = tx.Commit(ctx)
```
candiduslynx, line 210: this can be rolled into a commit/rollback pattern like this:
```go
func A() (err error) {
...
	tx, err := conn.BeginTx(ctx, pgx.TxOptions{})
	if err != nil {
		return err
	}
	defer func() {
		if err == nil {
			if err = tx.Commit(ctx); err != nil {
				err = fmt.Errorf("failed to commit transaction: %w", err)
			}
		}

		// this will also work for tx commit errors
		if err != nil {
			if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
				// we want to preserve the original err still, so only log
				c.logger.Err(rollbackErr).Msg("failed to rollback transaction")
			}
		}
	}()
	...
}
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -167,6 +170,50 @@ func (c *Client) dropTable(ctx context.Context, tableName string) error {
...
+	}
+	defer conn.Release()
+	// start transaction
+	tx, err := conn.BeginTx(ctx, pgx.TxOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+
+	sanitizedTableName := pgx.Identifier{tableName}.Sanitize()
+	sanitizedPKName := pgx.Identifier{getPKName(&schema.Table{Name: tableName})}.Sanitize()
```
erezrokah, line 204: Can we sure this constraint name always exist in this scenario? Maybe we should ignore if it doesn't instead of failing?

It won't exist probably when we do Postgres->Postgres, but then `TableColumnChangeTypeMoveToCQOnly` should be false I think

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -108,6 +116,7 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
 			Name:       columnName,
 			PrimaryKey: isPrimaryKey,
 			NotNull:    notNull,
+			Unique:     isUnique,
```
erezrokah, line 119: I'm not sure if we should also set `Unique: true` when the column is a PK. I'm not sure it matters, but I think that if we set a column to both `PrimaryKey: true` and `Unique: true`, then create the relevant table, then try to migrate it, we will always load it as `PrimaryKey: true` and `Unique: false` so we'll detect a schema change

erezrokah (commented):
Added a couple of comments

plugins/destination/postgresql/client/migrate.go
bbernays: This function is only called for `TableColumnChangeTypeMoveToCQOnly` change... So I think that it is fine, If users are messing around with the `_cq_id` column we cannot be sure of anything... 
```
case schema.TableColumnChangeTypeMoveToCQOnly:
			return c.migrateToCQID(ctx, tableName, change.Current)
```

plugins/destination/postgresql/client/list_tables.go
erezrokah: Might worth checking other destinations too since we do have that change type now

bbernays: I am removing all of the changes for uniqueness and can handle this in a different PR

erezrokah (approved):

plugins/destination/postgresql/client/migrate.go
```diff
@@ -167,6 +170,56 @@ func (c *Client) dropTable(ctx context.Context, tableName string) error {
...
+
+	// Drop existing primary key
+	_, err = tx.Exec(ctx, "alter table "+sanitizedTableName+" drop constraint "+sanitizedPKName)
+	if err != nil {
+		c.logger.Error().Err(err).Str("table", tableName).Msg("Failed to drop primary key")
+		return err
+	}
+
+	// Create new Primary Key with CQID
+	_, err = tx.Exec(ctx, "ALTER TABLE "+sanitizedTableName+" ADD CONSTRAINT "+sanitizedPKName+" Primary Key ("+pgx.Identifier{schema.CqIDColumn.Name}.Sanitize()+")")
```
candiduslynx, line 215: a nit: use the same casing across the queries added (L 208 uses lower-case, this one is mixed)

candiduslynx (approved):
only a tiny nit

candiduslynx (approved):

PR #16229, opened 2024-01-22
feat: Report progress for sync runs

erezrokah:
#### Summary

Goes with https://github.com/cloudquery/cloudquery-api-go/pull/109 and https://github.com/cloudquery/plugin-sdk/pull/1473 to report progress for managed syncs

PR #16230, opened 2024-01-22
refactor: Extract `NewClientWithResponses` to api package

erezrokah:
#### Summary

Extracted from https://github.com/cloudquery/cloudquery/pull/16229 to simplify the diff.

Moves some duplicated code into a helper package


cli/cmd/addon_download.go
```diff
@@ -73,13 +72,9 @@ func runAddonDownload(ctx context.Context, cmd *cobra.Command, args []string) er
 		return fmt.Errorf("invalid addon ref %q: version must start with 'v'", args[0])
 	}
 
-	c, err := cloudquery_api.NewClientWithResponses(managedplugin.APIBaseURL(),
-		cloudquery_api.WithRequestEditorFn(func(ctx context.Context, req *http.Request) error {
-			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
-			return nil
-		}))
+	c, err := api.NewClient(getEnvOrDefault(envAPIURL, defaultAPIURL), token.Value)
```
disq, line 75: why not refactor api url envs as well?

erezrokah: Ah, just remembered, because of https://github.com/cloudquery/cloudquery/blob/085548ac922f4c0adee8a024e0f09ea4d9e99ab1/cli/internal/team/team.go#L16 that passes only a URL, so needs another refactoring

disq (approved):

cli/cmd/addon_download.go
erezrokah: I'll do it in a separate PR to reduce the diff with this one

Issue #16218, opened 2024-01-20
bug: arm64 docker images are actually built for amd64


JumaX:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Env: OSX Sonoma 14.1
Docker Desktop: 4.26.1
Hardware: M1 Max, 32GB of RAM

Building a local docker environment using the following images fails:

`ghcr.io/cloudquery/cloudquery:latest-linux-arm64`

> Error response from daemon: image with reference ghcr.io/cloudquery/cloudquery:latest-linux-arm64 was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64

`ghcr.io/cloudquery/cloudquery:4.4-linux-arm64`

> Error response from daemon: image with reference ghcr.io/cloudquery/cloudquery:4.4-linux-arm64 was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64

Docker inspect output on both confirms they are built for `amd64`:

`
        "Architecture": "amd64",
        "Os": "linux",
        "Size": 47424823,
        "VirtualSize": 47424823,
`

Using DockerDesktop you can also view the image hierarchy and can see that these images are based on a root image for Alpine 3.19 for X86.

Links to here: 
https://github.com/alpinelinux/docker-alpine/blob/97c57449282d97cfa1c0b64669aed9afbf08645a/x86_64/Dockerfile 

<img width="1218" alt="Screenshot 2024-01-20 at 12 29 01" src="https://github.com/cloudquery/cloudquery/assets/1376580/fdf2238c-71ef-4d3f-a194-b53125f2de40">

### Expected Behavior

The expected behaviour is that images tagged with `linux/arm64` would be built for the arm64 platform and not `amd64`.

### CloudQuery (redacted) config

Not Applicable.

### Steps To Reproduce

Env: OSX Sonoma 14.1
Docker Desktop: 4.26.1
Hardware: M1 Max, 32GB of RAM

1. Pull either image:

`docker pull ghcr.io/cloudquery/cloudquery:latest-linux-arm64`

`docker pull ghcr.io/cloudquery/cloudquery:4.4-linux-arm64`

2. Use a simple docker compose file like the following:

```
version: "3.8"
services:
  cloudquery_demo:
    image: ghcr.io/cloudquery/cloudquery:latest-linux-arm64
    platform: linux/arm64
    container_name: cloudquery
    pull_policy: never
    volumes:
      - ./config.yml:/cache/.cq
[...]

PR #16231, opened 2024-01-22
fix: Build arm64 docker images with arm64 platform

disq:
Could fix https://github.com/cloudquery/cloudquery/issues/16218

https://goreleaser.com/cookbooks/multi-platform-docker-images/#creating-multi-platform-docker-images-with-goreleaser also mentions qemu, but I'm not sure it's needed after checking https://github.com/docker/setup-qemu-action/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc


erezrokah (commented):
I think you'll need to add:

```yaml
-
  name: Set up Docker Buildx
  uses: docker/setup-buildx-action@v3
```

Before https://github.com/cloudquery/cloudquery/blob/9379d5348bdc3abae8bad2f426eb7a513e7b2a83/.github/workflows/release_cli.yml#L43

cli/.goreleaser.yaml
```diff
@@ -32,7 +35,10 @@ dockers:
       - "ghcr.io/cloudquery/cloudquery:latest-linux-arm64"
       - "ghcr.io/cloudquery/cloudquery:{{.Version}}-linux-arm64"
       - "ghcr.io/cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}-linux-arm64"
+    use: buildx
     build_flag_templates:
+      - "--pull"
```
hermanschaaf, line 40: Does buildx not account for the platform when checking if the image is already in the cache, or what is the reason for adding `--pull` ?

erezrokah (commented):
We might want to follow up with a manifest configuration, as currently we tag `amd64` as:

```
- "ghcr.io/cloudquery/cloudquery:latest"
- "ghcr.io/cloudquery/cloudquery:{{.Version}}"
- "ghcr.io/cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}"
```

erezrokah:
Also probably you could test this via `goreleaser release --snapshot --clean -f ./cli/.goreleaser.yaml` (you need the pro version from `brew install goreleaser/tap/goreleaser-pro` (API key in 1Password)

cli/.goreleaser.yaml
disq: I've added it as per Erez's [instruction](https://github.com/cloudquery/cloudquery/issues/16218#issuecomment-1903804664) feel free to take over this issue if you're comfortable with this as I'm really not.

hermanschaaf: Oh I see, it's from the goreleaser example. I guess it doesn't hurt, we just won't use the cache. I also don't know whether that's necessary 😅

erezrokah: I think the reason to add `--pull` is to refresh any base images our image depends on. For example if we depend on `v2` tag of an image we'll pull the latest changes to that tag.

hermanschaaf: Ah I see, makes sense yeah.

disq:
@erezrokah added (not sure correctly?) in [`f2e31af` (#16231)](https://github.com/cloudquery/cloudquery/pull/16231/commits/f2e31af9deb5d3e9d9fd5a5fa7b2327e4eaece75) PTAL

goreleaser seems to work (once I switched to docker default context)

cli/.goreleaser.yaml
```diff
@@ -32,9 +35,28 @@ dockers:
...
+
+- name_template: "cloudquery/cloudquery:latest"
+  image_templates:
+  - "cloudquery/cloudquery:latest-linux-amd64"
+  - "cloudquery/cloudquery:latest-linux-arm64"
+
+- name_template: "cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}"
+  image_templates:
+  - "cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}-linux-amd64"
+  - "cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}-linux-arm64"
```
erezrokah, line 58:
```suggestion
- name_template: "cloudquery/cloudquery:{{.Version}}"
  image_templates:
  - "ghcr.io/cloudquery/cloudquery:{{.Version}}-linux-amd64"
  - "ghcr.io/cloudquery/cloudquery:{{.Version}}-linux-arm64"

- name_template: "ghcr.io/cloudquery/cloudquery:latest"
  image_templates:
  - "ghcr.io/cloudquery/cloudquery:latest-linux-amd64"
  - "ghcr.io/cloudquery/cloudquery:latest-linux-arm64"

- name_template: "ghcr.io/cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}"
  image_templates:
  - "ghcr.io/cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}-linux-amd64"
  - "ghcr.io/cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}-linux-arm64"
```

And I think you need to remove:

```
      - "ghcr.io/cloudquery/cloudquery:latest"
      - "ghcr.io/cloudquery/cloudquery:{{.Version}}"
      - "ghcr.io/cloudquery/cloudquery:{{ .Major }}.{{ .Minor }}"
```

from the amd64 configuration.

disq:
@erezrokah Makes much more sense now thanks

erezrokah (approved):

erezrokah (approved):

PR #16235, opened 2024-01-23
refactor: Remove duplication for getting API env variable value

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/cloudquery/pull/16230 with some more cleanup


disq (approved):
constants in `constants.go` can be removed also (except for `defaultAccountsURL`... maybe refactor those into `api` too?)

Issue #15272, opened 2023-11-15
feat: Update scaffold release workflow to publish to the Hub instead of GitHub


erezrokah:
### Which problem is this feature request solving?

The workflow in https://github.com/cloudquery/cloudquery/blob/c13aefce4af8e65a37014ad5614e9c764beb512d/scaffold/cmd/templates/source/.github/workflows/release.yaml.tpl uses Go Releaser to publish to GitHub.

### Describe the solution you'd like

We should update the workflow to publish to the Hub.
This won't work out of the box since users would need to configure a GitHub secret with an API key to publish successfully 

### Pull request (optional)

- [X] I can submit a pull request

PR #16302, opened 2024-01-24
feat!: Update scaffold to push release to CloudQuery Hub

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/15272


scaffold/cmd/templates/source/.github/workflows/release.yaml.tpl
```diff
@@ -16,23 +16,72 @@ jobs:
         uses: booxmedialtd/ws-action-parse-semver@7784200024d6b3fc01253e617ec0168daf603de3
         with:
           input_string: ${{"{{"}}github.ref_name{{"}}"}}
+          # `version_extractor_regex` is required to parse 'v1.0.0'.
```
erezrokah, line 19: Are you sure we need this?
I think it supports `v1.0.0` out of the box, see example input here:
https://github.com/cloudquery/cloudquery/actions/runs/7628688141/job/20780338631#step:3:3

Please note we use `ref_name` which is the clean tag (doesn't have the `refs/tags/` prefix) like the example in https://github.com/booxmedialtd/ws-action-parse-semver/tree/7784200024d6b3fc01253e617ec0168daf603de3#with-an-extractor-regex

The action seems to use https://github.com/booxmedialtd/ws-action-parse-semver/blob/7784200024d6b3fc01253e617ec0168daf603de3/src/parser.js#L1 which uses this regex https://github.com/npm/node-semver/blob/816c7b2cbfcb1986958a290f941eddfd0441139e/internal/re.js#L115

Also the tag will always have `v` prefix since that's the trigger https://github.com/cloudquery/cloudquery/pull/16302/files#diff-67f6cf97f98a6d73520a36f8d9c461783f9d2b24259009eb0c34a9991e4e8a4dR5

erezrokah (changes requested):
Looks great overall. I'm not sure we need `version_extractor_regex`

scaffold/cmd/templates/source/.github/workflows/release.yaml.tpl
candiduslynx: 556b92b

Issue #16298, opened 2024-01-23
bug: cloudquery sync errors when user is logged in with multiple cloudquery accounts


k-rheinheimer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

AWS fails to be successfully configured if user is concurrently logged in to cloudquery cli with 2 accounts

### Expected Behavior

the cli should default to using the most recent authentication

### CloudQuery (redacted) config

aws.yml
``` yaml
kind: source
spec:
  # Source spec section
  name: aws
  path: cloudquery/aws
  registry: cloudquery
  version: "v23.6.0"
  tables: [
           "aws_ec2_network_acls",
           "aws_cloudwatch_alarms",
           "aws_apigateway_rest_api_stages",
           "aws_ec2_security_groups",
           "aws_cloudtrail_trail_event_selectors",
           "aws_cloudwatchlogs_metric_filters",
           "aws_sns_subscriptions"
          ]
  destinations: ["postgresql"]
  spec:
    # Optional parameters
    # regions: []
    # accounts: []
    # org: nil
    # concurrency: 50000
    # initialization_concurrency: 4
    # aws_debug: false
    # max_retries: 10
    # max_backoff: 30
    # custom_endpoint_url: ""
    # custom_endpoint_hostname_immutable: nil # required when custom_endpoint_url is set
    # custom_endpoint_partition_id: "" # required when custom_endpoint_url is set
    # custom_endpoint_signing_region: "" # required when custom_endpoint_url is set
    # use_paid_apis: false
    # table_options: nil
    # scheduler: dfs # options are: dfs, round-robin or shuffle
    # use_nested_table_rate_limiting: false 
    # enable_api_level_tracing: false 
    ```
    
    postgresql.yml
    ``` yaml
    kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  registry: "cloudquery"
  version: "v7.2.0"

  spec:
    connection_string: "***********"
    # can be also specified in DSN format which can hold special characters in password
    # connection_string: "user=postgres password=pass+0-[word host=localhost port=5432 dbname=postgres"
    # Optional parameters:
    # pgx_log_level: error
    # batch_size: 10000 # 10K entries
    # batch_size_bytes: 100000000 # 100 MB
    # batch_timeout: 60s

    

### Steps To Reproduce

1. run cloudquery login as first user (not correctly configured for AWS)
2. run cloudquery login as second user (correctly configured)
[...]

PR #16310, opened 2024-01-24
fix: Properly set team on login

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/16298

The following scenarios have to be taken care of:
1. The user explicitly selects team via `--team` flag
2. The user has previously selected team
   1. The team is available after login – use it
   2. The team is unavailable – use the opt 3
3. Check that 1 & only 1 team is available for the user – & use it

### DIFF
--- a/cli/cmd/login.go
+++ b/cli/cmd/login.go
@@ -250,5 +250,6 @@ func setTeamOnLogin(ctx context.Context, cmd *cobra.Command, token string) error
 	cmd.Printf("Your current team is not set.\n\n")
 	cmd.Printf("Teams available to you: " + strings.Join(teams, ", ") + "\n\n")
 	cmd.Printf("To set your current team, run `cloudquery switch <team>`\n\n")
-	return fmt.Errorf("team not set. Available teams: [" + strings.Join(teams, ", ") + "]")
+	// we don't fail here immediately, as there are some additional commands the user can run in this state
+	return nil
 }
--- a/cli/cmd/login.go
+++ b/cli/cmd/login.go
@@ -211,7 +211,7 @@ func setTeamOnLogin(ctx context.Context, cmd *cobra.Command, token string) error
 	}
 
 	if cmd.Flags().Changed("team") {
-		// don't care about the current value
+		// don't care about the current cached config value as the user explicitly passes the `team` flag
 		currentTeam := cmd.Flag("team").Value.String()
 		err = cl.ValidateTeamAgainstTeams(currentTeam, teams)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


cli/cmd/login.go
```diff
@@ -177,60 +177,78 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
...
+	// list all available teams
+	teams, err := cl.ListAllTeams(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to list teams: %w", err)
 	}
+
 	if cmd.Flags().Changed("team") {
-		selectedTeam := cmd.Flag("team").Value.String()
-		err = cl.ValidateTeam(ctx, selectedTeam)
+		// don't care about the current value
```
erezrokah, line 214:
```suggestion
		// don't care about the current cached config value as the user explicitly passes the `team` flag
```

cli/cmd/login.go
```diff
@@ -177,60 +177,78 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
...
+		cmd.Printf("Your current team is set to %s.\n", currentTeam)
+		return nil
+	}
 
-	return nil
+	// 0 or > 1 teams available
+	cmd.Printf("Your current team is not set.\n\n")
+	cmd.Printf("Teams available to you: " + strings.Join(teams, ", ") + "\n\n")
+	cmd.Printf("To set your current team, run `cloudquery switch <team>`\n\n")
+	return fmt.Errorf("team not set. Available teams: [" + strings.Join(teams, ", ") + "]")
```
erezrokah, line 253: Did we fail before if teams > 1? Not sure we should fail the login command. The user will get another error message if they use a command that requires a team to be set

cli/cmd/login.go
```diff
@@ -177,60 +177,78 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
...
+			return fmt.Errorf("failed to set team: %w", err)
+		}
+		cmd.Printf("Your current team is set to %s.\n", currentTeam)
+		return nil
+	}
 
-	return nil
+	// 0 or > 1 teams available
+	cmd.Printf("Your current team is not set.\n\n")
+	cmd.Printf("Teams available to you: " + strings.Join(teams, ", ") + "\n\n")
```
erezrokah, line 251: If the user doesn't have any teams we'll print an empty array. Maybe we should direct the user to create a team on the cloud Website instead?

erezrokah (changes requested):
Great change, added a few comments

cli/cmd/login.go
candiduslynx: I think that would be a separate change unrelated to the fixed bug. I can do that as a follow-up

cli/cmd/login.go
candiduslynx: I removed error in ea441bb43e09d71ade958d3711e3ea768e9b0d30

erezrokah (approved):

PR #16316, opened 2024-01-25
fix!: Disable static linking for Snowflake destination plugin

candiduslynx:
Closes https://github.com/cloudquery/cloudquery-issues/issues/966

### DIFF
--- a/plugins/destination/snowflake/go.mod
+++ b/plugins/destination/snowflake/go.mod
@@ -148,3 +148,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.27.2-0.20240128165327-6183308309c0
--- a/plugins/destination/snowflake/go.mod
+++ b/plugins/destination/snowflake/go.mod
@@ -149,4 +149,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.27.2-0.20240128165327-6183308309c0
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.27.2-0.20240128170311-b603a6cfe7bf
--- a/.github/workflows/publish_plugin_to_hub_snowflake.yml
+++ b/.github/workflows/publish_plugin_to_hub_snowflake.yml
@@ -39,7 +39,7 @@ jobs:
     container:
       image: ghcr.io/cloudquery/golang-cross:v10.0.0
       env:
-        CGO_ENABLED: 1
+        CGO_ENABLED: 0 # aha!
         CC: /usr/bin/gencc.sh
         CXX: /usr/bin/gencpp.sh
         GOAMD64: v1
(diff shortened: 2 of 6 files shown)


erezrokah:
I believe there's a reason @maaarcelino added this, so I would let him review

candiduslynx:
> I believe there's a reason @maaarcelino added this, so I would let him review

I think it was caused by the goreleaser file, but, as mentioned in the issue, it seems that the overrides were only applied to the windows build & no to the linux one, so it might not have a reason.

maaarcelino:
In the [goreleaser file for Snowflake](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/snowflake/.goreleaser.yaml#L51) we have static linking for the Linux target. The reason this was discovered was due to [this issue for the SQLite plugin](https://github.com/cloudquery/cloudquery/issues/14465) for when the plugin was ran under Docker using the Alpine image. That image doesn't include the necessary C libraries for the plugin to run and static linking solved the issue.

candiduslynx:
> In the [goreleaser file for Snowflake](https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/snowflake/.goreleaser.yaml#L51) we have static linking for the Linux target. The reason this was discovered was due to [this issue for the SQLite plugin](https://github.com/cloudquery/cloudquery/issues/14465) for when the plugin was ran under Docker using the Alpine image. That image doesn't include the necessary C libraries for the plugin to run and static linking solved the issue.

@maaarcelino 

As I mentioned in the issue, GoReleaser seems to apply overrides to only the first entry (windows), so the linux one isn't actually built with the proper ldflags. That explains why the build from the GitHub release works, while the CQ Hub one fails.

maaarcelino:
Can you test this change under Docker/Alpine Linux?

erezrokah:
How does this change impact the CloudQuery registry? We don't use GoReleaser for it.

- GoReleaser -> GitHub releases
- Our [own build logic](https://github.com/cloudquery/plugin-sdk/blob/e399d1538f6b6044f302ab556cd2c4f8cb91343b/serve/package.go#L110) -> Hub releases

candiduslynx:
It's in the discovery process, but ATM I can say that the builds we have on GH aren't 100% statically linked, while the ones produced by package command still require some libraries to be present on the system.

erezrokah:
> It's in the discovery process, but ATM I can say that the builds we have on GH aren't 100% statically linked, while the ones produced by package command still require some libraries to be present on the system.

The builds from GoReleaser work inside GitHub Actions, and the ones we generate via the `package` command don't as described in the linked issue. Shouldn't the fix be in the `package` command logic?

candiduslynx:
> > It's in the discovery process, but ATM I can say that the builds we have on GH aren't 100% statically linked, while the ones produced by package command still require some libraries to be present on the system.
> 
> The builds from GoReleaser work inside GitHub Actions, and the ones we generate via the `package` command don't as described in the linked issue. Shouldn't the fix be in the `package` command logic?

The issue with the `GoReleaser` is that the override didn't actually work, so we didn't add extra flags to the linux build.

I'm trying to find a working set of flags that will allow the plugin to be statically built in the GoReleaser now, &, after that, I'll port the fix to other plugins (sqlite, duckdb) + to the SDK.

plugins/destination/snowflake/go.mod
```diff
@@ -148,3 +148,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.27.2-0.20240128170311-b603a6cfe7bf
```
disq, line 152: Link: https://github.com/cloudquery/plugin-sdk/commit/b603a6cfe7bf

candiduslynx: I'll mention everything once I'm done with repro/testing

candiduslynx:
Will reopen a proper PR

PR #16329, opened 2024-01-29
fix: Don't use `CGO` unless specifically selected to

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/966
Requires https://github.com/cloudquery/plugin-sdk/pull/1492


candiduslynx:
I've added tags to test releasing:
* `plugins-destination-duckdb-v111.1.1-rc111`
* `plugins-destination-snowflake-v111.1.1-rc111`
* `plugins-destination-sqlite-v111.1.1-rc111`

PR #16358, opened 2024-01-30
feat: Add JSON schema support to Datadog source plugin

mnorbury:
fixes: https://github.com/cloudquery/cloudquery/issues/16336

### DIFF
--- a/plugins/source/datadog/client/spec.go
+++ b/plugins/source/datadog/client/spec.go
@@ -1,5 +1,7 @@
 package client
 
+import _ "embed"
+
 type Spec struct {
 	Accounts    []Account `json:"accounts"`
 	Concurrency int       `json:"concurrency"`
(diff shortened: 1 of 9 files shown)


plugins/source/datadog/client/spec.go
```diff
@@ -1,5 +1,7 @@
 package client
 
+import _ "embed"
+
 type Spec struct {
```
hermanschaaf, line 5: It would be great to add the descriptions from https://hub.cloudquery.io/plugins/source/cloudquery/datadog/v4.2.8/docs?search=datadog#configuration-datadog-spec as comments here, so it will be propagated to the JSONSchema as well

hermanschaaf (approved):
Nit about the descriptions, otherwise LGTM 👍

PR #16369, opened 2024-01-30
feat: Add JSON Schema to Test destination plugin

disq:
Implements https://github.com/cloudquery/cloudquery/issues/16352

### DIFF
--- a/plugins/destination/test/client/schema.json
+++ b/plugins/destination/test/client/schema.json
@@ -7,7 +7,8 @@
       "properties": {
         "error_on_write": {
           "type": "boolean",
-          "description": "If true, will return an error on write rather than consume from the channel"
+          "description": "If true, will return an error on write rather than consume from the channel",
+          "default": false
         }
       },
       "additionalProperties": false,
--- a/plugins/destination/test/client/schema.json
+++ b/plugins/destination/test/client/schema.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/cloudquery/cloudquery/plugins/destination/test/client/spec",
+  "$ref": "#/$defs/Spec",
+  "$defs": {
+    "Spec": {
+      "properties": {
+        "error_on_write": {
+          "type": "boolean",
+          "description": "If true, will return an error on write rather than consume from the channel"
+        }
+      },
+      "additionalProperties": false,
+      "type": "object"
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


plugins/destination/test/client/schema.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/cloudquery/cloudquery/plugins/destination/test/client/spec",
+  "$ref": "#/$defs/Spec",
+  "$defs": {
+    "Spec": {
+      "properties": {
+        "error_on_write": {
+          "type": "boolean",
+          "description": "If true, will return an error on write rather than consume from the channel"
```
candiduslynx, line 10: you can add default here via annotating the struct:
```go
type Spec struct {
	// If true, will return an error on write rather than consume from the channel
	ErrorOnWrite bool `json:"error_on_write,omitempty" jsonschema:"default=false"`
}
```

plugins/destination/test/client/spec_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+func TestJSONSchema(t *testing.T) {
+	jsonschema.TestJSONSchema(t, JSONSchema, []jsonschema.TestCase{
+		{
+			Name: "empty spec",
+			Spec: `{}`,
+		},
+		{
+			Name: "spec with err_on_write",
+			Spec: `{"error_on_write": true}`,
```
candiduslynx, line 17: add test for `{"error_on_write":null}` & `{"error_on_write":123}` as well

plugins/destination/test/go.mod
```diff
@@ -4,10 +4,15 @@ go 1.21.4
 
 require (
 	github.com/apache/arrow/go/v15 v15.0.0-20240114144300-7e703aae55c1
+	github.com/cloudquery/codegen v0.3.12
 	github.com/cloudquery/plugin-sdk/v4 v4.28.0
+	github.com/invopop/jsonschema v0.12.0
 	github.com/rs/zerolog v1.31.0
 )
 
+// github.com/cloudquery/jsonschema @ cqmain
```
candiduslynx, line 13: better move the replace to the end of the mod file (otherwise it might produce severla additiona sections & not keep the direct & indirect imports separated

candiduslynx (approved):

candiduslynx:
@disq could you also add make gen & check for diff steps to the workflow?

PR #16389, opened 2024-01-31
feat: Add JSON schema to Gremlin destination plugin

disq:
Implements https://github.com/cloudquery/cloudquery/issues/16350


plugins/destination/gremlin/client/spec.go
```diff
@@ -1,35 +1,50 @@
...
+
+	"github.com/invopop/jsonschema"
+	orderedmap "github.com/wk8/go-ordered-map/v2"
 )
 
 type Spec struct {
-	Endpoint string `json:"endpoint"`
-	Insecure bool   `json:"insecure"`
+	// Endpoint for the database. Supported schemes are `wss://` and `ws://`, the default port is `8182`.
+	Endpoint string `json:"endpoint" jsonschema:"required,minLength=1"`
```
candiduslynx, line 15:
```suggestion
	Endpoint string `json:"endpoint" jsonschema:"required,pattern=^ws[s]://[^\n]+$"`
```

plugins/destination/gremlin/client/spec.go
```diff
@@ -91,3 +109,108 @@ func (s *Spec) Validate() error {
 
 	return nil
 }
+
+func (Spec) JSONSchemaExtend(sc *jsonschema.Schema) {
+	authMode := sc.Properties.Value("auth_mode")
+	authMode.Enum = []any{authModeNone, authModeBasic, authModeAWS}
```
candiduslynx, line 115: IMO this should be just done in the `authMode` type schema. Options:
* Implement `JSONSchema` method for `authMode` that will just return the schema for the type
* Implement `JSONSchemaExtend` method for `authMode`

See example in CLI: https://github.com/cloudquery/cloudquery/blob/main/cli/internal/specs/v0/pk_mode.go#L47

plugins/destination/gremlin/client/spec.go
```diff
@@ -91,3 +109,108 @@ func (s *Spec) Validate() error {
...
+
+	sc.OneOf = []*jsonschema.Schema{
+		{
+			// if auth_mode is aws, require aws_region to be set
+			Properties: func() *orderedmap.OrderedMap[string, *jsonschema.Schema] {
+				one := uint64(1)
+				properties := jsonschema.NewProperties()
+
+				authMode := *sc.Properties.Value("auth_mode")
+				authMode.Enum = []any{authModeAWS}
```
candiduslynx, line 125: instead of [`enum`](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.2) use [`const`](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.1.3)

plugins/destination/gremlin/client/spec.go
```diff
@@ -91,3 +109,108 @@ func (s *Spec) Validate() error {
...
 	return nil
 }
+
+func (Spec) JSONSchemaExtend(sc *jsonschema.Schema) {
+	authMode := sc.Properties.Value("auth_mode")
+	authMode.Enum = []any{authModeNone, authModeBasic, authModeAWS}
+
+	sc.OneOf = []*jsonschema.Schema{
+		{
+			// if auth_mode is aws, require aws_region to be set
```
candiduslynx, line 119: this should be implemented via `if` & `then` parts of schema

plugins/destination/gremlin/client/spec.go
```diff
@@ -91,3 +109,108 @@ func (s *Spec) Validate() error {
...
+					return properties
+				}(),
+				Required: []string{"auth_mode"},
+			},
+		},
+	}
+
+	sc.AllOf = []*jsonschema.Schema{
+		{
+			// if username is set, require password to be set and auth_mode to be basic
```
candiduslynx, line 155: shouldn't it be the other way around?
if mode is basic, require username & password.

`Validate` func suggests making `auth_mode:none => username & password should have max len = 0`

candiduslynx (changes requested):

plugins/destination/gremlin/client/spec.go
candiduslynx: @disq please also verify with tests that bad string format won't work while proper ones do work

plugins/destination/gremlin/client/spec.go
disq: It's the same thing but I'll rewrite it with const

candiduslynx:
@disq ping me again once the conflicts are resolved & tests pass

plugins/destination/gremlin/client/schema.json
```diff
@@ -0,0 +1,224 @@
...
+              "aws_region"
+            ]
+          }
+        },
+        {
+          "if": {
+            "properties": {
+              "auth_mode": {
+                "type": "string",
+                "enum": [
```
candiduslynx, line 45: remove enum here. Should be performed via copying the contents:
```go
authMode := *sc.Properties.Value("auth_mode")
authMode.Enum = nil
authMode.Const = ...
```

plugins/destination/gremlin/client/schema.json
```diff
@@ -0,0 +1,224 @@
...
+          "then": {
+            "properties": {
+              "password": {
+                "type": "string",
+                "minLength": 1,
+                "description": "Password to connect to the database. Required when `auth_mode` is `basic`."
+              },
+              "auth_mode": {
+                "type": "string",
+                "enum": [
```
candiduslynx, line 100: remove enum here, too

plugins/destination/gremlin/client/schema.json
```diff
@@ -0,0 +1,224 @@
...
+          "then": {
+            "properties": {
+              "username": {
+                "type": "string",
+                "minLength": 1,
+                "description": "Username to connect to the database. Required when `auth_mode` is `basic`."
+              },
+              "auth_mode": {
+                "type": "string",
+                "enum": [
```
candiduslynx, line 138: here

plugins/destination/gremlin/client/schema.json
```diff
@@ -0,0 +1,224 @@
...
+              "username",
+              "auth_mode"
+            ]
+          }
+        }
+      ],
+      "properties": {
+        "endpoint": {
+          "type": "string",
+          "pattern": "^ws(s?)://[^\n\\:]+(\\:[0-9]{1,5})?$",
```
candiduslynx, line 158: no need to group, _ do we really need to check for port here? maybe just use a simple verification?
```suggestion
          "pattern": "^wss?\://[^\n]+$",
```

plugins/destination/gremlin/client/spec_test.go
```diff
@@ -0,0 +1,132 @@
...
+			Name: "spec with endpoint",
+			Spec: `{"endpoint": "ws://localhost:8182"}`,
+		},
+		{
+			Name: "spec with wss endpoint, without port",
+			Spec: `{"endpoint": "wss://localhost"}`,
+		},
+		{
+			Name: "spec with bad endpoint 1",
+			Spec: `{"endpoint": "http://localhost:8182"}`,
```
candiduslynx, line 26: I'm not sure we need to validate port ourselves, as the Validate logic didn't do it in the first place

candiduslynx (changes requested):

plugins/destination/gremlin/client/schema.json
disq: you want verification but not too much 😂

plugins/destination/gremlin/client/spec_test.go
disq: removed

candiduslynx (approved):

PR #16439, opened 2024-02-01
feat: Add JSON schema to `gcs` destination plugin

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/16407


erezrokah (approved):

disq (changes requested):
main.go not updated with WithJSONSchema?

candiduslynx:
@disq added schema to plugin init

disq (approved):

plugins/destination/gcs/client/spec/schema_test.go
```diff
@@ -0,0 +1,199 @@
...
+			Name: "csv file spec",
+			Spec: `{
+	"format": "csv",
+	"format_spec": {
+		"skip_header": true,
+		"delimiter": "#"
+	},
+	"path": "abc",
+    "bucket": "abc"
+}`,
```
disq, line 23: erm, indent?

candiduslynx: it doesn't matter here + we use the raw string

PR #16440, opened 2024-02-01
feat: Add JSON schema to meilisearch destination plugin

disq:
Closes https://github.com/cloudquery/cloudquery/issues/16436


plugins/destination/meilisearch/docs/overview.md
```diff
@@ -61,20 +61,19 @@ This is the spec used by the Meilisearch destination plugin.
...
 
 - `batch_size_bytes` (`integer`) (optional) (default: `4194304` (= 4 MiB))
 
-  This parameter controls the maximum size of items that may be grouped together to be written as a single write.
+  Maximum size of items that may be grouped together to be written in a single write.
 
 - `batch_timeout` (`duration`) (optional) (default: `20s`)
 
-  This parameter controls the timeout for writing a single batch.
+  Timeout for writing a single batch.
```
candiduslynx, line 76: actually it's the max interval between batches & should reflect so (the doc is wrong here)

candiduslynx (changes requested):

plugins/destination/meilisearch/docs/overview.md
disq: Out of scope, I will update all `batch_timeout` parameter docs in a separate PR?

PR #16442, opened 2024-02-01
chore: No 'This parameter controls' in docs, correct `batch_timeout` explanation

disq:
Destination plugins only. ~Will update after last few PRs are merged~


plugins/destination/azblob/client/spec/spec.go
```diff
@@ -24,17 +24,17 @@ type Spec struct {
...
 	BatchSize *int64 `json:"batch_size" jsonschema:"minimum=1,default=10000"`
 
-	// This parameter controls the maximum size of items that may be grouped together to be written in a single object.
+	// Maximum size of items that may be grouped together to be written in a single object.
 	//
 	// Defaults to `52428800` (50 MiB) unless `no_rotate` is `true` (will be `0` then).
 	BatchSizeBytes *int64 `json:"batch_size_bytes" jsonschema:"minimum=1,default=52428800"`
 
-	// This parameter controls the maximum interval between batch writes.
+	// Inactivity time before starting a new batch.
```
candiduslynx, line 37: This change is incorrect

candiduslynx (changes requested):

plugins/destination/azblob/client/spec/spec.go
disq: how's it looking now?

PR #16446, opened 2024-02-01
feat: Add JSON schema to `kafka` destination plugin

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/16406


erezrokah (approved):
🚀

PR #16465, opened 2024-02-01
feat: Add JSON schema to `s3` destination plugin

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/14618


plugins/destination/s3/client/read.go
```diff
@@ -19,10 +20,11 @@ func (c *Client) Read(ctx context.Context, table *schema.Table, res chan<- arrow
 	if !c.spec.NoRotate {
 		return fmt.Errorf("reading is not supported when no_rotate is false. Table: %q", table.Name)
 	}
-	if strings.Contains(c.spec.Path, PathVarUUID) {
+	if c.spec.PathContainsUUID() {
 		return fmt.Errorf("reading is not supported when path contains uuid variable. Table: %q", table.Name)
 	}
-	name := strings.ReplaceAll(c.spec.Path, PathVarTable, table.Name)
+
+	name := c.spec.ReplacePathVariables(table.Name, uuid.NewString(), time.Time{})
```
candiduslynx, line 27: this is a bit different that the present code, but it's in line with `file` plugin.

disq (approved):

PR #16480, opened 2024-02-02
fix: Use `write_mode` instead of `write-mode` in docs

erezrokah:
#### Summary

Fixes some typos in the docs


erezrokah:
Tests are failing since I opened this from a fork so no access to secrets

disq (approved):

PR #16490, opened 2024-02-02
feat: Add JSON schema to bigquery destination

disq:
Closes https://github.com/cloudquery/cloudquery/issues/16447

### DIFF
--- a/plugins/destination/bigquery/client/schema.json
+++ b/plugins/destination/bigquery/client/schema.json
@@ -31,7 +31,8 @@
             "hour",
             "day"
           ],
-          "description": "The time partitioning to use when creating tables. The partition time column used will always be `_cq_sync_time` so that all rows for a sync run will be partitioned on the hour/day the sync started."
+          "description": "The time partitioning to use when creating tables. The partition time column used will always be `_cq_sync_time` so that all rows for a sync run will be partitioned on the hour/day the sync started.",
+          "default": "none"
         },
         "service_account_key_json": {
           "type": "string",
--- a/plugins/destination/bigquery/client/schema.json
+++ b/plugins/destination/bigquery/client/schema.json
@@ -1,0 +1,70 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/cloudquery/cloudquery/plugins/destination/bigquery/client/spec",
+  "$ref": "#/$defs/Spec",
+  "$defs": {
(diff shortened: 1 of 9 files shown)


plugins/destination/bigquery/client/schema.json
```diff
@@ -0,0 +1,70 @@
...
+          "minLength": 1,
+          "description": "The name of the BigQuery dataset within the project, e.g. `my_dataset`.\n This dataset needs to be created before running a sync or migration."
+        },
+        "dataset_location": {
+          "type": "string",
+          "description": "The data location of the BigQuery dataset. If set, will be used as the default location for job operations.\nPro-tip: this can solve \"dataset not found\" issues for newly created datasets."
+        },
+        "time_partitioning": {
+          "type": "string",
+          "enum": [
```
candiduslynx, line 29: is there a default?

plugins/destination/bigquery/client/spec.go
```diff
@@ -91,3 +117,15 @@ func isValidJson(content string) error {
 	}
 	return nil
 }
+
+func (Spec) JSONSchemaExtend(sc *jsonschema.Schema) {
+	sc.Properties.Value("batch_timeout").Default = "10s"
+}
+
+func (TimePartitioningOption) JSONSchemaExtend(sc *jsonschema.Schema) {
+	sc.Type = "string"
```
candiduslynx, line 126: add default, too

plugins/destination/bigquery/client/spec.go
```diff
@@ -91,3 +117,15 @@ func isValidJson(content string) error {
 	}
 	return nil
 }
+
+func (Spec) JSONSchemaExtend(sc *jsonschema.Schema) {
+	sc.Properties.Value("batch_timeout").Default = "10s"
+}
+
+func (TimePartitioningOption) JSONSchemaExtend(sc *jsonschema.Schema) {
```
candiduslynx, line 125: this can be used instead:
```suggestion
func (TimePartitioningOption) JSONSchema() *jsonschema.Schema) {
```

candiduslynx (changes requested):

plugins/destination/bigquery/client/spec.go
candiduslynx: ah, right, you'll have to provide full spec manually in this case

candiduslynx (approved):

PR #16493, opened 2024-02-02
feat: Add JSON Schema to Okta source plugin

disq:
Closes https://github.com/cloudquery/cloudquery/issues/16453


plugins/source/okta/client/schema.json
```diff
@@ -0,0 +1,66 @@
...
+          "type": "integer",
+          "minimum": 1,
+          "description": "Number of concurrent requests to be made to Okta API.",
+          "default": 10000
+        }
+      },
+      "additionalProperties": false,
+      "type": "object",
+      "required": [
+        "token",
```
candiduslynx, line 61: if that's required, there won't be any way to set this via `OKTA_API_TOKEN`

plugins/source/okta/client/spec.go
```diff
@@ -10,22 +11,37 @@ import (
...
 type (
 	Spec struct {
-		Token       string     `json:"token,omitempty"`
-		Domain      string     `json:"domain,omitempty"`
-		RateLimit   *RateLimit `json:"rate_limit,omitempty"`
-		Debug       bool       `json:"debug,omitempty"`
-		Concurrency int        `json:"concurrency,omitempty"`
+		// Token for Okta API access.
+		// You can set this with an `OKTA_API_TOKEN` environment variable.
+		Token string `json:"token" jsonschema:"required,minLength=1"`
```
candiduslynx, line 16: This shouldn't be required + this shouldn't impose length restrictions, as the `OKTA_API_TOKEN` is the alternative to set this.
IMO, the correct way is to leave the default representation & doc that empty/missing will result in `OKTA_API_TOKEN` env var check.
```suggestion
		Token string `json:"token"`
```

candiduslynx: Alt: just drop the env var & make a breaking change, then rebase the code

plugins/source/okta/client/spec/gen/main.go
```diff
@@ -0,0 +1,26 @@
...
+	"path"
+	"runtime"
+
+	"github.com/cloudquery/cloudquery/plugins/source/okta/client"
+	cqjsonschema "github.com/cloudquery/codegen/jsonschema"
+)
+
+func main() {
+	fmt.Println("Generating JSON schema for plugin spec")
+	cqjsonschema.GenerateIntoFile(new(client.Spec), path.Join(currDir(), "../..", "schema.json"),
```
candiduslynx, line 15: I prefer moving `spec.go` to `spec` pkg, so that all about spec is in a single pkg (schema, too). nit

plugins/source/okta/client/spec.go
```diff
@@ -10,22 +11,37 @@ import (
...
+		Debug bool `json:"debug,omitempty" jsonschema:"default=false"`
+
+		// Number of concurrent requests to be made to Okta API.
+		Concurrency int `json:"concurrency" jsonschema:"minimum=1,default=10000"`
 	}
 	RateLimit struct {
-		MaxBackoff time.Duration `json:"max_backoff,omitempty"`
-		MaxRetries int32         `json:"max_retries,omitempty"`
+		// Max backoff interval to be used.
+		MaxBackoff time.Duration `json:"max_backoff,omitempty" jsonschema:"minimum=30,default=30"`
```
candiduslynx, line 31: move to configtype.Duration, as `time.Duration` is wrongly treated as integer

plugins/source/okta/docs/overview.md
```diff
@@ -55,12 +55,12 @@ Make sure you use [environment variable expansion](/docs/advanced-topics/environ
...
 ### Rate limit spec
 
-- `max_backoff` (`duration`, optional. Default: `5s`)
+- `max_backoff` (`duration`, optional. Default: `30s`)
 
   Max backoff interval to be used.
   If the value specified is less than the default one, the default one is used.
 
-- `max_retries` (`int32`, optional. Default: `3`)
+- `max_retries` (`int32`, optional. Default: `2`)
```
candiduslynx, line 63:
```suggestion
- `max_retries` (`integer`, optional. Default: `2`)
```

plugins/source/okta/go.mod
```diff
@@ -129,3 +130,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// github.com/cloudquery/jsonschema @ cqmain
+replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0-20240202134451-d771afde32fb
```
candiduslynx, line 135: use latest from the `cqmain` (recently updated)

candiduslynx (changes requested):

plugins/source/okta/client/spec/gen/main.go
disq: spec-related code here is too short for that, would mean more changes in the codebase in other places.

plugins/source/okta/client/spec_test.go
```diff
@@ -0,0 +1,86 @@
...
+			Spec: `{"token": "tok", "domain": "https://domain.okta.com", "rate_limit": {"max_backoff": "60s"}}`,
+		},
+		{
+			Name: "spec with token and domain and invalid rate limit",
+			Spec: `{"token": "tok", "domain": "https://domain.okta.com", "rate_limit": {"max_backoff": true}}`,
+			Err:  true,
+		},
+		{
+			Name: "spec with token and domain and zero rate limit",
+			Spec: `{"token": "tok", "domain": "https://domain.okta.com", "rate_limit": {"max_backoff": 0}}`,
```
candiduslynx, line 53: rl tests could be extracted into a separate func, see https://github.com/cloudquery/cloudquery/blob/main/plugins/source/aws/client/spec/organization_test.go

plugins/source/okta/client/spec_test.go
```diff
@@ -0,0 +1,86 @@
...
+			Name: "spec with token and domain and null rate limit",
+			Spec: `{"token": "tok", "domain": "https://domain.okta.com", "rate_limit": null}`,
+		},
+		{
+			Name: "spec with token and domain and rate limit",
+			Spec: `{"token": "tok", "domain": "https://domain.okta.com", "rate_limit": {"max_backoff": "60s"}}`,
+		},
+		{
+			Name: "spec with token and domain and invalid rate limit",
+			Spec: `{"token": "tok", "domain": "https://domain.okta.com", "rate_limit": {"max_backoff": true}}`,
```
candiduslynx, line 48: remove token from tests (only leave for the tests that are actually testing token)

candiduslynx (commented):
I still think it's a good time to just drop the env var in a separate PR, then rebase & reintroduce token as a required param

plugins/source/okta/client/spec_test.go
candiduslynx: Or merge https://github.com/cloudquery/cloudquery/pull/16498 & reintroduce requirement for `token` (better IMO)

candiduslynx:
@disq now that the https://github.com/cloudquery/cloudquery/pull/16498 is merged, you'll need to merge/rebase the branch & require `token` once again

plugins/source/okta/client/spec_test.go
disq: done in https://github.com/cloudquery/cloudquery/pull/16493/commits/e8180cc2682a690d964c3790ea1b4d9bcb25b7d8

candiduslynx (approved):

PR #16497, opened 2024-02-03
feat: Add `azure_sql_managed_server_security_alert_policies` table

KMConner:
#### Summary

Added `azure_sql_managed_server_security_alert_policies` table, which retrieves Managed Server Security Alert Policies of Managed Instance.


yevgenypats:
Hey @KMConner , thanks for the contribution but unfourtuently we can't accept it at the moment as those plugins are open-core - https://github.com/cloudquery/cloudquery/tree/main?tab=readme-ov-file#open-source-and-open-core. You can of-course have your own build if you a current need but those tables will be coming up in our enterprise version.

Sorry this wasn't visible enough - I'll try to update it in the issue templates as well. Happy to answer any other questions

PR #16540, opened 2024-02-07
feat: Enable Skipping of Organization Discovery

bbernays:
#### Summary

If a user runs the GCP in parallel it can run into rate limits for discovering organizations. In cases when there are no org resources this allows users to skip those api calls altogether

### DIFF
--- a/plugins/source/gcp/client/spec/spec.go
+++ b/plugins/source/gcp/client/spec/spec.go
@@ -105,6 +105,9 @@ type Spec struct {
 
 	// Service Account impersonation configuration.
 	ServiceAccountImpersonation *CredentialsConfig `json:"service_account_impersonation"`
+
+	// If enabled CloudQuery will skip syncing organization level resources
+	SkipOrganizationResources bool `json:"skip_organization_resources"`
 }
 
 func (s *Spec) Validate() error {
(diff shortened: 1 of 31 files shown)


plugins/source/gcp/client/spec/spec.go
```diff
@@ -105,6 +105,9 @@ type Spec struct {
 
 	// Service Account impersonation configuration.
 	ServiceAccountImpersonation *CredentialsConfig `json:"service_account_impersonation"`
+
+	// If enabled CloudQuery will skip syncing organization level resources
+	SkipOrganizationResources bool `json:"skip_organization_resources"`
```
disq, line 110:
```suggestion
	SkipOrganizationResources bool `json:"skip_organization_resources" jsonschema:"default=false"`
```

plugins/source/gcp/resources/plugin/client.go
```diff
@@ -69,6 +72,19 @@ func (c *Client) Sync(ctx context.Context, options plugin.SyncOptions, res chan<
 	if err != nil {
 		return err
 	}
+	var skippedOrgTables []string
+	for _, table := range tables {
+		if c.syncClient.Spec.SkipOrganizationResources && table.Multiplex != nil {
+			// get the name of the multiplexer via reflection
+			multiplexerName := runtime.FuncForPC(reflect.ValueOf(table.Multiplex).Pointer()).Name()
+			if strings.HasPrefix(multiplexerName, "github.com/cloudquery/cloudquery/plugins/source/gcp/client.OrgMultiplex") {
```
erezrokah, line 80: Might be confusing to see this code in the private version of the repo, maybe only check the function name?

erezrokah (approved):

PR #16546, opened 2024-02-08
feat: Validate plugin spec before init

candiduslynx:
This PR adds spec validation before passing it to the plugin.

It also extracts the init logic into a handy `initPlugin` func.

The decision structure is the following (we allow a bit of leeway here if the schema provided isn't a correct JSON schema: the CLI will **skip validation** _instead of_ failing the sync/init)

```mermaid
flowchart TD
    A{No Connection?} ----> |true| S
    A{No Connection?} --> |false| G[Get spec schema]
    subgraph validate
    G --> |non-empty schema| P[parse schema]
    G --> |failed call| E[error-out]
    P --> |OK| V[validate spec]
    V --> |failed| E
    end
    P --> |failed| S
    G --> |unimplemented| S
    G --> |empty schema| S
    V --> |OK| I[init]
    S[Skip validation] --> I  
```

### DIFF
--- a/cli/cmd/specs.go
+++ b/cli/cmd/specs.go
@@ -169,7 +169,7 @@ func parseJSONSchema(jsonSchema string) (*jsonschema.Schema, error) {
 			// We add resource as `file`, but there's none, actually.
 			// So, we need to prettify message a bit.
 			return nil, fmt.Errorf("jsonschema compilation failed: %w",
-				errors.New(strings.Replace(se.Err.Error(), "jsonschema: '' ", "jsonschema: ", 1)))
+				errors.New(strings.Replace(se.Err.Error(), "jsonschema: '' ", "", 1)))
 		}
 		return nil, err
 	}
--- a/cli/cmd/specs.go
+++ b/cli/cmd/specs.go
@@ -168,7 +168,6 @@ func parseJSONSchema(jsonSchema string) (*jsonschema.Schema, error) {
 		if errors.As(err, &se); se != nil && se.Err != nil {
 			// We add resource as `file`, but there's none, actually.
 			// So, we need to prettify message a bit.
-
 			return nil, fmt.Errorf("jsonschema compilation failed: %w",
 				errors.New(strings.Replace(se.Err.Error(), "jsonschema: '' ", "jsonschema: ", 1)))
 		}
--- a/cli/cmd/specs.go
+++ b/cli/cmd/specs.go
@@ -1,10 +1,18 @@
 package cmd
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/cloudquery/cloudquery/cli/internal/specs/v0"
+	"github.com/cloudquery/plugin-pb-go/pb/plugin/v3"
 	pbSpecs "github.com/cloudquery/plugin-pb-go/specs"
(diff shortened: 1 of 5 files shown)


cli/cmd/specs.go
```diff
@@ -86,3 +94,72 @@ func CLIDestinationSpecToPbSpec(spec specs.Destination) pbSpecs.Destination {
...
+func parseJSONSchema(jsonSchema string) (*jsonschema.Schema, error) {
+	c := jsonschema.NewCompiler()
+	c.Draft = jsonschema.Draft2020
+	c.AssertFormat = true
+
+	if err := c.AddResource("schema.json", strings.NewReader(jsonSchema)); err != nil {
+		return nil, err
+	}
+
+	return c.Compile("schema.json")
```
disq, line 164: On error, the error is returned in an ugly format. We might want to parse and beautify this a little bit.
> jsonschema: 'file:///long/package/path/and/then/schema.json' does not validate with https://json-schema.org/draft/2020-12/schema#/type: rest of the message here

Maybe do something like (from the other project...)
```go
	var e2 *jsonschema.SchemaError
	if errors.As(err, &e2); e2 != nil && e2.Err != nil {
		msg := e2.Err.Error()
		// comment here explaining why we're doing this
		msg = strings.Replace(msg, "jsonschema: '' ", "jsonschema: ", 1)
		return errors.New(msg)
	}
	return err
```

disq, line 164: On error, the error is returned in an ugly format. We might want to parse and beautify this a little bit.
> jsonschema: 'file:///long/package/path/and/then/schema.json' does not validate with https://json-schema.org/draft/2020-12/schema#/type: rest of the message here

Maybe do something like (from the other project...)
```go
	var e2 *jsonschema.SchemaError
	if errors.As(err, &e2); e2 != nil && e2.Err != nil {
		msg := e2.Err.Error()
		// comment here explaining why we're doing this
		msg = strings.Replace(msg, "jsonschema: '' ", "jsonschema: ", 1)
		return errors.New(msg)
	}
	return err
```

candiduslynx: 7f2834a78933fe51d3091a6d13ee5eb21ba6f191

erezrokah (approved):

cli/cmd/specs.go
```diff
@@ -86,3 +95,84 @@ func CLIDestinationSpecToPbSpec(spec specs.Destination) pbSpecs.Destination {
...
+	}
+
+	sc, err := c.Compile("schema.json")
+	if err != nil {
+		var se *jsonschema.SchemaError
+		if errors.As(err, &se); se != nil && se.Err != nil {
+			// We add resource as `file`, but there's none, actually.
+			// So, we need to prettify message a bit.
+			return nil, fmt.Errorf("jsonschema compilation failed: %w",
+				errors.New(strings.Replace(se.Err.Error(), "jsonschema: '' ", "jsonschema: ", 1)))
```
disq, line 172:
```suggestion
				errors.New(strings.Replace(se.Err.Error(), "jsonschema: '' ", "", 1)))
```

if we're adding the prefix then no need for a second `jsonschema: `

disq (approved):

hermanschaaf (approved):

PR #16552, opened 2024-02-08
fix: Clarify that `aws_debug: true` needs debug log level

erezrokah:
#### Summary

Tried to enable AWS debug logs during working on auth related issues and didn't understand why I didn't see them

### DIFF
--- a/plugins/source/aws/client/spec/spec.go
+++ b/plugins/source/aws/client/spec/spec.go
@@ -21,7 +21,7 @@ type Spec struct {
 	// In AWS organization mode, CloudQuery will source all accounts underneath automatically.
 	Organization *Organization `json:"org"`
 
-	// If `true`, will log AWS debug logs, including retries and other request/response metadata.
+	// If `true`, will log AWS debug logs, including retries and other request/response metadata. Requires passing `--log-level debug` to the CloudQuery CLI.
 	AWSDebug bool `json:"aws_debug,omitempty"`
 
 	// Defines the maximum number of times an API request will be retried.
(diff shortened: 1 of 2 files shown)


plugins/source/aws/client/spec/spec.go
```diff
@@ -21,7 +21,7 @@ type Spec struct {
 	// In AWS organization mode, CloudQuery will source all accounts underneath automatically.
 	Organization *Organization `json:"org"`
 
-	// If `true`, will log AWS debug logs, including retries and other request/response metadata.
+	// If `true`, will log AWS debug logs, including retries and other request/response metadata. Requires passing `--log-level debug` to the CloudQuery CLI.
```
hermanschaaf, line 24:
```suggestion
	// If `true`, will log AWS debug logs, including retries and other request/response metadata. Requires debug-level logs to be enabled.
```

This comment might show in managed syncs, where passing parameters to the CLI isn't possible at the moment 🤔 Can we make it a bit more general maybe?

erezrokah: Not sure we would expose this in managed syncs (otherwise we'd need to enable CLI debugging for this to work anyways)

hermanschaaf (approved):

plugins/source/aws/client/spec/spec.go
candiduslynx: IMO we need to enable debugging as an extra flag (either debug level or `debug_log: bool`)

candiduslynx (approved):

bbernays (approved):
The the CQ CLI needs to have Debug logs enabled because of how we implemented the AWS Logger:

https://github.com/cloudquery/cloudquery/blob/9a0e6e5adfc2ac317ffcadca9af8c04cb703b9e1/plugins/source/aws/client/client.go#L245-L251

Only `warn` messages get printed, all others are `debug`

PR #16553, opened 2024-02-08
fix: Extract `HUBSPOT_APP_TOKEN` to spec

disq:
Since this is token's first appearance on the spec I don't think we should remove the env var yet.

### DIFF
--- a/plugins/source/hubspot/client/spec/spec.go
+++ b/plugins/source/hubspot/client/spec/spec.go
@@ -1,21 +1,32 @@
 package spec
 
-import _ "embed"
+import (
+	_ "embed"
+	"fmt"
+	"os"
+)
 
 const (
 	defaultConcurrency = 1000
 )
 
 type Spec struct {
+	// In order for CloudQuery to sync resources from your HubSpot setup, you will need to authenticate with your HubSpot account. You will need to create a [HubSpot Private App](https://developers.hubspot.com/docs/api/private-apps), and copy the App Token here.
+	// If not specified `HUBSPOT_APP_TOKEN` environment variable will be used instead.
+	AppToken string `json:"app_token,omitempty" jsonschema:"minLength=1"`
 	// Max number of requests per second to perform against the Hubspot API.
-	MaxRequestsPerSecond *int `yaml:"max_requests_per_second,omitempty" json:"max_requests_per_second,omitempty" jsonschema:"minimum=1,default=5"`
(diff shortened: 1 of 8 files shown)


plugins/source/hubspot/client/spec/spec.go
```diff
@@ -1,21 +1,32 @@
...
-	TableOptions TableOptions `yaml:"table_options,omitempty" json:"table_options,omitempty"`
+	TableOptions TableOptions `json:"table_options,omitempty"`
 	// Concurrency setting for the CloudQuery scheduler.
-	Concurrency int `yaml:"concurrency,omitempty" json:"concurrency,omitempty" jsonschema:"minimum=1,default=1000"`
+	Concurrency int `json:"concurrency,omitempty" jsonschema:"minimum=1,default=1000"`
 }
 
-func (spec *Spec) SetDefaults() {
+func (s *Spec) SetDefaults() {
+	if s.AppToken == "" {
```
candiduslynx, line 26:
```suggestion
	if len(s.AppToken) == 0 {
```

plugins/source/hubspot/client/spec/spec.go
```diff
@@ -24,13 +35,20 @@ func (spec *Spec) SetDefaults() {
...
 
-	if spec.MaxRequestsPerSecond == nil || *spec.MaxRequestsPerSecond <= 0 {
-		spec.MaxRequestsPerSecond = &defaultRateLimitPerSecond
+	if s.MaxRequestsPerSecond <= 0 {
+		s.MaxRequestsPerSecond = defaultRateLimitPerSecond
 	}
 
-	if spec.Concurrency == 0 {
-		spec.Concurrency = defaultConcurrency
+	if s.Concurrency == 0 {
```
candiduslynx, line 42:
```suggestion
	if s.Concurrency <= 0 {
```

plugins/source/hubspot/docs/overview.md
```diff
@@ -31,7 +31,11 @@ The following example sets up the HubSpot plugin, and connects it to a postgresq
 
 This is the specs that can be used by the HubSpot source Plugin.
 
+- `app_token` (`string`, optional, default: `HUBSPOT_APP_TOKEN` environment variable)
```
candiduslynx, line 34:
```suggestion
- `app_token` (`string`)  (optional) (default: `HUBSPOT_APP_TOKEN` environment variable value)
```
as in other plugins.

Additionally, it might be useful to update other params & reolace stuff like `int` to `integer`

candiduslynx (approved):

PR #16555, opened 2024-02-08
fix: Handle addition of Primary Key Components

bbernays:
#### Summary

When migrating from regular composite PKs to PKComponents a PKComponent is added that is not already a PK the current postgres plugin will fail because it will attempt to remove the NOT NULL constraint.

Example of issue:

version 1: table has pks on `id`
version 2: table has pk on `id`, `user`
version 3: table has pk on `_cq_id` and pk component on `id` and `user`

If a user upgrades from v1 to v3 the destination will fail.


erezrokah (approved):
I think we could also ignore the constraint not exists error regardless as we shouldn't fail if we try to remove something that doesn't exist

bbernays:
> I think we could also ignore the constraint not exists error regardless as we shouldn't fail if we try to remove something that doesn't exist

I agree completely but I couldn't find any way to ignore the error as the entire transaction is put into an error state.

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -88,22 +88,21 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
...
-				// 1. We saw the table with this name
-				// 2. If this is still empty on insert, the table in the database doesn't have a PK constraint
-				c.pgTablesToPKConstraints[tableName] = ""
+
+		// We always want to record that we saw the table, even if it doesn't have a PK constraint.
+		if _, ok := c.pgTablesToPKConstraints[tableName]; !ok {
+			c.pgTablesToPKConstraints[tableName] = pkConstraintDetails{}
+		}
+		if pkName != "" {
+			if constraint, ok := c.pgTablesToPKConstraints[tableName]; ok {
```
candiduslynx, line 97: You create a copy here & don't update the entry in map, iirc

candiduslynx (changes requested):

plugins/destination/postgresql/client/list_tables.go
bbernays: I'm not sure I follow, on line 101 I update the entry in the map

candiduslynx: Right. What I mean is instead of using the values in the map & doing the overwrite you can use `map-value-pointer` way:
```go
type someStruct struct {
    fields []string
}

aMap := map[string]*someStruct{}

//...

entry, ok := aMap["key"]
if !ok {
    entry := new(someStruct) // or &someStruct{fields:...}
    aMap["key"] = entry
}
// now in ANY case entry is OK to use directly
entry.fields = append(entry.fields, "some", "more", "values")
```

bbernays: done in c21bf852fcdebd57f0fe2341a44ceded16aaad58

plugins/destination/postgresql/client/list_tables.go
```diff
@@ -88,22 +88,18 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
...
+
+		// We always want to record that we saw the table, even if it doesn't have a PK constraint.
+		if _, ok := c.pgTablesToPKConstraints[tableName]; !ok {
+			c.pgTablesToPKConstraints[tableName] = &pkConstraintDetails{}
+		}
+		if pkName != "" {
+			c.pgTablesToPKConstraints[tableName].name = pkName
+			// we want to store any current column that is part of the PK
+			c.pgTablesToPKConstraints[tableName].columns = append(c.pgTablesToPKConstraints[tableName].columns, columnName)
+			table.PkConstraintName = pkName
```
candiduslynx, line 100:
```suggestion
		entry, ok := c.pgTablesToPKConstraints[tableName]
		if !ok {
			entry = new(pkConstraintDetails)
			c.pgTablesToPKConstraints[tableName] = entry
		}
		if pkName != "" {
			entry.name = pkName
			// we want to store any current column that is part of the PK
			entry.columns = append(entry.columns, columnName)
			table.PkConstraintName = pkName
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -87,7 +87,9 @@ func (c *Client) normalizeTable(table *schema.Table) *schema.Table {
 		col.Type = c.PgToSchemaType(c.SchemaTypeToPg(col.Type))
 		normalizedTable.Columns = append(normalizedTable.Columns, col)
 		// pgTablesToPKConstraints is populated when handling migrate messages
-		normalizedTable.PkConstraintName = c.pgTablesToPKConstraints[table.Name]
+		if c.pgTablesToPKConstraints[table.Name] != nil {
+			normalizedTable.PkConstraintName = c.pgTablesToPKConstraints[table.Name].name
```
candiduslynx, line 91:
```suggestion
		if entry, _ := c.pgTablesToPKConstraints[table.Name]; entry != nil {
			normalizedTable.PkConstraintName = entry.name
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -211,21 +213,42 @@ func (c *Client) migrateToCQID(ctx context.Context, table *schema.Table, _ schem
...
+			}
+		}
+		tx, err = conn.BeginTx(ctx, pgx.TxOptions{
+			IsoLevel: pgx.Serializable,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to begin transaction: %w", err)
+		}
+	}
+	for _, colName := range c.pgTablesToPKConstraints[tableName].columns {
```
candiduslynx, line 244:
```suggestion
	entry, _ := c.pgTablesToPKConstraints[tableName]
	if entry == nil {
		entry = new(pkConstraintDetails)
	}
	for _, colName := range entry.columns {
```

candiduslynx (changes requested):
a couple comments on pointer handling

plugins/destination/postgresql/client/migrate.go
bbernays: All tables are guaranteed to be in the map at this point

bbernays: Done in 1ee5ff6ca48caa7ae4bf950f406f026c85f54f70

plugins/destination/postgresql/client/list_tables.go
bbernays: Done in 1ee5ff6ca48caa7ae4bf950f406f026c85f54f70

candiduslynx (approved):

PR #16618, opened 2024-02-13
fix: Stop DynamoDB resolvers from making API calls that always fail

bbernays:
#### Summary

`aws_dynamodb_table_replica_auto_scalings`: This is only supported when GlobalTableVersion is `2019.11.21` not `2017.11.29`

`aws_dynamodb_global_tables` do not support tags. Leaving the column in place to not break backwards compatibility


plugins/source/aws/resources/services/dynamodb/global_tables.go
```diff
@@ -34,9 +34,9 @@ This table only contains version 2017.11.29 (Legacy) Global Tables. See aws_dyna
 				PrimaryKeyComponent: true,
 			},
 			{
-				Name:     "tags",
-				Type:     sdkTypes.ExtensionTypes.JSON,
-				Resolver: resolveDynamodbGlobalTableTags,
+				Name:          "tags",
```
candiduslynx, line 37: this will make this column empty, maybe we need to just drop it?

hermanschaaf (approved):

disq (approved):

PR #16655, opened 2024-02-16
fix: Correctly gather Tags for RDS & DynamoDB recovery points.

AshCorr:
#### Summary

 - Fix tags for RDS Vault Recovery points not being collected.
   - We were skipping collecting tags due to a `case` statement missing for RDS 
 - Fix tags for DyanmoDB Vault Recovery points not being collected.
   - Back in [this PR](https://github.com/cloudquery/cloudquery/pull/15781) I started skipping ListTags when the `ResourceArn` was set to `dynamodb`, this is wrong, it should be when the `RecoveryPointArn` is set to `dynamodb` or else we end up skipping all DynamoDB tags. 

#### Screenshots
Before
image
After
image


bbernays (approved):

bbernays:
@AshCorr - Thank you for this PR! It looks like the tests are not passing. Do you want to fix the test or should I handle it?

PR #16668, opened 2024-02-16
feat: Use Appender interface

disq:
Avoid writing to parquet if we can (appends only)

~Using upstream driver has some shortcomings from the get go: The interface decides on the schema based on passed types, so having untyped `nil`s don't help. But typed nils are also not handled. The [fork/PR](https://github.com/marcboeker/go-duckdb/pull/150) fixes these issues but I still get 0 rows on the test db (or motherduck)~ The PR is now merged upstream and tagged as 1.6.1.

~I had to use my [fork](https://github.com/disq/go-duckdb/tree/feat/allow-nulls-in-appender-first-row) to get around nils, another way to this would be if the `appender.initColTypes` was exported, then we could just pass on concrete types for init and then use nullable data as normal.~

~Another issue: We create JSON types as `json` but appender doesn't know about it. Sending `string` instead (which is what the json alias points to internally in duckdb, varchar) doesn't work. Might need the newer appender implementation for this to go forward.~

Example config:
```yaml
kind: source
spec:
  name: test
  path: cloudquery/test
  registry: cloudquery
  version: v4.0.8
#  tables: ["*"]
  tables: [ "test_some_table" ]
  destinations: ["duckdb"]
  spec:
    num_clients: 1
    num_rows: 4
    num_sub_rows: 1
#    num_clients: 10
#    num_rows: 10000
    num_sub_rows: 0
#    num_sub_rows: 10
---
kind: destination
spec:
  name: duckdb
  version: v0.0.0
  registry: grpc
  path: localhost:7777
#  write_mode: "overwrite"
  write_mode: "append"
  spec:
    connection_string: tmp_duck.db
#    connection_string: md:my_db
    # Optional parameters
    batch_size: 10000
#    batch_size_bytes: 14194304 # 14 MiB
    # debug: false
```


plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,114 @@
...
+		case arrow.STRING:
+			return ptrOf[string]()
+		case arrow.BINARY, arrow.LARGE_BINARY, arrow.FIXED_SIZE_BINARY:
+			return ptrOf[[]byte]()
+		default:
+			return ptrOf[string]()
+		}
+	}
+
+	if arrow.TypeEqual(arr.DataType(), types.NewUUIDType()) {
```
disq, line 53: This is probably ineffective

plugins/destination/duckdb/client/client.go
```diff
@@ -56,12 +57,11 @@ func New(ctx context.Context, logger zerolog.Logger, spec []byte, _ plugin.NewCl
 			internalPlugin.Team, internalPlugin.Kind, internalPlugin.Name, strings.TrimPrefix(internalPlugin.Version, "v"), runtime.GOOS, runtime.GOARCH)
 	}
 
-	c.connector, err = duckdb.NewConnector(c.spec.ConnectionString, nil)
+	dbconnector, err := duckdb.NewConnector(c.spec.ConnectionString, nil)
```
candiduslynx, line 60: did you forget to init `c.conn` somewhere around here?

plugins/destination/duckdb/client/client.go
```diff
@@ -88,8 +86,20 @@ func (c *Client) Close(ctx context.Context) error {
...
+	err := c.db.Close()
 	c.db = nil
+
+	if c.conn != nil {
+		err2 := c.conn.Close()
+		if err2 != nil {
+			fmt.Println("conn close failed: ", err2)
+		}
+		c.conn = nil
+		if err == nil {
```
candiduslynx, line 98: I'd just go with simple
```go
return errors.Join(c.db.Close(), c.conn.Close())
```

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+func getValue(arr arrow.Array, i int) any {
+	if !arr.IsValid(i) {
+		if arrow.TypeEqual(arr.DataType(), types.NewUUIDType()) {
+			return duckdb.UUID{}
+		}
+
+		// return typed nil
+		switch arr.DataType().ID() {
```
candiduslynx, line 20: IMO it's better to do a type switch instead (see ch dst):
```go
switch dt := arr.DataType().(type)
```
with stuff like `*arrow.TImestampType`

plugins/destination/duckdb/client/transform_gotype.go
candiduslynx: can be incorporated with the type switch as `case *types.UUIDType:`

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/marcboeker/go-duckdb"
+)
+
+func getValue(arr arrow.Array, i int) any {
+	if !arr.IsValid(i) {
+		if arrow.TypeEqual(arr.DataType(), types.NewUUIDType()) {
+			return duckdb.UUID{}
+		}
+
+		// return typed nil
```
candiduslynx, line 19: should be moved to a separate func

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,115 @@
...
+			return ptrOf[string]()
+		}
+	}
+
+	if arrow.TypeEqual(arr.DataType(), types.NewUUIDType()) {
+		v, _ := arr.(*types.UUIDArray).Value(i).MarshalBinary()
+		return duckdb.UUID(v)
+	}
+
+	switch arr.DataType().ID() {
```
candiduslynx, line 59: use type switch here, too

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,115 @@
...
+	var res [][]driver.Value
+	for i := int64(0); i < record.NumRows(); i++ {
+		nc := record.NumCols()
+		row := make([]driver.Value, nc)
+		for j := 0; int64(j) < nc; j++ {
+			row[j] = getValue(record.Column(j), int(i))
+		}
+		res = append(res, row)
+	}
+	return res
```
candiduslynx, line 109: prealloc should work fine here
```suggestion
	res := make([][]driver.Value, record.NumRows())
	nc := record.NumCols()
	for i := range res {
		res[i] = make([]driver.Value, nc)
	}

	for j := int64(0); j < nc; j++ {
		col := record.Column(j)
		for i := range res {
			 res[i][j] = getValue(col, i)
		}
	}
	return res
```

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func ptrOf[T any]() *T {
+	var val *T
+	return val
+}
```
candiduslynx, line 115: I wonder if thsi can be simplified via
```suggestion
func nilPtrOf[T any]() *T {
	return nil
}
```

plugins/destination/duckdb/client/write.go
```diff
@@ -125,12 +127,60 @@ func (c *Client) Write(ctx context.Context, msgs <-chan message.WriteMessage) er
...
+	fmt.Println("appendRows", table.Name)
+	appender, err := duckdb.NewAppenderFromConn(c.conn, "", table.Name)
+	if err != nil {
+		return fmt.Errorf("failed to create appender for %s: %w", table.Name, err)
+	}
+
+	for _, msg := range msgs {
+		arr := transformRecordToGoType(msg.Record)
+		for i := range arr {
+			b, _ := json.Marshal(arr[i])
```
candiduslynx, line 140: err shouldn't be ignored IMO

candiduslynx (changes requested):
some comments

plugins/destination/duckdb/client/client.go
disq: No, c.conn is for appender.

plugins/destination/duckdb/client/write.go
disq: it's just for `fmt.Println("appendRows", string(b))`

plugins/destination/duckdb/client/transform_gotype.go
disq: it works <3

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -42,6 +43,20 @@ func getTypedNilValue(arr arrow.Array) any {
 		return nilPtrOf[string]()
 	case *arrow.BinaryType, *arrow.LargeBinaryType, *arrow.FixedSizeBinaryType:
 		return nilPtrOf[[]byte]()
+	case *arrow.ListType:
```
candiduslynx, line 46: can we use [`arrow.ListLikeType`](https://pkg.go.dev/github.com/apache/arrow/go/v15@v15.0.0/arrow#ListLikeType) interface to match both lists (NB: also matches maps)?

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -63,6 +78,12 @@ func getValue(arr arrow.Array, i int, firstRow bool) any {
 		ts := arr.(*array.Timestamp)
 		timeUnit := ts.DataType().(*arrow.TimestampType).Unit
 		return ts.Value(i).ToTime(timeUnit)
+	case *arrow.Date32Type:
+		ts := arr.(*array.Date32)
```
candiduslynx, line 82: don't switch over `arr.DataType().(type)`, use `switch arr := arr.(type)`

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -42,6 +43,20 @@ func getTypedNilValue(arr arrow.Array) any {
...
+	case *arrow.ListType:
+		arr := arr.(*array.List)
+		v := getTypedNilValue(arr.ListValues())
+		return reflect.New(reflect.SliceOf(reflect.TypeOf(v))).Interface()
+	case *arrow.LargeListType:
+		arr := arr.(*array.LargeList)
+		v := getTypedNilValue(arr.ListValues())
+		return reflect.New(reflect.SliceOf(reflect.TypeOf(v))).Interface()
+	case *arrow.StructType:
+		// Can't create a Go struct dynamically and maps are unsupported: use string
```
candiduslynx, line 55: can we do JSON?

disq: As long as it matches what read() (which uses parquet) expects in type and data, we can try.

plugins/destination/duckdb/client/transform_gotype.go
disq: yeah it matches maps as well, but we can order it carefully

disq: Done in https://github.com/cloudquery/cloudquery/pull/16668/commits/3ce23706feccf97e67ba6820483d56044115844d

plugins/destination/duckdb/client/transform_gotype.go
disq: Turns out ValueStr does exactly that for Struct:

```go
func (a *Struct) ValueStr(i int) string {
	if a.IsNull(i) {
		return NullValueStr
	}

	data, err := json.Marshal(a.GetOneForMarshal(i))
	if err != nil {
		panic(err)
	}
	return string(data)
}
```

plugins/destination/duckdb/client/transform_gotype.go
candiduslynx: I still see it here, maybe it was done only in the values but not in nils?

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,159 @@
...
+		return nilPtrOf[string]()
+	case *arrow.MapType:
```
candiduslynx, line 57: should be moved before `arrow.ListLikeType` case

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,159 @@
...
+				val.Index(i).SetZero() // not sure if this is necessary
+				continue
+			}
+			// slice of pointers, make everything a pointer
+			sv := reflect.ValueOf(getValue(slice, i))
+			psv := reflect.New(sv.Type())
+			psv.Elem().Set(sv)
+			val.Index(i).Set(psv)
+		}
+		return val.Interface()
```
candiduslynx, line 132: I wonder if returning `[]any` with values would work better here...

plugins/destination/duckdb/client/transform_gotype.go
disq: Fixed in https://github.com/cloudquery/cloudquery/pull/16668/commits/34abf5cb6234ebd7cbf36ac97b05fe9f2b564723

plugins/destination/duckdb/client/transform_gotype.go
disq: smart, but no cigar:
> 10:04AM ERR failed to write batch error="failed to append row to test_testdata_table: type mismatch for column 42: \nexpected: []int64, actual: []interface {}" duration=1.473458 len=1 module=duckdb-dest table=test_testdata_table

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -12,51 +12,45 @@ import (
 )
 
 func getTypedNilValue(arr arrow.Array) any {
-	switch arr.DataType().(type) {
-	case *types.UUIDType:
+	switch arr.(type) {
```
candiduslynx, line 15:
```suggestion
	switch arr := arr.(type) {
```

disq: if I do that then I can't merge cases.

candiduslynx: why? you definitely can

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	for i, col := range arrowCols { // i: arrow column index, col: arrow column name
+		j, ok := columnMap[col] // look up the column index in the destination table
+		if !ok {
+			continue
+		}
+		arr := record.Column(i)
+		for k := range res {
+			res[k][j] = getValue(arr, k)
```
disq, line 148: @candiduslynx juicy

plugins/destination/duckdb/go.mod
```diff
@@ -134,3 +134,11 @@ require (
 
 // github.com/cloudquery/jsonschema @ cqmain
 replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0-20240202134451-d771afde32fb
+
+// replace github.com/marcboeker/go-duckdb => ../../../../../disq/go-duckdb
```
disq, line 138: These will need removing once we finalize this PR

plugins/destination/duckdb/client/client_test.go
```diff
@@ -40,6 +49,9 @@ func TestPlugin(t *testing.T) {
 			SkipIntervals: true,
 			// not supported in duckDB for now
 			SkipLargeTypes: true,
+			// not supported in Appender:
+			SkipMaps:    true,
+			SkipStructs: true,
```
candiduslynx, line 54: I wonder if we can remove this & use string/JSON?

plugins/destination/duckdb/client/write.go
```diff
@@ -139,14 +186,16 @@ func (c *Client) WriteTableBatch(ctx context.Context, name string, msgs message.
...
-	if len(table.PrimaryKeys()) == 0 {
-		copyStart := time.Now()
-		defer func() {
-			c.logger.Debug().Str("table", table.Name).Str("duration", time.Since(copyStart).String()).Msg("copy file to table")
-		}()
-		return c.copyFromFile(ctx, name, tmpFile, table)
-	}
-
+	/*
+		if len(table.PrimaryKeys()) == 0 {
```
candiduslynx, line 190: remove?

plugins/destination/duckdb/client/types.go
```diff
@@ -38,16 +38,21 @@ func transformTypeForWriting(dt arrow.DataType) arrow.DataType {
...
 	case *arrow.StructType:
-		builder := new(strings.Builder)
-		builder.WriteString("struct(")
-		for i, field := range dt.Fields() {
-			if i > 0 {
-				builder.WriteString(", ")
+		// use json for struct types for appender compatibility
+		return "json"
+	/*
+		case *arrow.StructType:
```
candiduslynx, line 44: remove obsolete code?

plugins/destination/duckdb/client/transform_gotype.go
```diff
@@ -0,0 +1,157 @@
...
+		for i := 0; i < slice.Len(); i++ {
+			if slice.IsNull(i) {
+				val.Index(i).SetZero() // not sure if this is necessary
+				continue
+			}
+			// slice of pointers, make everything a pointer
+			sv := reflect.ValueOf(getValue(slice, i))
+			psv := reflect.New(sv.Type())
+			psv.Elem().Set(sv)
+			val.Index(i).Set(psv)
```
candiduslynx, line 124: please extract this into a separate func

candiduslynx (approved):

disq:
I'm going to pick this up after https://github.com/cloudquery/cloudquery/pull/16772 is merged

plugins/destination/duckdb/client/client_test.go
disq: Struct as JSON done in https://github.com/cloudquery/cloudquery/pull/16668/commits/ecaf3a261cd96eaf8cf0e62cc7b80003674c5553 will now work on maps

plugins/destination/duckdb/client/write.go
disq: I uncomment it from time to time to test direct parquet copys.

plugins/destination/duckdb/client/client_test.go
disq: Maps as JSON done in https://github.com/cloudquery/cloudquery/pull/16668/commits/6eaf9da7f2ad42923ed578d05531230276168661 PTAL

PR #16682, opened 2024-02-19
fix: Isolate plugin environments in the cloud

hermanschaaf:
This change allows source and destination plugins to only be passed the environment variables they are meant to have access to. (Currently only applicable to Cloud syncs, but we could make it a general feature later, should we want to)

Depends on https://github.com/cloudquery/plugin-pb-go/pull/246


cli/cmd/sync.go
```diff
@@ -281,3 +297,14 @@ func sync(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+func pluginEnv(environ []string, pluginName, kind string) []string {
+	env := make([]string, 0)
+	prefix := "__" + kind + "_" + pluginName + "__"
```
hermanschaaf, line 303: This elaborate prefixing scheme is used to prevent things like a plugin being named `OTEL` and then getting access to OTEL environment variables.

disq, line 303: Technically you can still have the same plugin and kind from different teams on a single sync...

hermanschaaf: Assuming I didn't make a mistake, this should be the `name` field in the plugin spec, so you can't have two of the same

cli/cmd/sync.go
```diff
@@ -281,3 +297,14 @@ func sync(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+func pluginEnv(environ []string, pluginName, kind string) []string {
```
disq, line 301:
```suggestion
func filterPluginEnv(environ []string, pluginName, kind string) []string {
```

Maybe?

hermanschaaf: Yeah makes sense, updated

cli/go.mod
```diff
@@ -145,3 +145,6 @@ require (
 
 // github.com/cloudquery/jsonschema @ cqmain
 replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0-20240202134451-d771afde32fb
+
+// github.com/cloudquery/godebouncer @ fix-race
+replace github.com/vnteamopen/godebouncer => github.com/cloudquery/godebouncer v0.0.0-20240220152946-054f7dbbcf18
```
hermanschaaf, line 150: The new test uncovered a race condition in the godebouncer library. I'm fixing it with a replace directive here, but there is also an upstream PR https://github.com/vnteamopen/godebouncer/pull/28

disq (approved):

PR #16717, opened 2024-02-20
feat: Allow testing of set environment variables

hermanschaaf:
Adds code to the test plugin to allow us to test the environment variables that have been made available to the plugin. This is in order to test https://github.com/cloudquery/cloudquery/pull/16682


plugins/source/test/client/spec.go
```diff
@@ -11,6 +15,11 @@ type Spec struct {
 
 	// Number of rows to generate (per row of parent) in test_sub_table.
 	NumSubRows *int `json:"num_sub_rows" jsonschema:"minimum=0,default=10"`
+
+	// Required environment variables. The plugin will fail if these are not set
+	// to the correct values. Specified in `key=value` format. Use `key=` to specify
+	// that the environment variable should be not set or empty.
+	RequiredEnv []string `json:"required_env" jsonschema:""`
```
disq, line 22: `pattern` would work here for json schema, but I'm not sure about the correct escaping.

```suggestion
	RequiredEnv []string `json:"required_env" jsonschema:"pattern=^[^\=]+\="`
```

Maybe. Could also just do `minLength=2` (would still allow empty slice)

disq (approved):

PR #16727, opened 2024-02-20
fix: Sanitize ID in queries, better debug logging

disq:
Extracted from https://github.com/cloudquery/cloudquery/pull/16668


plugins/destination/duckdb/client/client.go
```diff
@@ -95,16 +93,10 @@ func (c *Client) Close(ctx context.Context) error {
...
-	}
 	if c.spec.Debug {
 		rowsAffected, rowsErr := r.RowsAffected()
-		if rowsErr == nil {
-			c.logger.Debug().Str("query", query).Any("values", args).Int64("rowsAffected", rowsAffected).Msg("exec query")
-		} else {
-			c.logger.Debug().Str("query", query).Any("values", args).Err(rowsErr).Msg("exec query")
-		}
+		errs := errors.Join(err, rowsErr)
+		c.logger.Debug().Str("query", query).Any("values", args).Int64("rowsAffected", rowsAffected).Err(errs).Msg("exec query")
```
disq, line 99: This will log failed queries as well as successful ones.

plugins/destination/duckdb/client/client_test.go
```diff
@@ -21,10 +22,18 @@ func TestPlugin(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	p.SetLogger(zerolog.New(zerolog.NewTestWriter(t)).Level(zerolog.DebugLevel))
```
disq, line 25: This will help us see batchwriter errors (little bit clunky since it's not tied to each test but the whole TestPlugin context) during tests.

plugins/destination/duckdb/client/write.go
```diff
@@ -76,9 +76,9 @@ func (c *Client) upsert(ctx context.Context, tmpTableName string, table *schema.
 		if written > 0 {
 			sb.WriteString(", ")
 		}
-		sb.WriteString(col.Name)
+		sb.WriteString(sanitizeID(col.Name))
```
disq, line 79: `binary` is now(? updated duckdb driver probably?) failing here so we need to sanitize all.

hermanschaaf (approved):

PR #16772, opened 2024-02-20
feat: Add binary and large type handling

disq:
This makes it fully sync the testdata table from the test source plugin (with [fixed](https://github.com/cloudquery/plugin-sdk/pull/1530) list scalars...) using parquet loading.

Also removes `interval` references which didn't work in parquet anyway.


plugins/destination/duckdb/client/transform.go
```diff
@@ -61,6 +61,30 @@ func transformArray(arr arrow.Array) arrow.Array {
...
+		defer bldr.Release()
+		for i := 0; i < arr.Len(); i++ {
+			if arr.IsNull(i) {
+				bldr.AppendNull()
+				continue
+			}
+			val := arr.Value(i)
+			bldr.Append(val)
+		}
+		return bldr.NewArray().(*array.String)
```
candiduslynx, line 87: As you have a typed builder, you can simply can new string array method

plugins/destination/duckdb/client/transform.go
```diff
@@ -61,6 +61,30 @@ func transformArray(arr arrow.Array) arrow.Array {
...
+		defer bldr.Release()
+		for i := 0; i < arr.Len(); i++ {
+			if arr.IsNull(i) {
+				bldr.AppendNull()
+				continue
+			}
+			val := arr.Value(i)
+			bldr.Append(val)
+		}
+		return bldr.NewArray().(*array.Binary)
```
candiduslynx, line 75: Same here

candiduslynx (approved):

PR #16796, opened 2024-02-21
feat: Add `github_repository_sboms` table to sync Github SBOM data

AshCorr:
#### Summary

Add a `github_repository_sboms` table to sync data from [Github's Software Bill Of Materials (SBOM) API](https://docs.github.com/en/rest/dependency-graph/sboms?apiVersion=2022-11-28)

This API can tell us which dependencies(NPM, Maven, etc) a repository has. With the Snyk plugin currently out of commission this provides a crucial way of gathering insights of which dependencies are being used across a Github organization.

This PR includes commits from https://github.com/cloudquery/cloudquery/pull/16797 . I've seperated it into 2 PRs to reduce the cognitive load on reviewers, you may want to review this PR starting from [7d92c1d0ae963f6793d3168e8b44c3a2b5a59f2d](https://github.com/cloudquery/cloudquery/pull/16796/commits/7d92c1d0ae963f6793d3168e8b44c3a2b5a59f2d)

image


plugins/source/github/docs/tables/github_repository_sboms.md
```diff
@@ -0,0 +1,26 @@
...
+
+This table depends on [github_repositories](github_repositories.md).
+
+## Columns
+
+| Name          | Type          |
+| ------------- | ------------- |
+|_cq_id|`uuid`|
+|_cq_parent_id|`uuid`|
+|org (PK)|`utf8`|
```
erezrokah, line 17: Looks like we're missing a PK that to identify this resource (`org` and `repository_id` are set when referencing those columns).

Would `spdxid` work or we need to include `spdx_version` too?

plugins/source/github/resources/services/repositories/sboms.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/google/go-github/v59/github"
+)
+
+func sboms() *schema.Table {
+	return &schema.Table{
+		Name:      "github_repository_sboms",
+		Resolver:  fetchSboms,
+		Multiplex: client.OrgRepositoryMultiplex,
+		Transform: client.TransformWithStruct(&github.SBOMInfo{}),
```
erezrokah, line 16:
```suggestion
		Transform: client.TransformWithStruct(&github.SBOMInfo{}, transformers.WithPrimaryKeys("SPDXID")),
```

Per my previous comment we should do this or add the version too

erezrokah (changes requested):
Thanks for the PR @AshCorr, I think we just need to figure out which columns to use as PKs, then fix the failing test

plugins/source/github/docs/tables/github_repository_sboms.md
AshCorr: Just pushed a change to fix the failing test - My bad, forgot to mock the DependencyGraphService.

I could be wrong, but I think the Org and the Repository ID are already part of the primary key? I don't think we would really want to add any more columns, there should only be 1 SBOM per repository.

AshCorr: All repos will probaly have the same `spdxid` and `spdx_version`, so I'm not sure they're suitable candidates for the primary key. 
image

erezrokah: > I could be wrong, but I think the Org and the Repository ID are already part of the primary key? I don't think we would really want to add any more columns to the PK, there will only be 1 SBOM per repository.

Correct Org and the Repository ID. If there can only be one SBOM per repository we don't need to add additional columns. Going over the docs that seems to be the case

erezrokah (approved):

erezrokah:
This is release in `v8.0.0` of the GitHub plugin, more in https://hub.cloudquery.io/plugins/source/cloudquery/github/latest/docs

Thanks @AshCorr 🚀

PR #16797, opened 2024-02-21
feat!: Bump `go-github` to v59 in Github plugin

AshCorr:
#### Summary

Bump `go-github` to v59 in the Github source plugin.

This allows Cloudquery to take advantage of some of the new Github APIs, specifically the [SBOM API](https://docs.github.com/en/rest/dependency-graph/sboms?apiVersion=2022-11-28) in https://github.com/cloudquery/cloudquery/pull/16796

I've tested this PR in our Cloudquery environment and did not notice any issues arising from this bump, although to be fair we don't collect every table due to resource and time constraints.

Looking at the [Release notes](https://github.com/google/go-github/releases) for go-github I don't see any breaking changes particularly concerning apart from a change of timestamp types in [v50](https://github.com/google/go-github/releases/tag/v50.0.0)

### DIFF
--- a/plugins/source/github/client/client.go
+++ b/plugins/source/github/client/client.go
@@ -13,7 +13,7 @@ import (
 	"github.com/beatlabs/github-auth/key"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/gofri/go-github-ratelimit/github_ratelimit"
-	"github.com/google/go-github/v49/github"
+	"github.com/google/go-github/v59/github"
 	"github.com/rs/zerolog"
 	"golang.org/x/oauth2"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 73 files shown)


plugins/source/github/client/client.go
```diff
@@ -233,7 +233,7 @@ func githubClientForHTTPClient(httpClient *http.Client, logger zerolog.Logger, e
 	}
 
 	if enterpriseSettings != nil {
-		return github.NewEnterpriseClient(enterpriseSettings.BaseURL, enterpriseSettings.UploadURL, rateLimiter)
```
AshCorr, line 236: `NewEnterpriseClient` has been deprecated and throws errors when running `make lint`, the replacement is to use `WithEnterpriseURLs`

erezrokah (approved):
This looks great @AshCorr 💯 Thanks for the PR.
The timestamp changes are not breaking for us as we already know to resolve GitHub timestamp representation as an arrow timestamp https://github.com/cloudquery/cloudquery/blob/8192d6d99898692e9232d29d9ac71f5d2c6d8d4a/plugins/source/github/client/transformers.go#L23

the changes to float are indeed schema breaking (as we don't migrate changes in data types at the moment), so I've updated the PR title from `feat:` to `feat!:` to signal our automation to release this under a major version bump

PR #16820, opened 2024-02-22
fix: Handle multiple changes to a single table

bbernays:
#### Summary

Prior to this change if multiple automigratable changes happened to a single table only the first one would be executed. This change makes it so all changes will be migrated

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -107,11 +107,19 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
 	for _, change := range changes {
 		switch change.Type {
 		case schema.TableColumnChangeTypeAdd:
-			return c.addColumn(ctx, tableName, change.Current)
+			err := c.addColumn(ctx, tableName, change.Current)
+			if err != nil {
+				return err
+			}
+			continue
 		case schema.TableColumnChangeTypeRemove:
 			continue
 		case schema.TableColumnChangeTypeMoveToCQOnly:
-			return c.migrateToCQID(ctx, table, change.Current)
+			err := c.migrateToCQID(ctx, table, change.Current)
+			if err != nil {
+				return err
+			}
+			continue
 		default:
 			return fmt.Errorf("unsupported column change type: %s for column: %v from %v", change.Type.String(), change.Current, change.Previous)
 		}


disq (approved):

erezrokah (approved):

plugins/destination/postgresql/client/migrate.go
```diff
@@ -107,25 +108,59 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
 	for _, change := range changes {
 		switch change.Type {
 		case schema.TableColumnChangeTypeAdd:
-			return c.addColumn(ctx, tableName, change.Current)
+			err := c.addColumn(ctx, tableName, change.Current)
+			if err != nil {
+				return err
+			}
+			continue
```
erezrokah, line 115:
```suggestion
```

I'm not sure we need the `continue`?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -107,25 +108,59 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
...
 		case schema.TableColumnChangeTypeRemove:
```
erezrokah, line 116: Can we remove `case schema.TableColumnChangeTypeRemove` and `default:` altogether?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -107,25 +108,59 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
...
 			continue
 		case schema.TableColumnChangeTypeMoveToCQOnly:
-			return c.migrateToCQID(ctx, table, change.Current)
+			err := c.migrateToCQID(ctx, table, change.Current)
+			if err != nil {
+				return err
+			}
+			continue
```
erezrokah, line 123:
```suggestion
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -107,25 +108,59 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
...
 		}
 	}
 	return nil
 }
 
 func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
+	// The SDK can detect more granular changes than we can handle
+	// We know that when the `TableColumnChangeTypeMoveToCQOnly` is present there will be other changes that were found as well
+	// As long as the only change is to remove PK from columns and add it to _cq_id, we can skip handling the changes
+	// But we need to make sure there is no other changes
```
erezrokah, line 135:
```suggestion
	// But we need to make sure there are no other changes
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -107,25 +108,59 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
...
+		}
+	}
+
 	for _, change := range changes {
 		switch change.Type {
 		case schema.TableColumnChangeTypeAdd:
 			if change.Current.PrimaryKey || change.Current.NotNull {
 				return false
 			}
+			continue
```
erezrokah, line 163:
```suggestion
```

plugins/destination/postgresql/client/migrate.go
```diff
@@ -135,8 +170,15 @@ func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
 				}
 				return false
 			}
+			continue
```
erezrokah, line 173:
```suggestion
```

erezrokah (commented):
Added a few comments mostly on style.
If will be good to add a few tests for the migration logic (at least for the one that triggered the bug)

PR #16837, opened 2024-02-23
feat: Add support for Private docker registries

bbernays:
#### Summary

- [x] Depends on https://github.com/cloudquery/plugin-pb-go/pull/248

I have verified that this functionality works with AWS ECR and with private Github container registry (ghcr.io) by clearing all local images from docker then running `go run main.go install src.yml` where the source was like this:

```
kind: source
spec:
  name: "typeform"
  registry: "docker"
  # docker_registry_auth_token: ${DOCKER_AUTH}
  # path: "012345678910.dkr.ecr.us-east-2.amazonaws.com/bernays:typeform"
  path: "ghcr.io/bbernays/typeform:v1.3.0"
```


bbernays:
> I think we need to run
> 
> https://github.com/cloudquery/cloudquery/blob/f957b492fc28e34096ccd585063ee546937fae74/cli/Makefile#L15
> 
> too

Agreed, but also need to get this merged and then released https://github.com/cloudquery/plugin-pb-go/pull/248

website/pages/docs/reference/source-spec.md
```diff
@@ -46,6 +46,40 @@ The name field may be used to uniquely identify a particular source configuratio
...
+```
+
+GitHub Container Registry:
+The username is `USERNAME` and you use a personal access token as the password. More information can be found [here](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic)
+
+Generating the token for GitHub Container Registry would look like this:
+
+```shell
+export CR_PAT=YOUR_TOKEN
+echo -n "{\"username\":\"USERNAME\",\"password\":\"$$CR_PAT\"}" | base64
```
erezrokah, line 80:
```suggestion
echo -n "{\"username\":\"USERNAME\",\"password\":\"$CR_PAT\"}" | base64
```

Is the double `$$` on purpose?

erezrokah (changes requested):
Looks good, one comment on the docs

erezrokah (approved):

PR #16863, opened 2024-02-26
feat: Deprecate `registry: github`

candiduslynx:
Closes https://github.com/cloudquery/cloudquery-issues/issues/1349


cli/internal/specs/v0/destination.go
```diff
@@ -50,17 +48,3 @@ func (d *Destination) UnmarshalSpec(out any) error {
 func (d *Destination) Validate() error {
 	return d.Metadata.Validate()
 }
-
-func (d Destination) VersionString() string {
```
candiduslynx, line 54: moved to `Metadata` where it belongs

erezrokah (approved):

PR #16872, opened 2024-02-26
fix: GitHub app authentication with GitHub enterprise

erezrokah:
#### Summary

We had a report on Discord that GitHub app authentication doesn't work with GitHub enterprise.
This PR switches the library we use to do app authentication do the one document in the GitHub plugin https://github.com/google/go-github?tab=readme-ov-file#as-a-github-app
instead of the one we're currently using. It has better docs on how to do app auth with GitHub enterprise:
https://github.com/bradleyfalzon/ghinstallation?tab=readme-ov-file#github-enterprise-example

It also updates the code to use `WithAuthToken` when using a personal access token instead of `oauth2.StaticTokenSource` as that's also how the Go client documents it:
https://github.com/google/go-github?tab=readme-ov-file#authentication

I'll tag a release candidate and ask the Discord user to try it out


erezrokah:
> I'll tag a release candidate and ask the Discord user to try it out

User confirms this works

candiduslynx (approved):

PR #16889, opened 2024-02-26
feat: Provide `aws_neptune_host` option for Neptune IAM authentication over proxy

disq:
Recently tested AWS Neptune IAM auth over HAProxy and it didn't work.

### DIFF
--- a/plugins/destination/gremlin/client/client.go
+++ b/plugins/destination/gremlin/client/client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
@@ -95,9 +96,23 @@ func (c *Client) getAuthInfo(ctx context.Context, baseURL string) (gremlingo.Aut
 		// emptyStringSHA256 is a SHA256 of an empty string
 		const emptyStringSHA256 = `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
 
-		req, err := http.NewRequest(http.MethodGet, baseURL, strings.NewReader(""))
+		urlToSign := baseURL
+		if c.spec.AWSNeptuneHost != "" {
+			u, err := url.Parse(baseURL)
+			if err != nil {
+				return nil, err
+			}
+			u.Host = c.spec.AWSNeptuneHost // signer.SignHTTP will use this as the Host header
+			urlToSign = u.String()
+		}
+
+		req, err := http.NewRequest(http.MethodGet, urlToSign, strings.NewReader(""))
 		if err != nil {
 			return nil, err
+		}
+
+		if c.spec.AWSNeptuneHost != "" {
+			req.Header.Set("Host", c.spec.AWSNeptuneHost)
 		}
 
 		cfg, err := config.LoadDefaultConfig(ctx)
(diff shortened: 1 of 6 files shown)


plugins/destination/gremlin/client/client.go
```diff
@@ -95,11 +96,25 @@ func (c *Client) getAuthInfo(ctx context.Context, baseURL string) (gremlingo.Aut
...
+			urlToSign = u.String()
+		}
+
+		req, err := http.NewRequest(http.MethodGet, urlToSign, strings.NewReader(""))
 		if err != nil {
 			return nil, err
 		}
 
+		if c.spec.AWSNeptuneHost != "" {
+			req.Header.Set("Host", c.spec.AWSNeptuneHost)
```
disq, line 115: This is the "real" Host value the server will verify the signature against.

plugins/destination/gremlin/client/client.go
```diff
@@ -95,11 +96,25 @@ func (c *Client) getAuthInfo(ctx context.Context, baseURL string) (gremlingo.Aut
...
 		const emptyStringSHA256 = `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
 
-		req, err := http.NewRequest(http.MethodGet, baseURL, strings.NewReader(""))
+		urlToSign := baseURL
+		if c.spec.AWSNeptuneHost != "" {
+			u, err := url.Parse(baseURL)
+			if err != nil {
+				return nil, err
+			}
+			u.Host = c.spec.AWSNeptuneHost // signer.SignHTTP will use this as the Host header
```
disq, line 105: AWS Signer doesn't take into account if the passed-in `req.Header` has a `Host` field, it simply parses the URL and uses that. So we need to fake it.

plugins/destination/gremlin/client/spec.go
```diff
@@ -33,6 +33,9 @@ type Spec struct {
 	// AWS region to use for AWS IAM authentication. Required when `auth_mode` is `aws`.
 	AWSRegion string `json:"aws_region"`
 
+	// AWS Neptune host header to use with AWS IAM authentication. Required when `auth_mode` is `aws`.
+	AWSNeptuneHost string `json:"aws_neptune_host" jsonschema:"pattern=^[\\w\\.-]+$"`
```
disq, line 37: This is completely optional (if you're running a sync inside the same VPC as your Neptune, you wouldn't need it) but it still needs to be just hostname-valid-chars. No slash, no `scheme://`.

plugins/destination/gremlin/client/client.go
```diff
@@ -95,11 +96,25 @@ func (c *Client) getAuthInfo(ctx context.Context, baseURL string) (gremlingo.Aut
 		// emptyStringSHA256 is a SHA256 of an empty string
 		const emptyStringSHA256 = `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
 
-		req, err := http.NewRequest(http.MethodGet, baseURL, strings.NewReader(""))
+		urlToSign := baseURL
+		if c.spec.AWSNeptuneHost != "" {
```
candiduslynx, line 100:
```suggestion
		if len(c.spec.AWSNeptuneHost) > 0 {
```
As it's a single lookup

candiduslynx (approved):

PR #16942, opened 2024-02-27
fix!: Properly replace JSON values when using `athena: true`

candiduslynx:
Instead of marshaling the `[]byte` array which obviously results in base64 encoding, pass objects to the `Append` directly.

BEGIN_COMMIT_OVERRIDE
fix: Properly replace JSON values when using `athena: true` (https://github.com/cloudquery/cloudquery/pull/16942)
BREAKING-CHANGE: Properly replace JSON values when using `athena: true` (https://github.com/cloudquery/cloudquery/pull/16942). Previously, `json` columns would have been sanitized & then marshaled twice, resulting in `base64` encoded bytes value. Now, the `json` columns have a proper object value.
END_COMMIT_OVERRIDE

### DIFF
--- a/plugins/destination/s3/client/write_test.go
+++ b/plugins/destination/s3/client/write_test.go
@@ -44,17 +44,17 @@ func TestSanitizeJSONKeys(t *testing.T) {
 		},
 		"foo:bar":     "baz",
 		"foo-bar-baz": []any{"baz", map[string]any{"foo:bar": "baz"}},
-		"string": map[string]string{
+		"string": map[string]any{
 			"foo-bar": "baz",
 		},
-		"int": map[string]int{
+		"int": map[string]any{
 			"foo-bar": 123,
 		},
-		"pointer": map[string]*string{
-			"foo-bar": &[]string{"baz"}[0],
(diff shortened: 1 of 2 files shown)


plugins/destination/s3/client/write_test.go
```diff
@@ -44,35 +44,35 @@ func TestSanitizeJSONKeys(t *testing.T) {
...
+		"string": map[string]any{
 			"foo-bar": "baz",
 		},
-		"int": map[string]int{
+		"int": map[string]any{
 			"foo-bar": 123,
 		},
-		"pointer": map[string]*string{
-			"foo-bar": &[]string{"baz"}[0],
+		"pointer": map[string]any{
```
candiduslynx, line 53: we could instead go for `marshal -> unmarshal -> call sanitizeJSONKeysForObject` route

disq (approved):

yevgenypats (approved):

bbernays:
This is a great change, but I think this should be a breaking change. Users could have worked around this issue by decoding the base64 object.

bbernays (approved):

PR #16945, opened 2024-02-27
fix: Use env var replacement in example configuration

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/16944

### DIFF
--- a/plugins/source/gitlab/docs/_configuration.md
+++ b/plugins/source/gitlab/docs/_configuration.md
@@ -11,8 +11,8 @@ spec:
 
   # Gitlab specific configuration
   spec:
-    access_token: "<YOUR_ACCESS_TOKEN_HERE>"
-    base_url: "<INSTANCE_URL>" # optional, leave empty for GitLab SaaS
-    # Optional parameters
-    # concurrency: 10000
+    # required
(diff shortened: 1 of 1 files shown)


plugins/source/gitlab/docs/_configuration.md
```diff
@@ -11,8 +11,8 @@ spec:
...
   # Gitlab specific configuration
   spec:
-    access_token: "<YOUR_ACCESS_TOKEN_HERE>"
-    base_url: "<INSTANCE_URL>" # optional, leave empty for GitLab SaaS
-    # Optional parameters
-    # concurrency: 10000
+    # required
+    access_token: "${GITLAB_ACCESS_TOKEN}"
+    # optional, leave empty for GitLab SaaS
+    # base_url: "<INSTANCE_URL>"
```
candiduslynx, line 17:
```suggestion
    # base_url: "${GITLAB_INSTANCE_URL}"
```

erezrokah: Actually we won't want since this is a comment (it will show up in the managed sync wizard as a default secret to configure). Also I'm not sure it should be an default env variable, see:

image

candiduslynx (approved):

PR #16955, opened 2024-02-28
fix: Remove stream_arn regex validation

mnorbury:
The frontend was showing the following validation error:

`jsonschema: '/stream_arn' does not validate with https://github.com/cloudquery/cloudquery/plugins/destination/firehose/client/spec/spec#/$ref/properties/stream_arn/pattern: does not match pattern '^arn:[^:]+:firehose:[^:]+:[^:]+:deliverystream/[^:]+$'`

Which I think is down to the `/` needing to be escaped. The `spec_test.go` tests still pass with this, so will give it a go.

### DIFF
--- a/plugins/destination/firehose/client/spec/schema.json
+++ b/plugins/destination/firehose/client/spec/schema.json
@@ -7,7 +7,7 @@
       "properties": {
         "stream_arn": {
           "type": "string",
-          "pattern": "^arn:[^:]+:firehose:[^:]+:[^:]+:deliverystream/[^:]+$",
+          "pattern": "^arn:[^:]+:firehose:[^:]+:[^:]+:deliverystream\\/[^:]+$",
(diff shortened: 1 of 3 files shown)


erezrokah (commented):
Let's remove this validation? I think it's too strict. The frontend will also fail if you use `${STREAM_ARN}` env variable I think

mnorbury:
> Let's remove this validation? I think it's too strict. The frontend will also fail if you use `${STREAM_ARN}` env variable I think

Okay, removed the pattern in https://github.com/cloudquery/cloudquery/pull/16955/commits/ac70a1ca33501d3807be91b34e7a573ee71f5311. 

How will `${STREAM_ARN}` fail in the frontend? The docs currently have it as `${FIREHOSE_STREAM_ARN}`.

erezrokah:
> How will `${STREAM_ARN}` fail in the frontend? The docs currently have it as `${FIREHOSE_STREAM_ARN}`.

The frontend will try to match the regex with `${FIREHOSE_STREAM_ARN}` (sorry didn't remember the exact name) and fail

PR #16967, opened 2024-02-29
fix: Clean env prefixes for cloud syncs

hermanschaaf:
This replaces dashes in plugin names with underscores, and makes the entire prefix upper case, in order to get valid, consistent environment variables.

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -302,7 +302,8 @@ func sync(cmd *cobra.Command, args []string) error {
 
 func filterPluginEnv(environ []string, pluginName, kind string) []string {
 	env := make([]string, 0)
-	prefix := "__" + kind + "_" + pluginName + "__"
+	cleanName := strings.ReplaceAll(pluginName, "-", "_")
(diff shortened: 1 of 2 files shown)


erezrokah (approved):

disq (approved):

cli/cmd/sync.go
```diff
@@ -302,7 +302,8 @@ func sync(cmd *cobra.Command, args []string) error {
 
 func filterPluginEnv(environ []string, pluginName, kind string) []string {
 	env := make([]string, 0)
-	prefix := "__" + kind + "_" + pluginName + "__"
+	cleanName := strings.ReplaceAll(pluginName, "-", "_")
+	prefix := strings.ToUpper("__" + kind + "_" + cleanName + "__")
```
erezrokah, line 306: Does this need a backend change too?

hermanschaaf: Yes, but we have the CLI version pinned in the backend so I will make an update that upgrades and makes the corresponding change there shortly

PR #16973, opened 2024-02-29
fix: Fix Primary key on `aws_applicationautoscaling_scaling_activities`

bbernays:
#### Summary

fix pk... Same resource can have multiple scaling activities


erezrokah (approved):

disq:
No point in merging this and will cause a conflict in the other repo.

PR #16985, opened 2024-03-01
fix: Adding environment variables to example

mnorbury:


plugins/destination/kafka/docs/_configuration.md
```diff
@@ -11,17 +11,19 @@ spec:
   version: "VERSION_DESTINATION_KAFKA"
   write_mode: "append"
   spec:
-    brokers: ["<broker-host>:<broker-port>"]
-    format: "csv" # options: parquet, json, csv
+    # required - list of brokers to connect to
+    brokers: ["${KAFKA_BROKER}"]
```
mnorbury, line 15: Having an environment variable will make the UI experience good for a single broker, but what if we have more than one broker? Maybe we shouldn't use an environment variable here.

erezrokah: Is the broker a secret? If not we can drop the env variable and let the user decide

erezrokah (approved):

plugins/destination/kafka/docs/_configuration.md
mnorbury: No, so I've removed it as suggested.

PR #16986, opened 2024-03-01
fix: Get dirname using path module

erezrokah:
#### Summary

Current code has a syntax error https://github.com/cloudquery/cloudquery-private/actions/runs/8109948590/job/22166179264#step:3:25

Follow up to https://github.com/cloudquery/cloudquery/pull/16891


erezrokah:
Merging as this is blocking some dependencies updates

PR #17009, opened 2024-03-04
feat: Adjustable sub-columns in `test_sub_table` for the test source plugin

disq:
Useful in debugging wide row handling in destinations.


plugins/source/test/resources/plugin/client.go
```diff
@@ -98,6 +90,15 @@ func Configure(_ context.Context, logger zerolog.Logger, spec []byte, opts plugi
 	if err := config.Validate(); err != nil {
 		return nil, fmt.Errorf("failed to validate spec: %w", err)
 	}
+
+	if opts.NoConnection {
```
hermanschaaf, line 94: why did the `NoConnection` section move? I thought we shouldn't validate config if NoConnection is given

disq: The config doesn't have any required params so it's to accomodate the default values for the table config. Test plugin is doing its own thing with documented but dynamic tables etc.

hermanschaaf (approved):

plugins/source/test/resources/plugin/client.go
erezrokah: I think this fails packaging since we pass and empty spec

PR #17013, opened 2024-03-04
feat: Use Hardcoded Regions if `EC2:DescribeRegion` is denied

bbernays:
#### Summary

Currently if CQ cannot call `EC2:DescribeRegions` then the sync will always fail. This PR enables users to sync regions that are explicitly defined (ie not defined via `*`) if `EC2:DescribeRegions` because of permissions errors


plugins/source/aws/client/account.go
```diff
@@ -100,17 +101,25 @@ func findEnabledRegions(ctx context.Context, logger zerolog.Logger, accountName
...
 		}
 		filteredRegions := filterDisabledRegions(localRegions, enabledRegions)
 		if len(filteredRegions) > 0 {
 			return filteredRegions
 		}
 	}
+	if !isAllRegions(localRegions) && len(deniedErrors) == len(regionsToCheck) && funk.Subset(regionsToCheck, deniedErrors) {
+		logger.Info().Str("account", accountName).Msg("Access denied for all regions, assuming all specified regions are enabled")
+		return localRegions
+	}
```
candiduslynx, line 122: I suggest adding a warning here add well, to indicate that no reasons are left (nit)

candiduslynx (approved):

disq (approved):

Issue #15704, opened 2023-12-14
feat: add `cloudquery validate` subcommand to validate a configuration directory


asaba-hashi:
### Which problem is this feature request solving?

I would like to check that my configuration is syntactically valid and internally consistent without invoking a full `sync` operation or making any actual API requests using real credentials.

For my use case, the following errors would be nice to catch in CI:

1. Syntactically correct yaml
1. Invalid plugin versions
1. Mismatched plugin versions
1. Missing required fields
1. `spec` should also be validated for syntactic correctness and internal consistency with it's plugin provider.
1. etc

Some of these can be checked with variations and combinations of other commands, but the ask is for:

1. A single subcommand.
2. No side-effects (such as plugin downloads or source/destination API calls).
3. List all problems found instead of failing at the first one.

### Describe the solution you'd like

Add a `cloudquery validate` subcommand that takes similar arguments to `sync`, that just does the minimal loading a checks of the configuration.  If necessary, the user should supply dummy variables values for testing.  Given the following file:

```yaml
kind: source
spec:
  name: aws
  path: cloudquery/aws
  registry: cloudquery
  version: "v19.09.2"
  spec:
    accounts:
       - id: 123456789012
         role: {ROLE_ARN}
         role_session_name: CloudQuerySession
  tables: ["aws_s3_buckets"]
  destinations: ["postgresql"]
```

`cloudquery validate .` should fail with a message listing all detectable errors (i.e. not stopping after the first one).

```
ROLE_ARN is not defined.

plugin version is invalid for aws
```

These checks would be similar to something like what [nomad job validate](https://developer.hashicorp.com/nomad/docs/commands/job/validate) is able to check.

The command should not make any changes or fetch new versions of plugins.

### Pull request (optional)

- [ ] I can submit a pull request

erezrokah:
Hi @asaba-hashi, thanks for opening the issue.

> 1. Invalid plugin versions

You should be able to achieve this by running `cloudquery plugin install` (see https://docs.cloudquery.io/docs/reference/cli/cloudquery_plugin_install). The command will fail if a plugin version is incorrect, or download all plugins if all versions are correct. Downloaded plugins are cached for the next run

> 2. Mismatched plugin versions

You should be able to achieve this by running `cloudquery migrate` (see https://docs.cloudquery.io/docs/reference/cli/cloudquery_migrate). The command verifies the versions are compatible and then migrates any relevant tables.
[...]

PR #17086, opened 2024-03-06
feat: Add `test-connection` command

erezrokah:
#### Summary

Work in progress for https://github.com/cloudquery/cloudquery/issues/15704 and also https://github.com/cloudquery/cloudquery-issues/issues/1170 (internal issue).

This command uses the built in `init` gRPC call to validate plugin initialization. This relies on plugins to do all the relevant checks during init. Some plugins do it, some don't so we'll need to add plugin's support for it.

Example for AWS->Postgres without AWS creds and wrong Postgres password:

```
Loading spec(s) from examples/aws-postgres.yml
Error: failed to init source aws (cloudquery/aws@v23.1.0): rpc error: code = Internal desc = failed to init plugin: failed to initialize client: no AWS accounts were successfully configured. See warning messages in the logs for details
failed to init destination postgresql (cloudquery/postgresql@v7.2.0): rpc error: code = Internal desc = failed to init plugin: failed to initialize client: failed to get current database: failed to connect to `host=localhost user=postgres database=postgres`: failed SASL auth (FATAL: password authentication failed for user "postgres" (SQLSTATE 28P01))
exit status 1
```


cli/cmd/validate.go
```diff
@@ -0,0 +1,135 @@
...
+
+func NewCmdValidate() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "validate [files or directories]",
+		Short:   validateShort,
+		Long:    validateShort,
+		Example: validateExample,
+		Args:    cobra.MinimumNArgs(1),
+		RunE:    validate,
+		Hidden:  true,
```
erezrokah, line 33: Hidden on purpose for now to test this a bit before making it public for everyone

cli/cmd/validate.go
```diff
@@ -0,0 +1,135 @@
...
+
+func validate(cmd *cobra.Command, args []string) error {
+	cqDir, err := cmd.Flags().GetString("cq-dir")
+	if err != nil {
+		return err
+	}
+
+	ctx := cmd.Context()
+	log.Info().Strs("args", args).Msg("Loading spec(s)")
+	fmt.Printf("Loading spec(s) from %s\n", strings.Join(args, ", "))
```
mnorbury, line 46: Should this be here as you have the log line before?

cli/cmd/validate.go
```diff
@@ -0,0 +1,135 @@
...
+		destinationRegInferred[i] = destination.RegistryInferred()
+	}
+
+	sourceClients, err := managedplugin.NewClients(ctx, managedplugin.PluginSource, sourcePluginConfigs, opts...)
+	if err != nil {
+		return enrichClientError(sourceClients, sourceRegInferred, err)
+	}
+	defer func() {
+		if err := sourceClients.Terminate(); err != nil {
+			fmt.Println(err)
```
mnorbury, line 105: Can we use the logger?

mnorbury (commented):
Looks good, left a few comments about potential left over `fmt` and using the logger for error messages.

cli/cmd/validate.go
erezrokah: So I think we need both, see https://github.com/cloudquery/cloudquery/blob/c6e06aa2fb25010e0266538200dcb7212aa17248/cli/cmd/sync.go#L99

one goes to the log, the other shows up in the console via `stdout`

cli/cmd/validate.go
erezrokah: This is similar to https://github.com/cloudquery/cloudquery/blob/c6e06aa2fb25010e0266538200dcb7212aa17248/cli/cmd/sync.go#L163, where we want to show the error in `stdout`

disq (commented):
LGTM but the title should be "feat: Add `test-connection` command"?

Validate would be the offline (no-conn) version of this.

cli/cmd/test_connection.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/cloudquery/cloudquery/cli/internal/auth"
+	"github.com/cloudquery/cloudquery/cli/internal/specs/v0"
+	"github.com/cloudquery/plugin-pb-go/managedplugin"
+	"github.com/cloudquery/plugin-pb-go/pb/plugin/v3"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+)
+
+const (
+	testConnectionShort   = "Test plugins connection to sources and destinations"
```
disq, line 17:
```suggestion
	testConnectionShort   = "Test plugin connections to sources and destinations"
```

cli/cmd/test_connection.go
```diff
@@ -0,0 +1,135 @@
...
+	testConnectionExample = `# Test plugins connection to sources and destinations
```
disq, line 18:
```suggestion
	testConnectionExample = `# Test plugin connections to sources and destinations
```

cli/cmd/test_connection.go
```diff
@@ -0,0 +1,135 @@
...
+cloudquery test-connection ./directory
+# Test plugins connection from directories and files
```
disq, line 20:
```suggestion
# Test plugin connections from directories and files
```

disq (changes requested):
Some wording changes to avoid having to use `plugin's`.

erezrokah:
> LGTM but the title should be "feat: Add `test-connection` command"?

Ah yeah the command was renamed

disq (approved):

PR #17087, opened 2024-03-06
feat: Add `invocation-id` to all log statements

bbernays:
#### Summary

This change will retain the same uuid for all source/destinations in the same sync rather than a unique sync id per source

example:
```
{"level":"info","module":"cli","log-id":"7871e9f8-c747-4f44-98e0-a32d216e503c","args":["config.yml"],"time":"2024-03-06T23:47:49Z","message":"Loading spec(s)"}
```


cli/cmd/logging.go
```diff
@@ -48,6 +48,6 @@ func initLogging(noLogFile bool, logLevel *enum.Enum, logFormat *enum.Enum, logC
 		}
 	}
 	mw := io.MultiWriter(writers...)
-	log.Logger = zerolog.New(mw).Level(zerologLevel).With().Str("module", "cli").Timestamp().Logger()
+	log.Logger = zerolog.New(mw).Level(zerologLevel).With().Str("module", "cli").Str("log-id", invocationUUID.String()).Timestamp().Logger()
```
erezrokah, line 51:
```suggestion
	log.Logger = zerolog.New(mw).Level(zerologLevel).With().Str("module", "cli").Str("trace-id", invocationUUID.String()).Timestamp().Logger()
```

Maybe a better name? Also maybe have a hidden flag/env variable to configure it incase one would like to have predictable trace IDs

erezrokah (commented):
Added a couple of comments. I agree this makes sense to have, was it requested by a user?

cli/cmd/root.go
```diff
@@ -67,6 +69,12 @@ func NewCmdRoot() *cobra.Command {
 			// PersistentPreRunE runs after argument parsing, so errors during parsing will result in printing the help
 			cmd.SilenceUsage = true
 			var err error
+
+			invocationUUID, err = uuid.NewRandom()
```
erezrokah, line 73:
```suggestion
			traceID, err = uuid.NewRandom()
```

cli/cmd/logging.go
bbernays: Unless I missed something, I don't think that this gets propagated to our traces. So maybe something like `invocation-id` would be better.

cli/cmd/root.go
```diff
@@ -67,6 +69,12 @@ func NewCmdRoot() *cobra.Command {
...
+			if err != nil {
+				return fmt.Errorf("failed to generate invocation uuid: %w", err)
```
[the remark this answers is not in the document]
bbernays: If someone is using this to monitor there syncs then I think it is better to error out rather than run something that is not able to be properly monitored

erezrokah (approved):

PR #17103, opened 2024-03-11
doc: Remove source plugin directory

hermanschaaf:
Only the BitBucket source plugin was still documented here; however the presence of the source plugin tab in mobile is confusing users. We should move the BitBucket plugin docs to the Hub instead.


erezrokah (approved):
This looks good but if someone tries to access the docs via backlinks they will 404. Maybe we can add it to the Hub and add a redirect

hermanschaaf:
@erezrokah Yeah good point, I'll wait for the plugin to be available in the Hub before merging this

Issue #8690, opened 2023-03-02
Add a Sync Metadata Table


getglad:
### Is your feature request related to a problem? Please describe.
Presently, to observe a `sync`'s operational health, I have to parse the logs. At scale across multiple workers, I end up building out a dashboard that turns that log parsing into data tables and charts.

### Describe the solution you'd like
The summary data regarding a resource tuple such as `(account, region, resource)` that is in the logs could be written to the `destination` as a `SyncMetadata` artifact.

Also recording sync start and stops would be valuable.

### Describe alternatives you've considered
Presently I parse the logs for an AWS sync as below.

`parse @message "table sync finished client=*:* errors=* module=* resources=* table=*" as Account, Region,  Errors, tableModule, Total, tableName`

It works as a reasonable basis to produce a data table that looks like the below.

| timestamp | Account | Region | tableName | Total | Errors |
|---|---|---|---|---|---|
| v1 | v2 | v3 | v4 | v5 | v6 |

### Additional context
I've built some data tables around the errors such as

`parse @message "error=\"operation error *: *, *\" client=*:* module=* table=*" as Service, Action, errorMsg, Account, tableClientRegion, tableModule, tableName`

And for sync-level errors that cause the process to entirely bail out
`parse @message "Error: *" as errorMsg`

But IMO error lines in the `destination` are less valuable than the summary output - knowing that errors have occurred and that the logs should be examined.

michalz-rely:
I think the issue https://github.com/cloudquery/cloudquery/issues/8146 is related

erezrokah:
Hi 👋  For those watching this issue, we're working on a guide for parsing and observing CloudQuery logs using Datadog. You can see it here https://cloudquery-r4i7sxw4c-cloudquery.vercel.app/how-to-guides/datadog-observability (PR https://github.com/cloudquery/cloudquery/pull/10285), and we would love your feedback.

I also opened https://github.com/cloudquery/cloudquery/issues/10266 as a result, and we might add more fields to the logs base on use cases so they can be easily queries.

Please let us know if this could work for you to solve the same use case a sync metadata table solves. I would also point out that a sync metadata table might not work when CloudQuery fails to access the destination, or have any unexpected error during the sync.

erezrokah:
Sorry for the late follow up, the official guide is here https://www.cloudquery.io/how-to-guides/datadog-observability

erezrokah:
[...]

PR #17112, opened 2024-03-11
feat: Add local sync summary

bbernays:
#### Summary

This pr adds in a sync summary that gets outputted to a user defined location. If the user doesn't define the location via the cli flag then the summary will not be persisted.

The file is a .jsonl format


erezrokah (approved):
Code looks good, I believe this closes https://github.com/cloudquery/cloudquery/issues/8690

disq (approved):

yevgenypats (changes requested):
let's schedule another meeting on that to discuss the use case

erezrokah (approved):

yevgenypats (changes requested):
sync summary should go into cli first with json output at least as first version, I suggest closing this for now

cli/cmd/sync_test.go
```diff
@@ -165,3 +329,21 @@ func TestSync_IsolatedPluginEnvironmentsInCloud(t *testing.T) {
 		})
 	}
 }
+
+func readSummaries(t *testing.T, filename string) []syncSummary {
+	p, err := os.ReadFile(filename)
+	assert.NoError(t, err)
+
+	lines := bytes.Split(p, []byte{'\n'})
```
disq, line 337: Does this test work OK on windows too?

bbernays: What is the behavior you would rather see?

disq: platform dependent line separators and a provision in this test to replace CRLF with LFs before proceeding :)

cli/cmd/migrate_v3.go
```diff
@@ -98,6 +98,7 @@ func migrateConnectionV3(ctx context.Context, sourceClient *managedplugin.Client
 				return handleSendError(err, writeClients[i], "migrate")
 			}
 		}
+
```
erezrokah, line 101:
```suggestion
```

cli/cmd/summary.go
```diff
@@ -0,0 +1,74 @@
...
+	_, err := os.Stat(dirPath)
+	if err != nil {
+		// if error is that the directory does not exist, create it
+		if os.IsNotExist(err) {
+			err = os.MkdirAll(dirPath, 0755)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
erezrokah, line 72:
```suggestion
	err := os.MkdirAll(dirPath, 0755)
	if err != nil {
		return err
	}
```

Can't we do this instead to simplify? Looks like `os.MkdirAll` handles the checks for us:

```go
func MkdirAll(path string, perm FileMode) error {
	// Fast path: if we can tell whether path is a directory or file, stop with success or error.
	dir, err := Stat(path)
	if err == nil {
		if dir.IsDir() {
			return nil
		}
		return &PathError{Op: "mkdir", Path: path, Err: syscall.ENOTDIR}
	}
```

cli/cmd/sync.go
```diff
@@ -35,6 +35,7 @@ func NewCmdSync() *cobra.Command {
 	}
 	cmd.Flags().Bool("no-migrate", false, "Disable auto-migration before sync. By default, sync runs a migration before syncing resources.")
 	cmd.Flags().String("license", "", "set offline license file")
+	cmd.Flags().String("sync-summary-location", "", "Sync summary file location")
```
erezrokah, line 38:
```suggestion
	cmd.Flags().String("summary-location", "", "Sync summary file location")
```

Since we're under the `sync` command I think we can drop the `sync` prefix, WDYT?
Also can we add an example to the `syncExample` `const`?

cli/cmd/summary.go
```diff
@@ -0,0 +1,74 @@
...
+func persistSummary(filename string, summaries []syncSummary) error {
+	// if filename is not specified then we don't need to persist the summary and can return
+	if filename == "" {
+		return nil
+	}
+	err := checkFilePath(filename)
+	if err != nil {
+		return fmt.Errorf("failed to validate summary file path: %w", err)
+	}
+	for _, summary := range summaries {
```
erezrokah, line 35: What's the reason to use `jsonl` instead of JSON and dump the whole array instead? Are we expecting consumers to read the output line by line? It's dumped in a single operation anyway and not streamed

erezrokah (changes requested):
Looks good, had a few questions/comments

cli/cmd/summary.go
bbernays: Appending to JSONL is much simpler because we don't have to deal with competing syncs writing to the same location at the same time or subsequent syncs overwriting the file.

Also because each source <> destination is an entry this makes it much easier to see what succeeded in the case of partial successes

cli/cmd/sync_v3.go
```diff
@@ -361,11 +362,30 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 		}
 	}
 
-	totals := sourceClient.Metrics()
+	syncSummaries := make([]syncSummary, len(destinationsClients))
 	for i := range destinationsClients {
 		m := destinationsClients[i].Metrics()
 		totals.Warnings += m.Warnings
 		totals.Errors += m.Errors
+		syncSummaries[i] = syncSummary{
```
erezrokah, line 370: Another question, what's the reason we're flattening the data? We could have a top level array of sources, then in each source point to an array of destinations with the summary

cli/cmd/summary.go
erezrokah: > because we don't have to deal with competing syncs writing to the same location at the same time or subsequent syncs overwriting the file

Can you explain what `competing syncs` mean? Do you mean multiple processes of the CLI? Not sure how concurrent writes will work in that case

> subsequent syncs overwriting the file

Don't we want to overwrite the file? Maybe the question to ask it how are we expecting someone to consume the file? If it's always being appended to won't consumers need to handle ignoring information they've already read?

erezrokah: > Also because each source <> destination is an entry this makes it much easier to see what succeeded in the case of partial successes

What does partial success mean? Don't we exist early on errors? We could still write the data for the sources that were successful it that's what you mean by partial success, and maybe add additional information on the ones that didn't ran. Otherwise you'd need to compare the spec to the JSON to know which ones were successful

cli/cmd/sync_v3.go
erezrokah: We've decided to release this as preview and keep the data flat for now until further feedback

cli/cmd/summary.go
erezrokah: We've decided to release this as preview and keep the implementation as is for now until further feedback

erezrokah (approved):

PR #17117, opened 2024-03-11
feat: Include Object Paths in S3 sync summary

bbernays:
This PR adds in a column that will include the full keys of all objects synced in the sync

`objects_synced` contains a `map[string][]string` where the map key is the table name and the array of strings are the object keys

```
{
  "_cq_source_name": "aws",
  "_cq_sync_time": "2024-03-11 17:11:16.369293",
  "cli_version": "development",
  "destination_name": "s3",
  "destination_path": "localhost:7778",
  "destination_version": null,
  "errors": 9,
  "objects_synced": {
    "aws_s3_buckets": [
      "aws_s3_buckets/8c862045-ed27-43f4-a70d-2b5f955e0748.parquet"
    ]
  },
  "resources": 29,
  "source_name": "aws",
  "source_path": "cloudquery/aws",
  "source_version": "v25.2.0",
  "sync_id": "a61ca899-1210-497b-ba1d-f442c26792a3",
  "warnings": 1
}
```

This is an extension of https://github.com/cloudquery/cloudquery/pull/17112

### DIFF
--- a/plugins/destination/s3/client/spec/spec.go
+++ b/plugins/destination/s3/client/spec/spec.go
@@ -181,10 +181,21 @@ func (s *Spec) Validate() error {
 
 func (s *Spec) ReplacePathVariables(table string, fileIdentifier string, t time.Time) string {
 	name := strings.ReplaceAll(s.Path, varTable, table)
+
+	if table == "cloudquery_sync_summary" {
+		name = strings.ReplaceAll(name, varFormat, "json")
+		fullSuffix := string(s.Format) + s.Compression.Extension()
+		if strings.HasSuffix(s.Path, fullSuffix) {
+			name = strings.TrimSuffix(name, fullSuffix)
+			name += "json"
+		}
+	}
+
 	if strings.Contains(name, varFormat) {
 		e := string(s.Format) + s.Compression.Extension()
 		name = strings.ReplaceAll(name, varFormat, e)
 	}
+
 	name = strings.ReplaceAll(name, varUUID, fileIdentifier)
 	name = strings.ReplaceAll(name, varYear, t.Format("2006"))
 	name = strings.ReplaceAll(name, varMonth, t.Format("01"))
--- a/plugins/destination/s3/client/write.go
+++ b/plugins/destination/s3/client/write.go
@@ -24,13 +24,26 @@ func (c *Client) WriteTable(ctx context.Context, msgs <-chan *message.WriteInser
 	var s *filetypes.Stream
 
 	for msg := range msgs {
-		if s == nil {
-			table := msg.GetTable()
+		fileClient := c.Client
+		table := msg.GetTable()
 
+		if s == nil {
+			if _, ok := c.objectKeys[table.Name]; !ok {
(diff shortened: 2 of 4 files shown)


erezrokah (approved):

plugins/destination/s3/client/spec/spec.go
```diff
@@ -181,10 +181,21 @@ func (s *Spec) Validate() error {
 
 func (s *Spec) ReplacePathVariables(table string, fileIdentifier string, t time.Time) string {
 	name := strings.ReplaceAll(s.Path, varTable, table)
+
+	if table == "cloudquery_sync_summary" {
```
erezrokah, line 185: Shouldn't this be under the option to enable the summary?

plugins/destination/s3/client/write.go
```diff
@@ -24,13 +24,26 @@ func (c *Client) WriteTable(ctx context.Context, msgs <-chan *message.WriteInser
...
+		table := msg.GetTable()
 
+		if s == nil {
+			if _, ok := c.objectKeys[table.Name]; !ok {
+				c.objectKeys[table.Name] = []string{}
+			}
 			objKey := c.spec.ReplacePathVariables(table.Name, uuid.NewString(), time.Now().UTC())
+			c.objectKeys[table.Name] = append(c.objectKeys[table.Name], objKey)
+
+			if table.Name == "cloudquery_sync_summary" {
```
erezrokah, line 37: This seems too tightly coupled with the CLI that uses the same name for the table

bbernays: It is designed to be an extension of the CLI summary table, all this functionality is doing is adding a column to that table.

yevgenypats (changes requested):
Few questions:
1) do we already have sync summary message? is it new
2) will objectkeys will run out of memory and it is not bound by anything?

depends on 1 and 2 we might want to implement sync summary only on the CLI in that case and then only add file that sais this sync finished and have a path of {{sync_id}} in the config instead of sending summary to every destination (unless we already do that?)

bbernays:
> Few questions:
> 
> 1. do we already have sync summary message? is it new
It will be added in  https://github.com/cloudquery/cloudquery/pull/17112
> 2. will objectkeys will run out of memory and it is not bound by anything?
I have added bounding that will limit it to only log the first 10,000 keys
> 
> depends on 1 and 2 we might want to implement sync summary only on the CLI in that case and then only add file that sais this sync finished and have a path of {{sync_id}} in the config instead of sending summary to every destination (unless we already do that?)

Adding {{sync_id}} only partially fixes the problem... S3 paths are used for at least 2 distinct things:
1. A means to search for objects: you can search for a path prefix to find all objects with said prefix
2. A means to partition data in S3/Athena for time based data. Adding the UUID into that means that it would be much harder to do because the UUID must be included in the definition

With the above solution assuming less than 10,000 objects in a single sync we get best of both worlds

erezrokah:
Going to close this as stale. Please re-open if needed

Issue #17204, opened 2024-03-14
bug: ClickHouse destination uses `DateTime64(3)` for `date64` mapping


candiduslynx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I also think that we should update the type mapping for `date64` in the ClickHouse plugin.

Currently `DateTime64(3)` (aka ms precision) is used:
https://github.com/cloudquery/cloudquery/blob/aba7a68ef4121240c9dec89b0652e085070f4a85/plugins/destination/clickhouse/typeconv/ch/types/column_type.go#L51-L53

ClickHouse will use `DateTime` instead: https://clickhouse.com/docs/en/sql-reference/formats#data-format-arrow.

I think the behavior between plugin & ClickHouse should be consistent: in Apache Arrow `date64` is still a date, so behavior when the `date64` isn't properly matching the date boundary is UB.

### Expected Behavior

`date64` is mapped to `DateTime` ClickHouse type

### CloudQuery (redacted) config

N/A

### CloudQuery (redacted) logs

N/A

### CloudQuery version

latest

### Pull request (optional)

- [X] I can submit a pull request

erezrokah:
What's the impact of this bug? Is it data corruption? Can you give an example?

candiduslynx:
I don't think there is any corruption, it's just an inconsistency between how ClickHouse treats the Apache Arrow data written directly & our conversions.

PR #17208, opened 2024-03-15
fix!: Use documented Apache Arrow type conversion

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery/issues/17204
BEGIN_COMMIT_OVERRIDE
fix: Use documented Apache Arrow type conversion (#17208)
BREAKING-CHANGE: Use documented Apache Arrow type conversion (#17208). The following type conversions were made to match the ClickHouse [type convention](https://clickhouse.com/docs/en/sql-reference/formats#data-format-arrow):
* `date64` is mapped to [`DateTime`](https://clickhouse.com/docs/en/sql-reference/data-types/datetime)
* `time32` & `time64` are mapped to [`DateTime64`](https://clickhouse.com/docs/en/sql-reference/data-types/datetime64) with proper precision

The only official CloudQuery source plugin that uses any of `time32`, `time64`, `date32` or `date64` types is [ServiceNow](https://hub.cloudquery.io/plugins/source/cloudquery/servicenow/latest/tables) with `date64` type. For all other source plugins we don't expect the change to be actually breaking.
However, due to actually changing the resulting column types, we deem it necessary to outline this change with a major version bump.

END_COMMIT_OVERRIDE


disq:
> No official CloudQuery source plugin currently produces the values of

https://github.com/cloudquery/cloudquery-private/blob/main/plugins/source/servicenow/resources/plugin/table_schema.go#L96

```go
	case "glide_date":
		return arrow.FixedWidthTypes.Date64
```

candiduslynx:
> > No official CloudQuery source plugin currently produces the values of
> 
> https://github.com/cloudquery/cloudquery-private/blob/main/plugins/source/servicenow/resources/plugin/table_schema.go#L96
> 
> ```go
> 	case "glide_date":
> 		return arrow.FixedWidthTypes.Date64
> ```

updated the desc

erezrokah (approved):
Looks good, I think it would have made sense to split the refactoring bits to make the PR easier to review, but not a blocker

PR #17235, opened 2024-03-18
feat: Pass the given spec to the `tables` handler of the plugin

disq:
This would be ignored by most source plugins (given there's a `NoConnection` flag on the call) but still gives the option to inquire tables remotely to the plugin.


cli/cmd/tables_v3.go
```diff
@@ -14,13 +14,13 @@ import (
...
 
-func tablesV3(ctx context.Context, sourceClient *managedplugin.Client, path string, format string) error {
+func tablesV3(ctx context.Context, sourceClient *managedplugin.Client, sourceSpec any, path string, format string) error {
 	err := registerOnce()
 	if err != nil {
 		return err
 	}
 	sourcePbClient := pluginPb.NewPluginClient(sourceClient.Conn)
-	if err := initPlugin(ctx, sourcePbClient, nil, true); err != nil {
+	if err := initPlugin(ctx, sourcePbClient, sourceSpec, true); err != nil {
```
disq, line 23: Here we could check if sourceSpec is an empty object and pass `nil` if it is, simplifying the `if opts.NoConnection && (len(specBytes) == 0 || bytes.Equal(specBytes, []byte("null")) || bytes.Equal(specBytes, []byte("{}"))) {` check in the plugin.

disq: Done in https://github.com/cloudquery/cloudquery/pull/17235/commits/1d4199ab03e601f0941d57cd13655bf60b904fc7
Normalizing to `null`  was the only backwards-compatible option, if I had used `{}` then the if statement doesn't change in order to support older cli versions.

erezrokah: > if I had used `{}` then the if statement doesn't change in order to support older cli versions.

Which if statement are you referring to?

disq: > > if I had used `{}` then the if statement doesn't change in order to support older cli versions.
> 
> Which if statement are you referring to?

In the ServiceNow plugin, quoted on the first comment in this thread ^^^

erezrokah: Ah got it, old CLI versions pass `nil` which serializes to `[]byte("null")`?

erezrokah (approved):

cli/cmd/tables_v3.go
disq: Yes

PR #17238, opened 2024-03-19
feat: Add JSON schema

erezrokah:
#### Summary

The sync from the open source repo to the private one is failing https://github.com/cloudquery/cloudquery-private/actions/runs/8335271827/job/22810448075#step:4:1241

since https://github.com/cloudquery/cloudquery-private/pull/1090 and https://github.com/cloudquery/cloudquery-private/pull/1089 (adding JSON schema to AWS Pricing and Hacker News) should have been done in the open source repository as those are public and free plugins.

Because ⬆️ were done in the private source repo each time their SDK is updated in the open source repo we get a conflict.

Also the JSON schema feature was never released for those plugins as they are released from this repo:
https://github.com/cloudquery/cloudquery/pull/17194
https://github.com/cloudquery/cloudquery/pull/17199


plugins/source/hackernews/client/spec.go
```diff
@@ -1,13 +1,16 @@
...
 	"time"
 )
 
 type Spec struct {
-	ItemConcurrency int    `json:"item_concurrency"`
-	StartTime       string `json:"start_time"`
+	// The number of items to fetch concurrently
+	ItemConcurrency int `json:"item_concurrency" jsonschema:"minimum=1,default=100"`
+	// RFC3339 formatted timestamp. Syncing will begin with posts after this date. If empty everything is synced.
+	StartTime string `json:"start_time" jsonschema:"format=date-time"`
```
candiduslynx, line 13: you also need to check that `"start_time": ""` is allowed

candiduslynx (approved):

plugins/source/hackernews/client/spec.go
candiduslynx: removed `automerge` bc of this

erezrokah: Looks like the code gen doesn't support `oneof_type` with additional properties https://github.com/invopop/jsonschema/issues/113.
Regardless I believe the description is wrong, we should allow either a valid date-time or omit the property per  https://github.com/cloudquery/cloudquery/blob/5ecf6fc92eebfc0a17e129b7ba65c4e5d28265bb/plugins/source/hackernews/docs/overview.md?plain=1#L38

I updated the description to reflect this

PR #17242, opened 2024-03-19
feat: Support `private_key` env variable

erezrokah:
#### Summary

Allows referencing the private key content directly instead of a path to the key.
Part of https://github.com/cloudquery/cloudquery-issues/issues/1361 (internal issue).

This should allow running the Oracle source plugin in environments when one can't populate the file system with the private key

### DIFF
--- a/plugins/source/oracle/client/configProvider.go
+++ b/plugins/source/oracle/client/configProvider.go
@@ -20,7 +20,7 @@ type rawPrivateKeyConfigProvider struct {
 }
 
 func (rawPrivateKeyConfigProvider) PrivateRSAKey() (*rsa.PrivateKey, error) {
-	const(
+	const (
 		tfVarEnvironmentVariable = tfVarEnvironmentVariable + "private_key"
 		ocCLIEnvironmentVariable = ocCLIEnvironmentVariable + "private_key"
 	)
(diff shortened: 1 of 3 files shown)


plugins/source/oracle/client/configProvider.go
```diff
@@ -0,0 +1,58 @@
...
+	ocCLIEnvironmentVariable = "OCI_CLI"
+)
+
+type rawPrivateKeyConfigProvider struct {
+	common.ConfigurationProvider
+}
+
+func (rawPrivateKeyConfigProvider) PrivateRSAKey() (*rsa.PrivateKey, error) {
+	tfVarEnvironmentVariable := fmt.Sprintf("%s_%s", tfVarEnvironmentVariable, "private_key")
+	ocCLIEnvironmentVariable := fmt.Sprintf("%s_%s", ocCLIEnvironmentVariable, "private_key")
```
candiduslynx, line 24:
```suggestion
	const(
		tfVarEnvironmentVariable = tfVarEnvironmentVariable + "private_key"
		ocCLIEnvironmentVariable = ocCLIEnvironmentVariable + "private_key"
	)
```
use const

candiduslynx (commented):
I only see consts as an improvement path, that's it

candiduslynx (approved):

PR #17294, opened 2024-03-21
doc: Document Apache Arrow types mapping

candiduslynx:


erezrokah:
Hi @candiduslynx can you explain what this PR fixes? Maybe with some before/after examples

candiduslynx:
@erezrokah 
> Hi @candiduslynx can you explain what this PR fixes? Maybe with some before/after examples

~1. `Null` type now maps to `Nothing` instead of `String`~
I've retested this & actually removed this as we can't create `Nothing` column

2. Decimal type conversion wont upscale the precision to at least 19 now
3. Doc the conversion

plugins/destination/clickhouse/typeconv/ch/types/decimal.go
```diff
@@ -5,41 +5,18 @@ import (
 	"strconv"
 
 	"github.com/apache/arrow/go/v15/arrow"
-	"golang.org/x/exp/constraints"
 )
 
 // https://clickhouse.com/docs/en/sql-reference/data-types/decimal
 func decimalType(decimalType arrow.DecimalType) (string, error) {
 	precision, scale := decimalType.GetPrecision(), decimalType.GetScale()
-	if scale > 76 {
```
erezrokah, line 14: Should this be labeled a `fix`? Also is it a breaking change? And what does it fix?

candiduslynx: I think we don't have an issue here, as the arrow documents decimal types to have max precision as 76: https://arrow.apache.org/docs/r/reference/data-type.html
So I'd say this change is just removing an obsolete check

erezrokah (approved):

Issue #17319, opened 2024-03-25
chore: CLI CI tests are flaky 


erezrokah:
See https://github.com/cloudquery/cloudquery/actions/runs/8413322363/job/23035387099?pr=17241 (also there's a race condition there)

PR #17320, opened 2024-03-25
chore: Isolate cmd tests cq dir and log file

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/17319

Some tests were sharing the default CQ dir `.cq` and the same log file, also no need to do `os.RemoveAll` on `t.TempDir()` as the test cleans it up automatically, but we do need to register the `CloseLogFile()` cleanup after `t.TempDir()` due to https://pkg.go.dev/testing#B.Cleanup:
> Cleanup functions will be called in last added, first called order.


cli/cmd/switch_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !windows
```
erezrokah, line 1: Took a bit of trial an error of skipping each test individually 
image

but this test fails consistently on WIndows, not sure way as the test process crashes with an error message https://github.com/cloudquery/cloudquery/actions/runs/8471308709/job/23210981327#step:6:66
```
Your current team is not set.

Teams available to you: my-team

To switch teams, run `cloudquery switch <team>`
2024/03/28 17:35:49 [DEBUG] GET http://127.0.0.1:53317/teams?page=1&per_page=100
Successfully switched teams to my-team.
2024/03/28 17:35:49 [DEBUG] GET http://127.0.0.1:53317/teams?page=1&per_page=100
FAIL	github.com/cloudquery/cloudquery/cli/cmd	[58](https://github.com/cloudquery/cloudquery/actions/runs/8471308709/job/23210981327#step:6:59).642s
```

Decided it's best to disable it for now (we also disable Windows test for `infer_test.go` which is another test that reads the STDOUT for the CLI. Not sure it's related, or maybe related to the XDG lib and something failing during its `Reload`

cli/cmd/addon_download_test.go
```diff
@@ -16,6 +17,11 @@ import (
 )
 
 func TestAddonDownload(t *testing.T) {
+	cqDir := t.TempDir()
+	logFileName := path.Join(cqDir, "cloudquery.log")
+	t.Cleanup(func() {
+		CloseLogFile()
+	})
```
bbernays, line 24: This seems very formulaic and used in nearly all of the tests, do we want to have a single helper function that just returns a list that can be appended to the args array

```
func setupArgs(t *testing.T) []string{
	cqDir := t.TempDir()
	logFileName := path.Join(cqDir, "cloudquery.log")
	t.Cleanup(func() {
		CloseLogFile()
	})
	return []string{"--cq-dir", cqDir, "--log-file-name", logFileName}
 }

Comment (erezrokah) [reply cli/cmd/addon_download_test.go:24]: Yes we can 👍 Let me follow up with a refactor commit

**Review by bbernays — APPROVED** [review]

File: cli/cmd/sync_test.go [hunk 3]
```diff
@@ -58,29 +58,6 @@ func TestSync(t *testing.T) {
 		{
 			name:   "multiple_destinations",
 			config: "multiple-destinations.yml",
-			summary: []syncSummary{
```
erezrokah, line 61: I believe this worked before since all tests were sharing the `.cq` directory which means they also share `summaryPath`.
If you check `multiple-destinations.yml` you'll see the source version is different than listed here

erezrokah: Also had to do https://github.com/cloudquery/cloudquery/pull/17320/commits/206bbaf7190d5a7aeda3ddc59b545716d741b7db to fix other sync summary tests due to the re-use of the output path

PR #17329, opened 2024-03-25
fix: Update docker-offline to reflect login requirement

bbernays:
#### Summary

Update docker-offline to reflect login requirement


website/pages/docs/deployment/docker-offline.mdx
```diff
@@ -35,13 +38,14 @@ FROM ghcr.io/cloudquery/cloudquery:latest
 WORKDIR /app
 COPY ./build.spec.yaml /app/build.spec.yaml
 
+ARG CLOUDQUERY_API_KEY
```
pilvikala, line 41: I think it's better to pass the API key as an environment variable rather than building it into the container image. At first, the api key gets stored in the container image and secondly, you'd need to rebuild the container every time you change the API key.

bbernays: This tutorial is for `docker offline` usage. People that follow this tutorial wouldn't expect there to be any network calls after the build phase

erezrokah: So maybe we should split it into 2 phases, one downloading the plugins, then second running them, similar to https://github.com/cloudquery/plugin-sdk-javascript/blob/00afbb59b11a6783a340dfa57d66f6fe12504a13/Dockerfile#L17

bbernays: We already have https://docs.cloudquery.io/docs/deployment/docker which shows how to run a sync via a docker container (which also should be updated to include an CQ Api Key but will do that in a different pr).

What is the use case that this tutorial should be helping users solve?

erezrokah: I think we want to show people how to package plugins with the container so they don't have to download them later.
If we can do it without embedding the API key in the final container that would be best

bbernays: As far as I am aware, `ARG`s are only available during the `docker build` stage. If this is not true I am happy to rework it

https://arc.net/l/quote/ybsollqe

bbernays: Updated it

erezrokah (approved):

PR #17331, opened 2024-03-26
fix: Ensure login callback is only handled once

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/1454 (internal issue).
It seems that a recent change in Brave browser causes the login callback to be called multiple times, causing us to panic when we close an already closed channel

### DIFF
--- a/cli/cmd/login.go
+++ b/cli/cmd/login.go
@@ -11,7 +11,7 @@ import (
 	"os/signal"
 	"strconv"
 	"strings"
-	goSync "sync"
+	gosync "sync"
 	"syscall"
 	"time"
 
@@ -98,7 +98,7 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 	mux := http.NewServeMux()
 	refreshToken := ""
 	gotToken := make(chan struct{})
-	var callbackLock goSync.Mutex
+	var callbackLock gosync.Mutex
 	callbackHandled := false
 	mux.HandleFunc("/callback", func(w http.ResponseWriter, r *http.Request) {
 		callbackLock.Lock()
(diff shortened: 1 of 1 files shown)


cli/cmd/login.go
```diff
@@ -11,6 +11,7 @@ import (
 	"os/signal"
 	"strconv"
 	"strings"
+	gosync "sync"
```
disq, line 14: `gosync`, why? is there another package or variable with this name, or just editor mishap?

cli/cmd/login.go
```diff
@@ -97,10 +98,19 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 	mux := http.NewServeMux()
 	refreshToken := ""
 	gotToken := make(chan struct{})
+	var callbackLock gosync.Mutex
+	callbackHandled := false
 	mux.HandleFunc("/callback", func(w http.ResponseWriter, r *http.Request) {
+		callbackLock.Lock()
```
disq, line 104: Should use `sync.Once` which is for this type of thing.

cli/cmd/login.go
erezrokah: There's another sync package

cli/cmd/login.go
erezrokah: Also we still want to do the redirect to `success-close`

erezrokah: Ok so https://github.com/cloudquery/cloudquery/pull/17331/commits/f1b1807890f84c0dfd5d868ffa03575c7ccf8527 looks better indeed

disq (approved):

Issue #17335, opened 2024-03-26
bug: PostgreSQL destination saving state sometimes fails


disq:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

 failed to sync unmanaged client: failed to save state: rpc error: code = Internal desc = write failed: failed to execute batch with pgerror: severity: ERROR, code: 42601, message: zero-length delimited identifier at or near """", detail :, hint: , position: 91, internal_position: 0, internal_query: , where: , schema_name: , table_name: , column_name: , data_type_name: , constraint_name: , file: scan.l, line: 1188, routine: scanner_yyerror: ERROR: zero-length delimited identifier at or near """" (SQLSTATE 42601)

### Expected Behavior

No error

### CloudQuery (redacted) config

-

### CloudQuery (redacted) logs

-

### CloudQuery version

postgresql destination v7.6.0

### Pull request (optional)

- [X] I can submit a pull request

hermanschaaf:
Probably related: https://github.com/cloudquery/cloudquery/issues/13585

PR #17337, opened 2024-03-26
fix: Make sure to scan table list during state backend operations

disq:
Fixes https://github.com/cloudquery/cloudquery/issues/17335
Fixes https://github.com/cloudquery/cloudquery/issues/13585

The issue was a mixed `len` vs. `nil` check and `listTables` creating the struct before filling it. Should probably add a lock.


plugins/destination/postgresql/client/list_tables.go
```diff
@@ -127,5 +127,7 @@ func (c *Client) listTables(ctx context.Context) (schema.Tables, error) {
 			Unique: isUnique && uniqueName == tableName+"_"+columnName+"_key",
 		})
 	}
+
+	c.pgTablesToPKConstraints = tableInfo
```
hermanschaaf, line 131: Yeah let's add a lock here

disq: Added in https://github.com/cloudquery/cloudquery/pull/17337/commits/358cd4f8e7f5e74b9209b31d5b9a8be432bc6a74

hermanschaaf (approved):

plugins/destination/postgresql/client/insert.go
```diff
@@ -13,15 +13,32 @@ import (
...
 )
 
-// InsertBatch inserts records into the destination table. It forms part of the writer.MixedBatchWriter interface.
-func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts) error {
+func (c *Client) pgTables(ctx context.Context) (map[string]struct{}, error) {
+	c.pgTablesToPKConstraintsMu.RLock()
 	// This happens when the CLI was invoked with `sync --no-migrate`
-	if c.pgTablesToPKConstraints == nil {
+	if len(c.pgTablesToPKConstraints) == 0 {
+		c.pgTablesToPKConstraintsMu.RUnlock()
```
bbernays, line 20: Why are we unlocking then locking again?

disq: because listTables requires an unlocked mutex

plugins/destination/postgresql/client/migrate.go
```diff
@@ -298,10 +298,15 @@ func (c *Client) migrateToCQID(ctx context.Context, table *schema.Table, _ schem
 			return fmt.Errorf("failed to begin transaction: %w", err)
 		}
 	}
+
+	c.pgTablesToPKConstraintsMu.Lock()
 	entry := c.pgTablesToPKConstraints[tableName]
 	if entry == nil {
 		entry = new(pkConstraintDetails)
+		c.pgTablesToPKConstraints[tableName] = entry
```
disq, line 306: Extra line unrelated to this fix.

PR #17343, opened 2024-03-26
feat: Fail on Init with incorrect credentials

bbernays:
#### Summary

 Fail on Init with incorrect credentials


plugins/destination/bigquery/client/client.go
```diff
@@ -89,3 +95,17 @@ func (c *Client) Close(ctx context.Context) error {
 	}
 	return c.client.Close()
 }
+
+func (c *Client) validateCreds(ctx context.Context) error {
+	tableRef := c.client.Dataset(c.spec.DatasetID).Table("cq-test-table")
+	_, err := tableRef.Metadata(ctx)
+	if err != nil {
+		if e, ok := err.(*googleapi.Error); ok {
+			// if the error is 404, the table does not exist, but the credentials are valid
```
erezrokah, line 104: Could this be a result of a permissions error? e.g. the permissions allow syncing a specific dataset but not accessing another resulting in 404?

bbernays: Good point, but there is nothing else that I can see that we can use to validate the credentials... We have no idea what tables are going to be synced. Maybe we can document that we need access to this table?

bbernays: I have updated this to fetch the dataset rather than a table... In our schema we require a dataset to be created before sync or migration

plugins/destination/bigquery/client/client.go
```diff
@@ -89,3 +95,18 @@ func (c *Client) Close(ctx context.Context) error {
 	}
 	return c.client.Close()
 }
+
+func (c *Client) validateCreds(ctx context.Context) error {
+	datasetRef := c.client.Dataset(c.spec.DatasetID)
+	_, err := datasetRef.Metadata(ctx)
+	if err != nil {
+		if e, ok := err.(*googleapi.Error); ok {
+			// if the error is 404, the table does not exist, but the credentials are valid
```
erezrokah, line 104:
```suggestion
```

erezrokah (approved):

hermanschaaf (approved):

Issue #17041, opened 2024-03-06
bug(github): Secondary rate limits encountered for SBOM table


hermanschaaf:
With similar config to https://github.com/cloudquery/cloudquery/issues/17040, user encountered this error:

```
2024-03-06T09:07:19Z WRN GitHub secondary rate limit detected for API call: https://api.github.com/repos/xxxx/ops-deployments/dependency-graph/sbom. Sleeping until 2024-03-06T15:00:57+05:30 module=github-src
```

First, that timestamp should be in UTC, so it's easier to compare. Second, it seems to be sleeping for a really long time? Third, we should try and adapt the SBOM table logic so that we don't encounter secondary rate limits.

hermanschaaf:
This also happens for the GitHub workflows table (see https://discord.com/channels/872925471417962546/873606591335759872/1214858764411469824)

erezrokah:
Relevant code in https://github.com/cloudquery/cloudquery/blob/31575d22d0000362ac51693f6d958a7fb148b520/plugins/source/github/client/client.go#L234

I think the sleep until is calculated from the GitHub response headers https://github.com/gofri/go-github-ratelimit/blob/9d9726540b07a6a728cc6beb4a28cdc8a3a8f91a/github_ratelimit/ratelimit.go#L169

erezrokah:
So I think a quick win would be to change the default for `skip_archived_repos` from `false` to `true` under a breaking change (major version bump)

erezrokah:
I tried using the approach from https://github.com/google/go-github?tab=readme-ov-file#conditional-requests to cache requests. I only got cache hits when using a persistent cache (e.g. disk cache) and between different sync runs, so using a cache doesn't help much for a single sync.

erezrokah:
Opened https://github.com/google/go-github/issues/3114 for reference

erezrokah:
https://github.com/google/go-github/pull/3117 should help

erezrokah:
Going to close this in favor of the more actionable https://github.com/cloudquery/cloudquery/issues/17475 and https://github.com/google/go-github/issues/3114

PR #17347, opened 2024-03-26
feat!: Auto detect concurrency setting from rate limit

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery/issues/17041

This is a breaking change since it changes the default concurrency from `10000` (GitHub default rate limit is 5000 requests per hour).
We can use an endpoint to detect the rate limit and use that as a starting value for the concurrency.


erezrokah:
This doesn't really help, I found that setting concurrency to a default of 1000 seems to work best to avoid the secondary rate limit. For the primary limit, we can't really do anything about it since it works like a quota bucket and we use the quota on each API request, regardless how fast/slow we make those

PR #17368, opened 2024-03-27
feat: Add support for conditional requests via local cache

erezrokah:
#### Summary

Implements https://github.com/google/go-github?tab=readme-ov-file#conditional-requests

Tested it and indeed saw some requests returned from the cache and not counting towards the limit


erezrokah:
Downside is that we use https://github.com/gregjones/httpcache that is archived but couldn't find a better alternative

plugins/source/github/client/client.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/gofri/go-github-ratelimit/github_ratelimit"
 	"github.com/google/go-github/v59/github"
+	"github.com/gregjones/httpcache"
+	"github.com/gregjones/httpcache/diskcache"
```
bbernays, line 17: Maybe we want to fork the repo if we are going to use it to ensure security vulnerabilities are flagged?

erezrokah: We could, let me do a follow up commit

plugins/source/github/client/client.go
```diff
@@ -91,10 +93,15 @@ func New(ctx context.Context, logger zerolog.Logger, spec Spec) (schema.ClientMe
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse InstallationID %v: %w", auth.InstallationID, err)
 		}
+		transport := http.DefaultTransport
+		if spec.LocalCachePath != "" {
+			cache := diskcache.New(spec.LocalCachePath)
```
bbernays, line 98: Do we have to validate that `spec.LocalCachePath` is a directory or file?

plugins/source/github/client/spec.go
```diff
@@ -25,6 +25,8 @@ type Spec struct {
 	DiscoveryConcurrency int `json:"discovery_concurrency,omitempty" jsonschema:"default=1"`
 	// Include archived repositories when discovering repositories.
 	IncludeArchivedRepos bool `json:"include_archived_repos,omitempty"`
+	// Path to a local cache directory. If set, the plugin will cache the GitHub API responses in this directory. Defaults to an empty string (no cache)
```
bbernays, line 28:
```suggestion
	// Path to a local directory that will hold the cache. If set, the plugin will cache the GitHub API responses in this directory. Defaults to an empty string (no cache)
```

plugins/source/github/client/client.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/17368/commits/03203fa62649e3728d39f0840958f90799f2f485

The cache package already does `os. MkdirAll` which should throw a relevant error, but that we might want to fail early here instead of on first write to the cache

bbernays (approved):

PR #17373, opened 2024-03-28
feat: Flush batch on Primary Key collision

bbernays:
#### Summary

Prior to this PR if a batch had rows (for the same table) with duplicate PK values we would just error out and stop the sync. With this change on a PK collision we flush the batch prior to adding the new value. This means all data will hit the DB (so users can still use Triggers)

### DIFF
--- a/plugins/destination/postgresql/client/insert.go
+++ b/plugins/destination/postgresql/client/insert.go
@@ -56,6 +56,7 @@ func getPKValues(r arrow.Record, pkIndex []int) []string {
 	for ri := range results {
 		for _, i := range pkIndex {
 			col := r.Column(i)
+			results[ri] += fmt.Sprintf("%d", i)
 			results[ri] += col.String()
 		}
 	}
--- a/plugins/destination/postgresql/client/insert.go
+++ b/plugins/destination/postgresql/client/insert.go
@@ -100,6 +100,7 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
 			detail = tableDetail{
 				indexes: table.PrimaryKeysIndexes(),
 				sql:     sql,
+				values:  make(map[string]bool),
 			}
 			details[tableName] = detail
 		}
--- a/plugins/destination/postgresql/client/insert.go
+++ b/plugins/destination/postgresql/client/insert.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/apache/arrow/go/v15/arrow"
 	"github.com/cloudquery/plugin-sdk/v4/message"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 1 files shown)


erezrokah (commented):
Can we detect the error instead? Then insert the batch one item at a time?
That way we won't have to calculate the PK value for every record for something that doesn't happen that often

Also it would be nice to add a test for this

plugins/destination/postgresql/client/insert.go
```diff
@@ -59,27 +87,61 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
...
 			if len(table.PrimaryKeysIndexes()) > 0 {
 				sql = c.upsert(table)
 			} else {
 				sql = c.insert(table)
 			}
-			queries[tableName] = sql
+			detail = tableDetail{
+				indexes: table.PrimaryKeysIndexes(),
+				sql:     sql,
+				values:  make(map[string]bool),
```
hermanschaaf, line 103:
```suggestion
				values:  make(map[string]struct{}),
```

using empty `struct{}` here should save a bit of memory

plugins/destination/postgresql/client/insert.go
```diff
@@ -34,6 +35,33 @@ func (c *Client) pgTables(ctx context.Context) (map[string]struct{}, error) {
...
+}
+
+// extracts PKValues from an Arrow record into a string
+// An arrow record may have multiple rows and multiple PK columns so a list of pk value strings is returned
+func getPKValues(r arrow.Record, pkIndex []int) []string {
+	results := make([]string, r.NumRows())
+	for ri := range results {
+		for _, i := range pkIndex {
+			col := r.Column(i)
+			results[ri] += col.String()
```
hermanschaaf, line 59: Edge case, but it seems like we need to use a 2D `[][]string` here, otherwise these two rows will be counted as a collision:


row | pk_col1 | pk_col2 |
----|-----|-----|
1     | aa    | b
2    | a      | ab

In this case it wouldn't really matter since it will just cause a premature flush, but there is also a warning message, so it would be better if it was correct in all cases

plugins/destination/postgresql/client/insert.go
```diff
@@ -59,27 +87,61 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
...
+			if len(detail.indexes) == 0 {
+				batch.Queue(detail.sql, rowVals...)
+				continue
+			}
+			// If the PK value is not in the map of existing pk values, add it to the map and batch the query
+			if _, ok := detail.values[pkVal[i]]; !ok {
+				detail.values[pkVal[i]] = true
+				batch.Queue(detail.sql, rowVals...)
+			} else {
+				c.logger.Warn().Msgf("Duplicate primary key value found for table %s in batch. Flushing batch and creating a new one", tableName)
```
hermanschaaf, line 123: I think we don't need a warning message, as this isn't a problem users of the postgres plugin need to concern themselves with and can happen during normal operations; a debug-level message would be better in my opinion.

hermanschaaf (changes requested):
Couple of comments; also can we add a test for this?

hermanschaaf:
> Can we detect the error instead? Then insert the batch one item at a time?

This could work, but instead of inserting one by one (which could very significantly impact performance), we can use code similar to this PR to make sure we insert batches without any duplicates

plugins/destination/postgresql/client/insert.go
bbernays: Nice catch. I fixed this by adding the col index to the stringified value. This way the pk value can continue to be the map key

bbernays:
Tests have been added for this in the Plugin-sdk: https://github.com/cloudquery/plugin-sdk/pull/1584

disq (approved):

erezrokah:
> Can we detect the error instead?

@bbernays is there a reason to run this logic on each bunch? Can we not detect the error instead? It seems pretty rare that this happens so not sure we want to run a not so simple logic every time

bbernays:
> > Can we detect the error instead?
> 
> @bbernays is there a reason to run this logic on each batch? Can we not detect the error instead? It seems pretty rare that this happens so not sure we want to run a not so simple logic every time

If we detect the error afterwards then what? Insert one-by-one?

erezrokah:
> If we detect the error afterwards then what? Insert one-by-one?

We can apply the same logic as in this PR (as @hermanschaaf suggested)

bbernays:
> > If we detect the error afterwards then what? Insert one-by-one?
> 
> We can apply the same logic as in this PR (as @hermanschaaf suggested)

At the time of the batch being flushed we don't have access to the `msg` we only have the transformed values and the SQL query.

erezrokah:
> At the time of the batch being flushed we don't have access to the `msg` we only have the transformed values and the SQL query. So if we still have to store all of the same data, then it is probably better to just always check for duplicates prior to adding to the batch

Got it, what is the error we're getting? Is it something we can handle in the Postgres level via a query? Also does it happen for both `insert` and `upsert` queries? Because we already have https://github.com/cloudquery/cloudquery/blob/cd73a0a8fd881658694e3fd2434fe6704578d2a7/plugins/destination/postgresql/client/insert.go#L143

PR #17452, opened 2024-04-01
chore: Add API key to GitHub Actions deployment

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/17420

### DIFF
--- a/website/pages/docs/deployment/github-actions.md
+++ b/website/pages/docs/deployment/github-actions.md
@@ -11,6 +11,12 @@ In this tutorial, we will show you how to load AWS resources into a Postgres dat
 
 ## Prerequisites
 
+### Generating a CloudQuery API key
+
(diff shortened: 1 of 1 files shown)


website/pages/docs/deployment/github-actions.md
```diff
@@ -11,6 +11,12 @@ In this tutorial, we will show you how to load AWS resources into a Postgres dat
 
 ## Prerequisites
 
+### Generating a CloudQuery API key
+
+Downloading plugins requires users to be authenticated, normally this means running `cloudquery login` but that is not doable in a CI environment like GitHub Actions. The recommended way to handle this is to use an API key. More information on generating an API Key can be found [here](/docs/deployment/generate-api-key)
```
candiduslynx, line 16:
```suggestion
Downloading plugins requires users to be authenticated, normally this means running `cloudquery login` but that is not doable in a CI environment like GitHub Actions. The recommended way to handle this is to use an API key. More information on generating an API Key can be found [here](/docs/deployment/generate-api-key).
```

candiduslynx (changes requested):

candiduslynx (approved):

Issue #14193, opened 2023-09-29
feat: Update hubspot plugin to do incremental updates


konung:
### Which problem is this feature request solving?

Doing a full sync everytime  is not neccesary. 

### Describe the solution you'd like

Can be implemented similar to Shopify's plugin

### Pull request (optional)

- [ ] I can submit a pull request

PR #17470, opened 2024-04-03
feat: Add incremental sync support for the Hubspot source plugin

maaarcelino:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/14193


plugins/source/hubspot/docs/overview.md
```diff
@@ -19,13 +10,43 @@ You can use the `max_requests_per_second` configuration option to change this va
...
+  backend_options:
+    table_name: "cq_state_hubspot"
+    connection: "@@plugins.DESTINATION_NAME.connection"
+  tables: ["*"]
+  spec:
+    # required, unless the HUBSPOT_APP_TOKEN environment variable is set
+    app_token: "${HUBSPOT_APP_TOKEN}"
+    # optional, default is 5.
+    # See https://developers.hubspot.com/docs/api/usage-details#rate-limits
+    # max_requests_per_second: 5
```
candiduslynx, line 41:
```suggestion
    app_token: "${HUBSPOT_APP_TOKEN}" # required, unless the HUBSPOT_APP_TOKEN environment variable is set
    # max_requests_per_second: 5 # optional, see https://developers.hubspot.com/docs/api/usage-details#rate-limits
```

plugins/source/hubspot/resources/plugin/client.go
```diff
@@ -0,0 +1,156 @@
...
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
+			grpc.WithDefaultCallOptions(
+				grpc.MaxCallRecvMsgSize(maxMsgSize),
+				grpc.MaxCallSendMsgSize(maxMsgSize),
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to dial grpc source plugin at %s: %w", options.BackendOptions.Connection, err)
+		}
+		stateClient, err = state.NewClient(ctx, c.backendConn, options.BackendOptions.TableName)
```
candiduslynx, line 78: It seems that having `backendConn` as `Client` member isn't actually required. Let's get rid of it & only use it in this file.

plugins/source/hubspot/resources/plugin/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+func titleTransformer(table *schema.Table) error {
+	if table.Title != "" {
+		return nil
+	}
+	exceptions := maps.Clone(docs.DefaultTitleExceptions)
+	for k, v := range customExceptions {
+		exceptions[k] = v
+	}
+	csr := caser.New(caser.WithCustomExceptions(exceptions))
```
candiduslynx, line 153: extract to init only once

plugins/source/hubspot/resources/services/crm/contacts_fetch.go
```diff
@@ -45,11 +51,11 @@ func fetchContacts(ctx context.Context, meta schema.ClientMeta, parent *schema.R
...
-		after = next.After
-		if after == "" {
+		if next.After == "" {
 			break
 		}
+		after = next.After
 	}
 
-	return nil
+	return setCursor(ctx, cqClient, key, after)
```
candiduslynx, line 60: IMO, instead of `after` we should use the last value.

Example: we got the response in a single page (so `next.After` is empty). The next sync will resync all of the data.

plugins/source/hubspot/resources/services/crm/cursor.go
```diff
@@ -0,0 +1,27 @@
...
+	if cqClient.Backend != nil && len(cursor) > 0 {
+		if err := cqClient.Backend.SetKey(ctx, key, cursor); err != nil {
+			return fmt.Errorf("failed to store cursor to backend: %w", err)
+		}
+	}
+	return nil
+}
+
+func getCursor(ctx context.Context, cqClient *client.Client, key string) (string, error) {
+	if cqClient.Backend == nil {
```
candiduslynx, line 19: we can just remove this check per setting `NopBackend` client (that won't do anything & return successfully).

plugins/source/hubspot/resources/services/crm/cursor.go
```diff
@@ -0,0 +1,27 @@
+package crm
+
+import (
+	"context"
+	"fmt"
+	"github.com/cloudquery/cloudquery/plugins/source/hubspot/client"
+)
+
+func setCursor(ctx context.Context, cqClient *client.Client, key string, cursor string) error {
+	if cqClient.Backend != nil && len(cursor) > 0 {
```
candiduslynx, line 10: we can just remove this check per setting `NopBackend` client (that won't do anything & return successfully).
```suggestion
	if len(cursor) > 0 {
```

plugins/source/hubspot/resources/services/crm/owners_fetch.go
```diff
@@ -13,7 +13,13 @@ func fetchOwners(ctx context.Context, meta schema.ClientMeta, parent *schema.Res
 	hubspotClient := owners.NewAPIClient(owners.NewConfiguration())
 	cqClient := meta.(*client.Client)
 
-	var after string
+	const key = "owners"
```
candiduslynx, line 16: IMO the key should be the table name instead of the part of the table name

plugins/source/hubspot/resources/services/crm/cursor.go
```diff
@@ -0,0 +1,27 @@
...
+		if err := cqClient.Backend.SetKey(ctx, key, cursor); err != nil {
```
candiduslynx, line 11: we should also prepend the key with the prefix (`c.ID()`?), if the user has different hubspot syncs).
See aws implementation for this

candiduslynx (changes requested):

plugins/source/hubspot/resources/services/crm/contacts_fetch.go
maaarcelino: What do you mean? `after` stores the last pagination cursor. If there was no cursor available (because there is only one page) then there is no state to store and yes, some data will be resynced (but we're talking about less than one page of data). There is nothing we can do about that, though.

plugins/source/hubspot/resources/services/crm/cursor.go
maaarcelino: Done in [602cc6c](https://github.com/cloudquery/cloudquery/pull/17470/commits/602cc6c03fc88a6d35901a8b83d811dc1bf77e1f)

plugins/source/hubspot/resources/plugin/client.go
maaarcelino: Done in [602cc6c](https://github.com/cloudquery/cloudquery/pull/17470/commits/602cc6c03fc88a6d35901a8b83d811dc1bf77e1f)

plugins/source/hubspot/docs/_configuration.md
```diff
@@ -1,16 +0,0 @@
-```yaml copy
```
erezrokah, line 1: Can we keep this file? We use it for managed syncs (or the example from `overview.md`) but we don't want to show `backend_options` in managed syncs since it's not supported yet

maaarcelino: I can revert the file, but how do you want to document `backend_options` for local syncs?

erezrokah: Keep your changes to `overview.md`, those will be use in https://hub.cloudquery.io/plugins/source/cloudquery/hubspot/latest/docs (where we hide the `_configuration.md` file)

erezrokah:
Are we sure we can use the pagination token as the state? What if a contact is deleted for example?

erezrokah:
> What if a contact is deleted for example?

Or even a contact that we already synced was updated - if we use the pagination token we won't sync it again

plugins/source/hubspot/resources/plugin/client.go
```diff
@@ -63,9 +62,8 @@ func (c *Client) Sync(ctx context.Context, options plugin.SyncOptions, res chan<
 	if options.BackendOptions == nil {
 		c.logger.Info().Msg("No backend options provided, using no state backend")
 		stateClient = &state.NoOpClient{}
-		c.backendConn = nil
 	} else {
-		c.backendConn, err = grpc.DialContext(ctx, options.BackendOptions.Connection,
+		backendConn, err := grpc.DialContext(ctx, options.BackendOptions.Connection,
```
candiduslynx, line 66: defer close? (although it's not done anywhere else, so we might have leak here in other plugins)

erezrokah:
So checked the Spotify plugin and it doesn't handle the deleted items too, but it does use `updated_at` for state. We should probably do the same here? Looks like we can sort based on it using the search API https://community.hubspot.com/t5/APIs-Integrations/CRM-API-v3-Sort-based-on-the-modified-date/m-p/350513

plugins/source/hubspot/resources/services/crm/cursor.go
```diff
@@ -6,22 +6,23 @@ import (
...
-	cursor, err := cqClient.Backend.GetKey(ctx, key)
+	cursor, err := cqClient.Backend.GetKey(ctx, generateKey(cqClient, key))
 	if err != nil {
 		return "", fmt.Errorf("failed to retrieve cursor from backend: %w", err)
 	}
 	return cursor, nil
 }
+
+func generateKey(cqClient *client.Client, tableName string) string {
+	return fmt.Sprintf("%s-%s", cqClient.ID(), tableName)
```
candiduslynx, line 27: don't do `fmt` calls here, they're costly
```suggestion
	return cqClient.ID() + "-" + tableName
```

concatenating up to 5 string in Go, however, isn't

candiduslynx (changes requested):

PR #17477, opened 2024-04-04
doc: Add doc for connecting to ClickHouse Cloud

disq:
Had trouble finding out about the `secure` param (and the error msg isn't helpful) as it's buried deep in:
> See [SDK documentation](https://github.com/ClickHouse/clickhouse-go#dsn) for more details.


plugins/destination/clickhouse/docs/overview.md
```diff
@@ -96,6 +96,13 @@ spec:
       - "{replica}"
 ```
 
+### Connecting to ClickHouse Cloud
+
+To connect to [ClickHouse Cloud](https://clickhouse.com/cloud), you need to set the `secure=true` parameter, and the port is `9440`. Use a connection string similar to:
```
candiduslynx, line 101: ref https://clickhouse.com/docs/en/cloud-quick-start#5-using-the-clickhouse-client as well?

candiduslynx: & https://clickhouse.com/docs/en/integrations/go#connection-settings-1

candiduslynx: possibly, for the connections_string docs as well (& json schema)

disq: added in https://github.com/cloudquery/cloudquery/pull/17477/commits/d2d135f698b85a03545d69276f2f4876c8bf08b7

disq: The "See [SDK documentation](https://github.com/ClickHouse/clickhouse-go#dsn) for more details." covers the rest I think, and they can always deep dive further.

candiduslynx (approved):

PR #17530, opened 2024-04-04
feat: Add `validate-config` command

bbernays:
#### Summary

This PR adds a strict `validate-config` command that ensures that all specified configs match with the plugin JSON Spec. This validation is more strict than what we require during a `sync` but if a config passes this validation it will pass the validation at sync time


cli/cmd/test_connection.go
```diff
@@ -87,6 +87,7 @@ func newCmdTestConnection() *cobra.Command {
 		RunE:    testConnection,
 		Hidden:  true,
 	}
+
```
erezrokah, line 90:
```suggestion
```

website/pages/docs/reference/cli/cloudquery_validate-config.md
```diff
@@ -0,0 +1,47 @@
+---
```
erezrokah, line 1: You also need to update https://github.com/cloudquery/cloudquery/blob/f509e8b1c961427979cdb3a525e7bc88bdaea308/cli/cmd/doc_test.go#L11

erezrokah (approved):

cli/cmd/validate_config.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/cloudquery/cloudquery/cli/internal/auth"
+	"github.com/cloudquery/cloudquery/cli/internal/specs/v0"
+	"github.com/cloudquery/plugin-pb-go/managedplugin"
+	"github.com/cloudquery/plugin-pb-go/pb/plugin/v3"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+)
+
+const (
+	validateConfigShort   = "Validate configs"
```
disq, line 17:
```suggestion
	validateConfigShort   = "Validate config"
```

cli/cmd/validate_config.go
```diff
@@ -0,0 +1,144 @@
...
+	validateConfigLong    = "Validate configs without requiring any credentials or connections. This validation is stricter than the validation done during `sync`, but if it passes this validation it will pass the sync validation."
```
disq, line 18:
```suggestion
	validateConfigLong    = "Validate configuration without requiring any credentials or connections. This validation is stricter than the validation done during `sync`, but if it passes this validation it will pass the sync validation."
```

cli/cmd/validate_config_test.go
```diff
@@ -0,0 +1,57 @@
...
+	cqDir := t.TempDir()
+	defer os.RemoveAll(cqDir)
+
+	for _, tc := range configs {
+		t.Run(tc.name, func(t *testing.T) {
+			defer CloseLogFile()
+			testConfig := path.Join(currentDir, "testdata", tc.config)
+			logFileName := path.Join(cqDir, "cloudquery.log")
+			cmd := NewCmdRoot()
+			cmd.SetArgs([]string{"validate-config", testConfig, "--cq-dir", cqDir, "--log-file-name", logFileName})
```
erezrokah, line 40: Let's use https://github.com/cloudquery/cloudquery/blob/f509e8b1c961427979cdb3a525e7bc88bdaea308/cli/cmd/cmd_test.go#L8 here to ensure different tests share the log file/cq dir

cli/cmd/testdata/validate-config-error.yml
```diff
@@ -0,0 +1,20 @@
+kind: source
+spec:
+  name: aws
+  path: cloudquery/aws
+  registry: cloudquery
+  version: "v25.5.0"
```
erezrokah, line 6: Our CLI tests are unauthenticated so using this version might not work

bbernays: I am seeing other weird issues as well even locally

erezrokah: https://github.com/cloudquery/cloudquery/pull/17530/commits/38c85d8443bffe1f6153bea3c2e4df029563663a should fix it and uses `cloudflare` as it's faster to download

erezrokah:
I would also point us that this won't validate the `tables` configuration is valid. For that we would need to init the plugin and call the `Tables` gRPC method

bbernays:
> I would also point us that this won't validate the `tables` configuration is valid. For that we would need to init the plugin and call the `Tables` gRPC method

Added that in

Issue #17475, opened 2024-04-03
feat: Make GitHub source plugin Workflow Runs into an incremental tables


erezrokah:
### Which problem is this feature request solving?

This should help with GitHub rate limits and sync times as even small organizations can have many workflow runs

### Describe the solution you'd like

We could use the `created` parameter https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository

### Pull request (optional)

- [X] I can submit a pull request

erezrokah:
We can't make workflow runs incremental until we can sort the results in `created_at` ascending order (or we persist state only after getting all workflows, but that's not feasible since we'll hit rate limits before on most repositories).

https://github.com/cloudquery/cloudquery/pull/17539 is released in [`v9.2.0`](https://github.com/cloudquery/cloudquery/releases/tag/plugins-source-github-v9.2.0) and allows filtering workflow runs to get only recent ones (e.g. `last 3 days`)

PR #17539, opened 2024-04-05
feat: Add table option to filter workflow runs

erezrokah:
#### Summary

Tried to do https://github.com/cloudquery/cloudquery/issues/17475 using `created_at` column but that's not really possible as while we can filter based on a `created_at` range the results are always returned in descending order, so we always get newest results first.
Which means that if the sync stops in the middle and we only get part of the data, next sync we'd have to split the range to get the runs before the oldest we retrieved and after the newest (and do a split each time we have partial state).

We could maybe save the state only after retrieving all the pages (and not after each page) which should ensure we don't have data to catch up on in the next sync.

This PR does something else, it adds a table option to let users specify the range they'd like to sync.

Opening for feedback


plugins/source/github/client/spec.go
```diff
@@ -28,6 +30,20 @@ type Spec struct {
 	IncludeArchivedRepos bool `json:"include_archived_repos,omitempty"`
 	// Path to a local directory that will hold the cache. If set, the plugin will cache the GitHub API responses in this directory. Defaults to an empty string (no cache)
 	LocalCachePath string `json:"local_cache_path,omitempty"`
+
+	// Table options to set for specific tables.
+	TableOptions TableOptions `json:"table_options,omitempty"`
+}
+
+type TableOptions struct {
+	WorkflowRuns WorkflowRunsOptions `json:"github_workflow_runs,omitempty"`
```
candiduslynx, line 39: maybe add comment here as well (for doc gen purposes)?

plugins/source/github/resources/services/actions/workflow_runs.go
```diff
@@ -31,11 +32,29 @@ func fetchWorkflowRuns(ctx context.Context, meta schema.ClientMeta, _ *schema.Re
...
 	for {
 		workflowRuns, resp, err := c.Github.Actions.ListRepositoryWorkflowRuns(ctx, *repo.Owner.Login, *repo.Name, actionOpts)
 		if err != nil {
 			return err
 		}
+		// When setting created_at, the API will return a maximum of 1000 records, so if we reached the limit we need to get the other records
+		// Workflows runs are sorted by created_at in descending order, so we need to scope down the query
+		if len(workflowRuns.WorkflowRuns) == 0 && actionOpts.Page > 0 && actionOpts.Created != "" {
+			actionOpts.Page = 0
+			actionOpts.Created = c.Spec.TableOptions.WorkflowRuns.ParsedTimeSince + ".." + lastCreatedAt
```
candiduslynx, line 49: shouldn't we use `lastCreatedAt` value on the 2nd loop iteration & so on (`lastCreatedAt` + ".." + time.Now()...`)?

erezrokah: > shouldn't we use `lastCreatedAt` value on the 2nd loop iteration & so on (`lastCreatedAt` + ".." + time.Now()...`)?

Since the results are sorted by `created_at` descending, you'd always get the newest results, hence we need to scope the query "back in time" instead of always getting the latest values.
To clarify more:
1. `>=ParsedTimeSince` will give us the most recent 1000 workflow runs
2. On each loop we save the 1000th result `CreatedAt` then do `ParsedTimeSince...CreatedAt` which will give us the next 1000

This is all assuming the query has more than 1000 result (otherwise we won't this code).

If we could do `lastCreatedAt..time.Now()` we would have been able to use `lastCreatedAt` as state and not need this option

Let me write a test for this to make it clear

candiduslynx: > Since the results are sorted by `created_at` descending
yep, now that makes sense, no need for extra test, just doc it in the comments

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/17539/commits/81782372d3871945f779486d97a3b767cfe0a4a6

erezrokah:
Ok forgot to add docs, doing it now

candiduslynx (approved):

Issue #17524, opened 2024-04-04
feat: `tables` command only return tables specified in config


bbernays:
### Which problem is this feature request solving?

As a user I want to know exactly what tables will be synced by a config.

### Describe the solution you'd like

Right now if I use the `cloudquery tables config.yml` command this will output all of the tables for the plugins included in that command.

My ideal solution would be to expose a CLI flag that specifies whether the CLI should send `tables:[*]` or use the settings specified in the config file to generate the list of tables

### Pull request (optional)

- [ ] I can submit a pull request

erezrokah:
We weren't able to do this before as docs generation was done in the plugins and didn't expose an option to control the list of tables. With https://github.com/cloudquery/cloudquery/pull/12758 should be not that hard to do

erezrokah:
This is done via https://github.com/cloudquery/cloudquery/pull/17584

PR #17551, opened 2024-04-08
feat: Add `filter` flag to `tables` command

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/17524

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -48,6 +48,8 @@ jobs:
           skip-build-cache: true
       - name: Build
         run: go build .
+      - name: Collect Workflow Telemetry
+        uses: catchpoint/workflow-telemetry-action@v2
       - name: Test
         run: make test
       - name: gen
--- a/cli/cmd/tables.go
+++ b/cli/cmd/tables.go
@@ -20,7 +20,7 @@ cloudquery tables ./directory
 cloudquery tables ./directory --format markdown
 # You can also specify an output directory. The default is ./cq-docs
 cloudquery tables ./directory --output-dir ./docs
-# You can also filter which tables are included in the output. The default is all, you can override it with --filter to include only tables referenced in the spec
+# You can also filter which tables are included in the output. The default is all, use --filter=spec to include only tables referenced in the spec
 cloudquery tables ./directory --filter spec
 `
 )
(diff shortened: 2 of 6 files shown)


bbernays (approved):

.github/workflows/cli.yml
```diff
@@ -49,7 +49,11 @@ jobs:
       - name: Build
         run: go build .
       - name: Test
-        run: make test
+        if: matrix.os == 'ubuntu-latest'
+        run: go test -v -race ./...
+      - name: Test
+        if: matrix.os != 'ubuntu-latest'
+        run: go test -v ./...
```
candiduslynx, line 56: I'm worried about this change. Why do we need it?

cli/cmd/addon_publish_test.go
```diff
@@ -91,9 +89,7 @@ func TestAddonPublishEmbedded(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if diff := cmp.Diff(wantCalls, gotCalls); diff != "" {
-		t.Fatalf("mismatch (-want +got):\n%s", diff)
-	}
+	require.Empty(t, cmp.Diff(wantCalls, gotCalls), "unexpected calls to the server")
```
candiduslynx, line 92: Please, move refactoring tests to a separate pr

cli/cmd/tables.go
```diff
@@ -34,6 +36,7 @@ func NewCmdTables() *cobra.Command {
 	}
 	cmd.Flags().String("output-dir", "cq-docs", "Base output directory for generated files")
 	cmd.Flags().String("format", "json", "Output format. One of: json, markdown")
+	cmd.Flags().String("filter", "all", "Filter tables. One of: all, spec")
```
candiduslynx, line 39: Maybe make default empty?

erezrokah: I think `all` conveys better what would be the output the default behavior, WDYT?

candiduslynx: IMO `--filter` is a restrictive option, so, it it's empty/not specified, `allow all` is implied.
So I'd prefer saying that the default is `""`

erezrokah: I think from a user perspective the behavior is the same. If they don't specify the flag it will return all tables. Also specifying `--filter all` seem nicer than? Otherwise it looks like a boolean flag I think

Issue #17092, opened 2024-03-08
feat(github): Add support to list api-tokens 


manjunathsc:
### Which problem is this feature request solving?

I would like to fetch  api-tokens created in a GitHub 

### Describe the solution you'd like

I would like to fetch  api-tokens created in a GitHub 

### Pull request (optional)

- [ ] I can submit a pull request

manjunathsc:
API Tokens - this objects defines an API token created a by a user to allow SW access to github resources.

Use Case
Which tokens and API service integrations are enabled in my environment is an important data point for security audits.

It would be super useful if these resources could be added to github.

bbernays:
@manjunathsc- Is this the information you are looking for: https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#list-requests-to-access-organization-resources-with-fine-grained-personal-access-tokens

If it is not the information you want can you point to the API that does return the data you are interested in?

erezrokah:
Per https://github.com/cloudquery/cloudquery/issues/17092#issuecomment-2050675390 as it's no clear what to implement I think we can close this as stale, pending further comments on the issue

erezrokah:
Re-opening so @candiduslynx can take another look

candiduslynx:
Currently this [request](https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28#list-fine-grained-personal-access-tokens-with-access-to-organization-resources) isn't implemented in the SDK we use, so closing

PR #17555, opened 2024-04-08
feat: Check Connection when user is using PAT

bbernays:
#### Summary

Check Connection when user is using PAT


plugins/source/github/client/services.go
```diff
@@ -15,6 +15,7 @@ type GithubServices struct {
 	Repositories    RepositoriesService
 	Teams           TeamsService
 	DependencyGraph DependencyGraphService
+	Users           UsersService
```
erezrokah, line 18: You need to initialize the `Users` service via https://github.com/cloudquery/cloudquery/blob/e4d94d08c9ebebf08ded2a7b1b33fe5bcd0976c4/plugins/source/github/client/client.go#L164

plugins/source/github/resources/services/users/keys.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/transformers"
+	"github.com/google/go-github/v59/github"
+)
+
+func keys() *schema.Table {
+	return &schema.Table{
+		Name:      "github_user_keys",
+		Resolver:  fetchMembers,
+		Transform: client.TransformWithStruct(&github.Key{}, transformers.WithPrimaryKeys("ID")),
```
erezrokah, line 17: Can we add a description and link to https://docs.github.com/en/rest/users/keys?apiVersion=2022-11-28#list-public-ssh-keys-for-the-authenticated-user?

plugins/source/github/resources/services/users/users.go
```diff
@@ -0,0 +1,35 @@
...
+		Name:     "github_users",
+		Resolver: fetchUser,
+		// No multiplexer is needed here because we are fetching the current user
+		Transform: client.TransformWithStruct(&github.User{}, transformers.WithPrimaryKeys("ID")),
+		Columns:   []schema.Column{client.OrgColumn},
+		Relations: []*schema.Table{keys()},
+	}
+}
+
+func fetchUser(ctx context.Context, meta schema.ClientMeta, _ *schema.Resource, res chan<- any) error {
```
erezrokah, line 23: Can we move this code to the discovery/client init phase? That way we don't need the extra table. Not sure it's that useful

erezrokah (changes requested):
Added a few comments, also I believe this is a different resource than https://github.com/cloudquery/cloudquery/issues/17092 which is API Tokens (relevant APIs here I think https://docs.github.com/en/rest/orgs/personal-access-tokens?apiVersion=2022-11-28

plugins/source/github/resources/services/users/users.go
bbernays: My thought is if we have to make the API call then might as well persist the data

erezrokah: I think this would also serve as a good connection test when someone is using token based auth

erezrokah (approved):

PR #17556, opened 2024-04-08
chore: Use `require.NoError` to check for error with `fakeObject`

bbernays:
#### Summary

use same pattern used in other plugins (aws)


erezrokah (approved):

erezrokah:
Looks good, just needed a `go mod tidy`

PR #17568, opened 2024-04-09
fix: Don't call `RowsAffected` on query error

erezrokah:
#### Summary

Fixes the panic from https://github.com/cloudquery/cloudquery/actions/runs/8607255034/job/23587408786?pr=17562#step:8:10


plugins/destination/duckdb/client/client.go
```diff
@@ -91,10 +91,15 @@ func (c *Client) Close(ctx context.Context) error {
...
+		return err
+	}
 	if c.spec.Debug {
 		rowsAffected, rowsErr := r.RowsAffected()
-		errs := errors.Join(err, rowsErr)
-		c.logger.Debug().Str("query", query).Any("values", args).Int64("rowsAffected", rowsAffected).Err(errs).Msg("exec query")
+		if rowsErr != nil {
+			return rowsErr
+		}
+		c.logger.Debug().Str("query", query).Any("values", args).Int64("rowsAffected", rowsAffected).Msg("exec query")
```
disq, line 102: This wouldn't log rowsErr or the query err and you'd only get debug messages for successful queries.

erezrokah: Cool, so I guess we can keep this log 👍

erezrokah: Well actually won't returning the error make it so it's logged always?

erezrokah: Anyway maybe keep it since we don't want to fail if we couldn't write a debug log

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/17568/commits/c11c4020358acfe53b29433e213df8236f439dbf

erezrokah: Actually when I look at the previous code, we would return `rowsErr` even if `err` was `nil`. Not sure that is what we want if the intention was only to log it

disq: Previous code would return _all_ errors (if any) thanks to the `errs := errors.Join(err, rowsErr)` line.

erezrokah: > Previous code would return _all_ errors (if any) thanks to the `errs := errors.Join(err, rowsErr)` line

It won't since it panics 🙃

disq: https://github.com/cloudquery/cloudquery/commit/c11c4020358acfe53b29433e213df8236f439dbf isn't the same as before, ideally if c.spec.Debug is on it should log the query and error before returning (not running rows affected)

An easier approach could be to check if err is nil inside c.spec.Debug and run rowsaffected and join the errors (or just don't care about the rowsaffected error) inside the block.

erezrokah: Maybe https://github.com/cloudquery/cloudquery/pull/17568/commits/011c48b3ca6763afce698c5564f3f2387d558cd6 to reduce the nesting?

disq (approved):

Issue #17579, opened 2024-04-10
bug: Docker plugins from CloudQuery registry don't work with team API keys


erezrokah:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using an API key with Docker plugins from CloudQuery registry fail with an error:
```
Error: team required. Hint: use `cloudquery switch` to set a team
```

### Expected Behavior

CLI should be able to work with Docker plugins from CloudQuery registry and team API keys

### CloudQuery (redacted) config

```yaml
kind: source
spec:
  name: airtable
  registry: docker
  path: docker.cloudquery.io/cloudquery/source-airtable:v2.1.2
  tables:
    - "*"
  destinations: ["postgresql"]
  spec:
    access_token: "redacted"
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v8.0.2"
  spec:
    connection_string: "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"
```

### Steps To Reproduce

1. Run `CLOUDQUERY_API_KEY=<your-api-key> cloudquery plugin install spec.yml` with the above spec

### CloudQuery (redacted) logs

N/A

### CloudQuery version

`cloudquery version 5.13.1`

### Pull request (optional)

- [X] I can submit a pull request

PR #17580, opened 2024-04-10
fix: Get team for API tokens

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/17579

The team name is propagated to the Docker registry to use in the authentication flow here:
https://github.com/cloudquery/plugin-pb-go/blob/78b50f093a96d903467634c085f3be46dcfaf8db/managedplugin/docker.go#L109


cli/internal/auth/team.go
```diff
@@ -1,22 +1,38 @@
...
+		if err != nil {
+			return "", fmt.Errorf("failed to create api client: %w", err)
+		}
+		resp, err := apiClient.ListTeamsWithResponse(ctx, &cqapi.ListTeamsParams{})
+		if err != nil {
+			return "", fmt.Errorf("failed to list teams: %w", err)
+		}
+		return resp.JSON200.Items[0].Name, nil
+	default:
+		return os.Getenv("_CQ_TEAM_NAME"), nil
```
erezrokah, line 36: Placeholder for Cloud Syncs where `_CQ_TEAM_NAME` is set (we don't use it at the moment)

candiduslynx (approved):

Issue #17547, opened 2024-04-08
Doc: Add guide on how to run Docker plugins when running the CLI inside Docker


erezrokah:
This has been popping up once in a while and causing quite a bit of friction to our users (see https://github.com/cloudquery/cloudquery/issues/17531).

It's possible (though not recommended) to run Docker inside docker. Which should add a guide for it

bbernays:
Maybe instead of documenting `docker in docker` we should document how to run a docker plugin as a standalone gRPC server. We can show users how they can create a docker-compose to ensure that multiple sources/destinations are all running before running the sync on a separate container

erezrokah:
> We can show users how they can create a docker-compose to ensure that multiple sources/destinations are all running before running the sync on a separate container

That could work to I'll look into it 👍

Issue #17531, opened 2024-04-05
bug: AIRTABLE Plugin: sync failed with error message from the log


shotleybuilder:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Airtable plug-in
From the log:
2024-04-05T08:59:55Z ERR exiting with error error="failed to pull Docker image: Error response from daemon: Head \"https://docker.cloudquery.io/v2/cloudquery/source-airtable/manifests/v2.1.2\": unknown: The server could not parse your request body" module=cli

### Expected Behavior

I'm not sure what to expect since is the first time I've used cloudquery

### CloudQuery (redacted) config

```yaml config
kind: source
# Common source-plugin configuration
spec:
  name: airtable
  registry: docker
  path: docker.cloudquery.io/cloudquery/source-airtable:v2.1.2
  tables:
    # Sync all tables under bases matching the name `base_name_to_sync`
    #- "*__UK__*"
    # Sync all tables matching the name `table_name_to_sync`
    #- "*__*__LRT"
    # Sync all tables matching the name `table_name_to_sync` under bases matching the name `base_name_to_sync`
    - "*__UK__LRT"
  destinations: ["supabase"]
  # airtable-specific configuration
  spec:
    access_token: "redacted" # required
    # endpoint_url: "https://api.airtable.com" # Optional, defaults to `https://api.airtable.com`
    # concurrency: 10000 # Optional, defaults to `10000`

```

### CloudQuery (redacted) logs

```
2024-04-05T08:55:42Z ERR exiting with error error="failed to list Docker images: permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get \"http://%2Fvar%2Frun%2Fdocker.sock/v1.43/images/json?filters=%7B%22reference%22%3A%7B%22docker.cloudquery.io%2Fcloudquery%2Fsource-airtable%3Av2.1.2%22%3Atrue%7D%7D\": dial unix /var/run/docker.sock: connect: permission denied" module=cli
2024-04-05T08:59:55Z INF Loading spec(s) args=["cloudquery-config/airtable.yml","cloudquery-config/postgresql.yml"] module=cli
2024-04-05T08:59:55Z WRN when using the CloudQuery registry, it's recommended to log in via `cloudquery login`. Logging in allows for better rate limits and downloading of premium plugins module=cli
[...]

PR #17581, opened 2024-04-10
chore: Add guide on how to run Docker plugins when running the CLI inside Docker

erezrokah:
#### Summary

Based on https://github.com/cloudquery/cloudquery/issues/17547#issuecomment-2042904677

Fixes https://github.com/cloudquery/cloudquery/issues/17547
Fixes https://github.com/cloudquery/cloudquery/issues/17531

Might need better phrasing and name for the title


website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
```diff
@@ -0,0 +1,112 @@
...
+    configs:
+      - spec.yml
+    depends_on:
+      airtable:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+  airtable:
+    container_name: airtable
+    image: docker.cloudquery.io/cloudquery/source-airtable:v2.1.2
```
bbernays, line 60: Is there a standardized way that users can know/find  the docker image so this can be relevant to all dockerized plugins?

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
```diff
@@ -0,0 +1,112 @@
+# Running Docker Plugins With the CLI Docker Image
```
bbernays, line 1:
```suggestion
# Running Docker Plugins
```

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
erezrokah: The path is `docker.cloudquery.io/<team>/<kind>-<name>:<version>`

Each plugin has this documented

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
```diff
@@ -0,0 +1,112 @@
...
+  tables: ["*"]
+  destinations: ["postgresql"]
+---
+kind: destination
+spec:
+  name: "postgresql"
+  path: "cloudquery/postgresql"
+  version: "VERSION_DESTINATION_POSTGRESQL"
+```
+3. Run `cloudquery plugin install spec.yml` to pull the Docker plugin image locally.
```
bbernays, line 32: Could we just have them run `docker pull <image>`?

erezrokah: No that won't work as our registry is private. Also `docker login docker.cloudquery.io` won't work as our registry requires you to pass a team name in a special header https://github.com/cloudquery/plugin-pb-go/blob/78b50f093a96d903467634c085f3be46dcfaf8db/managedplugin/docker.go#L121
There's a way to configure the additional headers but it's a bit complicated

pilvikala (commented):
I think I expected something a bit different (not using plugins as containers in a docker-compose) but to my surprise, we did not have this one either so it's actually quite a good improvement to our docs!

I think this is struggling a bit between trying to make this very basic (like opening postgres ports while specifying easy password in plain text) and too advanced (with healthchecks and stuff). I'd recommend to choose one approach and keep it. I'd start with the basic thing to provide the minimum example with a callout that this is an example to show how it's supposed to work and should not be used as is in production.
In short, I'd cut the dependencies, restrictions on pulling

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
```diff
@@ -0,0 +1,112 @@
...
+    depends_on:
+      airtable:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+  airtable:
+    container_name: airtable
+    image: docker.cloudquery.io/cloudquery/source-airtable:v2.1.2
+    # We use `cloudquery login` and `cloudquery plugin install spec.yml` to pull the image locally
+    pull_policy: never
```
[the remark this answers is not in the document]
erezrokah: Explained in https://github.com/cloudquery/cloudquery/pull/17581#discussion_r1559613735, only our CLI can pull from our private registry since we pass the team name

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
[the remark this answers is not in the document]
erezrokah: Only our CLI can pull from our private registry since we pass the team name

erezrokah:
> I think I expected something a bit different (not using plugins as containers in a docker-compose) but to my surprise, we did not have this one either so it's actually quite a good improvement to our docs!
> 
> I think this is struggling a bit between trying to make this very basic (like opening postgres ports while specifying easy password in plain text) and too advanced (with healthchecks and stuff). I'd recommend to choose one approach and keep it. I'd start with the basic thing to provide the minimum example with a callout that this is an example to show how it's supposed to work and should not be used as is in production. In short, I'd cut the dependencies, restrictions on pulling, just to focus on the core here.

This is the minimal example :) Without the health checks the CLI would start syncing before the plugin or the DB is ready and fail. The pull restrictions are required since we need to pass a team name to our registry as a part of the authentication flow (we need to know which team downloads which plugin).

As for the approach - to run multiple containers you can use compose or k8s which is more complicated I think. Maybe there's a simpler approach I'm not aware of (other than "docker in docker").

Please note this is doc is under the "advanced" section of our docs

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
bbernays: It doesn't seem that hard to specify the custom header... https://docs.docker.com/engine/reference/commandline/cli/#custom-http-headers

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
```diff
@@ -0,0 +1,112 @@
+# Running Docker Plugins
```
erezrokah, line 1:
```suggestion
# Using CloudQuery Docker plugins inside a containerized environment 
```

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
erezrokah: You need to modify your `config.json` and the headers are sent will all requests to all registries I believe

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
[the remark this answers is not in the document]
erezrokah: Changed it to `Using CloudQuery Docker Registry Plugins Inside a Containerized Environment`

website/pages/docs/advanced-topics/docker-cli-docker-plugins.md
erezrokah: We also pass the plugin version since docker doesn't https://github.com/cloudquery/plugin-pb-go/blob/78b50f093a96d903467634c085f3be46dcfaf8db/managedplugin/docker.go#L120 which is dynamic and we can't put it in the `config.json`

pilvikala (approved):

PR #17584, opened 2024-04-10
feat: Add filter flag to tables command

erezrokah:
#### Summary

Instead of https://github.com/cloudquery/cloudquery/pull/17551

### DIFF
--- a/cli/cmd/tables.go
+++ b/cli/cmd/tables.go
@@ -20,7 +20,7 @@ cloudquery tables ./directory
 cloudquery tables ./directory --format markdown
 # You can also specify an output directory. The default is ./cq-docs
 cloudquery tables ./directory --output-dir ./docs
-# You can also filter which tables are included in the output. The default is all, you can override it with --filter to include only tables referenced in the spec
+# You can also filter which tables are included in the output. The default is all, use --filter=spec to include only tables referenced in the spec
 cloudquery tables ./directory --filter spec
 `
 )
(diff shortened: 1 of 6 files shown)


bbernays (approved):

cli/cmd/tables_test.go
```diff
@@ -22,6 +25,9 @@ func getTablesCommand(t *testing.T, config string, format string) (*cobra.Comman
 	if format != "" {
 		args = append(args, "--format", format)
 	}
+	if filter != "" {
```
candiduslynx, line 28: (nit): easier to do is to get string len (a single lookup) instead of str cmp

cli/cmd/tables.go
```diff
@@ -34,6 +36,7 @@ func NewCmdTables() *cobra.Command {
 	}
 	cmd.Flags().String("output-dir", "cq-docs", "Base output directory for generated files")
 	cmd.Flags().String("format", "json", "Output format. One of: json, markdown")
+	cmd.Flags().String("filter", "all", "Filter tables. One of: all, spec")
```
candiduslynx, line 39:
```suggestion
	cmd.Flags().String("filter", "", "Filter tables. One of: \"\", spec")
```
the same suggestion as in the original PR

erezrokah: I think it's a bit strange to pass `--filter` with an empty value since it's a enum flag. If you drop it will return all tables

PR #17587, opened 2024-04-11
doc: Fix filenames in markdown

pilvikala:
Noticed a few styling issues. The `copy` appeared to have no effect other than actually hiding the filenames from the yaml config examples, so I fixed that.


erezrokah (approved):

erezrokah:
> The `copy` appeared to have no effect other than actually

Yeah I think we changed it a while back to have the copy button enabled by default instead of opt-in so no need to add it manually like we did before

PR #17601, opened 2024-04-11
feat: Validate AWS Tag Structure is not `[]`

bbernays:
#### Summary

Adds a test to validate that the JSON object is not `[]`


plugins/source/aws/client/testing.go
```diff
@@ -150,3 +153,36 @@ func validateNoEmptyColumnsExcept(t *testing.T, tables schema.Tables, messages m
...
+	tablesWithIssues := []string{}
+	for _, table := range tables.FlattenTables() {
+		if table.Column("tags") == nil {
+			continue
+		}
+
+		records := messages.GetInserts().GetRecordsForTable(table)
+		for _, resource := range records {
+			for colIndex, arr := range resource.Columns() {
+				if table.Columns[colIndex].Name != "tags" {
```
candiduslynx, line 168: get index by name & use only that?
https://pkg.go.dev/github.com/cloudquery/plugin-sdk/v4@v4.38.0/schema#ColumnList.Index

plugins/source/aws/client/testing.go
```diff
@@ -150,3 +153,36 @@ func validateNoEmptyColumnsExcept(t *testing.T, tables schema.Tables, messages m
...
+					continue
+				}
+				for i := 0; i < arr.Len(); i++ {
+					if arr.IsValid(i) {
```
candiduslynx, line 172: less indent could be better via
```go
if arr.IsNull(i) { continue }
```

candiduslynx (changes requested):

candiduslynx (approved):

plugins/source/aws/client/testing.go
```diff
@@ -150,3 +153,27 @@ func validateNoEmptyColumnsExcept(t *testing.T, tables schema.Tables, messages m
...
+	for _, table := range tables.FlattenTables() {
+		index := table.Columns.Index("tags")
+		if index == -1 {
+			continue
+		}
+		records := messages.GetInserts().GetRecordsForTable(table)
+		for _, resource := range records {
+			arr := resource.Column(index)
+			for i := 0; i < arr.Len(); i++ {
+				val := arr.GetOneForMarshal(i).(json.RawMessage)
```
candiduslynx, line 168: Won't it panic on nil?

candiduslynx (changes requested):

plugins/source/aws/client/testing.go
bbernays: It should never be null. If it was empty the test would have failed at this point

candiduslynx (approved):

PR #17602, opened 2024-04-11
feat!: Add path variable support to GCS destination plugin. 

### LINKED ISSUE #15097: feat: GCS Destination Templating

Currently with an extract from Postgres -> GCS the data is all landed in a single directory within a bucket. While this isn't necessarily problematic, it's much more convenient to have tables landed in their own directories, and with additional metadata about the export. I imagine this is the same across any file based storage.

It would be really helpful to be able to add some mild templating to `path` strings so that cloudquery can arbitrarily decide where to land data. I'm proposing a structure where the `spec.spec.path` for a gcs destination looks something like: `path/to/file/{source_schema}/{source_table}/{sync_timestamp}`. I realize this probably doesn't work when the source isn't a DB but sync_timestamp would be valuable.

Given this structure if I were to export two tables: `public.actions` and `public.orders` and the sync ran at `2023-11-01 14:52:51` the destination for these files would be something like: `path/to/file/public/actions/20231101145251/*.csv` and `path/to/file/public/orders/20231101145251/*.csv`.

This organization and structure makes it easier to reason about files landed in GCS, and should help to make ingestion into consumers easier too.

### Pull request (optional)

shubham-padia:
Fixes #15097 . Since the PR is fairly large, please review it commit by commit along with a read of the commit message descriptions to get a better sense of the changes. 

#### Summary

Add path variable support to GCS destination plugin. **This is a breaking change**. We check if the path string has any of the template variables like {{FORMAT}}, {{TABLE}} and so on. If they do, we follow the behaviour of S3 plugin wherein we replace the template variables and that's it. If there are no such variables, we revert back to the current behaviour in the code block shown above.

Please read the issue comments to check on how the decisions for this PR were made. 

This PR follows similar approach to the S3 plugin. I've added some extra tests for client.read() errors which we should probably copy back to S3 and File plugins in a follow up issue. 

Note: I have not made additions to CHANGELOG.md, I'm assuming the maintainer will do it just before merge, in order to make sure of the right tag at the time of merge. Please lmk if I need to make that change

<details> 
<summary> GCS screenshot for path `new/{{TABLE}}/{{UUID}}.{{FORMAT}}` </summary>

Screenshot 2024-04-12 at 1 46 50 AM

</details>

- [x] Read the [contribution guidelines](https://github.com/cloudquery/cloudquery/blob/main/CONTRIBUTING.md) 🧑‍🎓

### DIFF
--- a/plugins/destination/gcs/client/client.go
+++ b/plugins/destination/gcs/client/client.go
@@ -22,6 +22,7 @@ type Client struct {
 	streamingbatchwriter.UnimplementedDeleteStale
 	streamingbatchwriter.UnimplementedDeleteRecords
 
+	syncID string
 	logger zerolog.Logger
 	spec   *spec.Spec
 
@@ -32,9 +33,10 @@ type Client struct {
 	writer *streamingbatchwriter.StreamingBatchWriter
 }
 
-func New(ctx context.Context, logger zerolog.Logger, s []byte, _ plugin.NewClientOptions) (plugin.Client, error) {
+func New(ctx context.Context, logger zerolog.Logger, s []byte, newClientOpts plugin.NewClientOptions) (plugin.Client, error) {
 	c := &Client{
 		logger: logger.With().Str("module", "gcs").Logger(),
+		syncID: newClientOpts.InvocationID,
 	}
 
 	if err := json.Unmarshal(s, &c.spec); err != nil {
--- a/plugins/destination/gcs/client/client_test.go
+++ b/plugins/destination/gcs/client/client_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/rs/zerolog"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const bucket = "cq-dest-gcs"
@@ -39,9 +40,28 @@ func TestPlugin(t *testing.T) {
 		})
 
 		t.Run("write/"+string(ft), func(t *testing.T) {
-			testPluginCustom(t, &s)
+			testPluginCustom(t, &s, false, "")
 		})
 	}
+	t.Run("should give an error while reading when no_rotate is false", func(t *testing.T) {
+		s := spec.Spec{
+			Bucket:   bucket,
+			Path:     t.TempDir(),
+			NoRotate: false,
+			FileSpec: filetypes.FileSpec{Format: filetypes.FormatTypeCSV},
+		}
+		testPluginCustom(t, &s, true, "reading is not supported when `no_rotate` is false")
+	})
+
+	t.Run("should give an error while reading when {{UUID}} path variable is present", func(t *testing.T) {
+		s := spec.Spec{
+			Bucket:   bucket,
+			Path:     "{{UUID}}",
+			NoRotate: true,
+			FileSpec: filetypes.FileSpec{Format: filetypes.FormatTypeCSV},
+		}
+		testPluginCustom(t, &s, true, "reading is not supported when path contains uuid variable")
+	})
 }
 
 func testPlugin(t *testing.T, s *spec.Spec) {
@@ -66,7 +86,7 @@ func testPlugin(t *testing.T, s *spec.Spec) {
 	)
 }
 
-func testPluginCustom(t *testing.T, s *spec.Spec) {
+func testPluginCustom(t *testing.T, s *spec.Spec, expectErrorWhenReading bool, readErrorString string) {
 	ctx := context.Background()
 
 	var client plugin.Client
@@ -119,10 +139,12 @@ func testPluginCustom(t *testing.T, s *spec.Spec) {
 	}
 
 	readRecords, err := readAll(ctx, client, table)
-	if err != nil {
-		t.Fatal(fmt.Errorf("failed to sync: %w", err))
+	if expectErrorWhenReading {
+		require.ErrorContains(t, err, readErrorString)
+		return
+	} else {
+		require.NoError(t, err)
 	}
-
 	totalItems := plugin.TotalRows(readRecords)
 	assert.Equalf(t, int64(2), totalItems, "expected 2 items, got %d", totalItems)
 }
--- a/plugins/destination/gcs/client/client_test.go
+++ b/plugins/destination/gcs/client/client_test.go
@@ -40,7 +40,7 @@ func TestPlugin(t *testing.T) {
 		})
 
 		t.Run("write/"+string(ft), func(t *testing.T) {
-			testPluginCustom(t, &s, false, "")
+			testPluginCustom(t, &s, "")
 		})
 	}
 	t.Run("should give an error while reading when no_rotate is false", func(t *testing.T) {
@@ -50,7 +50,7 @@ func TestPlugin(t *testing.T) {
 			NoRotate: false,
 			FileSpec: filetypes.FileSpec{Format: filetypes.FormatTypeCSV},
 		}
-		testPluginCustom(t, &s, true, "reading is not supported when `no_rotate` is false")
+		testPluginCustom(t, &s, "reading is not supported when `no_rotate` is false")
 	})
 
 	t.Run("should give an error while reading when {{UUID}} path variable is present", func(t *testing.T) {
@@ -60,7 +60,7 @@ func TestPlugin(t *testing.T) {
 			NoRotate: true,
 			FileSpec: filetypes.FileSpec{Format: filetypes.FormatTypeCSV},
 		}
-		testPluginCustom(t, &s, true, "reading is not supported when path contains uuid variable")
+		testPluginCustom(t, &s, "reading is not supported when path contains uuid variable")
 	})
 }
 
@@ -86,7 +86,7 @@ func testPlugin(t *testing.T, s *spec.Spec) {
 	)
 }
 
-func testPluginCustom(t *testing.T, s *spec.Spec, expectErrorWhenReading bool, readErrorString string) {
+func testPluginCustom(t *testing.T, s *spec.Spec, readErrorString string) {
 	ctx := context.Background()
 
 	var client plugin.Client
@@ -139,7 +139,7 @@ func testPluginCustom(t *testing.T, s *spec.Spec, expectErrorWhenReading bool, r
 	}
 
 	readRecords, err := readAll(ctx, client, table)
-	if expectErrorWhenReading {
+	if readErrorString != "" {
 		require.ErrorContains(t, err, readErrorString)
 		return
 	}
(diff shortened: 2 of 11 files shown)


plugins/destination/gcs/client/spec/schema.go
```diff
@@ -41,7 +41,109 @@ func (s Spec) JSONSchemaExtend(sc *jsonschema.Schema) {
...
+			}(),
+			Required: []string{"no_rotate"},
+		},
+		Then: &jsonschema.Schema{
+			Not: pathWithUUID,
+		},
+	}
+
+	/* batching enabled -> require {{UUID}} in path or require no path variables in path,
+	since we will use UUID by default if batch */
```
shubham-padia, line 109: Note: Differing behaviour from S3 plugin in order to maintain backwards compatibility for paths without path variables

plugins/destination/gcs/client/client.go
```diff
@@ -22,6 +22,7 @@ type Client struct {
 	streamingbatchwriter.UnimplementedDeleteStale
 	streamingbatchwriter.UnimplementedDeleteRecords
 
+	syncID string
```
bbernays, line 25: SyncID needs to be set in `client.New()`

You can see how it is done in the S3 destination: https://github.com/cloudquery/cloudquery/blob/9b07ab328eb8b3ef7fab61c9d8458244cc33af93/plugins/destination/s3/client/client.go#L44

shubham-padia: Thanks!

Pushed the changes at https://github.com/cloudquery/cloudquery/compare/c2b3f14b168cd5d8b9f8c5a8474ee94a7c50ea5b..693d590150666a0282a090934885d80fd34d5df9

Test sync on GCS with path as `new/{{TABLE}}/{{UUID}}_{{SYNC_ID}}.{{FORMAT}}`

Screenshot 2024-04-12 at 2 02 00 AM

shubham-padia:
I checked the failing runs:

1. Fixed the grammar mistake in the documentation at https://github.com/cloudquery/cloudquery/compare/693d590150666a0282a090934885d80fd34d5df9..ccbbec85f57557e6582cf6ef4ce72ad103240f7d
2. For https://github.com/cloudquery/cloudquery/actions/runs/8652900255, $ACTIONS_ID_TOKEN_REQUEST_TOKEN is not available to this PR since I think this is running from a fork?
3. For https://github.com/cloudquery/cloudquery/actions/runs/8652900268, where `wait_for_required_workflows` fails, it might have to do something with kodiakhq and auto merging? I could look more into it but a maintainer might be able to help me out much quicker if it's a small issue.

erezrokah:
> 2\. For https://github.com/cloudquery/cloudquery/actions/runs/8652900255, $ACTIONS_ID_TOKEN_REQUEST_TOKEN is not available to this PR since I think this is running from a fork?

💯 Credentials are not available for forked PR, I'll open a copy of this PR from a non forked branch to validate

> 3\. where `wait_for_required_workflows` fails

Is a custom workflow we wrote to facilitate branch protection enforcement in monorepos without having to run workflows for all monorepos components due to https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

It's failing since due to the issue with the credentials

plugins/destination/gcs/client/client_test.go
```diff
@@ -120,7 +139,13 @@ func testPluginCustom(t *testing.T, s *spec.Spec) {
...
 	readRecords, err := readAll(ctx, client, table)
 	if err != nil {
+		if expectErrorWhenReading {
+			assert.ErrorContains(t, err, readErrorString)
+			t.Skip()
+		}
 		t.Fatal(fmt.Errorf("failed to sync: %w", err))
+	} else if expectErrorWhenReading {
+		t.Fatal(fmt.Errorf("expected error while performing client.read(), got None"))
 	}
```
erezrokah, line 149:
```suggestion
	if expectErrorWhenReading {
		require.ErrorContains(t, err,readErrorString)
	} else {
		require.NoError(t, err)
	}
```

Could this work instead? Regardless what's the reason we have `t.Skip()` in the test?

plugins/destination/gcs/docs/overview.md
```diff
@@ -44,7 +54,7 @@ This is the (nested) spec used by the CSV destination Plugin.
 
 - `no_rotate` (`boolean`) (optional) (default: `false`)
 
-  If set to `true`, the plugin will write to one file per table.
+  If set to `true`, the plugin will write to one file per table.
```
erezrokah, line 57:
```suggestion
  If set to `true`, the plugin will write to one file per table.
```

plugins/destination/gcs/client/spec/schema_test.go
```diff
@@ -143,6 +162,30 @@ func TestSpecJSONSchema(t *testing.T) {
...
+			Spec: `{"format": "csv", "path": "abc", "bucket": "b", "no_rotate":false}`,
+		},
+		{
+			Name: "no_rotate:true & path:abc",
+			Spec: `{"format": "csv", "path": "abc", "bucket": "b", "no_rotate":true}`,
+		},
+		{
+			Name: "no_rotate:false & path:{{TABLE}}",
+			Spec: `{"format": "csv", "path": "{{TABLE}}", "bucket": "b", "no_rotate":false}`,
+			Err:  true,
```
erezrokah, line 186: What's the reason this is returning an error and `no_rotate:false & path:abc` doesn't?
Is it to allow to default to the old behavior when there are no path variables and enforce `{{UUID}}` when there are?

erezrokah (commented):
Looks great @shubham-padia, added a few minor comments, sorry for the delayed review

plugins/destination/gcs/client/spec/schema_test.go
erezrokah: OK I think https://github.com/cloudquery/cloudquery/pull/17602/files#r1561583882 answers it

erezrokah:
Looks like tests are passing https://github.com/cloudquery/cloudquery/actions/runs/8665051208/job/23762862668?pr=17608. For the changelog we generate those automatically based on the commit message.

To trigger a major version bump the PR title should start with `feat!:` (see https://github.com/cloudquery/cloudquery/blob/main/CONTRIBUTING.md#commit-messages)

plugins/destination/gcs/docs/overview.md
shubham-padia: Fixed at https://github.com/cloudquery/cloudquery/compare/ccbbec85f57557e6582cf6ef4ce72ad103240f7d..abd53a174e96587e90770b0af5469ecf0a793957

plugins/destination/gcs/client/client_test.go
shubham-padia: Yes, that works much better, thanks @erezrokah. This is my first time using Go, so this is much appreciated! 

I was using t.Skip() to skip the rest of the test when we encounter an error but I saw [this file](https://github.com/cloudquery/cloudquery/blob/7a0936dcb17412460cce67e2f7733db2261fb892/cli/internal/publish/images/images_test.go#L270) uses return, so I changed it to that.

erezrokah: OK so I think we can drop the `return` in both cases. That's the reason to use `require` over `assert`. `require` stops the test on failure https://github.com/stretchr/testify?tab=readme-ov-file#require-package

plugins/destination/gcs/client/client_test.go
```diff
@@ -17,9 +17,10 @@ import (
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/rs/zerolog"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-const bucket = "cq-dest-gcs"
+const bucket = "test-cloudquery"
```
erezrokah, line 23: Noticing this now, what's the reason to change the bucket name?

shubham-padia: Sorry, committed this by mistake, I need to set this everytime to run tests, since GCS won't allow me to name my bucket `cq-dest-gcs`since bucket names need to be globally unique in GCS, I was gonna raise this issue on discord to see how we can make testing a bit easier, but haven't had the time to raise it yet.

shubham-padia: Fixed at https://github.com/cloudquery/cloudquery/compare/abd53a174e96587e90770b0af5469ecf0a793957..2b4bb86f422b2eeaf52046b94623ddc88108aff1

plugins/destination/gcs/client/client_test.go
shubham-padia: @erezrokah I think we still need this return, require will stop the tests on failure, but failure in this case would mean either not getting an error or the error not containing the string that we provided. If the error contains the message, the `require` assertion would pass successfully and move on to the next part of the tests

plugins/destination/gcs/client/client_test.go
```diff
@@ -119,10 +139,12 @@ func testPluginCustom(t *testing.T, s *spec.Spec) {
 	}
 
 	readRecords, err := readAll(ctx, client, table)
-	if err != nil {
-		t.Fatal(fmt.Errorf("failed to sync: %w", err))
+	if expectErrorWhenReading {
+		require.ErrorContains(t, err, readErrorString)
+		return
```
erezrokah, line 144:
```suggestion
```

erezrokah (approved):

plugins/destination/gcs/client/client_test.go
erezrokah: You're right 👍 How about https://github.com/cloudquery/cloudquery/pull/17602/commits/a25ccda535682d27c6e322d9668e8866acd98086 to early return and reduce the `if/else` nesting?

I also added another commit to remove the boolean flag in https://github.com/cloudquery/cloudquery/pull/17602/commits/0982124946b8b6a767963c5d01bcfba490f0fe6a

WDYT?

plugins/destination/gcs/client/client_test.go
erezrokah: I think we can expose it via an environment variable to be configurable (in a separate PR), WDYT?

erezrokah:
@shubham-padia please let me know what you think of my recent changes. I think the PR is ready to be merged unless there's something I missed

plugins/destination/gcs/client/client_test.go
shubham-padia: Yes, much cleaner, thanks!

plugins/destination/gcs/client/client_test.go
shubham-padia: Yes!

shubham-padia:
@erezrokah Those changes look good, thanks! This should be good to merge!

erezrokah:
We'll release the new version https://github.com/cloudquery/cloudquery/pull/17611 on Tuesday as a part of our weekly cadence of releases

plugins/destination/gcs/client/client_test.go
shubham-padia: Opened #17612 , will open a PR for it when I get some time

PR #17605, opened 2024-04-12
fix: Require either `api_token` or `api_key,api_email` in schema

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery-issues/issues/1515 (internal issue). This aligns the JSON schema with the [plugin requirements](https://hub.cloudquery.io/plugins/source/cloudquery/cloudflare/latest/docs#configuration-cloudflare-spec) to have either `api_token` or `api_key` and `api_email` required.

This is not a breaking change since we only log a warning on schema failures.

Technically you can configure the plugin only using a `CLOUDFLARE_API_TOKEN` env variable or (`CLOUDFLARE_API_KEY, CLOUDFLARE_EMAIL`) but that's not the standard in our plugins and we should remove it


plugins/source/cloudflare/client/spec.go
```diff
@@ -21,5 +25,12 @@ func (s *Spec) SetDefaults() {
 	}
 }
 
+func (Spec) JSONSchemaExtend(sc *jsonschema.Schema) {
+	sc.OneOf = []*jsonschema.Schema{
+		{Required: []string{"api_token"}},
+		{Required: []string{"api_key", "api_email"}},
+	}
```
bbernays, line 32: Can we add a test of what happens if both `api_token` and `api_key, api_email` are specified?

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/17605/commits/13db042e0302a9a52943c8d5c30937a20fcea37b

PR #17660, opened 2024-04-15
fix!: Rename `aws_db_proxies` to `aws_rds_db_proxies`

bbernays:
#### Summary

Follow naming convention of `plugin_service_resource` to allow users to use `aws_rds*` to sync all RDS resources


erezrokah:
Does this require a follow up on our policies?

bbernays:
> Does this require a follow up on our policies?

I don't think so, I did a quick search and the table name didn't come up in the policies repo

Issue #13878, opened 2023-09-16
bug(sqlite): Missing linux arm64 binary for sqlite destination plugin


iainlbc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

May we please have an asset added for linux arm64 to the sqlite destination plugin releases? 

### Expected Behavior

Expected to find binary archive for arm64 linux

### CloudQuery (redacted) config

```
---
kind: destination
spec:
  registry: "local"
  name: "sqlite"
  path: "./sqlite"
  version: "v2.4.9"
  spec:
    connection_string: data/sqlite.db
```

### CloudQuery (redacted) logs

I was originally intending to use this plugin within a dagger pipeline, and got this error:
```
180: [1.59s] Error: failed to get plugin url: failed to find plugin cloudquery/sqlite version v2.4.9
```

Then I tried manually installing the sqlite plugin via a dockerfile, and it cannot find the arm64 binary:
```
...
FROM builder AS cloudquery-sqlite
ARG CLOUDQUERY_SQLITE_VERSION=2.4.9
RUN helper-curl tar "--strip-components=2 plugins/destination/sqlite" \
    https://github.com/cloudquery/cloudquery/releases/download/plugins-destination-sqlite-v${CLOUDQUERY_SQLITE_VERSION}/sqlite_linux_@GOARCH.zip
...
```

Returns:
```
65: [0.33s] curl: (22) The requested URL returned error: 404
65: [0.33s] URL not found: https://github.com/cloudquery/cloudquery/releases/download/plugins-destination-sqlite-v2.4.9/sqlite_linux_arm64.zip
```

### CloudQuery version

3.16.3

### Pull request (optional)

- [ ] I can submit a pull request

iainlbc:
It looks like this is explicitly ignored here: https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/sqlite/.goreleaser.yaml#L33-L38

erezrokah:
> It looks like this is explicitly ignored here: https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/sqlite/.goreleaser.yaml#L33-L38

We no longer use Go Releaser so https://github.com/cloudquery/cloudquery/blob/6342e0251db735ffbcb4e133456463410d9f9725/plugins/destination/sqlite/main.go#L23 is the code to update

candiduslynx:
Hi @iainlbc!
Could you test with v2.8.0?

[...]

PR #17673, opened 2024-04-17
feat: Publish `linux/arm64` build

candiduslynx:
Closes https://github.com/cloudquery/cloudquery/issues/13878


erezrokah (approved):

candiduslynx:
> I think there was a reason it was disabled

The package doesn't fail on my machine, so I suspect it's a workable solution.

We can always revert this change if there are any issues reported by users (currently there's no paper trail on why the cross compile to linux/arm64 was disabled).

Issue #8138, opened 2023-02-16
Partitioning by day in overwrite mode


yevgenypats:
Users brought up the need to partition by day (or any other interval) in overwrite mode. This will need rounding the `sync_time` and add it to the primary keys instead of dropping them all together.

hermanschaaf:
Nice idea! Bit of a nitpick, but I think this feature should probably be a configuration option for overwrite mode, rather than for append mode? Because you'd be overwriting / upserting, we're just adding a bit of additional logic to that.

bbernays:
Maybe we could start with a tutorial on how to do it manually? I think only specific versions of our destinations would be able to take advantage of this feature... For example partitioning in Postgres has only been viable in 13+ while we support 11+

bbernays:
Proposal:

- Add in a new destination attribute named `partition_time_period` where it takes an integer number of minutes where we will round the `sync_time` too
- If value is >0 then we should add the `sync_time` to the PK (even if `pk_mode` is set to `cq-id-only`)
- When value is >0 and `write_mode != overwrite` sync should fail with an error

erezrokah:
Initial implementation released in CLI version `v5.16.0`, see docs in https://docs.cloudquery.io/docs/reference/destination-spec#sync_group_id-preview

PR #17683, opened 2024-04-17
feat: Add sync group ID destination spec configuration

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/8138. Still needs docs and a bit more testing, opening for feedback.
Using the following configurations (assuming `gcp-1.yml` and `gcp-2.yml` run separately in parallel on a daily basis):

```yaml filename=gcp-1.yml
kind: source
spec:
  name: "gcp-cq-playground"
  path: "cloudquery/gcp"
  version: "v12.3.5"
  tables: ["gcp_bigquery_datasets"]
  destinations: ["postgresql"]
  spec:
    project_ids: [cq-playground]
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  registry: "cloudquery"
  version: "v8.0.3"
  external_sync_group_id: "{{YEAR}}-{{MONTH}}-{{DAY}}-{{HOUR}}"
  write_mode: "overwrite"
  spec:
```

```yaml filename=gcp-2.yml
kind: source
spec:
  name: "gcp-cq-integration-tests"
  path: "cloudquery/gcp"
  version: "v12.3.5"
  tables: ["gcp_bigquery_datasets"]
  destinations: ["postgresql"]
  spec:
    project_ids: [cq-integration-tests]
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v8.0.3"
  write_mode: "overwrite"
  external_sync_group_id: "{{YEAR}}-{{MONTH}}-{{DAY}}-{{HOUR}}"
  spec:
```

Results look like 
image

Users can also utilise env var replacement to inject their own value.
Still not sure if this should be a source or destination option.

### DIFF
--- a/cli/internal/specs/v0/destination.go
+++ b/cli/internal/specs/v0/destination.go
@@ -3,6 +3,16 @@ package specs
 import (
 	"bytes"
 	"encoding/json"
+	"strings"
+	"time"
+)
+
+const (
+	varYear   = "{{YEAR}}"
+	varMonth  = "{{MONTH}}"
+	varDay    = "{{DAY}}"
+	varHour   = "{{HOUR}}"
+	varMinute = "{{MINUTE}}"
 )
 
 // Destination plugin spec
@@ -17,6 +27,8 @@ type Destination struct {
 
 	// Destination plugin PK mode
 	PKMode PKMode `json:"pk_mode,omitempty" jsonschema:"default=default"`
+
+	ExternalSyncId string `json:"external_sync_id,omitempty"`
 
 	// Destination plugin own (nested) spec
 	Spec map[string]any `json:"spec,omitempty"`
@@ -47,4 +59,13 @@ func (d *Destination) UnmarshalSpec(out any) error {
 
 func (d *Destination) Validate() error {
 	return d.Metadata.Validate()
+}
+
+func (s *Destination) RenderedExternalSyncId(t time.Time) string {
+	renderedValue := strings.ReplaceAll(s.ExternalSyncId, varYear, t.Format("2006"))
+	renderedValue = strings.ReplaceAll(renderedValue, varMonth, t.Format("01"))
+	renderedValue = strings.ReplaceAll(renderedValue, varDay, t.Format("02"))
+	renderedValue = strings.ReplaceAll(renderedValue, varHour, t.Format("15"))
+	renderedValue = strings.ReplaceAll(renderedValue, varMinute, t.Format("04"))
+	return renderedValue
 }
--- a/cli/internal/specs/v0/destination.go
+++ b/cli/internal/specs/v0/destination.go
@@ -61,8 +61,8 @@ func (d *Destination) Validate() error {
 	return d.Metadata.Validate()
 }
 
-func (s *Destination) RenderedExternalSyncGroupId(t time.Time) string {
-	renderedValue := strings.ReplaceAll(s.ExternalSyncGroupId, varYear, t.Format("2006"))
+func (d *Destination) RenderedExternalSyncGroupId(t time.Time) string {
+	renderedValue := strings.ReplaceAll(d.ExternalSyncGroupId, varYear, t.Format("2006"))
 	renderedValue = strings.ReplaceAll(renderedValue, varMonth, t.Format("01"))
 	renderedValue = strings.ReplaceAll(renderedValue, varDay, t.Format("02"))
 	renderedValue = strings.ReplaceAll(renderedValue, varHour, t.Format("15"))
--- a/cli/internal/specs/v0/destination.go
+++ b/cli/internal/specs/v0/destination.go
@@ -28,7 +28,7 @@ type Destination struct {
 	// Destination plugin PK mode
 	PKMode PKMode `json:"pk_mode,omitempty" jsonschema:"default=default"`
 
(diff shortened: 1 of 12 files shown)


disq:
Why not call it a `PartitionKey`?

cli/internal/transformer/transformer.go
```diff
@@ -10,9 +10,10 @@ import (
 )
 
 const (
-	cqSyncTime     = "_cq_sync_time"
-	cqSourceName   = "_cq_source_name"
-	cqIDColumnName = "_cq_id"
+	cqSyncTime       = "_cq_sync_time"
+	cqSourceName     = "_cq_source_name"
+	cqIDColumnName   = "_cq_id"
+	cqExternalSyncId = "_cq_external_sync_id"
```
bbernays, line 16: I get caught up on the name, because this ID is not designed to identify a single sync it is designed to identify a group of syncs so maybe something like `_cq_sync_group_id`?

bbernays:
> Why not call it a `PartitionKey`?

I don't think we should mix in DB terminology that we are not prepared to integrate with. We call columns PKs because they are actual PrimaryKeys, I don't think we have any intention of actually managing partitions and partition keys in various DBs

cli/internal/transformer/transformer.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/17683/commits/05f5c40aa8fe693119f8d4c94da19e2b7c35e532

bbernays:
How is this intended to be used with regards to write mode?
-`append` This will just add a column (all PKs are removed)
- `overwrite` - This will retain a record for each resource in each sync group
- `overwrite-delete-stale`- Deletion happens based on the sync config name, so any historical data will be deleted


Because of the inconstant behavior maybe we should not allow `_cq_sync_group_id` to be used with a destination using `overwrite-delete-stale`?

cli/internal/specs/v0/destination.go
```diff
@@ -18,6 +28,8 @@ type Destination struct {
 	// Destination plugin PK mode
 	PKMode PKMode `json:"pk_mode,omitempty" jsonschema:"default=default"`
 
+	ExternalSyncGroupId string `json:"external_sync_group_id,omitempty"`
```
hermanschaaf, line 31: I don't really mind, but why have `external` in the name, and not call it `sync_group_id`?

hermanschaaf:
> How is this intended to be used with regards to write mode?

`overwrite-delete-stale`: Whereas normally we would truncate only on `_cq_source_name`, when `_cq_sync_group_id ` is present this mode should truncate based on a combination of (`_cq_source_name`, `_cq_sync_group_id`), removing all matching entries with smaller `cq_sync_time`.

This would allow you to safely re-run a sync within a group multiple times without getting stale records in that group. I think this is important.

It will, however, have the side-effect that it will keep past sync groups indefinitely. Maybe we need to think about a way to enable cleaning those up?

cli/internal/specs/v0/destination.go
erezrokah: I was trying to be specific with the fact it's an "external" value, I'll rename it

erezrokah:
> Because of the inconstant behavior maybe we should not allow `_cq_sync_group_id` to be used with a destination using `overwrite-delete-stale`?

Yeah I'm thinking the same thing. It's not very useful for `overwrite-delete-stale` unless we change the truncate query like @hermanschaaf suggested, which means updating all the plugins that support `overwrite-delete-stale`.

Somehow this reminds of what https://github.com/cloudquery/plugin-sdk/pull/1141 is trying to solve.

I'll take a look a bit on `overwrite-delete-stale` and see what we can do there

erezrokah:
Added docs, tests and erroring out on using with `overwrite-delete-stale`

website/pages/docs/reference/destination-spec.md
```diff
@@ -101,6 +101,21 @@ Specifies the Primary Keys that the destination will configure when using the `o
 
 Supported by destination plugins released on 2023-03-21 and later
 
+<!-- vale off -->
+
+### sync_group_id
```
disq, line 106: Maybe mark this as a preview feature?

disq (approved):

yevgenypats:
> > This would allow you to safely re-run a sync within a group multiple times without getting stale records in that group. I think this is important.
> > It will, however, have the side-effect that it will keep past sync groups indefinitely. Maybe we need to think about a way to enable cleaning those up?
> 
> Since this requires a bit more design an additional work (in our gRPC protocol, SDK, destinations and CLI) how about we do it in a follow up and after getting feedback for this PR?
> 
> The use case we're trying to solve is `Correlating historical data from multiple concurrent syncs`, I think with this PR, assuming you run syncs on a cadence (e.g. daily), using either `overwrite` or `append` that should work with the tradeoffs:
> 
> 1. In `overwrite` if a re-run a sync within the group you might have stale data for that partition (which you can later clean up using sync name, time and partition)
> 2. In `append` if a re-run a sync within the group you'll have duplicated that for that partition (which you can later clean up using sync name, time and partition)
> 
> I agree having it for `overwrite-delete-stale` is useful, suggesting we could split up the work into 2 iterations

sounds good to me

Issue #13287, opened 2023-08-22
bug: ARN Created with local AccountID Rather than OwnerID


bbernays:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently a number of ARN resolvers are using the `client.AccountID` to populate the accountID portion of the ARN. This means that if a resource is owned by another and shared, that the ARN will be wrong.

impacted functions are:

- `resolveHostArn`
- `resolveEbsSnapshotArn`
- `resolveInternetGatewayArn`
- `resolveNetworkAclArn`
- `resolveNetworkInterfaceArn`
- `resolveRouteTableArn`
- `resolveSecurityGroupArn`
- `resolveVpcEndpointServiceArn`
- `resolveVpcEndpointArn`
- `resolveVpcArn`
- `resolveSnapshotARN`
- `resolveDocumentARN`

### Expected Behavior

instead of `client.AccountID` the `arn` should use the `OwnerID` field from the API response

### CloudQuery (redacted) config

NA

### Steps To Reproduce

NA

### CloudQuery (redacted) logs

NA

### CloudQuery version

NA

### Pull request (optional)

- [X] I can submit a pull request

candiduslynx:
`resolveVpcEndpointServiceArn` is nowhere to be found, others are addressed in the linked PRs

PR #17686, opened 2024-04-18
fix: Use owner account ID when resolving resource ARN

candiduslynx:
Part of https://github.com/cloudquery/cloudquery/issues/13287

BEGIN_COMMIT_OVERRIDE
fix: Use owner account ID when resolving resource ARN (https://github.com/cloudquery/cloudquery/pull/17686). The following tables are affected:
* `aws_ec2_hosts`
* `aws_ec2_internet_gateways`
* `aws_ec2_network_acls`
* `aws_ec2_network_interfaces`
* `aws_ec2_route_tables`
* `aws_ec2_security_groups`
* `aws_ec2_vpc_endpoints`
* `aws_ec2_vpcs`

END_COMMIT_OVERRIDE


plugins/source/aws/resources/services/ec2/network_interfaces.go
```diff
@@ -63,7 +63,7 @@ func resolveNetworkInterfaceArn(_ context.Context, meta schema.ClientMeta, resou
 		Partition: cl.Partition,
 		Service:   "ec2",
 		Region:    cl.Region,
-		AccountID: cl.AccountID,
+		AccountID: aws.ToString(item.OwnerId),
```
bbernays, line 66: Now that the `AccountID` is no longer guarenteed to be the same as `cl.AccountID` doesn't that mean for all of these resources we should change the `client.DefaultAccountIDColumn(false)` to be `true`?

candiduslynx: Yeah, makes sense

candiduslynx: acecb5c4d9cfce63945f5620c915fe9425c6dd9b

bbernays (approved):

PR #17709, opened 2024-04-19
Fix AWS_SHARED_CREDENTIALS configurations

ei-grad:
For some reason setting the region here breaks configurations which use AWS_SHARED_CREDENTIALS with a session credentials for MFA or assumed role workflow.

#### Summary


bbernays:
Hi @ei-grad - This change will definitely break many users syncs. Can you open up an issue with the details about what is not working for your setup so we can find a less dramatic solution?

erezrokah:
Going to close this PR. @ei-grad Please open an issue first so we can discuss the change

PR #17766, opened 2024-04-23
feat: Add support for `{{TABLE_HYPHEN}}` Path variable for S3 Destination Plugin

VictorCovalski:
### Summary
Creates a path variable named `{{TABLE_HYPHEN}}`, which outputs the table name separated by hyphens instead of undescores.


erezrokah (commented):
Thanks for the PR @VictorCovalski 🚀  I wonder if we should take a more generic approach for transforming template variables values, similar to https://mozilla.github.io/nunjucks/templating.html#filters

For example to hyphen the table name one would use `{{TABLE | toSnake }}`

I agree your PR is the most consistent with our current implementation, just trying to figure out if we can implement this in a way that benefits more uses case

erezrokah:
The test failure is because the PR is from a fork and our CI needs credentials to access an S3 bucket we use for testing. You shouldn't worry about that. As for my previous [comment](https://github.com/cloudquery/cloudquery/pull/17766#pullrequestreview-2016689729) no need to change anything yet, we're discussing this internally

erezrokah:
Also do you mind sharing the reason you'd like to have the name in hyphens?

plugins/destination/s3/docs/overview.md
```diff
@@ -37,6 +37,7 @@ This is the (nested) spec used by the CSV destination Plugin.
   The path supports the following placeholder variables:
 
   - `{{TABLE}}` will be replaced with the table name
+  - `{{TABLE_HYPHEN}}` will be replaced with the table name with hyphens instead of undescores.
```
candiduslynx, line 40:
```suggestion
  - `{{TABLE_HYPHEN}}` will be replaced with the table name with hyphens instead of underscores.
```

candiduslynx (dismissed):

VictorCovalski:
> Also do you mind sharing the reason you'd like to have the name in hyphens?

Sure! We're testing a integration between a internal service and CloudQuery, this service needs to have the S3 path with the table name with only hyphens and not underscores.
Right now we're solving this by setting the path like this `aws-securityhub-findings/{{YEAR}}-{{MONTH}}-{{DAY}}/{{TABLE}}-{{UUID}}.parquet` but we don't want to create and maintain a configuration file for each table.

erezrokah (approved):

candiduslynx:
/gen sha=9f06350405f42e41c4ac227ff4ec20e5007656dc dir=plugins/destination/s3

erezrokah:
Released in `v6.1.0` https://hub.cloudquery.io/plugins/destination/cloudquery/s3/v6.1.0/versions

PR #17782, opened 2024-04-25
fix: Append multiple system values at once

candiduslynx:


erezrokah (commented):
I'm having a hard time understanding if this is a `fix` or a `refactor`.
Can you explain the user visible impact of this fix?
e.g. if a user runs a sync with/without this change what would they see differently?

candiduslynx:
> I'm having a hard time understanding if this is a `fix` or a `refactor`. Can you explain the user visible impact of this fix? e.g. if a user runs a sync with/without this change what would they see differently?

I think this may very well be a `refactor` (I'm renaming the PR for this).
The users of the sources capable of multiple rows in a single record (pg?) will see significant improvement for the time consumed appending these values, but it will not have an impact for the simple single-row-per-record source plugins.

erezrokah:
> will see significant improvement for the time consumed appending these values

What would be the improvement? Do we have benchmarks?
Maybe we can group all the related multi row work under a top level issue? It seems we have a bunch of related fixes so we'll need to prioritize them with other work

candiduslynx:
After benchmarking it seems not a lot of diff in speed is observed, so made the diff smaller: 7b7680cde3f68a373311f5a18bcc825b23f3dc71

cli/internal/transformer/transformer.go
```diff
@@ -153,9 +154,8 @@ func (t *RecordTransformer) Transform(record arrow.Record) arrow.Record {
 		cols = append(cols, syncGroupIdBldr.NewArray())
 	}
 
-	for i := range sc.Fields() {
-		cols = append(cols, record.Column(i))
-	}
+	cols = cols[:len(sc.Fields())+t.internalColumns] // resize back as we have the capacity
+	copy(cols[t.internalColumns:], record.Columns())
```
erezrokah, line 158: I might be missing something but we have 3 conditional `append` to `cols` a bit before (also inside `TransformSchema`)
What happens if `t.internalColumns` is `3` for example but we didn't do any `append`.
Won't that mean that we'll have misalignment in the data?

I believe that can happen when someone is syncing to a Postgres DB destination, than using our Postgres source on the synced tables

erezrokah (commented):
Added a couple of comments

cli/internal/transformer/transformer.go
erezrokah: Regardless it would be helpful to understand how the refactor improves the code

cli/internal/transformer/transformer.go
```diff
@@ -130,11 +130,12 @@ func (t *RecordTransformer) Transform(record arrow.Record) arrow.Record {
 	newSchema := t.TransformSchema(sc)
 	nRows := int(record.NumRows())
 
-	cols := make([]arrow.Array, 0, len(sc.Fields())+t.internalColumns)
+	cols := make([]arrow.Array, 0, len(sc.Fields())+t.internalColumns) // alloc together with the proper capacity
 	if t.withSyncTime && !sc.HasField(cqSyncTime) {
+		ts, _ := arrow.TimestampFromTime(t.syncTime, arrow.Microsecond)
 		syncTimeBldr := array.NewTimestampBuilder(memory.DefaultAllocator, &arrow.TimestampType{Unit: arrow.Microsecond, TimeZone: "UTC"})
 		for i := 0; i < nRows; i++ {
-			syncTimeBldr.AppendTime(t.syncTime)
```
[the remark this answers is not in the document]
candiduslynx: Convert the value once instead of every time

cli/internal/transformer/transformer.go
candiduslynx: Ah, I see the issue with the copy here (we may skip append if the data has the col).
I'll fix it with a single append then

cli/internal/transformer/transformer.go
```diff
@@ -130,11 +130,12 @@ func (t *RecordTransformer) Transform(record arrow.Record) arrow.Record {
 	newSchema := t.TransformSchema(sc)
 	nRows := int(record.NumRows())
 
-	cols := make([]arrow.Array, 0, len(sc.Fields())+t.internalColumns)
+	cols := make([]arrow.Array, 0, len(sc.Fields())+t.internalColumns) // alloc together with the proper capacity
```
candiduslynx, line 133:
```suggestion
	cols := make([]arrow.Array, 0, len(sc.Fields())+t.internalColumns)
```

cli/internal/transformer/transformer.go
erezrokah: > Internal columns depends on the amount of added columns.

`internalColumns` is incremented when the option is configured https://github.com/cloudquery/cloudquery/blob/881902078f8293ac5cb374e48c91fddd60b3ab4d/cli/internal/transformer/transformer.go#L38

`cols` is appended based on a slightly different condition https://github.com/cloudquery/cloudquery/blob/881902078f8293ac5cb374e48c91fddd60b3ab4d/cli/internal/transformer/transformer.go#L148

erezrokah (approved):

PR #17791, opened 2024-04-29
ci: Verify `cli/cmd` tests

candiduslynx:


erezrokah:
The issue was resolved via a backend fix to our Hub

PR #17842, opened 2024-05-01
feat: Sync Metadata table

bbernays:


cli/cmd/summary.go
```diff
@@ -5,22 +5,33 @@ import (
...
+	DestinationVersion  string    `json:"destination_version"`
+	DestinationWarnings uint64    `json:"destination_warnings"`
+	Resources           uint64    `json:"resources"`
+	SourceErrors        uint64    `json:"source_errors"`
+	SourceName          string    `json:"source_name"`
+	SourcePath          string    `json:"source_path"`
+	SourceVersion       string    `json:"source_version"`
+	SourceWarnings      uint64    `json:"source_warnings"`
+	SyncID              string    `json:"sync_id"`
+	SyncStartTime       time.Time `json:"sync_start_time"`
```
hermanschaaf, line 34: I think let's call this `sync_time` so that the relationship with `_cq_sync_time` is less ambiguous

cli/cmd/summary.go
```diff
@@ -5,22 +5,33 @@ import (
...
-	DestinationVersion  string `json:"destination_version"`
-	DestinationWarnings uint64 `json:"destination_warnings"`
-	Resources           uint64 `json:"resources"`
-	SourceErrors        uint64 `json:"source_errors"`
-	SourceName          string `json:"source_name"`
-	SourcePath          string `json:"source_path"`
-	SourceVersion       string `json:"source_version"`
-	SourceWarnings      uint64 `json:"source_warnings"`
-	SyncID              string `json:"sync_id"`
+	CliVersion          string    `json:"cli_version"`
```
hermanschaaf, line 21:
```suggestion
	CLIVersion          string    `json:"cli_version"`
```

cli/cmd/summary.go
hermanschaaf: Or we could even call it `cq_sync_time` and `cq_sync_id` to be totally clear

hermanschaaf:
I tried it out locally, but got an error:

```
{"level":"info","module":"cli","invocation-id":"a8834301-9983-43f4-94fc-e7898d17d5a1","address":"/var/folders/_c/hm9lkvfj7fz_rykn71rmr8gh0000gn/T/cq-gQebwPMMfbrFKgWU.sock","message":"Got stop signal. Plugin server shutting down","signal":"interrupt","time":"2024-05-02T11:03:45Z"}
Error: failed to sync v3 source k8s: failed to send write request (migrate sync summary table): rpc error: code = Internal desc = SendMsg called after CloseSend
{"level":"error","module":"cli","invocation-id":"a8834301-9983-43f4-94fc-e7898d17d5a1","error":"failed to sync v3 source k8s: failed to send write request (migrate sync summary table): rpc error: code = Internal desc = SendMsg called after CloseSend","time":"2024-05-02T11:03:45Z","message":"exiting with error"}
```

cli/cmd/summary.go
```diff
@@ -62,3 +73,86 @@ func checkFilePath(filename string) error {
 	dirPath := filepath.Dir(filename)
 	return os.MkdirAll(dirPath, 0755)
 }
+
+func generateSummaryTable() *schema.Table {
+	tableName := "cloudquery_sync_summary"
```
hermanschaaf, line 78: `cloudquery_sync_summary` -> `cloudquery_sync_summaries`, for consistency with our resource naming convention? 😄

cli/cmd/summary.go
```diff
@@ -62,3 +73,86 @@ func checkFilePath(filename string) error {
...
+	t := schema.Tables{{
+		Name:      tableName,
+		Transform: transformers.TransformWithStruct(&syncSummary{}),
```
hermanschaaf, line 81: It would be nice if we could ensure that zero values for warnings and errors show up as zero, rather than null, in the table

hermanschaaf: I think setting it to `Required: true` should fix it, but haven't tried yet

hermanschaaf:
> I tried it out locally, but got an error:

☝️ I pushed a fix for this

cli/cmd/summary.go
hermanschaaf: Nevermind, I see this is because we use an automated mapping from field name to column name

cli/cmd/summary.go
hermanschaaf: I pushed an update for this as well

savme (approved):
Works great 🚀

cli/cmd/summary.go
```diff
@@ -62,3 +73,87 @@ func checkFilePath(filename string) error {
...
+	tableName := "cloudquery_sync_summaries"
+	t := schema.Tables{{
+		Name: tableName,
+		Transform: transformers.TransformWithStruct(
+			&syncSummary{},
+		),
+		Columns: []schema.Column{},
+	}}
+	if err := transformers.TransformTables(t); err != nil {
+		panic(err)
```
erezrokah, line 87: Can we return an error instead of panicking?

cli/cmd/summary.go
```diff
@@ -62,3 +73,87 @@ func checkFilePath(filename string) error {
...
+
+		transformedSchema := destinationTransformers[i].TransformSchema(summaryTableSchema)
+		transformedSchemaBytes, err := plugin.SchemaToBytes(transformedSchema)
+		if err != nil {
+			return err
+		}
+		wr := &plugin.Write_Request{}
+
+		// Respect the noMigrate flag
+		if !noMigrate {
```
erezrokah, line 110: Since we support `noMigrate` shouldn't we also call `generateSummaryTable` inside `migrate_v3.go`?

cli/cmd/summary.go
erezrokah, line 21: Referring @hermanschaaf's comment could we make this `CLIVersion` and then add an exception in the caser below?

cli/cmd/summary.go
```diff
@@ -62,3 +73,87 @@ func checkFilePath(filename string) error {
...
+	}
+	return t[0]
+}
+
+func sendSummary(writeClients []plugin.Plugin_WriteClient, destinationSpecs []specs.Destination, destinationsClients []*managedplugin.Client, destinationTransformers []*transformer.RecordTransformer, summary *syncSummary, noMigrate bool) error {
+	summaryTable := generateSummaryTable()
+	summaryTableSchema := summaryTable.ToArrowSchema()
+
+	defaultCaser := caser.New()
+	for i := range destinationsClients {
```
erezrokah, line 97: So this will send all summaries to all configured destinations correct?
Does it make sense to have it multiple destinations? Or a destination should have only it's "own" summary?

bbernays: I think it is implemented that a destination will only get the information about its self. In the for loop where we send the data to the destination, we set all of the destination specific information so it only ever gets information about itself.

If there is something I am missing let me know!

cli/cmd/sync_test.go
```diff
@@ -125,7 +125,8 @@ func TestSync(t *testing.T) {
 			if len(tc.summary) > 0 {
 				summaries := readSummaries(t, summaryPath)
 				// have to ignore SyncID because it's random and plugin versions since we update those frequently using an automated process
-				diff := cmp.Diff(tc.summary, summaries, cmpopts.IgnoreFields(syncSummary{}, "SyncID", "DestinationVersion", "SourceVersion"))
+				// also ignore SyncTime because it's a timestamp
+				diff := cmp.Diff(tc.summary, summaries, cmpopts.IgnoreFields(syncSummary{}, "SyncID", "DestinationVersion", "SourceVersion", "SyncTime"))
```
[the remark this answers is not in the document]
bbernays: Done in 7f1efbc

cli/cmd/summary.go
bbernays: Done in b12c0d9

cli/cmd/summary.go
bbernays: Done in [960a887](https://github.com/cloudquery/cloudquery/pull/17842/commits/960a8873f08ee5e4d4586adf1d648b57ae23d58c)

cli/cmd/migrate_v3.go
```diff
@@ -101,6 +101,10 @@ func migrateConnectionV3(ctx context.Context, sourceClient *managedplugin.Client
 				return handleSendError(err, writeClients[i], "migrate")
 			}
 		}
+
+		if err := migrateSummaryTable(writeClients[i], destinationTransformers[i], destinationSpecs[i]); err != nil {
```
erezrokah, line 105: Shouldn't we check the `send_sync_summary` option is enabled first?

cli/cmd/summary.go
```diff
@@ -62,3 +73,100 @@ func checkFilePath(filename string) error {
...
+	return t[0], nil
+}
+
+func migrateSummaryTable(writeClient plugin.Plugin_WriteClient, destTransformer *transformer.RecordTransformer, spec specs.Destination) error {
+	summaryTable, err := generateSummaryTable()
+	if err != nil {
+		return err
+	}
+	summaryTableSchema := summaryTable.ToArrowSchema()
+	transformedSchema := destTransformer.TransformSchema(summaryTableSchema)
```
erezrokah, line 98: Using `destTransformer` here generates `_cq_sync_time` and `_cq_source_name` columns 
image

erezrokah: Probably will generate `_cq_sync_group_id` is enabled by the config

cli/cmd/summary.go
```diff
@@ -5,22 +5,33 @@ import (
...
+	DestinationVersion  string    `json:"destination_version"`
+	DestinationWarnings uint64    `json:"destination_warnings"`
+	Resources           uint64    `json:"resources"`
+	SourceErrors        uint64    `json:"source_errors"`
+	SourceName          string    `json:"source_name"`
+	SourcePath          string    `json:"source_path"`
+	SourceVersion       string    `json:"source_version"`
+	SourceWarnings      uint64    `json:"source_warnings"`
+	SyncID              string    `json:"cq_sync_id"`
+	SyncTime            time.Time `json:"cq_sync_time"`
```
erezrokah, line 34:
```suggestion
	SyncID              string    `json:"sync_id"`
	SyncTime            time.Time `json:"sync_time"`
```

Unrelated to this PR and non blocking but maybe we should drop the `cq_` prefix as we don't use it anywhere in this struct, nor it's needed if the data goes into a dedicated file/table

cli/cmd/summary.go
erezrokah: Ok you're right each destination will get its own information, but since [we're looping over the summaries](https://github.com/cloudquery/cloudquery/pull/17842/files#diff-ee974bad39e59893cee87960039e8c360d83713c82eefb050fdb5eb4ea22d47dR380) each summary will be sent multiple times to the same destination, for example with the following configuration:

```yaml
kind: source
spec:
  name: "gcp"
  path: "cloudquery/gcp"
  registry: "cloudquery"
  version: "v12.7.0"
  tables: ["gcp_storage_buckets"]
  spec:
    project_ids: ["cq-playground"]
  destinations: ["file", "postgresql"]
---
kind: destination
spec:
  name: "file"
  path: "cloudquery/file"
  version: "v5.0.0"
  write_mode: "append"
  send_sync_summary: true
  spec:
    path: "examples/files/{{TABLE}}/{{UUID}}.{{FORMAT}}"
    format: "json"
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v8.0.4"
  send_sync_summary: true
  spec:
```

I got 2 files with the exact same summary under `examples/files/cloudquery_sync_summaries` and two entries in the Postgres summary table:
image

erezrokah: Also looks like `cli_version`, `cq_sync_id` and `cq_sync_time` are missing

cli/cmd/sync_v3.go
```diff
@@ -375,21 +362,41 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 			SourceErrors:        totals.Errors,
 			SourceWarnings:      totals.Warnings,
 			SyncID:              uid,
+			SyncTime:            syncTime,
```
erezrokah, line 365: Noticing this now but it seems there's a bug with `SourceErrors` and `SourceWarnings`. We add `m.Warnings` and `m.Errors` (the destination metrics) a bit before to `totals.Warnings` and `totals.Errors`, so we each loop accumulates all the warnings and errors

bbernays: Fixed in separate pr: https://github.com/cloudquery/cloudquery/pull/17914

cli/cmd/migrate_v3.go
bbernays: Because `send_sync_summary` is on each destination we would have to iterate through all of the destinations. The first part of the for loop where we iterate through the destination we do check if it is enabled though: https://github.com/cloudquery/cloudquery/blob/97babd42ea9edb27b874d3a30a51cd5e11bdf7e5/cli/cmd/summary.go#L122-L124

erezrokah: > Because `send_sync_summary` is on each destination we would have to iterate through all of the destinations. The first part of the for loop where we iterate through the destination we do check if it is enabled though:

In this file (`cli/cmd/migrate_v3.go`) we don't call `send_sync_summary`, but call `migrateSummaryTable` directly so I don't see where we check the config option.
You can test by running `cloudquery migrate` with the following config:

```yaml
kind: source
spec:
  name: "gcp"
  path: "cloudquery/gcp"
  registry: "cloudquery"
  version: "v12.7.0"
  tables: ["gcp_storage_buckets"]
  spec:
    project_ids: ["cq-playground"]
  destinations: ["postgresql"]
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  version: "v8.0.4"
  spec:
```

And see that the summary table is created even when the option is disabled

bbernays: Nice catch fixed in [4fdcd88](https://github.com/cloudquery/cloudquery/pull/17842/commits/4fdcd8881f685c4123983adca14f344da77babc3)

cli/cmd/summary.go
```diff
@@ -62,3 +73,109 @@ func checkFilePath(filename string) error {
...
+		return err
+	}
+	defaultCaser := caser.New()
+	for i := range destinationsClients {
+		if !destinationSpecs[i].SyncSummary {
+			continue
+		}
+
+		// Only send the summary to the destination that matches the current destination
+		if destinationSpecs[i].Name != summary.DestinationName || destinationSpecs[i].Version != summary.DestinationVersion || destinationSpecs[i].Path != summary.DestinationPath {
```
erezrokah, line 131: This fix works, but maybe we can drop the looping over destinations and pass a single destination instead?

cli/cmd/summary.go
bbernays: Fixed

cli/cmd/summary.go
bbernays: I think that is a good thing, to keep consistency with all of our other tables

cli/cmd/sync_v3.go
```diff
@@ -345,53 +345,66 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
 			DestinationVersion:  destinationSpecs[i].Version,
 			DestinationPath:     destinationSpecs[i].Path,
 		}
 	}
-	err = persistSummary(summaryLocation, syncSummaries)
-	if err != nil {
+	if err := persistSummary(summaryLocation, syncSummaries); err != nil {
 		log.Warn().Err(err).Msg("Failed to persist sync summary")
 	}
+	for _, summary := range syncSummaries {
```
erezrokah, line 379: I might be missing something here, but what's the reason we need 2 loops?
Can't we have one loop over `destinationsClients`:
```
  if !destinationSpecs[i].SyncSummary {
    continue
  }
  summary := syncSummaries[i]
  if err := sendSummary(writeClients[i], destinationSpecs[i], destinationsClients[i], destinationTransformers[i], &summary, noMigrate); err != nil {
    return err
  }
}
```

cli/cmd/sync_v3.go
```diff
@@ -345,53 +345,66 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
+		log.Info().Interface("summary", summary).Msg("Sync summary")
+		for i := range destinationsClients {
+			if !destinationSpecs[i].SyncSummary {
+				continue
+			}
+			// Only send the summary to the destination that matches the current destination
+			if destinationSpecs[i].Name != summary.DestinationName || destinationSpecs[i].Version != summary.DestinationVersion || destinationSpecs[i].Path != summary.DestinationPath {
+				continue
+			}
+			if err := sendSummary(writeClients[i], destinationSpecs[i], destinationsClients[i], destinationTransformers[i], &summary, noMigrate); err != nil {
```
erezrokah, line 389: Let's say we have an error sending the summary to the first destination? Shouldn't we try sending the summary for the next ones?

Maybe we can collect all errors via `errors.Join` then report the ones that failed

cli/cmd/summary.go
```diff
@@ -62,3 +73,104 @@ func checkFilePath(filename string) error {
...
+	if err != nil {
+		return err
+	}
+
+	csr := caser.New(caser.WithCustomInitialisms(map[string]bool{"CLI": true}), caser.WithCustomExceptions(map[string]string{"cli": "CLI"}))
+
+	if !destinationSpec.SyncSummary {
+		return nil
+	}
+
```
erezrokah, line 131:
```suggestion
```

Not sure we need this check here as we do it outside `sendSummary`

cli/cmd/sync_v3.go
```diff
@@ -345,52 +345,67 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
+		}
+
+		log.Info().Interface("summary", summary).Msg("Sync summary")
+		if !destinationSpecs[i].SyncSummary {
+			continue
+		}
+		// Only send the summary to the destination that matches the current destination
+		if destinationSpecs[i].Name != summary.DestinationName || destinationSpecs[i].Version != summary.DestinationVersion || destinationSpecs[i].Path != summary.DestinationPath {
+			continue
+		}
```
erezrokah, line 387:
```suggestion
```

Do we need this condition? Aren't we setting the summary fields based on the destination a bit before?

cli/cmd/sync_v3.go
```diff
@@ -345,52 +345,67 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
+		if err := sendSummary(writeClients[i], destinationSpecs[i], destinationsClients[i], destinationTransformers[i], &summary, noMigrate); err != nil {
+			metadataDataErrors = errors.Join(metadataDataErrors, err)
+		}
 	}
-	err = persistSummary(summaryLocation, syncSummaries)
-	if err != nil {
-		log.Warn().Err(err).Msg("Failed to persist sync summary")
+	// a single error for one of the destination should not stop all of the other table from being written
```
erezrokah, line 392:
```suggestion
```

I would say remove this comment or move it above `metadataDataErrors = errors.Join(metadataDataErrors, err)`

erezrokah (approved):

erezrokah:
@bbernays I removed `automerge` as we should add some docs for this

PR #17843, opened 2024-05-01
feat!: Store timestamps as `timestamp_tz`

bbernays:
#### Summary

Currently we are storing all timestamps as `timestamp_ntz` this is "wall clock" time meaning that it has no timezone associated with it. the `arrow.TimestampType` is:

```
TimestampType is encoded as a 64-bit signed integer since the UNIX epoch (2017-01-01T00:00:00Z). The zero-value is a second and time zone neutral. Time zone neutral can be considered UTC without having "UTC" as a time zone
```

Which makes it a better fit for users to be able to utilize the data in their queries.


hermanschaaf:
@bbernays Tests are failing on this PR, just FYI

bbernays:
The tests are failing because of issues with timezone comparison... In my local machine if I add `&timezone=UTC` to the connection string it works... still trying to figure out if there is a workable solution or not

plugins/destination/snowflake/client/client.go
```diff
@@ -43,7 +43,7 @@ func New(_ context.Context, logger zerolog.Logger, spec []byte, _ plugin.NewClie
 	if err != nil {
 		return nil, err
 	}
-	db, err := sql.Open("snowflake", dsn+"&BINARY_INPUT_FORMAT=BASE64&BINARY_OUTPUT_FORMAT=BASE64")
```
erezrokah, line 46: What will happen if someone already has `timezone` configured in their DSN?

bbernays: Not sure, but can test it out

erezrokah: If we store times with a timezone what's the reason we need it?

bbernays: Its a “usability” issue in that the default timezone is `America/Los_Angeles` (or whatever the user sets for the account or session)… This means that if a user compares `current_timestamp()` to `sync_time` (or any other timestamp synced by CQ) then they need to deal with casting and altering timezones of timestamps rather than just direct comparison. If we switch to to `timezone_tz` then all of the comparisons will just work out of the box

erezrokah: > If we switch to to `timezone_tz` then all of the comparisons will just work out of the box

I agree we need to switch to `timezone_tz`. What's the reason we need to set the timezone in the DSN? If we save the timezone diff in the the column already

bbernays: The `datetime` object that we actually send to the DB has no timezone information associated with it as the Arrow Datetime object specifically documents that it has no timezone but can be assumed to be UTC

erezrokah: I mean it shouldn't matter what's the timezone in the DSN as `timezone_tz` always save the data in diff from UTC format unless I misread the docs

erezrokah: > The `datetime` object that we actually send to the DB has no timezone information associated with it as the Arrow Datetime object specifically documents that it has no timezone but can be assumed to be UTC

Shouldn't we convert everything to UTC before insert then? Or by setting UTC on the DSN it causes the Go driver to do it for us?

bbernays: Exactly, by setting the session Timezone Snowflake uses that timezone when there isn't one specified by the datetime object

erezrokah: So we probably need to forbid setting it in the DSN? Or only set to UTC if not present so we still let users an option to configure this?

bbernays: I think we should forbid a user from setting any of the parameters we set in the DSN
`timezone`, `BINARY_INPUT_FORMAT`,  and`BINARY_OUTPUT_FORMAT`

erezrokah: OK so we can do it as a follow up to not block this PR

erezrokah (approved):

disq (approved):

plugins/destination/snowflake/client/client.go
bbernays: Issue to track the validation work: https://github.com/cloudquery/cloudquery/issues/17899

PR #17958, opened 2024-05-14
chore: Remove old integration pages

erezrokah:
#### Summary

Not sure why the Website build didn't fail in https://github.com/cloudquery/cloudquery/pull/17903, but it fails on `main`:

```
Error: No _configuration.md file found for source: firestore
--
15:37:38.719 | at <anonymous> (/vercel/path0/website/scripts/prebuild.tsx:166:17)
```

We shouldn't generate integration pages at all and redirect per https://github.com/cloudquery/cloudquery/pull/16681


website/components/pluginData.tsx
```diff
@@ -86,7 +86,7 @@ export const ALL_PLUGINS: Plugin[] = [
     logoDark: "/images/logos/plugins/firestore-dark.svg",
     id: "firestore",
     kind: "source",
-    availability: "free",
+    availability: "premium",
```
erezrokah, line 89: There's more cleanup to be done to remove any plugin data from the old Website, but didn't want to dig too deep into it

website/vercel.json
```diff
@@ -870,114 +870,6 @@
       "source": "/docs/plugins/destinations/sqlite/:path*",
       "destination": "https://hub.cloudquery.io/plugins/destination/cloudquery/sqlite"
     },
-    {
```
erezrokah, line 873: We have to remove the existing redirects otherwise https://github.com/cloudquery/cloudquery/pull/16681/files#diff-21378a71578c160a16aa86a2ccc6120c09d1d10a3deaf16af651a228bd0cfde4R1054 won't match

jackcloudquery (approved):

PR #17993, opened 2024-05-16
feat: Test connection using dedicated endpoint

disq:
Opening for comments.

A run with 2 plugin test connections (one success and one fail) and one plugin failure (not related to TestConnection)
(The postgresql plugin is an older public release with TestConnection not implemented, falling back on Init)

```
$ ./cli test-connection ./testsrc.yml ./gandi.yml 
Loading spec(s) from ./testsrc.yml, ./gandi.yml
Test Connection Results
source       gandi (cloudquery/gandi@v3.2.1)          Success
destination  pg@github (cloudquery/postgresql@v7.0.1) Failure: OTHER      rpc error: code = Internal desc = failed to init plugin: failed to initialize client: failed to get current database: failed to connect to `host=localhost user=postgres database=cq`: dial error (dial tcp 127.0.0.1:15432: connect: connection refused)
Error: at least one test connection failed
(exit code 1) $
```

With postgres implementing the test connector and a configuration with `host=badhost`

```
❯ ./cli test-connection test-connection.yaml
Loading spec(s) from test-connection.yaml
Test Connection Results
source       hackernews (cloudquery/hackernews@v3.1.8)  Success
destination  postgresql (grpc@localhost:7777)           Failure: INVALID_CREDENTIALS 	 no such host "badhost"
Error: at least one test connection failed
```


cli/go.mod
```diff
@@ -157,3 +157,7 @@ replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0
 
 // github.com/cloudquery/godebouncer @ fix-race
 replace github.com/vnteamopen/godebouncer => github.com/cloudquery/godebouncer v0.0.0-20240220152946-054f7dbbcf18
+
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/disq/plugin-sdk/v4 v4.0.0-20240516151118-4958828f53c9
+
+replace github.com/cloudquery/plugin-pb-go => github.com/cloudquery/plugin-pb-go v1.19.13-0.20240516105108-3fdf6ac15ea2
```
disq, line 163: To be removed before merge when `plugin-sdk` and `plugin-pb-go` are released.

erezrokah:
Can you share the specs you used for test? How can the test connection fail for the test plugin (`failed to test source test`)?
For the second spec not sure `host= badhost` is invalid credentials (credentials usually refer to user/pass), but that's a plugin issue

disq:
```yaml
# testsrc.yml
kind: source
spec:
  name: test
  version: v4.1.0
  path: cloudquery/test
  tables: ["*"]
  destinations: ["pg"]
  spec:
    num_clients: 1
    num_rows: 10
    num_sub_rows: 20000
    num_sub_cols: 200
```

```yaml
# gandi.yml
kind: source
spec:
  name: gandi
  path: cloudquery/gandi
  registry: cloudquery
  version: "v3.2.1"
  tables:
    - "*"
  destinations:
    - pg
  spec:
    api_key: my-valid-key
    endpoint_url: https://api.sandbox.gandi.net/
---
kind: destination
spec:
  name: "pg"
  path: "cloudquery/postgresql"
  registry: "github"
  version: "v7.0.1"
  spec:
    connection_string: "postgresql://postgres:pass@localhost:15432/cq"
```

disq:
@erezrokah The "failed to test source test" issue was fixed in https://github.com/cloudquery/cloudquery/pull/17993/commits/4477eb21193a47aa5c45e035bcf366604ff42a06 and the next cleanup commit. I just updated the PR description to reflect. It can still generate `failed to test source|destination %v: %w` type errors if for example it cannot communicate via gRPC with the plugin etc.

cli/cmd/test_connection_test.go
```diff
@@ -52,3 +53,25 @@ func TestTestConnection(t *testing.T) {
 		})
 	}
 }
+
+// errorsContainAny checks if any error in a slice contains at least one of the substrings.
+func errorsContainAny(errs []error, substrings []string) bool {
```
erezrokah, line 58: What's the benefit of these helpers instead of doing `require.ErrorContains` in a loop like before? Don't we want to test that the first substring appeared in the first error, and the second substring in the second error?

disq: We have multiple errors from `Unwrap() []error` so the previous code won't work. We can require the same count between `len(resultOfUnWrap)` and `len(want)` and try matching them in order, but that means the test expects a certain order in the results, which it currently doesn't.

erezrokah: Doesn't that mean that if cloudflare errors out but postgres not the test will still pass? Also should we check a better substring than the plugin name? Something like "test failed"?

disq: If the plugin name is not in the list of _errors_ then it's safe to assume the test succeeded. We could also embed the whole result in the failure error, let me do that so it's simpler.

disq: Refactored in https://github.com/cloudquery/cloudquery/pull/17993/commits/7991515a124020497ec7273ce834bf1a64e62aa4 , CLI output doesn't change.

cli/cmd/test_connection.go
```diff
@@ -291,3 +303,26 @@ func testPluginConnection(ctx context.Context, client plugin.PluginClient, spec
 		FailureDescription: resp.FailureDescription,
 	}, nil
 }
+
+// fetchFailedTestResults fetch the failed test results.
+//
+// The function returns any failed test results, or nil if all tests passed. The hackernews plugin is excluded from the
+// failed test results since it was previously used as a test case for the test connection command.
+func fetchFailedTestResults(results []testConnectionResult) (*testConnectionResult, error) {
```
disq, line 311: Maybe `filterFailedTestResults` instead?

cli/cmd/test_connection.go
```diff
@@ -291,3 +303,26 @@ func testPluginConnection(ctx context.Context, client plugin.PluginClient, spec
...
+	var failedResults []testConnectionResult
+
+	for _, result := range results {
+		if !result.Success && !strings.HasPrefix(result.pluginRef, "hackernews") {
```
disq, line 315: Should exclude the specific version of `hackernews` and `file` plugins if possible:
`cloudquery/file v4.0.4`
`cloudquery/hackernews v3.0.25`

cli/cmd/test_connection.go
```diff
@@ -171,34 +186,146 @@ func testConnection(cmd *cobra.Command, args []string) error {
...
+// filterFailedTestResults fetch the failed test results.
+//
+// The function returns any failed test results, or nil if all tests passed. The hackernews plugin is excluded from the
+// failed test results since it was previously used as a test case for the test connection command.
+func filterFailedTestResults(results []testConnectionResult) (*testConnectionResult, error) {
+	var failedResults []testConnectionResult
+
+	for _, result := range results {
+		if !result.Success {
+			if strings.Contains(result.pluginRef, "cloudquery/file v4.0.4") || strings.Contains(result.pluginRef, "cloudquery/hackernews v3.0.25") {
```
disq, line 316: It's in the format `cloudquery/gandi@v3.2.1` I think my comment was misleading.
```suggestion
			if strings.Contains(result.pluginRef, "cloudquery/file@v4.0.4") || strings.Contains(result.pluginRef, "cloudquery/hackernews@v3.0.25") {
```

cli/cmd/test_connection.go
```diff
@@ -171,34 +185,146 @@ func testConnection(cmd *cobra.Command, args []string) error {
...
+			maxLength = len(testResult.pluginRef)
+		}
+	}
+
+	connFailures := make([]testConnectionResult, 0, len(testConnectionResults))
+	for i, testResult := range testConnectionResults {
+		if i == 0 {
+			cmd.Println("Test Connection Results")
+		}
+		cmd.Printf("%-12s %s %s", testResult.pluginKind, testResult.pluginRef, strings.Repeat(" ", maxLength-len(testResult.pluginRef)+1))
```
erezrokah, line 235:
```suggestion
		fmt.Printf("%-12s %s %s", testResult.pluginKind, testResult.pluginRef, strings.Repeat(" ", maxLength-len(testResult.pluginRef)+1))
```

cli/cmd/test_connection.go
```diff
@@ -171,34 +185,146 @@ func testConnection(cmd *cobra.Command, args []string) error {
...
+		if testResult.Success {
+			cmd.Println("Success")
```
erezrokah, line 237:
```suggestion
			fmt.Println("Success")
```

Not saying it's better or worse, but this is how we do it in other places in the CLI

cli/cmd/test_connection.go
```diff
@@ -171,34 +185,146 @@ func testConnection(cmd *cobra.Command, args []string) error {
...
+			continue
+		}
+		cmd.Println("Failure:", testResult.FailureCode, "\t", testResult.FailureDescription)
```
erezrokah, line 240:
```suggestion
		fmt.Println("Failure:", testResult.FailureCode, "\t", testResult.FailureDescription)
```

cli/cmd/test_connection.go
```diff
@@ -171,34 +185,146 @@ func testConnection(cmd *cobra.Command, args []string) error {
...
 
 	return allErrors
 }
+
+type testConnectionFailures struct {
+	failed []testConnectionResult
+}
+
+func (*testConnectionFailures) Error() string {
+	// The errs are already shown in the console, so we don't need to show them again here
```
erezrokah, line 256: Are the errors shown when you do `--log-console` too?

cli/cmd/test_connection.go
```diff
@@ -171,34 +185,146 @@ func testConnection(cmd *cobra.Command, args []string) error {
...
+		FailureCode:        resp.FailureCode,
+		FailureDescription: resp.FailureDescription,
+	}, nil
+}
+
+// filterFailedTestResults fetch the failed test results.
+//
+// The function returns any failed test results, or nil if all tests passed. The hackernews plugin is excluded from the
+// failed test results since it was previously used as a test case for the test connection command.
+func filterFailedTestResults(results []testConnectionResult) (*testConnectionResult, error) {
```
erezrokah, line 310: Wait what's the reason we need to filter these plugins? I don't think their connection test can ever fail

erezrokah (commented):
A few more comments

cli/cmd/test_connection.go
disq: Screenshot 2024-05-17 at 15 28 57
I think this needs a PR :)

cli/cmd/test_connection.go
disq: Yes, now better, with proper `plugin_ref` etc.
```
2024-05-17T14:33:56Z INF Loading spec(s) args=["./pgrpc.yml"] invocation-id=dafde1e3-5e32-4fad-8fb5-c7ad7a885284 module=cli
2024-05-17T14:33:56Z INF Testing destination destination="pg (grpc@localhost:7778)" invocation-id=dafde1e3-5e32-4fad-8fb5-c7ad7a885284 module=cli
2024-05-17T14:33:56Z INF Test connection completed invocation-id=dafde1e3-5e32-4fad-8fb5-c7ad7a885284 module=cli testresults=[{"failure_code":"OTHER","failure_description":"rpc error: code = Internal desc = failed to init plugin: failed to initialize client: failed to get current database: failed to connect to `host=localhost user=postgres database=cq`: dial error (dial tcp 127.0.0.1:15432: connect: connection refused)","plugin_kind":"destination","plugin_ref":"pg (grpc@localhost:7778)","success":false}]
Error: at least one test connection failed
2024-05-17T14:33:56Z ERR exiting with error error="at least one test connection failed" invocation-id=dafde1e3-5e32-4fad-8fb5-c7ad7a885284 module=cli
```

cli/cmd/test_connection.go
disq: Those should be there until we update the cloud side to use the new cli.

erezrokah: Ah because with existing Cloud the new CLI will report failures for this?

erezrokah: I don't think it matters much this feature is only enabled for the `cloudquery` team

disq: If we want the cloud api side to work, this filtering needs to be there at least for now.

erezrokah (approved):

PR #17999, opened 2024-05-17
feat: Add explicit Test Connection support for DuckDB

disq:
Part of the explicit test connection support effort.


plugins/destination/duckdb/client/client.go
```diff
@@ -89,6 +78,29 @@ func (c *Client) Close(ctx context.Context) error {
...
+		return &plugin.TestConnError{
+			Code:    plugin.TestConnFailureCodeInvalidSpec,
+			Message: fmt.Errorf("failed to unmarshal spec: %w", err),
+		}
+	}
+	s.SetDefaults()
+
+	connector, err := duckdb.NewConnector(amendConnectionString(s.ConnectionString), nil)
+	if err != nil {
+		return err
```
disq, line 93: go-duckdb errors seem to be clear and not so deeply wrapped, that's why I omitted the `processError(err)` step here.

bbernays (approved):

PR #18033, opened 2024-05-21
chore: Doc update about incrementals supporting append destinations

disq:
Supported since SDK 4.37.0 ([PR](https://github.com/cloudquery/plugin-sdk/pull/1604) which added this feature) but it's only enabled for BigQuery destination, client initialized with `state.NewClientWithOptions` and `state.ClientOptions{Versioned: true}`)


website/pages/docs/advanced-topics/managing-incremental-tables.md
```diff
@@ -9,7 +9,7 @@ Incremental tables are tables that fetch only the data that changed since the la
...
 
-In order to resume from a previous position, incremental tables store some state, known as the **cursor**. When using the CloudQuery CLI, the cursor state is stored in a **backend**. Any destination plugin that supports `overwrite` mode can be used as a backend. This state destination is often the same destination data is being written to for the sync, but it doesn't need to be.
+In order to resume from a previous position, incremental tables store some state, known as the **cursor**. When using the CloudQuery CLI, the cursor state is stored in a **backend**. Any destination plugin can be used as a backend. This state destination is often the same destination data is being written to for the sync, but it doesn't need to be. In some destinations which support `append` mode, each state update will generate a new row (with a new version) in the state table and may need to be cleaned up periodically (this feature is not currently enabled for all `append`-only destinations). Destinations supporting `overwrite` mode don't have this issue.
```
hermanschaaf, line 12:
```suggestion
In order to resume from a previous position, incremental tables store some state, known as the **cursor**. When using the CloudQuery CLI, the cursor state is stored in a **backend**. Any destination plugin can be used as a backend. This state destination is often the same destination data is being written to for the sync, but it doesn't need to be. In some destinations that support only `append` mode, each state update will generate a new row (with a new version) in the state table and may need to be cleaned up periodically. Destinations supporting `overwrite` mode don't have this issue.
```

erezrokah:
> but it's only enabled for BigQuery destination

I think it's enabled per source, and we did it only for BigQuery

erezrokah (approved):

PR #18044, opened 2024-05-24
fix: Remove outdated login hint message

erezrokah:
#### Summary

This message is outdated and doesn't work well when someone has the plugins pre-downloaded already (e.g. using an offline license).


candiduslynx (approved):

erezrokah:
OK had to do another commit to only remove the warning. The initial PR changed so we return an error instead of `nil` which would break stuff.

The issue is that we can't really check if a token is needed just by looking at the spec, we need to see if the plugins are also pre-downloaded which would require a bit of work

PR #18051, opened 2024-05-24
fix: Pin alpine version in CLI Dockerfile

erezrokah:
#### Summary

To avoid issues like https://github.com/alpinelinux/docker-alpine/issues/396
I'll add a renovate config to handle updates via PRs

### DIFF
--- a/cli/Dockerfile.goreleaser
+++ b/cli/Dockerfile.goreleaser
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:3.19.1
 RUN apk --no-cache add git
 ENTRYPOINT ["/app/cloudquery"]
 WORKDIR /app


.github/renovate.json5
```diff
@@ -38,6 +38,9 @@
   gomod: {
     fileMatch: ["go\\.mod\\.tpl$"],
   },
+  dockerfile: {
+    fileMatch: ["Dockerfile\\.goreleaser$", "Dockerfile\\.ubuntu\\.goreleaser$"],
```
bbernays, line 42: This means the renovate will be handling upgrading both aline and ubuntu right?

erezrokah: > This means the renovate will be handling upgrading both aline and ubuntu right?

💯 Yes

bbernays (approved):

PR #18126, opened 2024-06-04
fix!: Properly support Apache Arrow types with non-default options

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/1786


candiduslynx:
will reopen soon

PR #18128, opened 2024-06-05
feat: Track login, sync started, sync ended, generic command start, end events

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery-issues/issues/1158 (internal issue).
First step in deprecating the current analytics client.

Opening for feedback


cli/cmd/sync.go
```diff
@@ -124,6 +125,10 @@ func sync(cmd *cobra.Command, args []string) error {
 	// in a cloud sync environment, we pass only the relevant environment variables to the plugin
 	osEnviron := os.Environ()
 
+	analytics.TrackSyncStarted(ctx, invocationUUID, analytics.SyncStartedEvent{
```
erezrokah, line 128: Easy to start started for all sources + destinations in a single place, for sync finished see below

cli/cmd/sync_v3.go
```diff
@@ -425,6 +426,16 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 		Str("result", msg).
 		Msg("Sync summary")
 
+	analytics.TrackSyncFinished(ctx, invocationUUID, analytics.SyncFinishedEvent{
```
erezrokah, line 429: A bit tricky to track sync finished without refactoring the CLI a bit.
Since we iterate over sources and print a `Sync completed successfully` per source not sure what sync finished means (all sources? per source?). We also exist on first source failure so how do we want to report a sync that the first source was successful, but the second one not?

To avoid refactoring the CLI, `TrackSyncFinished` is reported per source (same like sync summary and other log messages)

pilvikala: Makes sense, but we should then track the Sync Started per source as well.

cli/cmd/sync_v3.go
```diff
@@ -425,6 +426,16 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
+		Source:        sourceSpec,
```
pilvikala, line 430: I think we're pretty much interested in the plugin name (team name + plugin name), and the same for the destinations. That's how we track the syncs in Cloud as well.

cli/cmd/sync_v3.go
```diff
@@ -425,6 +426,16 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
+		Destinations:  destinationSpecs,
+		Errors:        totals.Errors,
+		Warnings:      totals.Warnings,
+		Duration:      syncTimeTook,
+		Result:        msg,
```
pilvikala, line 435: This would be better to have as simple "Succeeded" and "Failed". On cloud syncs, we have Status, and StatusReason.

cli/internal/analytics/client.go
```diff
@@ -0,0 +1,213 @@
...
+	if client == nil {
+		return
+	}
+
+	details := getEventDetails(ctx)
+	if details == nil {
+		return
+	}
+
+	_ = client.Enqueue(rudderstack.Track{
```
pilvikala, line 128: Is this enough to "identify" the user and the team?

cli/internal/analytics/client.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	toRedact := make([]*specs.Destination, len(event.Destinations))
+	for i := range event.Destinations {
+		toRedact[i] = &event.Destinations[i]
+	}
+
+	_ = client.Enqueue(rudderstack.Track{
+		UserId: details.userId,
+		Event:  "Sync Finished",
```
pilvikala, line 193: This should be tied to a group as well, the team property should be set too.

cli/internal/analytics/client.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	details := getEventDetails(ctx)
+	if details == nil {
+		return
+	}
+
+	_ = client.Enqueue(rudderstack.Track{
+		UserId: details.userId,
+		Event:  "Sync Started",
```
pilvikala, line 156: This should be tied to a group as well, the team property should be set too.

cli/internal/analytics/client.go
```diff
@@ -0,0 +1,213 @@
...
+)
+
+type eventDetails struct {
+	userId      string
+	currentTeam string
+	environment string
+}
+
+func initRudderStack() {
+	writeKey := env.GetEnvOrDefault("CQ_RUDDERSTACK_WRITE_KEY", "2h38sP5iH58EYKBTRsGByJDDr6r")
```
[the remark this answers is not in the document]
erezrokah: There's a bit more logic to enabling analytics so it's done in `root.go` https://github.com/cloudquery/cloudquery/pull/18128/files#diff-eb154bbd601602eeae895e7af6a12b88cbdb7d585a2d32216c398f68cabcdff5R95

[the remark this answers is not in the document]
hermanschaaf: Ah I missed that this is an uppercase `Init` function, so not automatically invoked by Go.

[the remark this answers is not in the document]
erezrokah: I'll rename it to avoid confusion

cli/cmd/root.go
```diff
@@ -27,12 +28,35 @@ Open source data integration at scale.
...
 )
 
+func wrapRunE(cmd *cobra.Command) {
+	// Ideally we could use PersistentPostRunE, but it doesn't get called if the command has an error
+	// See https://github.com/spf13/cobra/issues/1893
+	if cmd.RunE != nil {
+		wrapped := cmd.RunE
+		cmd.RunE = func(cmd *cobra.Command, args []string) error {
+			commandName := cmd.Name()
+			if cmd.Parent() != nil {
```
erezrokah, line 44: Not sure if it's worth handling more command nesting (we have only two levels, e.g. `plugin install`)

cli/internal/analytics/client.go
```diff
@@ -0,0 +1,229 @@
...
+		// sync_run_id to match with cloud events
+		Set("invocation_uuid", invocationUUID).
+		Set("sync_run_id", invocationUUID).
+		Set("team", details.currentTeam).
+		Set("environment", details.environment).
+		Set("sync_name", event.Source.Name).
+		Set("source_path", event.Source.Path).
+		Set("destination_paths", destinationPaths).
+		Set("user_id", details.user.ID).
+		Set("user_email", details.user.Email)
```
hermanschaaf, line 170: I missed this in my first review, but should we have sync group ID here (or somewhere) as well?

hermanschaaf (approved):

cli/internal/analytics/client.go
erezrokah: Do you mean https://docs.cloudquery.io/docs/reference/destination-spec#sync_group_id-preview? It's per destination, so I think if we add that maybe we should add an additional property with `source_spec` and `destinations_specs` with all the options we'd like to track.

hermanschaaf: Yeah we can leave it out for now, but I thought it would be nice to know when parallel syncs are related to one another.

PR #18161, opened 2024-06-06
fix: Progressbar related error when `log-console` option is used

disq:
Goes with https://github.com/cloudquery/plugin-pb-go/pull/336

~needs https://github.com/cloudquery/plugin-pb-go/pull/337 released~


candiduslynx (commented):
waiting for go mod without replace

candiduslynx (approved):

erezrokah (approved):

PR #18191, opened 2024-06-06
feat!: Change `skip_dependent_tables` default from `false` to `true`

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/1810, an internal issue to improve CLI defaults.

BEGIN_COMMIT_OVERRIDE
feat: Change `skip_dependent_tables` default from `false` to `true` (#18191)

BREAKING-CHANGE: Change `skip_dependent_tables` default from `false` to `true` (#18191), to avoid new tables implicitly being synced when added to plugins. See more in the [reference docs](https://docs.cloudquery.io/docs/reference/source-spec#skip_dependent_tables)

END_COMMIT_OVERRIDE

> We'll first do a CLI release with https://github.com/cloudquery/cloudquery/pull/18128 (next week), then do another major release with this change


cli/internal/specs/v0/source.go
```diff
@@ -68,6 +70,10 @@ func (s *Source) SetDefaults() {
 	if s.Spec == nil {
 		s.Spec = make(map[string]any)
 	}
+	if s.SkipDependentTables == nil {
+		b := true
+		s.SkipDependentTables = &b
```
candiduslynx, line 75: dark magic:
```suggestion
		s.SkipDependentTables = &[]bool{true}[0]
```

candiduslynx (approved):

cli/internal/specs/v0/source.go
disq: just do `lo.ToPtr(true)` without dark magic 😂

erezrokah: > just do `lo.ToPtr(true)` without dark magic 😂

Yeah didn't want to add it just for this 😄

bbernays (approved):

PR #18198, opened 2024-06-10
fix!: Support Apache Arrow types with non-default options

candiduslynx:
BEGIN_COMMIT_OVERRIDE
fix: Support Apache Arrow types with non-default options (#18198). Previously some Apache Arrow types (namely, `timestamp` with non-default time zone) would result in plugin failure.
BREAKING-CHANGE: Apache Arrow `timestamp` now always maps to BigQuery `timestamp` type (#18198). Previously, `timestamp` with `nanoseconds` precision was mapped to a `record` type.

END_COMMIT_OVERRIDE

Instead of https://github.com/cloudquery/cloudquery/pull/18126
Fixes https://github.com/cloudquery/cloudquery-issues/issues/1786 (internal issue)

### DIFF
--- a/plugins/destination/bigquery/client/client_test.go
+++ b/plugins/destination/bigquery/client/client_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/cloudquery/plugin-sdk/v4/plugin"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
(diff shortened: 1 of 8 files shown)


hermanschaaf (approved):

plugins/destination/bigquery/client/client_test.go
```diff
@@ -40,7 +41,8 @@ func TestPlugin(t *testing.T) {
 		plugin.WithTestDataOptions(schema.TestSourceOptions{
 			SkipMaps: true,
 			// https://github.com/cloudquery/cloudquery/issues/12022
-			SkipTimes: true,
+			SkipTimes:     true,
```
erezrokah, line 44: If we are skipping times, what's the reason we need to configure the precision?

candiduslynx: time32 & time64 only, not timestamp

PR #18200, opened 2024-06-10
fix: Isolate env variables in `test-connection` command if necessary

disq:
This was missing from the `test-connection` command (we have it in `sync`) and it will be used by the cloud side.
Currently we don't test multiple connections at the same time, but when/if we do, this will come in handy. It also simplifies the handling code as `test-connection` and `sync` behaves the same.


cli/cmd/test_connection.go
```diff
@@ -112,6 +112,10 @@ func testConnection(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
 	updateSyncTestConnectionStatus(cmd.Context(), log.Logger, cloudquery_api.SyncTestConnectionStatusStarted)
 
+	// in the cloud sync environment, we pass only the relevant environment variables to the plugin
+	isolatePluginEnvironment := apiAuth.NewTokenClient().GetTokenType() == apiAuth.SyncTestConnectionAPIKey
```
erezrokah, line 116: There's a fancy `CQ_CLOUD` env you could use, see https://github.com/cloudquery/cloudquery/pull/18128/files#diff-9adce87fe0e31bd984d6a819b5bbbcca2b0f2239249ae2a06f1d502f895fb7adR35

erezrokah: Maybe also update the sync command while you're at it

disq: Unless the `sync` command is already using that, it would be outside of the scope.

disq: Updated

erezrokah (approved):

disq:
Merge at will.

PR #18231, opened 2024-06-12
fix!: Remove `client_id` spec option

candiduslynx:
Closes https://github.com/cloudquery/cloudquery-issues/issues/1735
Library bump per https://github.com/IBM/sarama/pull/2706


plugins/destination/kafka/client/client.go
```diff
@@ -54,7 +55,7 @@ func New(_ context.Context, logger zerolog.Logger, s []byte, opts plugin.NewClie
 	c.conf.Metadata.Full = true
 	c.conf.Version = sarama.V1_0_0_0
 	c.conf.Metadata.Full = true
-	c.conf.ClientID = c.spec.ClientID
+	c.conf.ClientID = `cwc|1c04a227-aef8-47a9-9353-e20bbb6a9616|cq-destination-kafka|` + internalPlugin.Version
```
erezrokah, line 58: Shouldn't we set this value only when Kafka is configured to be used with Confluent? I don't think we want to send this ID to any Kafka configuration

erezrokah (changes requested):

plugins/destination/kafka/client/client.go
candiduslynx: As it's an open value (like our Sentry endpoints were), we can just use that & not worry for the differences.

candiduslynx:
Funny thing: the library we're using doesn't allow the client ID format we want to set

plugins/destination/kafka/client/client.go
erezrokah:
```
// A user-provided string sent with every request to the brokers for logging,
// debugging, and auditing purposes. Defaults to "sarama", but you should
// probably set it to something specific to your application.
ClientID
```

Since the value ends up visible to users and they can use it for logging, seems still useful to have? We should only disallow updating it when Confluent it the target to my understanding

candiduslynx: I still think it's actually a good idea to just have it hardcoded + version, as we do for the user-agents in some APIs.
The users would still be able to cross-reference this if they need to.

erezrokah: In other plugins we send a generic version of the user agent, this one is specific to one flavor of Kafka. Also this would make the change a non breaking one I think

erezrokah: Is it hard detecting we're on Confluent from the brokers list for example?

candiduslynx: Nothing is documented to be only confluent cloud endpoint.
& if they generate the endpoints for different clouds, we'll be even more in a pickle.

erezrokah (approved):

PR #18239, opened 2024-06-13
fix: Skip relations of incremental tables from delete stale

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/1844 (internal issue).
Relations of incremental tables gets cleared due to delete stale, we should exclude them I think.
The fix works since we always sync parents from before children, so parents get excluded first.

~~Still needs to test this.~~

Blocked by https://github.com/cloudquery/plugin-sdk/pull/1742 (I mean the code works, but the bug won't be fixed until source plugins upgrade to latest SDK)


cli/cmd/sync_v3.go
```diff
@@ -326,10 +327,15 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
...
 			}
-			tableName := tableNameFromSchema(sc)
+			table, err := schema.NewTableFromArrowSchema(sc)
+			if err != nil {
+				return err
+			}
 
-			if !isStateBackendEnabled || !tableIsIncremental(sc) {
-				tablesForDeleteStale[tableName] = true
+			if isStateBackendEnabled && (table.IsIncremental || (table.Parent != nil && skippedFromDeleteStale[table.Parent.Name])) {
```
candiduslynx, line 335: IIRC the marshaled date will have only flattened stuff, so it's impossible to fix this on the destination side.

When running the sync we could to table -> schema.Tables{table).Flatten -> exclude all

erezrokah: There's another fix in the SDK needed for this to send the parent in the migrate message which I still need to test https://github.com/cloudquery/plugin-sdk/compare/main...erezrokah:plugin-sdk:fix/send_parent_migrate?expand=1

erezrokah: Works with https://github.com/cloudquery/plugin-sdk/pull/1742

candiduslynx: This at the very least implies that the parent migrate is sent before the relation migrate (which isn't documented anywhere as such).
Approving, but this should be taken into account & possibly documented as a requirement.

candiduslynx (approved):

cli/cmd/sync_v3.go
erezrokah, line 335:
```suggestion
			// This works since we sync and send migrate messages for parents before children
			if isStateBackendEnabled && (table.IsIncremental || (table.Parent != nil && skippedFromDeleteStale[table.Parent.Name])) {
```

PR #18282, opened 2024-06-17
feat: Improve tracking of sync command.

marianogappa:
In order to improve the user journey for cloudquery users, we need to better understand the issues they face when they first start working with the cli.

The journey is roughly:

- download CLI
- run the login command
- create account
- configure the sync
- run it

For the most part, we can track these interactions on a team+user basis, except that we don't have visibility of failures when a sync doesn't finish, except a non-authed `command_sync_end` with `status: error` with no error details.

I had a chat with @pilvikala and we need the following:
- `sync_run_completed` should be tracked even on errors
- Ideally it should contain team+user, but at least team is needed
- It's better to have an existing event name for this (i.e. `sync_run_completed`) instead of creating a new one for the new cases or new information

This PR makes sure that:
- `sync_run_completed` is always sent, even when the sync is aborted due to a failure
- an `aborted_due_to_error` property was added, which contains the error that caused the command to not finish successfully
- This event contains team+user information

Note that the naming of the event can be slightly confusing: `sync_run_completed` might sound like a success. IMO it is a thing for a process to "complete with errors", and in fact syncs do complete successfully with some row sync errors too. But if this is a problem I can create a new event.

Screenshot 2024-06-18 at 09 28 20


erezrokah (approved):
Nice

Issue #18274, opened 2024-06-14
bug(tests): Meilisearch CI is failing 


erezrokah:
https://github.com/cloudquery/cloudquery/actions/runs/9519814526/job/26243817464?pr=18273#step:2:123

Started failing for no apparent reason.

Could be related to https://github.com/meilisearch/meilisearch/issues/3832

erezrokah:
Based on the [docs](https://www.meilisearch.com/docs/learn/configuration/instance_options#environment) the key should be 16 bytes long at least (I reproduced the CI error locally)

PR #18292, opened 2024-06-18
chore: Update Meilisearch health check

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/issues/18274


.github/workflows/dest_meilisearch.yml
```diff
@@ -37,11 +37,10 @@ jobs:
         ports:
         - 7700:7700
         options: >-
-                 --health-cmd "wget --no-verbose --tries=1 --spider http://localhost:7700/health || exit 1"
```
erezrokah, line 40: Maybe GitHub actions updated their `wget` version. When I run the old check locally on a health instance I get:
```
http://localhost:7700/health:
Remote file does not exist -- broken link!!!
```

murarustefaan (approved):

PR #18296, opened 2024-06-18
fix: Cleanup on logout

candiduslynx:
Fixes https://github.com/cloudquery/cloudquery-issues/issues/815

### DIFF
--- a/cli/cmd/logout.go
+++ b/cli/cmd/logout.go
@@ -2,14 +2,11 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
 
-	authpb "github.com/cloudquery/cloudquery-api-go/auth"
-	"github.com/cloudquery/cloudquery-api-go/config"
 	"github.com/cloudquery/cloudquery/cli/internal/auth"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cli/cmd/logout.go
```diff
@@ -41,7 +44,7 @@ func newCmdLogout() *cobra.Command {
 }
 
 func runLogout(_ context.Context, cmd *cobra.Command) error {
-	err := auth.Logout()
+	err := errors.Join(auth.Logout(), authpb.RemoveRefreshToken(), config.UnsetValue("team"))
```
marianogappa, line 47: Looks like `auth.Logout()` is doing `RemoveRefreshToken()` and `config.UnsetValue("team")`, but sequentially and stopping on errors rather than sequentially and not stopping on errors.

Shouldn't we just change the behaviour of logout? 🤔

candiduslynx: 335ee72579983358e88f08263a4d8355739cec9f

cli/internal/auth/logout.go
```diff
@@ -1,22 +1,22 @@
...
-		return fmt.Errorf("failed to remove refresh token: %w", err)
+	var errs []error
+	if err := cqapiauth.RemoveRefreshToken(); err != nil {
+		errs = append(errs, fmt.Errorf("failed to remove refresh token: %w", err))
 	}
 
-	err = config.UnsetValue("team")
-	if err != nil {
-		return fmt.Errorf("failed to unset team: %w", err)
+	if err := config.UnsetValue("team"); err != nil {
```
erezrokah, line 17: @candiduslynx were you able to reproduce the issue? Seems like it shouldn't happen normally since we're already clearing the team from the config

candiduslynx: If we have an issue removing refresh token this might happen

erezrokah: I understand, but did you ran into an issue where we get that error to remove the token? Based on the issue description not sure it's related as no errors where reported during logout

PR #18336, opened 2024-06-20
fix: Don't reuse memory pools between tables

candiduslynx:
Speedup memory GC
Testing on multi-row
Pre change:
s3 mrow

Post change
s3-uploader mrow


erezrokah (approved):
@candiduslynx maybe add some details, i.e. the benchmarking @marianogappa did to better explain this change.

marianogappa (approved):

PR #18397, opened 2024-06-25
feat: Add per table sync metrics file flag

erezrokah:
#### Summary

Goes with https://github.com/cloudquery/plugin-sdk/pull/1751

Opening for reference


cli/cmd/sync.go
```diff
@@ -123,7 +124,16 @@ func sync(cmd *cobra.Command, args []string) error {
 	// in a cloud sync environment, we pass only the relevant environment variables to the plugin
 	osEnviron := os.Environ()
 
+	otelReceiver, err := otel.StartOtelReceiver(ctx)
+	if err == nil {
+		defer otelReceiver.Shutdown(ctx)
+	}
+
 	for _, source := range sources {
+		if source.OtelEndpoint == "" {
```
erezrokah, line 133: Probably need to be a spec option instead of the default

cli/internal/otel/receiver.go
```diff
@@ -0,0 +1,261 @@
...
+	if c.tracesFile != nil {
+		err = errors.Join(err, c.tracesFile.Close())
+	}
+	if c.metricsFile != nil {
+		err = errors.Join(err, c.metricsFile.Close())
+	}
+	return err
+}
+
+func newDefaultSpanConsumer(tracesFile *os.File) ConsumeSpan {
```
erezrokah, line 65: I'm writing spans and metrics to files but we could do pretty much what we want we the data

cli/internal/otel/receiver.go
```diff
@@ -0,0 +1,358 @@
...
+	consumeMetric   ConsumeMetric
+}
+
+func (c *Consumer) Shutdown(ctx context.Context) error {
+	var err error
+	if c.quit != nil {
+		close(c.quit)
+	}
+	if c.metricsFile != nil {
+		err = errors.Join(err, c.metricsFile.Close())
```
marianogappa, line 73: This joins ` c.metricsFile.Close()` with `nil`. I assume it was a rewrite leftover?

cli/internal/otel/receiver.go
```diff
@@ -0,0 +1,358 @@
...
+func newDefaultSpanConsumer() ConsumeSpan {
+	return func(ctx context.Context, span Span) {
+		// do nothing
+	}
+}
+
+func newDefaultMetricConsumer(metricsFile *os.File, quit chan any) ConsumeMetric {
+	tableLock := sync.Mutex{}
+	metricsMap := make(map[string]*TableMetric)
+	ticker := time.NewTicker(20 * time.Second)
```
marianogappa, line 87: Should this be configurable?

cli/internal/otel/receiver.go
```diff
@@ -0,0 +1,358 @@
...
+				metrics.Table,
+				metrics.ClientId,
+				metrics.StartTime,
+				metrics.EndTime,
+				metrics.Resources,
+				metrics.Errors,
+				metrics.Panics,
+			}
+			if metrics.EndTime == nil {
+				row[3] = "N/A"
```
marianogappa, line 125: I'm aware that `row`'s type is `[]interface{}`, but unsure how the code that uses it works. Is it ok to mix `EndTime *time.Time` with `string` on `row[3]`?

marianogappa (approved):
I added a few comments. It's not a bulletproof review because I'm a little unfamiliar with Go's otel (I've used it in Python). I assume you've tested this code and it largely works.

cli/internal/otel/receiver.go
erezrokah: Or so without this the table with render `<nil>` for nil values, I'll post an example for how the content of the file looks like in a bit

cli/internal/otel/receiver.go
erezrokah: Fixed in https://github.com/cloudquery/cloudquery/pull/18397/commits/0f79d15bac686b591c2de5e0192d9a42075fe349

cli/internal/otel/receiver.go
erezrokah: Could be, not sure how to expose it yet? Another CLI flag seems like an overkill

erezrokah:
This works only with plugins released with https://github.com/cloudquery/plugin-sdk/pull/1751 so I'll wait before merging this

erezrokah:
> This works only with plugins released with [cloudquery/plugin-sdk#1751](https://github.com/cloudquery/plugin-sdk/pull/1751) so I'll wait before merging this

OK so I'll actually mark this as a hidden flag so I can release the CLI, then unhide it once all plugins are updated

PR #18441, opened 2024-07-03
fix: Replace string type mapping with nvarchar(max) in the MSSQL destination plugin

murarustefaan:


savme (approved):

bbernays (approved):
This looks good! I would still add a check in the `write` function that will validate that any string PKs are below the limit, otherwise the entire batch fails. This could be done in a followup PR as well

murarustefaan:
@bbernays I think we should do this in a followup as we might need to change the 4000 limit as well.
Based on [the MS docs](https://learn.microsoft.com/en-us/sql/relational-databases/tables/primary-and-foreign-key-constraints?view=sql-server-ver16), the PK limit is 900 bytes, which would mean we would need to turn the mapping of string PKs into `nvarchar(900)`.

disq (approved):

PR #18447, opened 2024-07-03
feat: Add support for plugin UI uploads

disq:
Part of https://github.com/cloudquery/cloudquery-issues/issues/1864 (internal issue)


cli/cmd/plugin_uiassets_upload.go
```diff
@@ -0,0 +1,88 @@
...
+
+const (
+	pluginUIAssetsUploadShort = "Upload plugin UI assets to CloudQuery Hub."
+	pluginUIAssetsUploadLong  = `Upload plugin UI assets to CloudQuery Hub.
+
+This uploads UI assets for a specific plugin version from a local build directory to CloudQuery hub.
+`
+	pluginUIAssetsUploadExample = `
+# Upload plugin UI assets to CloudQuery Hub
+cloudquery plugin uiassets upload test-team/source/test-plugin@v1.0.0`
```
hermanschaaf, line 25: I think we don't really need a separate command for this--it should be good enough to use `cloudquery plugin publish` and check for the `--ui-dir` flag? If the concern is that it's slow to re-upload the binaries every time, maybe let's fix that instead by comparing the checksums before doing a new upload

cli/cmd/testdata/ui-build/index.html
```diff
@@ -0,0 +1 @@
+boo
```
hermanschaaf, line 1:
```suggestion
<strong>boo</strong>
```

cli/internal/publish/images/images.go
```diff
@@ -335,6 +335,7 @@ func (f *imageFinder) Transform(node *ast.Document, reader text.Reader, pc parse
 						}
 						if len(el.Title) == 0 { // if no title, make sure the element ends with the link
 							parts := bytes.SplitN(lineLiteral, el.Destination, 2)
+							// nolint:revive
```
hermanschaaf, line 338: just curious, what's the issue being suppressed here?

cli/cmd/plugin_uiassets_upload.go
disq: What if you're not publishing a plugin and simply want to fix a bug in the UI? The command architecture/setup is the same as `plugin docs upload` (which also offers `download` and `upload --sync` though)

cli/internal/publish/images/images.go
disq: more than 5 levels of deep indentation. I'm guessing the linter was upgraded recently on the CLI repo and this is causing linter error.

cli/cmd/plugin_uiassets_upload.go
hermanschaaf: > What if you're not publishing a plugin and simply want to fix a bug in the UI?

I think it's simpler (and less maintenance) to have one command that does it all -- `cloudquery plugin publish` no matter what you changed. We just need to make it a bit smarter so it's not a drag to re-upload all the binaries every time.

cli/internal/publish/images/images.go
hermanschaaf: got it, okay - we don't have to fix it here but maybe lets add a comment to explain why it's being suppressed, so some good citizen can refactor it in another PR

cli/cmd/plugin_uiassets_upload.go
disq: - publish requires a manifest.json so you'd have to create a manifest file every time there's an issue, so you'd have to add a new option to get the plugin name/version directly which would complicate the code and be confusing.
- publish requires the plugin to not be published already, so you'd have to add another flag, or maybe make it so that it skips that half of the publush flow when the ui assets flag is specified along with the plugin reference flag
- publish is a documented command so you'd have to document all this, with exceptions ("if you use this flag, this portion of the publish flow won't be run", etc)
- `plugin uiassets` is a hidden command with no downside/documentation burden, just like `plugin docs` is

hermanschaaf: I see, thanks for the clarification. 

I was thinking that the UI publishing would be immutable after the plugin is finalized, so more like binaries, less like docs. If you fix a UI bug, you would upload a new version of the plugin, just like if you fixed a bug in the spec or config validation. I think this is similar to the Grafana model, where a UI bug will remain in a plugin until you upgrade the version.

Not versioning UI changes as part of the plugin version could lead to confusing issues down the line, if these are being cached somewhere other than our own bucket. But okay, I think we can live with it for now if it's a hidden command, meant only to be used in special circumstances.

cli/internal/publish/images/images.go
disq: As in https://github.com/cloudquery/cloudquery/pull/18450

cli/cmd/plugin_uiassets_upload.go
disq: We're versioning the plugin UI assets separately so that there's no chance of caching issues - each UI "version" has its own uuid prefix. Grafana kept it simple but it's part of their core product, where with us the UI is completely separate from framework/plugins and only used in cloud. Because of this separate versioning needed to happen to keep things sane.

marianogappa: What will the workflow be if we'd need to support some UI issue for a plugin? I've seen in support that we end up having to try a sync by checking out some tag of an older version, because the customer hasn't/can't update to the latest version.

If we had a separate UUID-based versioning scheme for the UI Framework, would it be cumbersome to see a certain version of the UI?

cli/internal/publish/pluginui.go
```diff
@@ -0,0 +1,134 @@
...
+
+func quickContentType(filename string) (string, error) {
+	filebytes := make([]byte, 512)
+
+	fp, err := os.Open(filename)
+	if err != nil {
+		return "", fmt.Errorf("failed to open file: %w", err)
+	}
+	defer fp.Close()
+	if _, err := fp.Read(filebytes); err != nil {
```
marianogappa, line 118: I think this fails if the file is < 512 bytes.

```
    if err != nil && err != os.EOF {
        return "", fmt.Errorf("failed to read file: %w", err)
    }
```

cli/cmd/plugin_uiassets_upload.go
marianogappa: I had a chat with @hermanschaaf about this to get context and I tend to agree with having separate commands to begin with until we have some real-world experience and see the tradeoffs better.

marianogappa (approved):
Added comments.

cli/cmd/plugin_publish_test.go
```diff
@@ -181,6 +184,124 @@ func TestPluginPublishFinalize(t *testing.T) {
...
+			w.Write([]byte(`{}`))
+		case "/upload-darwin":
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{}`))
+		case "/upload-uiasset":
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{}`))
+		case "/plugins/cloudquery/source/test/versions/v1.2.3/uiassets":
+			checkAuthHeader(t, r)
+			if r.Method == http.MethodPost {
```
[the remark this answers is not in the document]
disq: Correct but linter would've shouted at me for having that probably 😂

cli/cmd/plugin_publish_test.go
```diff
@@ -8,10 +8,13 @@ import (
 	"net/http/httptest"
 	"os"
 	"strings"
+	gosync "sync"
```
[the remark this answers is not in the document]
disq: `sync` seems to be a symbol in this package, could be a variable. Let me check locally...

oh, it's because of:
`func sync(cmd *cobra.Command, args []string) error`

And the codebase agrees with me as the `cmd` package as multiple places where `gosync "sync"` line exists, apparently.

cli/cmd/plugin_publish_test.go
```diff
@@ -181,6 +184,124 @@ func TestPluginPublishFinalize(t *testing.T) {
...
+	}
+	gotCalls := map[string]int{}
+	mu := &gosync.Mutex{}
+	uiID := ""
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		mu.Lock()
+		gotCalls[r.Method+" "+r.URL.Path]++
+		mu.Unlock()
+		// t.Log(r.Method, r.URL.Path)
```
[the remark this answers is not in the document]
disq: Not really but it's useful for developing. Left it because it's a test file, after all.

cli/internal/publish/pluginui.go
disq: The test passes with `index.html` which is a strong **boo**, so it seems to work... rest of the `filebytes` would be left as zero bytes, probably.

cli/cmd/plugin_publish_test.go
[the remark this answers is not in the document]
disq: Again, I wouldn't write this like that if it wasn't a one time test code.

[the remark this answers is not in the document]
disq: OK added double switch in https://github.com/cloudquery/cloudquery/pull/18447/commits/51bf299438877ca727c6f7dc09dc2ba8dad8aaa7 😂

cli/internal/analytics/client.go
```diff
@@ -128,7 +128,7 @@ func getSyncCommonProps(invocationUUID uuid.UUID, event SyncStartedEvent, detail
 		Set("source_path", event.Source.Path).
 		Set("destination_paths", destinationPaths).
 		Set("user_id", details.user.ID).
-		Set("user_email", string(details.user.Email))
+		Set("user_email", details.user.Email)
```
erezrokah, line 131: What's the reason for this change? This was a bug fix https://github.com/cloudquery/cloudquery/pull/18208

PR #18479, opened 2024-07-05
fix: Fix pagination logic for responses on Typeform plugin.

marianogappa:
Within the Typeform plugin, the client's logic for pagination of `forms_responses` is a recursive call that omits the `form_id`. It's likely that we never encountered this problem because the page size is 1000.


hermanschaaf (approved):
Nice 👍

PR #18492, opened 2024-07-08
feat: Implement support for Incremental Table on Typeform's FormResponses table

marianogappa:
fixes https://github.com/cloudquery/cloudquery/issues/18490

Implements support for Incremental Table on Typeform's FormResponses table.

Bumps plugin-sdk-python, which enables Python clients to use a StateClient.

Before:

```bash
✓ Code/test-typeform-postgres-sync  $ cli sync cloudquery-config                                                      ⏱ 10:35:07
Loading spec(s) from cloudquery-config
Starting sync for: typeform (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.0.7)]
Sync completed successfully. Resources: 457, Errors: 0, Warnings: 0, Time: 2
```

Enable PR
```bash
git checkout mariano/state-client
```

After:
```bash
✓ Code/test-typeform-postgres-sync  $ cli sync cloudquery-config                                                      ⏱ 10:35:19
Loading spec(s) from cloudquery-config
Starting sync for: typeform (grpc@localhost:7777) -> [postgresql (cloudquery/postgresql@v8.0.7)]
Sync completed successfully. Resources: 20, Errors: 0, Warnings: 0, Time: 1s
```


plugins/source/typeform/plugin/tables/form_responses.py
```diff
@@ -29,11 +30,13 @@ def __init__(self) -> None:
...
                 Column("tags", JSONType()),
             ],
+            is_incremental=True,
         )
+        self._resolver = FormResponsesResolver(table=self)
 
     @property
     def resolver(self):
-        return FormResponsesResolver(table=self)
+        return self._resolver
```
marianogappa, line 39: This is a dangerous anti-pattern we should be aware of. It was creating a brand new resolver every time the property was being referenced (note that by adding `@property`, you don't call `resolver()`, but just `resolver`).

hermanschaaf (approved):

plugins/source/typeform/plugin/tables/form_responses.py
dkeyer-twilio: @marianogappa What's the risk of this usage, out of curiosity? Is it performance only?

marianogappa: @dkeyer-twilio oh sorry, I should have been more clear. There's no performance issue.

- `resolver` here is acting as a `getter`, and one doesn't expect a getter to create a brand new object.
- Objects are stateful. One wouldn't normally expect to set some state on an object and then not have it later.
- In this case, the `.sync()` method calls `resolver.set_state_client(...)`, and this method would erase that state.

I think this pattern causes very unintuitive behaviour. I certainly lost an hour on it.

PR #18499, opened 2024-07-08
fix: Don't close metrics file before last print

erezrokah:
#### Summary

We were closing the metrics file before the last print was done, so the last values where not printed.
This fixes it. Follow up to https://github.com/cloudquery/cloudquery/pull/18498


cli/internal/otel/receiver.go
```diff
@@ -145,6 +142,7 @@ func newDefaultMetricConsumer(metricsFile *os.File, quit chan any) ConsumeMetric
 		clientId := metric.Attributes["sync.client.id"].(string)
 
 		tableLock.Lock()
+		defer tableLock.Unlock()
```
erezrokah, line 145: This ensures `renderTable` is handled only after we finished dealing with the latest batch

marianogappa (approved):

PR #18537, opened 2024-07-09
chore: Update publish workflows to include `cloud-config-ui`

disq:
Part of https://github.com/cloudquery/cloudquery-issues/issues/1864 (internal issue)


marianogappa:
As discussed offline with @disq this is very hard to review, but it has been tested in the private repo so it's reasonable to enable it. I can let @erezrokah have the final say and approve it.

erezrokah (approved):

PR #18538, opened 2024-07-09
fix: Handle concurrent map writes

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/1999 (hopefully).
While the comment in the code is correct and all messages for the same table are processed sequentially, I don't think there's a guarantee that `WriteTable` is not called from multiple concurrent Go routines

### DIFF
--- a/plugins/destination/s3/client/write.go
+++ b/plugins/destination/s3/client/write.go
@@ -68,11 +68,12 @@ func (c *Client) WriteTable(ctx context.Context, msgs <-chan *message.WriteInser
 			table := msg.GetTable()
 			objKey := c.spec.ReplacePathVariables(table.Name, uuid.NewString(), time.Now().UTC(), c.syncID)
 			// if object was already initialized, use the same key
-			// We don't need any locking here because all messages for the same table are processed sequentially
+			c.initializedTablesLock.Lock()
 			if val, ok := c.initializedTables[table.Name]; ok {
 				objKey = val
 				delete(c.initializedTables, table.Name)
 			}
+			c.initializedTablesLock.Unlock()
 
 			var err error
 			s, err = c.createObject(ctx, table, objKey)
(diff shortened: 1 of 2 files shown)


plugins/destination/s3/client/write.go
```diff
@@ -68,11 +68,12 @@ func (c *Client) WriteTable(ctx context.Context, msgs <-chan *message.WriteInser
 			table := msg.GetTable()
 			objKey := c.spec.ReplacePathVariables(table.Name, uuid.NewString(), time.Now().UTC(), c.syncID)
 			// if object was already initialized, use the same key
-			// We don't need any locking here because all messages for the same table are processed sequentially
+			c.initializedTablesLock.Lock()
```
erezrokah, line 71: I don't think we can use a `RWMutex` sync we want the entire operation to be atomic

plugins/destination/s3/client/write.go
```diff
@@ -68,11 +68,12 @@ func (c *Client) WriteTable(ctx context.Context, msgs <-chan *message.WriteInser
 			table := msg.GetTable()
 			objKey := c.spec.ReplacePathVariables(table.Name, uuid.NewString(), time.Now().UTC(), c.syncID)
 			// if object was already initialized, use the same key
-			// We don't need any locking here because all messages for the same table are processed sequentially
```
bbernays, line 71: Why is this assumption not true anymore? Or was it never true?

erezrokah: Looking at the batch code (even before https://github.com/cloudquery/plugin-sdk/pull/1719) I'm not sure we ever had a guarantee that `WriteTable` is not called from multiple Go routines

bbernays (approved):

erezrokah:
We decided to ship this hot fix for s3 then look at the SDK code to see what's causing `WriteTable` to be called from multiple Go routines

PR #18544, opened 2024-07-11
feat: Add s3 destination connection test

savme:
#### Summary

This PR implements the ConnectionTester interface for the S3 destination, enabling the `test-connection` command.

Screenshot 2024-07-11 at 11 35 56


plugins/destination/s3/client/client.go
```diff
@@ -105,7 +109,7 @@ func New(ctx context.Context, logger zerolog.Logger, s []byte, opts plugin.NewCl
 		}
 
 		if _, err := manager.NewUploader(c.s3Client).Upload(ctx, params); err != nil {
-			return nil, fmt.Errorf("failed to write test file to S3: %w", err)
+			return nil, errors.Join(errTestWriteFailed, err)
```
blesniewski, line 112: (just asking for my own curiosity) are there any benefits to this approach vs the `Errorf` wrap? or just to uniformly be using the `errors` package?

savme: I'm using this specific error in the `errors.Is` check later. So using `errors.Join` seemed like the most straightforward approach here

blesniewski (approved):

PR #18569, opened 2024-07-15
feat: Auto migrate removal of unique constraints

erezrokah:
#### Summary

Preparation for https://github.com/cloudquery/cloudquery/pull/17129. This ensures the MySQL destination can handle the change in https://github.com/cloudquery/cloudquery/pull/17129


plugins/destination/mysql/client/migrate.go
```diff
@@ -69,7 +69,7 @@ func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
 				return false
 			}
 		case schema.TableColumnChangeTypeRemoveUniqueConstraint:
-			return true
```
erezrokah, line 72: This was a bug, we need `continue` to ensure we take into account all changes

plugins/destination/mysql/client/schema.go
```diff
@@ -128,6 +131,17 @@ func (c *Client) addColumn(ctx context.Context, table *schema.Table, column *sch
 	return err
 }
 
+func (c *Client) removeUniqueConstraint(ctx context.Context, table *schema.Table, column *schema.Column) error {
+	builder := strings.Builder{}
+	builder.WriteString("ALTER TABLE ")
+	builder.WriteString(identifier(table.Name))
+	builder.WriteString(" DROP INDEX ")
+	builder.WriteString(identifier(column.Name))
```
erezrokah, line 139: This is the default name for the constraint, we could try and read the name as well, but since we're the ones adding it with the default name I think that's ok for now

maaarcelino: The name suggests this function drops unique indexes only, but I think it can drop any index, right?

erezrokah: Correct. renamed the function to clarify. Other indices won't be identified as a unique constraint removal, but as a `TableColumnChangeTypeUpdate` (https://github.com/cloudquery/plugin-sdk/blob/316e1e30736bdb0dc2f1e38755e0437629373ae5/schema/table.go#L502), so there are not auto migratable per https://github.com/cloudquery/cloudquery/blob/d84ff3499707c024e78e8cdf98c6519129a92a96/plugins/destination/mysql/client/migrate.go#L74

maaarcelino (approved):

PR #18571, opened 2024-07-15
fix: Allow `plugin install` command to be run with a single-plugin spec

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/18542, so you can install a single plugin without linking it to a destination/source


marianogappa (approved):
I _think_ this change means that if they do run plugin install with 2 plugins we wouldn't do this check:

```
// here we check if source with different versions use the same destination and error out if yes
```

PR #18714, opened 2024-07-29
feat: PostgreSQL destination plugin cloud config UI

jackcloudquery:
#### Summary

PostgreSQL destination plugin UI
Update scaffold with e2e tests


plugins/destination/postgresql/cloud-config-ui/src/form/index.tsx
```diff
@@ -0,0 +1,263 @@
...
+              <TextField
+                error={!!fieldState.error}
+                fullWidth={true}
+                required={true}
+                helperText={
+                  fieldState.error?.message || (
+                    <>
+                      Specifies the update method to use when inserting rows.{' '}
+                      <a
+                        href="https://docs.cloudquery.io/docs/reference/destination-spec#write_mode"
```
cduflo, line 232: does this work with the iframe embedded in the main app? Or do we need to use the messenger open_url here?

plugins/destination/postgresql/cloud-config-ui/src/form/radio.tsx
```diff
@@ -0,0 +1,90 @@
...
+    disabled?: boolean;
+  }>;
+}
+
+/**
+ * This component displays a group of radio buttons.
+ *
+ * @public
+ */
+export function RadioGroupSelector<Value extends string | number | boolean>({
```
cduflo, line 31: we have ExclusiveToggle in the plugin-config-ui-lib that is pretty close to this, not sure if you want to swap it out?

cduflo (approved):
Nice! Nothing blocking, just 2 questions 👍

plugins/destination/postgresql/cloud-config-ui/src/form/index.tsx
jackcloudquery: Nice catch, thanks!

plugins/destination/postgresql/cloud-config-ui/src/form/radio.tsx
jackcloudquery: Thanks, this is actually leftover

PR #18741, opened 2024-07-30
feat: Add Indexes to remove need for full index scan on deletion

bbernays:


plugins/destination/postgresql/client/migrate.go
```diff
@@ -399,3 +402,25 @@ func (c *Client) removeUniqueConstraint(ctx context.Context, table *schema.Table
...
 	return table.Name + "_cqpk"
 }
+
+func (c *Client) createPerformanceIndexes(ctx context.Context, table *schema.Table) error {
+
+	columns := []string{"_cq_source_name", "_cq_sync_time"}
+
+	for _, col := range columns {
+		if table.Columns.Get(col) == nil {
+			c.logger.Debug().Msgf("no performance indexes created because %s column is not present", col)
```
disq, line 412: Below `columns[1]` is not used at all, why are we checking for its existence? Are we trying to determine if the `_cq_source_name` col is legit/cq-owned?

bbernays: columns[1] should be used.

fixed in [d58dffc](https://github.com/cloudquery/cloudquery/pull/18741/commits/d58dffc440389daa442d8c9a29757cdc0ce609b5)

erezrokah:
> We can't add this as part of the table definition because we don't have a field that can contain data about arbitrary indexes, so we just always ensure the index is present.

Do we want to add one? We could add an additional field here https://github.com/cloudquery/plugin-sdk/blob/a0c96a69b4a87e8d5c018876130d92f9111c4040/schema/column.go#L45

bbernays:
We could add it, but it wouldn't be simple because we have to retain the order of columns in the index.

Maybe we can wait and see if another plugin needs this functionality and add it to the SDK in that case?

erezrokah:
> Maybe we can wait and see if another plugin needs this functionality and add it to the SDK in that case?

OK, then I would put this behind a spec option, not enabled by default.

erezrokah (approved):

PR #18761, opened 2024-07-31
chore: Update API for sync test connection updates

disq:
Stop using deprecated endpoint so we can remove it.

Part of https://github.com/cloudquery/cloudquery-issues/issues/2152 (internal issue)

### DIFF
--- a/cli/cmd/test_connection.go
+++ b/cli/cmd/test_connection.go
@@ -72,22 +72,42 @@ func updateSyncTestConnectionStatus(ctx context.Context, logger zerolog.Logger, 
 	}
 
 	log.Info().Str("status", string(status)).Msg("Sending sync test connection to API")
-	requestBody := cloudquery_api.UpdateSyncTestConnectionJSONRequestBody{
-		Status: status,
+
+	var statusCode int
+	switch os.Getenv("_CQ_SYNC_TEST_CONNECTION_KIND") {
+	case "source":
+		requestBody := cloudquery_api.UpdateSyncTestConnectionForSyncSourceJSONRequestBody{
+			Status: status,
+		}
+		if failedTestResult != nil {
+			requestBody.FailureCode = &failedTestResult.FailureCode
+			requestBody.FailureReason = &failedTestResult.FailureDescription
+		}
+		res, err := apiClient.UpdateSyncTestConnectionForSyncSourceWithResponse(ctx, teamName, syncTestConnectionUUID, requestBody)
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to send sync test connection result to API")
+			return
+		}
+		statusCode = res.StatusCode()
+	default:
+		requestBody := cloudquery_api.UpdateSyncTestConnectionForSyncDestinationJSONRequestBody{
+			Status: status,
+		}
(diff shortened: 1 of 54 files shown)


cli/cmd/test_connection.go
```diff
@@ -318,16 +338,12 @@ func testPluginConnection(ctx context.Context, client plugin.PluginClient, spec
...
 //
-// The function returns any failed test results, or nil if all tests passed. The hackernews plugin is excluded from the
-// failed test results since it was previously used as a test case for the test connection command.
+// The function returns any failed test results, or nil if all tests passed.
 func filterFailedTestResults(results []testConnectionResult) (*testConnectionResult, error) {
 	var failedResults []testConnectionResult
 
 	for _, result := range results {
 		if !result.Success {
-			if strings.Contains(result.PluginRef, "cloudquery/file@v4.0.4)") || strings.Contains(result.PluginRef, "cloudquery/hackernews@v3.0.25)") {
```
disq, line 328: This code is unnecessary because test connections have been using a relaxed spec reader for some time now, and dummy sources/dests are not included in cloud invocations anymore.

hermanschaaf (approved):

cli/cmd/test_connection.go
```diff
@@ -72,22 +72,42 @@ func updateSyncTestConnectionStatus(ctx context.Context, logger zerolog.Logger,
...
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to send sync test connection result to API")
+			return
+		}
+		statusCode = res.StatusCode()
 	}
-	if res.StatusCode() != http.StatusOK {
-		log.Warn().Str("status", res.Status()).Int("code", res.StatusCode()).Msg("Failed to send test connection to API")
+	if statusCode != http.StatusOK {
+		log.Warn().Str("status", string(status)).Int("code", statusCode).Msg("Failed to send test connection result to API")
```
disq, line 108: There's a tiny inconsistency fix here - `res.Status()` is the response message but `string(status)` is the status of the test conn we've tried updating it to.

erezrokah:
What's the reason this was a `chore`? Should have been a `fix` no?

cli/cmd/test_connection.go
```diff
@@ -72,22 +72,42 @@ func updateSyncTestConnectionStatus(ctx context.Context, logger zerolog.Logger,
...
+			requestBody.FailureCode = &failedTestResult.FailureCode
+			requestBody.FailureReason = &failedTestResult.FailureDescription
+		}
+		res, err := apiClient.UpdateSyncTestConnectionForSyncSourceWithResponse(ctx, teamName, syncTestConnectionUUID, requestBody)
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to send sync test connection result to API")
+			return
+		}
+		statusCode = res.StatusCode()
+	default:
```
erezrokah, line 92: Maybe the default should be to do nothing? And on destination it should update the destination?
That way if we add other kinds (e.g. transformer) this won't break

disq: Handled in https://github.com/cloudquery/cloudquery/pull/18790

PR #18794, opened 2024-07-31
feat: Add Hacker News source plugin cloud UI

rhino1998:
Adds a cloud plugin UI for the Hacker News source plugin


hermanschaaf (approved):
I pushed some small stylistic changes to your branch, but looks good! Let's see what it looks like once deployed, I think the only other thing will be to add a little Hacker News icon at the top, like we do for other plugins (we can do this in a follow-up though)

Screenshot 2024-08-01 at 09 07 29

PR #18797, opened 2024-07-31
chore: Enforce postgres UI config build in CI

erezrokah:
#### Summary

This PR enforces the CI added in https://github.com/cloudquery/cloudquery/pull/18714 passes before merging a PR to the PostgreSQL destination plugin


erezrokah:
Looks like the CI fails when running from a fork

PR #18827, opened 2024-08-02
chore: Add more govet lint rules

erezrokah:
#### Summary

Add more `govet` rules to the


plugins/destination/clickhouse/typeconv/arrow/values/unwrap.go
erezrokah, line 1: I think these are false positive

marianogappa (approved):

PR #18833, opened 2024-08-02
feat: Support table name changes on basic transformer.

marianogappa:
Adds a new transformation kind to the basic transformer that enables renaming tables.

Used like:

```yaml copy
      - kind: change_table_names
        tables: ["*"]
        new_table_name_template: "c_{{.OldName}}"
```

Unfortunately, `sync_v3` has to be updated to support this transformation, because the sync process deletes stale rows, and it supplies the old table names, and there's no way to ask the transformer how did you change the names.

Therefore, I added a very small `tableNameChanger` that learns table name changes by sniffing schema bytes on `TransformSchema` calls (via `MigrateTableMessage`), and updates the names where it needs to.

Screenshot 2024-08-02 at 15 03 53
Screenshot 2024-08-02 at 15 04 02
Screenshot 2024-08-02 at 15 04 27


plugins/transformer/basic/client/spec/spec.go
```diff
@@ -31,42 +38,56 @@ func (s *Spec) SetDefaults() {
...
 	var err error
 	for _, t := range s.TransformationSpecs {
-		switch t.Kind {
-		case KindRemoveColumns:
-			if len(t.Columns) == 0 {
-				err = errors.Join(err, fmt.Errorf("'columns' field must be specified for remove_columns transformation"))
-			}
-			if t.Name != "" {
-				err = errors.Join(err, fmt.Errorf("'name' field must not be specified for remove_columns transformation"))
+		requiredFields, ok := kindToRequiredFields[t.Kind]
```
erezrokah, line 53: Non blocking, but what's the reason to use reflection? I'm worried this can break when someone renames a field

plugins/transformer/basic/docs/overview.md
```diff
@@ -3,6 +3,7 @@ This CloudQuery transformer plugin provides basic transformation capabilities:
 - Removing columns
 - Adding literal string columns
 - Obfuscating string columns
+- Renaming tables using a name template
```
erezrokah, line 6: Maybe add the list of supported template variables here too (`{{.OldName}}` is the only one correct?)

cli/internal/tablenamechanger/table_name_changer.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"github.com/cloudquery/cloudquery/cli/internal/specs/v0"
+	"github.com/cloudquery/plugin-pb-go/pb/plugin/v3"
+	"github.com/cloudquery/plugin-sdk/v4/schema"
+)
+
+// TableNameChanger manages table name changes due to transformers, on a per-destination basis.
+// It can learn about table name changes and apply those changes on further steps.
+type TableNameChanger struct {
+	tableNameChanges map[string]map[string]string
```
erezrokah, line 12: Should this be concurrent safe?

erezrokah (commented):
Looks good, added a few comments

plugins/transformer/basic/client/spec/spec.go
marianogappa: Can you think of an elegant way of doing this without reflection? If someone renames a field this will panic in the tests.

cli/internal/tablenamechanger/table_name_changer.go
marianogappa: No need. There are 3 places that update it. Two are on the same goroutine, and the third happens after a `Wait` on that one.

marianogappa: But I can make it concurrency-safe 🤷 It's no big deal.

If you think about it, if we have to make it concurrency-safe, I think it's not safe, is it? 🤔 We don't want to update table names until they're all modified.

The per-destination multiplexing happens in loops, so it's all sequential.

plugins/transformer/basic/docs/overview.md
marianogappa: Changed to:

```
- Renaming tables using a name template (use `{{.OldName}}` to refer to the original name, see example below)
```

plugins/transformer/basic/client/spec/spec.go
erezrokah: I would keep the `switch/case` actually as it's easier to follow (and also not more code).

But if we want a generic way maybe a more declarative approach like https://pkg.go.dev/github.com/go-playground/validator/v10#section-readme (it uses reflection under the hood too I think)

erezrokah (approved):

plugins/transformer/basic/client/spec/spec.go
erezrokah: Again it's non blocking I think

marianogappa: Reverted to switch/case.

erezrokah (approved):

PR #18867, opened 2024-08-07
feat: Add logs to Datadog dashboard

erezrokah:
#### Summary

This adds a new section to our Datadog dashboard with information on throttled APIs:
image

Clicking on a column will users open the logs in the relevant location:
image
image

I've also added the the same change to https://github.com/DataDog/integrations-extras/pull/2451.

https://github.com/DataDog/integrations-extras/pull/2451 also adds custom log pipelines to extract the retry attempts into attributes instead of relying on the `retrying request` message.
However if we only distribute the dashboard (like we do on our website) we can't ship the pipelines with the dashboard (we need the integration for that).


website/public/assets/datadog-dashboard.json
```diff
@@ -145,7 +145,41 @@
...
+            "formulas": [{ "formula": "a" }],
+            "queries": [
+              {
+                "data_source": "logs",
+                "name": "a",
+                "indexes": ["*"],
+                "compute": { "aggregation": "count" },
+                "group_by": [],
+                "search": {
+                  "query": "(@retry.attempt:>=2 OR \"retrying request\") @service.name:$service.value"
```
erezrokah, line 171: `retrying request` query works on AWS and all plugins that use https://github.com/hashicorp/go-retryablehttp with https://github.com/cloudquery/plugin-sdk/blob/25ed3d25a529a22f351ab92e22fb03a19c9557d4/helpers/retryablehttp/logger.go

`@retry.attempt:>=2` works out of the box for GCP since we send a `retry.attempt` as a log field.

For Azure we need a log pipeline added in https://github.com/cloudquery/integrations-extras/blob/47a83adf4d7aa921a2fd432a676a059ce436fdee/cloudquery/assets/logs/cloudquery.source.azure.yaml#L18

marianogappa (approved):

erezrokah:
@marianogappa added another commit to show permission errors https://github.com/cloudquery/cloudquery/pull/18867/commits/083921aa94535128f926cf76ce57d8da5f3078f1 (also updated the title of the Dashboard to match the integration)

Issue #18039, opened 2024-05-22
Unable to use incremental sync in local - failed to sync unmanaged client: failed to create state client: rpc error


Prabhanjali:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I am using AWS Plugin 24.1.0, Postgresql version 7.0.0, AWS CLI is 5.20.1.
While trying to use a incremental sync for aws_inspector2_findings table, I am running into the below issues.

**Error:**
```
 unexpected error from sync client receive: rpc error: code = Unavailable desc = failed to sync records: failed to sync unmanaged client: failed to create state client: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp: address: too many colons in address"
```

### Expected Behavior

Incremental sync to work without error

### CloudQuery (redacted) config

source.yml
```
kind: source
spec:
  # Source spec section
  name: aws_xxxxxx
  path: cloudquery/aws
  version: "v24.1.0"
  tables: ["aws_inspector2_findings"]
  destinations: ["postgresql"]
  backend_options:
    table_name: "cq_state_aws"
    connection: "@@plugins.postgresql.connection"

```

postgresql.yml

```
kind: destination
spec:
  name: postgresql
  path: "cloudquery/postgresql"
  version: "v7.0.0"
  write_mode: "overwrite-delete-stale"
  migrate_mode: "forced"
  spec:
    connection_string: "xxxxxxxxxx"
```

### CloudQuery (redacted) logs

```
..
```

### CloudQuery version

5.20.1

### Pull request (optional)

- [ ] I can submit a pull request

erezrokah:
Looks like we're using a unix only feature for the state client:
https://github.com/cloudquery/plugin-pb-go/blob/d34a2c510353ebcf4c8b7fc655e59f1a7ae52711/managedplugin/plugin.go#L255
https://github.com/cloudquery/cloudquery/blob/85f0d9b82012afa58b6b783b41a01957422b40f0/cli/cmd/sync_v3.go#L118

Where we should be doing https://github.com/cloudquery/plugin-pb-go/blob/d34a2c510353ebcf4c8b7fc655e59f1a7ae52711/managedplugin/plugin.go#L514

PR #18880, opened 2024-08-08
feat: Enable StateClient usage under docker plugins.

marianogappa:
`StateClient`s currently don't work in docker-based plugins because they cannot connect to destination sockets.

This CLI change detects that a source has:
- the `Docker` registry setting
- the `backend_options` set

And in this case it makes the following config changes:
- Forces destinations to use TCP instead of sockets
- Adds the `host.docker.internal` extra host to the docker create configuration
- Sniffs the source's variables and replaces instances of localhost with this special hostname

As a result, normal syncs continue to work exactly the same as before, but syncs that suffered from this config problem now work seamlessly with the new settings:

Screenshot 2024-08-08 at 16 33 32
Screenshot 2024-08-08 at 16 36 32

(Note that I'm using an unreleased pb-go replace)


cli/cmd/sync.go
```diff
@@ -195,6 +200,9 @@ func sync(cmd *cobra.Command, args []string) error {
 			managedplugin.WithTeamName(teamName),
 			managedplugin.WithLicenseFile(licenseFile),
 		}
+		if isThereADockerSourcePluginWithBackend {
```
erezrokah, line 203: Shouldn't we do this only for destinations that are used as state?

erezrokah: Also a follow PR can be to update the logic to consider if the OS is Windows per https://github.com/cloudquery/cloudquery/issues/18039

marianogappa: I made the following changes:
- UseTCP is more strict; only enabling it when the destination is used as state
- I propagated the state that a source `shouldReplaceLocalhost` all the way to `ReplaceVariables`. Before, it was replacing if a source had a `RegistryDocker`, which is not ideal because it doesn't necessarily need to use state. I don't think it's a bug because if it doesn't use a backend it probably doesn't need any replacing, but it's more correct now.
- Added tests where possible.

Regarding the Windows edge case, I'd fix it separately (in another PR, as you say)

erezrokah (approved):

PR #18906, opened 2024-08-09
feat: Update filetypes

erezrokah:
#### Summary

So we can create an RC with https://github.com/cloudquery/filetypes/pull/567


erezrokah:
Done in https://github.com/cloudquery/cloudquery/pull/18917

PR #18913, opened 2024-08-12
fix: Update XKCD and hackernews to fix tracking in cloud UI

jackcloudquery:
Enhanced eslint rules for both plugins


hermanschaaf (approved):
Maybe we should set up some automation to keep the linting (and other) config files that are common across plugins in sync?

jackcloudquery:
> Maybe we should set up some automation to keep the linting (and other) config files that are common across plugins in sync?

yeah, that's a good idea, I will think about something

PR #18963, opened 2024-08-19
fix: Update UI Scaffold

cduflo:


erezrokah (approved):

jackcloudquery (approved):

erezrokah:
> But this is just a scaffold, new version of what needs to be triggered?

If we want this change to be available via `brew` or GitHub Actions releases we need to publish a new release. See here https://docs.cloudquery.io/docs/developers/creating-new-plugin/go-source#initializing-your-plugin-with-the-scaffold-tool how this tool is consumed

erezrokah:
Also https://github.com/cloudquery/cloudquery/releases/tag/scaffold-v3.3.0

jackcloudquery:
> > But this is just a scaffold, new version of what needs to be triggered?
> 
> If we want this change to be available via `brew` or GitHub Actions releases we need to publish a new release. See here https://docs.cloudquery.io/docs/developers/creating-new-plugin/go-source#initializing-your-plugin-with-the-scaffold-tool how this tool is consumed

Oh, I see, okay, I will leave to @cduflo to update

PR #18974, opened 2024-08-20
chore: Skip sending events for internal teams

rhino1998:
#### Summary

This skips emitting events if the team is marked internal. To do this we have to fetch the team internal flag on login and store it in the configuration.

Needs https://github.com/cloudquery/cloudquery-api-go/pull/224


cli/internal/analytics/client.go
```diff
@@ -20,9 +20,10 @@ var (
 )
 
 type eventDetails struct {
-	user        cqapi.User
-	currentTeam string
-	environment string
+	user                cqapi.User
+	currentTeam         string
+	currentTeamInternal bool
```
marianogappa, line 25: I would call this `isCurrentTeamInternal`. It's just clearer what it is when the type is not immediately visible.

marianogappa (approved):

PR #18985, opened 2024-08-22
fix: Better test connection error codes

disq:
~Still not great, but a small improvement to get more error codes out.~ I think this should cover most cases now.

### DIFF
--- a/plugins/destination/mysql/client/test_connection.go
+++ b/plugins/destination/mysql/client/test_connection.go
@@ -7,6 +7,11 @@ import (
 	"github.com/rs/zerolog"
 )
 
+const (
+	codeInvalidSpec   = "INVALID_SPEC"
+	codeConnectFailed = "CONNECT_FAILED"
+)
+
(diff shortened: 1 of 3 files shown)


plugins/destination/mysql/client/test_connection.go
```diff
@@ -8,21 +8,13 @@ import (
 	"github.com/rs/zerolog"
 )
 
-const (
```
savme, line 11: what's the reason for removing constants?

savme: I'd say it's better to use constants instead of duplicating strings in tests. However, if you feel differently, we can go ahead and merge it

savme (approved):

plugins/destination/mysql/client/test_connection.go
disq: Readded in https://github.com/cloudquery/cloudquery/pull/18985/commits/588ddd7fac4a971d712faca6eb49c4fcfd134353

hermanschaaf (approved):

PR #18991, opened 2024-08-22
chore: UI FormWrapper

cduflo:
- [x] needs bump https://github.com/cloudquery/plugin-config-ui-connector/pull/4
- [x] needs bump https://github.com/cloudquery/plugin-config-ui-lib/pull/25


erezrokah:
Maybe we should make this a `chore:` if it's a UI only fix? If it needs to be a `fix:` can we provide a more description changelog message via the PR title?

cduflo:
> Maybe we should make this a `chore:` if it's a UI only fix? If it needs to be a `fix:` can we provide a more description changelog message via the PR title?

yes - thanks for the reminder @erezrokah  👍

jackcloudquery (approved):

PR #18998, opened 2024-08-23
feat: Add display name support to XKCD Source UI

rhino1998:
#### Summary

This adds support for reconfiguring the display name to the XKCD source plugin config UI


rhino1998:
Covered by #19032

PR #18999, opened 2024-08-23
feat: Add display name support to Hacker News Source UI

rhino1998:
#### Summary

This allows configuration of the display name in the Hacker News source plugin ui


erezrokah:
Can we make this a `chore:`? I think since we have https://github.com/cloudquery/cloudquery/pull/18965 UI only changes are deployed on merges to `main` and don't need to trigger a new release. We make those into a `chore` to avoid adding change log entries to the release only for UI changes

rhino1998:
Covered by #19032

PR #19016, opened 2024-08-26
feat: Add init command

erezrokah:
#### Summary

Part of https://github.com/cloudquery/cloudquery-issues/issues/2312 (internal issue).
This PR adds an `cloudquery init` command to automatically generate a spec file based on the source and destination plugins the user choses.

I'll create a [release candidate](https://github.com/cloudquery/cloudquery/releases/tag/cli-v6.4.1-rc1) with this in a bit cc @pilvikala.

TODO:

- [x] Add tracking events
- [x] Fix Windows tests (Decided to skip the tests that require input prompts on Windows for simplicity)


cli/cmd/init.go
```diff
@@ -0,0 +1,344 @@
...
+
+func defaultConfigForPlugin(plugin cqapi.ListPlugin) *strings.Builder {
+	sb := strings.Builder{}
+	sb.WriteString("kind: " + string(plugin.Kind) + "\n")
+	sb.WriteString("spec:\n")
+	sb.WriteString("  name: " + plugin.Name + "\n")
+	sb.WriteString("  path: " + plugin.TeamName + "/" + plugin.Name + "\n")
+	sb.WriteString("  version: " + *plugin.LatestVersion + "\n")
+	return &sb
+}
```
marianogappa, line 143: I think it's more maintainable to make this more readable, but I leave that to your discretion, I don't know how much we expect this to change.

I'd have two separate functions for source & destination (so that you don't have the source spec in two different places, and they can evolve separately in the future at the cost of a little repetition now), and use text/template instead of a string builder.

If you want to make the change, here's the rewrite for this one:

```go
func defaultConfigForPlugin(plugin cqapi.ListPlugin) *strings.Builder {
	tmpl := `kind: {{.Kind}}
spec:
  name: {{.Name}}
  path: {{.TeamName}}/{{.Name}}
  version: {{.LatestVersion}}
`
	var buf bytes.Buffer
	t := template.Must(template.New("config").Parse(tmpl))
	t.Execute(&buf, plugin)

	sb := strings.Builder{}
	sb.WriteString(buf.String())
	return &sb
}
```

cli/cmd/init.go
```diff
@@ -0,0 +1,344 @@
...
+		successful.Println("Sync spec file generated successfully!")
+		fmt.Println("You can now run the sync with the following command:")
+		bold.Printf("cloudquery sync %s\n", specPath)
+	} else {
+		successful.Println("Sync spec file generated successfully!")
+		fmt.Println("To run the sync please login with the following command:")
+		bold.Println("cloudquery login")
+		fmt.Println("Then you can run the sync with the following command:")
+		bold.Printf("cloudquery sync %s\n", specPath)
+	}
```
marianogappa, line 342: So assuming that this tool's purpose is to make it as easy as possible for anyone to configure a sync, it would be grand to provide a link to the docs for the chosen plugins right here, because this tool makes you not have to think about how to configure the sync, but you still have to do the auth and that still makes you go to the same page to learn how.

marianogappa (approved):
Have you given sufficient thought to the command name "init"? In principle I find it a little confusing, like it's something one is supposed to run when first installing the cli.

I'd be looking for `make-spec` or something, but I can see some people disliking a long name. As long as we've given it some thought it's ok 👍

cli/cmd/init.go
erezrokah: Good idea, refactored in https://github.com/cloudquery/cloudquery/pull/19016/commits/ac53790416a338fd30fa5ebc0b75766d6267650f and added tests

cli/cmd/init.go
erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/19016/commits/38a34dea80ec3b91f9db3e469c3f5dcd8ca2e0f6, also brought up in https://github.com/cloudquery/cloudquery-issues/issues/2312#issuecomment-2311879206

erezrokah:
> Have you given sufficient thought to the command name "init"? In principle I find it a little confusing, like it's something one is supposed to run when first installing the cli.
> 
> I'd be looking for `make-spec` or something, but I can see some people disliking a long name. As long as we've given it some thought it's ok 👍

I was following:
- https://vercel.com/docs/cli/init
- https://docs.npmjs.com/cli/v10/commands/npm-init
- https://cli.netlify.com/commands/init/

Neither one is required on first install

pilvikala:
I think `init` is good. Here's another example from Terraform:
> The terraform init command initializes a working directory containing configuration files and installs plugins for required providers.

PR #19025, opened 2024-08-28
chore: Associate cloud analytics with team service account

rhino1998:
#### Summary

This causes analytics to be emitted using the team service account user ID/email when run in the cloud environment


erezrokah (approved):
Looks good 🚀 We can keep this as `chore:` to avoid spamming the change log

erezrokah:
> Looks good 🚀 We can keep this as `chore:` to avoid spamming the change log

Will be released in https://github.com/cloudquery/cloudquery/pull/19050

PR #19034, opened 2024-08-30
feat: Implement transformations in migrate cmd.

marianogappa:
fixes https://github.com/cloudquery/cloudquery-issues/issues/2370

When I implemented transformer plugins, I simply didn't even know that the `migrate` command existed honestly. The reason that the migrate command doesn't apply transformations is because the `migrate` command is an entirely separate command, and only knows about sources & destinations.

This PR is more or less a copy-pasta of the sync_v3 code, which:
- Instantiates transformer plugins
- Upon sending the `MigrateTable` message, applies `TransformSchema` for all existing transformers onto the schema.

Note that there can be multiple destinations, and an unknown heterogeneous topology of transformer pipelines for each (in case the code looks tricky).

I ran some tests that show how `cloudquery migrate .` now applies transformations:
Screenshot 2024-08-30 at 10 20 16
Screenshot 2024-08-30 at 10 30 26

But when there are no transformers, it continues to work properly:
Screenshot 2024-08-30 at 10 19 50
Screenshot 2024-08-30 at 10 19 32

### DIFF
--- a/cli/cmd/migrate_v3.go
+++ b/cli/cmd/migrate_v3.go
@@ -13,18 +13,30 @@ import (
 )
 
 // nolint:dupl
-func migrateConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, destinationsClients managedplugin.Clients, sourceSpec specs.Source, destinationSpecs []specs.Destination) error {
+func migrateConnectionV3(ctx context.Context, sourceClient *managedplugin.Client, destinationsClients managedplugin.Clients, sourceSpec specs.Source, destinationSpecs []specs.Destination, transformersForDestination map[string][]*managedplugin.Client, transformerSpecsByName map[string]specs.Transformer) error {
 	destinationStrings := make([]string, len(destinationSpecs))
 	for i := range destinationSpecs {
 		destinationStrings[i] = destinationSpecs[i].VersionString()
 	}
+	transformerStrings := []string{}
+	for _, transformerSpec := range transformerSpecsByName {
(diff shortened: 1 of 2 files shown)


cli/cmd/migrate_v3.go
```diff
@@ -40,7 +52,7 @@ func migrateConnectionV3(ctx context.Context, sourceClient *managedplugin.Client
 			opts = append(opts, transformer.WithRemovePKs())
 			opts = append(opts, transformer.WithCQIDPrimaryKey())
 		}
-		destinationTransformers[i] = transformer.NewRecordTransformer(opts...)
+		destinationRecordTransformers[i] = transformer.NewRecordTransformer(opts...)
```
marianogappa, line 55: These are unrelated (we already had a concept of RecordTransformer that adds `_cq_*` fields. I renamed them to make it _less_ confusing.

bbernays: Can we remove the old record transformers and move them to the new transformers? Maybe we can have "default" transformers?

marianogappa: Yes, I can do that. But that requires non-trivial changes in `sync_v3` too. I'd prefer to do that separately fi that's ok?

marianogappa: (added issue https://github.com/cloudquery/cloudquery-issues/issues/2386)

erezrokah (approved):

PR #19067, opened 2024-09-03
feat: Add option to use Appender API

disq:
https://github.com/cloudquery/cloudquery/pull/16668 continued (with updated library support) which had to be scrapped before ([PR](https://github.com/cloudquery/cloudquery/pull/16870)) due to version incompatibility.

This time I've added this as an option, if it works (maps are unsupported in the API*, structs are represented as strings due to language limitations), it could be significantly faster than encoding and decoding from parquet.

* Maps are apparently supported as of https://github.com/marcboeker/go-duckdb/pull/237 but they haven't been implemented in this PR.


plugins/destination/duckdb/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/cloudquery/codegen v0.3.19
 	github.com/cloudquery/plugin-sdk/v4 v4.60.0
 	github.com/google/uuid v1.6.0
-	github.com/marcboeker/go-duckdb v1.7.0
+	github.com/marcboeker/go-duckdb v1.7.2-0.20240830073305-42d190898463
```
disq, line 11: This includes a [fix](https://github.com/marcboeker/go-duckdb/commit/42d190898463ad2b407151ebc05156850fd29591) to have better error handling/messages during `appender.Close()`.

plugins/destination/duckdb/client/write.go
```diff
@@ -115,6 +116,54 @@ func (c *Client) copyFromFile(ctx context.Context, tableName string, fileName st
...
+
+	tableState, err := c.getCacheTableInfo(ctx, table.Name)
+	if err != nil {
+		return fmt.Errorf("table %s not found in duckdb: %w", table.Name, err) // should never happen as appender would've failed
+	}
+
+	fields := msgs[0].Record.Schema().Fields()
+
+	for _, msg := range msgs {
+		arr, err := transformRecordToGoType(msg.Record, fields, tableState.Columns)
```
disq, line 153: `appender.AppendRow` gets a `driver.Value` so we need to convert Arrow records to Go types.

erezrokah (commented):
Thanks for the PR @disq, does it fix a user reported issue? Asking as it's not clear from the docs what's the advantage of enabling the appender spec option.

Regardless, could we add a test that verifies the same data is written with/without the option?

disq:
It's a way of loading bulk data into duckdb via API ([api docs](https://duckdb.org/docs/data/appender.html))
I'm not sure if it's really faster than writing to + loading from parquet, but in theory (since all the intermediary steps involved are in memory) it should be.

I will add a test later.

erezrokah:
Going to close this as stale, at least until we get reports on performance issues with DuckDB. Please re-open if you think we still need this with benchmarks of how it improves on our current approach

PR #19095, opened 2024-09-05
fix: Use per OS login command in quick start guide

erezrokah:
#### Summary

In our quick start guides we tell users to run `cloudquery login`. However we use the same command for all operating systems. `cloudquery login` won't work for Linux (unless added to the PATH) or Windows
image

This PR fixes the issue by using `cloudquery login` on macOS, `./cloudquery login` on Linux and `cloudquery.exe login` on Windows. The fix is consistent with how we tell users to run a sync a bit below in the guide:

#### macOS
image
#### Linux
image
#### Windows
image

PR #19099, opened 2024-09-05
refactor: Rename Linux and Windows configure sync command imports

erezrokah:
#### Summary

Some cleanup after https://github.com/cloudquery/cloudquery/pull/19095 to rename imports to be consistent with https://github.com/cloudquery/cloudquery/blob/b7c965cf8e31e05ff4b1e7477e90240db897fcec/website/components/mdx/_configure-macOS.mdx?plain=1#L16


erezrokah:
Actually going to close this and do another cleanup

PR #19100, opened 2024-09-05
refactor: Inline quick start guide configure components

erezrokah:
#### Summary

At the moment the quick start guide is split into many components making it harder to update it. This was done to support multiple operating systems via components composition. However we can use the same approach as in https://github.com/cloudquery/cloudquery/pull/19095 (a `switch/case` to handle the logic) instead.
That allows us to have a single component with all the guide, making it easier to update


website/components/mdx/_configure.mdx
```diff
@@ -0,0 +1,150 @@
...
+        case "windows":
+            return <ConfigureSyncCommandWindow />
+        case "macos":
+            return <ConfigureSyncCommandMacos />
+        default:
+            return <ConfigureSyncCommandMacos />
+    }
+}
+
+## Configure Destination Plugin
```
erezrokah, line 19: So instead of something like 
image

Where the only difference between operating systems is the import for `configure-sync-command`, I inlined everything into a single component. 

The logic to handle multiple operating systems is via the `SyncCommand` component switch case

website/components/mdx/_configure.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+You can find authentication for other plugins in their corresponding [READMEs](https://github.com/cloudquery/cloudquery/tree/main/plugins/source).
+
+## Start Syncing
+
+Now you can start syncing the data from your source plugins to the specified destinations.
+
+You can run the sync by specifying the directory with the configuration files or each file individually
+
+<SyncCommand os={os} />
```
erezrokah, line 145: This will show the correct version of the sync command per OS

website/components/mdx/_configure.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+You can find authentication for other plugins in their corresponding [READMEs](https://github.com/cloudquery/cloudquery/tree/main/plugins/source).
+
+## Start Syncing
+
+Now you can start syncing the data from your source plugins to the specified destinations.
+
+You can run the sync by specifying the directory with the configuration files or each file individually
+
+<SyncCommand os={props.os} />
```
erezrokah, line 145: This will show the correct version of the sync command per OS

marianogappa (approved):

PR #19103, opened 2024-09-05
chore: MySQL UI generalization

cduflo:
#### Summary


plugins/destination/mysql/cloud-config-ui/src/form/index.tsx
```diff
@@ -1,188 +1,57 @@
...
-    await handleSubmit(getCurrentValues());
-  };
-
-  const onSubmit = handleFormSubmit(handleTestConnection);
-
-  const parsedTestConnectionError = useMemo(
-    () => (testConnectionError ? parseTestConnectionError(testConnectionError) : undefined),
-    [testConnectionError],
-  );
+  const getCurrentValues = useCallback(() => prepareSubmitValues(form.getValues()), [form]);
```
jackcloudquery, line 40:
```suggestion
  const getCurrentValues = useCallback(() => prepareSubmitValues(getValues()), [getValues]);
```

plugins/destination/mysql/cloud-config-ui/src/utils/formSchema.ts
```diff
@@ -1,88 +1,135 @@
...
+          .default(connectionObj.host ?? ''),
+        port: yup
+          .string()
+          .max(5)
+          .matches(/^($)|(\d+)$/, 'Port must be a number')
+          .default(connectionObj.port ?? ''),
+        database: yup
+          .string()
+          .max(63)
+          .default(connectionObj.database ?? ''),
```
jackcloudquery, line 57: None of those should be required?

jackcloudquery (approved):
Awesome!

plugins/destination/mysql/cloud-config-ui/src/utils/formSchema.ts
cduflo: Per the go-sql-driver docs, it appears that a DNS string of `/` is valid, making all segments optional.
It is a little confusing in that it states both:
`Except for the databasename, all values are optional. So the minimal DSN is: /dbname`
and 
`If you do not want to preselect a database, leave dbname empty: /`
So the only value that is documented to be both required and optional is dbname 😂 

My take is that all can be optional and the test connection will be the source of truth if the connection gets made. Your thoughts?

ref: https://github.com/go-sql-driver/mysql?tab=readme-ov-file#dsn-data-source-name

cduflo: @jackcloudquery  - sorry, forgot to tag you

jackcloudquery: Agreed

jackcloudquery (approved):

PR #19105, opened 2024-09-06
fix: Close Transformers pipeline on both source and destination closes

murarustefaan:


cli/internal/transformerpipeline/pipeline.go
```diff
@@ -3,10 +3,12 @@ package transformerpipeline
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 
-	"github.com/cloudquery/plugin-pb-go/pb/plugin/v3"
 	"golang.org/x/sync/errgroup"
+
```
disq, line 10:
```suggestion
```
Would need reordering imports though

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -68,6 +70,11 @@ func (lp *TransformerPipeline) Send(data []byte) error {
 	if lp.clientWrappers[len(lp.clientWrappers)-1].nextSendFn == nil {
 		return errors.New("OnOutput must be registered before Send is called, otherwise what do I do with the transformed data?")
 	}
+
+	if lp.clientWrappers[0].closed {
+		return fmt.Errorf("cannot send data to a closed pipeline")
```
disq, line 75:
```suggestion
		return errors.New("cannot send data to a closed pipeline")
```

disq (commented):
Human linting :)

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -91,6 +103,7 @@ type clientWrapper struct {
 	client     plugin.Plugin_TransformClient
 	nextSendFn func(*plugin.Transform_Request) error
 	nextClose  func() error
+	closed     bool
```
marianogappa, line 106: how about `isClosed` instead?

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -82,6 +87,13 @@ func (lp *TransformerPipeline) OnOutput(fn func([]byte) error) error {
 }
 
 func (lp *TransformerPipeline) Close() error {
+	// Closing the pipeline happens on both source as well as destination close.
```
marianogappa, line 90: I think we should change this function signature to `Close()`, that is, not return `error`.

I honestly don't understand what it means for `CloseSend()` to return an error, or what I'd do differently in either case, but we're not gonna do anything, because once `closed` is `true` we won't call it again.

This also simplifies `sync_v3`, because we don't have to check the `error`. I think it's wrong to `return err` where the pipeline is being closed anyway.

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -82,6 +87,13 @@ func (lp *TransformerPipeline) OnOutput(fn func([]byte) error) error {
...
+	// Not handling this will result in a close of closed channel panic.
+	if lp.clientWrappers[0].closed {
+		return nil
+	}
+	lp.clientWrappers[0].closed = true
```
marianogappa, line 95: I'd `ClosedSend` before setting this to `true`, and ignore the error.

cli/internal/transformerpipeline/pipeline.go
disq: Can also use `sync.Once`?

murarustefaan: Done

cli/internal/transformerpipeline/pipeline.go
murarustefaan: Done

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -81,16 +86,24 @@ func (lp *TransformerPipeline) OnOutput(fn func([]byte) error) error {
...
+func (lp *TransformerPipeline) Close() {
+	// Closing the pipeline happens on both source as well as destination close.
+	// Not handling this will result in a close of closed channel panic.
+	if lp.clientWrappers[0].isClosed {
+		return
+	}
+
 	// Close the first transformer. The rest will follow gracefully, otherwise records will be lost.
-	return lp.clientWrappers[0].client.CloseSend()
+	lp.clientWrappers[0].client.CloseSend()
```
erezrokah, line 97:
```suggestion
	_ = lp.clientWrappers[0].client.CloseSend()
```

You might need this to tell the linter you're ignoring the error from `CloseSend()`

erezrokah (approved):

cli/internal/transformerpipeline/pipeline.go
erezrokah: But you can ignore this if the linter passes

PR #19123, opened 2024-09-07
fix(main): Write to mysql in bulk

### LINKED ISSUE #19122: bug: mysql destination does not actually bulk write to the database

### Current Behavior

When syncing thousands of rows into a table, it is noticeably slow. On a MySQL server, you can execute `show full processlist;` and get a snapshot of queries running at that moment in time. Using this query, you can monitor the types of `INSERT` queries being run by CloudQuery. Using this, you can confirm that the plugin is only inserting one row at a time and not bulk inserting even though the plugin claims to bulk write.

https://github.com/cloudquery/cloudquery/blob/40cfc754055aa04d2e73456e97076273ec0bb83c/plugins/destination/mysql/client/spec.go#L23

Here is an example query being run by the MySQL plugin: (the tables are from a plugin i am working on)

```
INSERT INTO `jamf_computer_inventory_services` (`_cq_sync_time`, `_cq_source_name`, `_cq_id`, `_cq_parent_id`, `computer_id`, `name`) VALUES (?,?,?,?,?,?) ON DUPLICATE KEY UPDATE `_cq_sync_time` = VALUES(`_cq_sync_time`), `_cq_source_name` = VALUES(`_cq_source_name`), `_cq_id` = VALUES(`_cq_id`), `_cq_parent_id` = VALUES(`_cq_parent_id`), `computer_id` = VALUES(`computer_id`), `name` = VALUES(`name`)
```

### Expected Behavior

You can run bulk inserts by simply providing multiple value blocks separated by commas.

```
INSERT INTO `jamf_computer_inventory_services` (`_cq_sync_time`, `_cq_source_name`, `_cq_id`, `_cq_parent_id`, `computer_id`, `name`)
VALUES (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), (?,?,?,?,?,?), ...
ON DUPLICATE KEY UPDATE `_cq_sync_time` = VALUES(`_cq_sync_time`), `_cq_source_name` = VALUES(`_cq_source_name`), `_cq_id` = VALUES(`_cq_id`), `_cq_parent_id` = VALUES(`_cq_parent_id`), `computer_id` = VALUES(`computer_id`), `name` = VALUES(`name`)
```

kind: destination
spec:
  name: "mysql"
  path: "cloudquery/mysql"
  registry: "cloudquery"
  version: "v5.2.7"
  spec:
    connection_string: "..."

### Steps To Reproduce

1. Pick a source that will lead to thousands of records being inserted into your destination
2. Watch queries being executed using `show full processlist`
3. See that the query is not inserting in bulk, it is inserting one-by-one.

unnecessary

### CloudQuery version

6.5.0

### Additional Context

i have a PR upcoming to fix this

### Pull request (optional)

jsifuentes:
#### Summary

Fixes #19122

Before, the `INSERT` query was built within the `appendTableBatch` and `overwriteTableBatch` functions. But we aren't yet aware of how many rows need to be inserted, so we don't know how many value blocks to add. Rows are iterated within `writeResources` to create the final records to insert, so this is an opportunity to calculate a row count. So I moved the query building to within `writeResources` and constructed the `INSERT` query based on the amount of rows to be inserted.

This reduced the time to write 27,000 records from 20 minutes to under 1 minute.


yevgenypats:
Hey @jsifuentes , Thanks so much for submitting this PR! There might some short delay for the review here as our team is at an offsite at the moment.

erezrokah (approved):
Thanks for the PR @jsifuentes  looks great. Sorry for the late review, I wanted to make sure we're not hitting query size limits. Looks like our batching limits prevent that from happening (unless someone increases them simnifically).

I also added a test to see, on my local machine the improvement of the test time was fro m `127.894s` to `39.591s` which is a great 🚀

erezrokah:
Released in https://github.com/cloudquery/cloudquery/releases/tag/plugins-destination-mysql-v5.2, thanks again @jsifuentes 💯

PR #19126, opened 2024-09-09
feat: Add new transform that adds a column with the current timestamp

bbernays:
#### Summary

This PR adds a new transformation type that allows users to add a column whose value is the timestamp at the point in time that the transformer handles the data


plugins/transformer/basic/client/spec/spec.go
```diff
@@ -7,10 +7,11 @@ import (
 )
 
 const (
-	KindRemoveColumns    = "remove_columns"
-	KindAddColumn        = "add_column"
-	KindObfuscateColumns = "obfuscate_columns"
-	KindChangeTableNames = "change_table_names"
+	KindRemoveColumns      = "remove_columns"
+	KindAddColumn          = "add_column"
```
erezrokah, line 11: I'll let @marianogappa comment as well, but I'd rather add support for placeholders in `add_column` like `{{NOW}}` or something like that, which is similar to other places we inject dynamic values (e.g. https://hub.cloudquery.io/plugins/destination/cloudquery/s3/latest/docs#overview-s3-spec)

bbernays: While I really like that idea, `add_column` only adds in a string value while we want this to be a timestamp column

erezrokah: We could support configuring a type and try to convert the value to it

erezrokah: Or if the value equals `{{NOW}}` use a timestamp type

marianogappa: We can go as deep as we want and I'm not really the one to answer how deep. That's the real question to answer here, and I think @bbernays implicitly answered it: a `now()` column is valuable, and _also_ interpolating `now()` in strings doesn't add much value (that's what I infer).

We'd have to document how that works, and add edge cases to the code. Happy to work on it if it's worth it, though.

plugins/transformer/basic/docs/overview.md
```diff
@@ -2,6 +2,7 @@ This CloudQuery transformer plugin provides basic transformation capabilities:
 
 - Removing columns
 - Adding literal string columns
+- Adding a column with the current timestamp
```
marianogappa, line 5: Should we clarify the semantics of the value here or somewhere? i.e. that the value is the microsecond-resolution current timestamp as defined by Go's `time.Now()`, executed at the time of syncing each batch? Perhaps that's too granular, but maybe at least that two rows may have a slightly different timestamp?

marianogappa (commented):
There's no validation for the `spec` of this `Kind` and no tests for this.

Also, just pointing out the risk of adding too many bespoke transformations. I'm sure you're aware of this point and decided this one was worth it; just stating that this risk exists.

plugins/transformer/basic/client/spec/spec.go
bbernays: I would prefer to avoid overloading the transformer types at this time, while we are still learning about how users want to use the functionality

plugins/transformer/basic/client/transformers/transformers.go
```diff
@@ -78,6 +81,13 @@ func AddLiteralStringColumnAsLastColumn(name, value string) TransformationFn {
 	}
 }
 
+func AddTimestampColumnAsLastColumn(name string) TransformationFn {
+	return func(record arrow.Record) (arrow.Record, error) {
+		t := time.Now()
```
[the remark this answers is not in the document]
bbernays: fixed in [2322d17](https://github.com/cloudquery/cloudquery/pull/19126/commits/2322d171761f673978d42ee1432dc8e8d4f30e7e)

plugins/transformer/basic/docs/overview.md
bbernays: I have updated the description in [2beda7f](https://github.com/cloudquery/cloudquery/pull/19126/commits/2beda7fd5e826522c534fa6c959e9b0035fe41c2)

plugins/transformer/basic/docs/_configuration.md
```diff
@@ -16,6 +16,9 @@ spec:
         tables: ["xkcd_comics"]
         name: "source"
         value: "xkcd"
+      - kind: add_current_timestamp_column
+        tables: ["xkcd_comics"]
+        name: "current_timestamp"
```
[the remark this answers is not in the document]
bbernays: fixed in [ae44766](https://github.com/cloudquery/cloudquery/pull/19126/commits/ae447664500add2dc23f4c2623cc281cbaeff981)

marianogappa (approved):

PR #19130, opened 2024-09-11
fix: Data race on transformers pipeline closed check

disq:
Fixes the data race on `isClosed`, which we introduced in https://github.com/cloudquery/cloudquery/pull/19105

```
2024-09-11T16:10:36Z INF top level table resolver started client=Client#4 invocation_id=0ed60c97-05cf-4365-9789-e9fe27561951 module=test-source table=test_paid_table
panic: send on closed channel

goroutine 45 [running]:
github.com/cloudquery/cloudquery/cli/internal/transformerpipeline.(*identityTransformer).Send(0x14000a26a20?, 0x10580de10?)
        /Users/me/go/src/github.com/cloudquery/cq/cli/internal/transformerpipeline/identity.go:24 +0x28
github.com/cloudquery/cloudquery/cli/internal/transformerpipeline.(*TransformerPipeline).Send(0x140000c4460, {0x14000a6ea80, 0x1d48, 0x2a80})
        /Users/me/go/src/github.com/cloudquery/cq/cli/internal/transformerpipeline/pipeline.go:76 +0xcc
github.com/cloudquery/cloudquery/cli/cmd.syncConnectionV3.func6()
        /Users/me/go/src/github.com/cloudquery/cq/cli/cmd/sync_v3.go:422 +0x96c
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /Users/me/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:78 +0x58
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
        /Users/me/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:75 +0x98


==================
WARNING: DATA RACE
Read at 0x00c00057c868 by goroutine 71:
  github.com/cloudquery/cloudquery/cli/internal/transformerpipeline.(*TransformerPipeline).Send()
      /Users/me/go/src/github.com/cloudquery/cq/cli/internal/transformerpipeline/pipeline.go:72 +0xb4
  github.com/cloudquery/cloudquery/cli/cmd.syncConnectionV3.func6()
      /Users/me/go/src/github.com/cloudquery/cq/cli/cmd/sync_v3.go:422 +0xcfc
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/me/go/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:78 +0x7c

[...]

PR #19165, opened 2024-09-16
fix: Fix transformers-related hang.

marianogappa:
fixes https://github.com/cloudquery/cloudquery-issues/issues/2436

Several different users have reported that, since transformers have been implemented, certain syncs are hanging forever.

The transformations implementation seems to work well in these cases:
- A sync completes with no errors
- A sync finishes with errors due to the source
- A sync finishes with errors due to the transformation

However, when a destination has an error, at least sometimes it hangs forever.

Each transformer is an implementation of a `grpc.BidiStreamingClient`, which uses these two methods:

```go
	Send(*Req) error
	Recv() (*Res, error)
```

Which in turn call these:
```go
	// SendMsg is generally called by generated code. On error, SendMsg aborts
	// the stream. If the error was generated by the client, the status is
	// returned directly; otherwise, io.EOF is returned and the status of
	// the stream may be discovered using RecvMsg.
	//
	// SendMsg blocks until:
	//   - There is sufficient flow control to schedule m with the transport, or
	//   - The stream is done, or
	//   - The stream breaks.
	//
	// SendMsg does not wait until the message is received by the server. An
	// untimely stream closure may result in lost messages. To ensure delivery,
	// users should ensure the RPC completed successfully using RecvMsg.
	//
	// It is safe to have a goroutine calling SendMsg and another goroutine
	// calling RecvMsg on the same stream at the same time, but it is not safe
	// to call SendMsg on the same stream in different goroutines. It is also
	// not safe to call CloseSend concurrently with SendMsg.
	//
	// It is not safe to modify the message after calling SendMsg. Tracing
	// libraries and stats handlers may use the message lazily.
	SendMsg(m any) error
	// RecvMsg blocks until it receives a message into m or the stream is
	// done. It returns io.EOF when the stream completes successfully. On
	// any other error, the stream is aborted and the error contains the RPC
	// status.
	//
	// It is safe to have a goroutine calling SendMsg and another goroutine
	// calling RecvMsg on the same stream at the same time, but it is not
	// safe to call RecvMsg on the same stream in different goroutines.
	RecvMsg(m any) error
```

Both `SendMsg` and `RecvMsg` can hang forever, and this is what users are reporting.

However we may want to mitigate this issue (let's say the flow of data or the orchestration for closing is wrong), it's probably gonna be fragile if we still rely on those methods not hanging.

Thus, what I've done is running those two methods (`Recv()` and `Send()`) asynchronously. The Pipeline thread will invoke these, but it will exit if someone closes the pipeline (by using a `select`).

I've tested running @disq 's reproduction test 100 times; it no longer hangs nor panics:
https://github.com/cloudquery/cloudquery/pull/19130#issuecomment-2348504017

**It correctly fails on a test error on insert configuration**
Screenshot 2024-09-16 at 12 46 59

**It correctly succeeds on a success test configuration**
Screenshot 2024-09-16 at 12 47 29

**It still doesn't swallow the last batch, as tested by running a stable sync multiple times and seeing the same resource count**
Screenshot 2024-09-16 at 12 57 17


erezrokah (approved):

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -70,31 +71,49 @@ func (lp *TransformerPipeline) Send(data []byte) error {
...
+	go func() {
+		err := lp.clientWrappers[0].client.Send(&plugin.Transform_Request{Record: data})
+		sendCh <- err
+	}()
+
+	select {
+	case err := <-sendCh:
+		return err
+	case <-time.After(1 * time.Second): // Check if pipeline is closed every second
+		if lp.clientWrappers[0].isClosed {
```
disq, line 90: There's still a data race here, the race detector catches it once in a while.

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -106,22 +125,45 @@ type clientWrapper struct {
...
+	}()
+
+	for {
+		select {
+		case <-time.After(1 * time.Second): // Check if pipeline is closed every second
+			if s.isClosed {
+				return s.nextClose()
+			}
+		case req, ok := <-recvCh: // Propagate records to next transformer
+			if !ok {
```
disq, line 156: Nothing seems to close `recvCh` (so `ok` is never false?)

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -106,22 +125,45 @@ type clientWrapper struct {
...
-		data, err := s.client.Recv()
-		if err == io.EOF {
-			err := s.nextClose()
-			return err
-		}
-		if err != nil {
-			return err
+
+	recvCh := make(chan *plugin.Transform_Request)
+	errCh := make(chan error)
```
disq, line 134: Again, ideally do a `defer close(errCh)` in the goroutine (but handle the closed case in the `select` below by using the `val, ok := <-errCh` syntax)

cli/internal/transformerpipeline/pipeline.go
marianogappa: You're right, I didn't even look into that; I was optimising for fixing the issue.

I replaced the `bool` with an `atomic.Bool` and now the races disappeared.

cli/internal/transformerpipeline/pipeline.go
marianogappa: That is a very good point. I removed this unreachable code.

cli/internal/transformerpipeline/pipeline.go
```diff
@@ -70,31 +71,49 @@ func (lp *TransformerPipeline) Send(data []byte) error {
...
-		return errors.New("cannot send data to a closed pipeline")
+		return nil
 	}
 
-	return lp.clientWrappers[0].client.Send(&plugin.Transform_Request{Record: data})
+	sendCh := make(chan error)
+
+	// Send can block forever (e.g. if grpc buffer is full), so we run it asynchronously
+	// and check if pipeline is closed every second.
+	go func() {
```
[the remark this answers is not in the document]
marianogappa: What is the benefit of explicitly closing this channel? Btw, this goroutine is never explicitly exited either; the `defer` might as well be outside of it.

cli/internal/transformerpipeline/pipeline.go
marianogappa: Same question regarding this close. What's the benefit of closing it?

disq (approved):

cli/internal/transformerpipeline/pipeline.go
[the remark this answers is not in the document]
disq: Is the `client.Send()` method blocking ad infinitum (provided no error was encountered?) I thought (well, the comment above says...) it would only block if grpc buffer is full, or such cases.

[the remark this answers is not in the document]
disq: Closing would be to stop having unused open channels (do they use file descriptor limits? probably not, but don't know)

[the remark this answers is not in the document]
marianogappa: The case that we're hitting is that:
- the source (via `n` workers) is sending non-stop
- the destination was fine but then errors
- the `send` doesn't get `recv`'d many times in a row until the buffer is full

But with my fix, within <= 1 second, the pipeline is closed.

This should be a single channel that is needlessly around for about 1 sec, so I think we're fine.

PR #19169, opened 2024-09-17
feat: Add sharding support

erezrokah:
#### Summary

We'll need to release a few plugins with https://github.com/cloudquery/plugin-sdk/pull/1891 first, hence the future date in the command description

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -42,7 +42,7 @@ func NewCmdSync() *cobra.Command {
 	cmd.Flags().String("license", "", "set offline license file")
 	cmd.Flags().String("summary-location", "", "Sync summary file location. This feature is in Preview. Please provide feedback to help us improve it.")
 	cmd.Flags().String("tables-metrics-location", "", "Tables metrics file location. This feature is in Preview. Please provide feedback to help us improve it. Works with plugins released on 2024-07-10 or later.")
-	cmd.Flags().String("shard", "", "Allows splitting the sync process into multiple shards. This feature is in Preview. Please provide feedback to help us improve it. Works with plugins released on 2024-09-24 or later.")
+	cmd.Flags().String("shard", "", "Allows splitting the sync process into multiple shards. This feature is in Preview. Please provide feedback to help us improve it. For a list of supported plugins visit https://docs.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel")
 
 	return cmd
 }
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -5,6 +5,7 @@ import (
 	"math"
 	"os"
 	"slices"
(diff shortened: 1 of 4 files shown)


cli/cmd/sync_v3.go
```diff
@@ -118,6 +123,9 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 	// https://github.com/golang/go/issues/41087
 	syncTime := time.Now().UTC().Truncate(time.Microsecond)
 	sourceName := sourceSpec.Name
+	if shard != nil {
+		sourceName = fmt.Sprintf("%s-%d/%d", sourceName, shard.num, shard.total)
```
erezrokah, line 127: This is needed to support `overwrite-delete-stale` otherwise shards delete each other's data at the end of the sync.

We can use a better name for it, maybe `%s-shard-%d/%d`

cli/cmd/sync.go
```diff
@@ -39,6 +42,7 @@ func NewCmdSync() *cobra.Command {
 	cmd.Flags().String("license", "", "set offline license file")
 	cmd.Flags().String("summary-location", "", "Sync summary file location. This feature is in Preview. Please provide feedback to help us improve it.")
 	cmd.Flags().String("tables-metrics-location", "", "Tables metrics file location. This feature is in Preview. Please provide feedback to help us improve it. Works with plugins released on 2024-07-10 or later.")
+	cmd.Flags().String("shard", "", "Allows splitting the sync process into multiple shards. This feature is in Preview. Please provide feedback to help us improve it. Works with plugins released on 2024-09-24 or later.")
```
marianogappa, line 45: This does not mention that it only works for plugins using the default scheduler (e.g. not S3), but I realise it's a little tricky to put this in words.

marianogappa (approved):

cli/cmd/sync.go
erezrokah: Yeah, it doesn't work with Docker plugins as well 🙃 

I think that will be in the update of https://docs.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel and https://docs.cloudquery.io/docs/deployment/github-actions

cli/cmd/sync_v3.go
erezrokah, line 127:
```suggestion
		sourceName = fmt.Sprintf("%s_%d_%d", sourceName, shard.num, shard.total)
```

erezrokah:
I'll merge this once we have at least AWS, GCP and Azure published with this support

cli/cmd/sync.go
erezrokah: OK so did https://github.com/cloudquery/cloudquery/pull/19169/commits/fff4d1d79a46f1eeff5702d1ca8f14970fe58fff

PR #19173, opened 2024-09-17
feat: Handle updates via `Merge`

bbernays:
#### Summary


plugins/destination/snowflake/client/write.go
```diff
@@ -77,6 +84,63 @@ func (c *Client) WriteTableBatch(ctx context.Context, name string, msgs message.
...
+	if _, err := c.db.ExecContext(ctx, sql); err != nil {
+		return fmt.Errorf("failed to merge file into table: %s: %w", sql, err)
+	}
+	return nil
+}
+
+func (c *Client) createColumnsList(table *schema.Table) string {
+	// creates a string like:
+	// $1:COL1::TEXT as COL1, $1:COL2::NUMBER as COL2, $1:COL3::TIMESTAMP_TZ as COL3
+	columnString := ""
```
disq, line 99: Maybe create it as a known-length slice and strings.Join at the end? Still takes the same amount of lines, but less allocs possibly? Looks simpler too.

columns := make([]string, len(table.Columns)
in loop: coumns[i] = fmt.Sprintf...
return strings.Join(columns, ",")

murarustefaan: done, changed all of them as it was easier to read when I needed to change the ON AND separator

murarustefaan:
I also increased the batch size to 5x as the difference even in our x-small warehouse was huge, going from 2.5s / 1000rows to 3.2s / 5000rows.

erezrokah (approved):

PR #19220, opened 2024-09-20
fix: Add missing OTEL logs component

erezrokah:
#### Summary

This fixes an issue with the `--tables-metrics-location` flag, that when you pass it you'd see the following log warnings:

```
2024-09-20T13:28:20Z WRN otel error error="failed to send logs to http://localhost:51875/v1/logs: 404 Not Found" invocation_id=95b765f7-2472-4c7a-b062-c0c799743bf7 module=cli
```

The reason is that now plugins send OTEL logs data as well, but the CLI is not set up to receive them. This was an oversight on my end when adding logs support to plugins.

The PR also cleans up some unused options

### DIFF
--- a/cli/internal/otel/receiver.go
+++ b/cli/internal/otel/receiver.go
@@ -172,7 +172,7 @@ func (Consumer) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
 }
 
 // ConsumeLogs implements consumer.Logs.
-func (c Consumer) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
+func (Consumer) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
 	// Do nothing, the CLI only needs metrics to print the table metrics file
 	return nil
 }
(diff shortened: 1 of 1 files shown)


cli/internal/otel/receiver.go
```diff
@@ -261,47 +296,30 @@ func getFreePort() (int, error) {
 
 type OtelReceiverOption func(*Consumer)
 
-func WithSpanConsumer(consumerSpan ConsumeSpan) OtelReceiverOption {
```
erezrokah, line 264: These are unused

cli/internal/otel/receiver.go
```diff
@@ -261,47 +296,30 @@ func getFreePort() (int, error) {
...
 	}
 
-	if c.consumeSpan == nil {
-		c.consumeSpan = newDefaultSpanConsumer()
+	for _, opt := range opts {
+		opt(&c)
 	}
 
-	if c.consumeMetric == nil {
-		if c.metricsFilename == "" {
```
erezrokah, line 293: `metricsFilename` is never empty. We always call `StartOtelReceiver` with a non empty metrics file

PR #19289, opened 2024-10-03
chore: Fix mssql failing CI

murarustefaan:


disq (approved):

maaarcelino (approved):

erezrokah:
I think we were testing on 2017 since that's the minimal version supported https://hub.cloudquery.io/plugins/destination/cloudquery/mssql/latest/docs#overview

image

Maybe we can drop the support for 2017, but move to 2019? Not sure.

See https://endoflife.date/mssqlserver

murarustefaan:
Yes, makes sense.

PR #19314, opened 2024-10-03
chore: Update MSSQL supported versions

murarustefaan:


erezrokah (commented):
Looks good but we should make this a breaking change if we drop official support

murarustefaan:
To me it does not seem we should as there are no breaking changes / configurations users need to do. But it makes sense though from user POV though so I'm gonna do that.

PR #19317, opened 2024-10-03
chore: Run tests in parallel, sync before read (file)

disq:
Getting the file plugin ready for the [StreamingBatchWriter update](https://github.com/cloudquery/plugin-sdk/pull/1921). ~Incorporates https://github.com/cloudquery/filetypes/pull/579 and https://github.com/cloudquery/plugin-sdk/pull/1921~


plugins/destination/file/go.mod
```diff
@@ -102,3 +102,7 @@ require (
 
 // github.com/cloudquery/jsonschema @ cqmain
 replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0-20240220124159-92878faa2a66
+
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.64.2-0.20241003143736-33b6e395606e
```
disq, line 106: Waiting for an SDK release.

plugins/destination/file/client/read.go
```diff
@@ -19,6 +19,11 @@ func (c *Client) Read(_ context.Context, table *schema.Table, res chan<- arrow.R
 		return fmt.Errorf("reading is not supported when `path` contains UUID variable. Table: %q", table.Name)
 	}
 	name := c.spec.ReplacePathVariables(table.Name, uuid.NewString(), time.Time{})
+
+	if syncAfterWrite {
+		time.Sleep(500 * time.Millisecond)
```
disq, line 24: I've used `assert.EventuallyWithT` in the cloud storage destinations (gcs/azblob/S3) but here we use the `TestInsert` functionality from the SDK (`testPlugin` - except for `parquet` type or compressed formats) and it needs to be able to see the rows immediately after the write call. For parquet or compressed formats, we use `testPluginCustom` which follows the `write / close / read` pattern.

murarustefaan (approved):

PR #19320, opened 2024-10-03
chore: Run tests in parallel, wait before read (S3)

disq:
Getting the S3 plugin ready for the [StreamingBatchWriter update](https://github.com/cloudquery/plugin-sdk/pull/1921). ~Incorporates https://github.com/cloudquery/filetypes/pull/579 and https://github.com/cloudquery/plugin-sdk/pull/1921~


plugins/destination/s3/go.mod
```diff
@@ -108,3 +108,7 @@ require (
 
 // github.com/cloudquery/jsonschema @ cqmain
 replace github.com/invopop/jsonschema => github.com/cloudquery/jsonschema v0.0.0-20240220124159-92878faa2a66
+
+replace github.com/cloudquery/plugin-sdk/v4 => github.com/cloudquery/plugin-sdk/v4 v4.64.2-0.20241003143736-33b6e395606e
```
disq, line 112: Waiting for an SDK release.

plugins/destination/s3/go.mod
```diff
@@ -108,3 +108,7 @@ require (
...
+
+replace github.com/cloudquery/filetypes/v4 => github.com/cloudquery/filetypes/v4 v4.4.1-0.20241003143110-56c4670c4f4f
```
disq, line 114: Waiting for a filetypes release, although probably not needed.

plugins/destination/s3/client/client_test.go
```diff
@@ -125,6 +127,8 @@ func testPluginCustom(t *testing.T, s *spec.Spec) {
 		t.Fatal(fmt.Errorf("failed to close client: %w", err))
 	}
 
+	time.Sleep(2 * time.Second)
```
savme, line 130: can we signal completion instead of waiting an arbitrary amount of time?

disq: But it has already been completed, code runs fine without the sleep if you're lucky (if S3 behaves well). This is to "unflake" the test. We could also mock S3 here, but that would probably go outside of the scope of this PR, which is to get the StreamingBatchWriter updates safely over the line.

savme: Got it! Totally makes sense to keep mocking out of scope of this PR.

Perhaps, something like [assert.Eventually](https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually) could be a better fit here? Feel free to disregard the suggestion and proceed with the merge, though

savme (approved):

plugins/destination/s3/client/client_test.go
disq: Good tip, implemented in all three PRs. Can't do it in the file PR unfortunately, without changing the test concept.

PR #19342, opened 2024-10-10
fix: Auto detect SVG content-type

disq:


cli/internal/publish/pluginui_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestQuickContentType(t *testing.T) {
+	tests := []struct {
+		filename string
+		data     []byte
+		want     string
+	}{
+		{
+			filename: "octetstream.svg",
+			data:     []byte(`<svg xmlns="http://www.w3.org/2000/svg" width="2500" height="330" viewBox="0 0 719.713 94.943"><g fill-rule="evenodd" clip-rule="evenodd"><path d="M17.32,16.76l-.53.53-1.09-1.1a4.33,4.33,0,0,0,.53-.52Z" style="fill:#4285f4"/></g></svg>`),
```
erezrokah, line 19: Can we embed these files instead to make it easier to read the tests?

cli/internal/publish/pluginui.go
```diff
@@ -120,14 +120,22 @@ func quickContentType(filename string) (string, error) {
 	}
 
 	contentType := http.DetectContentType(filebytes)
-	if contentType == "application/octet-stream" {
+	switch {
+	case contentType == "application/octet-stream":
```
erezrokah, line 124: We can use the `mime` package similar to https://github.com/golang/go/issues/15888#issuecomment-222457018 WDYT?

disq: Done in https://github.com/cloudquery/cloudquery/pull/19342/commits/a5579c5b3cfd96a495f3756dc62b78112ab7ffeb

cli/internal/publish/pluginui_test.go
disq: Done in https://github.com/cloudquery/cloudquery/pull/19342/commits/a5579c5b3cfd96a495f3756dc62b78112ab7ffeb

hermanschaaf:
Linting error in the latest commit:

```
 Error: cli/internal/publish/pluginui_test.go:55:64: copyTempFile - result 1 (error) is always nil (unparam)
  func copyTempFile(t *testing.T, src, destName string) (string, error) {
```

Issue #19338, opened 2024-10-08
bug: test-connection provides false positive for plugins where it is not implemented


jschafer-kp:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I call `test-connection` on the [Microsoft Entra ID plugin](https://hub.cloudquery.io/plugins/source/cloudquery/entraid/latest/docs), it returns a successful message
```
Test Connection Results
source       entraid (cloudquery/entraid@v1.4.3)  Success
```
However, when I look at the logs, I see this error message
```
2024-10-08T15:05:22Z ERR finished call grpc.code=Unimplemented grpc.component=server grpc.error="rpc error: code = Unimplemented desc = TestConnection feature is not implemented in this plugin" grpc.method=TestConnection grpc.method_type=unary grpc.service=cloudquery.plugin.v3.Plugin grpc.start_time=2024-10-08T10:05:22-05:00 grpc.time_ms=1.367 invocation_id=<removed> module=cli peer.address= protocol=grpc
```

### Expected Behavior

I would expect that the call would show an error or message indicating that the feature has not been implemented, even for preview plugins.

### CloudQuery (redacted) config

```
kind: source
# Common source-plugin configuration
spec:
  name: entraid
  path: cloudquery/entraid
  registry: cloudquery
  version: "v1.3.2"
  tables: ["entraid_users"]
  destinations: ["postgresql"]

  # Entra ID specific configuration
  spec:
    # Optional parameters
    # concurrency: 50000

```

### Steps To Reproduce

1. With the above config
2. Run `test-connection <config file name>
3. See successful message on output
4. See error in logs

### CloudQuery (redacted) logs

```
2024-10-08T16:09:59Z INF Loading spec(s) args=["entraid.yml"] invocation_id=6c646d1d-b6bc-4506-80bb-592bf5d44548 module=cli
2024-10-08T16:10:00Z INF Plugin server listening address=/var/folders/l6/0d71k0vj4yqg69ck51fn_5440000gn/T/cq-dTARmNQnCKYnbcje.sock invocation_id=6c646d1d-b6bc-4506-80bb-592bf5d44548 module=cli
2024-10-08T16:10:01Z INF Testing source invocation_id=6c646d1d-b6bc-4506-80bb-592bf5d44548 module=cli source="entraid (cloudquery/entraid@v1.4.3)"
[...]

PR #19349, opened 2024-10-11
fix: Fix incorrect test-connection result for non-cloud requests.

marianogappa:
fixes https://github.com/cloudquery/cloudquery/issues/19338.

The `test-connection` command is incorrectly swallowing `Unimplemented` errors for plugins that don't support the `ConnectionTester` interface.

This behaviour must be kept for Cloud-based syncs, but this PR fixes it for non-Cloud, namely CLI-based connection tests.

Here's a test with the `Sonarqube` source plugin, which doesn't implement the interface at the moment. I'm running it with a fixed client binary (`cli` here) and the latest stable binary (`cloudquery` here)

**Before:**
```
$ cloudquery test-connection sonarqube_to_postgresql.yaml
Loading spec(s) from sonarqube_to_postgresql.yaml
Test Connection Results
source       sonarqube (grpc@localhost:7777)            Success
destination  postgresql (cloudquery/postgresql@v8.6.0)  Success
```

**After:**
```
$ cli test-connection sonarqube_to_postgresql.yaml
Loading spec(s) from sonarqube_to_postgresql.yaml
Test Connection Results
source       sonarqube (grpc@localhost:7777)            Failure: UNIMPLEMENTED 	 TestConnection feature is not implemented in this plugin
destination  postgresql (cloudquery/postgresql@v8.6.0)  Success
Error: at least one test connection failed
$
```


erezrokah (approved):

disq (approved):

erezrokah:
Maybe some our test plugins that we use in CI don't implement the test connection. That could explain the CI failures

marianogappa:
@erezrokah yep it was the `test` plugin and @disq added support for it just now. I'm just rebasing master I think we're good.

disq:
We'll need the updated versions published.

marianogappa:
Ah it downloads them, gotcha.

PR #19350, opened 2024-10-11
feat: Add test connection handling

disq:
Should help with https://github.com/cloudquery/cloudquery/pull/19349


marianogappa (approved):

erezrokah (approved):

marianogappa:
Indeed

erezrokah:
We'll have to this how these are used in the CLI tests, I think we use old versions from GitHub to avoid needing an API key (that's needed to download plugins). Maybe it's time to change our tests to require an API key

PR #19373, opened 2024-10-14
fix: Send content type when uploading images

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/2567 (internal issues). We need to update the CLI to adjust for a change in our backend requiring a content type when uploading images. This happens in the CLI when publishing plugins or addons docs that reference local images in their markdown doc files.


cli/internal/publish/pluginui.go
```diff
@@ -107,22 +106,3 @@ func readFlatDir(base string) (files []string, err error) {
 	})
 	return files, err
 }
-
-func quickContentType(filename string) (string, error) {
```
erezrokah, line 111: Moved this and the tests under `images` package. Seems more suitable now that it's re-used

cli/internal/publish/images/images.go
```diff
@@ -51,10 +72,20 @@ func ProcessDocument(ctx context.Context, c *cloudquery_api.ClientWithResponses,
 	fmt.Println("Preparing to upload images...")
 
 	reqs := make([]cloudquery_api.TeamImageCreate, 0, len(ims))
-	for k := range ims {
+	for k, v := range ims {
+		if len(v) == 0 {
+			continue
+		}
+		absFile := v[0].absFile
```
erezrokah, line 79: Not sure why this is an array, looks like we're using only the first file below anyways

cli/internal/publish/images/images.go
```diff
@@ -39,6 +41,25 @@ type listKey struct {
...
+		return "", fmt.Errorf("failed to open file: %w", err)
+	}
+	defer fp.Close()
+	if _, err := fp.Read(filebytes); err != nil {
+		return "", fmt.Errorf("failed to read file: %w", err)
+	}
+
+	contentType := mime.TypeByExtension(filepath.Ext(filename))
+	if contentType == "" {
+		contentType = http.DetectContentType(filebytes)
```
marianogappa, line 58: If this doesn't detect a content type, should it error?

cli/internal/publish/images/images.go
```diff
@@ -92,8 +123,9 @@ func ProcessDocument(ctx context.Context, c *cloudquery_api.ClientWithResponses,
 		}
 		item := item
```
marianogappa, line 124: Since you're doing this, you might as well do `ims[listKey{name: item.Name, sum: item.Checksum}][0]` and then you don't have to write it twice.

cli/internal/publish/images/images.go
disq, line 58: Maybe open and read file only `if contentType == ""`? Reverse the if and do an early return, otherwise read the first 512 and use the http detect method.

cli/internal/publish/images/images.go
```diff
@@ -39,6 +41,25 @@ type listKey struct {
 	name, sum string
 }
 
+func QuickContentType(filename string) (string, error) {
```
marianogappa, line 44: How about the existing name `DetectContentType`? There's no verb on this name.

cli/internal/publish/images/images_test.go
```diff
@@ -366,3 +367,64 @@ func convertStringToRefKey(input map[string][]reference) map[listKey][]reference
...
+			tryNoExt: true,
+		},
+		{
+			filename: "octetstream.svg",
+			want:     "image/svg+xml",
+		},
+		{
+			filename: "textplain.svg",
+			want:     "image/svg+xml",
+		},
```
marianogappa, line 391: Is there value in checking (in case in a regression there's no `http.DetectContentType`) for the header logic? I don't mind of not.

The 🤖 says these are good for PNG (hex):

```
89 50 4E 47 0D 0A 1A 0A 00 00 00 0D 49 48 44 52
```

And this for JPG:

```
FF D8 FF E0 00 10 4A 46 49 46 00 01 01 01 00 48 00 48
```

marianogappa (approved):
Minor comments

cli/internal/publish/images/images.go
disq: Looks like `findMarkdownImages` returns each image reference separately (and they would all have the same absFile value)

cli/internal/publish/images/images.go
```diff
@@ -92,8 +123,9 @@ func ProcessDocument(ctx context.Context, c *cloudquery_api.ClientWithResponses,
 		}
 		item := item
 		absFile := ims[listKey{name: item.Name, sum: item.Checksum}][0].absFile
+		contentType := ims[listKey{name: item.Name, sum: item.Checksum}][0].contentType
```
disq, line 126: This should be item.RequiredHeaders and that should be passed down to uploadFile

cli/internal/publish/images/images.go
disq: From the godoc:
>  if it cannot determine a more specific one, it returns "application/octet-stream".

erezrokah: Fixed in https://github.com/cloudquery/cloudquery/pull/19373/commits/269d066b6501c4059bb52924177756c35ad5bbb9

cli/internal/publish/images/images.go
erezrokah: Fixed in https://github.com/cloudquery/cloudquery/pull/19373/commits/269d066b6501c4059bb52924177756c35ad5bbb9

cli/internal/publish/images/images.go
erezrokah: Something's wrong with the code gen as this is what I see for `RequiredHeaders`:
```go
// RequiredHeaders Required HTTP headers to include for the upload
RequiredHeaders map[string]interface{} `json:"required_headers"`
```

Looks like it should be `RequiredHeaders map[string][]string` (the frontend client looks ok)

disq: Interesting but setting each key to a `[]string` should still work?

erezrokah: I'm verifying now the runtime type is `map[string][]string` so casting to it will work

erezrokah: OK https://github.com/cloudquery/cloudquery/pull/19373/commits/eeb17507c5917915b96b051c07c50a022848cad1 fixes this, tested locally with the EntraID plugin which is currently failing to publish because of this issue:

```
Publishing plugin cloudquery/entraid@v1.0.0 to CloudQuery Hub...
Uploading table schemas...
Uploading docs...
Preparing to upload images...
Uploading images...
Uploading binaries CloudQuery Hub...
- Uploading linux_amd64...
- Uploading linux_arm64...
- Uploading windows_amd64...
- Uploading darwin_amd64...
- Uploading darwin_arm64...
Success!
```

disq (approved):

PR #19403, opened 2024-10-15
fix: Remove constraints from internal fields if needed

erezrokah:
#### Summary

This PR fixes an issue with `sync_group_id` in append mode where the PK constraint for it was't removed

### DIFF
--- a/cli/internal/transformer/transformer.go
+++ b/cli/internal/transformer/transformer.go
@@ -98,9 +98,11 @@ func (t *RecordTransformer) TransformSchema(sc *arrow.Schema) *arrow.Schema {
 				[]string{schema.MetadataTrue},
 			)})
 	}
-	for _, field := range sc.Fields() {
-		mdMap := field.Metadata.ToMap()
+	fields = append(fields, sc.Fields()...)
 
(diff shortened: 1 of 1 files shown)


murarustefaan (approved):

cli/internal/transformer/transformer.go
```diff
@@ -112,17 +114,15 @@ func (t *RecordTransformer) TransformSchema(sc *arrow.Schema) *arrow.Schema {
 			mdMap[schema.MetadataPrimaryKey] = schema.MetadataTrue
 		}
 
-		newMd := arrow.MetadataFrom(mdMap)
-
-		fields = append(fields, arrow.Field{
+		transformedFields = append(transformedFields, arrow.Field{
```
disq, line 117:
```suggestion
		transformedFields[i] arrow.Field{
```

disq (commented):
Maybe... sligtly more perfy?

erezrokah:
> Maybe... sligtly more perfy?

I think it's the same per wise since in both cases the array will be pre-allocated. But I don't mind, maybe we'll be saving the size check with `append`

PR #19409, opened 2024-10-15
fix: Wait for last print of table metrics before exiting

erezrokah:
#### Summary

This PR fixes an issue with the table metrics flag where we would not print the last update of the metrics, since we didn't wait for the Go routine that does it to finish

### DIFF
--- a/cli/internal/otel/receiver.go
+++ b/cli/internal/otel/receiver.go
@@ -50,10 +50,8 @@ type Consumer struct {
 }
 
 func (c *Consumer) Shutdown(ctx context.Context) {
-	if c.quit != nil {
-		close(c.quit)
-		c.wg.Wait()
-	}
+	close(c.quit)
+	c.wg.Wait()
 }
 
 func newMetricConsumer(metricsFile *os.File, quit chan any, wg *sync.WaitGroup) func(context.Context, pluginMetric) {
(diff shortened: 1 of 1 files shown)


cli/internal/otel/receiver.go
```diff
@@ -46,15 +46,17 @@ type Consumer struct {
 	metricsFilename string
 	metricsFile     *os.File
 	consumeMetric   func(context.Context, pluginMetric)
+	wg              *sync.WaitGroup
 }
 
 func (c *Consumer) Shutdown(ctx context.Context) {
 	if c.quit != nil {
 		close(c.quit)
+		c.wg.Wait()
```
disq, line 55: Shouldn't we wait regardless of the quit channel being set or not?

erezrokah: I think we can remove the `nil` check it will never be `nil` per https://github.com/cloudquery/cloudquery/blob/92c81752ddc95617c48326df97aea353400525fa/cli/internal/otel/receiver.go#L233

erezrokah: Done in https://github.com/cloudquery/cloudquery/pull/19409/commits/c168f1d8a165f773546e7df29fb95b1c671407b3

disq (approved):

PR #19448, opened 2024-10-18
fix: Return correct error from transformers pipeline

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/2591 (internal issue)

We have 2 Go routines in play in `sync_v3`, one for the transformer pipeline and one for handling sync messages on the gRPC channel.
Any of those Go routines can return an error, and the error group will report the first one which is not consistent in case the pipelined was closed due to a destination error where `OnOutput` returns the error and closes the pipeline.
In the latter case we should signal the sync Go routine to exit early but not with an error that will get reported from the pipeline Go routine.

### DIFF
--- a/cli/cmd/sync_v3.go
+++ b/cli/cmd/sync_v3.go
@@ -413,9 +413,6 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 				},
 			}
 			if err := writeClientsByName[destinationName].Send(wr); err != nil {
-				// Close the transformations pipeline when the destination is closed.
-				pipeline.Close()
-
 				return handleSendError(err, writeClientsByName[destinationName], "insert")
 			}
 			return nil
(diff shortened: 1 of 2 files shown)


cli/cmd/sync_v3.go
```diff
@@ -413,9 +413,6 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 				},
 			}
 			if err := writeClientsByName[destinationName].Send(wr); err != nil {
-				// Close the transformations pipeline when the destination is closed.
-				pipeline.Close()
```
erezrokah, line 417: We already close the pipeline in the calling function https://github.com/cloudquery/cloudquery/blob/50ebeb7b78779281a1e22c79d676e3a14a8f668a/cli/internal/transformerpipeline/pipeline.go#L108

marianogappa (approved):

PR #19449, opened 2024-10-18
fix: Stop progress bar when the context is done

erezrokah:
#### Summary

Related to https://github.com/cloudquery/cloudquery-issues/issues/2591 and the bug can been seen in the output of that issue:

```
Loading spec(s) from examples/azure_to_postgresql.yml
Starting sync for: azure (cloudquery/azure@v13.0.0) -> [postgresql (cloudquery/postgresql@v8.6.3)]
/ Syncing resources... (23/-, 13 resources/s) [2s] Error: failed to sync v3 source azure: EOF
- Syncing resources... (23/-, 10 resources/s) [2s] %   
```

There's an additional progress bar print after the error was printed. We should instead stop the progress bar as soon as possible when context is done.


erezrokah:
Looking at the CI failures

cli/cmd/sync_v3.go
```diff
@@ -424,6 +406,27 @@ func syncConnectionV3(ctx context.Context, source v3source, destinations []v3des
 		pipelineByDestinationName[destinationName] = pipeline
 	}
 
+	// Add a ticker to update the progress bar every 100ms
+	t := time.NewTicker(100 * time.Millisecond)
+	newResources := int64(0)
+	go func() {
+		for {
+			select {
+			case <-gctx.Done():
```
marianogappa, line 415: Was it deliberate to change the context variable here? This adds the errgroup with all the pipeline goroutines, which is cancelled even when there's no error (i.e. when the goroutines finish). This change isn't necessarily a bad thing, but the code was stopping when the original ctx closed, which was mainly when the user SIGTERM'd.

marianogappa: Happy to approve if this is fine, let me know. I didn't because you added automerge.

erezrokah: It was deliberate as we should stop the progress bar as soon as the sync finishes (regardless if there's an error or not). Otherwise the timer keeps ticking on an interval and can update the progress after the sync Go routine ended

marianogappa (approved):

PR #19450, opened 2024-10-18
feat: Add hidden `cq-columns-not-null` flag to `migrate` and `sync` commands

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/2596 (internal issue).
This is intended to unblock customers that are partitioning over `_cq_sync_time` until we add proper partitioning support, that's why the flag is hidden (and also give us time to update destinations to handle migrating the columns automatically instead of requiring users to drop their schema)


savme (approved):

marianogappa (approved):
I wasn't part of the discussion so you know best if this is good to go.

My only comment (other than alignment nit) is that the original ticket had an issue with a single column and this feature adds `NOT NULL` constraint for all, right? So we're good until the next customer comes along needing this column to be `NOT NULL`, but a different one cannot be `NOT NULL` because there are already `NULL` rows. At that point we need a new flag, because this one is already being used by Yahoo 🤷

An alternative is a comma-separated list / array of fields rather than a bool, or separate hidden flags per field; more cumbersome and more customisable. We might never need this, it's trickier to implement and has more error scenarios, so I defer to your discussions and opinion.

erezrokah:
> My only comment (other than alignment nit) is that the original ticket had an issue with a single column and this feature adds `NOT NULL` constraint for all, right? So we're good until the next customer comes along needing this column to be `NOT NULL`, but a different one cannot be `NOT NULL` because there are already `NULL` rows. At that point we need a new flag, because this one is already being used by an existing customer 🤷
> 
> An alternative is a comma-separated list / array of fields rather than a bool, or separate hidden flags per field; more cumbersome and more customisable. We might never need this, it's trickier to implement and has more error scenarios, so I defer to your discussions and opinion.

I believe setting the (current) internal columns to `Nullable` is more of an oversight than something we planned. They are always populated and will never be null.
The next step to this PR will be to add [an SDK schema change](https://github.com/cloudquery/plugin-sdk/blob/3cc0a31dd78ed2f3fc3e05567661a549374a7733/schema/table.go#L49) to detect adding not null constraints and have destinations migrate it automatically (example [here](https://github.com/cloudquery/cloudquery/blob/8d7a62ce00bc5f3ba565727eac8b288b1cda7116/plugins/destination/postgresql/client/migrate.go#L144)).

The reason we've added this flag and simply not change the columns to be `NOT NULL` is that will break all tables for all destinations that apply that constraint, without giving users an easy way to migrate

PR #19459, opened 2024-10-21
chore(main): Updated Core Concepts page

JoeKarlsson:
How to link with a 404 on it, so I removed it. I also updated our language from plug-ins to integrations and improved grammar.


website/pages/docs/core-concepts/plugins.md
erezrokah, line 1: Should we rename the filename? Otherwise the URL slug still says `plugins`, see 
image

If we rename the URL slug we'll need to set up a redirect as well.
Also https://github.com/cloudquery/cloudquery/blob/81fbc90078b3c75d30d0650eb85604bf876a6ab7/website/pages/docs/core-concepts/_meta.json#L2 needs updating so the side bar says `Integrations` instead of `Plugins

PR #19460, opened 2024-10-22
fix: Show DSN connection string format first

erezrokah:
#### Summary

The issue of users unable to connect to Postgres because they need to encode their password value in the URL when it contains special characters has come up multiple times before. An easy way to avoid it is to use the DSN format. It's much easier I think to quote a value in the DSN format (to handle spaces) vs percent encode special URI characters.

I changed the docs a bit to list the DSN format first.

Feel free to close if this doesn't make sense to you


plugins/destination/postgresql/client/spec/spec.go
```diff
@@ -17,11 +17,11 @@ const (
 )
 
 type Spec struct {
-	// Connection string to connect to the database. This can be a URL or a DSN, for example:
+	// Connection string to connect to the database. This can be a URI or a DSN, for example:
```
disq, line 20:
```suggestion
	// Connection string to connect to the database. This can be a DSN or a URI, for example:
```
Since we're prioritizing DSN, mention it first in the doc.

erezrokah: Did it in another place forgot the spec, will fix

murarustefaan (approved):

murarustefaan:
I think this makes sense, people used with URIs will use them either way while showing we support DSN's will also help the others.

PR #19462, opened 2024-10-22
feat: Implement JSON Flattener transformer.

marianogappa:
This is an initial draft PR for a future JSON Flattener transformer, to gather feedback early.

At the moment, there are several limitations:
- JSON fields on source table will be flattened into separate columns in the destination, while still preserving the original JSON fields.
- Only JSON fields with schema metadata will be flattened. This metadata is present for selected CloudQuery sources.
- Only single-level JSON object fields will be flattened. JSON arrays will not be flattened.
- Multi-level JSON objects will not be flattened recursively.

Customisability is very limited:
- Only table glob patterns are supported
- No custom flattening configuration; only based on existing typeSchema metadata
- Cannot limit which fields will be flattened

At the moment, these types are supported:
- utf8
- int64
- bool
- timestamp
- json
- float64

In principle, testing shows it works reasonably well; it succeeded to flatten the whole `aws` plugin:

(with a small edge case; I skipped this table for now: `aws_apprunner_services`)

```
$ cli sync .
Loading spec(s) from .
Starting sync for: aws (cloudquery/aws@v27.23.0) -> [postgresql (cloudquery/postgresql@v8.0.7)]
Sync completed with errors, see logs for details. Resources: 621239, Errors: 276, Warnings: 74, Time: 11m25s
```

Screenshot 2024-10-22 at 18 03 17
Screenshot 2024-10-22 at 18 03 26


erezrokah (commented):
Looks great 🚀 

CI is failing since you need to add a workflow file similar to https://github.com/cloudquery/cloudquery/blob/7c20418bd8fbb5b9a74726c90251fcde9a53b94a/.github/workflows/transf_basic.yml

Also missing is the release configuration (see steps in https://github.com/cloudquery/cloudquery/blob/7c20418bd8fbb5b9a74726c90251fcde9a53b94a/contributing/adding_a_new_plugin_to_cq_monorepo.md)

It would be nice to add to the docs an example table with the data pre/post transformation similar to the table we show here https://docs.cloudquery.io/docs/quickstart/macOS#examine-the-data

marianogappa:
@erezrokah added the CI configs but not the docs yet.

.release-please-manifest.json
```diff
@@ -103,5 +103,6 @@
   "plugins/source/bitbucket+FILLER": "0.0.0",
   "plugins/source/notion": "1.1.8",
   "plugins/source/xkcd": "1.4.4",
-  "plugins/transformer/basic": "1.1.5"
+  "plugins/transformer/basic": "1.1.5",
+  "plugins/transformer/json": "1.0.0"
```
erezrokah, line 107:
```suggestion
  "plugins/transformer/basic": "1.1.5"
```

This is not needed, our bot updates it when it releases a new version. By putting a version here you're telling the bot the existing released version is `1.0.0` so the next one will be `1.1.0`

erezrokah:
> @erezrokah added the CI configs but not the docs yet.

The place to add the release config is here https://github.com/cloudquery/cloudquery/blob/68dbfefd0392448d13ff50defc2205e067e421c2/release-please-config.json

plugins/transformer/jsonflattener/client/recordupdater/column_builders.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+
+	"github.com/apache/arrow/go/v17/arrow"
+	"github.com/cloudquery/cloudquery/plugins/transformer/jsonflattener/client/schemaupdater"
+	"github.com/cloudquery/plugin-sdk/v4/types"
+)
+
+// ColumnBuilder creates separate columns from ONE json column, based on the type schema.
+type ColumnBuilder interface {
```
dcelasun, line 12: nit: This doesn't need to be exported.

marianogappa: Unexported all related symbols.

erezrokah (approved):

PR #19504, opened 2024-10-28
chore: Add WebSite structured data

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery-issues/issues/2647 (hopefully), that is, fixes `Vercel` showing up as site name for some searches:
image


website/theme.config.tsx
```diff
@@ -6,6 +6,16 @@ import { Badge } from "./components/Badge";
...
+import { WebSite, WithContext } from "schema-dts";
+import Script from "next/script";
+
+const jsonLd: WithContext<WebSite> = {
+  "@context": "https://schema.org",
+  "@type": "WebSite",
+  name: "CloudQuery",
+  alternateName: ["CloudQuery Docs"],
+  url: "https://docs.cloudquery.io/",
+};
```
hermanschaaf, line 17: Is it okay to show this `WebSite` type on every page of the docs? I'm no expert on this but I would have thought that we should use "WebPage" entity on every page

erezrokah: > Is it okay to show this `WebSite` type on every page of the docs? I'm no expert on this but I would have thought that we should use "WebPage" entity on every page

Google says you only need to add it to your homepage 
image

We have a redirect from `https://docs.cloudquery.io/` to `https://docs.cloudquery.io/docs` so I'm not sure which one is considered to be the homepage actually

erezrokah: This PR adds it to every page but I can make it conditional based on the path

hermanschaaf: > This PR adds it to every page but I can make it conditional based on the path

Yeah, I don't know but if we can do that it seems like it would be a little more in line with what Google suggests

adam-cobb532: @erezrokah @hermanschaaf Using WebPage Schema and IsPartOf seems like a good solution here https://schema.org/WebPage

adam-cobb532:
Have added a separate issue about the /docs redirect given I think that could be causing us SEO issues. Will need to be resolved carefully though!
https://github.com/cloudquery/cloudquery-issues/issues/2652

website/theme.config.tsx
erezrokah: > @erezrokah @hermanschaaf Using WebPage Schema and IsPartOf seems like a good solution here https://schema.org/WebPage

I'm happy to do whatever, what should we put in `IsPartOf`?

adam-cobb532:
I’d recommend we do if Herman is aligned.

website/theme.config.tsx
erezrokah: > Let's use `WebSite` on the home page, and `WebPage` on other pages, with `IsPartOf` set to the home page URL on those other pages? But I think for Google all that really matters is the `WebSite` entry at the root level

This is done (see recent changes). But we have a permanent redirect from the root to `/docs`:
image

So not sure if Google reads the root data. The `/docs` path has this now 
image

hermanschaaf (approved):

PR #19596, opened 2024-11-13
feat: Implement PARTITION BY support on ClickHouse destination.

marianogappa:
Enables optional support for partitioning tables in ClickHouse destination.

Because this came up in discussions as a potential eyebrow raiser, note that I decided to expose the partitioning as a pass-through string, rather than a list of columns, which was suggested:

- According to docs and according to our research for using it, I've never seen a case of partitioning by more than one column, so a list of columns seems like a feature nobody needs.
- According to docs and according to our research for using it, it's much more common to partition by an expression rather than by a column, e.g. `toYYYYMM(_cq_sync_time)`. So partitioning by column names would make this feature unusable for many use cases, and also I wouldn't want us to need to implement a semantic validation of all possible expressions that can be used here.

#### Partitioning

This option allows to specify a partitioning strategy to be used for tables. It is an array of objects.

Each object has the following fields:

- `tables` (array of strings) (optional) (default: `["*"]`)

  List of glob patterns to match table names against. Follows the same rules as the top-level spec `tables` option.

  If a table matches both a pattern in `tables` and `skip_tables`, the table will be skipped.

  Partition strategy table patterns should be disjointed sets: if a table matches two partition strategies, an error will be raised at runtime.

- `skip_tables` (array of strings) (optional) (default: empty)

  List of glob patterns to skip matching table names against. Follows the same rules as the top-level spec `skip_tables` option.

  If a table matches both a pattern in `tables` and `skip_tables`, the table will be skipped.

  Partition strategy table patterns should be disjointed sets: if a table matches two partition strategies, an error will be raised at runtime.

- `partition_by` (string) (required)

  Partitioning strategy to use, e.g. `toYYYYMM(_cq_sync_time)`, the string is passed as is after "PARTITION BY" clause with no validation or quoting.

  An unset `partition_by` is not valid.

Example:

```yaml copy
partition:
- tables: ["*"]
  skip_tables: ["special_partition_table", "non_partitioned_table"]
  partition_by: "toYYYYMM(_cq_sync_time)"
- tables: ["special_partition_table"]
  partition_by: "toYYYYMMDD(_cq_sync_time)"
```


erezrokah (approved):
Looks good 👍 Makes sense to have it as a passthrough string and not a list of columns

disq (approved):

marianogappa:
I've tested it successfully
Screenshot 2024-11-13 at 16 04 48

PR #19641, opened 2024-11-19
feat: Implement ORDER BY spec support on ClickHouse destination.

marianogappa:
This PR implements support for setting custom `ORDER BY` clauses on any table or group of tables in the ClickHouse destination.

Because tables are defined on the source side and every destination can have different optimal indexing/sorting circumstances, it is not practical to define destination-specific sorting strategies at the source. Therefore, this enables setting it at the destination.

Screenshot 2024-11-19 at 16 08 06

Screenshot 2024-11-19 at 16 08 32


erezrokah (approved):

erezrokah (commented):
OK so I think there might be an issue here related to migrations, if someone changes the `order by` can we make sure that the plugin fails with a proper error? I think now it will silently ignore the change

plugins/destination/clickhouse/queries/tables.go
```diff
@@ -5,17 +5,18 @@ import (
...
 	"github.com/cloudquery/plugin-sdk/v4/glob"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 )
 
-func sortKeys(table *schema.Table) []string {
+func SortKeys(table *schema.Table) []string {
 	keys := make([]string, 0, len(table.Columns))
 	for _, col := range table.Columns {
-		if col.NotNull || col.PrimaryKey {
+		if (col.NotNull || col.PrimaryKey) && !isCompoundType(col) {
```
erezrokah, line 19: OK so I was thinking of doing https://github.com/cloudquery/cloudquery/compare/fix/CanonizedField?expand=1

That fixes the underlying issue in `CanonizedField` of losing the nullability and metadata and won't duplicate the logic of https://github.com/cloudquery/cloudquery/blob/a81edd6c5e7c2a25f3396ac80983d28c4af2f1c7/plugins/destination/clickhouse/typeconv/ch/types/nullable.go#L10

erezrokah (commented):
Added a comment, and I think we would probably need release this as a breaking change.
I would break this PR into 3 parts (in this order):
1. Detect changes to order and partition by and fail migration (or recreate tables in `force_migrate` mode
2. Fix the order by bug that causes compound types to always be included in the sort key
3. Add support for configuring `Order By`

WDYT?

marianogappa:
@erezrokah 1. and 2. must happen together because if I do 1., CI doesn't pass due to the bug

erezrokah:
> Oh, so you think that by adding the fix commit you created, 1. won't fail CI?

Not sure, but if we need to do all together that's fine we can tweak the changelog so it shows up as 3 different fixes, and explain why it's breaking

PR #19642, opened 2024-11-19
feat: Add license information

murarustefaan:


murarustefaan:
> Looks good but we should exclude the licenses file from the vale linter check in https://github.com/cloudquery/cloudquery/blob/bbbbc9f49f714b5c3be108d2ad9e5fd30ee1565d/.github/workflows/lint_markdown.yml

Yup, you're right, added a change that does that.

PR #19675, opened 2024-11-21
fix: Keeps nullability and metadata on field Canonize.

marianogappa:


plugins/destination/clickhouse/typeconv/schema.go
```diff
@@ -42,5 +42,13 @@ func CanonizedField(field arrow.Field) (*arrow.Field, error) {
 		return nil, err
 	}
 	// 2 - convert back to Apache Arrow
-	return arrow_types.Field(field.Name, fieldType)
+	temp, err := arrow_types.Field(field.Name, fieldType)
+	if err != nil {
+		return nil, err
+	}
+	clone := field
+	clone.Nullable = field.Nullable
```
erezrokah, line 50:
```suggestion
```

plugins/destination/clickhouse/typeconv/schema.go
```diff
@@ -42,5 +42,13 @@ func CanonizedField(field arrow.Field) (*arrow.Field, error) {
...
+	clone.Type = temp.Type
+	clone.Metadata = field.Metadata
```
erezrokah, line 52:
```suggestion
```

PR #19676, opened 2024-11-22
feat: Add FIPS version

erezrokah:
#### Summary

This PR is intended to add a FIPS version of the test source plugin, it works only for linux arm64 and amd64, since that's what `boringcrypto` [supports](https://github.com/golang/go/blob/4865aadc21acebc8039f914929f03c7042b2ae8d/src/crypto/boring/boring_test.go#L16). 

Needs https://github.com/cloudquery/plugin-sdk/pull/1974 for validation


.github/workflows/publish_plugin_to_hub_fips.yml
```diff
@@ -0,0 +1,122 @@
...
+  push:
+    tags:
+      - "plugins-source-test-v*.*.*"
+jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    outputs:
+      plugin_name: ${{ steps.split.outputs.plugin_name }}
+      plugin_kind: ${{ steps.split.outputs.plugin_kind }}
+      plugin_version: ${{ steps.split.outputs.plugin_version }}-fips
```
erezrokah, line 12: Notice we append `-fips` to the version here

murarustefaan (approved):

murarustefaan:
I'd say let's wait and see whether @bernays is also ok with this approach.

PR #19691, opened 2024-11-27
chore: Use predefined unimplemented type

disq:
Good for `grep`.


plugins/destination/clickhouse/client/client.go
```diff
@@ -23,16 +22,13 @@ type Client struct {
 	logger zerolog.Logger
 	writer *batchwriter.BatchWriter
 	plugin.UnimplementedSource
+	batchwriter.UnimplementedDeleteStale
```
disq, line 25: This is:
```go
type UnimplementedDeleteStale struct{}

func (UnimplementedDeleteStale) DeleteStale(context.Context, message.WriteDeleteStales) error {
	return fmt.Errorf("DeleteStale: %w", plugin.ErrNotImplemented)
}
```

erezrokah (approved):

PR #19699, opened 2024-11-27
feat: Add snyk code scan

murarustefaan:


erezrokah:
Don't think we need this? We can use Snyk as a GitHub app

erezrokah:
Closing as I assume we integrated this via another way. Please re-open if that's not the case

murarustefaan:
Yeah, we don't need it anymore.

PR #19861, opened 2024-12-04
refactor: Don't use request URL path to build zip path in addons download

erezrokah:
#### Summary

Fixes https://github.com/cloudquery/cloudquery/security/code-scanning/4
Fixes https://github.com/cloudquery/cloudquery/security/code-scanning/3
Fixes https://github.com/cloudquery/cloudquery/security/code-scanning/2

Technically those are all false positives since:
1. The request path is derived from the location [here](https://github.com/cloudquery/cloudquery/blob/83cab0b4b34bfbe33a89b444659087dfa5b4ef7b/cli/cmd/addon_download.go#L94), which is built from the addon parts which are validated [here](https://github.com/cloudquery/cloudquery/blob/83cab0b4b34bfbe33a89b444659087dfa5b4ef7b/cli/cmd/addon_download.go#L67) (though not validated for `..` in the argument)
2. If `/` or `..` are part of the addon name the request to get it will fail so we won't reach the code that downloads it.

Regardless I think we can refactor the code to make it more clear where that value is coming from

### DIFF
--- a/cli/cmd/addon_download.go
+++ b/cli/cmd/addon_download.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path"
+	"path/filepath"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


disq (approved):

cli/cmd/addon_download.go
```diff
@@ -110,5 +115,10 @@ func runAddonDownload(ctx context.Context, cmd *cobra.Command, args []string) er
 		return fmt.Errorf("addon download failed: %d %s", res.StatusCode, location)
 	}
 
-	return publish.DownloadAddonFromResponse(res, checksum, targetDir)
+	zipPath := "-"
+	if targetDir != "-" {
+		zipPath = filepath.Join(targetDir, fmt.Sprintf("%s_%s_%s_%s.zip", addonTeam, addonType, addonName, addonVersion))
```
erezrokah, line 120: OK this is safer I think

PR #19863, opened 2024-12-05
feat: Add "fail immediately" spec option to test source plugin

marianogappa:
Adds a `fail_immediately` spec option to the test source plugin.

Initially I also added a `fail_after_n_resources` option, but it's tricky to implement with multiple tables, and it would interact weirdly with batching. I didn't see much benefit to be honest.

This will be used to add end to end testing at the CLI stage.


plugins/source/test/resources/services/test_sub_table.go
```diff
@@ -53,6 +53,9 @@ func testSubTable(config client.Spec) *schema.Table {
 
 func fetchSubTableData(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- any) error {
 	cl := meta.(*client.Client)
+	if cl.Spec.FailImmediately {
+		return fmt.Errorf("failing immediately")
```
maaarcelino, line 57: WDYT about having a defined error that we return in all three of these resolvers? e.g. `var ErrFailImmediately = ...; return ErrFailImmediately`

marianogappa: Makes sense. Implemented.

maaarcelino (approved):
LGTM

PR #19914, opened 2024-12-09
fix: Create unique index per table, handle non existing tables

erezrokah:
#### Summary

There are 2 bugs related to performance index creation:
1. We use the same index name for all tables, so it gets created only for a single table across the database
2. We don't create it for new tables, only existing ones

The logic in the test is to handle CockroachDB

### DIFF
--- a/plugins/destination/postgresql/client/client_test.go
+++ b/plugins/destination/postgresql/client/client_test.go
@@ -23,7 +23,7 @@ import (
 func getTestConnection() string {
 	testConn := os.Getenv("CQ_DEST_PG_TEST_CONN")
 	if testConn == "" {
-		return "postgresql://postgres:pass@localhost:5442/postgres?sslmode=disable"
+		return "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"
 	}
 	return testConn
 }
--- a/plugins/destination/postgresql/client/client_test.go
+++ b/plugins/destination/postgresql/client/client_test.go
@@ -3,22 +3,59 @@ package client
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"testing"
 	"time"
 
+	"github.com/apache/arrow/go/v17/arrow"
 	"github.com/cloudquery/cloudquery/plugins/destination/postgresql/v8/client/spec"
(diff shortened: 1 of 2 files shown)


plugins/destination/postgresql/client/client_test.go
```diff
@@ -3,24 +3,61 @@ package client
...
+	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/jackc/pgx/v5/tracelog"
+	"github.com/stretchr/testify/require"
 )
 
 func getTestConnection() string {
 	testConn := os.Getenv("CQ_DEST_PG_TEST_CONN")
 	if testConn == "" {
-		return "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"
+		return "postgresql://postgres:pass@localhost:5442/postgres?sslmode=disable"
```
disq, line 25: Intended?

erezrokah: Nope

erezrokah, line 25:
```suggestion
		return "postgresql://postgres:pass@localhost:5432/postgres?sslmode=disable"
```

murarustefaan (approved):

plugins/destination/postgresql/client/migrate.go
```diff
@@ -420,7 +420,7 @@ func (c *Client) createPerformanceIndexes(ctx context.Context, table *schema.Tab
 		return nil
 	}
 
-	indexName := "cq_performance_idx"
+	indexName := table.Name + "_cqpi"
```
bbernays, line 423: This has an issue when the table name is 62 characters (or whatever the maximum length for an object name), we cannot add 5 more characters to the name. To make the fix backwards compatible we can just use a hash of the entire table name if the length is the max length

erezrokah: This is how we name PKs so only following that convention (it will fail for them too, see https://github.com/cloudquery/cloudquery-private/pull/5379

bbernays: Great point! existing PKs use `table_name + '_cqpk'` which is the same number of characters as `_cqpi` so no new issue here.

PR #19964, opened 2024-12-13
feat: Support all batch writers in test destination plugin.

marianogappa:
Adds spec options for all 3 batch writers on the test destination plugin.

My tests show that all 3 can be used. However, the streamBatchWriter CLI tests are getting stuck. This shouldn't block this PR, on the contrary, it might be good to add tests to the CLI using the `streamBatchWriter` and figure out why it's getting stuck.

Testing with this spec
Screenshot 2024-12-13 at 16 55 26

Test pass when omitting the streamBatchWriter
Screenshot 2024-12-13 at 17 34 44


plugins/destination/test/go.mod
```diff
@@ -5,13 +5,14 @@ go 1.22.7
 require (
 	github.com/apache/arrow/go/v17 v17.0.0
 	github.com/cloudquery/codegen v0.3.21
-	github.com/cloudquery/plugin-sdk/v4 v4.71.1
+	github.com/cloudquery/plugin-sdk/v4 v4.72.0
```
erezrokah, line 8: I'll hold off updating to this SDK version otherwise you'd have two arrow versions in the plugin, v17 and v18.

The update is done via https://github.com/cloudquery/cloudquery/pull/19963

erezrokah (approved):
Looks good so can be merged without the SDK update

plugins/destination/test/go.mod
marianogappa: Downgraded. The reason I initially upgraded was to see if the streambatchwriter issue was already fixed.

marianogappa:
Ran the tests after downgrading arrow. Tests still behave the same.

PR #20116, opened 2025-01-07
feat: Publish UI Bundle Archive

rhino1998:
This PR creates and uploads a `bundle.tar.gz` of UI assets as part of the plugin publishing process. 

The bundle file will be downloaded and extracted in cases where we want to download/mirror all the files needed to serve a UI for a plugin. Otherwise, we don't have a way to list all the files.


cli/internal/publish/pluginui.go
```diff
@@ -65,14 +81,36 @@ func UploadPluginUIAssets(ctx context.Context, c *cloudquery_api.ClientWithRespo
...
 
+	if bundleAsset.UploadURL == "" {
+		return errors.New("bundle asset URL not found in the response")
+	}
+
+	bundleReader, bundleWriter := io.Pipe()
+	defer bundleWriter.Close()
+
+	eg.Go(func() error {
+		defer bundleWriter.Close()
```
disq, line 106: This call is a duplicate, also in L103

hermanschaaf: Fixed in https://github.com/cloudquery/cloudquery/pull/20116/commits/b1fcdbed4c55d556bf08926355161c232832490f

cli/internal/publish/pluginui.go
```diff
@@ -65,14 +81,35 @@ func UploadPluginUIAssets(ctx context.Context, c *cloudquery_api.ClientWithRespo
...
 			return hub.UploadFileWithContentType(egCtx, asset.UploadURL, details[0], details[1])
 		})
 	}
 
+	if bundleAsset.UploadURL == "" {
+		return errors.New("bundle asset URL not found in the response")
+	}
+
+	bundleReader, bundleWriter := io.Pipe()
+	defer bundleWriter.Close()
```
hermanschaaf, line 103: Should `bundleReader` be closed as well?

erezrokah: We do close the file here https://github.com/cloudquery/cloudquery/pull/20116/files#diff-9151361555ef4493db87b7c5d100db9d6666f39a38245b40957be1a63d219a53L79 before calling the `io.Reader` variant of `UploadReaderWithContentType`

erezrokah (approved):
Approving assuming this works and to unblock. Maybe we can try publishing a plugin locally with this code if we hadn't done it.

We could also compress the file locally then upload the compressed file (like any other asset) instead of compressing during upload.
Might be less efficient but would make the code easier to follow I think

cli/internal/publish/pluginui.go
```diff
@@ -65,14 +81,35 @@ func UploadPluginUIAssets(ctx context.Context, c *cloudquery_api.ClientWithRespo
...
+
+	eg.Go(func() error {
```
erezrokah, line 105: Should we be re-using the same error group here? Asking since it's limited to 4 Go routines so we might be writing to the writer without a Go routine that reads from it (at least until one is available).
Don't think it can hang but might be better to have a dedicated error group with 2 Go routines, one for writing, another for reading

rhino1998: Refactored into a separate function w/ a nested errgroup [c1a544e](https://github.com/cloudquery/cloudquery/pull/20116/commits/c1a544edd411fe64d963c0497829a127f9504ff4)

cli/internal/publish/pluginui.go
rhino1998: I'm pretty sure `io.PipeReader` is safe to just discard (no internal references and no held resources) but [5dffd31](https://github.com/cloudquery/cloudquery/pull/20116/commits/5dffd3178590504bbe8b414e3cbde661fc7372c1) adds an explicit close.

cli/internal/publish/pluginui.go
rhino1998: The separate close cases were intentional. The one that was removed is needed for the reader to eventually hit EOF.

[c1a544e](https://github.com/cloudquery/cloudquery/pull/20116/commits/c1a544edd411fe64d963c0497829a127f9504ff4) makes this more explicit and adds better force cancelation for error cases.

hermanschaaf (approved):

rhino1998:
Feel free to merge this if you think it is ready. I've tested it against a local instance and it is working.

erezrokah (approved):

PR #20138, opened 2025-01-13
chore: Disable Airtable source plugin CI workflows

murarustefaan:


erezrokah (approved):

erezrokah (changes requested):
Actually I think this will break our CI as if a dependency update runs our "wait for all workflow" will expect the CI to run and then fail the PR

erezrokah:
I suggest we simply remove this plugin from https://github.com/cloudquery/cloudquery/blob/main/release-please-config.json so no new releases will get generated

murarustefaan:
Yeah i think that makes sense, good catch!

PR #20139, opened 2025-01-16
feat: Separate CLI tokens

disq:
Enables `CLOUDQUERY_CLI_API_URL` and `CLOUDQUERY_CLI_KEY` env var handling, only activated when `CQ_CLOUD` is truthy.

### DIFF
--- a/cli/internal/api/client.go
+++ b/cli/internal/api/client.go
@@ -6,8 +6,10 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 
 	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
+	"github.com/cloudquery/cloudquery-api-go/auth"
 	"github.com/cloudquery/cloudquery/cli/v6/internal/env"
 )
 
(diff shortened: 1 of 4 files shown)


cli/internal/api/client.go
```diff
@@ -12,26 +14,23 @@ import (
...
-	c, err := cloudquery_api.NewClientWithResponses(env.GetEnvOrDefault(envAPIURL, defaultAPIURL))
-	if err != nil {
-		return nil, fmt.Errorf("failed to create api client: %w", err)
-	}
-	return c, nil
+	return NewClient("")
+}
+
+// NewLocalClient creates a client that connects to the local API if possible. If not, it falls back to the regular API using `nonLocalToken`.
+func NewLocalClient(nonLocalToken string) (*cloudquery_api.ClientWithResponses, error) {
```
disq, line 32: Next step would be to go through cli client usage and replace some of the `NewClient` usage with `NewLocalClient` (such as sync, test conn ...)

cli/internal/api/client.go
```diff
@@ -74,3 +73,44 @@ func GetPluginVersion(cl *cloudquery_api.ClientWithResponses, teamName string, k
...
+
+	val := env.GetEnvOrDefault(envCLIAPIURL, regularAPI)
+	return val, val != regularAPI
+}
+
+func overrideToken(token string, getLocal bool) string {
+	if !getLocal {
+		return token
+	}
+	return env.GetEnvOrDefault(envCLIToken, "")
```
disq, line 96: Never using the given `token` (which is not local) with the local endpoint.

cli/internal/api/client.go
```diff
@@ -12,26 +14,23 @@ import (
...
 }
 
+// NewAnonymousClient creates a client that doesn't require authentication.
 func NewAnonymousClient() (*cloudquery_api.ClientWithResponses, error) {
-	c, err := cloudquery_api.NewClientWithResponses(env.GetEnvOrDefault(envAPIURL, defaultAPIURL))
-	if err != nil {
-		return nil, fmt.Errorf("failed to create api client: %w", err)
-	}
-	return c, nil
+	return NewClient("")
```
erezrokah, line 28: Won't this pass `Authorization: Bearer`? Will that work? Won't the service fail the request?

disq: No, empty tokens are now considered anonymous (see `opts` init in `newClient()`)

erezrokah: Ah ok you changed `NewClient` to call `newClient` 🙃

disq:
I think I have a better solution using `isolatePluginEnvironment`.

PR #20140, opened 2025-01-16
fix: Allow overriding API key per plugin in isolated env setting

disq:
Without this fix, two env vars of the same name would be returned.

### DIFF
--- a/cli/cmd/sync.go
+++ b/cli/cmd/sync.go
@@ -495,7 +495,7 @@ func filterPluginEnv(environ []string, pluginName, kind string) []string {
 	for _, v := range environ {
 		switch {
 		case strings.HasPrefix(v, "CLOUDQUERY_API_KEY="):
-			lowerPriority[v[:strings.Index(v, "=")]] = v
+			lowerPriority["CLOUDQUERY_API_KEY"] = v
 		case strings.HasPrefix(v, "_CQ_TEAM_NAME="),
 			strings.HasPrefix(v, "HOME="):
 			env = append(env, v)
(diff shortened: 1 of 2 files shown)


cli/cmd/sync.go
```diff
@@ -486,17 +486,28 @@ func sync(cmd *cobra.Command, args []string) error {
 }
 
 func filterPluginEnv(environ []string, pluginName, kind string) []string {
-	env := make([]string, 0)
+	env := make([]string, 0, len(environ))
```
erezrokah, line 489: `lowerPriority` works only on `CLOUDQUERY_API_KEY`?

Would it make sense to do a first pass to group the envs by `prefix`?
- With prefix - high priority
- Without prefix - low priority

Then it's easier to add them to env

erezrokah: Also don't we want to pass `CLOUDQUERY_API_URL` to plugins too?

disq: That would change how things work currently - but would only change for isolatePluginEnv enabled.

disq: but then isolate also disallows any env vars except for the given three to be present.

erezrokah: OK then so maybe change the code to have a boolean `use_cq_api_from_isolated_env` or something insted of a `lowerPriority` map so it's clear it's only supported for that specific use case

disq: OK done in https://github.com/cloudquery/cloudquery/pull/20140/commits/6eb0c7a47b4e969c631bea7b8b00b2575734d5fb

disq: But maybe team name should also be overridable? Let me do it again

disq: Looks like _CQ_TEAM_NAME is not used in the usage client when API key is specified. `cloudTokenSource` also has some more usage around api url/key, but we're not enabling `CQ_CLOUD` for plugins except when they're really invoked in the cloud so they are non issues at this point.

erezrokah (approved):

PR #20141, opened 2025-01-17
chore: Update `update docs` in release workflow

disq:


.github/workflows/release_cli.yml
```diff
@@ -105,7 +105,7 @@ jobs:
       - name: Update docs
         if: steps.semver_parser.outputs.prerelease == ''
         run: |
-          cp -r ./cli/docs/reference/*.md ./frontend/old_docs_site/pages/docs/reference/cli/
+          cp -r ./cli/docs/reference/*.md ./frontend/apps/docs/pages/docs/reference/cli/
```
disq, line 108: Not sure if correct, but this seems to be the only place to put these now

erezrokah (approved):

PR #20193, opened 2025-01-29
fix: Handle SnakeCasing for Initialisms

bbernays:
Our caser automatically handles TLS differently than normal words because it is included in our common initialisms

https://github.com/cloudquery/plugin-sdk/blob/b3be1b67a1b72a85424488eff7d931ae9007f686/caser/initialisms.go#L36

Before:
Screenshot 2025-01-29 at 12 53 19 PM

After:
Screenshot 2025-01-29 at 12 54 23 PM


erezrokah:
As discussed internally this should be fix in the Azure plugin JSON schema (internal link https://github.com/cloudquery/cloudquery-private/blob/c09d7f54f6ba8f3433f89631f77f5388994af629/plugins/source/azure/client/transformer.go#L14)

PR #20194, opened 2025-01-30
feat: Passtrough `AWS_`-prefixed environment variables to plugins

murarustefaan:


erezrokah (approved):

erezrokah (approved):
Looks good but looks like we can just return `parts[0]` from `splitEnv` per the linter

PR #20284, opened 2025-02-24
chore: Add source_name & invocation_id to start/end sync logs.

marianogappa:
This PR adds source_name & invocation_id to start/end sync logs. Otel consumers may use this information for synchronization purposes.

### DIFF
--- a/cli/cmd/sync_v3.go
+++ b/cli/cmd/sync_v3.go
@@ -197,8 +197,8 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
 		}
 	}
 
-	log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("Start sync")
-	defer log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("End sync")
+	log.Info().Str("source", sourceSpec.VersionString()).Str("source_name", sourceName).Str("invocation_id", invocationUUID.UUID.String()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("Start sync")
+	defer log.Info().Str("source", sourceSpec.VersionString()).Str("source_name", sourceName).Str("invocation_id", invocationUUID.UUID.String()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("End sync")
 
 	variables := specs.Variables{
 		Plugins: make(map[string]specs.PluginVariables),


marianogappa:
This doesn't work as is because CLI logs are not getting to otel. I think it makes sense for them to, because SDK runs this, so for example `table sync finished` and other sync logs do get picked up.

I think we need to run something like this at startup: https://github.com/cloudquery/plugin-sdk/blob/main/serve/opentelemetry.go#L120 cc @erezrokah you set this up initially, wdyt?

cli/cmd/sync_v3.go
```diff
@@ -197,8 +197,8 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
 		}
 	}
 
-	log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("Start sync")
-	defer log.Info().Str("source", sourceSpec.VersionString()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("End sync")
+	log.Info().Str("source", sourceSpec.VersionString()).Str("source_name", sourceName).Str("invocation_id", invocationUUID.UUID.String()).Strs("destinations", destinationStrings).Strs("transformers", transformerStrings).Time("sync_time", syncTime).Msg("Start sync")
```
erezrokah, line 200: Invocation Id is already added in https://github.com/cloudquery/cloudquery/blob/fed76694d622e511f7d61c2d0baedfbe39b15d60/cli/cmd/logging.go#L51 I think

erezrokah: We could use the same approach to add the source/destination name to each plugin logger in https://github.com/cloudquery/cloudquery/blob/fed76694d622e511f7d61c2d0baedfbe39b15d60/cli/cmd/sync.go#L217 and https://github.com/cloudquery/cloudquery/blob/fed76694d622e511f7d61c2d0baedfbe39b15d60/cli/cmd/sync.go#L266 while we're at it

erezrokah:
> This doesn't work as is because CLI logs are not getting to otel. I think it makes sense for them to, because SDK runs this, so for example `table sync finished` and other sync logs do get picked up.

Ah yeah true. OTEL is an SDK feature (and only for source plugins actually). OTEL started with traces and metrics to benchmark syncs, and later on we added logs.

While I agree it makes sense to send CLI logs to OTEL as well we'll have to think this from a configuration perspective as now OTEL is configured per source plugin (not even destinations), so we might need to configure it via CLI flags.

Would my suggestion in https://github.com/cloudquery/cloudquery/pull/20284#discussion_r1967207191 solve your use case?

erezrokah:
Let's close this PR and open a separate issue to either add OTEL to the CLI or pass the relevant fields we need in plugins logs to plugins?

marianogappa:
Closing since we're solving this differently.

PR #20344, opened 2025-03-12
fix: Don't fail on creating existing tables

erezrokah:
#### Summary

Postgres does the same https://github.com/cloudquery/cloudquery/blob/809586281f83fc9e7c1a95a9f7ae412f731d6cde/plugins/destination/postgresql/client/migrate.go#L341.

There could still be schema mismatches if two concurrent syncs try to migrate the same table, but if that conflicts probably inserting will fail a bit after


hermanschaaf (approved):

maaarcelino (approved):

murarustefaan (approved):
Makes sense!

PR #20399, opened 2025-03-19
feat: Enable time substitution in spec templates

przste-go:
#### Summary

This PR aims to enable relative time support in all spec templates used is CQ platform without additional type overriding on plugin side. I.e following aws table options can be automatically substituted using yaml parser. 

```   table_options:
      aws_cloudtrail_events:
        lookup_events:
          - start_time: ${time:1 days ago}
            lookup_attributes:
              - attribute_key:   EventName
                attribute_value: RunInstances 
```

BEGIN_COMMIT_OVERRIDE
feat: Enable time substitution in spec templates (#20399). See our [docs](https://docs.cloudquery.io/docs/advanced-topics/environment-variable-substitution#time-variable-substitution-example) for more information.

END_COMMIT_OVERRIDE


cli/internal/specs/v0/testdata/time_substitution.yml
```diff
@@ -0,0 +1,16 @@
+kind: source
+spec:
+  name: aws
+  version: v1.3.3
+  destinations: [postgresql]
+  path: cloudquery/aws
+  tables: [test]
+  spec:
+    # Specifying non-relative time for easier testing
+    table_options: ${time:2025-01-01}
```
erezrokah, line 10: Can we use a real example for a table in the AWS plugin? So people don't get confused and think this is something that would work

murarustefaan:
I don't think this needs to be marked as a breaking change, but is rather a feature people can opt-in to use and it should not affect anything else otherwise. Am I wrong?

murarustefaan:
@przste-go while permitted by some implementations, they are not allowed or supported by the RFC.

https://pubs.opengroup.org/onlinepubs/9799919799/
> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2024 consist solely of uppercase letters, digits, and the <underscore> ('_') from the characters defined in [Portable Character Set](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap06.html#tagtcjh_3) and do not begin with a digit.

I don't think it's worth creating this as a breaking change for something that 
1. is not used the way it should be and
2. has a very unlikely chance of happening.

Realistically, I tried to create these ENV's and I was not able to.
```
bash-3.2$ export test:=test
bash: export: `test:=test': not a valid identifier
bash-3.2$ export "test:=test"
bash: export: `test:=test': not a valid identifier
```

przste-go:
sounds fair

cli/internal/specs/v0/testdata/time_substitution.yml
przste-go: Replaced with real table_options 👍

akash1810:
I love this change! How might I create a relative date of format `YYYY-MM-DD`? This format is required for the [`aws_costexplorer_cost_custom` table's](https://hub.cloudquery.io/plugins/source/cloudquery/aws/latest/tables/aws_costexplorer_cost_custom) `TimePeriod` option.

przste-go:
Hi @akash1810 we have that feature planned for current development cycle, so it should be publicly available in upcoming weeks 🚀

PR #20437, opened 2025-03-24
feat: Set Primary Key(s) for table

bbernays:
#### Summary


plugins/transformer/basic/client/schemaupdater/schema_updater.go
```diff
@@ -95,3 +97,26 @@ func (s *SchemaUpdater) ChangeTableName(newTableNamePattern string) (*arrow.Sche
...
+	table, err := schema.NewTableFromArrowSchema(s.schema)
+	if err != nil {
+		return nil, err
+	}
+	oldPks := table.PrimaryKeys()
+	for _, oldPk := range oldPks {
+		if !slices.Contains(newPks, oldPk) {
+			table.Columns.Get(oldPk).PrimaryKey = false
+		}
+	}
```
murarustefaan, line 111: Should we actually have it as a single-source-of-truth for PKs, or rather allow toggling of extra fields as PKs?
cc @erezrokah

erezrokah: We probably need both, add PKs and remove PKs directives, also how would this look like in the spec? I don't think this PR exposes it?

bbernays: This would be the spec a user would use:
```
kind: transformer
spec:
  name: "basic"
  path: "cloudquery/basic"
  version: v2.1.0
  spec:
    transformations:
      - kind: set_primary_keys
        tables: ["*"]
        columns: ["_cq_id", "_cq_source_name"]
```

murarustefaan: Yeah exactly, I think providing the ability to add PKs columns is the way to go, rather then specify the only columns that should be PKs.

bbernays: Changed to `AddPrimaryKeys`

erezrokah (commented):
Let's also make sure to update https://github.com/cloudquery/cloudquery/blob/8db20d6bc12153dc750857fe853ce2e3d95db65c/plugins/transformer/basic/docs/_configuration.md

bbernays:
> 461f7dd

Done [461f7dd](https://github.com/cloudquery/cloudquery/pull/20437/commits/461f7dd8c71161b49a93188e2deaf66c1d7aee8a)

plugins/transformer/basic/client/schemaupdater/schema_updater.go
```diff
@@ -95,3 +96,18 @@ func (s *SchemaUpdater) ChangeTableName(newTableNamePattern string) (*arrow.Sche
...
+
+func (s *SchemaUpdater) AddPrimaryKeys(newPks []string) (*arrow.Schema, error) {
+	table, err := schema.NewTableFromArrowSchema(s.schema)
+	if err != nil {
+		return nil, err
+	}
+	for _, newPk := range newPks {
+		newCol := table.Columns.Get(newPk)
+		if newCol == nil {
+			return nil, errors.New("new primary key column not found in schema")
```
erezrokah, line 108: Can we add the missing column name to the error so it's easier to know what caused the error? Maybe the table name too

erezrokah: Also can we make any column type a PK? Maybe support only the primitive types?

bbernays: I updated the error in [b85b61a](https://github.com/cloudquery/cloudquery/pull/20437/commits/b85b61a4adc235ecf574cf1ec964dc3ed27ba941)

bbernays: I would stay away from restricting the columns that can be used as a PK as this is an advanced feature... It all depends on how the destination handles a type. For example SQLite stringifies all complex columns so technically they could be used for a PK

erezrokah (commented):
One last question is about JSON Path support. We support it for removal and obfuscation, will this work here too? In case someone wants to use an internal JSON property as a PK?

I guess you'd have to flatten the JSON first?

erezrokah (approved):

bbernays:
> One last question is about JSON Path support. We support it for removal and obfuscation, will this work here too? In case someone wants to use an internal JSON property as a PK?
> 
> I guess you'd have to flatten the JSON first?

Yeah I think that is correct

PR #20438, opened 2025-03-25
feat: Remove pk index naming from postgres destination

przste-go:
#### Summary

This PR aims to solve issues with long table names in Postgresql database. All identifiers cannot be longer than 63 characters limit. Due to custom naming scheme of primary keys if table had more than 58 characters adding `_cqpk` caused DDLs to fail. To prevent that 2 things have been changed:

1. Primary keys no longer use `_cqpk` instead we fallback to default postgres naming scheme of `_pkey` which gets auto-incremented in case of conflicting name. This change is backwards compatible as it's part of create if not exists DDL.
2. If table name is longer than 58 characters - name is being hashed when creating perfomance index `_cqpi`. This change is also backwards compatible as table names longer than 58 characters couldn't be created due to `_cqpk` primary key identifier preventing DDL statements to execute.


plugins/destination/postgresql/client/migrate.go
```diff
@@ -401,8 +399,13 @@ func (c *Client) removeUniqueConstraint(ctx context.Context, table *schema.Table
 	return nil
 }
 
-func getPKName(table *schema.Table) string {
-	return table.Name + "_cqpk"
+func (c *Client) getPKName(ctx context.Context, tableName string) string {
+	var pkConstraintName string
+	_ = c.conn.QueryRow(ctx, `select tco.constraint_name from information_schema.table_constraints tco join information_schema.key_column_usage kcu
```
maaarcelino, line 404: Technically this query can fail. Do you think it's not worth catching that error?

plugins/destination/postgresql/client/migrate.go
```diff
@@ -430,3 +436,13 @@ func (c *Client) createPerformanceIndexes(ctx context.Context, table *schema.Tab
 
 	return nil
 }
+
+func hashTableName(input string) string {
+	// Max table name length
+	if len(input) > 63 {
+		input = input[:63]
```
maaarcelino, line 443: how come you are trimming the input to the hash function? should it not matter the length of the input?

przste-go: That's by design because all identifiers are going to be trimmed to 63 characters anyway on db side. So if I'd use full table name which would be i.e 65 characters it would cause 2 indexes to be created. One from 65 character hash, the other one from 63 character hash.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -383,6 +376,11 @@ func (c *Client) createTableIfNotExist(ctx context.Context, table *schema.Table)
 		c.logger.Error().Err(err).Str("table", tableName).Str("query", sb.String()).Msg("Failed to create table")
 		return fmt.Errorf("failed to create table %s: %w"+sb.String(), tableName, err)
 	}
+	c.pgTablesToPKConstraints[tableName] = &pkConstraintDetails{
+		name:    c.getPKName(ctx, tableName),
```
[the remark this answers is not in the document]
maaarcelino: Oh yeah, my bad

maaarcelino (approved):

erezrokah (commented):
Can we add some tests on migration existing tables with the old naming convention?

przste-go:
@erezrokah Added the test checking if old index names are untouched after migration.

plugins/destination/postgresql/client/migrate.go
```diff
@@ -401,8 +407,17 @@ func (c *Client) removeUniqueConstraint(ctx context.Context, table *schema.Table
...
 }
 
-func getPKName(table *schema.Table) string {
-	return table.Name + "_cqpk"
+func (c *Client) getPKName(ctx context.Context, tableName string) (string, error) {
+	var pkConstraintName string
+	err := c.conn.QueryRow(ctx, `select tco.constraint_name from information_schema.table_constraints tco join information_schema.key_column_usage kcu
+     on kcu.constraint_name = tco.constraint_name
+     where tco.constraint_type = 'PRIMARY KEY'
+    and kcu.table_name = $1`, tableName).Scan(&pkConstraintName)
```
murarustefaan, line 415: Maybe a nitpick but since the other queries are uppercased, let's uppercase this one as well.

murarustefaan (approved):

plugins/destination/postgresql/client/migrate.go
przste-go: done 🚀

erezrokah (approved):

PR #20544, opened 2025-04-11
feat: Support migrating addition of `_cq_client_id` when order by is set

erezrokah:
#### Summary

When someone enables (for supported plugins) the `add_cq_client_id: true` spec option migration will fail at the moment as we're adding a new non null column.

This PR supports such migration with the caveat that it only works if `order_by` was already configured explicitly, otherwise `_cq_client_id` will be implicitly considered in the new sort key as it's a non null column and we can't do sort key changes (yet). Technically we can modify the sort key when adding a new column, but we'd have to make sure it's a prefix of the primary key https://clickhouse.com/docs/sql-reference/statements/alter/order-by so don't think it's worth supporting that.

The operation is fast to do, as it changes metadata only, and we don't need to set a default value. It will use the zero value of the string type, see https://clickhouse.com/docs/sql-reference/statements/alter/column#add-column. The data will be filled upon reading from the table as needed.


plugins/destination/clickhouse/client/migrate.go
```diff
@@ -123,6 +123,11 @@ func (c *Client) dropTable(ctx context.Context, table *schema.Table) error {
 }
 
 func needsTableDrop(change schema.TableColumnChange) bool {
+	// Support for adding the cq_client_id column without dropping the table
+	if change.Type == schema.TableColumnChangeTypeAdd && change.Current.Name == schema.CqClientIDColumn.Name {
+		return false
```
blesniewski, line 128: Is there any way at this point to know if `order_by` was configured explicitly? If I understand the PR description correctly, it's either configured and we can add the column, or it's not, and we need to drop the table?

erezrokah: > and we need to drop the table?

If there's an `order_by` mismatch it will fail a bit later (regardless of the `_cq_client_id` change) with a proper error message since `order_by` changes also require a full drop.

https://github.com/cloudquery/cloudquery/blob/c44d7f10658c6c98fb1c5306859efee26a8d27dd/plugins/destination/clickhouse/client/migrate.go#L209

blesniewski (approved):

PR #20554, opened 2025-04-15
feat: Document how to escape dots and slashes in JSON paths

stoovon:
[Customer](https://cloudqueryio.slack.com/archives/C07DW5044MU/p1744708833085219) wanted to remove `annotations.kubectl.kubernetes.io/last-applied-configuration`.

sjson interprets that as

```
{
  "annotations": {
    "kubectl": {
      "kubernetes": ...
    }
  }
}
```

which of course doesn't match.

A workaround is to pass the following in the spec:

```
      - kind: remove_columns
        tables: ["k8s_core_secrets"]
        columns: ["annotations.kubectl\\.kubernetes\\.io\\/last-applied-configuration"]
```

This PR makes that more clear in the docs.

#### Summary

### DIFF
--- a/plugins/transformer/basic/client/recordupdater/record_updater.go
+++ b/plugins/transformer/basic/client/recordupdater/record_updater.go
@@ -30,6 +30,33 @@ func New(record arrow.Record) *RecordUpdater {
 	}
 }
 
+func escapeString(input string) string {
+	if !strings.ContainsRune(input, '\'') {
+		return input
+	}
+
+	result := make([]string, 0, 2*len(input))
+	parts := strings.Split(input, "'")
+	for i, part := range parts {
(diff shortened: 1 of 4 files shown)


erezrokah (commented):
I suggest we document `sjson` escape logic in the transformer docs (or if it's already documented make it more visible), per https://github.com/tidwall/sjson?tab=readme-ov-file#path-syntax

Not sure we'd want to support additional escaping patterns

stoovon:
> I suggest we document `sjson` escape logic in the transformer docs (or if it's already documented make it more visible), per https://github.com/tidwall/sjson?tab=readme-ov-file#path-syntax
> 
> Not sure we'd want to support additional escaping patterns

Docs fix done in baab2a5.

erezrokah (approved):

PR #20689, opened 2025-05-04
fix: Removed engine name logging

john-s58:


murarustefaan:
I don't really see a reason removing the engine name from logging.

PR #20690, opened 2025-05-06
chore: Stop releasing Gremlin and Meilisearch

przste-go:
#### Summary

Gremlin and meilisearch plugins are becoming depracated and this PR aims to stop them from releasing to the hub. 


maaarcelino:
@przste-go do you want to remove the github actions for these plugins too?

przste-go:
> for these plugins too?

Probably not as those are only running tests @maaarcelino. If we decide to move those plugins to archive we can remove the jobs too

maaarcelino (approved):

PR #20728, opened 2025-05-08
feat: Add 'auto_obfuscate' transformation to basic transformer

blesniewski:
#### Summary

Adds automatic obfuscation transformation basing on the new metadata field added in https://github.com/cloudquery/plugin-sdk/pull/2134


blesniewski:
Not ready yet:
- Needs the SDK PR
- docs need to be updated

Also, not sure if:
- we should add way to skip certain tables and to set that in spec
- modifying the default output of `obfuscate_columns` is ok, or if this should have it's own logic to do the obfuscation

blesniewski:
One question remaining:
If there's a JSON column, the current `obfuscate` transformation behavior is that it won't obfuscate the entire column- it requires a json path: `failed to transform schema: column tags is not a string column`

So 2 options:
- if we want this capability, I'll probably split the logic between `obfuscate` and `auto_obfuscate`, and add handling (or adjust the `obfuscate` behavior if we want to keep sharing)
- if we don't want to be able to obfuscate entire json columns, I'll adjust the validation in SDK

I assume we do want it, just putting the question out for confirmation

blesniewski:
I've added handling for obfuscating entire JSON columns.

Drawback of the current solution - to not have to unmarshal every value for each handled column, we're calculating hash of the entire value, which results in JSON output, but without preserving the internal structure:
`{"redacted_by_cloudquery": "81f2a9ddc7ae49a...hash_value"}`

Additionally, the output of the `obfuscate` transformation has changed, now the hashes values in would be preceded by `Redacted by CloudQuery | `

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -30,6 +33,9 @@ func New(record arrow.Record) *RecordUpdater {
 	}
 }
 
+const RedactedByCQMessage = "redacted by CloudQuery |"
```
murarustefaan, line 36:
```suggestion
const RedactedByCQMessage = "Redacted by CloudQuery |"
```

murarustefaan: Also, do these consts need to be public?

plugins/transformer/basic/client/transformers/transformers_test.go
```diff
@@ -136,8 +136,8 @@ func TestTransform(t *testing.T) {
 			},
 			record: createTestRecord(),
 			validate: func(t *testing.T, record arrow.Record) {
-				require.Equal(t, "bac8d4414984861d5199b7a97699c728bee36c4084299b2ca905434cf65d8944", record.Column(1).(*array.String).Value(0), "Expected sha256 value in new_col column")
-				require.Equal(t, "dd0fff6ac351dd46cd26e2d5c61e479ce7c68ef12489e04284c0fd66648723cb", record.Column(1).(*array.String).Value(1), "Expected sha256 value in new_col column")
+				require.Equal(t, "redacted by CloudQuery | bac8d4414984861d5199b7a97699c728bee36c4084299b2ca905434cf65d8944", record.Column(1).(*array.String).Value(0), "Expected sha256 value in new_col column")
+				require.Equal(t, "redacted by CloudQuery | dd0fff6ac351dd46cd26e2d5c61e479ce7c68ef12489e04284c0fd66648723cb", record.Column(1).(*array.String).Value(1), "Expected sha256 value in new_col column")
```
murarustefaan, line 140:
```suggestion
				require.Equal(t, "Redacted by CloudQuery | bac8d4414984861d5199b7a97699c728bee36c4084299b2ca905434cf65d8944", record.Column(1).(*array.String).Value(0), "Expected sha256 value in new_col column")
				require.Equal(t, "Redacted by CloudQuery | dd0fff6ac351dd46cd26e2d5c61e479ce7c68ef12489e04284c0fd66648723cb", record.Column(1).(*array.String).Value(1), "Expected sha256 value in new_col column")
```

murarustefaan (changes requested):

murarustefaan (approved):

PR #20767, opened 2025-05-15
feat: Add ability to change client project_id

przste-go:
#### Summary

In some scenarios users want to execute queries in context of different project than referenced in the query itself. I.e executing pricinipal has permissions to access table in project B only when executing within context of project A. This wasn't possible using only project_id spec. Now default project for execution can also be fetched from the environment/credentials by using `*detect-project-id*` variable.


maaarcelino (approved):

plugins/destination/bigquery/docs/_configuration.md
```diff
@@ -18,9 +18,13 @@ spec:
...
 
 This example above expects the following environment variables to be set:
 
   * `PROJECT_ID` - The Google Cloud Project ID
-  * `DATASET_ID` - The Google Cloud BigQuery Dataset ID
\ No newline at end of file
+  * `DATASET_ID` - The Google Cloud BigQuery Dataset ID
+
+`client_project_id` variable can be used to run BigQuery queries in a project different from where the destination table is located. 
+If you set client_project_id to `*detect-project-id*`, it will automatically detect the project ID from the environment variable or application default credentials.
```
murarustefaan, line 30: @przste-go why not set the default to this? 🤔

przste-go: Because that'd change previous behaviour. I.e I'm logged to gcp using user principal which doesn't have project_id set, that'd cause plugin to start failing. Now it behaves exactly as it did with an option to reconfigure that in spec.

PR #20772, opened 2025-05-16
feat!: Add DeleteRecord handling to Clickhouse destination

maaarcelino:
This adds handling for `DeleteRecord` messages in ClickHouse along with tests. I've also enabled the `DeleteRecord` tests in Postgres as they were excluded by mistake.


plugins/destination/clickhouse/client/delete.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	for _, msg := range messages {
+		sql := generateDelete(msg.DeleteRecord)
+		params, err := extractPredicateValues(msg.DeleteRecord.WhereClause)
+		if err != nil {
+			return err
+		}
+
+		if err = c.conn.Exec(ctx, sql, params...); err != nil {
```
maaarcelino, line 26: I'd like to be able to use some kind of prepared statement or batching here, but CH does not support this for `DELETE` statements. Batching is only supported for `INSERT` queries.

.github/workflows/dest_clickhouse.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         working-directory: plugins/destination/clickhouse
     services:
       clickhouse:
-        image: clickhouse/clickhouse-server:22.1.2
+        image: clickhouse/clickhouse-server:24.12.1
```
erezrokah, line 33: Is this needed to support deletion? If so we should do a major version bump and update the docs https://hub.cloudquery.io/plugins/destination/cloudquery/clickhouse/latest/docs#overview

plugins/destination/clickhouse/client/delete.go
```diff
@@ -0,0 +1,99 @@
...
+	for i := range v.Len() {
+		r[i] = v.Index(i).Interface()
+	}
+	return r
+}
+
+func generateDelete(msg message.DeleteRecord) string {
+	var sb strings.Builder
+
+	sb.WriteString("DELETE FROM ")
```
erezrokah, line 68: This only works if the engine is of the `MergeTree` family, we should either document this limitation or check the engine and do `ALTER TABLE [table] DELETE` if it's not in the `MergeTree` family, per https://clickhouse.com/docs/deletes/overview

maaarcelino: We only support `MergeTree` engines in this plugin as per https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/clickhouse/client/spec/engine.go#L63-L65.

erezrokah: Good call, then it's fine we don't need to change anything here

przste-go (approved):

.github/workflows/dest_clickhouse.yml
maaarcelino: Yes that's to support the lightweight delete statements. Those were added in CH 23.3 ([release notes](https://clickhouse.com/blog/clickhouse-release-23-03)). The current latest version of CH is 25.4.4. I'm wondering what we should set as the new minimum version for this plugin. WDYT? 23.3?

erezrokah: Maybe even 25.3 as 24.8 will go out of support soon, and 25.3 has JSON type production level support

maaarcelino: Yeah it does need to be breaking, agreed. I think jumping to 25.x will be a problem for our platform tenants. On platform-preview we run version 24.12.1 of CH. I think I'll bump the plugin to 24.8.1.

PR #20870, opened 2025-06-04
feat: Transform table name for DeleteRecord

przste-go:
#### Summary

Plugins like AWS CUR send DeleteRecord as a first message. This means that tableNameChanger logic (transformer table rename) will not properly rename the table causing delete to be applied against wrong or non-existent table. This PR aims to trigger table renaming by sending dummy message to table transformers.


cli/cmd/sync_v3.go
```diff
@@ -709,3 +733,32 @@ func deleteStale(client safeWriteClient, tables map[string]bool, sourceName stri
...
+	md := arrow.NewMetadata([]string{schema.MetadataTableName}, []string{tableName})
+
+	// Create a minimal schema with one dummy string field
+	return arrow.NewSchema(
+		[]arrow.Field{
+			{Name: "_dummy", Type: arrow.BinaryTypes.String},
+		},
+		&md,
+	)
+}
```
murarustefaan, line 749: Shouldn't we process the entire schema, though? What is there are column changes as well?

przste-go: This was never implemented for delete messages but I see your point here. To properly solve that issue we would need to make deleteMessage a proper arrow record, but that'd mean that other kind of transformation would be happening on it which are not really necessary. As an alternative we could maybe create schema from already existing column names in the message and try to rename them based on transformer response 🤔. Though if someone uses RemoveColumn transformation that could have weird consequences.
 I think that this issue is unique to CUR plugin though, so maybe it's just something we don't have to implement just yet. But I'm open to making it better.

przste-go: @murarustefaan Wdyt ?

murarustefaan: I think we can do it incrementally, I'm fine with the current implementation.

PR #20872, opened 2025-06-05
fix: S3 destination avoid hang on migrate errors

disq:
When `write_empty_objects_for_empty_tables` option is used, if migration (writing empty object to the bucket) fails, the method returns early without closing the channel. This produces a hang in [StreamingBatchWriter](https://github.com/cloudquery/plugin-sdk/blob/main/writers/streamingbatchwriter/) (`clientCh <- r`)

### DIFF
--- a/plugins/destination/s3/client/write.go
+++ b/plugins/destination/s3/client/write.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"regexp"
 	"time"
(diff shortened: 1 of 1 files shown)


plugins/destination/s3/client/write.go
```diff
@@ -132,14 +134,12 @@ func (c *Client) MigrateTable(ctx context.Context, ch <-chan *message.WriteMigra
 		c.initializedTables[table.Name] = objKey
 		s, err := c.createObject(ctx, table, objKey)
 		if err != nil {
-			return err
-		}
-		err = s.Finish()
-		if err != nil {
-			return err
+			errs = errors.Join(errs, err)
```
disq, line 137: Not returning early is not perfect but the current contract with StreamingBatchWriter requires this channel to be consumed to completion.

bbernays (approved):

PR #20878, opened 2025-06-07
fix: Proper error join

disq:
`errors.Join(errs, err)` doesn't check if `errs` is already a `joinError`, so it ends up creating error pairs with multiple levels of wrapping.

Follow-up to https://github.com/cloudquery/cloudquery/pull/20872

### DIFF
--- a/plugins/destination/s3/client/write.go
+++ b/plugins/destination/s3/client/write.go
@@ -137,9 +137,7 @@ func (c *Client) MigrateTable(ctx context.Context, ch <-chan *message.WriteMigra
 			errs = append(errs, err)
 			continue
 		}
-		if err := s.Finish(); err != nil {
-			errs = append(errs, err)
-		}
+		errs = append(errs, s.Finish())
 	}
 	return errors.Join(errs...)
 }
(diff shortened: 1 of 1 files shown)


plugins/destination/s3/client/write.go
```diff
@@ -134,12 +134,12 @@ func (c *Client) MigrateTable(ctx context.Context, ch <-chan *message.WriteMigra
...
 		if err != nil {
-			errs = errors.Join(errs, err)
+			errs = append(errs, err)
 			continue
 		}
-		errs = errors.Join(errs, s.Finish())
+		errs = append(errs, s.Finish())
 	}
-	return errs
+	return errors.Join(errs...)
```
disq, line 145: We might also consider returning only the first few errors rather than the whole slice.

murarustefaan: I think this is fine for now, but if it's not a hard thing and would make sense, we could aggregate (de-dupe) errors by table and messages?

disq:
@murarustefaan I'm aiming to have the streamingbatchwriter stop processing and still work without hanging after the client handler returns an error (it won't need to drain the channel), if that makes sense, but we can also do that inside the plugin handlers too, since it's all table scoped.

PR #20882, opened 2025-06-09
fix: Enable S3 destination test connections

disq:
S3 destination test connections didn't work on some cases:

- if `test_write` was explicitly set to `false` (the default value when unspecified is `true`)
- If `{{SYNC_ID}}` is set in the conn path (it always failed with: `path contains {{SYNC_ID}}. Upgrade your CLI to use this path variable`) due to invocation ID from the CLI not being set for test connections

This PR solves both issues, by forcing `test_write` to be `true` for test connections only, and force-generating an invocation ID (separate from what the CLI generated or was provided, which is a current shortcoming of the proto)

### DIFF
--- a/plugins/destination/s3/client/test_connection.go
+++ b/plugins/destination/s3/client/test_connection.go
@@ -31,11 +31,13 @@ func NewConnectionTester(createClientFn NewClientFn) plugin.ConnectionTester {
 				return err
 			}
 			s.SetDefaults()
-			b := true
-			s.TestWrite = &b // Force test write to be enabled for connection testing
-			specBytes, err = json.Marshal(s)
-			if err != nil {
-				return fmt.Errorf("failed to marshal s3 spec: %w", err)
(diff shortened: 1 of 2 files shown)


plugins/destination/s3/client/test_connection.go
```diff
@@ -17,7 +21,29 @@ const (
 
 func NewConnectionTester(createClientFn NewClientFn) plugin.ConnectionTester {
 	return func(ctx context.Context, logger zerolog.Logger, specBytes []byte) error {
-		_, err := createClientFn(ctx, logger, specBytes, plugin.NewClientOptions{})
+		{
```
erezrokah, line 24: I would not change `TestWrite` as if someone disabled it they probably won't be happy having test files written to their bucket

disq:
Done in https://github.com/cloudquery/cloudquery/pull/20883

PR #20883, opened 2025-06-09
fix: Pass `invocationID` when initializing test connection

erezrokah:
#### Summary

Otherwise if the path contains `{{SYNC_ID}}` the test connection fails here https://github.com/cloudquery/cloudquery/blob/983062b2210f6a269743afda4c56c5df9e0cdc77/plugins/destination/s3/client/client.go#L66

A better solution would be to get the invocation id from the CLI but the current signature of `NewConnectionTester` doesn't allow for it.


erezrokah:
~~Done in https://github.com/cloudquery/cloudquery/pull/20882/files~~

disq (approved):

PR #20885, opened 2025-06-12
fix: SSL Config

bbernays:
#### Summary

### DIFF
--- a/plugins/destination/kafka/client/spec/spec.go
+++ b/plugins/destination/kafka/client/spec/spec.go
@@ -20,8 +20,6 @@ type TlsDetails struct {
 	KeyFile *string `json:"key_file_path,omitempty"`
 	// Path to the certificate authority file for TLS client authentication
 	CaFile *string `json:"ca_file_path,omitempty"`
-
-	Enabled bool
 }
 
 type Spec struct {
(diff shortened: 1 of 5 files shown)


plugins/destination/kafka/client/spec/spec.go
```diff
@@ -15,11 +15,13 @@ type topicDetails struct {
 
 type TlsDetails struct {
 	// Path to the certificate file for client authentication
-	CertFile *string `json:"cert_file,omitempty"`
+	CertFile *string `json:"cert_file_path,omitempty"`
 	// Path to the key file for client authentication
-	KeyFile *string `json:"key_file,omitempty"`
+	KeyFile *string `json:"key_file_path,omitempty"`
 	// Path to the certificate authority file for TLS client authentication
-	CaFile *string `json:"ca_file,omitempty"`
```
murarustefaan, line 22: If we intend to ship this as a non-breaking change we'd need to support the previous config keys as well (which is what I would recommend, with a mention in the docs that those are deprecated).

plugins/destination/kafka/client/spec/spec.go
```diff
@@ -66,6 +68,14 @@ func (s *Spec) SetDefaults() {
...
 		s.TopicDetails.ReplicationFactor = 1
 	}
+
+	if s.TlsDetails != nil && s.TlsDetails.CertFile != nil && s.TlsDetails.KeyFile != nil && s.TlsDetails.CaFile != nil {
+		s.TlsDetails.Enabled = true
+	} else {
+		s.TlsDetails = &TlsDetails{
+			Enabled: false,
+		}
+	}
```
murarustefaan: Maybe move Enabled as a function with the TlsDetails receiver (feel free to ignore if you think this approach is better)?

murarustefaan (changes requested):
We also need to update the new settings in the docs.

plugins/destination/kafka/client/spec/spec.go
bbernays: This was never released

plugins/destination/kafka/client/spec/spec.go
bbernays: Completed in [0bcdcef](https://github.com/cloudquery/cloudquery/pull/20885/commits/0bcdcefcd8dc4514a4d16a7f53060d2a425181d1)

plugins/destination/kafka/client/spec/spec.go
murarustefaan: Ah ok it's just coming in 5.5.0. Makes sense.

murarustefaan (approved):

PR #20895, opened 2025-06-17
feat: Configures otel log collector

wprzechrzta:
#### Summary

Adds support for open telemetry collector.

Tested on local Tilt cluster

Example otel records
```
───────
Timestamp:          2025-06-18 07:09:04.000000000
TimestampTime:      2025-06-18 07:09:04
TraceId:
SpanId:
TraceFlags:         0
SeverityText:       info
SeverityNumber:     9
ServiceName:        cloudquery-cli
Body:               table sync finished
ResourceSchemaUrl:  https://opentelemetry.io/schemas/1.26.0
ResourceAttributes: {'service.name':'cloudquery-cli','telemetry.sdk.language':'go','telemetry.sdk.name':'opentelemetry','telemetry.sdk.version':'1.36.0'}
ScopeSchemaUrl:
ScopeName:          cloudquery
ScopeVersion:
ScopeAttributes:    {}
LogAttributes:      {'client':'project:cq-playground','duration_ms':'2274.9195','errors':'0','invocation_id':'a219e90f-c8f2-4f55-84f1-c99fcacb9b69','module':'gcp-source','resources':'35','table':'gcp_storage_buckets'}

Row 55:
───────
Timestamp:          2025-06-18 07:09:08.000000000
TimestampTime:      2025-06-18 07:09:08
TraceId:
SpanId:
TraceFlags:         0
SeverityText:       info
SeverityNumber:     9
ServiceName:        cloudquery-cli
Body:               started call
ResourceSchemaUrl:  https://opentelemetry.io/schemas/1.26.0
ResourceAttributes: {'service.name':'cloudquery-cli','telemetry.sdk.language':'go','telemetry.sdk.name':'opentelemetry','telemetry.sdk.version':'1.36.0'}
ScopeSchemaUrl:
ScopeName:          cloudquery
ScopeVersion:
ScopeAttributes:    {}
LogAttributes:      {'grpc.component':'server','grpc.method':'Transform','grpc.method_type':'bidi_stream','grpc.service':'cloudquery.plugin.v3.Plugin','grpc.start_time':'2025-06-18T09:09:01+02:00','grpc.time_ms':'6126.349','invocation_id':'a219e90f-c8f2-4f55-84f1-c99fcacb9b69','module':'basic-transformer','peer.address':'','protocol':'grpc'}

Row 56:
```

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -246,7 +246,7 @@ func CloseLogFile() {
 }
 
 func otelConfig() (*otel.Config, error) {
-	insecureEndpoint, err := strconv.ParseBool(env.GetEnvOrDefault("OTEL_ENDPOINT_INSECURE", "true"))
+	insecureEndpoint, err := strconv.ParseBool(env.GetEnvOrDefault("OTEL_ENDPOINT_INSECURE", "false"))
 	if err != nil {
 		return nil, err
 	}


cli/internal/otel/opentelemetry.go
```diff
@@ -0,0 +1,185 @@
...
+	ctx          context.Context
+	invocationId string
+}
+
+func (h *otelLoggerHook) Run(e *zerolog.Event, level zerolog.Level, message string) {
+	record := otellog.Record{}
+	record.SetTimestamp(time.Now().UTC())
+	record.SetSeverity(otellogSeverity(level))
+	record.SetBody(otellog.StringValue(message))
+	// See https://github.com/rs/zerolog/issues/493, this is ugly but it works
```
erezrokah, line 106: I know this is how we do it in the plugin SDK, but there's a better way https://github.com/rs/zerolog/pull/682#issuecomment-2395677680, just I haven't had the time to update the SDK code.

However I'm good with this is @murarustefaan is good with it

cli/internal/otel/opentelemetry.go
```diff
@@ -0,0 +1,185 @@
...
+				attributeValue = otellog.BytesValue(v)
+			default:
+				attributeValue = otellog.StringValue(fmt.Sprintf("%v", v))
+			}
+			recordAttributes = append(recordAttributes, otellog.KeyValue{
+				Key:   k,
+				Value: attributeValue,
+			})
+		}
+		recordAttributes = append(recordAttributes, otellog.KeyValue{
```
erezrokah, line 156: Will log entries that have `invocation_id` already have that field duplicated or overridden?

wprzechrzta: These are overridden, no duplicates.

cli/internal/otel/opentelemetry.go
murarustefaan: I mean, the code in the PR comment you posted looks a bit more cleaner whilst achieving a similar outcome, so maybe let's rewrite this using that approach and we can do the SDK later.
@wprzechrzta

cli/cmd/root.go
```diff
@@ -79,7 +80,7 @@ func NewCmdRoot() *cobra.Command {
 			cmd.SilenceUsage = true
 			var err error
 
-			if logFile, err = initLogging(noLogFile, logLevel, logFormat, logConsole, logFileName); err != nil {
+			if logFile, loggingShutdownFn, err = initLogging(noLogFile, logLevel, logFormat, logConsole, logFileName); err != nil {
 				return err
 			}
```
murarustefaan: Maybe we should consider moving the OTEL initialization wrapper (at least) to the Sync call? 🤔 
That way we would not need to specify the logging endpoint again since it's already in the sync spec.

murarustefaan: @erezrokah I'm not sure whether there are any implications of doing this, but I do not see why we would not. We can also postpone doing this for later, but I thought since we are already touching this bit of code, we could do a small refactor there rather than add another environment variable.

erezrokah: > Maybe we should consider moving the OTEL initialization wrapper (at least) to the Sync call? 🤔
> That way we would not need to specify the logging endpoint again since it's already in the sync spec.

But then only the sync command logs would be sent, and not `migrate` `test connection`, etc.

erezrokah: > @erezrokah I'm not sure whether there are any implications of doing this, but I do not see why we would not

For the sake of the platform this is ok I think. It might make sense to make OTEL configurable at the CLI level and send OTEL data for all plugins and CLI (not only sources like we do now), but that's a bigger change and no one requested it 🙃

murarustefaan: Ok, this makes little sense since it would mean tighly coupling with the Source spec which is where the otel_endpoint is located at.

cli/cmd/logging.go
```diff
@@ -51,5 +56,16 @@ func initLogging(noLogFile bool, logLevel *enum.Enum, logFormat *enum.Enum, logC
...
+
+	otelEndpoint := env.GetEnvOrDefault("CLOUD_PLATFORM_OTEL_ENDPOINT", "")
+	if otelEndpoint != "" {
+		shutdownFn, err = otel.SetupOtel(context.Background(), log.Logger, otelEndpoint)
+		if err != nil {
+			return nil, shutdownFn, fmt.Errorf("failed to setup OpenTelemetry: %w", err)
+		}
+		log.Logger = log.Logger.Hook(otel.NewOTELLoggerHook(invocationUUID.String()))
+		log.Info().Str("otel_logs_endpoint", otelEndpoint).Msg("otel logs endpoint set")
+	}
```
murarustefaan: Let's move this outside the initLogging function to the root command, it would allow us to remove the redundant  `loggingShutdownFn` and also would make more sense since we should be able to enable OTEL if we'd want metrics / traces as well in the future.

murarustefaan: Also, rather than `CLOUD_PLATFORM_OTEL_ENDPOINT` let's use only `OTEL_ENDPOINT`.

Do we also need `OTEL_ENDPOINT_INSECURE`?

cli/internal/otel/opentelemetry.go
wprzechrzta: If you agree I would prefer to keep this implementation, I invested heavily in testing to ensure that results are correct. Also if I understood correctly we still need to parse payload and make conversion to otel record in logger.Write(buf []byte), eg. setting levels and LogAttributes. So code will be very similar only it will be based on buffer passed in Writer interface.

cli/cmd/logging.go
wprzechrzta: > Do we also need OTEL_ENDPOINT_INSECURE

for platform this is always true

wprzechrzta: I moved block outside, but still shutdownFn still needs to be called cobra.OnFinalize(){} hook, not in root command otherwise subcommands will execute on closed logger (nothing will be logged)

cli/cmd/root.go
wprzechrzta: Extracted this code outside initLogging() as suggested below

cli/internal/otel/opentelemetry.go
murarustefaan: Yeah I am good with that.

cli/cmd/logging.go
murarustefaan: > for platform this is always true

Regardless, the CLI does not only serve the platform so any change / feature would need to support the larger subset of usecases as long as it does not involve a huge amount of work. 

Therefore since this should be a small change, I would say we introduce the `_INSECURE` environment variable as well and we can definitely configure that in the relevant place as well, like we do to the endpoint.

murarustefaan (approved):

cli/cmd/root.go
```diff
@@ -224,3 +244,14 @@ func CloseLogFile() {
 		logFile.Close()
 	}
 }
+
+func otelConfig() (*otel.Config, error) {
+	insecureEndpoint, err := strconv.ParseBool(env.GetEnvOrDefault("OTEL_ENDPOINT_INSECURE", "true"))
```
murarustefaan: Let's maybe default this to false?

PR #20897, opened 2025-06-19
feat: Add support for overwrite-delete-stale in Clickhouse destination

maaarcelino:


plugins/destination/clickhouse/client/delete.go
```diff
@@ -9,15 +9,42 @@ import (
...
+
+func generateDeleteForDeleteStale(msg *message.WriteDeleteStale) string {
+	var sb strings.Builder
+	sb.WriteString("DELETE FROM ")
+	sb.WriteString(util.SanitizeID(msg.TableName))
+	sb.WriteString(" WHERE ")
+	sb.WriteString(util.SanitizeID(schema.CqSourceNameColumn.Name))
+	sb.WriteString(" = $1 AND ")
+	sb.WriteString(util.SanitizeID(schema.CqSyncTimeColumn.Name))
+	sb.WriteString(" < $2")
```
disq: We had to do some TZ conversion stuff for Snowflake, did you verify it's not required here?
https://github.com/cloudquery/cloudquery/blob/007bc00f0405f852db475e0ab28f8d8a11f4af1f/plugins/destination/snowflake/client/deletestale.go#L18-L21

maaarcelino: How did this manifest itself in Snowflake? I was basing this implementation based on Postgres, MSSQL and MySQL which don't require conversions. This also worked when I was testing it locally, but perhaps I missed an edge case test.

disq: The PR has the explanation https://github.com/cloudquery/cloudquery/pull/17843

maaarcelino: Added a TZ cast in [002b0b8](https://github.com/cloudquery/cloudquery/pull/20897/commits/002b0b8ae3b54ca57585f3cd5a44cf0ae8883281)

PR #20907, opened 2025-06-24
fix: Handle max concurrent queries errors

erezrokah:
#### Summary

This PR fixes an issue that's more evident on older ClickHouse versions (e.g. 22 that we don't support anymore) where `max_concurrent_queries` defaulted to `100`. On newer versions the default is 1500.

The result of hitting the limit is that some tables can be dropped unexpectedly, since `checkPartitionOrOrderByChanged` called from `checkForced` here https://github.com/cloudquery/cloudquery/blob/a69aa8ca76607ccb5f750775a1c23e7899ad8e69/plugins/destination/clickhouse/client/migrate.go#L31 will not return an error while `checkPartitionOrOrderByChanged` can return an error when called further down the line here https://github.com/cloudquery/cloudquery/blob/a69aa8ca76607ccb5f750775a1c23e7899ad8e69/plugins/destination/clickhouse/client/migrate.go#L149

Had to override the `max_concurrent_queries` configuration so using the docker compose file for that.

The solution implemented in this PR has 2 parts:
1. Save the state from when we check which tables need to be forced migration so we don't re-run the same queries again
2. Retry ClickHouse operations (tried to create a connection wrapper for this purpose, but that's a bit tricky since the error can be on `connection.Exec` `row.Scan` and so on, so I'd basically need to wrap all ClickHouse's driver interfaces.


plugins/destination/clickhouse/client/migrate.go
```diff
@@ -28,68 +32,103 @@ func (c *Client) MigrateTables(ctx context.Context, messages message.WriteMigrat
...
 			}()
 			if len(want.Columns) == 0 {
-				c.logger.Warn().Str("table", want.Name).Msg("Table with no columns, skip")
+				c.logger.Warn().Str("table", want.Name).Msg("Table with no columns, skipping")
 				return nil
 			}
 
-			have := have.Get(want.Name)
-			if have == nil {
+			tableName := want.Name
```
erezrokah, line 72: This is the part about re-using the tables changes state

plugins/destination/clickhouse/client/migrate.go
```diff
@@ -103,23 +142,21 @@ func unsafeChanges(changes []schema.TableColumnChange) []schema.TableColumnChang
 }
 
 func (c *Client) createTable(ctx context.Context, table *schema.Table, partition []spec.PartitionStrategy, orderBy []spec.OrderByStrategy) (err error) {
-	c.logger.Debug().Str("table", table.Name).Msg("Table doesn't exist, creating")
```
erezrokah, line 106: This was confusing as we called `dropTable` then `createTable` when doing force migration, so even if the table existed this log was printed

marianogappa (approved):
Congrats on finding and fixing this bug 👍 

My only concern here is the retry logic.

Obviously, without more context there's no smartness possible to the retry, like we do on InsertSplitter, so we agree that it's not gonna get a lot smarter than "try running the same query again later".

However, retrying up to 10 times, 1s +/- .5s later in the case of "Too many simultaneous queries" sounds 💣 🤔 no? I think this will by default increase congestion. My proposed remediation isn't very smart...I'd 10x the delay or something like that.

erezrokah:
> However, retrying up to 10 times, 1s +/- .5s later in the case of "Too many simultaneous queries" sounds 💣 🤔 no? I think this will by default increase congestion. My proposed remediation isn't very smart...I'd 10x the delay or something like that.

The default strategy is backoff + random jitter, but I don't mind increasing the retry delay. Based on my tests with the current values and 2000 concurrent syncs (from the test) we hardly every retry more than twice

plugins/destination/clickhouse/client/retry_helpers.go
```diff
@@ -0,0 +1,130 @@
...
+	onRetryFunc := func(logger zerolog.Logger, query string) func(n uint, err error) {
+		return func(attempt uint, err error) {
+			logger.Warn().Err(err).Uint("attempt", attempt).Msgf("Retrying query %s", query)
+		}
+	}
+
+	commonRetryOptions := []retry.Option{
+		retry.Attempts(5),
+		retry.Delay(3 * time.Second),
+		retry.MaxJitter(1 * time.Second),
```
maaarcelino, line 31: Would it make sense to expose these variables in the spec? Not blocking, just thinking out loud.

erezrokah: Not sure actually I think we can expose based on user feedback. Could be too much control. The default retry logic is backoff + random jitter so I think that should cover quite a lot of cases (especially on newer versions)

plugins/destination/clickhouse/client/retry_helpers.go
```diff
@@ -0,0 +1,130 @@
...
+	record, err := retry.DoWithData(
+		func() (arrow.Record, error) {
+			rows, err := conn.Query(ctx, queries.Read(table))
+			if err != nil {
+				return nil, err
+			}
+			defer rows.Close()
+
+			row := rowArr(rows.ColumnTypes())
+			builder := array.NewRecordBuilder(memory.DefaultAllocator, table.ToArrowSchema())
```
candiduslynx, line 114: you can create record builder once and then reset it before scanning in the retried function

PR #20908, opened 2025-06-25
feat: Add list of incremental tables to sync summary

disq:
Adds `IncrementalTables []string` to Sync Summary.

### DIFF
--- a/cli/cmd/summary.go
+++ b/cli/cmd/summary.go
@@ -36,6 +36,7 @@ type syncSummary struct {
 	SyncGroupID         *string   `json:"sync_group_id,omitempty"`
 	ShardNum            *int      `json:"shard_num,omitempty"`
 	ShardTotal          *int      `json:"shard_total,omitempty"`
+	IncrementalTables   []string  `json:"incremental_tables,omitempty"`
 }
 
 func persistSummary(filename string, summary syncSummary) error {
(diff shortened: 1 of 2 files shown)


cli/cmd/sync_v3.go
```diff
@@ -762,3 +766,14 @@ func getTransformedTableNameFromSchema(transformedSchemaBytes []byte) (string, e
 
 	return tableName, nil
 }
+
+func getIncrementalTables(tt schema.Tables) schema.Tables {
+	flattenedTables := tt.FlattenTables()
+	incTables := make(schema.Tables, 0, len(flattenedTables))
```
disq, line 772: We could over optimize and return names of _all_ and _incremental_ tables together from this function (as we've already flattened them) but I don't think it's needed.

cli/cmd/summary.go
```diff
@@ -36,6 +36,7 @@ type syncSummary struct {
 	SyncGroupID         *string   `json:"sync_group_id,omitempty"`
 	ShardNum            *int      `json:"shard_num,omitempty"`
 	ShardTotal          *int      `json:"shard_total,omitempty"`
+	IncrementalTables   []string  `json:"incremental_tables,omitempty"`
```
disq, line 39: `[]bool` would also work, but it would be harder to use.

disq:
Closing for now.

PR #20909, opened 2025-06-25
feat: Add error handling and per-table metrics

przste-go:
#### Summary

This PR adds handling of error messages from source plugins and add new statistics for platform calls. This solves two things:

1. Having proper error statistics when using plugins in development mode 
2. Adding ability to show detailed per table metrics in platform without need to rely on otel_traces


cli/cmd/sync_v3.go
```diff
@@ -608,27 +636,19 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
...
 	totals = sourceClient.Metrics()
 	sourceWarnings := totals.Warnings
-	sourceErrors := totals.Errors
+	var sourceErrors uint64
+	for _, val := range statsPerTable {
+		sourceErrors += uint64(val.Errors)
+	}
+	if totals.Errors > sourceErrors {
+		sourceErrors = totals.Errors
+	}
```
murarustefaan, line 651: Why? 🤔 Shouldn't these 2 be equal?

cli/cmd/sync_v3.go
```diff
@@ -373,10 +391,11 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
 				status = cloudquery_api.SyncRunStatusCompleted
 			}
 			obj := cloudquery_api.CreateSyncRunProgressJSONRequestBody{
-				Rows:     atomic.LoadInt64(&totalResources),
-				Errors:   int64(totals.Errors),
-				Warnings: int64(totals.Warnings),
-				Status:   &status,
+				Rows:          atomic.LoadInt64(&totalResources),
+				Errors:        int64(totals.Errors),
+				Warnings:      int64(totals.Warnings),
```
murarustefaan, line 396: How come we don't need to use atomics for these?

cli/cmd/sync_v3.go
przste-go: When using old plugin version that are not supporting error messages yet sourceErrors will be equal to 0. I pressume I can switch that check to if sourceErrors==0

cli/cmd/sync_v3.go
przste-go: Actually those ones are old counters. But we don't need atomics because counter is being increased in single thread. Currently all messages are being consumed in single gorutine and counter is being incremented there i.e
```				
                                  stats := statsPerTable[tableName]
				stats.Rows += record.NumRows()
				statsPerTable[tableName] = stats
```
but read in multiple gorutines, which means there is no benefit on adding atomics unless we plan to increment counter from another gorutine.

cli/cmd/sync_v3.go
przste-go: On top of that, there is also one more case when those number differ. totals.Errors is taken from the log parsing, so if there is an error log not associated with the table this totals.Errors will be bigger than sourceErrors 🤷

murarustefaan: Yup, makes **total** sense.

murarustefaan (approved):

PR #20916, opened 2025-06-26
feat: Enable Role Assumption for S3 Destination

bbernays:
#### Summary

Prior to this change a user could not assume a role without first generating a config file and using `local_profile`.  This change enables all of the role assumption that is possible in the AWS Source plugin, without the need for a local config file.

This is fully backwards compatible


murarustefaan (commented):
This looks good. Let's just make sure we document these in `docs/`

bbernays:
> This looks good. Let's just make sure we document these in `docs/`

completed in [fbd4954](https://github.com/cloudquery/cloudquery/pull/20916/commits/fbd49543b2759f76b1a609ef8db1387be5181e77)

murarustefaan (approved):

PR #20917, opened 2025-06-26
feat: Enable Dropping of Rows based on field values 

bbernays:
#### Summary

Example spec that enables users to dropdefault RDS Parameters
```
kind: source
spec:
  destinations: ["postgresql"]
  name: aws
  path: cloudquery/aws
  registry: cloudquery
  version: 'v32.21.0'
  tables:
    - aws_rds*
---
kind: transformer
spec:
  name: "basic"
  # path: "cloudquery/basic"
  # version: v2.4.0
  registry: grpc
  path: "localhost:7777"
  spec:
    transformations:
      - kind: drop_rows
        tables: ["aws_rds*parameters"]
        columns: ["source"]
        value: "system"
      - kind: drop_rows
        tables: ["aws_rds*parameters"]
        columns: ["source"]
        value: "engine-default"  

---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  registry: "cloudquery"
  version: "v8.8.3"
  transformers:
    - "basic"
  spec:
    connection_string: "postgres://postgres:pass@localhost:5432/postgres?sslmode=disable"
    ```

### CODE REVIEW

**Review by murarustefaan — COMMENTED** [review]
All good code-wise, but let's also add a usage example to `docs/_configuration` before merging.

File: plugins/transformer/basic/client/transformers/transformers_test.go [hunk 1]
```diff
@@ -186,6 +186,56 @@ func TestTransform(t *testing.T) {
 				require.Equal(t, "cq_sync_table1", newTableName)
 			},
 		},
+		{
```
erezrokah, line 189: Can we add some tests on non string values in records? `ValueStr` and `AppendValueFromString` have some gotchas depending on the types (e.g. not sure `AppendValueFromString` on `ValueStr` preserves the original value for all types).

Maybe try with:
- UUID
- Timestamp
- Floats
- Ints
- Lists
- Maps
- JSON

Another option is to do this only for string types

bbernays: More tests in [85ea7c7](https://github.com/cloudquery/cloudquery/pull/20917/commits/85ea7c73f2f479ec903379a47841b6712cef9d51)

but that change is reliant on [this PR](https://github.com/cloudquery/plugin-sdk/pull/2197) to make each row have different values. When every row has the same value, then all rows get dropped.

bbernays: I will document that it is not tested against Lists and Maps and therefore not supported

erezrokah: You could create the record manually instead of https://github.com/cloudquery/plugin-sdk/pull/2197 as well

bbernays: I would prefer to keep this kind of record generation in a single place and reuse it. Most of the fields are already dynamic, there were just a few that were hardcoded... Open to going in either direction that the team wants though.

plugins/transformer/basic/client/recordupdater/record_updater_test.go
```diff
@@ -160,6 +160,66 @@ func TestAutoObfuscateEntireJSONColumn(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, `1492571184685610752`, updatedRecord.Column(8).(*array.Uint64).ValueStr(0))
+	require.Equal(t, int64(6), updatedRecord.NumRows())
+
+	updatedRecord, err = updater.DropRows([]string{"date64"}, "2023-05-03")
+	require.NoError(t, err)
+	require.Equal(t, `2023-06-12`, updatedRecord.Column(19).(*array.Date64).ValueStr(0))
+	require.Equal(t, int64(5), updatedRecord.NumRows())
+
+	updatedRecord, err = updater.DropRows([]string{"timestamp_ns"}, "2025-06-27 10:40:35.000914Z")
```
erezrokah: How come dropping `timestamp_ns` removes the remaining 5 rows?

plugins/transformer/basic/client/recordupdater/record_updater_test.go
```diff
@@ -160,6 +160,66 @@ func TestAutoObfuscateEntireJSONColumn(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, "a6:ae:92:fb:b5:2c", updatedRecord.Column(16).(*types.MACArray).ValueStr(0))
+	require.Equal(t, int64(9), updatedRecord.NumRows())
+
+	updatedRecord, err = updater.DropRows([]string{"inet"}, "176.32.0.0/12")
+	require.NoError(t, err)
+	require.Equal(t, "139.0.0.0/10", updatedRecord.Column(15).(*types.InetArray).ValueStr(0))
+	require.Equal(t, int64(8), updatedRecord.NumRows())
+
+	updatedRecord, err = updater.DropRows([]string{"json"}, `{"test":["a","b",99733]}`)
```
erezrokah: Would this would work only if the JSON is compacted? Or does dropping `{ "test" : ["a", "b", 99733 ] }` works too?

plugins/transformer/basic/client/recordupdater/record_updater_test.go
bbernays: because all timestamp columns have the same value

erezrokah: Can we add a timestamp column that doesn't get removed so we can verify its content?

plugins/transformer/basic/client/transformers/transformers_test.go
murarustefaan: > I would prefer to keep this kind of record generation in a single place and reuse it

Yeah, same here.

plugins/transformer/basic/client/recordupdater/record_updater_test.go
bbernays: The value that the user specifies, must be compacted. Will document this

plugins/transformer/basic/client/recordupdater/record_updater_test.go
bbernays: I will add a separate test for this

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -158,6 +158,54 @@ func (r *RecordUpdater) ObfuscateSensitiveColumns() (arrow.Record, error) {
...
+		for j := 0; j < arr.Len(); j++ {
+			if slices.Contains(rowsToDrop, j) {
+				continue
+			}
+			if arr.IsNull(j) {
+				builder.AppendEmptyValue()
+				continue
+			}
+
+			if err := builder.AppendValueFromString(arr.ValueStr(j)); err != nil {
```
erezrokah: I think it's probably better to do something like this:

```
switch typ := arr.(type) {
case *types.MACArray:
	builder.(*types.MACBuilder).Append(typ.Value(j))
case *types.InetArray:
	builder.(*types.InetBuilder).Append(typ.Value(j))
case *types.JSONArray:
	builder.(*types.JSONBuilder).Append(typ.Value(j))
case *array.Binary:
	builder.(*array.BinaryBuilder).Append(typ.Value(j))
case *array.Date64:
	builder.(*array.Date64Builder).Append(typ.Value(j))
case *array.Timestamp:
	builder.(*array.TimestampBuilder).Append(typ.Value(j))
...
}
```

To avoid the conversion to string and back

erezrokah:
OK thinking about this a bit more, what's the reason we have `columns: ["source"]` (an array) and not a single column?
Can you map a single value to multiple columns? Also how would that work with this code?

Let's say you remove 2 rows from `column_1` then 3 rows from `column_2`, won't that create a mismatch of rows in the arrow record?

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -158,6 +158,54 @@ func (r *RecordUpdater) ObfuscateSensitiveColumns() (arrow.Record, error) {
...
+	rowsToDrop := make([]int, 0, r.record.NumRows())
+	for j, column := range cols {
+		if !slices.Contains(columnNames, r.record.ColumnName(j)) {
+			continue
+		}
+		for i := range column.Len() {
+			if slices.Contains(rowsToDrop, i) {
+				continue
+			}
+			if (!column.IsValid(i) && value == "") || (column.IsValid(i) && column.ValueStr(i) == value) {
```
erezrokah: `column.IsValid` means a `nil` value, don't we want to distinguish `nil` from empty string?

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -158,6 +158,54 @@ func (r *RecordUpdater) ObfuscateSensitiveColumns() (arrow.Record, error) {
...
+		return r.record, nil
+	}
+
+	newRowCount := int(r.record.NumRows()) - len(rowsToDrop)
+	newCols := make([]arrow.Array, len(cols))
+	newRow := 0
+	for _, arr := range cols {
+		builder := array.NewBuilder(memory.DefaultAllocator, arr.DataType())
+		for j := 0; j < arr.Len(); j++ {
+			if slices.Contains(rowsToDrop, j) {
```
erezrokah: Using a `map` for `rowsToDrop` might be more suitable

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -158,6 +158,54 @@ func (r *RecordUpdater) ObfuscateSensitiveColumns() (arrow.Record, error) {
...
+				continue
+			}
+			if arr.IsNull(j) {
+				builder.AppendEmptyValue()
```
erezrokah:
```suggestion
				builder.AppendNull()
```

Empty value is not the same as null values. This will replace all null values with empty ones

plugins/transformer/basic/client/recordupdater/record_updater.go
bbernays: Done in [f830fce](https://github.com/cloudquery/cloudquery/pull/20917/commits/f830fced38dfc49ee36a44af2088bf5dca43290b)

plugins/transformer/basic/client/recordupdater/record_updater.go
bbernays: If a user wanted to drop rows with a `nil` value how would they represent that in the yaml?

erezrokah:
OK came up with this (needs testing):

```
slices := make([]arrow.Record, 0, len(rowsToDropIndexes))
currentSliceStart := 0
for _, row := range rowsToDropIndexes {
	slices = append(slices, r.record.NewSlice(int64(currentSliceStart), int64(row)))
	currentSliceStart = row + 1
}

schema := r.record.Schema()
concatenatedCols := make([]arrow.Array, int(r.record.NumCols()))

for i := 0; i < int(r.record.NumCols()); i++ {
	var colChunks []arrow.Array
	for _, slice := range slices {
		colChunks = append(colChunks, slice.Column(i))
	}
	concat, err := array.Concatenate(colChunks, memory.DefaultAllocator)
	if err != nil {
		return nil, fmt.Errorf("failed to concatenate arrays: %w", err)
	}
	concatenatedCols[i] = concat
}
length := int64(concatenatedCols[0].Len())
result := array.NewRecord(schema, concatenatedCols, length)
```

This ⬆️ slices out the rows we want to remove, then merges everything back to a single record

plugins/transformer/basic/client/recordupdater/record_updater.go
erezrokah: We can make `value` a `*string`? Then omitting is from the spec means dropping nulls. Another option is to add a new spec option that says `drop_nulls` to make it clear

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -158,6 +158,55 @@ func (r *RecordUpdater) ObfuscateSensitiveColumns() (arrow.Record, error) {
...
+		return r.record, nil
+	}
+
+	newRowCount := int(r.record.NumRows()) - len(rowsToDrop)
+	newCols := make([]arrow.Array, len(cols))
+	newRow := 0
+	for _, arr := range cols {
+		builder := array.NewBuilder(memory.DefaultAllocator, arr.DataType())
+		for j := 0; j < arr.Len(); j++ {
+			if _, ok := rowsToDrop[j]; ok {
```
erezrokah:
```suggestion
			if rowsToDrop[j] {
```

works too since we're using a map of boolean values

plugins/transformer/basic/client/recordupdater/record_updater.go
bbernays: No longer needed after [420e2f3](https://github.com/cloudquery/cloudquery/pull/20917/commits/420e2f352cd59c13af38c307f238e01a26cc93cf)

plugins/transformer/basic/client/recordupdater/record_updater.go
bbernays: Converted the `value` to take a `*string` so we can tell the difference between empty string and nil

plugins/transformer/basic/client/recordupdater/record_updater_test.go
bbernays: Test added in [98e4245](https://github.com/cloudquery/cloudquery/pull/20917/commits/98e4245d8653f526400cb5da3a87ee363ff34fe3)

plugins/transformer/basic/docs/_configuration.md
```diff
@@ -36,5 +36,9 @@ spec:
       - kind: lowercase
         tables: ["xkcd_comics"]
         columns: ["title"]
+      - kind: drop_rows
+        tables: ["xkcd_comics"]
+        columns: ["year"]
+        values: "2023"
```
erezrokah:
```suggestion
        value: "2023"
```

Shouldn't this be `value`?

plugins/transformer/basic/docs/overview.md
```diff
@@ -58,3 +59,11 @@ You can also use the `obfuscate_sensitive_columns` transformation to automatical
 
 Note: transformations are applied sequentially. If you rename tables, the table matcher configuration of subsequent transformations will need to be updated to the new names.
 Note: escape syntax is [SJSON sytax](https://github.com/tidwall/sjson?tab=readme-ov-file#path-syntax).
+
+
+
+
+Edge Cases and limitations for `drop_rows` transformation:
+- The `drop_rows` transformation only supports non-list columns
+- If you are trying to drop based on a nil value in the column, you should leave the `values` field empty
```
erezrokah:
```suggestion
- To drop rows with `nil` values, configure `value: null`, `value: ~` or drop the `value` configuration altogether
```

plugins/transformer/basic/docs/overview.md
```diff
@@ -58,3 +59,11 @@ You can also use the `obfuscate_sensitive_columns` transformation to automatical
 
 Note: transformations are applied sequentially. If you rename tables, the table matcher configuration of subsequent transformations will need to be updated to the new names.
 Note: escape syntax is [SJSON sytax](https://github.com/tidwall/sjson?tab=readme-ov-file#path-syntax).
+
+
+
+
+Edge Cases and limitations for `drop_rows` transformation:
+- The `drop_rows` transformation only supports non-list columns
```
erezrokah:
```suggestion
- Only non-list columns are supported
```

We don't need to repeat `drop_rows`, we're already in the context of `drop_rows` limitations

plugins/transformer/basic/docs/overview.md
```diff
@@ -58,3 +59,11 @@ You can also use the `obfuscate_sensitive_columns` transformation to automatical
...
+- If you are trying to drop based on a nil value in the column, you should leave the `values` field empty
+- If you are trying to drop based on a JSON value, make sure that the value is the compacted version of the JSON. For example, if you want to drop rows where a JSON column `tags` has a value of `{"foo": "bar"}`, you should specify the value as `{"foo":"bar"}` without any whitespace.
```
erezrokah:
```suggestion
- To drop rows based on a JSON value, use the compacted version of the JSON. For example, if you want to drop rows where a JSON column `tags` has a value of `{"foo": "bar"}`, you should specify the value as `{"foo":"bar"}` without any whitespace.
```

erezrokah:
Looks great, added a few comments regarding the docs. One thing I'm still not sure about is the reason to have multiple `columns` mapped to a single value.

When are we expecting multiple columns to have the same value? What's the use case there? Maybe only for removing null values?

Otherwise seems a bit confusing

bbernays:
> Looks great, added a few comments regarding the docs. One thing I'm still not sure about is the reason to have multiple `columns` mapped to a single value.
> 
> When are we expecting multiple columns to have the same value? What's the use case there? Maybe only for removing null values?
> 
> Otherwise seems a bit confusing

I will handle the docs updates, but the comments look helpful!

As for multiple columns, yes the idea is that multiple tables could have different columns but the same value (`nil`). I also don't really want to add another attribute to the `spec`. I don't understand why the use of an array would be a super complicating feature. I think it would actually be detrimental to the usability of the plugin as a whole to have `column` and `columns` as valid inputs

erezrokah:
> usability of the plugin as a whole to have `column` and `columns` as valid inputs

Having both also works, agree `columns` is useful to quickly strip `nil` values

PR #20981, opened 2025-07-08
fix: Fix nested column obfuscation

przste-go:
#### Summary

This PR fixes obfuscation of nested columns

### DIFF
--- a/plugins/transformer/basic/client/recordupdater/record_updater.go
+++ b/plugins/transformer/basic/client/recordupdater/record_updater.go
@@ -433,8 +433,8 @@ func (*RecordUpdater) obfuscateJSONColumns(column arrow.Array, jcs []jsonColumn)
 		str := column.ValueStr(i)
 		for _, jc := range jcs {
 			val := gjson.Get(column.ValueStr(i), jc.columnPath)
-			if val.Exists() && val.Type == gjson.String {
-				if modified, err := sjson.Set(str, jc.columnPath, fmt.Sprintf("%s %x", redactedByCQMessage, sha256.Sum256([]byte(val.Str)))); err == nil {
+			if val.Exists() {
+				if modified, err := sjson.Set(str, jc.columnPath, fmt.Sprintf("%s %x", redactedByCQMessage, sha256.Sum256([]byte(val.Raw)))); err == nil {
 					str = modified
 					continue
 				}
--- a/plugins/transformer/basic/client/transformers/transformers_test.go
+++ b/plugins/transformer/basic/client/transformers/transformers_test.go
@@ -1,6 +1,7 @@
 package transformers
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/apache/arrow-go/v18/arrow"
@@ -8,6 +9,7 @@ import (
 	"github.com/apache/arrow-go/v18/arrow/memory"
 	"github.com/cloudquery/cloudquery/plugins/transformer/basic/client/spec"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
+	"github.com/cloudquery/plugin-sdk/v4/types"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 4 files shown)


plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -433,8 +433,8 @@ func (*RecordUpdater) obfuscateJSONColumns(column arrow.Array, jcs []jsonColumn)
 		str := column.ValueStr(i)
 		for _, jc := range jcs {
 			val := gjson.Get(column.ValueStr(i), jc.columnPath)
-			if val.Exists() && val.Type == gjson.String {
```
przste-go, line 436: This check was biggest problem because gjson in case of smth.#.X queries will return an array rather than a string

plugins/transformer/basic/client/recordupdater/record_updater_test.go
```diff
@@ -121,10 +122,10 @@ func TestAutoObfuscateColumns(t *testing.T) {
 		fmt.Sprintf("%s 528e5290f8ff0eb0325f0472b9c1a9ef4fac0b02ff6094b64d9382af4a10444b", redactedByCQMessage),
 		updatedRecord.Column(0).(*array.String).Value(1))
 	assert.Equal(t,
-		fmt.Sprintf(`{"foo":{"bar":["%s ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb","%s 3e23e8160039594a33894f6564e1b1348bbd7a0088d42c4acb73eeaed59c009d","c"]},"hello":"world"}`, redactedByCQMessage, redactedByCQMessage),
+		fmt.Sprintf(`{"foo":{"bar":["%s ac8d8342bbb2362d13f0a559a3621bb407011368895164b628a54f7fc33fc43c","%s c100f95c1913f9c72fc1f4ef0847e1e723ffe0bde0b36e5f36c13f81fe8c26ed","c"]},"hello":"world"}`, redactedByCQMessage, redactedByCQMessage),
```
przste-go, line 125: Not sure if we consider that a breaking change if it's only affecting json columns. I could probably make it backwards compatible if needed too

przste-go:
Only downside of this approach is that nested columns will be obfuscated with single sha value due to array being returned by gjson
{
  "top_foo" : [ {
    "foo" : "Redacted by CloudQuery | efe3e67a3f277ec61b7857edfa2df04488f78b8b9ef1ecd4b388ff656f143df3"
  }, {
    "foo" : "Redacted by CloudQuery | efe3e67a3f277ec61b7857edfa2df04488f78b8b9ef1ecd4b388ff656f143df3"
  }, {
    "foo" : "Redacted by CloudQuery | efe3e67a3f277ec61b7857edfa2df04488f78b8b9ef1ecd4b388ff656f143df3"
  } ]
}

plugins/transformer/basic/client/recordupdater/record_updater.go
```diff
@@ -433,8 +433,8 @@ func (*RecordUpdater) obfuscateJSONColumns(column arrow.Array, jcs []jsonColumn)
...
-				if modified, err := sjson.Set(str, jc.columnPath, fmt.Sprintf("%s %x", redactedByCQMessage, sha256.Sum256([]byte(val.Str)))); err == nil {
+			if val.Exists() {
+				if modified, err := sjson.Set(str, jc.columnPath, fmt.Sprintf("%s %x", redactedByCQMessage, sha256.Sum256([]byte(val.Raw)))); err == nil {
```
bbernays, line 438:
```suggestion
			if val.Exists() {
				if val.Type != gjson.String {
					// Check if value is an array and if so iterate through its elements and replace the # in the column path with the index
					for i, arrayVal := range val.Array() {
						updatedColumnPath := strings.Replace(jc.columnPath, "#", fmt.Sprintf("%d", i), 1)
						if modified, err := sjson.Set(str, updatedColumnPath, fmt.Sprintf("%s %x", redactedByCQMessage, sha256.Sum256([]byte(arrayVal.Raw)))); err == nil {
							str = modified
						}

					}
				} else {
					if modified, err := sjson.Set(str, jc.columnPath, fmt.Sprintf("%s %x", redactedByCQMessage, sha256.Sum256([]byte(val.Raw)))); err == nil {
						str = modified
						continue
					}
				}
```
What about something like this?

murarustefaan (approved):

plugins/transformer/basic/client/recordupdater/record_updater.go
przste-go: Yeah, that kinda works for basic scenario, but for deeply nested cases like 'object1.object2.#.nested_object1.nested_object2.#.nested2_object1' we are receiving array of arrays of values '[[1,2],[3,4]]' 
so we'd need replace # with indexes of elements. 
like:
object1.object2.0.nested_object1.nested_object2.0.nested2_object1
object1.object2.0.nested_object1.nested_object2.1.nested2_object1
etc. Of course that's also possible to do.
On top of that to be fully compliant we'd need to handle #() which is also a valid query that let's you conditionally select elements. 
image

przste-go: As nested json values are shouldn't be used as pk/pkc we decided with @murarustefaan to not implement that right now. But I've added todo to re-consider that in the future

PR #21021, opened 2025-07-14
feat: Enable Using AWS Creds with MongoDB Atlas 

bbernays:
#### Summary


erezrokah (commented):
We probably want to follow the examples from https://github.com/mongodb/mongo-go-driver/blob/v1.17.4/mongo/client_examples_test.go#L250

plugins/destination/mongodb/client/client.go
```diff
@@ -17,26 +22,86 @@ type Client struct {
...
+		configFns := []func(*config.LoadOptions) error{}
+		if c.spec.AWSCredentials.Default {
+			// Use default AWS credentials
+			cfg, err = config.LoadDefaultConfig(ctx)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			configFns = append(configFns, config.WithDefaultRegion("us-east-1"))
+			if c.spec.AWSCredentials != nil && c.spec.AWSCredentials.LocalProfile != "" {
```
maaarcelino:
```suggestion
			if c.spec.AWSCredentials.LocalProfile != "" {
```
there's already a check for `c.spec.AWSCredentials != nil` in line 46

plugins/destination/mongodb/client/client.go
```diff
@@ -17,26 +22,86 @@ type Client struct {
...
+				configFns = append(configFns, config.WithSharedConfigProfile(c.spec.AWSCredentials.LocalProfile))
+			}
+
+			cfg, err = config.LoadDefaultConfig(ctx, configFns...)
+			if err != nil {
+				return nil, fmt.Errorf("unable to load AWS SDK config: %w", err)
+			}
+
+			if c.spec.AWSCredentials != nil && c.spec.AWSCredentials.RoleARN != "" {
```
maaarcelino:
```suggestion
			if c.spec.AWSCredentials.RoleARN != "" {
```

plugins/destination/mongodb/client/spec/spec.go
```diff
@@ -48,5 +51,9 @@ func (s *Spec) Validate() error {
 	if s.Database == "" {
 		return errors.New("database is required")
 	}
+	if s.AWSCredentials != nil && (s.AWSCredentials.RoleARN != "" || s.AWSCredentials.RoleSessionName != "" || s.AWSCredentials.ExternalID != "" || s.AWSCredentials.LocalProfile != "") && s.AWSCredentials.Default {
```
maaarcelino: are there no required fields for the case when `s.AWSCredentials != nil`?

plugins/destination/mongodb/docs/overview.md
```diff
@@ -51,3 +51,47 @@ This is the (nested) spec used by the MongoDB destination Plugin.
...
+
+- `role_arn` (`string`)
+
+  If specified will use this to assume role.
+
+- `role_session_name` (`string`)
+
+  If specified will use this session name when assume role to `role_arn`.
+
+- `external_id` (`string`)
```
maaarcelino: This ties in to my comment regarding required params for AWS IAM credentials. These docs don't specify any field as `(optional)`, so are they required?

bbernays:
> We probably want to follow the examples from https://github.com/mongodb/mongo-go-driver/blob/v1.17.4/mongo/client_examples_test.go#L250

I'm not sure I follow. All of the examples assume that you already have the credentials to pass to the client... The examples show all of the different ways that credentials can be injected into a compute resource not how to grab them, which is what the aws SDK is doing for us in our code.

erezrokah:
> All of the examples assume that you already have the credentials to pass to the client...

AWS Lambda and ECS Container or EC2 Instance examples use default credentials without requiring any known credentials, they are grabbed later https://github.com/mongodb/mongo-go-driver/blob/9d878b710ac8d254d51eaa19bca8c87e03ebe2b1/x/mongo/driver/auth/aws_conv.go#L149

plugins/destination/mongodb/client/client.go
```diff
@@ -17,26 +22,86 @@ type Client struct {
...
+		}
+		awsCreds, err := cfg.Credentials.Retrieve(ctx)
+		if err != nil {
+			return nil, err
+		}
+		assumeRoleCredential := options.Credential{
+			AuthMechanism: "MONGODB-AWS",
+			Username:      awsCreds.AccessKeyID,
+			Password:      awsCreds.SecretAccessKey,
+		}
```
przste-go: Should we allow to Specify authSource too ? As per docs:
`authSource=<authentication database>
`
MongoDB database that contains your user's authentication data. If you omit the authSource parameter and placeholder value, the driver uses the default value admin.

bbernays: Can you link to the docs?

przste-go: https://www.mongodb.com/docs/kafka-connector/current/security-and-authentication/mongodb-aws-auth/#connector-connection-settings

przste-go: Ah it's admin only for kafka
`// AuthSource: the name of the database to use for authentication. This defaults to "$external" for MONGODB-AWS,
// MONGODB-OIDC, MONGODB-X509, GSSAPI, and PLAIN. It defaults to  "admin" for all other auth mechanisms. This can
// also be set through the "authSource" URI option (e.g. "authSource=otherDb").`
my bad

plugins/destination/mongodb/client/client.go
```diff
@@ -5,38 +5,58 @@ import (
...
+				"RoleSessionName": c.spec.AWSCredentials.RoleSessionName,
+				"ExternalID":      c.spec.AWSCredentials.ExternalID,
+				"Default":         fmt.Sprintf("%t", c.spec.AWSCredentials.Default),
+			},
+		}
+		// According to the docs: if ApplyURI is called before SetAuth, the Credential from SetAuth will overwrite the values from the connection string
+		mongoDBClientOptions = mongoDBClientOptions.SetAuth(assumeRoleCredential)
+	}
+
+	c.client, err = mongo.Connect(context.Background(), mongoDBClientOptions)
```
murarustefaan: Why not pass the context from the function params here? 🤔

plugins/destination/mongodb/client/mongodb_aws_credentials.go
```diff
@@ -0,0 +1,318 @@
...
+		if err != nil {
+			return awsCred, fmt.Errorf("failed to parse Default value: %w", err)
+		}
+		awsCred.Default = boolValue
+	}
+	return awsCred, nil
+}
+
+func getAWSConfig(awsCred spec.Credentials) (*aws.Config, error) {
+	ctx := context.Background()
```
murarustefaan, line 84: Why? Wouldn't it make more sense to get a passed context here?

murarustefaan (commented):
Looks good overall but were you able to test this?

plugins/destination/mongodb/client/client.go
bbernays: Good catch.  Fixed in [71bf20c](https://github.com/cloudquery/cloudquery/pull/21021/commits/71bf20c476e820ea394e1e13ff42872486d9f1ee)

plugins/destination/mongodb/client/mongodb_aws_credentials.go
bbernays: We can't pass in an actual context because `NewAuthenticator` cannot take a context.

The pattern that the Mongo Driver exposes is that it creates different `Authenticator`s and then registers them as "available` and then at Connect time it decides which one to use.

murarustefaan (approved):

PR #21048, opened 2025-07-15
feat: Support the new OTEL metric format removing the client_id metric point granularity

murarustefaan:
Should go with https://github.com/cloudquery/plugin-sdk/pull/2185 in order to allow the CLI to read metrics without the client_id metric point.

[old CLI, old SDK](https://github.com/user-attachments/files/21236264/before.txt)
[old CLI, new SDK](https://github.com/user-attachments/files/21289151/after.txt)
[updated CLI, old SDK](https://github.com/user-attachments/files/21289612/after.txt)
[updated CLI, new SDK](https://github.com/user-attachments/files/21236263/after.txt)

The above showcase that Duration metrics are unavailable if using an old CLI version with newer plugin versions, but that's a downside we have to work with.

### DIFF
--- a/cli/internal/otel/receiver.go
+++ b/cli/internal/otel/receiver.go
@@ -133,10 +133,7 @@ func newMetricConsumer(metricsFile *os.File, quit chan any, wg *sync.WaitGroup) 
 
 	return func(ctx context.Context, metric pluginMetric) {
 		table := metric.Attributes["sync.table.name"].(string)
-		clientId, ok := metric.Attributes["sync.table.client_id"].(string)
-		if !ok {
-			clientId = ""
-		}
+		clientId, _ := metric.Attributes["sync.table.client_id"].(string)
 
 		tableLock.Lock()
 		defer tableLock.Unlock()


erezrokah (approved):

cli/internal/otel/receiver.go
```diff
@@ -131,7 +133,10 @@ func newMetricConsumer(metricsFile *os.File, quit chan any, wg *sync.WaitGroup)
 
 	return func(ctx context.Context, metric pluginMetric) {
 		table := metric.Attributes["sync.table.name"].(string)
-		clientId := metric.Attributes["sync.client.id"].(string)
+		clientId, ok := metric.Attributes["sync.table.client_id"].(string)
```
disq:
```suggestion
		clientId, _ := metric.Attributes["sync.table.client_id"].(string)
```
You don't need the `if !ok` block below.

murarustefaan: Done in [`4381ce7` (#21048)](https://github.com/cloudquery/cloudquery/pull/21048/commits/4381ce7056979b47ecf3e1bd7cdfe9ea78c7f058)

PR #21052, opened 2025-07-16
feat: Add post login steps in the CLI

blesniewski:
Adds post login step guide to the CLI

### DIFF
--- a/cli/cmd/login.go
+++ b/cli/cmd/login.go
@@ -200,6 +200,14 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 
 	analytics.TrackLoginSuccess(ctx, invocationUUID.UUID)
 	cmd.Println("CLI successfully authenticated.")
+	cmd.Println(`Logged in. Next, initialize your sync configuration:
+	cloudquery init --source=<source> --destination=<destination>
+	
+Example:
+	cloudquery init --source=aws --destination=postgresql
+
+Then run your first sync:
+	cloudquery sync`)
 
 	return nil
 }
--- a/cli/cmd/login.go
+++ b/cli/cmd/login.go
@@ -204,7 +204,7 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 	cloudquery init --source=<source> --destination=<destination>
 	
 Example:
-	cloudquery init --source=aws --destination=postgresql
+	cloudquery init --source=aws --destination=sqlite
 
 Then run your first sync:
 	cloudquery sync`)
--- a/cli/cmd/login.go
+++ b/cli/cmd/login.go
@@ -200,7 +200,7 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 
 	analytics.TrackLoginSuccess(ctx, invocationUUID.UUID)
 	cmd.Println("CLI successfully authenticated.")
-	cmd.Println(`Logged in. Next, initialize your sync configuration:
+	cmd.Println(`Next, initialize your sync configuration:
 	cloudquery init --source=<source> --destination=<destination>
 	
 Example:


blesniewski:
```
❯ go run main.go login                         
Opened browser at https://accounts.cloudquery.io?returnTo=http://localhost:57535/callback. Waiting for authentication to complete.
Authenticating...
Your current team is set to cloudquery.
CLI successfully authenticated.
Logged in. Next, initialize your sync configuration:
        cloudquery init --source=<source> --destination=<destination>

Example:
        cloudquery init --source=aws --destination=postgresql

Then run your first sync:
        cloudquery sync

File: cli/cmd/login.go [hunk 1]
```diff
@@ -200,6 +200,14 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
...
 	analytics.TrackLoginSuccess(ctx, invocationUUID.UUID)
 	cmd.Println("CLI successfully authenticated.")
+	cmd.Println(`Logged in. Next, initialize your sync configuration:
+	cloudquery init --source=<source> --destination=<destination>
+	
+Example:
+	cloudquery init --source=aws --destination=postgresql
+
+Then run your first sync:
+	cloudquery sync`)
```
erezrokah, line 210: I think the example will always fail since you need to set a connection string for PostgresSQL via `${POSTGRESQL_CONNECTION_STRING}`, see result of `cloudquery init --source=aws --destination=postgresql` below:

```yaml
kind: source
spec:
  # Source spec section
  name: aws
  path: cloudquery/aws
  registry: cloudquery
  version: "v32.31.0"
  tables: ["aws_ec2_instances"]
  destinations: ["postgresql"]
  # Learn more about the configuration options at https://cql.ink/aws_source
  spec:
    # Optional parameters
    # regions: []
    # accounts: []
    # org: nil
    # concurrency: 50000
    # initialization_concurrency: 4
    # aws_debug: false
    # max_retries: 10
    # max_backoff: 30
    # custom_endpoint_url: ""
    # custom_endpoint_hostname_immutable: nil # required when custom_endpoint_url is set
    # custom_endpoint_partition_id: "" # required when custom_endpoint_url is set
    # custom_endpoint_signing_region: "" # required when custom_endpoint_url is set
    # use_paid_apis: false
    # table_options: nil
    # scheduler: shuffle # options are: dfs, round-robin, shuffle, or shuffle-queue
    # use_nested_table_rate_limiting: false 
    # enable_api_level_tracing: false 
---
kind: destination
spec:
  name: "postgresql"
  path: "cloudquery/postgresql"
  registry: "cloudquery"
  version: "v8.8.11"
  # Learn more about the configuration options at https://cql.ink/postgresql_destination
  spec:
    # set the environment variable in DSN format like "user=postgres password=pass+0-[word host=localhost port=5432 dbname=postgres sslmode=disable"
    # you can also specify it in URI format like "postgres://postgres:pass@localhost:5432/postgres?sslmode=disable". any special URI characters need to be percent-encoded
    connection_string: "${POSTGRESQL_CONNECTION_STRING}"
    # Optional parameters:
    # pgx_log_level: error
    # batch_size: 10000 # 10K entries
    # batch_size_bytes: 100000000 # 100 MB
    # batch_timeout: 60s

    # create_performance_indexes: false #create indexes that help with performance when using `write_mode: overwrite-delete-stale`
```

erezrokah: Maybe chose SQLite as that will always work assuming AWS is configured with default credentials

blesniewski:
> Shouldn't this be a feat:? Looks good otherwise

Doing a minor version bump just for a `cmd.Println` seemed excessive, I can change to feat if you prefer it that way

erezrokah:
> Doing a minor version bump just for a `cmd.Println` seemed excessive, I can change to feat if you prefer it that way

It's a new feature so that's what's expected when following SemVer. For `chore` we won't have a release at all

cli/cmd/login.go
```diff
@@ -200,6 +200,14 @@ func runLogin(ctx context.Context, cmd *cobra.Command) (err error) {
 
 	analytics.TrackLoginSuccess(ctx, invocationUUID.UUID)
 	cmd.Println("CLI successfully authenticated.")
+	cmd.Println(`Logged in. Next, initialize your sync configuration:
```
disq: I would prefer it this way:
> CLI successfully authenticated.
> Next, initialize your sync configuration:

"Logged in" seems extra, previous line ("CLI successfully authenticated.") already tells that.

PR #21058, opened 2025-07-16
fix: Don't re-use scan destination between rows scan

erezrokah:
#### Summary

I believe this should fix the flaky ClickHouse destination tests we've been seeing for a long while.

We were re-using the scan destination between scans which caused under some conditions to keep the previous read value instead of overwriting (I saw this happen when we first read a non null value, then supposed to read a null value, but I guess ClickHouse has some optimization so skip reading the null value)


savme (approved):

przste-go:
Nice find 👍

PR #21062, opened 2025-07-18
chore: Replace all golang-cross images with maintained version

przste-go:
This PR replaces cq golang-cross image with maintained version.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -17,6 +17,18 @@
       depNameTemplate: 'goreleaser/goreleaser-pro',
       datasourceTemplate: 'github-releases',
     },
+    {
+      customType: "regex",
+      fileMatch: [
+        ".github/workflows/.*\\.ya?ml$"
+      ],
+      matchStrings: [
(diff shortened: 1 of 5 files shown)


murarustefaan (approved):

savme (approved):

przste-go:
I'm starting to think we can actually get rid of all of those images like in here: https://github.com/cloudquery/cloudquery-private/pull/8321/files. WDYT @murarustefaan @savme ? It seems that for source plugins we are only using them for fips which shouldn't be needed

erezrokah:
> I'm starting to think we can actually get rid of all of those images like in here: https://github.com/cloudquery/cloudquery-private/pull/8321/files. WDYT @murarustefaan @savme ? It seems that for source plugins we are only using them for fips which shouldn't be needed

If I recall correctly we needed the images in the validate FIPS workflows since the old way of enabling FIPS required `CGO_ENABLED=1`. Since we don't need `CGO_ENABLED=1` anymore for FIPS plugins we can probably get rid of the images

PR #21063, opened 2025-07-18
chore: Remove golang-cross

przste-go:
#### Summary

Golang-cross is not needed for test-plugin. It should be only used for plugins that require cgo.


maaarcelino:
@przste-go "Most likely" does not fill me with confidence.. 😬 did you test this hypothesis?

erezrokah:
Looks like the test plugin is still using BoringCrypto so we'll need to change it first https://github.com/cloudquery/cloudquery/blob/85d4001576d3dc6daa2dc99e82d74cb321556676/plugins/source/test/main_fips.go

przste-go:
> Looks like the test plugin is still using BoringCrypto so we'll need to change it first https://github.com/cloudquery/cloudquery/blob/85d4001576d3dc6daa2dc99e82d74cb321556676/plugins/source/test/main_fips.go

Yes, just pushed changes removing boringcrypto

przste-go:
> I think we still need it for the `CGO_ENABLED=1` plugins no? I mean DuckDB, Snowflake and SQLite

Yeah you are right @erezrokah those are setting CGO: true explicitly

przste-go:
Ok, now I've left the changes only for source test plugin, all destination plugins left are using CGO

PR #21101, opened 2025-07-22
feat: Add support for table-level TTLs in ClickHouse destination

hermanschaaf:
This adds support for table-level TTLs in the ClickHouse destination. 

Similar to partitions and ORDER BY clauses, a TTL section can be added to the config like this, so every table's TTL can be specified individually, if desired:

```
ttl:
- tables: ["*"]
  skip_tables: ["no_ttl_table"]
  ttl: "INTERVAL 60 DAY"
```

Some aspects of this were harder to do than anticipated. For example, ClickHouse transforms the TTL expression to a different format when it is stored in the database, so to avoid unnecessary DDLs we do a comparison of normalized expressions inside ClickHouse to evaluate if two TTL expressions are, in fact, equal.


plugins/destination/clickhouse/queries/table.go
```diff
@@ -56,3 +57,19 @@ func tableNamePart(table, cluster string) string {
...
+	return fmt.Sprintf(`SHOW CREATE TABLE "%s"."%s"`, database, table)
+}
+
+func EqualTTLsQuery(table *schema.Table, ttl1, ttl2 string) string {
+	// ClickHouse allows different syntaxes for the same TTL expression,
+	// so we use a query to compare two given TTLs.
+	// However, to evaluate the expression, we replace referenced columns with a fixed date.
+	for i, col := range table.Columns {
+		// Replace column names with a fixed date to ensure the comparison works
+		ttl1 = strings.ReplaceAll(ttl1, col.Name, fmt.Sprintf("makeDate(1970, 1, 1) + INTERVAL %d SECOND", i))
```
dcelasun, line 71: nit: Since it's just a string comparison we don't really need `makeDate`, `now64()` would work as well.

dcelasun (approved):

plugins/destination/clickhouse/queries/table.go
hermanschaaf: It's not actually a string comparison, we need the expressions to be evaluated to compare them. I think `now64()` should _probably_ work but I'd rather play it safe by using a constant value

dcelasun: Yeah, string comparison wasn't the right word. As long as it's a constant expression it should work, and `now64()` is stable across invocations in a single query. But yes, `makeDate` is more explicitly constant.

PR #21102, opened 2025-07-23
feat: Add DeleteRecord handling to SQLite destination

jeromewir:
#### Summary

Following up on https://community.cloudquery.io/t/incremental-tables-how-to-handle-deletion/1205/6

This PR adds support for `DeleteRecord` messages so sources can specify which items can be deleted specifically.
I used a lot of the existing code present in the ClickHouse PR: https://github.com/cloudquery/cloudquery/pull/20772/files


plugins/destination/sqlite/client/delete.go
```diff
@@ -0,0 +1,92 @@
...
+	sb.WriteString(" WHERE ")
+	if len(msg.WhereClause) == 0 {
+		sb.WriteString("1")
+	} else {
+		for i, predicateGroup := range msg.WhereClause {
+			if len(predicateGroup.Predicates) == 0 {
+				continue
+			}
+			sb.WriteString("(")
+			for i, predicate := range predicateGroup.Predicates {
```
hermanschaaf: This inner loop is re-using `i` as an index variable which should work, I think, due to how variables can be shadowed in Go, but I would say is not recommended, mostly for readability reasons.

hermanschaaf (commented):
Thanks for the pull request @jeromewir! We'll review this a bit more closely, I just left one comment for now after a quick read through. Also, I changed the title from `feat!` to `feat` because I think this isn't a breaking change, right?

plugins/destination/sqlite/client/delete.go
jeromewir: Fixed!

jeromewir:
> Also, I changed the title from feat! to feat because I think this isn't a breaking change, right?

Oh, I didn't realize the `!` was for breaking change, sorry about that.
That's correct, the code is backward compatible

I just added a missing folder `typeconv`, which is mostly the same as the linked PR in the description with a slight change: I'm returning native objects instead of pointers.

PR #21107, opened 2025-07-24
feat: Add per table stats to sync summaries

erezrokah:
#### Summary

Adds per table stats for the sync summaries table (we already have total `resources` and `source_errors`)


cli/cmd/summary.go
```diff
@@ -18,24 +18,26 @@ import (
...
+	SourcePath          string            `json:"source_path"`
+	SourceVersion       string            `json:"source_version"`
+	SourceWarnings      uint64            `json:"source_warnings"`
+	SourceTables        []string          `json:"source_tables"`
+	SyncID              string            `json:"sync_id"`
+	SyncTime            time.Time         `json:"sync_time"`
+	SyncGroupID         *string           `json:"sync_group_id,omitempty"`
+	ShardNum            *int              `json:"shard_num,omitempty"`
+	ShardTotal          *int              `json:"shard_total,omitempty"`
+	ResourcesPerTable   map[string]uint64 `json:"resources_per_table,omitempty"`
```
erezrokah, line 39: Went with 2 separate columns as it's aligned with `resources`, `source_errors` and  `destination_errors`.
Also a map is easier to query on some databases (if I used a single column it will be a JSON column in most destinations)

cli/cmd/summary.go
```diff
@@ -18,24 +18,26 @@ import (
...
+	ErrorsPerTable      map[string]uint64 `json:"errors_per_table,omitempty"`
```
erezrokah, line 40: We can also do `source_errors_per_table` but I think it's clear these errors are from the source

stoovon (approved):

PR #21108, opened 2025-07-24
fix: Concurrent map access when reporting remote progress

erezrokah:
#### Summary

Our race detection tests didn't catch this since we only report remote progress for sync runs API keys https://github.com/cloudquery/cloudquery/blob/0e0deec32b95fac6af9f27e53e8fd1bb75921da2/cli/cmd/sync_v3.go#L89

I'll follow up with a test later


disq (approved):

cli/cmd/sync_v3.go
```diff
@@ -390,12 +392,13 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
 			if atomic.LoadInt64(&isComplete) == 1 {
 				status = cloudquery_api.SyncRunStatusCompleted
 			}
+			tableProgress := cloudquery_api.SyncRunTableProgress(statsPerTable.GetAll())
 			obj := cloudquery_api.CreateSyncRunProgressJSONRequestBody{
 				Rows:          atomic.LoadInt64(&totalResources),
 				Errors:        int64(totals.Errors),
 				Warnings:      int64(totals.Warnings),
 				Status:        &status,
-				TableProgress: &statsPerTable,
```
erezrokah, line 398: The progress reporting runs in a separate Go routine than the sync one, hence the race

PR #21165, opened 2025-07-29
feat: Better summary of schema changes

erezrokah:
#### Summary

Follow up to https://github.com/cloudquery/plugin-sdk/pull/2240

Example error

```
Loading spec(s) from examples/gcp_to_postgres.yml
Starting migration for: gcp (cloudquery/gcp@v18.20.1) -> [postgresql (grpc@localhost:8888)]
Error: failed to migrate v3 source gcp: rpc error: code = Internal desc = write failed: 
Can't migrate tables automatically, migrate manually or consider using 'migrate_mode: forced'. Non auto migratable tables changes:

gcp_artifactregistry_docker_images:
  - Type changed from "list<item: utf8, nullable>" to "json" for column "tags"

gcp_compute_routes:
  - Column "next_hop_inter_region_cost" added with type "int64"
  - Column "next_hop_med" added with type "int64"
  - Column "next_hop_origin" added with type "utf8"
  - Type changed from "list<item: utf8, nullable>" to "json" for column "tags"
```

PR #21171, opened 2025-07-29
feat: Better summary of schema changes

erezrokah:
#### Summary

Similar to https://github.com/cloudquery/cloudquery/pull/21165

### DIFF
--- a/plugins/destination/sqlite/client/migrate.go
+++ b/plugins/destination/sqlite/client/migrate.go
@@ -82,21 +82,19 @@ func (c *Client) normalizeField(field arrow.Field) *arrow.Field {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


plugins/destination/sqlite/client/migrate.go
```diff
@@ -132,51 +130,53 @@ func (*Client) canAutoMigrate(changes []schema.TableColumnChange) bool {
 
 // This is the responsibility of the CLI of the client to lock before running migration
 func (c *Client) MigrateTables(ctx context.Context, msgs message.WriteMigrateTables) error {
+	allTables := schema.Tables{}
```
erezrokah, line 133: This seems like a lot of changes but it only changes the logic that we first initialize all tables from all messages instead of going over each message one by one

disq (approved):

PR #21180, opened 2025-07-30
fix: Don't call Tables gRPC method at the beginning of the sync

erezrokah:
#### Summary

Since this PR https://github.com/cloudquery/cloudquery/pull/20909/files the S3 source plugin fails to sync and result in the following error `failed to get tables: tables only discovered during sync`.

The reason is that the S3 source dynamically discovers all files/tables during the sync, and errors out if you call the tables endpoint before the sync is done, see https://github.com/cloudquery/cloudquery-private/blob/334760529da4815dc83659c7f529154682b2c8a0/plugins/source/s3/resources/plugin/client.go#L91

Instead of calling the tables endpoint we can keep track of the tables during the sync. This works even if you pass `--no-migrate` since we skip those a bit after the new code https://github.com/cloudquery/cloudquery/blob/7b5980236005db31be536f4ffd2c390752f12309/cli/cmd/sync_v3.go#L600


cli/cmd/sync_v3.go
```diff
@@ -356,23 +356,6 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
...
-	sourceTables, err := getTables(ctx, sourcePbClient, &plugin.GetTables_Request{
-		Tables:              sourceSpec.Tables,
-		SkipTables:          sourceSpec.SkipTables,
-		SkipDependentTables: *sourceSpec.SkipDependentTables})
-
-	if err != nil {
-		return err
-	}
-	// Pre init stats per table
-	for _, table := range sourceTables {
```
erezrokah, line 368: We don't really need the init code as Go Zero value gives us the same functionality

dcelasun (approved):

murarustefaan (approved):

stoovon (approved):

PR #21228, opened 2025-08-14
fix: Test PostgreSQL pool connections before use

dcelasun:
If a PostgreSQL server becomes unreachable after a sync has started, connections from the pgx pool can hang indefinitely. As per upstream, the only way to know for sure if a connection is still valid is to try using it.

This commit adds a BeforeAcquire hook to ping the database before any query. If it fails, the connection is dropped from the pool.


hermanschaaf:
@dcelasun Were you able to reproduce the issue and verify that this change fixes it?

dcelasun:
> @dcelasun Were you able to reproduce the issue and verify that this change fixes it?

I tried by shutting down the server mid run, but I saw errors - it didn't just hang as was reported. I don't (yet) know the exact error message the reporter saw. I'll try another repro once I receive it.

erezrokah:
Let's close this until we have a reproduction then? Happy to re-open later

PR #21264, opened 2025-08-19
feat: Add a timeout to the migrate context

stoovon:
Previously, the acquire would hang indefinitely until a connection became available.

There's a default max conn idle time of 30 minutes, and default max conns is 4 (defined in pgxpool/pool.go).

Finally, if connectTimeOut is <= 0, connections will terminate after 2 minutes.

All of these behaviours can be mitigated:
- We could specify a lower number of max conns (4 / 2 = 2 for sync, and 2 for kinesis), such that we never have too many conns and deadlocks.
- We could recommend the customer specifies 8 conns in postgres.

- We can introduce a timeout (this PR) such that the acquire fails loudly. That gives us a very clear signal, which will help us to make the above changes.

- We can also set a global timeout but that's a deeper change in my view.

### DIFF
--- a/plugins/destination/postgresql/client/migrate.go
+++ b/plugins/destination/postgresql/client/migrate.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/cloudquery/plugin-sdk/v4/message"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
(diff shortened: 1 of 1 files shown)


plugins/destination/postgresql/client/migrate.go
```diff
@@ -17,11 +18,14 @@ import (
 
 // MigrateTableBatch migrates a table. It forms part of the writer.MixedBatchWriter interface.
 func (c *Client) MigrateTableBatch(ctx context.Context, messages message.WriteMigrateTables) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
hermanschaaf, line 21: I think this is risky - if I'm understanding correctly, this will mean that all the migrations in the batch _must_ complete within 30 seconds, otherwise the migration will fail. I don't think we want to impose such a restriction necessarily. At the very least, we should give this 30 second timeout to every individual table migration.

But apart from this, what will this change achieve? I could be wrong, but I think the deadlock error raised in Postgres also causes this code to fail already. The real question for me is how a deadlock situation arises inside Postgres

stoovon: I agree, we'd need to choose a value that makes sense but it would be arbitrary and would collar migrations. Probably too crude.

stoovon: I wasn't currently able to find evidence of a deadlock error in Postgres from the Slack thread but if we do have an error that's very helpful.

My understanding at the present time is that we don't have a deadlock error, but we do appear to have a gap in our connection pooling strategy (it works perfectly for EventBased _or_ conditional sync, but EventBased _and_ conditional sync).

PR #21268, opened 2025-08-22
feat: Retry on Postgres deadlock

stoovon:
#### Summary

- [We do not recommend customers configure overlapping syncs](https://docs.cloudquery.io/docs/advanced-topics/running-cloudquery-in-parallel#no-overlapping-syncs).
  - Overlapping syncs can cause deadlocks in the Postgres destination.
  - Avoiding overlapping syncs is the primary recommendation to prevent deadlocks.
- As a secondary safeguard, this PR introduces retries to handle most Postgres deadlock scenarios (other destinations are out of scope).
  - Retries are attempted up to 5 times, with random jitter between attempts.
    - Deadlocks are rare, even with multiple concurrent syncs. Five retries balances resilience and respect for Postgres backpressure.
  - Users can enable this behaviour via the `retry_on_deadlock` option.


plugins/destination/postgresql/client/migrate.go
```diff
@@ -139,6 +139,8 @@ func (c *Client) autoMigrateTable(ctx context.Context, table *schema.Table, chan
 			if err != nil {
 				return err
 			}
+		default:
```
stoovon, line 142: These changes were required by `golangci-lint v2` `enforce-switch-style`

erezrokah: Can we disable this rule? See https://github.com/cloudquery/plugin-sdk/blob/b8cf3902daef8d1bfbde5d7c7908c70f8085c0a6/.golangci.yml#L74

stoovon: Rule disabled in 9ab172b802 but I think we should still keep the default clauses.

stoovon:
> test to simulate many concurrent syncs

I borrowed and tuned the test but I was not quite able to repro the issue. Nevertheless, the test exercises the correct code path so I think it still has value in terms of showing how to set up a spec with `RetryOnDeadlock`

plugins/destination/postgresql/client/spec/spec.go
```diff
@@ -39,6 +39,8 @@ type Spec struct {
 
 	// Option to create specific indexes to improve deletion performance
 	CreatePerformanceIndexes bool `json:"create_performance_indexes,omitempty" jsonschema:"default=false"`
+
+	RetryOnDeadlock bool `json:"retry_on_deadlock,omitempty" jsonschema:"default=false"`
```
erezrokah: Can we make this the number of times to retry?

plugins/destination/postgresql/client/insert.go
```diff
@@ -87,21 +89,31 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
...
-		return nil
-	}
+	const maxRetries = 5
+	var attempt int
 
-	var pgErr *pgconn.PgError
-	if errors.As(err, &pgErr) {
-		return fmt.Errorf("failed to execute batch with pgerror: %s: %w", pgErrToStr(pgErr), err)
-	}
+	for {
```
erezrokah: Non blocking but in other places we use a library, see example https://github.com/cloudquery/plugin-pb-go/blob/175e2424ebf2f82cfd43339614dad225ee34f7f6/managedplugin/plugin.go#L340

plugins/destination/postgresql/client/spec/spec.go
stoovon: Refactored in fbc2339.

plugins/destination/postgresql/client/spec/spec.go
```diff
@@ -39,6 +39,9 @@ type Spec struct {
 
 	// Option to create specific indexes to improve deletion performance
 	CreatePerformanceIndexes bool `json:"create_performance_indexes,omitempty" jsonschema:"default=false"`
+
+	// Number of times to retry a transaction if a deadlock is detected by Postgres.
+	RetryOnDeadlock int64 `json:"retry_on_deadlock,omitempty" jsonschema:"default=0"`
```
erezrokah, line 44: We should also add this to the docs (under the docs directory)

plugins/destination/postgresql/client/insert.go
stoovon: Refactored in 28f6ada5af.

plugins/destination/postgresql/client/spec/spec.go
stoovon: Docs added in 2b03787.

plugins/destination/postgresql/client/insert.go
```diff
@@ -87,21 +90,41 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
...
+		if err != nil {
+			return err
+		}
+
 		return nil
-	}
+	}, retry.RetryIf(func(err error) bool {
+		var pgErr *pgconn.PgError
+		if errors.As(err, &pgErr) {
+			if pgErr.Code == "40P01" && attempt < c.spec.RetryOnDeadlock {
```
erezrokah: Don't think we need to track attempts and add jitter, the library has a default configuration https://github.com/avast/retry-go/blob/cbf4dbfb37a5861d7304d50e5e87b6ae72569dde/retry.go#L225

We can override it if needed

erezrokah: We might want to set `lastErrorOnly` to `true`

stoovon: Simplified in d21258e1e2.

stoovon: `lastErrorOnly` in a1d68b4d36

PR #21277, opened 2025-08-25
fix!: Update migration strategy

disq:
Backported some of the migration logic from Postgresql destination, with some additional changes
Faster schema state detection (3 SQLs most of the time, more if schema has more than 200 tables)

Should fix https://linear.app/cloudquery/issue/ENG-1884

Recommended review strategy is reading each commit separately.


plugins/destination/snowflake/client/migrate.go
```diff
@@ -66,175 +73,283 @@ func (c *Client) MigrateTables(ctx context.Context, msgs message.WriteMigrateTab
...
+func (c *Client) autoMigrateTable(ctx context.Context, table, oldTable *schema.Table, changes []schema.TableColumnChange) error {
+	tableName := table.Name
+	for _, change := range changes {
+		var err error
+
+		switch {
+		case change.Type == schema.TableColumnChangeTypeMoveToCQOnly:
+			// Drop PKs automatically (and add PK constraint to _cq_id in next step)
+			if len(oldTable.PrimaryKeys()) > 0 {
+				err = c.alterTableDropPK(ctx, tableName)
```
disq: We need to do this here because the order of changes was in reverse in my debugging:
changes[0] is `schema.TableColumnChangeTypeMoveToCQOnly`
changes[1] is `schema.TableColumnChangeTypeUpdate` (add PK)
changes[2] is `schema.TableColumnChangeTypeUpdate` (remove current PK)

Obviously adding another PK doesn't work unless you remove the current one.

disq: Fixed in https://github.com/cloudquery/cloudquery/pull/21277/commits/b285fe3b14417629d26b817bc3311062ea42d2d1 where we don't emit/handle `TableColumnChangeTypeMoveToCQOnly` at all, just do the PK stuff.

erezrokah (approved):

PR #21278, opened 2025-08-26
feat: Implement support for PgVector.

marianogappa:
Implements support for PgVector within the Postgres destination.

This new spec subobject was added:

```

- `pgvector_config` (`object`) (optional)

  Optional configuration to enable PgVector embedding support.

  Note: source plugin must sync the `_cq_id` column on target tables if this is enabled.

  - `tables` (`array`) (required)
    
    Tables to create embeddings for. A table with the `_embeddings` suffix will be created for each table in this list.

    - `table_name` (`string`) (required)

    - `embed_columns` (`array`) (required)

    Columns to create embeddings for.

    - `metadata_columns` (`array`) (required)

    These columns will be added as-is from the source table for context. The `_cq_id` column will be added automatically and an index will be created on it.

  - `text_splitter` (`object`) (optional)

    Optional text splitting configuration for the embeddings. If unset, defaults are used.

    - `recursive_text` (`object`) (required)

      - `chunk_size` (`integer`) (required)

      Chunk size for the text splitting.

      - `chunk_overlap` (`integer`) (required)

      Chunk overlap for the text splitting.

  - `embedding` (`object`) (required)

    Embedding API provider configuration. Currently only OpenAI is supported.

    - `dimensions` (`integer`) (required)
    
    The number of dimensions to use for the embeddings. For `text-embedding-3-small`, this is 1536. For `text-embedding-3-large`, this is 3072.

    - `api_key` (`string`) (required)

    The OpenAI API key to use for the embedding API.

    - `model_name` (`string`) (required)

    The model name to use for the embedding API. Currently, `text-embedding-3-small` and `text-embedding-3-large` are supported.
```

This new subobject is optional. If unset, everything works the same way as it used to.

If set, a new table with an `_embeddings` suffix is added for each configured table. The new table contains at least `_cq_id`, an `embedding` vector and a `chunk` column to describe which part of the content the embedding was made for. Extra metadata columns from the source table can be passed through as is.

It supports the recursive-text TextSplitter algorithm.

`_cq_id` is required on the source table (otherwise, an error is thrown). An index is added onto it, but the PK is removed from it because the `_embeddings` table has `n` rows per row on the source table (because embeddings are calculated for potentially a subset of the configured content). This is also why a different table is used, and why `_cq_id` is required.

Because rows are upserted via `_cq_id` on successive syncs, and there can be many instances of a given `_cq_id` whose cardinality can change between syncs, instead a `DELETE` and `INSERT` strategy is used.

The code always runs a `CREATE TABLE IF NOT EXISTS` for the `_embeddings` table (unlike regular tables which respond to `MigrateMessage` on GRPC), since it's possible that the source table already exists and thus the migrate messages are ignored for it.


erezrokah (commented):
We had issues before when we reach the max table name length for Postgres, so we should be careful when suffixing table names. Maybe we can put the embeddings in another DB schema with the same table names?

marianogappa:
@stoovon I'm pretty sure this case is covered. On a subsequent sync, all `_cq_ids` synced are deleted from the embeddings table before insertion (there's a note about why I do this in the PR description).

marianogappa:
>We had issues before when we reach the max table name length for Postgres, so we should be careful when suffixing table names. Maybe we can put the embeddings in another DB schema with the same table names?

@erezrokah this is a good catch but I think the alternatives need a little more thought. I'm not thrilled with using a different DB schema. Definitely cannot insert on the same table.

erezrokah:
> @erezrokah this is a good catch but I think the alternatives need a little more thought. I'm not thrilled with using a different DB schema. Definitely cannot insert on the same table.

If the schemas are the same for the embeddings maybe put all of them in the same table with an additional PK as the table name?

marianogappa:
>If the schemas are the same for the embeddings maybe put all of them in the same table with an additional PK as the table name?
I didn't quite understand this one 🤔 But this morning I was thinking why not just require a per-table-config embeddings table name on the spec and that's that, user will set it?

erezrokah:
> per-table-config embeddings table name on the spec and that's that, user will set it?

A config for the table names works too

PR #21304, opened 2025-09-01
fix: Auth docs cleanup for the most popular plugins

pilvikala:
Related to https://github.com/cloudquery/cloudquery-private/pull/8801.
Cleaning up Auth documentation for the most popular plugins to make sure it gets displayed properly in Getting Started.


plugins/destination/file/docs/_authentication.md
```diff
@@ -0,0 +1,5 @@
+---
```
erezrokah: `_authentication.md` and `_configuration.md` are hidden in the sidebar so you don't need to set a title for them.

Regardless we default to title casing the filename which will work in this case.

If you want this to show up in the sidebar you have to rename it to `authentication.md`.

Otherwise you can include this information via a placeholder similar to https://github.com/cloudquery/cloudquery/blob/c62490154a86d77dffdaa9f5e13b23b8f8aa5b96/plugins/source/square/docs/overview.md?plain=1#L13

plugins/destination/postgresql/docs/_authentication.md
```diff
@@ -0,0 +1,12 @@
+---
+hub-title: Authentication
+---
+
+## Authentication
```
erezrokah: Similar here, this won't show up anywhere unless you rename it or use the placeholder `:authentication`

plugins/destination/sqlite/docs/types.md
```diff
@@ -1,3 +1,8 @@
+---
+hub-title: Supported Types
```
erezrokah: If we change the title here, we should change it to all plugins that have a `types.md` file

plugins/destination/sqlite/docs/spec.md
```diff
@@ -0,0 +1,10 @@
+---
+hub-title: SQLite Spec
```
erezrokah: Aren't we already under the SQLite plugin? What's the reason we need to have `SQLite Spec` and not only `Spec`?

plugins/destination/sqlite/docs/overview.md
```diff
@@ -3,21 +3,12 @@ name: SQLite
 stage: GA
 title: SQLite Destination Plugin
 description: CloudQuery SQLite destination plugin documentation
+hub-title: SQLite Destination Plugin
```
erezrokah: What's wrong with `Overview` in the sidebar? `SQLite Destination Plugin` will be quite long I think

image

plugins/destination/sqlite/docs/_authentication.md
```diff
@@ -0,0 +1,8 @@
+---
+hub-title: Authentication
```
erezrokah: Similar comment here, this content won't show up in the sidebar

plugins/destination/sqlite/docs/_configuration.md
```diff
@@ -1,3 +1,8 @@
+---
+hub-title: Configuration
```
erezrokah: Similar comment here, this content won't show up in the sidebar

erezrokah (commented):
Agree that our plugins docs need a revamp but we should decide on the format to make sure they are all consistent, maybe the following sidebar structure

## Source plugins

```
Overview
Authentication
Configuration
Spec
...extra
```

## Destination plugins

```
Overview
Authentication
Configuration
Spec
Types
...extra
```

pilvikala:
I have removed most of the titles now and renamed one file.

plugins/destination/sqlite/docs/_authentication.md
```diff
@@ -0,0 +1,7 @@
+---
+hub-order: 2
+---
+
+## Authentication
```
erezrokah:
```suggestion
```

We already add the title when rendering the docs

plugins/destination/s3/docs/_authentication.md
erezrokah, line 1: You should rename this to `authentication.md` if you want it to appear in the sidebar

plugins/destination/postgresql/docs/authentication.md
```diff
@@ -0,0 +1,12 @@
+---
+hub-title: Authentication
+---
+
```
erezrokah:
```suggestion
```

`Authentication` is already the default based on the filename

plugins/destination/postgresql/docs/authentication.md
```diff
@@ -0,0 +1,12 @@
...
+## Authentication
```
erezrokah:
```suggestion
```

The header is redundant

plugins/destination/postgresql/docs/authentication.md
erezrokah: This comment is still unaddressed, we can remove `hub-title` as `Authentication` is the default

plugins/destination/s3/docs/_authentication.md
erezrokah, line 1: This file should be renamed to `authentication.md` if you want it to appear in the sidebar navigation.
Also if you do rename it to `authentication.md` you should remove these lines from the overview otherwise the content will get duplicated https://github.com/cloudquery/cloudquery/blob/main/plugins/destination/s3/docs/overview.md?plain=1#L214-L216

plugins/destination/s3/docs/_authentication.md
```diff
@@ -1,85 +1,37 @@
-The plugin needs to be authenticated with your account(s) in order to sync information from your cloud setup.
+---
+hub-title: Authentication
```
erezrokah:
```suggestion
```

`Authentication` is already the default title for a file named `authentication.md`

plugins/destination/s3/docs/authentication_advanced.md
```diff
@@ -0,0 +1,79 @@
+---
+hub-title: Advanced Authentication Configuration
```
erezrokah:
```suggestion
hub-title: Auth Details
```

I believe this is what we used in a another place to keep the sidebar navigation title shorter

plugins/destination/sqlite/docs/_configuration.md
erezrokah, line 1: This file should be renamed to `configuration.md`

plugins/destination/sqlite/docs/_authentication.md
erezrokah, line 1: This file should be renamed to `authentication.md`

plugins/destination/sqlite/docs/overview.md
```diff
@@ -3,21 +3,11 @@ name: SQLite
 stage: GA
 title: SQLite Destination Plugin
 description: CloudQuery SQLite destination plugin documentation
+hub-order: 1
 ---
+
 # SQLite Destination Plugin
```
erezrokah:
```suggestion
```

We can remove this I think as now we have duplicated headers 
image

plugins/destination/sqlite/docs/overview.md
```diff
@@ -3,21 +3,11 @@ name: SQLite
 stage: GA
 title: SQLite Destination Plugin
 description: CloudQuery SQLite destination plugin documentation
```
erezrokah:
```suggestion
```

Don't think we use these anymore

erezrokah (commented):
Added a few more comments

pilvikala:
@erezrokah I think it should be good now.

plugins/destination/s3/docs/overview.md
```diff
@@ -4,6 +4,7 @@ stage: GA
 title: S3 Destination Plugin
 description: CloudQuery S3 destination plugin documentation
 ---
+
 # S3 Destination Plugin
```
erezrokah:
```suggestion
```

plugins/destination/s3/docs/overview.md
```diff
@@ -4,7 +4,7 @@ stage: GA
 title: S3 Destination Plugin
 description: CloudQuery S3 destination plugin documentation
 ---
-# S3 Destination Plugin
+
```
erezrokah, line 7:
```suggestion
```

PR #21335, opened 2025-09-03
feat: Add `durations_per_table_ms` to `cloudquery_sync_summaries`

disq:
Might not be the best way to do it, ~but I don't think there's a way to know whether processing a table has finished.~ OTEL stats work per client ID while sync summary works per table.

### DIFF
--- a/plugins/destination/file/docs/_authentication.md
+++ b/plugins/destination/file/docs/_authentication.md
@@ -1,0 +1,1 @@
+This plugin does not require any authentication.
--- a/plugins/destination/file/docs/overview.md
+++ b/plugins/destination/file/docs/overview.md
@@ -4,6 +4,7 @@ stage: GA
 title: File Destination Plugin
 description: CloudQuery File destination plugin for exporting to CSV, JSON and Parquet
 ---
+
 # File Destination Plugin
 
 :badge
@@ -11,6 +12,10 @@ description: CloudQuery File destination plugin for exporting to CSV, JSON and P
 This destination plugin lets you sync data from a CloudQuery source to local files in various formats. It currently supports CSV, line-delimited JSON and Parquet.
 
 This plugin is useful in local environments, but also in production environments where scalability, performance and cost are requirements. For example, this plugin can be used as part of a system that syncs sources across multiple virtual machines, uploads Parquet files to a remote storage (such as S3 or GCS), and finally loads them to data lakes such as BigQuery or Athena in batch mode. If this is your end goal, you may also want to look at more specific destination cloud storage destination plugins such as [S3](/docs/plugins/destinations/s3/overview), [GCS](/docs/plugins/destinations/gcs/overview) or [Azure Blob Storage](/docs/plugins/destinations/azblob/overview).
+
+## Authentication
+
+:authentication
 
 ## Example
 
@@ -39,7 +44,7 @@ This is the (nested) spec used by the file destination Plugin.
 
 - `format` (`string`) (**required**)
 
-  Format of the output file.  Supported values are `csv`, `json` and `parquet`.
+  Format of the output file. Supported values are `csv`, `json` and `parquet`.
 
 - `format_spec` ([format_spec](#format_spec)) (optional)
 
@@ -97,6 +102,6 @@ Reserved for future use.
 
   Some Parquet readers require a specific root repetition option to be able to read the file. For example, importing Parquet files into [Snowflake](https://www.snowflake.com/en/) requires the root repetition to be `undefined`.
 
-- `max_row_group_length` (`integer`) (optional) (default: `134217728` (= 128 * 1024 * 1024))
+- `max_row_group_length` (`integer`) (optional) (default: `134217728` (= 128 _ 1024 _ 1024))
 
   The maximum number of rows in a single row group. Use a lower number to reduce memory usage when reading the Parquet files, and a higher number to increase the efficiency of reading the Parquet files.
--- a/plugins/destination/postgresql/docs/authentication.md
+++ b/plugins/destination/postgresql/docs/authentication.md
@@ -1,0 +1,6 @@
+In order to store information in a PostgreSQL database, cloudquery needs to be authenticated. Insert the PostgreSQL connection string in the `connection_string` field in the sync config file.
+
+Acceptable formats are:
+
+- URI: `postgres://postgres:pass@localhost:5432/postgres?sslmode=disable`. Any special URI characters need to be percent-encoded.
+- DSN: `"user=postgres password=pass+0-[word host=localhost port=5432 dbname=postgres sslmode=disable"`
--- a/plugins/destination/s3/docs/authentication.md
+++ b/plugins/destination/s3/docs/authentication.md
@@ -1,0 +1,36 @@
+---
+hub-order: 1
+---
+
+## Quickstart
+
+The plugin needs to be authenticated with your AWS account(s) in order to sync information from your cloud setup.
+
+The plugin requires only `PutObject` permissions (we will never make any changes to your cloud setup), so, following the principle of least privilege, it's recommended to grant it `PutObject` permissions.
+
+If you are running CloudQuery CLI locally, and have AWS CLI installed, [sign in with AWS CLI](https://docs.aws.amazon.com/signin/latest/userguide/command-line-sign-in.html).
+
+Test that your AWS CLI is working:
+
+```shell
+aws account list-regions
+```
+
+### Non-interactive Authentication
+
+There are multiple ways to authenticate with AWS, and the plugin respects the AWS credential provider chain. This means that AWS plugin will follow the following priorities when attempting to authenticate:
+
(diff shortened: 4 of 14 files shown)


cli/cmd/sync_v3.go
```diff
@@ -566,6 +579,9 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
 						return err
 					}
 				}
+				stats, _ = statsPerTable.Get(tableName)
+				stats.LastMessageTime = time.Now()
```
disq: Record last message time after it's been sent to all destinations.

cli/cmd/sync.go
```diff
@@ -167,20 +167,39 @@ func sync(cmd *cobra.Command, args []string) error {
...
 			if source.OtelEndpoint != "" {
 				sourcesWithOtelEndpoint = append(sourcesWithOtelEndpoint, source.Name)
 			}
 		}
-		if len(sourcesWithOtelEndpoint) > 0 {
+		if tableMetricsLocation != "" && len(sourcesWithOtelEndpoint) > 0 {
 			return fmt.Errorf("the `--tables-metrics-location` flag is not supported for sources with `otel_endpoint` configured. Either remove the `--tables-metrics-location` flag or do not configure `otel_endpoint` for the following sources: %s", strings.Join(sourcesWithOtelEndpoint, ", "))
 		}
-		otelReceiver, err = otel.StartOtelReceiver(ctx, otel.WithMetricsFilename(tableMetricsLocation))
-		if err == nil {
```
disq, line 182: Error handling was missing here

cli/cmd/sync.go
```diff
@@ -167,20 +167,39 @@ func sync(cmd *cobra.Command, args []string) error {
...
-			defer otelReceiver.Shutdown(ctx)
+		if len(sourcesWithOtelEndpoint) == 0 {
+			otelReceiver, err = otel.StartOtelReceiver(ctx, otel.WithMetricsFilename(tableMetricsLocation))
```
disq: Here we start otel receiver even with an empty tableMetricsLocation if we got here, if we need summary by `--summary-location` or `send_sync_summary`.

cli/internal/otel/receiver.go
```diff
@@ -135,7 +145,7 @@ func newMetricConsumer(metricsFile *os.File, quit chan any, wg *sync.WaitGroup)
 		}
 	}()
 
-	return func(ctx context.Context, metric pluginMetric) {
+	consumerFunc := func(ctx context.Context, metric pluginMetric) {
```
disq: Not just `return func(..) { .. }, func(..) { .. }` to preserve indentation for untouched code.

cli/internal/otel/receiver.go
```diff
@@ -160,6 +170,16 @@ func newMetricConsumer(metricsFile *os.File, quit chan any, wg *sync.WaitGroup)
...
 		case "sync.table.duration":
 			metrics.Duration = &metric.Value
+
+			d, ok := durationsMap[table]
+			if !ok {
+				d = new(time.Duration)
+				durationsMap[table] = d
+			}
+			dur := time.Duration(metric.Value * int64(time.Millisecond))
+			if d == nil || dur > *d { // Assume longest duration is the full sync duration between multiple clientIDs
```
disq: We can change this once `clientId` is removed.

cli/cmd/sync.go
```diff
@@ -167,20 +167,39 @@ func sync(cmd *cobra.Command, args []string) error {
...
+			if err != nil {
+				log.Warn().Err(err).Msg("Failed to start otelReceiver")
+			} else {
+				defer otelReceiver.Shutdown(ctx)
+			}
+		} else {
+			log.Warn().Strs("sources", sourcesWithOtelEndpoint).Msg("Sync Summary requested with sources having `otel_endpoint` configured. Using the configured `otel_endpoint` for table duration info.")
```
erezrokah: What does this mean? How can we use the configured `otel_endpoint` to get table duration?

disq: ok clarified warning in https://github.com/cloudquery/cloudquery/pull/21335/commits/fc3f5dd6d6aa35f8868a37b62c04b08e48cb26d4

cli/internal/otel/receiver.go
```diff
@@ -242,19 +272,24 @@ func WithMetricsFilename(filename string) OtelReceiverOption {
 
 func StartOtelReceiver(ctx context.Context, opts ...OtelReceiverOption) (*OtelReceiver, error) {
```
erezrokah, line 260: Maybe a cleaner API would be to pass an additional option to `StartOtelReceiver` where the option is a callback that gets invoked on each time we get a new duration metric?

disq: Good idea, done in https://github.com/cloudquery/cloudquery/pull/21335/commits/7fc42cebe2fec2f469acb45104b9dc857d7111ef

erezrokah (approved):

disq:
Just tested, seems to work.
```
{
    "cli_version": "development",
    "destination_errors": 0,
    "destination_name": "file",
    "destination_path": "cloudquery/file",
    "destination_version": "v5.4.22",
    "destination_warnings": 0,
    "resources": 5,
    "source_errors": 0,
    "source_name": "anthropic",
    "source_path": "cloudquery/anthropic",
    "source_version": "v1.0.0",
    "source_warnings": 0,
    "source_tables":
    [
        "anthropic_message_batches",
        "anthropic_organization_api_keys",
        "anthropic_organization_invites",
        "anthropic_organization_users",
        "anthropic_organization_workspace_members",
        "anthropic_organization_workspaces"
    ],
    "sync_id": "3259e73e-6e5e-4b31-a020-773290086ba6",
    "sync_time": "2025-09-04T08:09:24.284376Z",
    "resources_per_table":
    {
        "anthropic_message_batches": 0,
        "anthropic_organization_api_keys": 2,
        "anthropic_organization_invites": 0,
        "anthropic_organization_users": 1,
        "anthropic_organization_workspace_members": 1,
        "anthropic_organization_workspaces": 1
    },
    "errors_per_table":
    {
        "anthropic_message_batches": 0,
        "anthropic_organization_api_keys": 0,
        "anthropic_organization_invites": 0,
        "anthropic_organization_users": 0,
        "anthropic_organization_workspace_members": 0,
        "anthropic_organization_workspaces": 0
    },
    "durations_per_table_ms":
    {
        "anthropic_message_batches": 576,
        "anthropic_organization_api_keys": 283,
        "anthropic_organization_invites": 270,
        "anthropic_organization_users": 270,
        "anthropic_organization_workspace_members": 138,
        "anthropic_organization_workspaces": 402
    }
}
```

My only gripe is the repetition of table names, we should consider switching to a `table_stats` object and having all stats in one object.

cli/cmd/sync.go
```diff
@@ -167,20 +167,40 @@ func sync(cmd *cobra.Command, args []string) error {
...
 		}
-		if len(sourcesWithOtelEndpoint) > 0 {
+		if tableMetricsLocation != "" && len(sourcesWithOtelEndpoint) > 0 {
 			return fmt.Errorf("the `--tables-metrics-location` flag is not supported for sources with `otel_endpoint` configured. Either remove the `--tables-metrics-location` flag or do not configure `otel_endpoint` for the following sources: %s", strings.Join(sourcesWithOtelEndpoint, ", "))
 		}
-		otelReceiver, err = otel.StartOtelReceiver(ctx, otel.WithMetricsFilename(tableMetricsLocation))
-		if err == nil {
-			defer otelReceiver.Shutdown(ctx)
+		if len(sourcesWithOtelEndpoint) == 0 {
+			otelReceiver, err = otel.StartOtelReceiver(ctx, otel.WithMetricsFilename(tableMetricsLocation), otel.WithDurationCallback(tableDurations.Set))
```
hermanschaaf, line 196: Does the duration callback get called once per client, or once per table?

disq: ~It's called once per client,~ because of that it has to decide which duration to use (longer duration wins out)

disq: Maybe we can consider keeping lowest startTime and highest endTime and use the difference as the duration. Otherwise if we're limited by concurrency/scheduler (if not all client ids are processed at once) the reported duration would be inaccurate.

hermanschaaf: Yeah, the duration depends a lot on when the scheduler decides to schedule clients. One way to do this somewhat accurately might be to keep track of the start/end times of all the clients, and then run an algorithm over them to sum them up, but removing overlapping sections.


For example:


```
           A      B
client 1   |------|

              C         D
client 2      |---------|

                                 E        F
client 3                         |--------|
```

In this case the duration should be:

(B - A) - (B - C) + (D - C) + (F - E)

So then we're at least accounting for the scheduler possibly scheduling one client right at the start, and another right at the end, as might happen frequently in the case of shuffle schedulers, for example.

disq: (although we can't do that because `sync.table.start_time` and `sync.table.end_time` are deprecated and no longer being sent)

disq: erm, seems `sync.table.client_id` is also removed and no longer being sent, I think that means they are called per table and the SDK takes care (or _should_ take care) of the duration calcs - currently it's not that smart:
https://github.com/cloudquery/plugin-sdk/blob/6be5fe6411812cb59cbdd0d3b8c03361a34f3e71/scheduler/metrics/metrics.go#L192-L201

PR #21340, opened 2025-09-05
fix: Snowflake unique constraint drop

disq:
Detect unique constraint name correctly, to ease migrating from 4.x series.

v4.5.5 Doesn't use a constraint name when creating so the name ends up being a `SYS_CONSTRAINT_<uuid>`: https://github.com/cloudquery/cloudquery/blob/plugins-destination-snowflake-v4.5.5/plugins/destination/snowflake/client/migrate.go#L170

v5.1.1 Assumes we're using a static name: https://github.com/cloudquery/cloudquery/blob/plugins-destination-snowflake-v5.1.1/plugins/destination/snowflake/client/migrate.go#L171-L173

This PR matches the column name with the list of unique constraints and drops it if it's `SYS_CONSTRAINT_` or in CQ format (`<table>_<column>_key`).


stoovon:
Overall LGTM although I agree with gocritic

stoovon (approved):

plugins/destination/snowflake/client/migrate.go
```diff
@@ -167,13 +167,28 @@ func (c *Client) alterColumnDropNotNull(ctx context.Context, tableName string, c
 	return nil
 }
 
-func (c *Client) alterColumnDropUniqueConstraint(ctx context.Context, table *schema.Table, change schema.TableColumnChange) error {
-	// We only support the default unique constraint name
-	// If it is using a unique constraint that is not default it means CQ didn't create it so we shouldn't drop it
-	indexName := uniqueConstraintName(table.Name, change.ColumnName)
-	sql := "alter table identifier(?) drop constraint " + sanitizeColumn(indexName)
```
disq, line 174: We could've further simplified this PR by just doing this instead and not adding all the constraint name logic:
```go
sql := "alter table identifier(?) drop unique (" + sanitizeColumn(change.ColumnName) + ")"
```
(but then it would drop any type of unique constraint, not just `SYS_CONSTRAINT_*` and `<table>_<column>_KEY`)

PR #21358, opened 2025-09-15
feat: Cloud-API-based AI Onboarding command.

marianogappa:
This PR enables the Open AI-based chat onboarding mode on `cloudquery init`.

The Cloud API chat endpoints are already merged and functional but the client API changes haven't been released yet, so this is a WIP simply because I cannot pull that dependency yet, but the code itself should we final minus any bugs.

Upon merging, the `cloudquery init` command will start an AI chat to help the user set up the first sync.

This behaviour can be prevented in the following ways:
- Call `cloudquery init` with the `-source` and `-destination` flags already.
- Use the `-offline` flag to force the existing simple interactive mode.

Chats are per-user and may be subject to per team-day token limits.

Here's an example chat:
Screenshot 2025-09-15 at 16 04 47

### DIFF
--- a/cli/cmd/ai.go
+++ b/cli/cmd/ai.go
@@ -1,0 +1,122 @@
+package cmd
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
+	"github.com/cloudquery/cloudquery/cli/v6/internal/api"
+
+	"github.com/fatih/color"
+)
+
+var (
+	aiBold    = color.New(color.Bold)
+	aiSuccess = color.New(color.Bold, color.FgGreen)
+	aiInfo    = color.New(color.Bold, color.FgCyan)
+)
+
+func aiCmd(ctx context.Context, client *cloudquery_api.ClientWithResponses, teamName string) error {
+
+	// Track AI session start
+	TrackAISessionStarted(ctx, invocationUUID.UUID)
+
+	fmt.Println()
+	aiSuccess.Println("🤖 CloudQuery AI Assistant")
+	fmt.Println("I'm here to help you set up CloudQuery syncs!")
+	fmt.Println("Type 'exit' or 'quit' to end the conversation.")
+	fmt.Println()
+	fmt.Println("What are you trying to build with CloudQuery?")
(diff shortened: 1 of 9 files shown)


cli/cmd/ai.go
```diff
@@ -0,0 +1,116 @@
...
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
+	"github.com/cloudquery/cloudquery/cli/v6/internal/api"
+	"github.com/samber/lo"
+
```
disq: I know there's no `make goimports` target for the cli, but we should still remove this line and sort the imports.

marianogappa: Fixed.

cli/cmd/init.go
```diff
@@ -56,6 +56,7 @@ func newCmdInit() *cobra.Command {
 	cmd.Flags().String("destination", "", "Destination plugin name or path")
 	cmd.Flags().String("spec-path", "", "Output spec file path")
 	cmd.Flags().Bool("yes", false, "Accept all defaults")
+	cmd.Flags().Bool("offline", false, "Run in offline mode without requiring authentication")
```
erezrokah: Not sure `offline` is the correct name for it. We still send requests if it's true, and also will use user's credentials if they exist (though not require it).

Maybe call it `disable-ai`?

cli/cmd/init.go
```diff
@@ -270,6 +276,23 @@ func initCmd(cmd *cobra.Command, args []string) (initCommandError error) {
 		}
 	}
 
+	// TODO: For now, this is enabled only for @cloudquery.io users
```
erezrokah, line 279: Can we add true feature flagging? Adding a LaunchDarkly client shouldn't be that high effort?

marianogappa: There is a real feature flag active on the server side: https://app.launchdarkly.com/projects/default/flags/ai-onboarding-enabled/targeting?env=production&selected-env=production

The `NewConversation` endpoint returns 404 in this case and the code below is defaulting back to your code.

This email thing is just a last-minute requirement that Michal asked me for.

cli/cmd/init.go
marianogappa: Renamed

erezrokah (approved):

PR #21361, opened 2025-09-15
feat: Log CLI version

bbernays:
Currently the only way of knowing what version of the CLI is being run from the logs is for a successful sync to have completed and then looking in the sync_summary. This PR adds a log line at the beginning of all commands


disq (approved):
I remember logging all start args at some point, maybe it was in another life of our cli.

erezrokah (approved):

stoovon (approved):

bbernays:
> I remember logging all start args at some point, maybe it was in another life of our cli.

@disq i agree I thought we had already done this, but I went through the code and the output in my local log files and couldn't find it

PR #21369, opened 2025-09-19
chore: Implement AI Onboarding improvements.

marianogappa:
Implements AI Onboarding improvements:

- Upon running SIGINT or typing `exit` or `quit` on AI Onboarding, an `EndConversation` API call is performed.
- `--resume-conversation` flag is added so that a new `conversation_id` is not created before starting the chat. This allows a web chat to continue on terminal.

### DIFF
--- a/cli/cmd/ai.go
+++ b/cli/cmd/ai.go
@@ -40,6 +40,8 @@ func aiCmd(ctx context.Context, client *cloudquery_api.ClientWithResponses, team
 		}
 		return err
 	case <-ctx.Done():
+		// End the conversation when context is cancelled (e.g., Ctrl+C)
+		api.EndConversation(ctx, client, teamName)
 		fmt.Println("\nGoodbye! 👋")
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


dcelasun (approved):

erezrokah (approved):
Looks good, it should be a `fix:` if you want a new CLI release with it

cli/cmd/ai.go
```diff
@@ -76,6 +74,8 @@ func aiCmdInner(ctx context.Context, client *cloudquery_api.ClientWithResponses,
 
 		// Check for exit commands
 		if strings.ToLower(userInput) == "exit" || strings.ToLower(userInput) == "quit" {
+			// End the conversation before exiting
+			api.EndConversation(ctx, client, teamName)
```
disq, line 78: I would move this into the other line where we say Goodbye to the user (L39)

cli/cmd/ai.go
```diff
@@ -40,6 +40,8 @@ func aiCmd(ctx context.Context, client *cloudquery_api.ClientWithResponses, team
 		}
 		return err
 	case <-ctx.Done():
+		// End the conversation when context is cancelled (e.g., Ctrl+C)
+		api.EndConversation(ctx, client, teamName)
```
disq, line 44: Calling with cancelled context here, use `context.Background()` instead

cli/cmd/ai.go
marianogappa: Fix released https://github.com/cloudquery/cloudquery/pull/21372/files

cli/cmd/ai.go
marianogappa: Fix released https://github.com/cloudquery/cloudquery/pull/21372/files

PR #21384, opened 2025-09-22
fix: Not retry when making AI Onboarding requests.

marianogappa:
When there's a 5xx coming from an AI Onboarding request, by default the client is retrying many times. This is sending the very same request to OpenAI which then is rate-limiting due to 30k tokens being consumed within 1 minute. This PR uses a client that does not retry.


erezrokah (approved):

disq (approved):

cli/cmd/init.go
```diff
@@ -279,11 +281,16 @@ func initCmd(cmd *cobra.Command, args []string) (initCommandError error) {
 		if err != nil {
 			return err
 		}
+
+		apiClientWithoutRetries, err = api.NewClient(token.Value, cqapi.WithHTTPClient(&http.Client{}))
```
disq: why not `http.DefaultClient`?
```suggestion
		apiClientWithoutRetries, err = api.NewClient(token.Value, cqapi.WithHTTPClient(http.DefaultClient))
```

PR #21386, opened 2025-09-23
fix: Provide visual feedback of chat completion.

marianogappa:
The delay in responses on AI Onboarding can be confusing. This is because there's a lot of back and forth between CLI, Cloud API and OpenAI API. Also, for some Cloud API operations, there's many requests due to pagination. So the user may experience delay for some operations, and no feedback can be confusing.

This PR implements visual feedback as seen on this video:

https://github.com/user-attachments/assets/1338f112-53e9-4d13-a540-6d1cfc839a3c

### DIFF
--- a/cli/cmd/ai.go
+++ b/cli/cmd/ai.go
@@ -75,6 +75,9 @@ func startSpinner(ctx context.Context, messages []string) func() {
 	// Create a context for the spinner that can be cancelled
 	spinnerCtx, spinnerCancel := context.WithCancel(ctx)
 
+	// Channel to signal when cleanup is complete
+	cleanupDone := make(chan struct{})
+
 	// Start the message rotation
 	go func() {
 		ticker := time.NewTicker(4 * time.Second)
@@ -101,6 +104,7 @@ func startSpinner(ctx context.Context, messages []string) func() {
 	spinnerIndex := 0
 
 	go func() {
+		defer close(cleanupDone)
 		ticker := time.NewTicker(100 * time.Millisecond)
 		defer ticker.Stop()
 
@@ -111,8 +115,6 @@ func startSpinner(ctx context.Context, messages []string) func() {
 				stopped = true
 				mu.Unlock()
 				_ = bar.Clear()
-				// Give a moment for the clear to take effect
-				time.Sleep(50 * time.Millisecond)
 				return
 			case <-ticker.C:
 				mu.Lock()
@@ -140,8 +142,7 @@ func startSpinner(ctx context.Context, messages []string) func() {
 	// Return a stop function
 	return func() {
 		spinnerCancel()
-		// Give a moment for the cleanup to take effect
-		time.Sleep(100 * time.Millisecond)
+		<-cleanupDone
 	}
 }
 


cli/cmd/ai.go
```diff
@@ -23,6 +26,124 @@ var (
...
+func showSpinner(ctx context.Context, messages []string, done chan struct{}) {
+	var mu syncpkg.Mutex
+	var currentMessage string
+	var stopped bool
+
+	// Create a progressbar for the spinner
+	bar := progressbar.NewOptions(-1,
+		progressbar.OptionSetDescription(""),
+		progressbar.OptionSetWidth(50),
+		progressbar.OptionShowCount(),
```
disq: Is this used at all?
```suggestion
```

disq: This is what draws the `(0/-)` part.

cli/cmd/ai.go
```diff
@@ -9,12 +9,15 @@ import (
 	"os/exec"
 	"os/signal"
 	"strings"
+	syncpkg "sync"
```
disq: We traditionally call this `gosync`

cli/cmd/ai.go
```diff
@@ -90,6 +220,11 @@ func aiCmdInner(ctx context.Context, client *cloudquery_api.ClientWithResponses,
 				if err != nil {
 					return fmt.Errorf("failed to create spec file: %w", err)
 				}
+
+				// Show spinner while waiting for API response after creating spec file
+				spinnerDone := make(chan struct{})
```
disq: You can get rid of this and use a sub `CancelContext` and cancel that to stop the spinner, as the code doesn't do anything special while selecting from ctx.Done and spinnerDone.

cli/cmd/ai.go
```diff
@@ -90,6 +220,11 @@ func aiCmdInner(ctx context.Context, client *cloudquery_api.ClientWithResponses,
...
+				go showSpinner(ctx, specFileMessages, spinnerDone)
```
disq: You could refactor this into
`stop := startSpinner(ctx, specFileMessages)`
which would do the cancel context and the goroutine firing, returning a fn to call to stop.
(You still need to call `defer stop()` at some point here to prevent a ctx leak, if linter picks up on that)

pilvikala:
So where's my dancing bear?

cli/cmd/ai.go
marianogappa: Renamed

cli/cmd/ai.go
marianogappa: Refactored

cli/cmd/ai.go
marianogappa: Refactored

marianogappa:
@pilvikala I don't put dancing bears on public repos

pilvikala:
> @pilvikala I don't put dancing bears on public repos

This will have to do then :) Looks good!

cli/cmd/ai.go
```diff
@@ -23,6 +26,125 @@ var (
...
+				_ = bar.Add(0) // Trigger redraw
+			}
+		}
+	}()
+
+	// Return a stop function
+	return func() {
+		spinnerCancel()
+		// Give a moment for the cleanup to take effect
+		time.Sleep(100 * time.Millisecond)
```
disq: Isn't this arbitrary? Another channel, with `defer close(quit)` in the second goroutine and then we `<-quit` here would be cleaner

disq:
> Looks good, but can we drop the `(0/-)` counter? Doesn't seem like it adds any value

It's been removed, with https://github.com/cloudquery/cloudquery/pull/21386#discussion_r2371530108

cli/cmd/ai.go
marianogappa: Refactored

marianogappa:
Still looks good:
```
- ⠴ 📚 Consulting plugin documentation...  [10s]
```

PR #21391, opened 2025-09-25
feat: Add `time_partitioning_expiration` spec option

disq:
It's not per-table, but it's a start.

### DIFF
--- a/plugins/destination/bigquery/client/embeddings_migrate.go
+++ b/plugins/destination/bigquery/client/embeddings_migrate.go
@@ -14,18 +14,19 @@ func (c *ConcreteEmbeddingsClient) MigrateTables(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+
+		if ok, err := c.client.doesTableExist(ctx, c.client.client, tableConfig.TargetTableName); err != nil {
+			return err
+		} else if ok {
+			continue
+		}
+
 		tm := bigquery.TableMetadata{
 			Name:             tableConfig.TargetTableName,
 			Location:         "",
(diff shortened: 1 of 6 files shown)


plugins/destination/bigquery/client/migrate.go
```diff
@@ -215,13 +215,15 @@ func (c *Client) timePartitioning() *bigquery.TimePartitioning {
...
-			Field: "_cq_sync_time",
+			Type:       bigquery.HourPartitioningType,
+			Field:      schema.CqSyncTimeColumn.Name,
+			Expiration: c.spec.TimePartitioningExpiration.Duration(),
 		}
 	case TimePartitioningOptionDay:
 		return &bigquery.TimePartitioning{
-			Type:  "DAY",
-			Field: "_cq_sync_time",
+			Type:       bigquery.DayPartitioningType,
```
disq, line 224: BQ SDK also allows Month and Year partitioning but it's for another PR later on.

plugins/destination/bigquery/client/embeddings_migrate.go
```diff
@@ -22,12 +29,6 @@ func (c *ConcreteEmbeddingsClient) MigrateTables(ctx context.Context) error {
 			TimePartitioning: c.client.timePartitioning(),
 		}
 
-		if ok, err := c.client.doesTableExist(ctx, c.client.client, tableConfig.TargetTableName); err != nil {
```
disq, line 25: `doesTableExist` doesn't use `tm` (doesn't check against partitioning) so I moved it above the line where `tm` is initialized.

erezrokah (approved):

PR #21392, opened 2025-09-25
fix: Update team internal metadata upon switch and login

hermanschaaf:
Fixes a bug whereby the `internal` value in the cached config didn't get updated between login / switch events, meaning that analytics events could get sent (or not sent) incorrectly. 

This was a rare bug that only affected CloudQuery team members doing testing on the analytics system itself.

### DIFF
--- a/cli/internal/team/team.go
+++ b/cli/internal/team/team.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"slices"
 	"strings"
 
 	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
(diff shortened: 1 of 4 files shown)


cli/internal/team/team.go
```diff
@@ -87,3 +103,11 @@ func (c *Client) GetTeam(ctx context.Context, team string) (*Team, error) {
 
 	return resp.JSON200, nil
 }
+
+func Names(teams []Team) []string {
+	names := make([]string, 0, len(teams))
```
disq:
```suggestion
	names := make([]string, len(teams))
	for i := range teams {
	names[i] = teams[i].Name
	}
```

hermanschaaf: Updated in https://github.com/cloudquery/cloudquery/pull/21392/commits/b4ea48bb933ce65b7e0d9682a86aadb62043ff5b

PR #21401, opened 2025-09-26
fix: No AI helper when certain flags are provided

disq:
Fixes https://linear.app/cloudquery/issue/ENG-2099


disq:
Dupe of https://github.com/cloudquery/cloudquery/pull/21400

PR #21410, opened 2025-09-29
chore: Fix workflows permissions

erezrokah:
Needed since we changed the default permissions to read only


.github/workflows/missing_tables_summary.yml
erezrokah, line 1: Wasn't used (disabled manually in the UI(

PR #21472, opened 2025-10-06
fix: Change fips mode to on, prevent panic from crypto.

marianogappa:
We have the fips debug mode set to `only`, which causes panics or errors when certain crypto functions are run. This was not the intention of us setting that mode, so this PR sets it to `on`.

### DIFF
--- a/plugins/source/test/main_fips.go
+++ b/plugins/source/test/main_fips.go
@@ -1,6 +1,6 @@
 //go:build fipsEnabled
 
-//go:debug fips140=only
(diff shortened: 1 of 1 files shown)


murarustefaan (approved):
> This was not the intention of us setting that mode

I mean, enforcing only FIPS140-3 algorithms run in FIPS tagged bianries, but if the plugins are not working at all in combination with Go 1.25, it's the most reasonable thing to do until there's a more stable option.

https://github.com/golang/go/issues/74630

PR #21474, opened 2025-10-08
fix: Table with PK components should upsert.

marianogappa:
`schema.Column` defines both `PrimaryKey` and `PrimaryKeyComponent` for a column, and certain plugin's tables use one of the two but not the other.

The Postgres destination decides whether to insert or upsert based on the existence of PK columns on the table, but ignores PK components.

This causes tables with PK component only columns to fail with constraint violations on inserts, when the rows already exist on the DB side.

### DIFF
--- a/plugins/destination/postgresql/client/insert.go
+++ b/plugins/destination/postgresql/client/insert.go
@@ -67,7 +67,7 @@ func (c *Client) InsertBatch(ctx context.Context, messages message.WriteInserts)
 		if !ok {
 			// cache the query
 			table := c.normalizeTable(msg.GetTable())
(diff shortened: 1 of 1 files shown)


murarustefaan (approved):

erezrokah:
Also my expectation is that when you have `PrimaryKeyComponents` you still have `PrimaryKeys` - a single PK which is `_cq_id`

erezrokah:
OK we should always have at least one PK https://github.com/cloudquery/plugin-sdk/blob/c1edab76e81e05324647573f6a8428fb519f9a7a/schema/table.go#L124 so not sure what happens here.

Maybe it's append mode?

PR #21514, opened 2025-10-22
fix: Don't set constraints for staging tables

erezrokah:
#### Summary

Fixes https://linear.app/cloudquery/issue/ENG-2214/bugawspricing-missing-ondemand-terms-for-17percent-of-ec2-products (internal link), see https://community.cloudquery.io/t/cloudquery-aws-pricing-plugin-bug-missing-ondemand-terms-for-1-7-of-ec2-products/1235.

We don't (and can't) handle conflicts when we do `COPY` (in `copyFromFile`) to the staging table, so if a batch has conflicts we lose the data.

We don't really need to create staging tables with constraints, as we handle those when we insert the data into the final table


murarustefaan (approved):

hermanschaaf (approved):
Nice catch - would it be possible to add a test for this case to prevent a future regression?

erezrokah:
> would it be possible to add a test for this case to prevent a future regression?

Sure will add a test shortly

PR #21515, opened 2025-10-22
test: Test duplicates same batch

erezrokah:
Test for https://github.com/cloudquery/cloudquery/pull/21514

### DIFF
--- a/plugins/destination/duckdb/client/client.go
+++ b/plugins/destination/duckdb/client/client.go
@@ -63,8 +63,10 @@ func New(ctx context.Context, logger zerolog.Logger, spec []byte, _ plugin.NewCl
 	}
 
 	// Unregister the built-in UUID type to avoid conflicts with our own UUID type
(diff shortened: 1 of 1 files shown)


plugins/destination/duckdb/client/client.go
```diff
@@ -63,8 +63,10 @@ func New(ctx context.Context, logger zerolog.Logger, spec []byte, _ plugin.NewCl
 	}
 
 	// Unregister the built-in UUID type to avoid conflicts with our own UUID type
-	if err := arrow.UnregisterExtensionType(extensions.NewUUIDType().ExtensionName()); err != nil {
-		return nil, err
+	if _, ok := arrow.GetExtensionType(extensions.NewUUIDType().ExtensionName()).(*extensions.UUIDType); ok {
```
erezrokah, line 66: Have 2 tests unregistering the type meant a test will not have `uuid` support. Added a check to ensure we only unregister the default one

stoovon (approved):

PR #21516, opened 2025-10-23
feat: Publish FIPS versions

murarustefaan:
This pull request introduces support for FIPS-enabled builds for the BigQuery destination plugin, including workflow updates and code changes to differentiate between standard and FIPS builds. The changes ensure that FIPS-compliant binaries are built, tested, and published through the dedicated workflow.

### DIFF
--- a/.github/workflows/dest_bigquery.yml
+++ b/.github/workflows/dest_bigquery.yml
@@ -71,7 +71,7 @@ jobs:
 
   validate-fips:
     timeout-minutes: 30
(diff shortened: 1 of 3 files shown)


erezrokah (approved):

erezrokah:
We need to update this list https://github.com/cloudquery/cloudquery/blob/9bc35225bffa7bbd08768158adb05864f3142100/scripts/workflows/wait_for_required_workflows.js#L65

erezrokah (approved):

PR #21575, opened 2025-11-05
Added More Docs

hayden3456:
This PR adds source-linked, auto-generated docs for the codebase. Before you dismiss it: it includes inline source references, architecture diagrams, and I think it does a solid job. No hard feelings if you’d prefer not to include it.

If you find it useful, I’m building a documentation copilot to help explain why code is the way it is. It will:
- Auto-sync docs 
- Asks a few questions on PR's to gather understanding as to why
- Visualize the system at different levels of abstraction

I'm just building this out and would love to chat about what would help most so I can build something genuinely useful.
docforge.net

### DIFF
--- a/docs/autogenerated_docs/1-overview.md
+++ b/docs/autogenerated_docs/1-overview.md
@@ -1,0 +1,274 @@
+# Overview
+
+<details>
+<summary>Relevant source files</summary>
+
+The following files were used as context for generating this wiki page:
+
+- [.release-please-manifest.json](https://github.com/cloudquery/cloudquery/blob/5064c039/.release-please-manifest.json)
+- [cli/CHANGELOG.md](https://github.com/cloudquery/cloudquery/blob/5064c039/cli/CHANGELOG.md)
+- [cli/go.mod](https://github.com/cloudquery/cloudquery/blob/5064c039/cli/go.mod)
+- [cli/go.sum](https://github.com/cloudquery/cloudquery/blob/5064c039/cli/go.sum)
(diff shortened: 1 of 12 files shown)


erezrokah (commented):
Hi @hayden3456 and thanks for the contribution, with big contributions it's always better to open an issue first describing the problem you're trying to solve and a proposed approach to solve it, so we can discuss it on the issue. I recommend doing that first, as it increases the chances of a contribution being accepted.

For the content of this PR:

1. We already have autogenerated docs for CLI commands and our plugin specs are already well documented.

2. Have you validated the docs? In many cases they are incorrect, for example https://github.com/cloudquery/cloudquery/blob/ff919dd47494fd3a639163b0e17aa295e9dbec55/docs/autogenerated_docs/2-cli.md#configuration-format, some options for `registry` are incorrect and some are missing. In other cases the markdown is broken, see https://github.com/cloudquery/cloudquery/blob/ff919dd47494fd3a639163b0e17aa295e9dbec55/docs/autogenerated_docs/6-plugin-development.md#core-components and https://github.com/cloudquery/cloudquery/blob/ff919dd47494fd3a639163b0e17aa295e9dbec55/docs/autogenerated_docs/6-plugin-development.md#conclusion.
The summary is missing here https://github.com/cloudquery/cloudquery/blob/ff919dd47494fd3a639163b0e17aa295e9dbec55/docs/autogenerated_docs/3.2-destination-plugins.md#summary.

4. The docs include a lot of deprecated code paths that should not be included.

5. Our framework also has private, closed source code so the docs are missing context about that part.

I'll close the PR as I think it's better to discuss what's missing from the repo in an issue, then decide on what we want to add first, then open scoped PRs for each topic that's missing from the docs

PR #21633, opened 2025-11-17
fix: Properly handle escaping newlines in private keys

bbernays:
This PR enables users to not have to manually escape new line characters in private key files when using the native file replacement functionality


cli/internal/specs/v0/spec_reader.go
```diff
@@ -384,3 +385,17 @@ func stripYamlComments(b []byte) ([]byte, error) {
...
 }
+
+func escapeContent(content []byte) bool {
+	if !bytes.ContainsAny(content, "\n\r") {
+		return false
+	}
+	if json.Valid(content) {
+		return true
+	}
+	block, _ := pem.Decode([]byte(content))
```
stoovon:
```suggestion
	block, _ := pem.Decode(content)
```

cli/internal/specs/v0/spec_reader.go
```diff
@@ -384,3 +385,17 @@ func stripYamlComments(b []byte) ([]byte, error) {
 	}
 	return b, nil
 }
+
+func escapeContent(content []byte) bool {
```
hermanschaaf: It might be good to document what the `bool` that this function returns indicates, it's not easy to discern by looking at it

hermanschaaf: oh sorry no it's the naming that threw me off, maybe something like `shouldEscape(...) bool` ?

bbernays: Done.
[3f1f29b](https://github.com/cloudquery/cloudquery/pull/21633/commits/3f1f29bc84c2b78d7a50cd5bb53833ac7a90fff8)

[789e88d](https://github.com/cloudquery/cloudquery/pull/21633/commits/789e88dfc132ddee2f9a856354af8b89523291d0)

cli/internal/specs/v0/spec_reader.go
```diff
@@ -384,3 +385,17 @@ func stripYamlComments(b []byte) ([]byte, error) {
...
+func shouldEscapeFileContent(content []byte) bool {
+	if !bytes.ContainsAny(content, "\n\r") {
+		return false
+	}
+	if json.Valid(content) {
+		return true
+	}
+	block, _ := pem.Decode(content)
+
+	// if block is nil then content is a valid PEM block and should return true
```
disq: pem.Decode:
```
// Decode will find the next PEM formatted block (certificate, private key
// etc) in the input. It returns that block and the remainder of the input. If
// no PEM data is found, p is nil and the whole of the input is returned in
// rest. Blocks must start at the beginning of a line and end at the end of a line.
func Decode(data []byte) (p *Block, rest []byte) {
```

if block is nil, content isn't a valid pem block?

There is also an example in `pem_test.go` in the stdlib that does:
```go
	isEmpty := func(block *Block) bool {
		return block != nil && block.Type == "EMPTY" && len(block.Headers) == 0 && len(block.Bytes) == 0
	}
```

bbernays: https://github.com/cloudquery/cloudquery-private/blob/main/plugins/source/okta/client/spec.go#L87-L89

disq: I mean, the code is correct, but the comments are opposite.
```suggestion
	// if block is not nil then content is a valid PEM block and should return true
```

cli/internal/specs/v0/spec_reader.go
```diff
@@ -384,3 +385,17 @@ func stripYamlComments(b []byte) ([]byte, error) {
...
+	// If block is not nil then content is not a valid PEM block and should return false
```
disq:
```suggestion
	// If block is nil then content is not a valid PEM block and should return false
```

bbernays: [46add7c](https://github.com/cloudquery/cloudquery/pull/21633/commits/46add7cde0686081962fc3c852cae57f972389f2)

cli/internal/specs/v0/spec_reader.go
bbernays: Gotchya, yeah did not pick up on that

bbernays: Fixed in [46add7c](https://github.com/cloudquery/cloudquery/pull/21633/commits/46add7cde0686081962fc3c852cae57f972389f2)

erezrokah:
@bbernays is there anything blocking this PR?

PR #21635, opened 2025-11-18
fix: Allow empty destinations in spec during test connection

disq:
Follow-up to
https://github.com/cloudquery/cloudquery/pull/17993
https://github.com/cloudquery/cloudquery/pull/18542

### DIFF
--- a/cli/cmd/testdata/source-no-destination.yml
+++ b/cli/cmd/testdata/source-no-destination.yml
@@ -1,0 +1,7 @@
+kind: "source"
+spec:
+  name: "test"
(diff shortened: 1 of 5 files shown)


cli/cmd/testdata/source-no-destination.yml
```diff
@@ -0,0 +1,7 @@
+kind: "source"
+spec:
+  name: "test"
+  path: "cloudquery/test"
+  destinations: []
```
disq, line 5: This is the key, no destinations.

PR #21675, opened 2025-12-01
feat: Add optional tenant ID field to logger

hermanschaaf:
For deployments with multiple tenants syncing to a single log sync, this adds the possibility of having a `tenant_id` field in the logger.


cli/internal/env/env.go
```diff
@@ -13,3 +13,7 @@ func IsCloud() bool {
 	_, ok := os.LookupEnv("CQ_CLOUD")
 	return ok
 }
+
+func TenantID() string {
+	return os.Getenv("CQ_TENANT_ID")
```
disq: why not `_CQ_TENANT_ID`?

hermanschaaf: Why `_CQ_TENANT_ID`?

hermanschaaf: All other env vars we support right now start with `CQ_`, apart from some legacy ones that were previously used for Cloud

disq: Because it's an internal variable, not a user facing one. No user will ever have to configure this. Also all cloud-related env vars (apart from `CQ_CLOUD`, which "enables" the processing for these vars) start with `_`.

hermanschaaf: Ok, I'll make it `_CQ_TENANT_ID`

hermanschaaf:
> filterPluginEnv needs an update to pass this to plugins.

@disq I'll test, but I think it doesn't, because we're not expecting plugins to read this env variable, it's being read by the manager which is invoked from the CLI

hermanschaaf:
I tested and found it to work as expected, I don't think we need to change `filterPluginEnv`

PR #21705, opened 2025-12-09
chore: Introduce Platform API and use for sync,tables,migrate.

marianogappa:
From this issue: https://linear.app/cloudquery/issue/ENG-2457/chorecli-update-cli-to-use-platform-go-sdk-for-syncs

CLI uses Cloud API for a bunch of different things, but we migrated syncs to Platform API so CLI doesn't work anymore until we point at platform for those.

Since CLI still needs to use Cloud API for other things that are unsupported on Platform API (e.g. AI Onboarding and publishing addons), this means that both APIs are used in CLI.

The auth mechanism has an identical API, but generating a token on Cloud API does not make it available on Platform API, so it's important that this mechanism is segregated per command in CLI that uses it.

Therefore I created two internal subpackages `cloud` and `platform` that create clients and handle auth and analytics.

The following commands now use Platform API:
- sync
- tables
- migrate

Everything else uses Cloud API.

Screenshot 2025-12-09 at 17 20 39


marianogappa:
@erezrokah suggested cloning cli instead

PR #21707, opened 2025-12-11
chore: Upgrade cloudquery-api-go & SDK to latest. Remove API reports.

marianogappa:
Removes API calls against Cloud that were removed. Upgrades `cloudquery-api-go` & `plugin-sdk` to latest version, continues to compile & run syncs properly.

Screenshot 2025-12-11 at 15 27 18

⚠️  This shouldn't be merged until Platform syncs are running against `platform-cli`.

### DIFF
--- a/cli/cmd/sync_v3.go
+++ b/cli/cmd/sync_v3.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/apache/arrow-go/v18/arrow"
-	cloudquery_api "github.com/cloudquery/cloudquery-api-go"
 	"github.com/cloudquery/cloudquery-api-go/auth"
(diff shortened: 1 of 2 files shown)


disq:
Some more to remove.

marianogappa:
Nobody merge yet 🤣 Well it's a chore so it wouldn't cause a release but still

disq:
Maybe convert to draft to avoid accidental button presses.

erezrokah:
> Nobody merge yet 🤣 Well it's a chore so it wouldn't cause a release but still

It's ok to merge it won't break anything. Even a release won't break anything since in order for platform syncs to use this we need a PR to update the CLI in the platform repo, and the platform repo doesn't use this version of CLI anymore

PR #21711, opened 2025-12-12
fix: Send DeleteStale messages before the sync summary

bbernays:


cli/cmd/sync_v3.go
```diff
@@ -633,6 +633,22 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
...
 
+	for i := range destinationsClients {
+		if destinationSpecs[i].WriteMode == specs.WriteModeOverwriteDeleteStale {
+			// Table names might have changed due to transformers
+			updatedTablesForDeleteStale := tableNameChanger.UpdateTableNames(destinationSpecs[i].Name, tablesForDeleteStale)
+			if err := deleteStale(writeClients[i], updatedTablesForDeleteStale, sourceName, syncTime); err != nil {
+				return err
+			}
+		}
+		if _, err := writeClients[i].CloseAndRecv(); err != nil {
```
hermanschaaf: Unless I'm reading it wrong, this `CloseAndRecv` looks like it could be a problem for sending the summaries later, as well as the `destinationsPbClients[i].Close` line below it. We should probably keep these after the summary sending step

cli/cmd/sync_v3.go
```diff
@@ -633,6 +633,17 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
...
+		if destinationSpecs[i].WriteMode == specs.WriteModeOverwriteDeleteStale {
+			// Table names might have changed due to transformers
+			updatedTablesForDeleteStale := tableNameChanger.UpdateTableNames(destinationSpecs[i].Name, tablesForDeleteStale)
+			if err := deleteStale(writeClients[i], updatedTablesForDeleteStale, sourceName, syncTime); err != nil {
+				return err
+			}
+		}
+	}
+
+	syncDurationMs := time.Since(syncTime)
```
erezrokah: We should probably not re-calculate this to avoid `syncTimeTook = time.Since(syncTime)` having a slightly different value in the logs vs sync summary

bbernays: This is a mistake, should not be in this PR, removed it in [6fc795f](https://github.com/cloudquery/cloudquery/pull/21711/commits/6fc795fcea6bfb9fdcedc8fcf6e97d4179f0cb10)

cli/cmd/sync_v3.go
bbernays: Fixed in [602e9c7](https://github.com/cloudquery/cloudquery/pull/21711/commits/602e9c7d90014f53b5ecb56b5a00f9123423149d)

cli/cmd/sync_v3.go
bbernays: Changes to the SyncDuration are in [this PR](https://github.com/cloudquery/cloudquery/pull/21712)

hermanschaaf (approved):

PR #21732, opened 2025-12-19
feat: Retry HTTP5xxs on PutObject

stoovon:
#### Summary

Specific and limited retries on PutObject

### DIFF
--- a/plugins/destination/s3/client/write.go
+++ b/plugins/destination/s3/client/write.go
@@ -16,11 +16,13 @@ import (
 	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	awstypes "github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/aws/smithy-go/transport/http"
 	"github.com/cloudquery/filetypes/v4"
 	"github.com/cloudquery/plugin-sdk/v4/message"
 	"github.com/cloudquery/plugin-sdk/v4/schema"
 	"github.com/cloudquery/plugin-sdk/v4/types"
 	"github.com/google/uuid"
+	"github.com/hashicorp/go-retryablehttp"
 )
 
 var reInvalidJSONKey = regexp.MustCompile(`\W`)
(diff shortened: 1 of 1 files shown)


plugins/destination/s3/client/write.go
```diff
@@ -70,9 +72,73 @@ func (c *Client) createObject(ctx context.Context, table *schema.Table, objKey s
...
+				_, upErr := uploader.Upload(ctx, params)
+				if upErr == nil {
+					return nil
+				}
+				lastErr = upErr
+
+				// If context canceled, return immediately
+				if ctx.Err() != nil {
+					return ctx.Err()
+				}
```
erezrokah, line 109: Isn't this duplicated with the code a bit before?

plugins/destination/s3/client/write.go
```diff
@@ -70,9 +72,73 @@ func (c *Client) createObject(ctx context.Context, table *schema.Table, objKey s
...
+
+		// Retry configuration: 3 retries (4 attempts total) with exponential backoff
+		const maxAttempts = 4
+		const minBackoff = 100 * time.Millisecond
+		const maxBackoff = 3 * time.Second
+
+		// If the reader supports seeking, we can safely retry failed uploads by seeking back to the
+		// start before each attempt (so the uploader can re-read the full content).
+		// If it doesn't (NoRotate mode), fall back to a single attempt since the data can't be re-read.
+		if rs, ok := adjustedReader.(io.ReadSeeker); ok {
```
erezrokah, line 87: In other places (SDK, plugins) we use https://github.com/hashicorp/go-retryablehttp

erezrokah:
Also based on the error message

```
operation error S3: PutObject, exceeded maximum number of attempts, 3, https response error StatusCode: 503
```

Looks like we're already retrying? Maybe only need to increase the retries to fix it?

erezrokah:
OK so what I think had changed is https://github.com/cloudquery/cloudquery/pull/20400/files#diff-4c5075e67222b2f9d06cdc2dc8d561e0638393bb9ddc2026629d524e60b61e2fR82.

The defaults before https://github.com/cloudquery/cloudquery/pull/20400 were:
- https://github.com/aws/aws-sdk-go-v2/blob/5a2fe4ac852d59999de13a2cd0a3095a1aec8027/aws/retry/standard.go#L173 
- https://github.com/aws/aws-sdk-go-v2/blob/5a2fe4ac852d59999de13a2cd0a3095a1aec8027/aws/retry/standard.go#L28

3 for `MaxAttempts` and 20 seconds for `MaxBackoff`. The difference is that before we didn't set `RateLimiter` to `None` which means we potentially hit the server less times as there was some client side limiting

plugins/destination/s3/client/write.go
stoovon: I don't think so, because I think that uploader.Upload can mutate ctx.

stoovon:
Closing in favour of #21733

PR #21733, opened 2025-12-19
fix: Bring back client side rate limiter

erezrokah:
#### Summary

Instead of https://github.com/cloudquery/cloudquery/pull/21732.

https://github.com/cloudquery/cloudquery/pull/20400 has 2 slight changes in behavior
1. Disabled client side rate limiting
2. Increased `MaxBackoff` from `20` to `30`.

While I think 2. is ok, 1. can have unwanted effects such as https://github.com/cloudquery/cloudquery/pull/21732#issuecomment-3675267462

### DIFF
--- a/plugins/destination/s3/client/client.go
+++ b/plugins/destination/s3/client/client.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/aws/ratelimit"
 	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	awshttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/config"
@@ -82,7 +81,6 @@ func New(ctx context.Context, logger zerolog.Logger, s []byte, opts plugin.NewCl
 			return retry.NewStandard(func(so *retry.StandardOptions) {
 				so.MaxAttempts = *c.spec.MaxRetries
 				so.MaxBackoff = time.Duration(*c.spec.MaxBackoff) * time.Second
-				so.RateLimiter = ratelimit.None
 			})
 		}),
 	}


bbernays:
Under what conditions are we hitting s3 rate limits?

erezrokah:
> Under what conditions are we hitting s3 rate limits?

It retries on `https response error StatusCode: 503` so don't think it's S3 rate limits, but still something we retry on

bbernays:
> > Under what conditions are we hitting s3 rate limits?
> 
> 
> 
> It retries on `https response error StatusCode: 503` so don't think it's S3 rate limits, but still something we retry on

But client side rate limits just limits the number of concurrent requests, why would we want that in this case?

erezrokah:
> But client side rate limits just limits the number of concurrent requests, why would we want that in this case?

- Before: All retries hit the server
- After: ~~Only some retries hit the server, some don't hit the server and backoff without sending requests~~ Actually not correct, but I think the impact is that we throttle the retries

We can also increased the default max retries to 10.

bbernays:
> > But client side rate limits just limits the number of concurrent requests, why would we want that in this case?
> 
> * Before: All retries hit the server
> * After: ~Only some retries hit the server, some don't hit the server and backoff without sending requests~ Actually not correct, but I think the impact is that we throttle the retries
> 
> We can also increased the default max retries to 10.

Based on the docs, it seems like if the client side rate limiter limits the request then the request will just fail:
```
A retry that attempts to consume more tokens than what's available results in operation failure with a [QuotaExceededError](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/ratelimit#QuotaExceededError).
```
Which I don't think is the behavior we want... 

https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-retries-timeouts.html

erezrokah:
> Which I don't think is the behavior we want...

OK I assumed it will just retry it again. So let's close this for now, I've asked the customer to increase the max retries, let's see if that helps

PR #21760, opened 2025-12-30
fix!: Correct nullability for map items

erezrokah:
#### Summary

`MapOf` always sets items to `Nullable: true` https://github.com/apache/arrow-go/blob/2d0962ed55074f050193b82aaf80f8b5995a2ffc/arrow/datatype_nested.go#L536


erezrokah:
We can also not fix this until someone reports this, as this is a breaking schema change

hermanschaaf:
I could be wrong, but I think we don't use the map type in any official source plugins, so this would only (potentially) affect database sources as a breaking change.

PR #22655, opened 2026-04-20
fix(xkcd): check HTTP status code before decoding response body

mailtoboggavarapu-coder:
## Problem

`GetComic` and `GetLatest` in the xkcd internal client call `c.client.Get(url)` and immediately decode the response body as JSON, without checking the HTTP status code first.

This means a non-200 response (e.g. a `404 Not Found` for a non-existent comic number) will attempt to decode the error response body as a `Comic`, producing a misleading JSON decode error instead of a clear, actionable message.

## Fix

Add an explicit `resp.StatusCode != http.StatusOK` check in both `GetComic` and `GetLatest` before decoding, and return a descriptive error that includes the status code and the requested resource.

## Changes

- `plugins/source/xkcd/internal/xkcd/xkcd.go`: add `"fmt"` import and HTTP status code guards in `GetComic` and `GetLatest`

### DIFF
--- a/plugins/source/xkcd/internal/xkcd/xkcd.go
+++ b/plugins/source/xkcd/internal/xkcd/xkcd.go
@@ -2,6 +2,7 @@ package xkcd
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strconv"
 )
@@ -14,7 +15,6 @@ type Comic struct {
 	Link       string `json:"link"`
 	Year       string `json:"year"`
 	News       string `json:"news"`
-	SafeTitle  string `json:"safe_title"`
 	Transcript string `json:"transcript"`
 	Alt        string `json:"alt"`
 	Img        string `json:"img"`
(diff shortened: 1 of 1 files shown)


mailtoboggavarapu-coder:
Hi @cloudquery/maintainers — this PR adds HTTP status code guards in `GetComic` and `GetLatest` in the xkcd source plugin. Previously a 404 or 5xx from the XKCD API would silently fall through to the JSON decoder and produce a confusing decode error. All 4 CI checks are passing. Would appreciate a review and merge when you get a chance — thank you! 🙏

plugins/source/xkcd/internal/xkcd/xkcd.go
```diff
@@ -14,7 +15,6 @@ type Comic struct {
 	Link       string `json:"link"`
 	Year       string `json:"year"`
 	News       string `json:"news"`
-	SafeTitle  string `json:"safe_title"`
```
erezrokah, line 17: Any reason this was removed?

erezrokah (commented):
Thanks for the PR @mailtoboggavarapu-coder, can we scope this change to the `resp.StatusCode != http.StatusOK` checks and do the identifiers renaming in a separate PR?

erezrokah:
Closing as stale, please re-open when https://github.com/cloudquery/cloudquery/pull/22655#discussion_r3116390800 is addressed

PR #22656, opened 2026-04-20
fix: Validate item_concurrency upper bound in Spec

mailtoboggavarapu-coder:
## Problem

`Spec.Validate()` returned `nil` unconditionally, accepting any value for `item_concurrency` — including arbitrarily large numbers like `99999`. Spawning that many goroutines against the HackerNews Firebase API causes severe rate-limiting, connection pool exhaustion, and eventually an opaque runtime crash — with no clear message pointing to the configuration.

## Fix

- Add an upper-bound check (`> 1000`) in `Validate()` that returns a descriptive error
- Update the `jsonschema` tag to include `maximum=1000` so generated docs and OpenAPI tooling stay in sync
- Change the receiver from `*Spec` to `*Spec` (was a value receiver `Spec`, which is inconsistent with `SetDefaults`)

## Files changed

`plugins/source/hackernews/client/spec.go`

### DIFF
--- a/plugins/source/hackernews/client/spec.go
+++ b/plugins/source/hackernews/client/spec.go
@@ -38,3 +38,6 @@ func (s *Spec) Validate() error {
 	}
 	return nil
 }
+
+//go:embed schema.json
+var JSONSchema string
--- a/plugins/source/hackernews/client/spec.go
+++ b/plugins/source/hackernews/client/spec.go
@@ -2,13 +2,14 @@ package client
 
 import (
 	_ "embed"
+	"fmt"
 
 	"github.com/cloudquery/plugin-sdk/v4/configtype"
 )
 
 type Spec struct {
 	// The number of items to fetch concurrently
-	ItemConcurrency int `json:"item_concurrency" jsonschema:"minimum=1,default=100"`
+	ItemConcurrency int `json:"item_concurrency" jsonschema:"minimum=1,maximum=1000,default=100"`
 	// RFC3339 formatted timestamp. Syncing will begin with posts after this date.
 	// Relative values like "3 days ago" are also supported.
 	// If not specified, the plugin will default to 24 hours ago.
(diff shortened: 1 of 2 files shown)


mailtoboggavarapu-coder:
Hi team! 👋 Friendly ping — this PR has just been rebased on top of the latest main. It validates the `item_concurrency` upper bound in the HackerNews spec. Ready for review!

mailtoboggavarapu-coder:
Hi team! 👋 Friendly ping — this PR has just been rebased on latest main. It validates `item_concurrency` upper bound in the HackerNews spec. Ready for review!

mailtoboggavarapu-coder:
Hi team! Pinging on this PR which validates the `item_concurrency` upper bound in the HackerNews plugin spec. The branch has been synced with latest main. Happy to address any feedback!

plugins/source/hackernews/client/spec.go
```diff
@@ -31,10 +32,9 @@ func (s *Spec) SetDefaults() {
...
-	// validation for configtype.Time is done on unmarshalling
+func (s *Spec) Validate() error {
+	if s.ItemConcurrency > 1000 {
+		return fmt.Errorf("item_concurrency must be at most 1000, got %d", s.ItemConcurrency)
+	}
 	return nil
 }
-
-//go:embed schema.json
-var JSONSchema string
```
erezrokah, line 40: Hi @mailtoboggavarapu-coder what's the reason this is removed?

erezrokah (changes requested):
Hi @mailtoboggavarapu-coder added a comment, also you might need to run `make gen` to update the JSON schema

plugins/source/hackernews/client/spec.go
mailtoboggavarapu-coder: Thanks for catching this, @erezrokah! You're absolutely right — the `//go:embed schema.json` + `var JSONSchema string` declaration should not have been removed. It's consumed by `plugin.go` (via `plugin.WithJSONSchema(client.JSONSchema)`) and by `spec_test.go`, so removing it would break both the plugin binary and the test suite.

Fixed in the latest commit: restored the embed directive, and also regenerated `schema.json` to include the new `"maximum": 1000` constraint that corresponds to the `jsonschema:"maximum=1000"` struct tag (equivalent to running `make gen`). Branch is now synced with the latest `main` as well.

erezrokah (approved):

erezrokah:
/gen sha=425f5dd66d44c2593f69dad85ab194f016a60e2a dir=plugins/source/hackernews

erezrokah (approved):

PR #22705, opened 2026-04-24
fix: MongoDB destination opt-in write retries for transient network errors

erezrokah:
#### Summary

Fixes [ENG-3281](https://linear.app/cloudquery/issue/ENG-3281/atomic-write-lock-during-mongodb-destination-sync) — the MongoDB Go driver's built-in retryable-writes only retries once, so two consecutive network failures (e.g. TCP broken pipe against an Atlas private-link endpoint) abort a sync batch.

Adds an opt-in `write_retry` spec block (`max_attempts`, `max_backoff`) that wraps `InsertMany` and `BulkWrite` with `avast/retry-go/v5`. Retries fire on `mongo.IsNetworkError`, `mongo.IsTimeout`, EOF, the server-side `RetryableWriteError` label, and the driver's `RetryablePoolError` interface (which covers `PoolClearedError`).

Retries are **disabled by default** (`max_attempts: 1`) to avoid changing existing behavior — `write_mode: append` tables can produce duplicate documents on retry because the server-side dedup tuple (session ID + txnNumber) is internal to the driver. Users with `write_mode: overwrite` (idempotent under retry) opt in by setting `max_attempts >= 2`. Documented in the destination's overview with a callout.

Includes a deterministic regression test that uses an in-process TCP proxy to inject the same `NetworkError`-labeled errors that surface in production, plus unit tests for the retry policy that don't require docker.

### DIFF
--- a/.github/workflows/dest_mongodb.yml
+++ b/.github/workflows/dest_mongodb.yml
@@ -20,9 +20,9 @@ permissions:
   contents: read
 
 jobs:
-  plugins-destination-mongodb:
-    timeout-minutes: 30
-    name: "plugins/destination/mongodb"
+  prepare:
+    timeout-minutes: 15
+    name: "plugins/destination/mongodb/prepare"
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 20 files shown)


disq (approved):

erezrokah:
Closing as stale. If we still need this please re-open cc @bbernays

bbernays:
@copilot resolve the merge conflicts in this pull request

bbernays:
@erezrokah - the customer that requested this functionality has confirmed that the RC we provided them was working as designed. I have re-opened the PR with the intention that it is merged

erezrokah:
> @erezrokah - the customer that requested this functionality has confirmed that the RC we provided them was working as designed. I have re-opened the PR with the intention that it is merged

There are some CI changes that broke the wait for all workflow, will fix them and merge

PR #22756, opened 2026-04-30
feat: Wrap writes in `WithTransaction`

bbernays:
The Tests fail because a standalone MongoDB (ie one running in a docker) will fail if a transaction is used


plugins/destination/mongodb/client/delete_stale.go
```diff
@@ -12,12 +12,19 @@ func (c *Client) DeleteStale(ctx context.Context, msgs message.WriteDeleteStales
...
+					schema.CqSourceNameColumn.Name: msg.SourceName,
+					schema.CqSyncTimeColumn.Name:   bson.M{"$lt": msg.SyncTime},
+				})
+				return err
+			})
+		})
+		if err != nil {
 			return err
 		}
+
```
disq:
```suggestion
```

plugins/destination/mongodb/client/spec/spec.go
```diff
@@ -55,6 +55,8 @@ type WriteRetryConfig struct {
 
 	// Maximum backoff between retry attempts.
 	MaxBackoff *configtype.Duration `json:"max_backoff,omitempty" jsonschema:"default=10s"`
+
+	UseTransactions bool `json:"use_transactions,omitempty"`
```
murarustefaan, line 71: We should be able to discover automatically whether the provided URI is a replica set or standalone instance and use transaction whenever possible.

murarustefaan: https://pkg.go.dev/go.mongodb.org/mongo-driver/mongo/description#TopologyKind

stoovon (approved):

stoovon (commented):
Looks good subject to the comment

plugins/destination/mongodb/client/spec/spec.go
bbernays: By default I do not want transactions to be enabled (even if the server supports it) until we can validate that it is working as expected.

bbernays: In a future major version, we **should** change the default behavior, but today, this should be an opt-in feature rather than an opt-out

murarustefaan: Ok, I'm fine with this.

murarustefaan (approved):

murarustefaan:
Let's document the field though.

stoovon (approved):

bbernays:
@erezrokah - Can we merge this PR and then provide a RC to the customer?

erezrokah:
> @erezrokah - Can we merge this PR and then provide a RC to the customer?

Go for it, you can also create an RC from the branch directly by pushing a tag

erezrokah (approved):

PR #22861, opened 2026-05-20
fix: Add SEO meta descriptions to CLI reference doc generator

JoeKarlsson:
Adds a `description:` frontmatter field to each generated CLI reference page so the descriptions survive future CLI version regenerations instead of being overwritten.


erezrokah:
Closing as stale

PR #22944, opened 2026-06-11
feat: Support Databricks Lakebase as a destination

bbernays:
I have validated that both normal connection string as well as OAuth methods work for integrating with Lakebase


erezrokah (approved):
I would put this as a separate plugin based on our experience from CockroachDB support in this destination (it grew to 2 separate branches in the code in many places).

Also this would mean we don't need to bundle the Databricks Go SDK (and a set of other dependencies) with the Postgres plugin.

Don't have a strong opinion on this so approving the PR, just a preference from maintenance perspective cc @murarustefaan

murarustefaan (changes requested):
Initially discussed on migrating onto a separate plugin.

murarustefaan (approved):

murarustefaan:
Yeah ok gave this some thought, since the core is 100% pg and only credential acquisition changes, we can have it as part of PG destination. 

In my mind it is similar to how source plugins operate, different means of getting credentials, same shared core; or to how the Go drivers work - same interfaces / implementations, different drivers for getting credentials.

PR #22977, opened 2026-06-19
feat: implement graceful degradation and timeout propagation in sync …

Aayushiii25:
1. **Fixed Memory Leaks in TransformerPipeline:** 
   Replaced `time.After(1 * time.Second)` with a reusable `time.NewTicker` inside the `Send()` and `startBlocking()` loops. This prevents the allocation of thousands of lingering `time.Timer` objects during high-throughput syncs.

2. **Fixed Goroutine Leaks in TransformerPipeline:** 
   Introduced `context.Context` cancellation for the background `Recv()` goroutines. When the pipeline closes, the background goroutine now cleanly exits instead of being orphaned and blocking on a dead channel indefinitely.

3. **Best-Effort Parallel Destination Shutdown:** 
   Previously, destination cleanup in `sync_v2.go` and `sync_v3.go` was sequential and fail-fast. If the first destination failed to close, subsequent destinations were never closed (leaving dangling streams and missing summary data). This PR implements a parallel shutdown sequence using `sync.WaitGroup`, collecting all errors securely via `sync.Mutex` and `errors.Join`, ensuring all destinations attempt to close cleanly.

4. **Plugin Timeout Propagation:** 
   Previously, the CLI lacked timeout propagation for plugin lifecycle methods, meaning a stuck plugin would hang the entire CLI indefinitely. This PR introduces a `--plugin-timeout` CLI flag (defaulting to `0` for backward compatibility). It propagates a derived context to `Init`, `Configure`, `Close`, `CloseAndRecv`, and `testPluginConnection` gRPC calls to ensure the CLI can fail gracefully.

### DIFF
--- a/cli/cmd/specs.go
+++ b/cli/cmd/specs.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/cloudquery/cloudquery/cli/v6/internal/specs/v0"
 	"github.com/cloudquery/plugin-pb-go/pb/plugin/v3"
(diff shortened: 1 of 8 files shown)


erezrokah (commented):
Hi @Aayushiii25, thank you for your contribution. Since this PR seems to try and fix 4 different distinct issues, do you mind opening a bug or feature report per fix first?

Please detail the impact of the bug/feature and share a reproduction so we can validate it on our end.

Cheers

erezrokah:
Closing per my last comment https://github.com/cloudquery/cloudquery/pull/22977#pullrequestreview-4531895368. Please open separate issues first

PR #22992, opened 2026-06-25
feat: Headless platform-destination sync via cqpd_ token (no cloudquery login)

disq:
Lets `cloudquery sync` / `migrate` run against a CloudQuery Platform tenant using only a pre-minted `cqpd_` token — **no `cloudquery login`** — for the full flow: plugin download, destination auto-injection, premium-table auth, and version pinning. Everything the CLI needs (tenant, API URL, owning team) rides in the token's claims (`t`/`u`/`tm`).

## How a cqpd_ token is supplied
Either env works, and they behave identically (one `platformToken()` resolver):
- `CQ_PLATFORM_TOKEN=cqpd_…` — explicit, separable platform-destination credential.
- `CLOUDQUERY_API_KEY=cqpd_…` — the standard credential env, when it holds a cqpd_.

## What the CLI does with it
- **Download** (`DownloadAuth`): uses the cqpd_ as the plugin-download credential and the team from the `tm` claim → `managedplugin` hits the team-scoped `DownloadPluginAssetByTeam`, so downloads are authorized + recorded against the tenant's cloud team. The cqpd_ is syncs-scoped (can't enumerate teams), so the team comes from the claim, not a cloud call.
- **Inject** (`MaybeInjectDestination`): auto-injects the `platform` destination from the token alone — no cloud login / tenant discovery / session mint. `api_url` comes from the `u` claim (not injected into the spec).
- **Detect** (`DetectTenant`): `init` reports the target tenant from the `u` claim.
- **Premium auth** (`PropagatePluginCredential`): exports the cqpd_ as `CLOUDQUERY_API_KEY` for spawned plugins (which validate premium tables / report usage against cloud from their inherited env) — **only when unset**, so a user's own key wins.
- **Version pinning**: in the headless flow the CLI fetches the recommended plugin version from the tenant's `…/external-syncs/whoami` (via `u`, authed with the token) and pins the injected plugin. Precedence: `CQ_PLATFORM_PLUGIN_VERSION` env > whoami-recommended > CLI default (`v1.0.1`). Best-effort with debug logging on every branch.

Non-headless (logged-in) flow is unchanged — it still resolves the token/team via `cloudquery login` and gets the version from the session mint.

## Cross-repo dependencies
- platform: `tm` claim + admin mint (3082, merged), platform-direct mint `tm` (3117), whoami `plugin_version` + default `v1.0.1` (3162)
- cloud: relay `tm` into the mint (8642)
- plugin: derive `api_url` from the token (cloudquery-private#12858, merged) — needs a **`v1.0.1` release** for the default download to resolve.

## Tests
`DownloadAuth` (CQ_PLATFORM_TOKEN + CLOUDQUERY_API_KEY paths), `TeamFromToken`, direct-token inject (both envs), whoami version-pin (httptest), `PropagatePluginCredential` (set-when-unset / never-clobber / non-cqpd no-op), `DetectTenant` multi-tenant. Build/vet/`go test ./internal/platform/` green.


stoovon (approved):

cli/internal/platform/inject.go
```diff
@@ -122,25 +126,150 @@ func DetectTenant(ctx context.Context, token, teamName string) (apiURL string, o
...
+// PropagatePluginCredential makes a headless cqpd_ token available to spawned
+// source/destination plugins as CLOUDQUERY_API_KEY. Plugins validate premium
+// tables and report usage against cloud using that env var (read from their own
+// process env, which in local runs they inherit from us); a cqpd_ supplied only
+// via CQ_PLATFORM_TOKEN would otherwise be invisible to them. No-op for a
+// non-cqpd_ token, and it never overwrites an existing CLOUDQUERY_API_KEY — a
+// user who set their own (e.g. a team key) keeps it for plugin auth.
+func PropagatePluginCredential(token string) {
+	if strings.HasPrefix(token, cqpdPrefix) && os.Getenv("CLOUDQUERY_API_KEY") == "" {
+		_ = os.Setenv("CLOUDQUERY_API_KEY", token)
```
disq, line 157: This looks scary at first glance but it's only done for `cqpd_` tokens, where the user has manually created and copy-pasted this into their terminal/env.

erezrokah (approved):

PR #23046, opened 2026-07-03
feat: Store nested JSON timestamps as BSON dates

murarustefaan:
Adds support to the mongodb destination plugin for using the timestamp types in nested fields.


erezrokah (approved):
Can we also run a test sync on `k8s_core_pods` to make sure this works on `k8s_core_pods -> status_container_statuses.state.terminated.finishedAt`?

murarustefaan:
Already did!

PR #23119, opened 2026-07-16
feat: Add obfuscate_columns_except transformer

murarustefaan:
Opt-in field redaction, for syncing into environments where every field has to be explicitly approved before it's allowed out unredacted.

`obfuscate_columns` only does the opposite — you list what to hide — so an approve-first workflow means enumerating every sensitive field and then keeping that list current forever as the source schema grows. On something the size of the K8s schema that isn't maintainable.

So: `obfuscate_columns_except`. You list what to keep, everything else is handled according to `unmatched`.

```yaml
- kind: obfuscate_columns_except
  tables: ["k8s_core_pods"]
  unmatched: drop
  columns:
    - uid
    - context
    - namespace
    - status_phase
    - "spec_containers.#.image"
    - "spec_init_containers.#.image"
    - "status_container_statuses.#.state.terminated.finishedAt"
```

Against a real cluster that gives you:

```json
{
  "uid": "10377ba4-fcb1-456f-9a47-3c8cf93424ca",
  "context": "prod-1",
  "namespace": "vault",
  "status_phase": "Succeeded",
  "spec_containers": [ { "image": "hashicorp/vault:2.0.3" } ],
  "spec_init_containers": [],
  "status_container_statuses": [
    { "state": { "terminated": { "finishedAt": "2026-07-26T00:00:08Z" } } }
  ]
}
```

## unmatched

`drop` (default) removes non-approved fields entirely. `collapse` replaces each non-approved object or array with one marker. `redact` puts a marker on every individual leaf.

I built `redact` first, since "redact everything else" is the obvious reading of the ticket. Then I ran it against a real cluster and it produced about 8,000 `Redacted by CloudQuery` strings in a single column of a single table — 900KB of markers wrapped around 10KB of data anyone actually approved. Hence the other two modes. Numbers from 118 pods:

| unmatched | pods columns | spec_containers | markers |
|---|---|---|---|
| `drop` | 11 | 10 KB | 0 |
| `collapse` | 59 | 197 KB | 1,758 |
| `redact` | 59 | 923 KB | 8,515 |

`drop` also avoids a typing problem, below.

## Stuff worth knowing before you approve

**Primary keys.** A non-approved PK column loses whatever made the row unique, which quietly breaks upserts at the destination — under `drop` the column is gone, and with `include_sha: false` every row redacts to the same string. Rather than let that happen silently the transformer refuses to start and tells you to allowlist the column. For the K8s tables that means `uid`. Verified against the MongoDB destination: two consecutive syncs, 129 pods, 129 distinct uids, no collisions.

**Typed columns are never stringified.** If a non-approved root column is a bool/int/timestamp/inet/array we drop it rather than stuffing a redaction string into it, so the schema never claims a type the data doesn't have. Checked against Postgres — `k8s_core_pods` has 19 typed root columns and all 19 are gone in every mode, with only the `_cq_*` internals surviving. Approved typed columns come through untouched (`generation` bigint = 1, `spec_host_network` boolean = false, `status_pod_ip` inet = 10.42.2.2/32).

Nested JSON leaves are the exception, because their types travel in `TypeSchema` metadata rather than the Arrow schema. A redacted nested timestamp becomes a string, and under `collapse` a whole nested object becomes a string. The MongoDB destination handles that fine — it falls back to storing a string — and approved nested fields still convert properly (`finishedAt` lands as a real BSON Date in both `drop` and `redact`). Worth knowing if you're pointing this at some other typed destination.

**Don't forget the `#`.** Paths use the usual gjson syntax, so anything crossing an array needs `spec_containers.#.image`, not `spec_containers.image`. It's easy to write the second one by hand, and there's no error if you do — the path just matches nothing and the field you meant to keep gets dropped along with everything else. If you name a top-level column that doesn't exist you do get a hard error, so that case is covered.

## Also in here

`include_sha: false` drops the hash from the redaction marker, for setups that don't want it visible in the destination. Works on all three obfuscate kinds. Default stays `true`, and I checked the existing kinds produce byte-identical output to main.

Table selection is still the source's job. This only redacts inside tables you point it at, so `tables:` on the source has to be locked down separately.

## Testing

101 unit tests, plus end-to-end syncs from a live K8s cluster into both Postgres and MongoDB in all three modes. Backward compat checked by building the transformer from main and diffing output for the existing kinds — identical hashes and schema.

Left as follow-ups: there are two JSON traversal implementations in here now (gjson/sjson for the existing kinds, a tree walk for this one, since gjson can't express "everything except"), and the allowlist gets re-parsed per batch.


erezrokah (approved):

murarustefaan:
I'll ship this as an RC first.

PR #23162, opened 2026-07-26
fix: resolve 15 bugs across CLI and plugins

praneshnikhar:
**CLI bugs:**
- **Nil pointer dereference** (`cli/internal/publish/plugins.go`): `req.Header.Add` calls placed before the `err != nil` check from `http.NewRequestWithContext` — if URL parsing fails or context is cancelled, this panics instead of returning an error.
- **Goroutine leaks** (`cli/internal/transformerpipeline/pipeline.go`): Two unbuffered channel patterns where spawned goroutines block forever after the pipeline closes — the `Recv` goroutine in `startBlocking` and the `Send` goroutine in `Send`. Fixed by making channels buffered (size 1) and adding `isClosed` checks.
- **Zero metrics sent to analytics** (`cli/cmd/sync_v3.go`): The `mt metrics.Metrics` variable was declared but never populated — all sync metric payloads were sent as zeros. Now built from `statsPerTable` data.
- **Wrong transformer option** (`cli/cmd/sync_v2.go:34`): `WithRemovePKs()` called twice instead of `WithRemovePKs()` + `WithRemoveUniqueConstraints()` in the v2→v3 append-mode path — unique constraints were silently preserved when they should be removed.
- **Data races** (`cli/internal/analytics/client.go`): Global `client` and `cachedSyncEventDetails` variables accessed from multiple goroutines without synchronization. Added `sync.RWMutex` protection.
- **Dead code** (`cli/internal/publish/plugins.go`): Redundant second `os.ReadFile` that can never trigger (the file was already confirmed to exist by the first read).

**Plugin bugs:**
- **SQL injection** (ClickHouse `queries/table.go`, PostgreSQL `migrate.go`, DuckDB+SQLite `migrate.go`): Table/database names and TTL strings interpolated directly into SQL queries — now sanitized with `util.SanitizeID`, `pgx.Identifier`, or parameterized queries.
- **Nil dereference** (ClickHouse `client/table.go`): `*uint8` result dereferenced without nil check in `equalTTLs`. Also used `context.Background()` instead of the caller's context.
- **panic in production** (MongoDB `write.go`): `panic(err)` used instead of error returns when JSON unmarshal fails in `transformArr`. Replaced with proper error propagation through the call chain.
- **Nil tx rollback** (PostgreSQL `migrate.go`): Deferred `tx.Rollback(ctx)` on nil transaction when `BeginTx` fails mid-function (CockroachDB path). Added nil guard.
- **Resource leak** (GCS `write.go`): `*storage.Writer` not closed on `WriteHeader`/`WriteContent`/`WriteFooter` errors, and `w.Close()` on nil writer when message channel is empty.


erezrokah (changes requested):
Hi @praneshnikhar, thank you for your contribution. Can you please open a bug report with a reproduction per issue fixed in this PR? This will help us understand the impact per issue and prioritize the review accordingly

praneshnikhar:
suree! @erezrokah

praneshnikhar:
> Hi @praneshnikhar, thank you for your contribution. Can you please open a bug report with a reproduction per issue fixed in this PR? This will help us understand the impact per issue and prioritize the review accordingly

Done — opened issues #23164 through #23173, each with reproduction steps. Let me know if you need anything else!

PR #23174, opened 2026-07-28
fix: Prevent wiping out staged files during concurrent writes

bbernays:
#### Summary


plugins/destination/snowflake/client/write.go
```diff
@@ -155,6 +155,10 @@ func insertColumnsList(table *schema.Table) string {
 func (c *Client) setupWrite(ctx context.Context) error {
 	var setupErr error
 	c.setupWriteOnce.Do(func() {
+		if c.spec.SkipWriteSetup {
```
erezrokah, line 158: Can't we inject a random ID instead per plugin init? That would ensure they don't conflict.

I also suggest adding a concurrency load test similar to https://github.com/cloudquery/cloudquery/pull/20907/changes#diff-16b35c7fa05984f0e08cef54e2b7f6a4079474b4e6eb0a5a93e333ca54ed57e1R235

erezrokah: There's also `CREATE TEMPORARY STAGE` scoped to a session not sure if we can use it here

bbernays: I don't like the random-id per plugin because that will mean the stages will grow unbounded. `TEMPORARY` stage seems like a great solution. If we can validate that the CREATE/REPLACE is in fact the issue

bbernays: Actually the syntax seems to support `IF NOT EXISTS` which I think is the cleanest short term solution

PR #23228, opened 2026-08-03
fix: Update arrow-go to v18.7.0 and preserve int64 precision in nested values

erezrokah:
Supersedes #23163.

arrow-go v18.7.0 made `array.FromJSON` decode int64/uint64 exactly, but `AppendValueFromString` still routes nested values through float64, so list/struct round-trips lose precision and every Go destination's write test suite fails.

Route nested types through `UnmarshalOne` with `UseNumber` in each destination read path, and correct two float64-rounded literals in the transformer/basic test.

Blocked on cloudquery/plugin-sdk#2553 — the SDK's own test helpers hit the same lossy path, so CI here stays red until that ships. The file/s3/gcs/azblob destinations additionally need cloudquery/filetypes#757.


erezrokah:
Moved the fix into #23163 instead.

PR #23294, opened 2026-08-31
feat: Enable opt in usage of Copy rather than bulk inserts

bbernays:
#### Summary


erezrokah (commented):
Review comments

| # | Location | Severity | Issue | Fix |
|---|----------|----------|-------|-----|
| 1 | `client/copy_from.go:71` | High | Each ineligible group calls `insertBatchExec` separately, so one pipelined `SendBatch` per mixed batch becomes one per table. 10k-row batch over ~300 tables = ~300 sequential round trips instead of 1, and zero rows use COPY. Opposite of the PR's goal on remote DBs. | Accumulate ineligible groups' messages, call `insertBatchExec` once after the loop. |
| 2 | `client/copy_from.go:145` | High | `distinct on` dedupes on `table.PrimaryKeys()` (source schema) but `on conflict on constraint` targets `table.PkConstraintName` (database). On mismatch the merge aborts with `ON CONFLICT DO UPDATE command cannot affect row a second time` (21000) where the INSERT path succeeds. Repro: table created with PK `(account_id, arn)`, source moves to `_cq_id`-only PK, user syncs `--no-migrate`. | Dedupe on `pkConstraintDetails.columns` (already populated, currently unused), or make `eligible()` require it to equal `table.PrimaryKeys()`. |
| 3 | `client/copy_from.go:87` | High | **RLS**: Postgres refuses `COPY FROM` on a table with row-level security enabled. The PK-less direct-copy path hits this and fails the sync outright. Staging path is unaffected (temp table has no RLS, merge is a real INSERT). | Detect `relrowsecurity` in `listTables` and exclude those tables from `eligible()`, or document the incompatibility. |
| 4 | `client/copy_from_test.go:107` | Medium | No test asserts COPY actually ran. `TestCopyFrom_DuplicatePrimaryKeysInOneBatch` — the stated regression test for `distinct on` — would still pass with `distinct on` deleted if the writer ever splits the batch below `minCopyRows`. | Count `copyTable` invocations, or use `forceCopyFrom` with small row counts and assert the path. |
| 5 | `client/copy_from.go:87` | Medium | **Rules**: `COPY FROM` does not invoke rewrite rules. PK-less tables now copy straight into the target, so any `INSERT` rule on them silently stops firing. | Document alongside the existing trigger caveat in `overview.md`. |
| 6 | `client/copy_from.go:117` | Medium | Fresh UUID staging name per batch makes all three staging statements unique SQL. Under pgx's default `QueryExecModeCacheStatement` each batch adds 3 never-reusable prepared statements, churning the 512-entry LRU and evicting statements that are reused. | Use `QueryExecModeExec` for the staging statements, or a deterministic per-table name. |
| 7 | `docs/overview.md:78` | Medium | "Batches smaller than 100 rows use `INSERT` regardless" is wrong — the threshold is per table group within a batch. Users with `batch_size: 10000` will expect COPY and get INSERT everywhere. | Reword to "fewer than 100 rows for a given table within one batch". |
| 8 | `client/spec/spec.go:57` | Low | `useCopyFrom()` returns false whenever `pgvector_config` is set, but spec comment, `schema.json` and `overview.md` list only CockroachDB and CrateDB as exclusions. Flag is silently inert with no log line. | Add pgvector to both descriptions; log once at startup when the flag is inert. |
| 9 | `client/copy_from.go:118` | Low | Comment claims "LIKE without INCLUDING carries no constraints". Postgres always copies `NOT NULL` with `LIKE`; only CHECK, defaults, indexes and identity need `INCLUDING`. Conclusion drawn is right, reason is wrong. | Fix or drop — repo convention is no code comments unless asked, and this PR adds many. |
| 10 | `client/copy_from.go:18` | Low | `minCopyRows` is a package var that `forceCopyFrom` rewrites to 0. Global mutable state shared by every test in the package; breaks under `t.Parallel()` and trips `-race` (used by `make test`). | Move to a `Client` field defaulted from the spec. |

erezrokah (commented):
Also we should be aware of the difference in approaches, maybe document them

| Behavior | INSERT | COPY FROM |
|----------|--------|-----------|
| Type / format validation | yes | yes |
| `NOT NULL` | yes | yes |
| `CHECK`, domain constraints | yes | yes |
| `PRIMARY KEY` / `UNIQUE` | yes | yes |
| `FOREIGN KEY` | yes | yes |
| Column `DEFAULT` (omitted columns) | yes | yes |
| `BEFORE`/`AFTER ROW` triggers | fire per row | fire per row |
| Statement-level triggers | once per statement | once per COPY |
| `ON CONFLICT` | yes | **no** — why this PR needs a staging table |
| `RETURNING` | yes | no |
| Rewrite rules | applied | **not applied** |
| Row-level security | works | **errors out** |
| Per-row expressions | yes | no, literal values only |
| Wire cost | parse/bind/execute per statement | one statement, rows stream as CopyData |
| Value format | negotiated per type | pgx forces binary |
| Failure granularity | implicit txn per batch, all-or-nothing | one bad row aborts whole COPY |

disq (approved):
You could switch to `batchwriter` (instead of `mixedbatchwriter` - depending on the spec option specified of course) and get rid of most of the grouping code, but then you can't `reset()` nor fallback to mixed inserts. Good as a first release, if it tracks we can improve on it.

PR #23331, opened 2026-09-02
fix: Release CLI and Scaffold without GoReleaser Pro

erezrokah:
Drops the expired GoReleaser Pro licence from the CLI and Scaffold releases by merging each project's Pro-only `monorepo`/`includes` config into a single OSS-compatible file and letting release-please keep the GitHub release while the workflow uploads the artifacts, leaving tag names, asset names, Docker tags and Homebrew formulae unchanged.


erezrokah:
Closing, easier to use the paid version

PR #23361, opened 2026-09-18
feat: Attach hashed cloud account ids to sync_run_completed

erezrokah:
Behind the cq-cli-account-id-analytics flag, collects the distinct account identifier of each synced row for the aws, gcp, azure, k8s, github and cloudflare sources and attaches hashes of them, keyed on the per-team salt served by the API, with an exact distinct count, to the sync_run_completed event.


cli/internal/analytics/accounts.go
```diff
@@ -0,0 +1,161 @@
...
+	hashLength = 16
+)
+
+type columnSpec struct {
+	dimension string
+	table     string
+	column    string
+}
+
+var idColumnsBySource = map[string][]columnSpec{
```
erezrokah, line 30: We can start with this and then maybe move to the plugins level?

cli/internal/analytics/accounts.go
```diff
@@ -0,0 +1,161 @@
...
+	"k8s":        {{dimension: AccountDimension, column: "cloud_cluster_id"}},
+	"cloudflare": {{dimension: AccountDimension, column: "account_id"}},
+	"github": {
+		{dimension: AccountDimension, column: "org"},
+		{dimension: RepositoryDimension, table: "github_repositories", column: "id"},
+	},
+}
+
+type IDSummary struct {
+	Hashes    []string
```
bbernays: Can you confirm that we are not going to be collecting actual hashes?

For AWS account IDs, the level of effort to de-anonymize hashes is not insurmountable as AccountIDs are a limited finite set of

bbernays:
Does this need to be a new major version? This significantly deviates from our docs in terms of the data we collect by default: https://www.cloudquery.io/docs/cli/advanced/telemetry

cli/internal/analytics/accounts.go
erezrokah: Going to add a backend generated per team analytics salt to help with this, moved the PR to draft for now

cli/cmd/sync_v3.go
```diff
@@ -176,6 +177,7 @@ func syncConnectionV3(ctx context.Context, syncOptions syncV3Options) (syncErr e
 		totalResources = int64(0)
 		totals         = sourceClient.Metrics()
 		statsPerTable  = utils.NewConcurrentMap[string, SyncRunTableProgressValue]()
+		idCollector    = newIDCollector(ctx, sourceSpec.Path, destinationSpecs)
```
disq, line 180: This can result in a `nil` idCollector which gets deref'ed below.

cli/cmd/sync_v3.go
```diff
@@ -862,3 +866,30 @@ func getTransformedTableNameFromSchema(transformedSchemaBytes []byte) (string, e
...
+		UserID:      userID,
+		Team:        team,
+		Environment: environment,
+		CLIVersion:  Version,
+	})
+	if !enabled {
+		return nil
+	}
+
+	return analytics.NewIDCollector(sourcePath, analytics.TeamAnalyticsSalt(ctx, team))
```
disq, line 894: This can also end up in a `nil` return (if `salt` is empty for some reason, or if `!ok`)

cli/cmd/sync_v3.go
erezrokah: Added https://github.com/cloudquery/cloudquery/pull/23361/changes/b7962c8fdb06dd22de2f34c621ac3bcd784d1f73

cli/cmd/sync_v3.go
erezrokah: Added https://github.com/cloudquery/cloudquery/pull/23361/changes/b7962c8fdb06dd22de2f34c621ac3bcd784d1f73

erezrokah:
@greptileai review

disq (approved):

erezrokah:
> Does this need to be a new major version? This significantly deviates from our docs in terms of the data we collect by default: https://www.cloudquery.io/docs/cli/advanced/telemetry

We decided not to make this a major version, it's gate by current telemetry controls, we'll update the docs and share a message to customers before the enabling the change

